PR #1, opened 2023-05-16
move all sources into a cargo workspace

SuperFluffy:
this steps just moves everything out of `sources/` and places it into `crates/`. `cargo check` passes after adding making the crates workspace members in the root `Cargo.toml`.

The crates do not refer to each other by path, and they are also not yet cleaned up.


joroshiba (approved):

joroshiba (commented):
Should we leave the existing github workflows until we have consolidated?

joroshiba (approved):

PR #5, opened 2023-05-17
basic chain using tower-abci and penumbra stack

noot:
Implements a basic ABCI application that communicates with a tendermint process. The application only has account logic at this stage, and only one transaction type `accounts::transaction::Transaction` which represents a value transfer. It uses `penumbra_storage::Storage` as the state storage, and `penumbra_component::Component` to define an application component (in this case, there is only the `AccountsComponent`).

It uses the `tower_abci::Service` as the base layer which talks to tendermint. The application logic is built up around the requests made by tendermint. 

Currently, there are 3 accounts that have a hard-coded balance of 10e18 at genesis: `alice`, `bob`, and `carol`.

Usage:

```
cd crates/astria-sequencer && cargo build && ../../target/debug/astria-sequencer --genesis-file=test-genesis.json
```

Clone and build https://github.com/astriaorg/cometbft  (`make install && make install_abci`)
```
cometbft init
cometbft node
```

You can use `abci-cli` to query the application, for example, to query Alice's nonce: (note that the `0xdeadbeef` is totally arbitrary, abci-cli requires some hex string as an arg :/)

```sh
$ abci-cli query --path=accounts/nonce/alice 0xdeadbeef
I[2023-05-18|22:21:48.105] service start                                module=abci-client msg="Starting socketClient service" impl=socketClient
-> code: OK
-> height: 6
-> key: accounts/nonce/alice
-> key.hex: 6163636F756E74732F6E6F6E63652F616C696365
-> value: 
-> value.hex: 01000000
```

You can also curl endpoints directly:
```sh
$ curl http://localhost:26657/abci_info
{"jsonrpc":"2.0","id":-1,"result":{"response":{"data":"astria_sequencer","version":"0.1.0","app_version":"1","last_block_height":"1"}}}
```

After node has started, submit a tx:
```
$ curl http://localhost:26657/broadcast_tx_commit?tx=0x0003000000626f6205000000616c6963651516050000000000000000000000000001000000
{"jsonrpc":"2.0","id":-1,"result":{"check_tx":{"code":0,"data":null,"log":"","info":"","gas_wanted":"0","gas_used":"0","events":[],"codespace":"","sender":"","priority":"0","mempoolError":""},"deliver_tx":{"code":0,"data":null,"log":"","info":"","gas_wanted":"0","gas_used":"0","events":[],"codespace":""},"hash":"5CD86628A864412D505C5B575EE939F7B60AE507119A0F63749AFD2384B8BAEE","height":"8"}}
```

Note: the tx was generated with the test `test_transaction`; will need to make a prettier CLI for this

Inside cometbft you'll see the tx was included:
```
I[2023-05-17|22:06:46.600] received proposal                            module=consensus proposal="Proposal{2/0 (7640BDA8191E35BD9028FE6913D9C831ED095782B087E45C90905FDD476785FA:1:214894D427A1, -1) 94CE5F23AEE1 @ 2023-05-18T02:06:46.598356671Z}"
I[2023-05-17|22:06:46.601] received complete proposal block             module=consensus height=2 hash=7640BDA8191E35BD9028FE6913D9C831ED095782B087E45C90905FDD476785FA
I[2023-05-17|22:06:46.606] finalizing commit of block                   module=consensus height=2 hash=7640BDA8191E35BD9028FE6913D9C831ED095782B087E45C90905FDD476785FA root=35EE717EA57E28688B0DE0653C38247AAD65BF943D00EF3ADAF5023D2F1FF0A2 num_txs=1
I[2023-05-17|22:06:46.698] executed block                               module=state height=2 num_valid_txs=1 num_invalid_txs=0
I[2023-05-17|22:06:51.750] committed state                              module=state height=2 num_txs=1 app_hash=A0E6C663281AB4E36F7DDC647D192D502F2EE481A4CDE6D8C45B2B51658AD7DD
I[2023-05-17|22:06:51.753] indexed block exents                         module=txindex height=2
```

You'll also see a log inside the sequencer app:
```
2023-05-19T02:21:50.047549Z  INFO astria_sequencer::app: executed transaction tx=AccountsTransaction(Transaction { to: "bob", from: "alice", amount: 333333, nonce: 1 }) height=8
```

Query Alice's nonce again
```sh
$ abci-cli query --path=accounts/nonce/alice 0x00
I[2023-05-18|22:26:50.206] service start                                module=abci-client msg="Starting socketClient service" impl=socketClient
-> code: OK
-> height: 62
-> key: accounts/nonce/alice
-> key.hex: 6163636F756E74732F6E6F6E63652F616C696365
-> value: 
-> value.hex: 0101000000
```

Nonce should increase after the tx was sent :D (note the value is borsh-serialized, so doesn't correspond to the raw value exactly)

You can set the path to `accounts/balance/(alice, bob, carol)` to check balances also


crates/astria-sequencer/src/app.rs
```diff
@@ -0,0 +1,195 @@
...
+    // Invariant: state_tx and self.state are the only two references to the
+    // inter-block state.
+    fn apply(&mut self, state_tx: StateDelta<InterBlockState>) -> Vec<Event> {
+        let (state2, mut cache) = state_tx.flatten();
+        std::mem::drop(state2);
+        // Now there is only one reference to the inter-block state: self.state
+
+        let events = cache.take_events();
+        cache.apply_to(
+            Arc::get_mut(&mut self.state).expect("no other references to inter-block state"),
```
SuperFluffy, line 182: Let's propagate errors like this upstream and handle them in the event loop where `apply` is called (i.e. try to avoid panicking deeply inside execution logic).

crates/astria-sequencer/src/consensus.rs
```diff
@@ -0,0 +1,144 @@
...
+        self.app
+            .deliver_tx(&deliver_tx.tx)
+            .await
+            .unwrap_or_else(|e| {
+                // we don't want to panic on failing to deliver_tx as that would crash the entire
+                // node
+                tracing::error!(error = ?e, "deliver_tx failed");
+                vec![]
+            });
+        Ok(Default::default())
```
SuperFluffy, line 123: Should this not return a `BoxError` if `self.app.deliver_tx` fails?

SuperFluffy (changes requested):
Going to submit the first review now before continuing on.

Some high level observations:

+ I think this PR should use more of penumbra provided APIs;
+ This should have more instrumentation to understand how logic flows through the program.
+ The `expect` statements should replaced by error handling so that the program does not get taken down the moment something goes awry (except during early setup, where it's fine). Reading the abci spec I notice that it says that certain operations, if failed, should crash the program. I am not sure how I feel about crashing. I would favour non-zero exit statuses instead.

crates/astria-sequencer/src/app.rs
```diff
@@ -0,0 +1,195 @@
...
+    },
+    transaction::Transaction,
+};
+
+/// The application hash, used to verify the application state.
+/// TODO: this may not be the same as the state root hash?
+pub type AppHash = penumbra_storage::RootHash;
+
+/// The inter-block state being written to by the application.
+type InterBlockState = Arc<StateDelta<Snapshot>>;
```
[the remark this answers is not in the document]
noot: this was mirrored from penumbra: https://github.com/penumbra-zone/penumbra/blob/9cc2c644e05c61d21fdc7b507b96016ba6b9a935/app/src/app/mod.rs#L35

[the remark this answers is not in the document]
noot: I believe it's wrapped in an `Arc` because multiple components need to take ownership of it within one function to mutate it (ie write to the `StateDelta`)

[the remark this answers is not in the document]
noot: it is also relatively short-lived; it only exists within this current block, at the end of the block it gets replaced (see `App.commit`)

```
```

crates/astria-sequencer/src/app.rs
```diff
@@ -0,0 +1,195 @@
...
+
+/// The genesis state for the application.
+#[derive(Clone, Debug, serde::Deserialize, Default)]
+pub struct GenesisState {
+    pub accounts: Vec<(String, u64)>,
+}
+
+/// The Sequencer application, written as a bundle of [`Component`]s.
+#[derive(Clone)]
+pub struct App {
```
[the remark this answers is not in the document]
noot: I kind of like `App` since it's the application logic :p this seems to be how it's referred to in ABCI as well. I think `Sequencer` should refer to the binary logic (ie what's run inside `main` like you mentioned in a previous comment?)

.cargo/config.toml
```diff
@@ -0,0 +1,2 @@
+[build]
+rustflags = ["--cfg", "tokio_unstable"]
```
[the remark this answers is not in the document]
noot: the build breaks without this :/
```
error[E0433]: failed to resolve: could not find `Builder` in `task`
  --> /home/e/.cargo/git/checkouts/penumbra-54633d8978d5e293/9f33f72/storage/src/snapshot.rs:63:22
   |
63 |         tokio::task::Builder::new()
   |                      ^^^^^^^ could not find `Builder` in `task`
   |
help: consider importing one of these items
   |
1  | use std::thread::Builder;
   |
1  | use tempfile::Builder;
   |
1  | use tokio::runtime::Builder;
   |
help: if you import `Builder`, refer to it directly
   |
63 -         tokio::task::Builder::new()
63 +         Builder::new()
```

penumbra has this as well, I just took it from there

crates/astria-sequencer/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+hex = "0.4"
+penumbra-component = { git = "https://github.com/penumbra-zone/penumbra.git", rev = "9f33f728dab2312a92c619864a1be97fb81882d9" }
+penumbra-storage = { git = "https://github.com/penumbra-zone/penumbra.git", rev = "9f33f728dab2312a92c619864a1be97fb81882d9" }
+serde = { version = "1", features = ["derive"] }
+serde_json = "1"
+tendermint-proto = "0.29"
+tendermint = "0.29"
+tokio = { version = "1.28", features = [ "tracing" ] }
+tower = "0.4"
+tower-abci = { git = "https://github.com/penumbra-zone/tower-abci.git", rev = "11ecad76b3bc527dcc2d1652bb7fcaa6cb52046d" }
```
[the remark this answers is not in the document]
noot: updated

crates/astria-sequencer/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "astria-sequencer"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+anyhow = "1"
```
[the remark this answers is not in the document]
noot: I was kind of forced to use anyhow to use the penumbra traits, eg. if I use eyre/color-eyre:
```
method `end_epoch` has an incompatible type for trait
expected signature `fn(&'life0 mut Arc<S>) -> Pin<Box<(dyn futures::Future<Output = Result<(), anyhow::Error>> + std::marker::Send + 'async_trait)>>`
   found signature `fn(&'life0 mut Arc<S>) -> Pin<Box<(dyn futures::Future<Output = Result<(), ErrReport>> + std::marker::Send + 'async_trait)>>`rustc[Click for full compiler diagnostic](rust-analyzer-diagnostics-view:/diagnostic%20message%20%5B0%5D?0#file%3A%2F%2F%2Fhome%2Fe%2Fastria%2Fcrates%2Fastria-sequencer%2Fsrc%2Faccounts%2Fcomponent.rs)
component.rs(41, 5): change the output type to match the trait: `Pin<Box<(dyn futures::Future<Output = Result<(), anyhow::Error>> + std::marker::Send + 'async_trait)>>`
```

this is in `impl Component for AccountsComponent` for example, but shows up other places. do you know a good way around this? I agree it would be better to be consistent

crates/astria-sequencer/src/bin/app.rs
```diff
@@ -0,0 +1,54 @@
...
+    tracing_subscriber::fmt()
+        .with_env_filter(
+            EnvFilter::try_from_default_env().unwrap_or_else(|_| EnvFilter::new("info")),
+        )
+        .init();
+
+    let storage = penumbra_storage::TempStorage::new()
+        .await
+        .expect("should create temp storage");
+    let snapshot = storage.snapshot(u64::MAX).expect("should create snapshot");
```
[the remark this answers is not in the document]
noot: updated!

.cargo/config.toml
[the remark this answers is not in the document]
SuperFluffy: Cool, thank you for pointing this out. I dug through penumbra's code and this is from their tokio console integration.

crates/astria-sequencer/Cargo.toml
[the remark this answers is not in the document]
SuperFluffy: No way around this. Very unfortunate that they hard coded this into their public interfaces. :(

crates/astria-sequencer/src/consensus.rs
noot: I think I changed this because a failing `deliver_tx` (ie meaning the tx execution failed) would cause the entire blockchain to halt, as it wouldn't be able to make a block. I'll think about how to fix that, it should still do something different on error

crates/astria-sequencer/src/app.rs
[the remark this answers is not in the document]
SuperFluffy: Gotcha. Thank you for pointing to that. This design surprised me because usage `StateDelta` through `ArcStateDeltaExt` requires exclusive access. Since there are no other protections (like Mutex) it seems like a big burden on the developer to ensure that weak/strong references are always dropped in time.

crates/astria-sequencer/src/app.rs
```diff
@@ -0,0 +1,195 @@
...
+        let jmt_root = storage
+            .commit(state)
+            .await
+            .expect("must be able to successfully commit to storage");
+
+        let app_hash: AppHash = jmt_root;
+        tracing::debug!(?app_hash, "finished committing state");
+
+        // Get the latest version of the state, now that we've committed it.
+        self.state = Arc::new(StateDelta::new(storage.latest_snapshot()));
```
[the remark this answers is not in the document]
SuperFluffy: Yes, that makes sense.

crates/astria-sequencer/src/consensus.rs
SuperFluffy: I see - note that this makes it an infallible operation (as are many of the other async methods in this impl block).

I think I am fine with these functions being handlers that only return the final result `T` (as opposed to a `Result<T, E>` where, if `Err(e)` is encountered, the event loop calling `deliver_tx` makes a decision on what to return).

The only case where this could be different is if an invariant is violated and the event loop should exit.

crates/astria-sequencer/src/bin/app.rs
```diff
@@ -0,0 +1,18 @@
+use astria_sequencer::sequencer::Sequencer;
```
SuperFluffy, line 1: High level comment on this: let's `mv bin/app.rs main.rs` because the main point of this crate is to provide an executable.

crates/astria-sequencer/Cargo.toml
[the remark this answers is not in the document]
noot: yeah :/ the alternative is that we don't use their `Component` trait and define our own, although I'm not sure if that'll break compatibility with them in the future. for now that seems fine though, we don't need to strictly follow their `Component` trait

crates/astria-sequencer/src/bin/app.rs
```diff
@@ -0,0 +1,54 @@
...
+};
+use tower_abci::Server;
+use tracing::info;
+use tracing_subscriber::EnvFilter;
+
+pub const DEFAULT_LISTEN_ADDR: &str = "127.0.0.1:26658";
+
+#[tokio::main]
+async fn main() {
+    tracing_subscriber::fmt()
```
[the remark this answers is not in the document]
noot: is `init_env_filter` the main functionality I'd need? or should I add `init` as well?

crates/astria-sequencer/src/app.rs
noot: I think the idea here is that this will never panic unless there's developer error, should I still update it?

crates/astria-sequencer/src/bin/app.rs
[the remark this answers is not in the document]
SuperFluffy: You need the whole thing. The reason is that tracing_subscriber escapes rustc capturing stdout. So in order to properly test the service you need have a way to write the trace events to stdout or hide them.

crates/astria-sequencer/src/accounts.rs
```diff
@@ -0,0 +1,5 @@
+pub mod component;
```
SuperFluffy, line 1: Note on the name of `accounts.rs`: prefer to `mv accounts.rs accounts/mod.rs`. This is because `accounts.rs` might be very far away from `accounts/` when the directory strcuture which is very confusing to navigate (like on github). Many projects prefer `mod.rs` for that reason.

crates/astria-sequencer/src/accounts/accounts.rs
```diff
@@ -0,0 +1,133 @@
...
+pub const ACCOUNTS_PREFIX: &str = "accounts";
+
+fn storage_key(address: &str) -> String {
+    format!("{}/{}", ACCOUNTS_PREFIX, address)
+}
+
+pub struct AccountsComponent {}
+
+impl AccountsComponent {
+    pub fn new() -> Self {
```
SuperFluffy, line 29: Marker types don't need a constructor

crates/astria-sequencer/src/accounts/accounts.rs
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    pub async fn execute<S: StateWrite>(&self, state: &mut S) -> Result<()> {
+        let from_balance = state
+            .get_raw(&storage_key(&self.from))
+            .await?
+            .unwrap_or([0u8; 8].to_vec());
+        let from_balance = u64::from_be_bytes(from_balance.as_slice().try_into()?);
+        let to_balance = state
+            .get_raw(&storage_key(&self.to))
```
SuperFluffy, line 117: why not pre-compute the storage key in all instances during transaction construction?

crates/astria-sequencer/src/app.rs
SuperFluffy: We should make a note a the top of `app` that states this. Looking at the penumbra code, it seems to me that the main motivation to wrap all of this inside `Arc` is to be found here: https://github.com/penumbra-zone/penumbra/blob/07d93f77520cb1ed777c10077ea73d54f532a3fa/crates/core/app/src/app/mod.rs#L219-L233

In other words: they use `Arc` to do (potentially threaded) computations very deeply into mutating state, right before committing it. Except for `check_stateless` and `check_stateful` all other execution is strictly sequential (not even concurrent but really sequential) to the point of it crashing the program because otherwise the invariant is violated.

I have not found evidence to the contrary, but I might have missed it.

noot: will add a note. `check_stateless` is for a transaction, it doesn't actually deal with the `Arc<StateDelta>` at all. but yeah when the tx is executed or other state changes are made, they all appear to be sequentlal

crates/astria-sequencer/src/accounts/accounts.rs
noot: you mean when `accounts::Transaction` is constructed? blockchain transactions don't usually contain raw storage keys, transactions also get gossiped around the network so that would increase bandwidth, overall seems like something that isn't really done

crates/astria-sequencer/src/accounts/accounts.rs
noot: updated (and the above)

SuperFluffy (commented):
I know this is going to take more work in this PR, but I would love to have all methods in this PR instrumented to make sure we don't need a big refactor in the future.

See this as a very strong recommendation, but I am ok if you punt this for this PR and do it in a follow-up PR.


What I would block this PR on is testing: can we define a minimal test where a request is sent to the system and then read? It's fine if it just tests the happy path.

crates/astria-sequencer/src/sequencer.rs
```diff
@@ -0,0 +1,70 @@
...
+    app::App,
+    consensus::ConsensusService,
+    info::InfoService,
+    mempool::MempoolService,
+    snapshot::SnapshotService,
+};
+
+pub struct Sequencer {
+    #[allow(clippy::type_complexity)]
+    server: Server<
```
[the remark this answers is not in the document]
noot: what about getting rid of `Sequencer::new` and just having `Sequencer::run`?

crates/astria-sequencer/src/info.rs
```diff
@@ -0,0 +1,139 @@
...
+
+    Ok(response::Query {
+        key: key.into(),
+        value: value.into(),
+        height: Height::from(height as u32),
+        ..Default::default()
+    })
+}
+
+pub enum Query {
```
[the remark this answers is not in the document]
noot: yes, same with `Transaction`, will add

crates/astria-sequencer/src/mempool.rs
```diff
@@ -0,0 +1,92 @@
...
+                // TODO: status codes for various errors
+                match Transaction::from_bytes(&req.tx) {
+                    Ok(tx) => match tx.check_stateless() {
+                        Ok(_) => Poll::Ready(Ok(MempoolResponse::CheckTx(response::CheckTx {
+                            code: 0.into(),
+                            ..Default::default()
+                        }))),
+                        Err(e) => Poll::Ready(Ok(MempoolResponse::CheckTx(response::CheckTx {
+                            code: 1.into(),
+                            log: format!("{:?}", e),
```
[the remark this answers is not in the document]
noot: right, is there a better way to return a cleaned-up error?

crates/astria-sequencer/src/snapshot.rs
```diff
@@ -0,0 +1,73 @@
...
+impl Service<SnapshotRequest> for SnapshotService {
+    type Error = BoxError;
+    type Future = SnapshotServiceFuture;
+    type Response = SnapshotResponse;
+
+    fn poll_ready(&mut self, _cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        Poll::Ready(Ok(()))
+    }
+
+    fn call(&mut self, req: SnapshotRequest) -> Self::Future {
```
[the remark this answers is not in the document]
noot: yeah, if we need to fork tower-abci (and maybe change the anyhow to eyre also) let's open an issue and do in a follow-up

crates/astria-sequencer/src/transaction.rs
```diff
@@ -0,0 +1,108 @@
...
+        amount: Balance,
+        nonce: Nonce,
+    ) -> Self {
+        Self::AccountsTransaction(AccountsTransaction::new(to, from, amount, nonce))
+    }
+
+    pub fn to_bytes(&self) -> Result<Vec<u8>> {
+        let bytes = match self {
+            Self::AccountsTransaction(tx) => {
+                let mut bytes = vec![0u8];
```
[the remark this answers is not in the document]
noot: I was prepending the enum index, but yeah we should probably use a legit encoding scheme here, does protobuf handle rust enums? will try json in the meantime

noot:
@SuperFluffy 
> What I would block this PR on is testing: can we define a minimal test where a request is sent to the system and then read? It's fine if it just tests the happy path.

do you mean by starting the sequencer server and sending a request to it? I'll need to get the ABCI gRPC server here built to do that https://github.com/cometbft/cometbft/blob/main/proto/tendermint/abci/types.proto#L18

crates/astria-sequencer/src/accounts/state_ext.rs
```diff
@@ -0,0 +1,72 @@
+use anyhow::Result;
+use async_trait::async_trait;
+use penumbra_storage::{
+    StateRead,
+    StateWrite,
+};
+
+pub type Address = String;
+pub type Balance = u64; // might need to be larger
```
SuperFluffy, line 9: Please avoid type aliases for primitive types because it makes it difficult to parse and understand code (and figure out what's wrong in compiler errors).

Type aliases are generally used for simplifying complex types, for example `pub type Nested<Item> = N<N<N<N<N<N<N<N<L<Item>>>>>>>>>;` (from the penumbra codebase).

I think these types are prime candidates for newtype wrappers, for which you can either serde or borsh. You can then get rid of the entire parsing logic.

For serde, have a look at bincode.

For borsh: https://borsh.io/

crates/astria-sequencer/src/accounts/accounts.rs
SuperFluffy, line 1: This file is dead code (not included in the module tree). I was trying to make sense of it and only then realized that it's not even used. :(

crates/astria-sequencer/src/sequencer.rs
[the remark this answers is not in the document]
SuperFluffy: Yeah, I am happy with getting rid of `Sequencer::new`. You can do something like `let config = config::get().unwrap();` and `Sequencer::run_until_stopped(config).await`. That's in fact a common pattern.

crates/astria-sequencer/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+is-terminal = "0.4.7"
+penumbra-component = { git = "https://github.com/penumbra-zone/penumbra.git", tag = "v0.52.2" }
+penumbra-storage = { git = "https://github.com/penumbra-zone/penumbra.git", tag = "v0.52.2" }
+tendermint-proto = "0.29"
+tendermint = "0.29"
+tower = "0.4"
+tower-abci = "0.5.0"
+tower-actor = "0.1.0"
+
+color-eyre = { workspace = true }
```
SuperFluffy, line 19: delete?

crates/astria-sequencer/src/accounts/transaction.rs
```diff
@@ -0,0 +1,69 @@
...
+impl ActionHandler for Transaction {
+    fn check_stateless(&self) -> Result<()> {
+        Ok(())
+    }
+
+    async fn check_stateful<S: StateReadExt + 'static>(&self, state: &S) -> Result<()> {
+        let curr_nonce = state.get_account_nonce(&self.from).await?;
+
+        // TODO: do nonces start at 0 or 1? this assumes an account's first tx has nonce 1.
+        if curr_nonce >= self.nonce {
```
SuperFluffy, line 48: `anyhow::ensure!`

crates/astria-sequencer/src/accounts/transaction.rs
```diff
@@ -0,0 +1,69 @@
...
+            anyhow::bail!("invalid nonce");
+        }
+
+        let curr_balance = state.get_account_balance(&self.from).await?;
+        if curr_balance < self.amount {
```
SuperFluffy, line 53: `anyhow::ensure!`

crates/astria-sequencer/src/accounts/state_ext.rs
```diff
@@ -0,0 +1,72 @@
...
+    format!("{}/nonce", storage_key(address))
+}
+
+#[async_trait]
+pub trait StateReadExt: StateRead {
+    async fn get_account_balance(&self, address: &str) -> Result<Balance> {
+        let bytes = self
+            .get_raw(&balance_storage_key(address))
+            .await
+            .expect("storage error");
```
SuperFluffy, line 32: Why panic instead of returning an error?

crates/astria-sequencer/src/accounts/state_ext.rs
```diff
@@ -0,0 +1,72 @@
...
+        let Some(bytes) = bytes else {
+            // the account has not yet been initialized; return 0
+            return Ok(0);
```
SuperFluffy, line 35: This means you cannot distingish between "uninitialized" and "zero balance". Maybe `get_account_balance -> Result<Option<Balance>>`?

crates/astria-sequencer/src/accounts/state_ext.rs
```diff
@@ -0,0 +1,72 @@
...
+        };
+
+        let balance = u64::from_be_bytes(bytes[0..8].try_into()?);
```
SuperFluffy, line 38: If for whatever reason the byte vector is shorter this will panic.

crates/astria-sequencer/src/accounts/state_ext.rs
```diff
@@ -0,0 +1,72 @@
...
+        Ok(balance)
+    }
+
+    async fn get_account_nonce(&self, address: &str) -> Result<Nonce> {
+        let bytes = self
+            .get_raw(&nonce_storage_key(address))
+            .await
+            .expect("storage error");
```
SuperFluffy, line 46: same as above: why panic?

crates/astria-sequencer/src/accounts/state_ext.rs
```diff
@@ -0,0 +1,72 @@
...
+        let Some(bytes) = bytes else {
+            // the account has not yet been initialized; return 0
+            return Ok(0);
+        };
+
+        let balance = u64::from_be_bytes(bytes[0..8].try_into()?);
+        Ok(balance)
+    }
+
+    async fn get_account_nonce(&self, address: &str) -> Result<Nonce> {
```
SuperFluffy, line 42: instrument

SuperFluffy (changes requested):
Bunch of comments.

Big things that I would like to get done before we merge this into main:

1. no panics except for where invariants are violated (that's the `(&mut Arc)::get_mut` calls)
2. error context for all errors through `anyhow::Context::context` (note that that's different from `eyre::WrapErr::wrap_err`)
3. replace manual parsing of primitives by newtype wrappers + deriving borsh or serde (and then use borsh directly or bincode for serde)
4. instrument all functions

crates/astria-sequencer/src/accounts/state_ext.rs
noot: this is the behaviour I want for now (and mirrors most chains like ethereum) - all accounts technically already exist, but whether anyone knows the private key for them is unknown

crates/astria-sequencer/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+anyhow = "1"
+async-trait = "0.1.52"
+bytes = "1.4"
+is-terminal = "0.4.7"
+penumbra-component = { git = "https://github.com/penumbra-zone/penumbra.git", tag = "v0.52.2" }
+penumbra-storage = { git = "https://github.com/penumbra-zone/penumbra.git", tag = "v0.52.2" }
+tendermint-proto = "0.29"
+tendermint = "0.29"
+tower = "0.4"
+tower-abci = "0.5.0"
```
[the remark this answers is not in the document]
noot: I had to update tendermint to 0.31 as there was no penumbra release I could find that supported tendermint 0.30. tower-abci was updated to 0.7.0 because of that also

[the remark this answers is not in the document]
noot: we should probably bump the repo-wide tendermint version to 0.31 or 0.32 (the latest)

crates/astria-sequencer/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+
+[dependencies]
+anyhow = "1"
+async-trait = "0.1.52"
+bytes = "1.4"
+is-terminal = "0.4.7"
+penumbra-component = { git = "https://github.com/penumbra-zone/penumbra.git", tag = "v0.52.2" }
+penumbra-storage = { git = "https://github.com/penumbra-zone/penumbra.git", tag = "v0.52.2" }
+tendermint-proto = "0.29"
+tendermint = "0.29"
```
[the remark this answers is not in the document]
noot: addressed above

crates/astria-sequencer/src/accounts/accounts.rs
```diff
@@ -0,0 +1,127 @@
...
+
+    pub async fn check_stateful<S: StateRead + 'static>(&self, state: &S) -> Result<()> {
+        let account_state = state
+            .get_raw(&storage_key(&self.from))
+            .await?
+            .unwrap_or([0u8; 12].to_vec());
+        let nonce = u32::from_be_bytes(account_state[0..4].try_into()?);
+
+        // TODO: do nonces start at 0 or 1?
+        if nonce <= self.nonce {
```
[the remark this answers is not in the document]
noot: deleted this file, was here by accident

crates/astria-sequencer/src/accounts/state_ext.rs
noot: I updated these and put them in `account/types.rs` and used borsh for encoding/decoding before putting in storage. let me know what you think!

noot:
> Big things that I would like to get done before we merge this into main:
    no panics except for where invariants are violated (that's the (&mut Arc)::get_mut calls)
    error context for all errors through anyhow::Context::context (note that that's different from eyre::WrapErr::wrap_err)
    replace manual parsing of primitives by newtype wrappers + deriving borsh or serde (and then use borsh directly or bincode for serde)
    instrument all functions

@SuperFluffy I think these are all addressed; the only thing that may not be fully addressed is adding context for every error, as some errors already had context added lower-down on the stack so I wasn't sure if I should re-add another context. let me know if you want literally every error to have additional context and I can add!

crates/astria-sequencer/Cargo.toml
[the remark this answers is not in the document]
erwanor: fyi, we cut a release of `tower-abci@0.8` that plugs into `0.32`!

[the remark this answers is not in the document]
noot: nice, will probably do a repo-wide update when this is merged!

crates/astria-sequencer/src/app.rs
```diff
@@ -0,0 +1,371 @@
...
+
+        let mut state_tx = self
+            .state
+            .try_begin_transaction()
+            .expect("state Arc should not be referenced elsewhere");
+
+        state_tx.put_block_height(0);
+
+        // call init_chain on all components
+        AccountsComponent::init_chain(&mut state_tx, &genesis_state).await?;
```
joroshiba, line 92: Should we have some sort of Component manager to define modules and the order they will be called in similar to the Cosmos [manager](https://docs.cosmos.network/v0.47/building-modules/module-manager#manager)?

crates/astria-sequencer/src/accounts/types.rs
```diff
@@ -0,0 +1,114 @@
...
+    BorshDeserialize,
+    Serialize,
+    Deserialize,
+    PartialEq,
+    Eq,
+    PartialOrd,
+    Ord,
+    Debug,
+)]
+pub struct Nonce(u32);
```
joroshiba, line 94: Should we use a u64 here? I believe ethereum uses [uint64 for nonce](https://github.com/astriaorg/go-ethereum/blob/8b9ba616d1c4fbe2acf80c5f6773ea0446ce801f/core/types/transaction.go#L83), as does [cosmos for sequence](https://docs.cosmos.network/main/core/transactions#:~:text=uint64%20sequence%20%3D%205%3B)?

crates/astria-sequencer/src/accounts/transaction.rs
```diff
@@ -0,0 +1,72 @@
...
+#[async_trait]
+impl ActionHandler for Transaction {
+    fn check_stateless(&self) -> Result<()> {
+        Ok(())
+    }
+
+    async fn check_stateful<S: StateReadExt + 'static>(&self, state: &S) -> Result<()> {
+        let curr_nonce = state.get_account_nonce(&self.from).await?;
+
+        // TODO: do nonces start at 0 or 1? this assumes an account's first tx has nonce 1.
```
joroshiba, line 54: I believe each account nonce balance should start at `0` and the first transaction should be `1`

crates/astria-sequencer/src/accounts/state_ext.rs
```diff
@@ -0,0 +1,90 @@
...
+    fn put_account_balance(&mut self, address: &Address, balance: Balance) -> Result<()> {
+        let bytes = balance
+            .try_to_vec()
+            .context("failed to serialize balance")?;
+        self.put_raw(balance_storage_key(address.to_str()), bytes);
+        Ok(())
+    }
+
+    #[instrument(skip(self))]
+    fn put_account_nonce(&mut self, address: &Address, nonce: Nonce) -> Result<()> {
```
joroshiba, line 83: Do we need a `put_account_nonce` or should this have all the increment logic and be `increment_account_nonce`? Is there ever an instance in which account nonce should go up by more than one at a time?

crates/astria-sequencer/src/accounts/transaction.rs
```diff
@@ -0,0 +1,72 @@
...
+        ensure!(curr_nonce < self.nonce, "invalid nonce",);
+
+        let curr_balance = state.get_account_balance(&self.from).await?;
+        ensure!(curr_balance >= self.amount, "insufficient funds",);
+
+        Ok(())
+    }
+
+    async fn execute<S: StateWriteExt>(&self, state: &mut S) -> Result<()> {
```
joroshiba, line 63: Is there anything preventing the call of `execute` without having first validated the tx? What happens if there are insufficient funds when this is called?

In general I wonder if structurally we should be moving more of the logic of "tx creates these DB level changes" directly into the state implementation.

crates/astria-sequencer/src/accounts/transaction.rs
```diff
@@ -0,0 +1,72 @@
...
+        let from_balance = state.get_account_balance(&self.from).await?;
+        let from_nonce = state.get_account_nonce(&self.from).await?;
+        let to_balance = state.get_account_balance(&self.to).await?;
+        state.put_account_balance(&self.from, from_balance - self.amount)?;
```
joroshiba, line 67: Maybe this is just that lower level, but it feels like maybe the logic for state transitions which has race possibilities should be handled by the state itself? If putting the balance succeeds but nonce fails, we have issues.

crates/astria-sequencer/src/accounts/transaction.rs
```diff
@@ -0,0 +1,72 @@
...
+        state.put_account_nonce(&self.from, from_nonce + Nonce::from(1))?;
```
joroshiba, line 68: Should we be setting to the nonce of the transaction and validating that it is acceptable?

crates/astria-sequencer/src/accounts/transaction.rs
noot: cool, that's how it functions right now, the first tx has nonce 1

crates/astria-sequencer/src/app.rs
noot: the cosmos manager seems more needed because it's an SDK, and it needs to figure out what modules exist at runtime, whereas for us it's fine to just write them in and do it at compile time (this is what penumbra does too)

crates/astria-sequencer/src/accounts/types.rs
noot: yeah, can update, u64 just seemed a bit large lol

crates/astria-sequencer/src/accounts/state_ext.rs
noot: if we have an increment function then it'll have a get and a put, but usually wherever this function is called there's already a get, so it seems a bit redundant to have 2 gets as we want to minimize reads/writes.

crates/astria-sequencer/src/accounts/transaction.rs
noot: yeah, this function is only called inside `DeliverTx`, which always performs validation (stateless and non-stateless) before executing it. what do you mean by moving it directly into the state implementation?

crates/astria-sequencer/src/accounts/transaction.rs
noot: right, so these writes are all happening on a `StateDelta`, which is basically a list of pending writes that happen during a block's execution. the writes are all batched and only actually written to state once `Commit` is called. so yeah this effectively is atomic afaik

crates/astria-sequencer/src/accounts/transaction.rs
noot: the validation occurs outside this function in `check_stateful` (which happens right before `execute` is called) so I don't think we need to double validate

crates/astria-sequencer/src/accounts/state_ext.rs
joroshiba: Would we need the get if it was an increment? It seems we are only getting to do the increment.

crates/astria-sequencer/src/accounts/transaction.rs
joroshiba: I just want to make sure we are creating hardened code here. Should it be possible to call execute before validation? If no we should enforce that it isn't possible not hope someone doesn't forget that.

crates/astria-sequencer/src/accounts/types.rs
joroshiba: Better to plan on being successful than not!

crates/astria-sequencer/src/accounts/transaction.rs
joroshiba: Can we create an issue to add tests for error cases here?

joroshiba (approved):
LGTM. Great start for the chain overall.

I don't consider any of my comments blocking on this, but we should create issues and consider resolving separately.

SuperFluffy:
My last commits fixed compilation in CI. The reason is that CI defines a `RUSTFLAGS` env var which overrides the setting in `.config/cargo.toml`.

The other commits are about fixing clippy lints. I also increased the number of lints that are used for `astria-sequencer` to `clippy::pedantic`. This is now enforced in a separate step. The other crates will be fixed step by step.

crates/astria-sequencer/src/service/consensus.rs
```diff
@@ -0,0 +1,188 @@
...
+        while let Some(Message {
+            req,
+            rsp_sender,
+            span,
+        }) = self.queue.recv().await
+        {
+            // The send only fails if the receiver was dropped, which happens
+            // if the caller didn't propagate the message back to tendermint
+            // for some reason -- but that's not our problem.
+            rsp_sender.send(self.handle_request(req).instrument(span).await);
```
SuperFluffy, line 55: I broke-fixed a clippy lint. Fix:

Split this into 2 steps:

1. Match on `self.handle_request`.
    a. if an error, emit a `warn!(parent: &span, ...)` event with the error
2. `if let Err(e) = rsp_sender.send` with the ok-value, and then `warn!` in the same span

SuperFluffy: done

crates/astria-sequencer/src/accounts/transaction.rs
SuperFluffy: We should have a write-up about how state gets transformed when moving through these components.

SuperFluffy:
> > Big things that I would like to get done before we merge this into main:
> > no panics except for where invariants are violated (that's the (&mut Arc)::get_mut calls)
> > error context for all errors through anyhow::Context::context (note that that's different from eyre::WrapErr::wrap_err)
> > replace manual parsing of primitives by newtype wrappers + deriving borsh or serde (and then use borsh directly or bincode for serde)
> > instrument all functions
> 
> @SuperFluffy I think these are all addressed; the only thing that may not be fully addressed is adding context for every error, as some errors already had context added lower-down on the stack so I wasn't sure if I should re-add another context. let me know if you want literally every error to have additional context and I can add!

To document for posteriority: I added context to literally every error I could find.

SuperFluffy (approved):
Approving to unblock this.

Opening an issue to fix rocksdb in CI and docker. See also these github issues:

https://github.com/rust-rocksdb/rust-rocksdb/issues/310#issuecomment-1537137868
https://github.com/rust-rocksdb/rust-rocksdb/pull/354#issuecomment-1514852676

PR #6, opened 2023-05-17
implement github workflow for testing, linting

SuperFluffy:
This PR implements github workflows for running tests, formatting, and linting.

Because `astria-conductor` needs the `astria-sequencer-relayer` to be running but we want to avoid building container images as part of the tests, this PR circumvents the issue by:

1. building `astria-sequencer-relayer` as a binary before the tests are run;
2. copying the binary to `/tmp/astria/`;
3. mounting `/tmp/astria/astria-sequencer-relayer` into a barebones `debian:bookworm-slim` container as a `hostPath`.

This strategy can be repeated for future test requirements.

Note that `debian:bookworm-slim` (glibc 2.36) is preferred over `debian:buster-slim` (2.28) because of compatibility with `ubuntu-22.04` (glibc 2.35).

This PR also moves to rustls everywhere to avoid linking to openssl.

Issues that this PR will left unaddressed:

+ `gossipnet` tests stall indefinitely and never finish on CI (but work locally): #24 
+ `rs-cnc` needs celestia running for its roundtrip test to work (needs to be updated to folow `astria-conductor` and `astria-sequencer-relayer`): #25 
+ `astria-conductor` and `astria-sequencer-relayer` have a ton of duplicate code. This needs to be refactored into a common testing crate: #26


crates/astria-conductor-test/kubernetes/deployment.yml
```diff
@@ -132,14 +132,17 @@ spec:
...
             - configMapRef:
                 name: metro-env
-        - name: relayer
-          image: "ghcr.io/astriaorg/sequencer-relayer:sha-574f422-debug"
-          command: ["/usr/local/bin/relayer"]
+        - name: astria-sequencer-relayer
+          image: docker.io/library/debian:bookworm-slim
+          command: ["/usr/local/bin/astria-sequencer-relayer"]
           args:
             - "--sequencer-endpoint=http://localhost:1318"
```
steezeburger, line 139: 2/3 of these arguments are default values. `--sequencer-endpoint=http://localhost:1318` is the only one that is not a default because we're manually configuring Metro for 1318, but I think that was changed from 1317 to avoid port conflicts, which are no longer a concern.

Could we start Metro on 1317 again, which would allow us to get rid of all three of these arguments?

SuperFluffy: Yeah, we should totally do this. Let's open an issue though and do it in a follow-up PR to not delay this one.

Maybe we can do this while splitting out the individual containers into separate pods so that we don't have to worry about port overlaps ever again?

crates/astria-rs-cnc/Cargo.toml
```diff
@@ -12,12 +12,12 @@ bytes = "1.3.0"
...
 hex = "0.4.3"
 rand = "0.8.5"
-reqwest = { version = "0.11", default-features = false, features = ["json"] }
+reqwest = { version = "0.11", default-features = false, features = ["json", "rustls-tls"] }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 
 [features]
-default = ["reqwest/default-tls"]
+# default = ["reqwest/default-tls"]
```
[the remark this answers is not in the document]
SuperFluffy: Yes, thank you. I forgot to address this. I needed to remove the openssl dep due to ssl being linked into the final binary which causes some issues when mounting it into the k8s container.

I don't see a lot of need to drag around openssl so I'll remove the feature.

crates/astria-conductor-test/kubernetes/deployment.yml
steezeburger: That sounds like a great idea to me!

steezeburger (approved):
Looks great!

.github/workflows/rust.yml
```diff
@@ -0,0 +1,99 @@
...
+      # FIXME: the gossipnet and rs-cnc tests are currently broken in CI:
+      #        https://github.com/astriaorg/astria/issues/24
+      #        https://github.com/astriaorg/astria/issues/25
+      - name: Build tests
+        run: |
+          cargo test --release --no-run \
+          --workspace \
+          --exclude astria-gossipnet \
+          --exclude astria-rs-cnc
+      - name: Run light tests
```
itamarreif, line 46: why is this the same as "Build tests" right above it?

SuperFluffy: Great find! Thank you for spotting this!

joroshiba:
LGTM

PR #7, opened 2023-05-17
transition to single astria-proto directory

joroshiba:
- Adds workflow to run buf lint, and push build artifacts to `buf.build/astria/astria` (non main pushes are drafts)
- Combines `astria-execution-apis-rpc` and `astria-sequencer-relayer-proto` into `astria-proto`
- Makes all packages prefixed with `astria` following proto standards to allow cleaner imports. (ie `astria.execution.v1`)
- Adds a protobuf package name to `astria-sequencer-relayer-proto` 

Fixes https://github.com/astriaorg/astria/issues/10


crates/astria-proto/src/lib.rs
```diff
@@ -0,0 +1,7 @@
+pub mod execution {
```
SuperFluffy, line 1: Should we name the module matching the protobuf package name? Right now the `v1` bit gets lots.

That would be `pub mod astria_execution_v1`, which looks awully verbose (think `use astria_proto::astria_execution_v1::SomeType`)

Maybe nested modules? E.g.

```rust
pub mod execution {
    pub mod v1 {
        include!(...)
    }
}
```

SuperFluffy (commented):
This looks very good to me.

I would like to match the rust module names to the proto package names (see the comments). 

Same with the protobuf filenames themselves:

```
# protobuf package
astria.execution.v1
# current path
proto/astria/execution/v1/execution.proto
# suggested path
proto/astria.exeuction.v1.proto
```

crates/astria-proto/proto/astria/execution/v1/execution.proto
```diff
@@ -1,6 +1,6 @@
 syntax = 'proto3';
 
-package execution.v1;
+package astria.execution.v1;
```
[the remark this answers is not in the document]
joroshiba: This is a naming standard enforced by linting as configured within buf, but also followed by most orgs I've seen. Think it's best to stick with the normal standard.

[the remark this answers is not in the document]
SuperFluffy: Good to know, I attempted to create a `proto/astria.execution.v1` and moved the proto in there, and it complained with:
```
proto/astria.execution.v1/execution.proto:3:1:Files with package "astria.execution.v1" must be within a directory "astria/execution/v1" relative to root but were in directory "astria.execution.v1".
```

.github/workflows/proto.yml
```diff
@@ -0,0 +1,30 @@
+name: buf
+on:
+  pull_request:
+  push:
+    branches:
+        - main
+jobs:
+  build-lint-push-protos:
+    runs-on: ubuntu-latest
+    environment: BUF
```
SuperFluffy, line 10: Is this used anywhere? I believe it can be deleted.

crates/astria-proto/src/lib.rs
joroshiba: I like the idea of nested modules here, but I am not sure that I love the idea of duplicating `astria`. 

What if crates were:
`astria_proto::execution::v1`

The difference between package name and module is then that in protobuf we don't specify that it's proto, because obviously.

.github/workflows/proto.yml
joroshiba: Yes, the BUF_TOKEN secret is attached to this environment. It gives the "deployment" below

crates/astria-proto/buf.work.yaml
```diff
@@ -0,0 +1,3 @@
+version: v1
+directories:
+  - proto
```
SuperFluffy, line 3: Overlooked it earlier: please add a newline to this file.

SuperFluffy (approved):
Looks great, thank you for transitioning this.

I would like to merge this into `main` after #6 is merged.

.github/workflows/proto.yml
```diff
@@ -0,0 +1,30 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - uses: bufbuild/buf-setup-action@v1
+        with:
+          version: "1.15.1"
+          github_token: ${{ github.token }}
+      - uses: bufbuild/buf-lint-action@v1
+        with:
+          input: "crates/astria-proto/proto"
+      # TODO(GH-3): Uncomment out after reaching any production stage
```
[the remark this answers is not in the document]
joroshiba: It's a github issue, I have since transferred the issue to the monorepo and updated the number

PR #9, opened 2023-05-17
fix clippy and gossipnet test in CI

noot:


SuperFluffy:
Note that this still has to pass be activated in CI by changing the last step in the `test` job. Right now it's only testing `-p astria-conductor -p astria-sequencer-relayer`, so `astria-gossipnet` is not running.

noot:
closing, will open another PR

Issue #14, opened 2023-05-11
Replace logging by tracing, remove alert actor


SuperFluffy:
[`tracing`](https://docs.rs/tracing) is the goto solution to logging/tracing in async applications. Moving the system to the tracing ecosystem will also allow replacing the alert actor:
https://github.com/astriaorg/astria-conductor/blob/ff6580f7de96301ff44ee3400a7e6a5293d37044/astria-conductor/src/bin/conductor.rs#L103-L112

As a suggested implementation detail: the executor can also received [`tracing::Span`](https://docs.rs/tracing/latest/tracing/struct.Span.html)s so that it can delegate methods to be ran in the same span (or a child span) of the code that originally sent the message to it.

PR #34, opened 2023-05-22
use tracing for telemetry

SuperFluffy:
This commit replaces log and flexilogger by tracing and tracing-subscriber. The default log level is `info`, which can be overriden by using the `RUST_LOG` environment variable with directives as described in the tracing subscriber docs: https://docs.rs/tracing-subscriber/0.3.17/tracing_subscriber/filter/struct.EnvFilter.html

Because env var directives are strictly more powerful than setting verbosity levels on the command line the `log-level` flag is removed.

If astria-conductor is connected to a tty it will emit tracing events using human-readable format. If it is not connected to a tty it will emit JSON.

By default logging is disabled in integration tests. It can be enabled by setting the environment variable `TEST_LOG=1`.

Note that the service no longer writes explicitly writes logs to a file. Because astria-conductor is intended to be run as a service by a service manager or orchestration platform, it is assumed that persisting logs is handled externally.

Partially addresses #14


steezeburger (approved):

noot (approved):
this is great!!

PR #36, opened 2023-05-22
feat(gossipnet): discovery via DHT

### LINKED ISSUE #19: implement DHT discovery
right now, there's only local discovery with mDNS. need to add a DHT for non-local discovery

noot:
implements discovery via the kademlia DHT. this allows peers that are not on the same local network to discover each other.

additions:
- add identify protocol behaviour which is required to find out remote peer's listening addresses
- add kademlia DHT behaviour and "random walk" used to discover new peers 
- "random walk" works by asking your peers about their view of the network, using a random key as the search value

closes #19

### DIFF
--- a/crates/astria-gossipnet/tests/network.rs
+++ b/crates/astria-gossipnet/tests/network.rs
@@ -236,11 +236,11 @@ async fn test_dht_discovery() {
             let event = event.unwrap();
             if let Event::GossipsubPeerConnected(peer_id) = event {
                 println!("Charlie connected to {:?}", peer_id);
-                if charlie.peer_count() == 1 {
+                if charlie.num_peers() == 1 {
                     charlie.random_walk().await.unwrap();
                 }
 
-                if charlie.peer_count() == 2 {
+                if charlie.num_peers() == 2 {
                     charlie_tx.send(()).unwrap();
                     return;
                 }


crates/astria-gossipnet/src/network_stream.rs
```diff
@@ -0,0 +1,271 @@
...
+                        warn!("failed to parse multiaddr: {maddr_str}");
+                        continue;
+                    };
+
+                    self.multiaddrs.push(multiaddr);
+                    return Poll::Ready(Some(Event::NewListenAddr(address)));
+                }
+                SwarmEvent::ConnectionEstablished {
+                    peer_id,
+                    endpoint: _,
```
SuperFluffy, line 100: Use `..` if you want to ignore all other fields instead of binding them to `_`

crates/astria-gossipnet/src/network_stream.rs
```diff
@@ -0,0 +1,271 @@
...
+    #[cfg(feature = "dht")]
+    FoundClosestPeers(Vec<PeerId>),
+    #[cfg(feature = "dht")]
+    GetClosestPeersError(GetClosestPeersError),
+}
+
+impl futures::Stream for Network {
+    type Item = Event;
+
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
```
SuperFluffy, line 76: I think this could be a stream returning `type Item = Result<Event, Error>;` and let the consumer of the network handle the errors (like logging).

crates/astria-gossipnet/src/network_stream.rs
```diff
@@ -0,0 +1,271 @@
...
+                        ..
+                    },
+                )) => {
+                    debug!(
+                        "Routable peer: {peer:?}, Address: {address:?}",
+                        peer = peer,
+                        address = address,
+                    );
+                }
+                #[cfg(feature = "dht")]
```
SuperFluffy, line 170: These interspersed conditional compilation flags make it tough to read. Please delegate handling of kadmelia and mdns events to separate methods

crates/astria-gossipnet/src/network_stream.rs
```diff
@@ -0,0 +1,271 @@
...
+    ProvideError(AddProviderError),
+    #[cfg(feature = "dht")]
+    RoutingUpdated(PeerId, Addresses),
+    #[cfg(feature = "dht")]
+    FoundClosestPeers(Vec<PeerId>),
+    #[cfg(feature = "dht")]
+    GetClosestPeersError(GetClosestPeersError),
+}
+
+impl futures::Stream for Network {
```
SuperFluffy, line 73: This stream impl is in an odd state between neither being completely high level nor low level. It's partially exposing runtime details, while silently dropping others.

As a consumer it seems to me I should be mainly (only?) interested in `Message`, and I am not sure what I should do with the other variants. Silently drop them?

I believe we could simplify this interface to just yield messages and for the other events just create log entries.

SuperFluffy (changes requested):

crates/astria-gossipnet/Cargo.toml
```diff
@@ -4,12 +4,16 @@ version = "0.1.0"
...
 
 color-eyre = { workspace = true }
 futures = { workspace = true }
-tokio = { workspace = true, features = ["macros", "rt-multi-thread", "signal"] }
+tokio = { workspace = true, features = ["macros", "rt-multi-thread", "signal", "sync"] }
 tracing = { workspace = true }
+tracing-subscriber = { workspace = true, features = ["env-filter"] }
 
 [features]
+default = [ "dht" ]
```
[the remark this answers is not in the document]
noot: the reason I went with this method is due to this:
```
#[derive(NetworkBehaviour)]
pub(crate) struct GossipnetBehaviour {
    pub(crate) ping: ping::Behaviour,
    pub(crate) identify: identify::Behaviour,
    pub(crate) gossipsub: gossipsub::Behaviour,
    #[cfg(feature = "mdns")]
    pub(crate) mdns: mdns::tokio::Behaviour,
    #[cfg(feature = "dht")]
    pub(crate) kademlia: Kademlia<MemoryStore>,
}
```

this structure is passed into the `SwarmBuilder` to configure the network behaviour. however, if we don't go with conditional compilation, we'll need to have 4 different structures, one for each possible configuration :/

crates/astria-gossipnet/src/network_stream.rs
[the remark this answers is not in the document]
noot: yeah sorry, I can move it back if it's easier, but the main change that was done was just adding the statements under `// DHT events`

crates/astria-gossipnet/Cargo.toml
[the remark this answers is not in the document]
SuperFluffy: I had a look at `SwarmBehaviour` and noticed the [`Toggle`](https://docs.rs/libp2p/0.51.3/libp2p/swarm/behaviour/toggle/struct.Toggle.html) behaviour which should allow to layer the required behaviors and enable/disable them at construction? And they have `impl<T> NetworkBehaviour for Toggle<T> where T: NetworkBehaviour`: https://docs.rs/libp2p/0.51.3/libp2p/swarm/behaviour/toggle/struct.Toggle.html#impl-NetworkBehaviour-for-Toggle%3CTBehaviour%3E

BTW, this is similar to `tracing-subscriber`s `Layer` trait that has `impl<T> Layer for Option<L> where L: Layer` to allow enabling/disabling layers based on config.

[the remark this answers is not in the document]
noot: nice, updated to use `Toggle`, no more conditional compilation

crates/astria-gossipnet/src/network_stream.rs
noot: most of these I used in the tests, things like `NewListenAddr` are useful for consumers, but I'll remove the ones that aren't really used

crates/astria-gossipnet/src/network_stream.rs
```diff
@@ -0,0 +1,271 @@
...
+pub enum Event {
+    NewListenAddr(Multiaddr),
+    Message(Message),
+    #[cfg(feature = "mdns")]
+    MdnsPeersConnected(Vec<PeerId>),
+    #[cfg(feature = "mdns")]
+    MdnsPeersDisconnected(Vec<PeerId>),
+    PeerConnected(PeerId),
+    PeerSubscribed(PeerId, TopicHash),
+    #[cfg(feature = "dht")]
```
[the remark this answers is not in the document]
noot: removed all conditional compilation

crates/astria-gossipnet/src/network.rs
```diff
@@ -131,43 +134,119 @@ impl Network {
...
+                    mdns::Config::default(),
+                    local_peer_id,
+                )?)
+            } else {
+                None
+            };
+
             let behaviour = GossipnetBehaviour {
+                identify: identify::Behaviour::new(identify::Config::new(
+                    "gossipnet/0.1.0".into(),
```
joroshiba, line 163: Should we pull this into a constant?

noot: done!

crates/astria-gossipnet/src/network.rs
```diff
@@ -1,77 +1,84 @@
...
+    keypair: Option<Keypair>,
+    with_mdns: bool,
+    with_kademlia: bool,
 }
 
 impl NetworkBuilder {
     pub fn new() -> Self {
         Self {
             bootnodes: None,
             port: 0, // random port
```
SuperFluffy, line 78: nit (not for this PR): using a default port of `0` is probably too random.

crates/astria-gossipnet/src/network.rs
```diff
@@ -85,33 +92,31 @@ impl NetworkBuilder {
...
+        // TODO: load/store keypair from disk
+        self.keypair = Some(keypair);
+        self
     }
-}
 
-impl Default for NetworkBuilder {
-    fn default() -> Self {
-        Self::new()
+    pub fn with_mdns(mut self, with_mdns: bool) -> Self {
```
SuperFluffy, line 101: Please add a doc comment.

very minor nit (feel free to ignore): on a builder it's fine to just call this `fn mdns` or `fn set_mdns`

crates/astria-gossipnet/src/network.rs
```diff
@@ -85,33 +92,31 @@ impl NetworkBuilder {
         self
     }
 
-    pub fn build(self) -> Result<Network> {
-        Network::new(self.bootnodes, self.port)
+    pub fn keypair(mut self, keypair: Keypair) -> Self {
```
SuperFluffy, line 95: Please add a doc comment.

Also looking at `NetworkBuilder::build`, please add that a keypair will be randomly generated if not set.

crates/astria-gossipnet/src/network.rs
```diff
@@ -85,33 +92,31 @@ impl NetworkBuilder {
...
+        self
     }
-}
 
-pub struct Network {
-    pub multiaddr: Multiaddr,
-    swarm: Swarm<GossipnetBehaviour>,
-    terminated: bool,
-}
+    pub fn with_kademlia(mut self, with_kademlia: bool) -> Self {
```
SuperFluffy, line 106: nit: as above with `mdns`.

not-a-nit: please add a doc comment on what this function is doing.

crates/astria-gossipnet/src/network.rs
```diff
@@ -85,33 +92,31 @@ impl NetworkBuilder {
...
+        self
+    }
 
-impl Network {
-    pub fn new(bootnodes: Option<Vec<String>>, port: u16) -> Result<Self> {
-        // TODO: store this on disk instead of randomly generating
-        let local_key = identity::Keypair::generate_ed25519();
-        let local_peer_id = PeerId::from(local_key.public());
-        info!("local peer id: {local_peer_id:?}");
+    pub fn build(self) -> Result<Network> {
```
SuperFluffy, line 111: Please add a doc comment

crates/astria-gossipnet/src/network.rs
```diff
@@ -131,43 +136,119 @@ impl Network {
...
             gossipsub_config,
         )
         .map_err(|e| eyre!("failed to create gossipsub behaviour: {}", e))?;
 
         let mut swarm = {
-            #[cfg(feature = "mdns")]
-            let mdns = mdns::tokio::Behaviour::new(mdns::Config::default(), local_peer_id)?;
+            let kademlia = if self.with_kademlia {
+                let mut cfg = KademliaConfig::default();
+                cfg.set_query_timeout(Duration::from_secs(5 * 60));
```
SuperFluffy, line 147: Should the timeout be configurable? If not, please add this to the documentation, maybe on `fj with_kademlia`

crates/astria-gossipnet/src/network.rs
```diff
@@ -131,43 +136,119 @@ impl Network {
 
         // build a gossipsub network behaviour
         let gossipsub = gossipsub::Behaviour::new(
-            gossipsub::MessageAuthenticity::Signed(local_key),
+            gossipsub::MessageAuthenticity::Signed(keypair),
             gossipsub_config,
         )
         .map_err(|e| eyre!("failed to create gossipsub behaviour: {}", e))?;
```
SuperFluffy, line 142: use modern rust formatting (I also like to add backticks to make variable things stand out in errors or in general formatted logs):
```rust
        .map_err(|e| eyre!("failed to create gossipsub behaviour: `{e}`"))?;
```

crates/astria-gossipnet/src/network.rs
```diff
@@ -131,43 +136,119 @@ impl Network {
...
                 ping: ping::Behaviour::default(),
+                mdns: mdns.into(),
+                kademlia: kademlia.into(),
             };
             SwarmBuilder::with_tokio_executor(transport, behaviour, local_peer_id).build()
         };
 
-        let listen_addr = format!("/ip4/0.0.0.0/tcp/{}", port);
+        let listen_addr = format!("/ip4/0.0.0.0/tcp/{}", self.port);
         swarm.listen_on(listen_addr.parse()?)?;
```
SuperFluffy, line 177: i know not touched these lines in this PR, but if we are at it...: error context on both `?`

crates/astria-gossipnet/src/network.rs
```diff
@@ -131,43 +136,119 @@ impl Network {
...
 
-        if let Some(addrs) = bootnodes {
+        if let Some(addrs) = self.bootnodes {
             addrs.iter().try_for_each(|addr| -> Result<_> {
-                debug!("dialing {:?}", addr);
-                let remote: Multiaddr = addr.parse()?;
-                swarm.dial(remote)?;
-                debug!("dialed {addr}");
+                let mut maddr: Multiaddr = addr.parse()?;
```
SuperFluffy, line 181: error context

crates/astria-gossipnet/src/network.rs
```diff
@@ -131,43 +136,119 @@ impl Network {
...
+                swarm.dial(maddr.clone())?;
```
SuperFluffy, line 182: error context

crates/astria-gossipnet/src/network.rs
```diff
@@ -131,43 +136,119 @@ impl Network {
...
+
+                let Some(peer_id) = maddr.pop() else {
+                    return Err(eyre!("failed to parse peer id from addr: {}", addr));
```
SuperFluffy, line 185:
```suggestion
                    return bail!("failed to parse peer id from addr: `{addr}`");
```

crates/astria-gossipnet/src/network.rs
```diff
@@ -131,43 +136,119 @@ impl Network {
...
+                };
+
+                match peer_id {
+                    Protocol::P2p(peer_id) => {
+                        let peer_id = match PeerId::from_multihash(peer_id) {
+                            Ok(peer_id) => peer_id,
+                            Err(e) => {
+                                return Err(eyre!("failed to parse peer id from addr: {:?}", e));
```
SuperFluffy, line 193: Why `wrap_err_with` (or `map_err`) on `PeerId::from_multihash`?

crates/astria-gossipnet/src/network.rs
```diff
@@ -131,43 +136,119 @@ impl Network {
...
+                            }
+                        };
+
+                        if let Some(kademlia) = swarm.behaviour_mut().kademlia.as_mut() {
+                            kademlia.add_address(&peer_id, maddr);
+                        }
+                    }
+                    _ => {
```
SuperFluffy, line 201: IMO this is better to guard against refactoring:
```suggestion
                    other => {
```

crates/astria-gossipnet/src/network.rs
```diff
@@ -131,43 +136,119 @@ impl Network {
...
+                        return Err(eyre!("failed to parse peer id from addr: {}", addr));
```
SuperFluffy, line 202: Two notes:
1. use `bail!`
2. why format and return `addr` instead of `peer_id` that is being matched on?

crates/astria-gossipnet/src/network.rs
```diff
@@ -131,43 +136,119 @@ impl Network {
...
+}
+
+impl Network {
+    pub fn multiaddrs(&self) -> Vec<Multiaddr> {
+        self.multiaddrs.clone()
+    }
+
+    pub async fn bootstrap(&mut self) -> Result<()> {
+        if let Some(kademlia) = self.swarm.behaviour_mut().kademlia.as_mut() {
+            kademlia.bootstrap().map(|_| ()).map_err(|e| eyre!(e))
```
SuperFluffy, line 239: `wrap_err` with extra context instead of `map_err(|e| eyre!`

crates/astria-gossipnet/src/network.rs
```diff
@@ -131,43 +136,119 @@ impl Network {
...
     }
+}
+
+impl Default for NetworkBuilder {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+pub struct Network {
```
SuperFluffy, line 225: Please add a doc comment

crates/astria-gossipnet/src/network.rs
```diff
@@ -131,43 +136,119 @@ impl Network {
...
+    pub(crate) multiaddrs: Vec<Multiaddr>,
+    pub(crate) local_peer_id: PeerId,
+    pub(crate) swarm: Swarm<GossipnetBehaviour>,
+    pub(crate) terminated: bool,
+}
+
+impl Network {
+    pub fn multiaddrs(&self) -> Vec<Multiaddr> {
```
SuperFluffy, line 233: doc comment

crates/astria-gossipnet/src/network.rs
```diff
@@ -131,43 +136,119 @@ impl Network {
...
+        self.multiaddrs.clone()
+    }
+
+    pub async fn bootstrap(&mut self) -> Result<()> {
```
SuperFluffy, line 237: also here. What does this fn do?

crates/astria-gossipnet/src/network.rs
```diff
@@ -131,43 +136,119 @@ impl Network {
...
+        if let Some(kademlia) = self.swarm.behaviour_mut().kademlia.as_mut() {
+            kademlia.bootstrap().map(|_| ()).map_err(|e| eyre!(e))
+        } else {
+            bail!("kademlia is not enabled")
+        }
+    }
+
+    pub async fn random_walk(&mut self) -> Result<QueryId> {
```
SuperFluffy, line 245: what does this fn do?

crates/astria-gossipnet/src/network.rs
```diff
@@ -192,218 +273,8 @@ impl Network {
...
-                        return;
-                    }
-                }
-            }
-        });
 
-        let (res_a, res_b) = join!(alice_handle, bob_handle);
-        res_a.unwrap();
-        res_b.unwrap();
+    pub fn peer_count(&self) -> usize {
```
SuperFluffy, line 277: Doc comment. Also why not also just call it `num_peers`?

crates/astria-gossipnet/src/network_stream.rs
SuperFluffy: No need to refactor in this PR, but IMO creating types that are only meant to be used is a code smell.

crates/astria-gossipnet/src/network_stream.rs
```diff
@@ -0,0 +1,273 @@
...
+    RoutingUpdated(PeerId, Addresses),
+}
+
+impl futures::Stream for Network {
+    type Item = Result<Event>;
+
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        while let Poll::Ready(maybe_event) = self.swarm.poll_next_unpin(cx) {
+            let Some(event) = maybe_event else {
+                self.terminated = true;
```
SuperFluffy, line 62: Is this ever set to `false` actually?

crates/astria-gossipnet/src/network_stream.rs
```diff
@@ -0,0 +1,273 @@
...
+
+                    debug!("bootstrapping ok");
+                    None
+                }
+                _ => {
+                    debug!(query_id = ?id, result = ?result, "got query result");
+                    None
+                }
+            },
+            _ => {
```
SuperFluffy, line 267: To guard against refactoring: please match on `other` and use that in the log

crates/astria-gossipnet/src/network_stream.rs
```diff
@@ -0,0 +1,273 @@
...
+                            id,
+                            bootstrap.err()
+                        )));
+                    }
+
+                    debug!("bootstrapping ok");
+                    None
+                }
+                _ => {
+                    debug!(query_id = ?id, result = ?result, "got query result");
```
SuperFluffy, line 263: please match `other` and use that

crates/astria-gossipnet/Cargo.toml
```diff
@@ -4,12 +4,13 @@ version = "0.1.0"
...
 
 [dependencies]
-libp2p = { version = "0.51", features = ["gossipsub", "mdns", "noise", "macros", "ping", "tcp", "tokio", "yamux"] }
+libp2p = { version = "0.51", features = ["kad", "gossipsub", "identify", "mdns", "noise", "macros", "ping", "tcp", "tokio", "yamux"] }
+multiaddr = "0.17"
 
 color-eyre = { workspace = true }
 futures = { workspace = true }
-tokio = { workspace = true, features = ["macros", "rt-multi-thread", "signal"] }
+tokio = { workspace = true, features = ["macros", "rt-multi-thread", "signal", "sync"] }
```
SuperFluffy, line 12: Should this feature (actually, most of them) be under the normal dependencies? These look like they are only used as part of tests and live under `dev-dependencies`

SuperFluffy (approved):
Looks good to me. Approving this, but please address the comments re doc comments and handling errors.

crates/astria-gossipnet/src/network.rs
noot: yes probably :p

crates/astria-gossipnet/src/network.rs
noot: i guess there is a chance `peer_id` is not actually a peer ID, i'll rename it

crates/astria-gossipnet/src/network.rs
noot: see here: https://docs.rs/libp2p-kad/latest/libp2p_kad/struct.Kademlia.html#method.bootstrap it populates the routing table of the node's DHT

crates/astria-gossipnet/src/network.rs
noot: added a comment

crates/astria-gossipnet/src/network.rs
noot: arbitrary, I can rename

crates/astria-gossipnet/src/network_stream.rs
noot: no, it's only initially false by default, but never set to false again after being true.

noot: I'll remove this actually as it doesn't do anything atm

PR #37, opened 2023-05-22
Use pause:3.2 image which has arm and amd versions

steezeburger:
The latest `pause` image is quite old and does not support arm.
https://github.com/linkerd/linkerd2/issues/7558

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,5 @@
 target/
+
+
+# jetbrains
+.idea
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -1 +1,5 @@
 target/
+
+
+# jetbrains
+.idea
```
SuperFluffy, line 5: Quick comment on the draft: can you put this bit into a separate PR?

steezeburger: I understand wanting to keep PRs succinct and indexable, but I think this is fine in this instance. It's not related to our own code at all and feels excessive to spend extra time breaking it out into another PR.

SuperFluffy (approved):
That's a good find and apologies I didn't test this on macOS.

PR #38, opened 2023-05-22
turn per-crate into workspace dependencies

SuperFluffy:
Crate dependencies are made into workspace dependencies if used in more than one crate.


SuperFluffy:
Need to keep rand at 0.7 for now, because `ed25519-dalek:1.0.1` still requires it. It's bumped to 0.8 in `ed25519-dalek:rc.2`, but a) is that not a full release, and b) that breaks `libp2p-noise`'s `snow` dependency.

Once `libp2p` is updated to 0.52 we should be able to update rand 0.8.

noot (approved):

PR #44, opened 2023-05-29
feat(conductor): verify sequencer block commit

### LINKED ISSUE #16: validate tendermint commits on each block
when we read a block, we need to ensure that the block was actually finalized by the sequencer. since we have access to the sequencer's validator information, we need to ensure that the block's commit contains >2/3 validator stake signatures.

noot:
- verify that sequencer block commit has >2/3 voting power
- verify each `CommitSig` signature
- verify that `Commit` hashes to `last_commit_hash`

- note: newly proposed block contains the commit for the **previous** block!! should we still execute upon receiving a block, or only upon receiving the commit? or should we execute and mark it "head", then when we get the child block and the commit for the "head" block then mark it "softly" committed? I would have to do this in a follow-up as it will involve changing the gRPC methods.

closes #16


crates/astria-conductor/src/reader.rs
```diff
@@ -282,3 +349,235 @@ impl Reader {
...
+            commit.height
+        );
+    }
+
+    let mut total_voting_power = 0u64;
+    validator_set
+        .validators
+        .iter()
+        .try_for_each(|v| -> Result<()> {
+            total_voting_power += v.voting_power.parse::<u64>()?;
```
SuperFluffy, line 367: If voting power must be an integer, let's fix the type of `Validator` and make `voting_power: u64` so it already fails at ingestion, not deep into processing it.

SuperFluffy (changes requested):
`verify_commit` is a very complex function. Please also provide a test for the negative case (i.e. quorum not reached).

Also `get_new_blocks` is run as an async task in a loop within the async runtime (in `Reader::get_new_blocks`), which will stall the runtime at scale. Better run `verify_commit` in `tokio::task::spawn_blocking`

crates/astria-conductor/src/reader.rs
noot: unfortunately, the RPC returns this as a string, eg `"1000"`, so serde fails to deserialize it into `u64`. I added a custom type + deserialize implementation for now

SuperFluffy: ugh, that's terrible :(

noot:
@SuperFluffy comments addressed and added a test case for the quroum-not-reached case! will work on adding more test cases also

crates/astria-conductor/src/tendermint.rs
```diff
@@ -47,15 +49,15 @@ impl ValidatorSet {
 pub struct Validator {
     pub address: String,
     pub pub_key: KeyWithType,
-    pub voting_power: String,
-    pub proposer_priority: String,
+    pub voting_power: UintString,
```
SuperFluffy, line 52: Use the desired target types but use `serde(deserialize_with = "path::to::deser_fun")` (not tested, but should work for both integer types you are using):

```rust
#[derive(Deserialize, Debug)]
pub struct Validator {
    #[serde(deserialize_with = "deserialize_int_from_str")]
    voting_power: u64,
}

fn deserialize_int_from_str<'de, D, T>(deserializer: D) -> Result<T, D::Error>
where
    D: Deserializer<'de>,
    T: FromStr + serde::Deserialize<'de>,
    <T as FromStr>::Err: Display,
{
    let s: &str = Deserialize::deserialize(deserializer)?;
    s.parse::<T>().map_err(D::Error::custom)
}
```

Also, please provide a simple test case for `Validator` with these integers-as-strings set.

crates/astria-conductor/src/tendermint.rs
```diff
@@ -30,12 +36,8 @@ impl ValidatorSet {
...
     /// TODO: could there ever be two validators with the same priority?
     pub(crate) fn get_proposer(&mut self) -> eyre::Result<Validator> {
-        self.validators.sort_by(|v1, v2| {
-            v1.proposer_priority
-                .parse::<i64>()
-                .unwrap()
-                .cmp(&v2.proposer_priority.parse::<i64>().unwrap())
-        });
+        self.validators
+            .sort_by(|v1, v2| v1.proposer_priority.0.cmp(&v2.proposer_priority.0));
```
SuperFluffy, line 40: Not relevant for this PR but this `get_proposers` function is doing too much/should not have side effects. It should probably either just sort or use `iter().max_by` to find the proposer. Or at least be renamed.

But yeah, not relevant for this PR/

crates/astria-conductor/src/tendermint.rs
```diff
@@ -99,3 +101,43 @@ impl TendermintClient {
...
+            String::deserialize(deserializer)?
+                .parse::<i64>()
+                .map_err(|e| D::Error::custom(format!("{e}")))?
+                .try_into()
+                .map_err(|e| D::Error::custom(format!("{e}")))?,
+        ))
+    }
+}
+
+impl Serialize for UintString {
```
SuperFluffy, line 120: Do you need `serialize` given that you only derive `Deserialize` for the struct this is contained in?

crates/astria-conductor/src/reader.rs
```diff
@@ -282,3 +351,249 @@ impl Reader {
...
+) -> Result<()> {
+    if commit.height != validator_set.block_height {
+        bail!(
+            "commit height mismatch: expected {}, got {}",
+            validator_set.block_height,
+            commit.height
+        );
+    }
+
+    let mut total_voting_power = 0u64;
```
SuperFluffy, line 389: Do we need to be careful if the voting powers are too large, i.e. if their orders of magnitude are in the vicinity of u64::MAX? What protections do we have in place to guard against that?

crates/astria-conductor/src/reader.rs
```diff
@@ -282,3 +351,249 @@ impl Reader {
...
+    validator_set.validators.iter().for_each(|v| {
+        total_voting_power += v.voting_power.0;
```
SuperFluffy, line 391: maybe?
```suggestion
        total_voting_power += total_voting_power.saturating_add(v.voting_power.0);
```

crates/astria-conductor/src/reader.rs
```diff
@@ -282,3 +351,249 @@ impl Reader {
...
+            chain_id,
+            &validator.pub_key.key.0,
+            &vote.signature.0,
+        )?;
+
+        commit_voting_power += validator.voting_power.0;
+    }
+
+    ensure!(
+        commit_voting_power > total_voting_power * 2 / 3,
```
SuperFluffy, line 441: This has overflow potential. Prefer `total_voting_power / 3 * 2`.

I would love to have this as a free standing function and a bunch of unit tests just for it, i.e.:
```rust
fn does_commited_voting_power_have_quorum(commited: u64, total: u64) -> bool {
    commited > total / 3 * 2
}

**Review by SuperFluffy — COMMENTED** [review]
submitting comments; I forgot you are actively working on this

File: crates/astria-conductor/src/tendermint.rs (hunk 4 above)
Comment (noot) [reply crates/astria-conductor/src/tendermint.rs]: removed this type

File: crates/astria-conductor/src/reader.rs (hunk 5 above)
Comment (noot) [reply crates/astria-conductor/src/reader.rs:389]: I don't think that can happen, we are also receiving these values from a sequencer client that we're running or that we trust, so there should not be invalid values. however I'll double check what the values can be

File: crates/astria-conductor/src/reader.rs (hunk 6 above)
Comment (noot) [reply crates/astria-conductor/src/reader.rs]: see https://github.com/tendermint/tendermint/issues/2985

File: crates/astria-conductor/src/reader.rs (hunk 7 above)
Comment (noot) [reply crates/astria-conductor/src/reader.rs]: unit tests added!

File: crates/astria-conductor/src/reader.rs (hunk 5 above)
Comment (SuperFluffy) [reply crates/astria-conductor/src/reader.rs:389]: I would very much like to have some kind of early warning system in place just in case we f up something on the tendermint. It's true that we work with trusted inputs, but that does not mean that we should trust our inputs. :smile: 

The reason I am flagging this part is that it can silently break without us noticing.

What do you think about using `checked_add`?:

```rust
let Some(total_voting_power) = validator_set
    .validators
    .iter()
    .try_fold(0, |acc, &power| acc.checked_add(power)) else {
    bail!("total voting power exceeded u64:MAX");
};
```

crates/astria-conductor/src/tendermint.rs
```diff
@@ -29,33 +38,30 @@ impl ValidatorSet {
...
     /// the validator with the highest proposer priority is the proposer.
     /// TODO: could there ever be two validators with the same priority?
-    pub(crate) fn get_proposer(&mut self) -> eyre::Result<Validator> {
-        self.validators.sort_by(|v1, v2| {
-            v1.proposer_priority
-                .parse::<i64>()
-                .unwrap()
-                .cmp(&v2.proposer_priority.parse::<i64>().unwrap())
-        });
+    pub(crate) fn get_proposer(&self) -> eyre::Result<Validator> {
```
SuperFluffy, line 41: nice, ty for changing this

SuperFluffy (approved):
The changes look good, thank you very much for the tests.

I have added some thoughts about summing up the total voting power and provided a short snippet of code.

I am approving this PR but would very much like to have some kind of warning system in place *just in case* something weird is going on.

joroshiba (approved):
lgtm

PR #55, opened 2023-06-07
renamed prev_state_root in DoBlockRequest to prev_block_hash

akansjain:
In crates/astria-proto/proto/astria/execution/v1/execution.proto, there is a the gRPC method DoBlockRequest has a field called prev_state_root but what's actually passed here is the parent block hash, and should be called prev_block_hash.

Therefore, field prev_state_root is renamed to prev_block_hash in DoBlockRequest.


akansjain:
Hey @noot ,
Please review my PR.

joroshiba:
Thanks for the contribution, can you also update `astria-conductor` to use the new grpc as well [here](https://github.com/astriaorg/astria/blob/main/crates/astria-conductor/src/execution_client.rs#LL68C13-L68C28)

crates/astria-conductor/src/execution_client.rs
```diff
@@ -64,8 +64,8 @@ impl ExecutionClient for ExecutionRpcClient {
         timestamp: Option<Timestamp>,
     ) -> Result<DoBlockResponse> {
         let request = DoBlockRequest {
-            // TODO: this field name should actually be prev_block_hash!
-            prev_state_root: prev_block_hash,
+            // Updated field name: prev_block_hash
```
joroshiba, line 67: We can just remove this comment.

crates/astria-conductor/src/execution_client.rs
```diff
@@ -64,8 +64,8 @@ impl ExecutionClient for ExecutionRpcClient {
...
+            prev_block_hash: prev_block_hash.clone(),
```
joroshiba, line 68: I don't think the clone is necessary here, would prefer the shorthand.

noot, line 68:
```suggestion
            prev_block_hash,
```

crates/astria-conductor/src/execution_client.rs
joroshiba, line 67:
```suggestion
```

joroshiba (changes requested):

noot:
closing due to staleness

PR #57, opened 2023-06-07
Buf workflow only run when protos or workflow changes

### LINKED ISSUE #56: Update Buf workflow to only run for protos

joroshiba:
Fixes #56

This also establishes a pattern we can use elsewhere to limit test runs and builds of dependent images

### DIFF
--- a/.github/workflows/proto.yml
+++ b/.github/workflows/proto.yml
@@ -1,13 +1,29 @@
-name: buf
+name: Proto
 on:
   pull_request:
   push:
     branches:
         - main
 jobs:
-  build-lint-push-protos:
+  changed:
     runs-on: ubuntu-latest
-    environment: BUF
+    permissions:
+      pull-requests: read
+    outputs:
+      proto: ${{ steps.filters.outputs.proto }}
+    steps:
+      - uses: actions/checkout@v3
+      - uses: dorny/paths-filter@v2
+        id: filters
+        with:
+          filters: |
+            proto:
+              - 'crates/astria-proto/**'
+              - '.github/workflows/proto.yml'
+  lint:
+    runs-on: ubuntu-latest
+    needs: changed
+    if: ${{ needs.changed.outputs.proto == 'true' }}
     steps:
       - uses: actions/checkout@v3
       - uses: bufbuild/buf-setup-action@v1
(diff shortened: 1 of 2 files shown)


.github/workflows/proto.yml
```diff
@@ -1,13 +1,30 @@
...
+      proto: ${{ steps.filters.outputs.proto }}
+    steps:
+      - uses: actions/checkout@v3
+      - uses: dorny/paths-filter@v2
+        id: filters
+        with:
+          filters: |
+            proto:
+              - 'crates/astria-proto/**'
+              - '.github/workflows/proto.yml'
```
noot, line 22: what does this job do? it doesn't look like it runs anything?

joroshiba: It provides information about whether or not there were changes to the proto directory or build steps. The output then allows other jobs to run or be skipped based on the outcome.

I think having a seperate job for this is better than a step so it doesn't look like "build" passed when nothing was built.

noot: I see, should the `if: ${{ needs.changes.outputs.proto == 'true' }}` be added to the `push` job also?

joroshiba: Nope, it requires the build & lint to run

.github/workflows/proto.yml
```diff
@@ -1,13 +1,30 @@
...
+  
+  build-lint:
```
SuperFluffy, line 24: Nit: `lint-protos` might be more descriptive.

.github/workflows/proto.yml
```diff
@@ -1,13 +1,30 @@
-name: buf
+name: Proto
 on:
   pull_request:
   push:
     branches:
         - main
 jobs:
-  build-lint-push-protos:
+  changes:
```
SuperFluffy, line 8: Nit: `protos-changed:`?

SuperFluffy (approved):
Lgtm ;)

Just small nits

.github/workflows/proto.yml
joroshiba: Technically right now all it does is build and makes sure it completes. The lint is actually commented out since we are still in the "it's all alpha break it fine" stage.

.github/workflows/proto.yml
joroshiba: It's all under the "buf" job named "protos"?

noot (approved):

Issue #60, opened 2023-06-08
refactor sequencer-relayer


SuperFluffy:
```[tasklist]
### Tasks
- [x] allow configuring from environment variables (closed in #59)
- [ ] https://github.com/astriaorg/astria/issues/85
- [ ] https://github.com/astriaorg/astria/issues/86
- [ ] https://github.com/astriaorg/astria/issues/87
```

emhane:
can we use https://github.com/informalsystems/hermes ?

PR #70, opened 2023-06-09
init validator key during sequencer-relayer init

SuperFluffy:
This commit makes validator address and key
construction part of sequencer-relayer
initialization. The key and key constructor are
now crate private and thus an implementation
detail. Their only entrypoint is now through
passing a config to the relayer constructor.

Part of #60 

Note that the previous tests were checking if sha2 and bach32 worked correctly, not our use of them. They were therefore replaced.


crates/astria-sequencer-relayer/src/relayer.rs
```diff
@@ -21,37 +23,19 @@ use tracing::{
...
-
 pub struct Relayer {
     sequencer_client: SequencerClient,
     da_client: CelestiaClient,
     disable_writing: bool,
-    keypair: ed25519_dalek::Keypair,
-    validator_address: String,
-    validator_address_bytes: Vec<u8>,
+    validator: keys::Validator,
+    // validator_address: String,
```
joroshiba, line 37: Should these be removed?

SuperFluffy: Yes, thank you for catching

crates/astria-sequencer-relayer/src/keys.rs
```diff
@@ -1,92 +1,97 @@
...
-pub fn private_key_bytes_to_keypair(private_key: &[u8]) -> eyre::Result<Keypair> {
-    let private_key = ed25519_dalek::SecretKey::from_bytes(&private_key[..32])
-        .wrap_err("failed reading secret key from bytes")?;
-    let public_key = PublicKey::from(&private_key);
-    Ok(Keypair {
-        secret: private_key,
-        public: public_key,
-    })
+#[derive(Debug)]
+pub(crate) struct Validator {
```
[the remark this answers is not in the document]
SuperFluffy: Agree, I wanted to `git mv keys.rs validator.rs` or something like that, but it made reviewing things impossible (because everything was just green). I wonder if that has changed.

crates/astria-conductor/src/reader.rs
```diff
@@ -282,3 +281,27 @@ impl Reader {
         Ok(())
     }
 }
+
+fn public_key_to_bach32_address(key: &[u8]) -> Result<String> {
+    use sha2::{
+        digest::typenum::Unsigned as _,
+        Digest as _,
+    };
+    const ADDRESS_LENGTH: usize = 20;
```
noot, line 290: can use this also https://github.com/informalsystems/tendermint-rs/blob/32dad520d8996da0c0bf5b0fc8767fe3fd9fab8e/tendermint/src/account.rs#L19 this won't be changing though so doesn't really matter

crates/astria-sequencer-relayer/src/keys.rs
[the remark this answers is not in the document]
noot: yeah I'd rename this to `ValidatorFile` or something like that, or add a comment about where this came from

crates/astria-sequencer-relayer/src/keys.rs
```diff
@@ -1,92 +1,97 @@
...
+    pub(crate) address: Vec<u8>,
+    pub(crate) bach32_address: String,
```
noot, line 23:
```suggestion
    pub(crate) bech32_address: String,
```

crates/astria-sequencer-relayer/src/keys.rs
```diff
@@ -1,92 +1,97 @@
...
+        }
+
+        let f = File::open(&path).wrap_err("failed to open validator key file")?;
+        let Ok(read_keys): Result<JsonRepr, _> = serde_json::from_reader(f) else {
+            // silently drop the error to ensure the deserializer never leaks the keys
+            // now or in the future.
+            bail!("failed to read the validator key from file");
+        };
+        let keypair = Keypair::from_bytes(&read_keys.priv_key.value)
+            .wrap_err("failed generating ed25519 dalek keypair from validator private key")?;
```
noot, line 54:
```suggestion
            .wrap_err("failed generating ed25519 keypair from validator private key")?;
```

i think the dalek part is just naming they gave the crate :p

crates/astria-sequencer-relayer/src/keys.rs
```diff
@@ -1,92 +1,97 @@
...
+            priv_key: Key,
+        }
+
+        let f = File::open(&path).wrap_err("failed to open validator key file")?;
+        let Ok(read_keys): Result<JsonRepr, _> = serde_json::from_reader(f) else {
+            // silently drop the error to ensure the deserializer never leaks the keys
+            // now or in the future.
+            bail!("failed to read the validator key from file");
+        };
+        let keypair = Keypair::from_bytes(&read_keys.priv_key.value)
```
noot, line 53: does this work? I don't think it will, this function is expecting 64 bytes https://docs.rs/ed25519-dalek/latest/ed25519_dalek/constant.KEYPAIR_LENGTH.html

noot: need to pass in `priv_key | pub_key` (concatenated)

crates/astria-sequencer-relayer/src/keys.rs
```diff
@@ -1,92 +1,97 @@
...
-pub fn public_key_to_address(public_key: &[u8]) -> eyre::Result<String> {
-    let mut hasher = Sha256::new();
-    hasher.update(public_key);
-    let result = hasher.finalize();
-    let address_bytes: [u8; ADDRESS_LENGTH] = result[0..ADDRESS_LENGTH]
-        .try_into()
-        .expect("should convert slice to array of 20 bytes");
-    let address = bech32::encode("metrovalcons", address_bytes.to_base32(), Variant::Bech32)
+        let bach32_address = bech32::encode(
+            "metrovalcons",
```
noot, line 61: make this a const somewhere?

crates/astria-sequencer-relayer/src/keys.rs
SuperFluffy: This was wrong indeed. Thank you for noticing! I now avoid the issue entirely by leaning on the `PrivValidatorKey` type from `tendermint-config`.

SuperFluffy:
@noot I realized that this was re implementing a ton of logic that was already provided by tendermint and tendermint-config. This in turn triggered some more changes, resulting in us moving from `ed25519-dalek` to `ed25519-consensus` (from penumbra).

I left some comments in crates/astria-sequencer-relayer/src/validator.rs where I noticed that we have some difference from how tendermint is doing things.

crates/astria-sequencer-relayer/src/relayer.rs
```diff
@@ -21,37 +23,17 @@ use tracing::{
...
 use crate::{
-    base64_string::Base64String,
     data_availability::CelestiaClient,
-    keys::{
-        private_key_bytes_to_keypair,
-        validator_hex_to_address,
-    },
     sequencer::SequencerClient,
     sequencer_block::SequencerBlock,
+    validator,
```
joroshiba, line 29: is there a reason this can't be `validator::Validator` shortening up the references below?

crates/astria-sequencer-relayer/src/validator.rs
```diff
@@ -0,0 +1,106 @@
...
+  },
+  "priv_key": {
+    "type": "tendermint/PrivKeyEd25519",
+    "value": "1hBYYTBKxkMODNTW6Pk//kA023UAkpgSLhM0SjwndSkWP/Y3MbjTh/4KMclROExc1LvEljkYw+4VWUowpvzowA=="
+  }
+}
+"#;
+
+    #[test]
+    fn valid_validator_keys_can_be_read_from_file() {
```
joroshiba, line 101: Should we have a tests for what happens with invalid file in terms of format as well as non-existant file?

SuperFluffy: I don't think so. Thought about this and all tests I can come up with would be testing the external routines we are calling with extra steps.

crates/astria-sequencer-relayer/src/validator.rs
```diff
@@ -0,0 +1,105 @@
...
+    /// The bech32-encoded validator account address with "metrovalcons" as
+    /// as the human readable prefix.
+    ///
+    /// FIXME: Add a note on what this is used for.
+    /// NOTE 1: why apply bech32 to the validator *address* instead of its
+    ///         verification (i.e. public) key? tendermint provides `PublicKey::to_bech32`
+    /// NOTE 2: tendermint's `PublicKey::to_bech32` prepends the pubkey with extra bytes
+    ///         to make it amino compatible. Do we need this too?
+    #[zeroize(skip)]
+    pub(crate) bech32_address: String,
```
noot, line 36: i think we can remove all bech32 addresses in a follow-up as they're cosmos-specific

crates/astria-sequencer-relayer/src/validator.rs
```diff
@@ -0,0 +1,105 @@
...
+use zeroize::{
+    Zeroize,
+    ZeroizeOnDrop,
+};
+
+const BECH32_HUMAN_READABLE_PREFIX: &str = "metrovalcons";
+
+/// `Validator` holds the ed25519 keys to sign and verify tendermint
+/// messages. It also contains its address in the tendermint network
+/// and a bech32 encoded address that is used when communicating with metro.
```
noot, line 19:
```suggestion
/// and a bech32 encoded address that is used on the metro network.
```

crates/astria-sequencer-relayer/src/validator.rs
```diff
@@ -0,0 +1,105 @@
...
+    pub(crate) signing_key: ed25519_consensus::SigningKey,
+    #[zeroize(skip)]
+    /// The ed25519 verification key of this validator.
+    pub(crate) verification_key: ed25519_consensus::VerificationKey,
+}
+
+impl Validator {
+    /// Constructs a `Validator` from a json formatted tendermint private validator key.
+    ///
+    /// This file is frequently called `private_validator_key.json` and is generated durign
```
noot, line 47:
```suggestion
    /// This file is frequently called `private_validator_key.json` and is generated during
```

noot (approved):
looks good, only a few minor comments!

crates/astria-sequencer-relayer/src/validator.rs
SuperFluffy: Opened #97

joroshiba (approved):

Issue #42, opened 2023-05-25
Generate OCI images on github actions in monorepo


steezeburger:
Will require some Github setup so that `astria` has permission to push to the package registry

Can we only build images for a specific crate when that crate actually changes?
Should we build on every push? Only pushes to main? This is something we need to figure out.

PR #71, opened 2023-06-09
Use github flavored globbing for gh actions

steezeburger:


.github/workflows/docker-build.yml
```diff
@@ -7,10 +7,10 @@ on:
     branches:
       - "main"
     tags:
-      - "v[0-9]+.[0-9]+.[0-9]+--.*"
-      - "v[0-9]+.[0-9]+.[0-9]+-alpha.[0-9]+--.*"
-      - "v[0-9]+.[0-9]+.[0-9]+-beta.[0-9]+--.*"
-      - "v[0-9]+.[0-9]+.[0-9]+-rc[0-9]+--.*"
+      - "v[0-9]+.[0-9]+.[0-9]+--[a-z]+"
```
SuperFluffy, line 10: I *think* (please double check) that this is only going to work if you include a dash because our crates all have a dash.
```suggestion
      - "v[0-9]+.[0-9]+.[0-9]+--[a-z-]+"
```

steezeburger: Good catch! I only thought through this with `conductor` in mind. The tags aren't using the crate names, they are using the image names, e.g. `v0.0.1--conductor` and `v0.0.1--sequencer-relayer`, but `sequencer-relayer` definitely has a hyphen in it.

I think your suggestion is correct according to [the docs ](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet)

steezeburger: Wait actually, `conductor` doesn't have a hyphen in it, so I think it was correct originally since it would match any version of `v0.0.1--some-more-words` or `v0.0.1--word`

steezeburger: or maybe we should add `**`?
```suggestion
      - "v[0-9]+.[0-9]+.[0-9]+--[a-z]+**"
```

joroshiba (approved):

jbowen93 (approved):

steezeburger:
In support of https://github.com/astriaorg/astria/issues/42

PR #72, opened 2023-06-11
feat(sequencer): implement cryptographically-backed accounts and transaction signing

### LINKED ISSUE #64: sequencer: implement account keys and signing scheme
currently we will be using ed25519 to start with.

requires:
- creating address scheme based off public keys (probably just first 20 bytes of `sha256(public_key_bytes)` like every other chain for now)
- adding signature and public key of sender to transactions
- implement signing of transactions and signature verification of signed transactions

noot:
- implement `SignedTransaction` type, previous `Transaction` type is renamed to `UnsignedTransaction`
- update `Address` type to be first 20 bytes of sha256 hash of public key instead of just a string
- implementing signing and verification for txs
- verify tx signature in `check_stateless`
- transaction `from` address is now derived from the transaction signer public key
- question: how to handle keys properly for the test accounts? just hard-code them all somewhere?

closes #64


crates/astria-sequencer/src/accounts/types.rs
```diff
@@ -89,6 +164,45 @@ impl std::ops::Sub<u128> for Balance {
     }
 }
 
+impl Balance {
+    pub fn as_proto(&self) -> ProtoBalance {
+        ProtoBalance {
+            hi: (self.0 >> 64) as u64,
```
joroshiba, line 170: Should these make use of the functions on the primitive?

crates/astria-sequencer/src/crypto.rs
```diff
@@ -0,0 +1,46 @@
+pub(crate) use ed25519_dalek::{
+    Keypair,
+    PublicKey,
+    Signature,
+    Signer,
+    Verifier,
+};
+
+#[cfg(test)]
+mod test {
```
joroshiba, line 10: Are these tests just testing that a limited subset of an external crate work as anticipated?

crates/astria-sequencer/src/transaction/signed_transaction.rs
```diff
@@ -0,0 +1,155 @@
...
+    pub fn signer_address(&self) -> Result<Address> {
+        Address::try_from(&self.public_key)
+    }
+
+    #[must_use]
+    pub fn to_vec(&self) -> Vec<u8> {
+        let proto = ProtoSignedTransaction {
+            transaction: Some(match &self.transaction {
+                UnsignedTransaction::AccountsTransaction(tx) => {
+                    astria_proto::sequencer::v1::UnsignedTransaction {
```
joroshiba, line 68: should we alias this name as `ProtoUnsignedTransaction`?

crates/astria-sequencer/src/transaction/signed_transaction.rs
```diff
@@ -0,0 +1,155 @@
...
+                        value: Some(astria_proto::sequencer::v1::unsigned_transaction::Value::AccountsTransaction (
```
joroshiba, line 69: what's the difference between `unsigned_transaction` here and `UnsignedTransaction` above?

crates/astria-sequencer/src/accounts/types.rs
noot: yes! will update

crates/astria-sequencer/src/crypto.rs
noot: no, basically for testing we need test accounts, and if they're deterministic it makes our lives a lot easier. this "test" basically just generates the test accounts. I don't really think this should be here, but I'm not sure where they should live exactly, as I don't want to just hard code some keys in without showing where they came from

crates/astria-sequencer/src/transaction/signed_transaction.rs
noot: when prost generates this type:
```
message UnsignedTransaction {
    oneof value {
        AccountsTransaction accounts_transaction = 1;
    }   
}
```

it creates a module `unsigned_transaction` as follows to deal with the `oneof`:
```
/// Nested message and enum types in `UnsignedTransaction`.
pub mod unsigned_transaction {
    #[allow(clippy::derive_partial_eq_without_eq)]
#[derive(Clone, PartialEq, ::prost::Oneof)]
    pub enum Value {
        #[prost(message, tag="1")]
        AccountsTransaction(super::AccountsTransaction),
    }
}
```

`UnsignedTransaction` is a struct that contains a `Value`:
```
pub struct UnsignedTransaction {
    #[prost(oneof="unsigned_transaction::Value", tags="1")]
    pub value: ::core::option::Option<unsigned_transaction::Value>,
}
```

crates/astria-sequencer/src/crypto.rs
joroshiba: Gotcha, could we put a comment to that effect here? Not sure of a better place either.

crates/astria-sequencer/src/accounts/types.rs
joroshiba, line 1: Might be good to have more consistency in how to get our `Ok(Address({stuff}))` in here. Some are declaring into a separate array others pushing directly into the address. Personally I think the arr is a little more readable.

crates/astria-sequencer/src/crypto.rs
```diff
@@ -0,0 +1,46 @@
...
+    use super::*;
+    use crate::accounts::types::Address;
+
+    #[test]
+    fn generate_default_keys() {
+        let default_accounts: Vec<&[u8]> = vec![b"alice", b"bob", b"carol"];
+        let message = b"test";
+
+        for acc in default_accounts {
+            let mut hasher = sha2::Sha256::new();
```
joroshiba, line 23: Should this use the hasher utility function?

joroshiba (changes requested):
It looks like there might be a number of places we could use that new hasher utility function?

crates/astria-sequencer/src/accounts/types.rs
noot: sorry which lines/functions is this referring to specifically?

joroshiba: Lines 25, 26, 48, 99 all do something like `Ok(Address(arr))`

Lines 60, 76 have logic inside the `Ok(Address())` call

crates/astria-sequencer/src/accounts/types.rs
```diff
@@ -7,24 +9,97 @@ use serde::{
...
-    fn from(s: &str) -> Self {
-        Self(s.to_string())
+impl TryFrom<&crate::crypto::Keypair> for Address {
+    type Error = anyhow::Error;
+
+    fn try_from(keypair: &crate::crypto::Keypair) -> Result<Self, Self::Error> {
+        use sha2::Digest as _;
+        let mut hasher = sha2::Sha256::new();
+        hasher.update(keypair.public.as_bytes());
+        let hash = hasher.finalize();
```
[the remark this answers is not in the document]
noot: updated

crates/astria-proto/proto/astria/sequencer/v1/transaction.proto
```diff
@@ -0,0 +1,23 @@
+syntax = "proto3";
+
+package astria.sequencer.v1;
+
+import "astria/primitive/v1/types.proto";
+
+message SignedTransaction {
```
SuperFluffy, line 7: Please add comments on these.

crates/astria-proto/proto/astria/sequencer/v1/transaction.proto
```diff
@@ -0,0 +1,23 @@
...
+    UnsignedTransaction transaction = 3;
+}
+
+message UnsignedTransaction {
+    oneof value {
+        AccountsTransaction accounts_transaction = 1;
+    }   
+}
+
+message AccountsTransaction {
```
SuperFluffy, line 19: Judging from the implementation code, it is an error to leave e.g. `amount` unset. Can you also add that information here?

crates/astria-sequencer/src/accounts/transaction.rs
```diff
@@ -3,53 +3,108 @@ use anyhow::{
...
+                .amount
+                .as_ref()
+                .ok_or_else(|| anyhow::anyhow!("missing amount"))?
+                .into(),
+            nonce: Nonce::from(proto.nonce),
+        })
+    }
+}
+
+impl TryFrom<ProtoAccountsTransaction> for Transaction {
```
SuperFluffy, line 62: Can we get rid of all these `TryFrom` impls in favour of just using the `from_proto` and `to_proto` methods on the types? That would make it much easier to discover these methods and how they work.

I also don't see how these methods being used in a generic context - so let's not unnecessarily introduce interfaces.

crates/astria-sequencer/src/accounts/transaction.rs
```diff
@@ -3,53 +3,108 @@ use anyhow::{
...
-    fn check_stateless(&self) -> Result<()> {
+impl TryFrom<&ProtoAccountsTransaction> for Transaction {
+    type Error = anyhow::Error;
+
+    fn try_from(proto: &ProtoAccountsTransaction) -> Result<Self> {
+        Self::from_proto(proto)
+    }
+}
+
+impl From<Transaction> for ProtoAccountsTransaction {
```
SuperFluffy, line 78: Same for the From impls

crates/astria-sequencer/src/transaction/signed_transaction.rs
```diff
@@ -0,0 +1,155 @@
...
+    hash,
+    transaction::{
+        ActionHandler,
+        TransactionHash,
+        UnsignedTransaction,
+    },
+};
+
+#[derive(Debug, Clone)]
+pub struct SignedTransaction {
```
SuperFluffy, line 39: private

crates/astria-sequencer/src/transaction/signed_transaction.rs
```diff
@@ -0,0 +1,155 @@
...
+    ///
+    /// # Errors
+    ///
+    /// - If the public key cannot be converted into an address
+    pub fn signer_address(&self) -> Result<Address> {
+        Address::try_from(&self.public_key)
+    }
+
+    #[must_use]
+    pub fn to_vec(&self) -> Vec<u8> {
```
SuperFluffy, line 68: The fact that `to_vec` here (and for the types) actually serializes to protobuf is very surprising and unexpected. 

From a first glance it also looks like these `to_vec` are only used in `hash` calls`?

Instead of these indirections, please just make it explicit at the call site: prefer something like `Sha256::hash(signed_transaction.to_proto())

crates/astria-sequencer/src/transaction/mod.rs
```diff
@@ -0,0 +1,14 @@
+pub mod action_handler;
```
SuperFluffy, line 1: please make all of this private

SuperFluffy (changes requested):
First round of reviews:

please be liberal with commenting new types and their methods.
On a first pass, I would remove all `TryFrom` impls and prefer direct conversion methods on the types. This would make it much easier to discover.

Also please make all types crate-private by default, except for those types that are used as entrypoints into `Sequencer.`

crates/astria-sequencer/src/accounts/state_ext.rs
```diff
@@ -76,14 +76,14 @@ pub(crate) trait StateWriteExt: StateWrite {
...
-        self.put_raw(balance_storage_key(address.to_str()), bytes);
+        self.put_raw(balance_storage_key(&address.to_string()), bytes);
         Ok(())
     }
 
     #[instrument(skip(self))]
     fn put_account_nonce(&mut self, address: &Address, nonce: Nonce) -> Result<()> {
         let bytes = nonce.try_to_vec().context("failed to serialize nonce")?;
-        self.put_raw(nonce_storage_key(address.to_str()), bytes);
+        self.put_raw(nonce_storage_key(&address.to_string()), bytes);
```
SuperFluffy, line 86: In the same light as my other comments, why not pass `address` to `nonce_storage_key` and let it do the bytes-to-hex conversion?

crates/astria-proto/src/lib.rs
```diff
@@ -80,6 +80,16 @@ mod primitive_impls {
         }
     }
 
+    impl From<&Uint128> for u128 {
```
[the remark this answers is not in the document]
noot: sorry, where is the `prost` build config? don't see it in the build.rs file

crates/astria-proto/proto/astria/sequencer/v1/transaction.proto
noot: you mean having a zero-value transfer (or leaving the other values as zero)? that's allowed, except for nonce, which should be 1 for an account's first tx

crates/astria-sequencer/src/accounts/types.rs
```diff
@@ -7,24 +9,89 @@ use serde::{
...
+    fn try_from(s: &str) -> std::result::Result<Self, Self::Error> {
+        let bytes = hex::decode(s)?;
+        let arr: [u8; ADDRESS_LEN] = bytes
+            .try_into()
+            .map_err(|_| anyhow!("invalid address hex length"))?;
+        Ok(Address(arr))
+    }
+}
+
+impl TryFrom<&crate::crypto::PublicKey> for Address {
```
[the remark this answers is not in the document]
noot: in what way? would it be better to implement a `from_public_key()` for Address instead?

crates/astria-sequencer/src/transaction/unsigned_transaction.rs
```diff
@@ -0,0 +1,113 @@
...
+        Ok(match value {
+            ProtoAccountsTransaction(tx) => {
+                Self::AccountsTransaction(AccountsTransaction::from_proto(&tx)?)
+            }
+        })
+    }
+
+    /// Signs the transaction with the given keypair.
+    #[must_use]
+    pub fn sign(self, keypair: &Keypair) -> SignedTransaction {
```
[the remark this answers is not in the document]
noot: I made this public since they'll eventually be used by outside crates (ie. the sequencer signing client library) but will make private for now

crates/astria-sequencer/src/transaction/signed_transaction.rs
noot: this brings up a question; how should we serialize transactions that are sent over the network/RPC? protobuf or borsh? i assumed protobuf for now since that's sort of cosmos-standard but we don't actually need to stick with that.

joroshiba (approved):

crates/astria-proto/proto/astria/sequencer/v1/transaction.proto
SuperFluffy: I meant from a protobuf3 (+ rust) point of view. Since in protobuf3 technically all fields are optional (meaning it's not an error to not send them), Rust represents `amount` as an `Option`.

I think creating an `AccountsTransaction` without `amount` set is invalid from a business logic point of view, so I think we should document this here (probably in all our protobufs, but here is the first time I really noticed).

crates/astria-proto/src/lib.rs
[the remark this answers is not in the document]
SuperFluffy: You need to set it in the buf build yaml (toml?) file.

crates/astria-sequencer/src/transaction/hash.rs
```diff
@@ -0,0 +1,26 @@
+use anyhow::ensure;
+
+/// Represents the sha256 hash of an encoded transaction.
+pub(crate) struct Hash([u8; 32]);
```
SuperFluffy, line 4: Please add methods on `Hash` instead of relying on the `TryFrom` impls, i.e.

```rust
fn from_bytes(&[u8]) -> Result<Self> {
}
```

noot: just gonna delete this file cause it's not actually used anymore

crates/astria-sequencer/src/transaction/signed_transaction.rs
joroshiba: I'd prefer we stick to what is more "standard" as well as just minimizing the number of serialization formats we use across our stack. Looking at [performance comparisons](https://github.com/djkoloski/rust_serialization_benchmark), it looks like borsch would save us some time at the cost of space, saving over wire bandwidth, is probably worthwhile is my gut feel.

crates/astria-sequencer/src/transaction/unsigned.rs
```diff
@@ -0,0 +1,122 @@
...
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[non_exhaustive]
+pub(crate) enum Transaction {
+    AccountsTransaction(AccountsTransaction),
+}
+
+impl Transaction {
+    /// Attempts to encode the unsigned transaction into bytes.
+    #[must_use]
+    pub(crate) fn to_vec(&self) -> Vec<u8> {
```
SuperFluffy, line 30: I find this naming very confusing. Please just make this `fn to_proto -> ProtoUnsignedTransaction` and let the conversion to protobuf be explicit.

crates/astria-sequencer/src/transaction/unsigned.rs
```diff
@@ -0,0 +1,122 @@
...
+        match &self {
+            Transaction::AccountsTransaction(tx) => ProtoUnsignedTransaction {
+                value: Some(ProtoAccountsTransaction(tx.to_proto())),
+            },
+        }
+        .encode_length_delimited_to_vec()
+    }
+
+    /// Signs the transaction with the given keypair.
```
SuperFluffy, line 39:
```suggestion
    /// Signs the transaction with a signing key
```

crates/astria-sequencer/src/crypto.rs
```diff
@@ -0,0 +1,41 @@
...
+        let default_accounts: Vec<&[u8]> = vec![b"alice", b"bob", b"carol"];
+        let message = b"test";
+
+        for acc in default_accounts {
+            let bytes: [u8; 32] = hash(acc).try_into().unwrap();
+            let secret = SigningKey::from(bytes);
+            let public = secret.verification_key();
+
+            let sig = secret.sign(message);
+            assert!(public.verify(&sig, message).is_ok());
```
SuperFluffy, line 30: maybe `unwrap` instead so that the we get more info why this failed

crates/astria-sequencer/src/transaction/unsigned.rs
```diff
@@ -0,0 +1,122 @@
...
+    impl Transaction {
+        /// Attempts to decode an unsigned transaction from the given bytes.
+        ///
+        /// # Errors
+        ///
+        /// - If the bytes cannot be decoded into the prost-generated `UnsignedTransaction` type
+        /// - If the value is missing
+        /// - If the value is not a valid transaction type (ie. does not correspond to any
+        ///   component)
+        fn try_from_slice(bytes: &[u8]) -> Result<Self> {
```
SuperFluffy, line 91: As in the other case: I think the proto-deserialization should be made explicit, so this fn would be better as just a `Transaction::try_from_proto`

crates/astria-sequencer/src/transaction/unsigned.rs
```diff
@@ -0,0 +1,122 @@
...
+        match &self {
+            Transaction::AccountsTransaction(tx) => ProtoUnsignedTransaction {
+                value: Some(ProtoAccountsTransaction(tx.to_proto())),
+            },
+        }
+        .encode_length_delimited_to_vec()
+    }
+
+    /// Signs the transaction with the given keypair.
+    #[allow(dead_code)]
```
SuperFluffy, line 40: Not a fan of allowing dead code for the sake of running tests....

Can you just make this a free-standing fn inside the test?

crates/astria-sequencer/src/transaction/unsigned.rs
```diff
@@ -0,0 +1,122 @@
...
+    pub(crate) fn sign(self, secret_key: &SigningKey) -> SignedTransaction {
+        let signature = secret_key.sign(&self.hash());
+        SignedTransaction {
+            transaction: self,
+            signature,
+            public_key: secret_key.verification_key(),
+        }
+    }
+
+    pub(crate) fn hash(&self) -> Vec<u8> {
```
SuperFluffy, line 51: Please remove. `util::hash(&my_transaction)` isn't much more work but one less indirection to chase down.

crates/astria-sequencer/src/utils.rs
```diff
@@ -0,0 +1,6 @@
+pub(crate) fn hash(s: &[u8]) -> Vec<u8> {
```
SuperFluffy, line 1: Please add a small blurb, e.g.:
```rust
/// Utility function to hash a byte slice using sha256
```

crates/astria-proto/buf.gen.yaml
```diff
@@ -6,5 +6,6 @@ plugins:
     out: .
     opt:
      - extern_path=.tendermint.types=::tendermint-proto::types
+     - type_attribute=.astria.primitive.v1.Uint128=#[derive(Copy)]
```
SuperFluffy, line 9: Kinda nice.

I wonder how maintainable this will be in the long run should we ever want to add more type attributes or other things. But it's what we are given. :sweat_smile: : :shrug:

SuperFluffy (changes requested):
I think this PR is going in the right direction.

There are still a few changes I would like to see. Especially the implicit protobuf serialization/deserializations I would love to make explicit (both in conversion methods like `try_from_slice`), but also by removing `TryFrom` and other impls.

Can you also add a few simple tests for the conversion methods from/to proto? In my personal experience that's where I always found a lot of gotchas for my own code.

crates/astria-proto/proto/astria/sequencer/v1/transaction.proto
```diff
@@ -0,0 +1,29 @@
...
+}
+
+// `UnsignedTransaction` is a transaction that does not have an attached signature.
+message UnsignedTransaction {
+    oneof value {
+        AccountsTransaction accounts_transaction = 1;
+    }   
+}
+
+// `AccountsTransaction` represents a value transfer transaction.
```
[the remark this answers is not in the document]
noot: like just a `\n` at the end?

crates/astria-sequencer/src/accounts/types.rs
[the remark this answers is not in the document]
SuperFluffy: Sorry, didn't see. Yeah, that would be better. But even more so: for the key it would be good to call it smth like `Address:derive_from_public_key` or something like that.

crates/astria-proto/proto/astria/sequencer/v1/transaction.proto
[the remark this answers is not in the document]
SuperFluffy: Yeah, like an empty space. I.e.

```rust
// Title
//
// More stuff.
```

crates/astria-sequencer/src/accounts/types.rs
```diff
@@ -117,6 +192,12 @@ impl From<u32> for Nonce {
     }
 }
 
+impl From<Nonce> for u32 {
```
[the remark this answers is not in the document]
noot: like turning this into just methods on `Nonce`?

crates/astria-sequencer/src/transaction/signed.rs
```diff
@@ -0,0 +1,127 @@
...
+}
+
+impl Transaction {
+    /// Verifies the transaction signature.
+    /// The transaction signature message is the hash of the transaction.
+    ///
+    /// # Errors
+    ///
+    /// - If the signature is invalid
+    pub(crate) fn verify_signature(&self) -> Result<()> {
```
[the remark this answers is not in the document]
noot: a `SignedTransaction` is constructed in `check_tx` and `deliver_tx` (from external input) and also using `UnsignedTransactions::sign()`. so in the first two cases, we would have to do the signature verification inside `SignedTransaction::try_from_slice()`?

crates/astria-sequencer/src/accounts/types.rs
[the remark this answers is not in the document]
SuperFluffy: Yeah. So this would be `to_u32`.

I know I didn't do this for `U128`, but it felt justified because that's for converting between primitive-likes.

For things like `Nonce` I feel like there is more structure to justify these conversions.

crates/astria-sequencer/src/transaction/signed.rs
```diff
@@ -0,0 +1,127 @@
...
+            .verify(&self.signature, &self.transaction.hash())
+            .context("failed to verify transaction signature")
+    }
+
+    /// Returns the address which signed the transaction.
+    ///
+    /// # Errors
+    ///
+    /// - If the public key cannot be converted into an address
+    pub(crate) fn signer_address(&self) -> Result<Address> {
```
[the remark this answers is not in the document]
noot: we don't want to serialize it so we'd have to make two transaction types - one for including `address`, and one without it for sending over the wire which I don't really like. but yes conversion from `VerificationKey` to `Address` can never fail, I can update that function to not return `Result`

crates/astria-sequencer/src/transaction/signed.rs
```diff
@@ -0,0 +1,127 @@
...
+    ///
+    /// # Errors
+    ///
+    /// - If the bytes cannot be decoded into the prost-generated `SignedTransaction` type
+    /// - If the transaction value is missing
+    /// - If the transaction value is not a valid transaction type (ie. does not correspond to any
+    ///   component)
+    /// - If the signature cannot be decoded
+    /// - If the public key cannot be decoded
+    pub(crate) fn try_from_slice(bytes: &[u8]) -> Result<Self> {
```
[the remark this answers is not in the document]
noot: I think the `ProtoSignedTransaction` should probably be re-exported since importing it from astria_proto every time might be annoying, what do you think? 

would renaming this to `try_from_proto_bytes` or something also be better?

crates/astria-sequencer/src/transaction/signed.rs
```diff
@@ -0,0 +1,127 @@
...
+        };
+        Ok(signed_tx)
+    }
+}
+
+#[async_trait]
+impl ActionHandler for Transaction {
+    #[instrument]
+    fn check_stateless(&self) -> Result<()> {
+        self.verify_signature()?;
```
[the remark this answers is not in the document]
noot: replied above

crates/astria-sequencer/src/transaction/unsigned.rs
noot: updated, but I kind of like the `try_from_slice` because we won't need to import from `astria_proto` in any crates that eventually use this, as well as `prost::Message` for encoding/decoding. thoughts?

crates/astria-sequencer/src/transaction/unsigned.rs
noot: this is used in multiple unit tests so I didn't want to copy-paste, as well it's used here in the client crate: https://github.com/astriaorg/astria/pull/129 so it's no longer dead code

crates/astria-sequencer/src/transaction/unsigned.rs
noot: you mean to add a helper function that will take in a `Transaction` and hash it? if I remove this right now, I need to do

```
let signature = secret_key.sign(&crate::hash(&self.to_proto().encode_length_delimited_to_vec()));
```

which I don't really like

crates/astria-sequencer/src/transaction/unsigned.rs
SuperFluffy: ok

crates/astria-sequencer/src/transaction/signed.rs
```diff
@@ -0,0 +1,127 @@
...
+    accounts::{
+        transaction::Transaction as AccountsTransaction,
+        types::Address,
+    },
+    crypto::{
+        Signature,
+        VerificationKey,
+    },
+    transaction::{
+        unsigned::Transaction as UnsignedTransaction,
```
SuperFluffy, line 28: I realize this naming came as a fix from the clippy lint. If you rename `unsigned::Transaction` to `unsigned::Unsigned` and reexport it under `transaction`, you can refer to it as `transaction::Unsigned` without the exra `transaction::unsigned::Transaction as UnsignedTransaction` dance.

Apologies for not having made this clear last week. :(

crates/astria-sequencer/src/transaction/signed.rs
```diff
@@ -0,0 +1,127 @@
...
+        ActionHandler,
+    },
+};
+
+/// Represents a transaction signed by a user.
+/// It contains the signature and the public key of the user,
+/// as well as the transaction itself.
+#[derive(Debug, Clone)]
+pub(crate) struct Transaction {
```
SuperFluffy, line 37:
```suggestion
pub(crate) struct Signed {
```

crates/astria-sequencer/src/transaction/signed.rs
```diff
@@ -0,0 +1,127 @@
...
+    pub(crate) signature: Signature,
+    pub(crate) public_key: VerificationKey,
+    pub(crate) transaction: UnsignedTransaction,
```
SuperFluffy, line 40:
```suggestion
    pub(crate) transaction: transaction::Unsigned,
```

crates/astria-sequencer/src/transaction/signed.rs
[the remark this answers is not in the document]
SuperFluffy: Yes, either in `try_from_slice` (or `try_from_proto` ;) )

crates/astria-sequencer/src/transaction/signed.rs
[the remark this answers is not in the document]
SuperFluffy: It might be annoying - but at the moment that is extremely rarely down so we don't get a lot from the extra brevity so I don't think this would cause a lot of churn (also because this is a private interface).

On the other hand separating deserialization from verification helps in tests and in understanding how the types moves in and out of our program.

crates/astria-sequencer/src/transaction/unsigned.rs
```diff
@@ -0,0 +1,118 @@
...
+    hash,
+    transaction::signed::Transaction as SignedTransaction,
+};
+
+/// Represents an unsigned sequencer chain transaction.
+/// This type wraps all the different module-specific transactions.
+/// If a new transaction type is added, it should be added to this enum.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[non_exhaustive]
+pub(crate) enum Transaction {
```
SuperFluffy, line 23:
```suggestion
pub(crate) enum Unsigned {
```

crates/astria-sequencer/src/transaction/unsigned.rs
```diff
@@ -0,0 +1,118 @@
...
+};
+
+use crate::{
+    accounts::transaction::Transaction as AccountsTransaction,
+    crypto::SigningKey,
+    hash,
+    transaction::signed::Transaction as SignedTransaction,
+};
+
+/// Represents an unsigned sequencer chain transaction.
```
SuperFluffy, line 18: empty line after this, i.e.
```rust
/// Represents an ...
///
/// This type

File: crates/astria-sequencer/src/transaction/unsigned.rs [hunk 38]
```diff
@@ -0,0 +1,118 @@
...
+            Transaction::AccountsTransaction(tx) => ProtoUnsignedTransaction {
+                value: Some(ProtoAccountsTransaction(tx.to_proto())),
+            },
+        }
+    }
+
+    /// Signs the transaction with the given signing key.
+    #[allow(dead_code)]
+    #[must_use]
+    pub(crate) fn sign(self, secret_key: &SigningKey) -> SignedTransaction {
```
SuperFluffy, line 41: For transformations like convention is usually to call this `to_signed` or smth like that. But because this takes n argument maybe?
```suggestion
    pub(crate) fn to_signed_with_key(self, secret_key: &SigningKey) -> SignedTransaction {
```

crates/astria-sequencer/src/transaction/unsigned.rs
SuperFluffy: Well, this is crate-private and only used in one location (immediately above in `sign`). So it's not pretty, but at least I would see at a glance what's happening.

Also, f I were to see `self.hash()` I would be very surprised that it's not even the in-memory representation of my type that is being hashed, but the protobuf-serialized rust-protobuf in-memory representation of another type that my type is first converted to.

crates/astria-sequencer/src/transaction/unsigned.rs
```diff
@@ -0,0 +1,118 @@
...
+        /// - If the value is not a valid transaction type (ie. does not correspond to any
+        ///   component)
+        fn try_from_proto(proto: &ProtoUnsignedTransaction) -> Result<Self> {
+            let Some(ref value) = proto.value else {
+                bail!("invalid unsigned transaction; missing value");
+            };
+
+            Ok(match value {
+                ProtoAccountsTransaction(tx) => {
+                    Self::AccountsTransaction(AccountsTransaction::try_from_proto(tx)?)
```
SuperFluffy, line 95: please add error context

crates/astria-sequencer/src/transaction/unsigned.rs
noot: hmm, this gives me a clippy warning:
```
error: methods with the following characteristics: (`to_*` and `self` type is not `Copy`) usually take `self` by reference
  --> crates/astria-sequencer/src/transaction/unsigned.rs:42:29
   |
42 |     pub(crate) fn to_signed(self, secret_key: &SigningKey) -> Signed {
   |                             ^^^^
   |
   = help: consider choosing a less ambiguous name
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#wrong_self_convention
   = note: `-D clippy::wrong-self-convention` implied by `-D warnings`
```

noot: will change to `into_signed`

SuperFluffy (approved):
Thanks for the changes, this looks very nice now IMO.

Can you make an issue for implementing tests for the proto <> non-proto types?

noot:
@SuperFluffy opened #134

Issue #67, opened 2023-06-09
sequencer: implement parsing and updating of genesis file


noot:
cometbft creates a genesis.json file on `init`, however we want to be able to easily modify this file to add genesis balance allocations and validators. specifically we want to put the sequencer-specific type `GenesisState` into this genesis file.

requires:
- parsing the existing genesis file, or a way to easily add fields to the existing genesis file
- serializing/deserializing the app-specific `GenesisState` into this file
- (optional) CLI commands for adding balance allocations/validators to the genesis file

noot:
@sambukowski https://github.com/astriaorg/astria/blob/main/crates/astria-sequencer/test-genesis.json we want to put this into the cometbft genesis.json file

see instructions here for installing cometbft: https://github.com/astriaorg/astria/blob/main/crates/astria-sequencer/test-genesis.json you need to do the steps up to `cometbft init` which will put the genesis file in .cometbft/config/genesis.json

PR #74, opened 2023-06-13
Noot/sequencer fix

SuperFluffy:
Some general cleanup and especially instrumentation.


crates/astria-sequencer/src/accounts/query.rs
```diff
@@ -1,49 +1,37 @@
...
-};
+use crate::accounts::state_ext::StateReadExt;
 
 #[derive(Debug)]
-pub enum QueryRequest {
-    BalanceQuery(Address),
-    NonceQuery(Address),
+pub(crate) enum QueryRequest {
+    BalanceQuery(String),
+    NonceQuery(String),
```
noot, line 12: removal of the `Address` and `Balance` types are going to be annoying because I have to re-add them in my next PR for accounts that are key-backed, not just arbitrary strings. `Balance` also has to be a custom type because protobuf doesn't support u128 :/ can you add these types back? alternatively I'll have to re-add them in my next PR

crates/astria-sequencer/src/app.rs
```diff
@@ -74,13 +60,6 @@ impl App {
 
     #[instrument(name = "App:init_chain", skip(self))]
     pub async fn init_chain(&mut self, genesis_state: GenesisState) -> Result<()> {
-        // allocate to hard-coded accounts for testing
-        let mut accounts = genesis_state.accounts;
-        accounts.append(&mut default_genesis_accounts());
-        let genesis_state = GenesisState {
-            accounts,
-        };
-
```
noot, line 83: this *probably* shouldn't be here, except now the manual test instructions in the PR description/crate README will fail unless this is re-added since genesis parsing isn't implemented yet (#67) thus there are no funded accounts at genesis anymore.

noot: not sure what the exact right solution is here, maybe adding a feature flag for a test-genesis-state or something like that which adds this, until #67 is done?

crates/astria-sequencer/src/transaction.rs
```diff
@@ -78,24 +64,3 @@ impl ActionHandler for Transaction {
...
-            Address::from("alice"),
-            Balance::from(333333),
-            Nonce::from(1),
-        );
-        let bytes = tx.to_bytes().unwrap();
-        let tx2 = Transaction::from_bytes(&bytes).unwrap();
-        assert_eq!(tx, tx2);
-        println!("0x{}", hex::encode(bytes));
-    }
-}
```
noot, line 101: is this moved somewhere else?

crates/astria-tracing-tower/src/lib.rs
```diff
@@ -0,0 +1,24 @@
+/// A vendored copy of the unpublished [`tracing-tower`] crate:
+///
+/// Based on [`penumbra-tracing-tower`].
+/// [`tracing-tower`]: https://github.com/tokio-rs/tracing/tree/e603c2a254d157a25a7a1fbfd4da46ad7e05f555/tracing-towerpub mod trace;
```
noot, line 4:
```suggestion
/// [`tracing-tower`]: https://github.com/tokio-rs/tracing/tree/e603c2a254d157a25a7a1fbfd4da46ad7e05f555/tracing-tower
```

I think?

crates/astria-tracing-tower/src/lib.rs
```diff
@@ -0,0 +1,24 @@
+/// A vendored copy of the unpublished [`tracing-tower`] crate:
```
noot, line 1: are there differences between this and penumbra's crate?

crates/astria-sequencer/src/transaction.rs
SuperFluffy: This was basically testing if the serde deser worked so it's redundant in that sense.

crates/astria-sequencer/src/app.rs
SuperFluffy: Fair point. This is IMO core functionality. I will add it in.

crates/astria-tracing-tower/src/lib.rs
SuperFluffy: heh, it complained about `trace` not being available, so that's where it went. Either way, removed the dep in favour of just using penumbra.

crates/astria-tracing-tower/src/lib.rs
SuperFluffy: none, using upstream

crates/astria-sequencer/src/accounts/query.rs
SuperFluffy: brought them back. I also added a `Uint128` protobuf type in #77.

crates/astria-sequencer/src/app.rs
SuperFluffy: Added through a simple clap config, see `cargo run -- --help`

SuperFluffy:
I added context for all errors, added some instrumentation, and added CLI parsing to read the genesis state and change the listen addr.

crates/astria-sequencer/src/accounts/transaction.rs
```diff
@@ -49,24 +40,56 @@ impl ActionHandler for Transaction {
...
+        let curr_balance = state
+            .get_account_balance(&self.from)
+            .await
+            .context("failed getting `from` account balance")?;
         ensure!(curr_balance >= self.amount, "insufficient funds",);
 
         Ok(())
     }
 
+    // #[instrument(skip_all, fields(self.to.as_ref(), %self.from.as_ref(), %self.amount,
```
noot, line 60: remove?

crates/astria-sequencer/src/config.rs
```diff
@@ -0,0 +1,23 @@
...
+const DEFAULT_ABCI_LISTEN_ADDR: &str = "127.0.0.1:26658";
+
+#[derive(Debug, Deserialize, Parser, Serialize)]
+pub struct Config {
+    /// The endpoint on which Sequencer will listen for ABCI requests
+    #[arg(long, default_value_t = String::from(DEFAULT_ABCI_LISTEN_ADDR))]
+    pub(crate) listen_addr: String,
+    /// The path to the json encoded genesis file with a list of accounts.
+    #[arg(long)]
+    pub(crate) genesis_file: String,
```
noot, line 16: nice, eventually we need to merge this with the tendermint genesis file, as that's passed in via `InitChain`

SuperFluffy: yeah, I wasn't sure about that. I just blindly followed what was in the tests and figured it would be fine.

Note that I haven't updated the README for this. Can you give this a spin and do that?

noot: yep testing it now, will have to make a "test genesis" file for this

noot (approved):
works for me, merging into my branch, will add the test genesis file and update the README!

PR #77, opened 2023-06-13
feat(proto): add u128 pb type and conversion impls

SuperFluffy:


noot (approved):

crates/astria-proto/proto/astria/primitive/v1/types.proto
```diff
@@ -0,0 +1,8 @@
+syntax = "proto3";
+
+package astria.primitive.v1;
+
+message U128 {
```
joroshiba, line 5: Can we comment here on how to interpret (even if somewhat apparent), since any other languages would need to add custom support?

SuperFluffy: Hah, I forgot to push the comment. Sorry about that and I agree :-)

SuperFluffy:
Renamed `U128` to `Uint128`, bringing the naming closer to protobuf conventions.

joroshiba (approved):

PR #89, opened 2023-06-16
feat(sequencer-relayer): single entry point

### LINKED ISSUE #85: create a single entrypoint `SequencerRelayer` entrypoint

SuperFluffy:
This commit adds a `SequencerRelayer` type to
`astria-sequencer-relayer`, which serves as the
single entrypoint to the service. It creates all
constituent tasks from the service config, and
runs them until stopped.

This also lays the groundwork for writing
integration tests against `SequencerRelayer`
using the same codepaths as are used by the
binary.

In future commits all items that are currently
exposed will be made crate-private, except for
the `SequencerRelayer` type itself, config
generation, and telemetry initialization.

Because other workspace crates depend on items
defined within `astria-sequencer-relayer` they
are currently left unchanged.

Closes #85

### DIFF
--- a/crates/astria-sequencer-relayer/src/relayer.rs
+++ b/crates/astria-sequencer-relayer/src/relayer.rs
@@ -67,8 +67,8 @@ impl Relayer {
     /// Returns one of the following errors:
     /// + failed to read the validator keys from the path in cfg;
     /// + failed to construct a client to the sequencer;
-    /// + failed to construct a client to the data availability layer (if cfg.disable_writing is
-    /// set)
+    /// + failed to construct a client to the data availability layer (unless `cfg.disable_writing`
+    ///   is set).
     pub fn new(
         cfg: &crate::config::Config,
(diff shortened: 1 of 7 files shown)


crates/astria-sequencer-relayer/src/relayer.rs
```diff
@@ -53,39 +60,54 @@ impl State {
...
 
 impl Relayer {
+    /// Instantiates a `Relayer`.
+    ///
+    /// # Errors
+    ///
+    /// Returns one of the following errors:
+    /// + failed to read the validator keys from the path in cfg;
+    /// + failed to construct a client to the sequencer;
+    /// + failed to construct a client to the data availability layer (if cfg.disable_writing is
```
noot, line 70:
```suggestion
    /// + failed to construct a client to the data availability layer (if cfg.disable_writing is not set)
```

joroshiba (approved):

noot (approved):

PR #91, opened 2023-06-16
ci: make pedantic clippy the default

SuperFluffy:
This commit makes `clippy::pedantic` the default
lint group for the repository. Only a few crates
are currently exempt from it. All new crates are
expected to adhere to `clippy::pedantic`.

This commit also ensures that `astria-proto`
does not trigger `clippy::pedantic` warnings:

+ generated code is generally exempt
+ a few doc comment related lints are enabled to make for nicer rust docs from the generated code
+ the conversion methods for `Uint128` are updated to no longer need explicit casts; they expicitly move bytes into the right positions instead.

### DIFF
--- a/crates/astria-proto/src/lib.rs
+++ b/crates/astria-proto/src/lib.rs
@@ -93,8 +93,8 @@ mod primitive_impls {
         fn u128_roundtrips_work() {
             u128_roundtrip_check(0u128);
             u128_roundtrip_check(1u128);
-            u128_roundtrip_check(u64::MAX as u128);
-            u128_roundtrip_check(u64::MAX as u128 + 1u128);
+            u128_roundtrip_check(u128::from(u64::MAX));
+            u128_roundtrip_check(u128::from(u64::MAX) + 1u128);
             u128_roundtrip_check(1u128 << 127);
             u128_roundtrip_check((1u128 << 127) + (1u128 << 63));
             u128_roundtrip_check(u128::MAX);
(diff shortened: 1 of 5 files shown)


crates/astria-proto/src/lib.rs
```diff
@@ -24,8 +43,26 @@ mod primitive_impls {
...
+                l1,
+                l2,
+                l3,
+                l4,
+                l5,
+                l6,
+                l7,
+            ] = primitive.to_be_bytes();
+            let lo = u64::from_be_bytes([l0, l1, l2, l3, l4, l5, l6, l7]);
+            let hi = u64::from_be_bytes([h0, h1, h2, h3, h4, h5, h6, h7]);
```
noot, line 65: omg, this is an interesting diff, why does clippy prefer this method?

noot (approved):

crates/astria-proto/src/lib.rs
SuperFluffy: So clippy really hates `as` casts, which I kinda agree with. It's better to rely on `From` and `TryFrom` to make sure one does not get accidental precision loss, wrap arounds, etc (especially when refactoring at some point).

However, in this case it flagged a lossy conversion where I performed `u128 as u64`, which basically cuts off the upper 64 bits. That's technically correct and wanted - but I had to make a choice on whether to add an `#[allow(clippy::cast_possible_truncation)]` or to just spell it out explicitly, shuffling the bytes around as needed.

As you can see I went with the verbose but (in my mind) absolutely and totally cannot-be-confused solution. :laughing:

joroshiba (approved):

PR #103, opened 2023-06-16
fix(relayer): fix calculation of data_hash

### LINKED ISSUE #105: fix calculation of tendermint block `data_hash`
this error was seen when running the devnet setup.

```
 2023-06-16T21:50:03.515068Z  WARN astria_conductor::reader: sequencer block failed validation: data hash stored in block header does not match hash calculated from transactions
    at crates/astria-conductor/src/reader.rs:211
```

noot:
fixes data hash mismatch:
```
  2023-06-16T21:50:03.515068Z  WARN astria_conductor::reader: sequencer block failed validation: data hash stored in block header does not match hash calculated from transactions
    at crates/astria-conductor/src/reader.rs:211
```

transactions needed to be hashed first before being merkleized.

closes #105

### DIFF
--- a/crates/astria-sequencer-relayer/src/sequencer_block.rs
+++ b/crates/astria-sequencer-relayer/src/sequencer_block.rs
@@ -233,20 +233,11 @@ impl SequencerBlock {
             bail!("block has no data hash");
         };
 
-        tracing::info!(
-            "block transactions: {} {}",
-            self.sequencer_txs.len(),
-            self.rollup_txs.len(),
-        );
-
         let mut ordered_txs = vec![];
         ordered_txs.append(&mut self.sequencer_txs.clone());
-        self.rollup_txs.iter().for_each(|(_, tx)| {
-            tracing::info!("rollup txs: {:?}", tx);
-            ordered_txs.append(&mut tx.clone())
-        });
-
-        use sha2::Digest as _;
+        self.rollup_txs
+            .iter()
+            .for_each(|(_, tx)| ordered_txs.append(&mut tx.clone()));
 
         // TODO: if there are duplicate or missing indices, the hash will obviously be wrong,
         // but we should probably verify that earier to return a better error.
@@ -260,12 +251,6 @@ impl SequencerBlock {
             })
             .collect::<Vec<_>>();
         let data_hash = txs_to_data_hash(&txs);
-
-        tracing::info!(
-            "data hash from transactions: {}",
-            Base64String(data_hash.as_bytes().to_vec())
-        );
-        tracing::info!("data hash from block: {}", this_data_hash,);
 
         ensure!(
             data_hash.as_bytes() == this_data_hash.0,
--- a/crates/astria-sequencer-relayer/src/sequencer_block.rs
+++ b/crates/astria-sequencer-relayer/src/sequencer_block.rs
@@ -233,20 +233,39 @@ impl SequencerBlock {
             bail!("block has no data hash");
         };
 
+        tracing::info!(
+            "block transactions: {} {}",
+            self.sequencer_txs.len(),
+            self.rollup_txs.len(),
+        );
+
         let mut ordered_txs = vec![];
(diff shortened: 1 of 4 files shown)


crates/astria-sequencer-relayer/src/sequencer_block.rs
```diff
@@ -244,7 +244,11 @@ impl SequencerBlock {
         ordered_txs.sort_by(|a, b| a.block_index.cmp(&b.block_index));
         let txs = ordered_txs
             .into_iter()
-            .map(|tx| tx.transaction)
+            .map(|tx| {
+                let mut hasher = sha2::Sha256::new();
+                hasher.update(tx.transaction.0);
```
SuperFluffy, line 249: Put this block into its own function?

crates/astria-sequencer-relayer/src/transaction.rs
```diff
@@ -1,24 +1,37 @@
...
+
+    use super::*;
+
+    #[test]
+    fn txs_to_data_hash_test() {
+        // data_hash is calculated from the txs in a block, where the leaves of the merkle tree are
+        // the sha256 hashes of the txs
+        let tx = Base64String::from_string("CscBCsQBCg0vU2VxdWVuY2VyTXNnErIBCghldGhlcmV1bRJ4Avh1ggU5gIRZaC8AhQUD1cTyglIIlBtwp0/22gQLMRmQwVX9/9u8AvfuiA3gtrOnZAAAgMABoLnRqksJblEaolE6wbsAHYTAiSlA14+B5nvWuFrIfevnoBg+UGcWLC4eg1lZylqLnrL8okBc3vTS4qOO/J5sRtVDGixtZXRybzFsbDJobHAzM3J4eTdwN2s2YXhoeDRjdnFtdGcwY3hkZjZnemY5ahJ0Ck4KRgofL2Nvc21vcy5jcnlwdG8uc2VjcDI1NmsxLlB1YktleRIjCiEDJ/LvaMZTBcGX66geJOEmTm/fyyPTZKMUJoDtMDUmSPkSBAoCCAESGAoQCgV1dGljaxIHMTAwMDAwMBCAlOvcAyIIZXRoZXJldW0aQMhoTCUr84xgTkYxsFWDfHH2k+oHCPsKvbTpz8m5YrHfYMv6gdou6V8oj1v0B9ySD5VjMXQi1kJ9DZN6wD2buo8=".to_string()).unwrap();
+        let mut hasher = sha2::Sha256::new();
+        hasher.update(tx.0);
```
SuperFluffy, line 28: And then use that function from the above comment here instead of inlining

crates/astria-sequencer-relayer/src/transaction.rs
```diff
@@ -1,24 +1,37 @@
...
+        let hash = hasher.finalize();
+
+        let expected_hash =
+            Base64String::from_string("rRDu3aQf1V37yGSTdf2fv9GSPeZ6/p0wJ9pjBl8IqFc=".to_string())
```
SuperFluffy, line 32: Not a fan of these magic number tests. What does it tell us should they ever fail? It's very inscrutable.

I don't know if this can be easily solved in a unit test. Maybe spin up an instance of the thing that's supposed to understand this message?

SuperFluffy (approved):
Changes look fine except for moving the hashing into a free function.

Not a fan of the test. Dunno what would be better

joroshiba (approved):

joroshiba:
Can we attach an issue?

crates/astria-sequencer-relayer/src/transaction.rs
noot: yeah, the `header_verify_hashes` should be testing this, except the problem is we also need to submit a transaction to validate a data_hash that isn't just the empty merkle tree hash, but there's no tx signing library in rust for metro :| when we move to the rust sequencer, this will definitely be doable!

Issue #98, opened 2023-06-16
use tendermint-rs types for block commits


itamarreif:
Currently we are redefining CometBFT's `Commit`, `CommitSig`, `BlockId` and `PartSetHeader` types in our [`astria-proto/proto/sequencer/v1/data.proto`](https://github.com/astriaorg/astria/blob/main/crates/astria-proto/proto/astria/sequencer/v1/data.proto) (for wire types) and `sequencer-relayer/src/types.rs` (for domain types) as well as manually converting them. See the `to_proto()` and `from_proto()` funcs across `types.rs`, linked above.

This is done because `CometBFT` and `tendermint-rs`'s protos definitions for these types do not properly serialize Base64 strings, which are used for hashes and signatures. For example, types that do properly serialize to Base64 have the following proc macro in their definition: https://github.com/informalsystems/tendermint-rs/blob/main/rpc/src/endpoint/evidence.rs#L42

tasks:
- [ ] change the definitions in `astria-proto/proto/sequencer/v1/data.proto` to use the types imported from tendermint's protos and remove the copied definitions.
- [ ] remove the domain types in `sequencer-relayer/src/types.rs` and replace them with types imported from tendermint-rs.
- [ ] fix the inevitable problems that will result across `conductor`

noot:
to add more context on why this is blocked:
tendermint-rs's `Hash` type, which is contained is basically everything to do with blocks, commits, etc. supports deserialization from hex strings only, not base64 strings. however, since we are getting these types with hashes encoded as base64, we are unable to deserialize them into a `Hash`

for example, this is the result we get from metro:
```
$ curl http://localhost:1317/cosmos/base/tendermint/v1beta1/blocks/latest
{
  "block_id": {
    "hash": "JEEXrSESfa01mbd3nH6q4ZSGPKn8NHtlL5wivJB/iYA=",
    "part_set_header": {
      "total": 1,
      "hash": "MerMvPanzEuYLiBfsd1sQ7wvwVBbbxALOMzFHi8EKgw="
    }
  },
// extra stuff omitted
```

see the tendermint-rs type here: https://github.com/informalsystems/tendermint-rs/blob/main/proto/src/prost/v0_37/tendermint.types.rs#L63

it only is able to deserialize the `Hash` via hex.

note that this appears to be due to the cosmos-sdk specifically, as tendermint itself returns hashes encoded in hex: https://docs.tendermint.com/v0.37/rpc/#/Info/block

this means using "retro" will fix this issue.

Issue #50, opened 2023-06-07
newly proposed blocks should be marked "head", and commited should be marked "safe"


noot:
When tendermint proposes a new block, the block contains the commit for the parent block. However, the conductor currently treats this incorrectly, as when it receives a newly-proposed block, it executes it and marks it as safe/softly-committed, even though it's not actually finalized by tendermint yet. The correct logic when receiving a newly proposed block would be:
- execute and mark newly proposed block as "head" of chain
- verify commit contained in block (which is for the parent block), if it's verified, then mark parent block as "safe"
- receiving block from DA remains the same

Requires:
- updating the gRPC service to have methods such as:
```
ExecuteAndMarkAsHead(txs, prev_block_hash)
SetBlockAsSafe(block_hash)
SetBlockAsFinal(block_hash)
```

- updating the geth fork to support these methods
- updating conductor to correctly handle newly proposed vs committed blocks

blocked by #44

noot:
yeah i think this is probably a duplicate of #207 at this point, will close

> there is no need to validate commitment to the rollup data 

we still need to do this when reading from DA

PR #106, opened 2023-06-16
feat(conductor): implement BlockValidator type and validate blocks received from gossip

### LINKED ISSUE #104: conductor: validate block received from gossip
same checks need to be done on a gossip block vs a DA block (except we don't need to validate the blob signature, as there is no blob).

noot:
- pull out the block validator logic into its own `BlockValidator` module (thoughts on naming? maybe `BlockVerifier` is better since validator is already a jargon word?)
- validate blocks received from gossip, as previously only blocks received from DA were fully validated

closes #104

### DIFF
--- a/crates/astria-conductor/src/driver.rs
+++ b/crates/astria-conductor/src/driver.rs
@@ -1,11 +1,15 @@
 //! The driver is the top-level coordinator that runs and manages all the components
 //! necessary for this reader/validator.
 
-use std::sync::Mutex;
+use std::sync::{
+    Arc,
+    Mutex,
+};
 
 use astria_sequencer_relayer::sequencer_block::SequencerBlock;
 use color_eyre::eyre::{
     eyre,
+    Context,
     Result,
 };
 use futures::StreamExt;
@@ -35,6 +39,7 @@ use crate::{
         self,
         ReaderCommand,
     },
+    validation::BlockValidator,
 };
 
 /// The channel through which the user can send commands to the driver.
@@ -65,6 +70,8 @@ pub struct Driver {
 
     network: GossipNetwork,
 
+    block_validator: Arc<BlockValidator>,
+
     is_shutdown: Mutex<bool>,
 }
 
@@ -76,10 +83,13 @@ impl Driver {
         let (cmd_tx, cmd_rx) = mpsc::unbounded_channel();
         let (executor_join_handle, executor_tx) = executor::spawn(&conf, alert_tx.clone()).await?;
 
+        let block_validator = Arc::new(BlockValidator::new(&conf.tendermint_url)?);
+
         let (reader_join_handle, reader_tx) = if conf.disable_finalization {
             (None, None)
         } else {
-            let (reader_join_handle, reader_tx) = reader::spawn(&conf, executor_tx.clone()).await?;
+            let (reader_join_handle, reader_tx) =
+                reader::spawn(&conf, executor_tx.clone(), block_validator.clone()).await?;
             (Some(reader_join_handle), Some(reader_tx))
         };
 
@@ -90,6 +100,7 @@ impl Driver {
                 reader_tx,
                 executor_tx,
                 network: GossipNetwork::new(conf.bootnodes)?,
+                block_validator,
                 is_shutdown: Mutex::new(false),
             },
             executor_join_handle,
@@ -128,7 +139,15 @@ impl Driver {
             NetworkEvent::Message(msg) => {
                 debug!("received gossip message: {:?}", msg);
                 let block = SequencerBlock::from_bytes(&msg.data)?;
-                // TODO: validate this block!!!!!
+
+                // validate block received from gossip network
+                // note: `validate_sequencer_block` is async, but `handle_network_event` cannot be
+                // async due to libp2p restrictions.
+                let handle = tokio::runtime::Handle::current();
+                handle
+                    .block_on(self.block_validator.validate_sequencer_block(&block))
+                    .wrap_err("invalid block received from gossip network")?;
+
                 self.executor_tx
                     .send(ExecutorCommand::BlockReceivedFromGossipNetwork {
                         block: Box::new(block),
--- a/crates/astria-conductor/src/driver.rs
+++ b/crates/astria-conductor/src/driver.rs
@@ -128,6 +128,7 @@ impl Driver {
             NetworkEvent::Message(msg) => {
                 debug!("received gossip message: {:?}", msg);
                 let block = SequencerBlock::from_bytes(&msg.data)?;
+                // TODO: validate this block!!!!!
                 self.executor_tx
                     .send(ExecutorCommand::BlockReceivedFromGossipNetwork {
                         block: Box::new(block),
--- a/crates/astria-conductor/src/reader.rs
+++ b/crates/astria-conductor/src/reader.rs
@@ -1,7 +1,4 @@
-use std::{
-    collections::HashMap,
-    str::FromStr,
-};
+use std::sync::Arc;
 
 use astria_sequencer_relayer::{
     data_availability::{
@@ -10,48 +7,13 @@ use astria_sequencer_relayer::{
         SequencerNamespaceData,
         SignedNamespaceData,
     },
-    keys::public_key_to_address,
     sequencer_block::SequencerBlock,
-    types::{
-        Commit,
-        CommitSig,
-    },
-};
-use bech32::{
-    self,
-    ToBase32,
-    Variant,
 };
 use color_eyre::eyre::{
-    bail,
-    ensure,
     eyre,
     Result,
     WrapErr,
 };
-use ed25519_dalek::Verifier;
-use prost::Message;
-use tendermint::{
-    account::Id as AccountId,
-    block::{
-        parts,
-        CommitSig as TendermintCommitSig,
-        Height,
-        Id as BlockId,
-        Round,
-    },
-    chain::Id as ChainId,
-    crypto,
-    merkle,
-    vote::{
-        self,
-        CanonicalVote,
-    },
-    Hash,
-    Signature,
-    Time,
-};
-use tendermint_proto::types::CommitSig as RawCommitSig;
 use tokio::{
     sync::mpsc::{
         self,
@@ -64,17 +26,13 @@ use tracing::{
     debug,
     error,
     info,
-    instrument,
     warn,
 };
 
 use crate::{
     config::Config,
     executor,
-    tendermint::{
-        TendermintClient,
-        ValidatorSet,
-    },
+    validation::BlockValidator,
 };
 
 pub(crate) type JoinHandle = task::JoinHandle<Result<()>>;
@@ -89,10 +47,11 @@ type Receiver = UnboundedReceiver<ReaderCommand>;
 pub(crate) async fn spawn(
     conf: &Config,
     executor_tx: executor::Sender,
+    block_validator: Arc<BlockValidator>,
 ) -> Result<(JoinHandle, Sender)> {
     info!("Spawning reader task.");
     let (mut reader, reader_tx) =
-        Reader::new(&conf.celestia_node_url, &conf.tendermint_url, executor_tx).await?;
+        Reader::new(&conf.celestia_node_url, executor_tx, block_validator).await?;
     let join_handle = task::spawn(async move { reader.run().await });
     info!("Spawned reader task.");
     Ok((join_handle, reader_tx))
@@ -119,15 +78,15 @@ pub struct Reader {
     /// the last block height fetched from Celestia
     curr_block_height: u64,
 
-    tendermint_client: TendermintClient,
+    block_validator: Arc<BlockValidator>,
 }
 
 impl Reader {
     /// Creates a new Reader instance and returns a command sender and an alert receiver.
     pub async fn new(
         celestia_node_url: &str,
-        tendermint_url: &str,
         executor_tx: executor::Sender,
+        block_validator: Arc<BlockValidator>,
     ) -> Result<(Self, Sender)> {
         let (cmd_tx, cmd_rx) = mpsc::unbounded_channel();
         let celestia_client = CelestiaClientBuilder::new(celestia_node_url.to_owned())
@@ -143,7 +102,7 @@ impl Reader {
                 executor_tx,
                 celestia_client,
                 curr_block_height,
-                tendermint_client: TendermintClient::new(tendermint_url.to_owned())?,
+                block_validator,
             },
             cmd_tx,
         ))
@@ -203,15 +162,29 @@ impl Reader {
                     };
 
                     for data in datas {
-                        let block = match self.validate_sequencer_namespace_data(data).await {
+                        // validate data
+                        self.block_validator
+                            .validate_signed_namespace_data(&data)
+                            .await?;
+
+                        let block = match self.get_sequencer_block_from_namespace_data(&data).await
+                        {
                             Ok(block) => block,
                             Err(e) => {
                                 // this means someone submitted an invalid block to celestia;
                                 // we can ignore it
-                                warn!("sequencer block failed validation: {}", e);
+                                warn!("failed to get sequencer block from namespace data: {}", e);
                                 continue;
                             }
                         };
+
+                        if let Err(e) = self.block_validator.validate_sequencer_block(&block).await
+                        {
+                            // this means someone submitted an invalid block to celestia;
+                            // we can ignore it
+                            warn!("sequencer block failed validation: {}", e);
+                            continue;
+                        }
 
                         blocks.push(block);
                     }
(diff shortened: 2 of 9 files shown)


crates/astria-conductor/src/driver.rs
```diff
@@ -1,11 +1,15 @@
...
-use std::sync::Mutex;
+use std::sync::{
+    Arc,
+    Mutex,
+};
 
 use astria_sequencer_relayer::sequencer_block::SequencerBlock;
 use color_eyre::eyre::{
     eyre,
+    Context,
```
joroshiba, line 12: How is this used?

joroshiba (approved):
LTGM overall. On naming, I do think `BlockVerifier` is better do to overload of validator

crates/astria-conductor/src/driver.rs
noot: it's a trait, imported so that we can use `.context()` or `.wrap_err()` on our errors, for example:
```
                handle
                    .block_on(self.block_verifier.validate_sequencer_block(&block))
                    .wrap_err("invalid block received from gossip network")?;
```

noot:
@joroshiba agreed, i've renamed the type to `BlockVerifier`

crates/astria-conductor/src/driver.rs
SuperFluffy: For consistency, please import `WrapErr as _` (`Context` is for quickly migrate codebases from anyhow).

noot: done

crates/astria-conductor/src/driver.rs
```diff
@@ -128,7 +139,15 @@ impl Driver {
...
                 debug!("received gossip message: {:?}", msg);
                 let block = SequencerBlock::from_bytes(&msg.data)?;
-                // TODO: validate this block!!!!!
+
+                // validate block received from gossip network
+                // note: `validate_sequencer_block` is async, but `handle_network_event` cannot be
+                // async due to libp2p restrictions.
+                let handle = tokio::runtime::Handle::current();
+                handle
+                    .block_on(self.block_verifier.validate_sequencer_block(&block))
```
SuperFluffy, line 148: This makes me very uneasy. Can you link some documentation that explains this a bit further (namely the bit about `handle_network_event` not being async)?


If this is the case then `Driver::run` or the part that is interacting with the p2p network should be put into a `tokio::task::spawn_blocking`

crates/astria-conductor/src/reader.rs
```diff
@@ -89,10 +47,11 @@ type Receiver = UnboundedReceiver<ReaderCommand>;
 pub(crate) async fn spawn(
     conf: &Config,
     executor_tx: executor::Sender,
+    block_verifier: Arc<BlockVerifier>,
 ) -> Result<(JoinHandle, Sender)> {
     info!("Spawning reader task.");
     let (mut reader, reader_tx) =
-        Reader::new(&conf.celestia_node_url, &conf.tendermint_url, executor_tx).await?;
+        Reader::new(&conf.celestia_node_url, executor_tx, block_verifier).await?;
```
SuperFluffy, line 54: Please add error context when you touch lines that lack it, like here.

crates/astria-conductor/src/driver.rs
```diff
@@ -65,6 +70,8 @@ pub struct Driver {
 
     network: GossipNetwork,
 
+    block_verifier: Arc<BlockVerifier>,
```
SuperFluffy, line 73: Just for clarification, this is behind an `Arc` because it's shared between the reader and the driver?

crates/astria-conductor/src/validation.rs
```diff
@@ -0,0 +1,537 @@
...
+    instrument,
+    warn,
+};
+
+use crate::tendermint::{
+    TendermintClient,
+    ValidatorSet,
+};
+
+pub struct BlockVerifier {
```
SuperFluffy, line 62: Two notes:

1. please make this crate-private, e.g. `pub(crate)`, as well as it's currently `pub` methods.
2. please add a do comment about the purpose of this type.

crates/astria-conductor/src/validation.rs
SuperFluffy, line 1: Maybe name this module e.g. `block_verifier.rs`?

SuperFluffy (changes requested):
The business logic looks good to me.

Asking for changes on:
1. error context
2. public visibility

I also hate the `block_on` when handling network events. conductor has several cases like this where async tasks are `await`ed inside loops, effectively making the entire processing sequential. Blocking the async executor is the same thing, but worse.

Why not spin up a block task, move gossipnet + the `executor_tx` + the `block_verifier` into it and separate it completely from `Driver`? (driver could still hold the task handle)

crates/astria-conductor/src/driver.rs
noot: yeah, so to handle this async I need to make `Driver::run()` async, which then causes this error inside where `Driver::run()` is spawned here:
```
    tokio::task::spawn(async move {
        if let Err(e) = driver.run().await {
            panic!("Driver error: {}", e)
        }
    });
```

```
future cannot be sent between threads safely
the trait `Sync` is not implemented for `(dyn libp2p_core::transport::boxed::Abstract<(libp2p_identity::peer_id::PeerId, libp2p_core::muxing::boxed::StreamMuxerBox)> + Send + Unpin + 'static)`rustc[Click for full compiler diagnostic](rust-analyzer-diagnostics-view:/diagnostic%20message%20%5B0%5D?0#file%3A%2F%2F%2Fhome%2Fe%2Fastria%2Fcrates%2Fastria-conductor%2Fsrc%2Fbin%2Fastria-conductor.rs)
driver.rs(120, 55): future is not `Send` as this value is used across an await
spawn.rs(163, 21): required by a bound in `tokio::spawn`
```

this makes me think the driver shouldn't contain an instance of `GossipNetwork` and they should communicate over a channel or something, what do you think?

crates/astria-conductor/src/driver.rs
noot: yes exactly!

crates/astria-conductor/src/validation.rs
noot: I can't make it crate-private because tests/reader.rs needs it to be public :/ will add comment

SuperFluffy: Gah, ok. This is part of my future refactor then

crates/astria-conductor/src/driver.rs
SuperFluffy: Yeah, I agree. Let's put the loop over the gossip stream into its own dedicated task, since this need not all be done in the same `select!` and it can just be its own thing. So it can get its clone of `executor_tx` etc

Inside the loop over the gossip stream we could make use of a tokio `LocalSet` to handle the events asynchronously. But this could be left to another PR.

I'll investigate what's with the trait bounds when I'm back from vacation

noot: sounds good, I opened an issue: #111

noot: fixed with sync_wrapper :DD

noot:
@SuperFluffy I added `wrap_err`s to all errors in the diff that didn't have them! should be ready for review again

crates/astria-conductor/src/block_verifier.rs
```diff
@@ -0,0 +1,583 @@
...
+};
+use bech32::{
+    self,
+    ToBase32,
+    Variant,
+};
+use color_eyre::eyre::{
+    bail,
+    ensure,
+    Result,
```
dereksione, line 26: Nit: Personally, I would recommend aliasing this `Result` into something like `EryeResult` so that it's clear that the Error is being handled by Eyre

crates/astria-conductor/src/block_verifier.rs
```diff
@@ -0,0 +1,583 @@
...
+};
+
+/// `BlockVerifier` is responsible for verifying the correctness of a block
+/// before executing it.
+/// It has two public functions: `validate_signed_namespace_data` and `validate_sequencer_block`.
+///
+/// `validate_signed_namespace_data` is used to validate the data received from the data
+/// availability layer. `validate_sequencer_block` is used to validate the blocks received from
+/// either the data availability layer or the gossip network.
+pub struct BlockVerifier {
```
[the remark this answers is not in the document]
dereksione: The other option would be to import `eyre` and use `eyre::Result` in the code to disambiguate the different `Result` types

crates/astria-conductor/src/block_verifier.rs
```diff
@@ -0,0 +1,583 @@
...
+    /// - the root of the markle tree of all the header fields matches the block's block_hash
+    /// - the root of the merkle tree of all transactions in the block matches the block's data_hash
+    /// - validate the block was actually finalized; ie >2/3 stake signed off on it
+    pub(crate) async fn validate_sequencer_block(&self, block: &SequencerBlock) -> Result<()> {
+        // sequencer block's height
+        let height = block
+            .header
+            .height
+            .parse::<u64>()
+            .wrap_err("failed to parse height")?;
```
dereksione, line 142:
```suggestion
        let height: u64 = block
            .header
            .height
            .parse()
            .wrap_err("failed to parse height")?;
```

crates/astria-conductor/src/block_verifier.rs
```diff
@@ -0,0 +1,583 @@
...
+
+        // check if the proposer address matches the sequencer block's proposer
+        let received_proposer_address = bech32::encode(
+            "metrovalcons",
+            block.header.proposer_address.0.to_base32(),
+            Variant::Bech32,
+        )
+        .wrap_err("failed converting bytes to bech32 address")?;
+
+        if received_proposer_address != expected_proposer_address {
```
dereksione, line 165: Nit: Would prefer guard clauses use some version of `assert`

crates/astria-conductor/src/block_verifier.rs
```diff
@@ -0,0 +1,583 @@
...
+
+        // find proposer address for this height
+        let expected_proposer_address = validator_set
+            .get_proposer()
+            .wrap_err("failed to get proposer from validator set")?
+            .address;
+
+        // check if the proposer address matches the sequencer block's proposer
+        let received_proposer_address = bech32::encode(
+            "metrovalcons",
```
dereksione, line 159: This feels like a magic string. Should be a constant.

crates/astria-conductor/src/block_verifier.rs
```diff
@@ -0,0 +1,583 @@
...
+                // verify that the validator votes on the previous block have >2/3 voting power
+                let last_commit = block.last_commit.clone();
+                let chain_id = block.header.chain_id.clone();
+                tokio::task::spawn_blocking(move || -> Result<()> {
+                    ensure_commit_has_quorum(&last_commit, &validator_set, &chain_id)
+                })
+                .await?
+                .wrap_err("failed to ensure commit has quorum")?
+
+                // TODO: commit is for previous block; how do we handle this?
```
dereksione, line 193: Is there an issue for this TODO? Would prefer TODOs be Github issues that are referenced in code.

crates/astria-conductor/src/block_verifier.rs
```diff
@@ -0,0 +1,583 @@
...
+                };
+
+                if calculated_last_commit_hash.as_slice() != last_commit_hash.0 {
+                    bail!("last commit hash in header does not match calculated last commit hash");
+                }
+
+                // verify that the validator votes on the previous block have >2/3 voting power
+                let last_commit = block.last_commit.clone();
+                let chain_id = block.header.chain_id.clone();
+                tokio::task::spawn_blocking(move || -> Result<()> {
```
dereksione, line 187: I'm not sure what is happening here. Why are spawning a blocking task and then awaiting it immediately?

crates/astria-conductor/src/block_verifier.rs
```diff
@@ -0,0 +1,583 @@
...
+    for vote in &commit.signatures {
+        // we only care about votes that are for the Commit.BlockId (ignore absent validators and
+        // votes for nil)
+        if vote.block_id_flag != "BLOCK_ID_FLAG_COMMIT" {
+            continue;
+        }
+
+        // verify validator exists in validator set
+        let validator_address = bech32::encode(
+            "metrovalcons",
```
dereksione, line 269: Magic string

crates/astria-conductor/src/block_verifier.rs
dereksione: You use `ensure` later, so lets' stick to that

crates/astria-conductor/src/block_verifier.rs
```diff
@@ -0,0 +1,583 @@
...
+fn does_commit_voting_power_have_quorum(commited: u64, total: u64) -> bool {
+    if total < 3 {
+        return commited * 3 > total * 2;
+    }
+
+    commited > total / 3 * 2
+}
+
+// see https://github.com/tendermint/tendermint/blob/35581cf54ec436b8c37fabb43fdaa3f48339a170/types/vote.go#L147
+// TODO: we can change these types (CommitSig and Commit) to be the tendermint types
```
dereksione, line 331: Would love if this was an issue and the issue link was referenced here

crates/astria-conductor/src/block_verifier.rs
```diff
@@ -0,0 +1,583 @@
...
+// after the other relayer types are updated.
+fn verify_vote_signature(
```
dereksione, line 333: Nit: Why does this function have no whitespace between statements? I would prefer if our lints were consistent

crates/astria-conductor/src/block_verifier.rs
```diff
@@ -0,0 +1,583 @@
...
+    commit: &Commit,
+    chain_id: &str,
+    public_key_bytes: &[u8],
+    signature_bytes: &[u8],
+) -> Result<()> {
+    let public_key = ed25519_dalek::PublicKey::from_bytes(public_key_bytes)
+        .wrap_err("failed to create public key from vote")?;
+    let signature = ed25519_dalek::Signature::from_bytes(signature_bytes)
+        .wrap_err("failed to create signature from vote")?;
+    let canonical_vote = CanonicalVote {
```
dereksione, line 344: Nit: Would prefer it if we use the builder design pattern to declare structs

dereksione (changes requested):

crates/astria-conductor/src/block_verifier.rs
dereksione: The better option is to actually use `eyre::Result` in the returntypes to disambiguate and not deal with aliasing

crates/astria-conductor/src/block_verifier.rs
noot: yep #50, will add

crates/astria-conductor/src/block_verifier.rs
noot: @SuperFluffy ?

crates/astria-conductor/src/block_verifier.rs
noot: #98 Itamar had a lot of issues with this - basically the tendermint types can't actually be deserialized into from the cosmos RPC return strings

crates/astria-conductor/src/block_verifier.rs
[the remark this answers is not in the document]
noot: arbitrary, I can change it

crates/astria-conductor/src/block_verifier.rs
noot: example of builder pattern for structs? note that this is a type from tendermint-rs, not our type

dereksione (approved):

crates/astria-conductor/src/driver.rs
```diff
@@ -63,7 +69,9 @@ pub struct Driver {
     /// The channel used to send messages to the executor task.
     executor_tx: executor::Sender,
 
-    network: GossipNetwork,
+    network: SyncWrapper<GossipNetwork>,
```
SuperFluffy, line 72: Can you add comments on a) why this is necessary, and b) link to the rust internals discussion and the `Exclusive`type that is still unstable?

crates/astria-conductor/src/driver.rs
```diff
@@ -74,12 +82,22 @@ impl Driver {
         alert_tx: AlertSender,
     ) -> Result<(Self, executor::JoinHandle, Option<reader::JoinHandle>)> {
         let (cmd_tx, cmd_rx) = mpsc::unbounded_channel();
-        let (executor_join_handle, executor_tx) = executor::spawn(&conf, alert_tx.clone()).await?;
+        let (executor_join_handle, executor_tx) = executor::spawn(&conf, alert_tx.clone())
+            .await
+            .wrap_err("failed to construct block verifier")?;
```
SuperFluffy, line 87: Is this message correct? It fails to spawn the executor, right?

crates/astria-conductor/src/driver.rs
```diff
@@ -74,12 +82,22 @@ impl Driver {
...
+        );
 
         let (reader_join_handle, reader_tx) = if conf.disable_finalization {
             (None, None)
         } else {
-            let (reader_join_handle, reader_tx) = reader::spawn(&conf, executor_tx.clone()).await?;
+            let (reader_join_handle, reader_tx) =
+                reader::spawn(&conf, executor_tx.clone(), block_verifier.clone())
+                    .await
+                    .wrap_err("failed to construct Reader")?;
```
SuperFluffy, line 100: I know we refer to this as `reader`, but is there a better way to call this than `Reader` at least in the error? What is it reading?

crates/astria-conductor/src/driver.rs
```diff
@@ -102,14 +124,14 @@ impl Driver {
         info!("Starting driver event loop.");
         loop {
             select! {
-                res = self.network.0.next() => {
+                res = self.network.get_mut().0.next() => {
                     if let Some(res) = res {
-                        self.handle_network_event(res)?;
+                        self.handle_network_event(res).await.wrap_err("failed to handle network event")?;
```
SuperFluffy, line 129: In the case of an error, should the entire event loop abort? Or should this just log the error and continue the loop?

SuperFluffy (approved):
The use of `sync_wrapper` makes this much nicer. Thank you for adding it.

Approving the PR as I don't see any blockers, but please have a look at the comments and make a not on `SyncWrapper` and link the relevant discussions on Rust internal and the tracking issue of `Exclusive`.

crates/astria-conductor/src/block_verifier.rs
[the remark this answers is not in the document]
SuperFluffy: Honestly, I sometimes reach for newtype patterns, but i find tuple accesses hard to read.

IMO `block_verifier.tendermin_client` is much more obvious than `block_verifier.0`

crates/astria-conductor/src/block_verifier.rs
```diff
@@ -0,0 +1,616 @@
...
+            .get_proposer()
+            .wrap_err("failed to get proposer from validator set")?
+            .address;
+
+        // verify the namespace data signing public key matches the proposer address
+        let res_address = public_key_to_bech32_address(&data.public_key.0)
+            .wrap_err("failed to convert namespace data public key to address")?;
+        ensure!(
+            res_address == expected_proposer_address,
+            format!(
```
SuperFluffy, line 120: This `format!` is IMO not necessary. I think `ensure!` would print the value of `res_address` and `expected_proposer_address`

crates/astria-conductor/src/block_verifier.rs
```diff
@@ -0,0 +1,616 @@
...
+            METRO_VALIDATOR_ADDRESS_PREFIX,
+            block.header.proposer_address.0.to_base32(),
+            Variant::Bech32,
+        )
+        .wrap_err("failed converting bytes to bech32 address")?;
+
+        ensure!(
+            received_proposer_address == expected_proposer_address,
+            format!(
+                "proposer address mismatch: expected {}, got {}",
```
SuperFluffy, line 175: As above - I think you don't need to explicitly format here.

But even if you wanted to, you should probably prefer this:
```rust
        ensure!(
            received_proposer_address == expected_proposer_address,
            "proposer address mismatch: expected `{expected_propser_address}`, got `{received_proposer_address}`",
    )

crates/astria-conductor/src/driver.rs
noot: for the gossip network errors, we should continue (ie. we receive some invalid data and can't decode it)

it seems if we fail on `handle_driver_command`, we should probably abort. should I update here?

PR #124, opened 2023-06-23
sequencer accounts are now parsed into cometbft genesis.json by default

sambukowski:
#67 
This PR adds two features:
1. The astria-sequencer genesis.json accounts are merged automatically into the cometbft genesis.json when the sequencer is run.
2. A small utility for easily adding or removing accounts from the astria-sequencer genesis.json.


joroshiba (commented):
Should we have a default cometbft genesis file in here as well?

crates/astria-sequencer/src/config.rs
```diff
@@ -5,15 +5,20 @@ use serde::{
...
 const DEFAULT_ABCI_LISTEN_ADDR: &str = "127.0.0.1:26658";
+const DEFAULT_GENESIS_ACCOUNT_ALLOCATIONS_PATH: &str = "accounts-genesis.csv";
+const DEFAULT_COMETBFT_GENESIS_FILE: &str = "/.cometbft/config/genesis.json";
 
 #[derive(Debug, Deserialize, Parser, Serialize)]
 pub struct Config {
     /// The endpoint on which Sequencer will listen for ABCI requests
     #[arg(long, default_value_t = String::from(DEFAULT_ABCI_LISTEN_ADDR))]
     pub(crate) listen_addr: String,
     /// The path to the json encoded genesis file with a list of accounts.
```
SuperFluffy, line 16: I think this comment is not correct given that this now assumes a csv. However, I would probably just leave it as a json and not introduce extra dependencies.

crates/astria-sequencer/src/config.rs
```diff
@@ -5,15 +5,20 @@ use serde::{
...
-    #[arg(long)]
+    #[arg(long, default_value_t = String::from(DEFAULT_GENESIS_ACCOUNT_ALLOCATIONS_PATH))]
     pub(crate) genesis_file: String,
+    /// The path to the json encoded genesis file with a list of accounts.
+    #[arg(long, default_value_t = String::from(DEFAULT_COMETBFT_GENESIS_FILE))]
+    pub(crate) cometbft_genesis_file: String,
```
SuperFluffy, line 21: The naming is confusing - what's the difference between `cometbft_genesis_file` and `genesis_file`?

crates/astria-sequencer/accounts-genesis.csv
```diff
@@ -0,0 +1,4 @@
+"address","balance"
+"alice","1000000000000000000"
```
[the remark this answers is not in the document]
SuperFluffy: CSV files are very restrictive. I would prefer letting this be a json.

crates/astria-sequencer/src/genesis.rs
```diff
@@ -18,9 +24,54 @@ pub(crate) struct GenesisState {
...
+
+        Ok(Self {
+            accounts: records,
+        })
+    }
+
+    /// Add the Account data from `GenesisState` to the `CometBFT` genesis.json file
+    pub(crate) fn propagate_accounts_to(&self, path: &str) -> anyhow::Result<()> {
+        // build the absolute path to the json file you want to add the accounts to
+        let mut home_path = env::var("HOME")?;
```
SuperFluffy, line 45: Please don't read any extra env vars from here. The main entry point should be `config.rs`. `Config` should be the final source of truth for all environmental access and configuration.

crates/astria-sequencer/src/genesis.rs
```diff
@@ -18,9 +24,54 @@ pub(crate) struct GenesisState {
...
+            records.push(record);
+        }
+
+        Ok(Self {
+            accounts: records,
+        })
+    }
+
+    /// Add the Account data from `GenesisState` to the `CometBFT` genesis.json file
+    pub(crate) fn propagate_accounts_to(&self, path: &str) -> anyhow::Result<()> {
```
SuperFluffy, line 43: So this overrides the cometbft genesis file from under it? I don't know how I feel about this. IMO `sequencer` third tools should never change files in the config/cache/etc directories of other tools.

crates/astria-sequencer/src/genesis.rs
```diff
@@ -18,9 +24,54 @@ pub(crate) struct GenesisState {
...
+        let abs_destination_json_file_path = Path::new(&home_path);
+        let dest_file = File::open(abs_destination_json_file_path)
+            .context("failed to open destination genesis json file")?;
+
+        let mut dest_state: Value = serde_json::from_reader(&dest_file)
+            .context("failed deserializing genesis state from file")?;
+
+        // convert the accounts in GenesisState into a json Value
+        let mut json_map: serde_json::Map<String, Value> = serde_json::Map::new();
+        let mut accounts: Vec<Value> = Vec::new();
```
SuperFluffy, line 56: If you implement (or derive) an appropriate `serde::ser::Serialize` trait for `GenesisState` all this can be avoided.

crates/astria-sequencer/src/genesis.rs
```diff
@@ -29,3 +80,19 @@ pub(crate) struct Account {
     pub(crate) address: Address,
     pub(crate) balance: Balance,
 }
+
+fn merge_json_values(a: &mut Value, b: Value) {
```
SuperFluffy, line 84: I am not sure what the purpose of this (recursive) function is.

SuperFluffy (changes requested):
IMO too much is happening here during startup of `sequencer`.

By the looks of it this could (should?) be a free standing utility that is called on the cometbft genesis file to make it absolutely obvious what's happening and that we are editting the config dir of another service.

`Sequencer` should not writing to the cometbft config dirs during startup. I am also not sure about the flow here. Is cometbft init first run, then sequencer is started, followed by cometbft?

In a containerized environment there are no guarantees that this actually happens in that order.

crates/astria-sequencer/src/config.rs
noot: yeah, maybe I'd call the file with the initial accounts `initial_app_state` or something

crates/astria-sequencer/accounts-genesis.csv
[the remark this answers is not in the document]
noot: yeah I prefer json also, because there might be other things in the initial app state other than balances, so it'd be easier to separate them by module

crates/astria-sequencer/src/genesis.rs
noot: pretty sure this is what the cosmos-sdk does :/ maybe we should copy the cometbft default genesis file and write to that?

crates/astria-sequencer/src/genesis.rs
```diff
@@ -1,10 +1,16 @@
 use std::{
+    env,
     fs::File,
     path::Path,
 };
 
 use anyhow::Context;
```
noot, line 7: if this moves to its own crate `sequencer-utils` can change this to eyre

crates/astria-sequencer/src/genesis.rs
```diff
@@ -18,9 +24,54 @@ pub(crate) struct GenesisState {
 }
 
 impl GenesisState {
-    pub(crate) fn from_path<P: AsRef<Path>>(path: P) -> anyhow::Result<Self> {
-        let file = File::open(path).context("failed to open file with genesis state")?;
-        serde_json::from_reader(&file).context("failed deserializing genesis state from file")
+    /// Load Account information from a CSV file
+    pub(crate) fn from_csv<P: AsRef<Path>>(path: P) -> anyhow::Result<Self> {
```
noot, line 28: can you add a unit test for this?

Issue #120, opened 2023-06-21
composer: initialize crate with config, tracing, and basic test


itamarreif:
Initialize the `composer` crate's scaffolding with a `main()` function that initializes:
```[tasklist]
- [x] tracing
- [x] `Searcher` service from config
- [ ] implement and test a `ready` endpoint
```

Issue #102, opened 2023-06-16
composer: wrap and submit transactions to rust sequencer


joroshiba:

joroshiba:
Blocked by #110

SuperFluffy:
Closed as part of #192

Issue #101, opened 2023-06-16
composer: read transactions from EVM mempool using ethers


joroshiba:

SuperFluffy:
Closed as part of #192

PR #125, opened 2023-06-27
Itamarreif/composer init

itamarreif:
This PR implements the tasks in #120, #101 and #102.


itamarreif:
A couple questions on managing memory in async rust & error handling:

1. in [`Searcher::run_subtasks()`](https://github.com/astriaorg/astria/blob/itamarreif/composer-init/crates/astria-composer/src/searcher/mod.rs#L81), I borrow fields from `self` and those are moved into the task that's spawned. this doesn't compile since the task with the borrowed values may outlive `self`. What is the proper way to launch this task?

2. how should i structure the error handling in async rust?
i'm generally a bit confused about how to structure my error handling strategy in a rust async program.
my general intuition is that errors that can't be handled internally by a task will be bubbled up to the searcher's run function, and if they cant be handled by searcher they should be bubbled up to main.
however i am unclear on how to return a `SearcherError` from `Searcher::run_subtasks()` to `Searcher::run()` from within a subtask that's spawned in `run_subtasks()`?

crates/astria-composer/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+color-eyre = { workspace = true }
+figment = { workspace = true, features = ["toml", "env"] }
+tokio = { workspace = true }
+serde = { workspace = true, features = ["derive"]  }
+clap = { workspace = true, features = ["derive"] }
+thiserror = { workspace = true }
+tracing = { workspace = true }
+tracing-subscriber = { workspace = true, features = ["ansi", "json"] }
+reqwest = { workspace = true, features = ["json"] }
+# TODO: should these be moved to workspace?
```
SuperFluffy, line 19: Yeah, should be done. When adding dependencies to a project it makes sense to check if any of the other crates have that dependency.

crates/astria-composer/src/searcher/api.rs
```diff
@@ -0,0 +1,58 @@
...
+
+pub(super) async fn run(api_url: &SocketAddr, state: &SearcherState) -> Result<(), ApiError> {
+    let api_router = Router::new()
+        .route("/healthz", get(healthz))
+        .with_state(state.clone()); // TODO: should use an Arc here instead of cloning, otherwise state is stale
+
+    axum::Server::bind(api_url)
+        .serve(api_router.into_make_service())
+        .await
+        .map_err(ApiError::ServerError)
```
SuperFluffy, line 28: You don't need to wrap that error here. If you have several long-running tasks that are at the same "level", you can just spawn all of them into their respective tasks and `tokio::select!` or `tokio::join!` on them.

In fact, I did something like that here: https://github.com/astriaorg/astria/pull/89/files#diff-79f07f63704e1b426ed89a94ebf5327ed84d4f64dd76efb420b8f84eaeabe0bc

(still need to merge it)

crates/astria-composer/src/searcher/api.rs
```diff
@@ -0,0 +1,58 @@
...
+#[derive(Debug, Error)]
+pub enum ApiError {
+    #[error("hyper error")]
+    ServerError(#[from] hyper::Error),
+}
+
+pub(super) async fn run(api_url: &SocketAddr, state: &SearcherState) -> Result<(), ApiError> {
+    let api_router = Router::new()
+        .route("/healthz", get(healthz))
+        .with_state(state.clone()); // TODO: should use an Arc here instead of cloning, otherwise state is stale
```
SuperFluffy, line 23: You probably want a `Arc<State>` instead of a `State(Arc<T>)`, but it's hard to tell without knowing what enters the state. Try to express what kind of state you want to maintain and what state is supposed to be shared between API calls?

crates/astria-composer/src/types/mod.rs
SuperFluffy, line 1: I am not a big fan of creating a generic `types` module and then throwing all sorts of types in there. I prefer to create types closer to their point of use. For example, I think `types::searcher` should actually live in the `searcher` module and then be exported from there.

crates/astria-composer/src/types/searcher.rs
```diff
@@ -0,0 +1,39 @@
...
+use color_eyre::eyre;
+use tokio_stream::Stream;
+
+/// A stream of events emitted by a [Collector](Collector).
+// TODO
+pub type CollectorStream<'a, E> = Pin<Box<dyn Stream<Item = E> + Send + 'a>>;
+
+/// Collector trait, which defines a source of events.
+#[async_trait]
+pub trait Collector<E>: Send + Sync {
```
SuperFluffy, line 14: I know it's very enticing to start out with high level abstractions, but I would recommend starting concrete `collector`, `strategy`, and `executor` types (just one of each), and then going from there.

Chances are, you will never need the abstractions. And also you will have a better understanding of what it is you are trying to abstract away, reducing the chances of your abstractions being leaky: https://en.wikipedia.org/wiki/Leaky_abstraction

crates/astria-composer/src/searcher/mod.rs
```diff
@@ -0,0 +1,117 @@
...
+    pub fn new(config: Config) -> Result<Self, SearcherError> {
+        // configure rollup tx collector
+        // configure rollup tx bundler
+        // configure rollup tx executor
+
+        // init searcher state
+        let state = State(Arc::new(()));
+
+        // parse api url from config and init api router
+        let api_url = config
```
SuperFluffy, line 47: I know this is not how it's done in the other crates of this repo (although the plan is to slowly change that).

But if you do this whole thing from scratch, just create a `Config` that already contains the final types that you expect. So if you know that `api_url` must be a `SocketAddr`, then just set it to that type in `Config` and let figment/serde handle that.

crates/astria-composer/src/types/searcher.rs
itamarreif: agreed, this was copied over from [atremis](https://github.com/paradigmxyz/artemis/blob/main/crates/artemis-core/src/types.rs) when i started the repo but don't think i need to use it so removed.

crates/astria-composer/src/types/mod.rs
itamarreif: I only left the `rollup` submodule in there because I think those are types that will be shared between the `BlockBuilder` and `Searcher` later on.

crates/astria-composer/Cargo.toml
itamarreif: should that be in this pr or a different one? would change lines in `conductor` and `sequencer-relayer` as well

Cargo.toml
```diff
@@ -9,6 +9,7 @@ members = [
   "crates/astria-sequencer",
   "crates/astria-sequencer-relayer",
   "crates/astria-sequencer-relayer-test",
+  "crates/astria-composer"
```
noot, line 12: nit but put this in alphabetical order?

crates/astria-composer/src/config/mod.rs
```diff
@@ -0,0 +1,190 @@
...
+fn default_log() -> String {
+    DEFAULT_LOG.into()
+}
+
+#[derive(Debug, Deserialize, Serialize, PartialEq)]
+/// The high-level config for creating an astria-composer service.
+pub struct Config {
+    /// Log level. One of debug, info, warn, or error
+    #[serde(default = "default_log")]
+    pub log: String,
```
noot, line 31: do we need a log option or can we just go wtth ` RUST_LOG`?

crates/astria-composer/src/config/searcher.rs
```diff
@@ -0,0 +1,119 @@
...
+    Serialize,
+};
+use thiserror::Error;
+
+use super::cli;
+use crate::types::rollup::ChainId;
+
+const DEFAULT_SEQUENCER_URL: &str = "127.0.0.1:1317";
+const DEFAULT_API_URL: &str = "127.0.0.1:8080";
+const DEFAULT_CHAIN_ID: &str = "ethereum";
```
noot, line 27: should there be a default chain ID at all? i guess makes it easier for testing?

crates/astria-composer/src/config/searcher.rs
```diff
@@ -0,0 +1,119 @@
...
+const DEFAULT_EXECUTION_RPC_URL: &str = "127.0.0.1:50051";
```
noot, line 28: why do we need this? what are we submitting to the gRPC execution server?

crates/astria-composer/src/config/searcher.rs
```diff
@@ -0,0 +1,119 @@
...
+    Deserialize,
+    Serialize,
+};
+use thiserror::Error;
+
+use super::cli;
+use crate::types::rollup::ChainId;
+
+const DEFAULT_SEQUENCER_URL: &str = "127.0.0.1:1317";
+const DEFAULT_API_URL: &str = "127.0.0.1:8080";
```
noot, line 26: which API?

crates/astria-composer/Cargo.toml
joroshiba: This is the appropriate place to make the change.

crates/astria-composer/src/config/searcher.rs
itamarreif: right now it's just `/healthz`, this should be changed to just a port number tho.

crates/astria-composer/src/config/searcher.rs
```diff
@@ -0,0 +1,119 @@
...
+use serde::{
+    Deserialize,
+    Serialize,
+};
+use thiserror::Error;
+
+use super::cli;
+use crate::types::rollup::ChainId;
+
+const DEFAULT_SEQUENCER_URL: &str = "127.0.0.1:1317";
```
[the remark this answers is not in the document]
itamarreif: agreed, just a placeholder for now

crates/astria-composer/src/config/searcher.rs
```diff
@@ -0,0 +1,119 @@
...
+    #[serde(default = "default_sequencer_url")]
+    pub sequencer_url: SocketAddr,
+
+    /// Chain ID that we want to connect to
+    #[serde(default = "default_chain_id")]
+    pub chain_id: ChainId,
+
+    /// Address of the RPC server for execution
+    #[serde(default = "default_execution_rpc_url")]
+    pub execution_rpc_url: SocketAddr,
```
joroshiba, line 56: What are we executing? 

I'm guessing the idea here is like as a block builder, but this might require some rethinking of APIs implemented on sequenced chains. The execution API drives the chain and adds blocks to the chain, not the idea here presumably. That would be work done by conductor.

crates/astria-composer/src/config/searcher.rs
itamarreif: will be using this to initialize an [`ethers::Provider`websocket connection](https://www.gakonst.com/ethers-rs/providers/ws.html) (the actual value will probably change, this is just a placeholder copied over from `conductor`)

crates/astria-composer/src/config/searcher.rs
itamarreif: yeah just easier for testing so you don't have to specify it as cli arg or env variable every time.

crates/astria-composer/src/config/searcher.rs
itamarreif: will be using this to initialize an ethers::Providerwebsocket connection (the actual value will probably change, this is just a placeholder copied over from conductor)

crates/astria-composer/src/config/mod.rs
itamarreif: good point

crates/astria-composer/src/config/searcher.rs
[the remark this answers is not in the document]
itamarreif: changed the default value to `rolluptest`.

some clarification: this is the chain_id value that will go into the [`SecondaryAction` tx](https://github.com/astriaorg/astria/blob/noot/secondary-txs/crates/astria-proto/proto/astria/sequencer/v1/transaction.proto#L36) that the composer will be submitting to the sequencer. When I continue with implementing the read & write operations this will probably be moved to a `RollupConfig` struct that'll hold the execution node's rpc addr for reading, the chain_id to write to and whatever other config will be needed. so it will eventually be `RollupConfig::chain_id` or something like that.

if you have any name suggestions i'm not at all attached, but this is mostly a placeholder for now.

crates/astria-composer/src/config/searcher.rs
itamarreif: https://github.com/astriaorg/astria/pull/125#discussion_r1248066300

crates/astria-composer/src/config/searcher.rs
joroshiba: Ok should definitely not be 50051 but instead 8546, can we rename to something like ETH_WS_URL?

Port 50051 is the port for driving execution through our gRPC [Execution API](https://github.com/astriaorg/astria/blob/main/crates/astria-proto/proto/astria/execution/v1/execution.proto)

crates/astria-composer/src/config/searcher.rs
joroshiba: As noted above I think we should rename this, since this is a eth_websocket connection

itamarreif: renamed to `execution_ws_url` and changed the port to 8546

joroshiba (approved):

crates/astria-composer/src/config/searcher.rs
noot: ok, so it's the API of the composer itself. cool

crates/astria-composer/src/searcher/api.rs
```diff
@@ -0,0 +1,59 @@
...
+        }
+        let (status, msg) = match self {
+            Self::Ok => (axum::http::StatusCode::OK, "ok"),
+            Self::_Degraded => (axum::http::StatusCode::GATEWAY_TIMEOUT, "degraded"),
+        };
+        let mut response = axum::Json(HealthzBody {
+            status: msg,
+        })
+        .into_response();
+        *response.status_mut() = status;
```
SuperFluffy, line 51: I think you might be able to call `(status, json_body).into_response()` because of this: https://docs.rs/axum/latest/axum/response/trait.IntoResponse.html#impl-IntoResponse-for-(StatusCode,+R)

itamarreif:
# quick summary before handoff 7/10
## tx collector
The collector (wrapper around ethers provider) is implemented in `src/searcher/collector.rs`. The constructor initializes from config info and `run()` subscribes to a stream of pending transactions from the node, converting each transaction that arrives to a `NewTx` `Event` (defined in `searcher/mod.rs`). These events are forwarded into the event channel.

## bundler
The bundler currently takes a single `NewTx` event and turns the `ethers::Transaction` in it to a `SendSequencerSecondaryTx` action with an `sequencer::Unsigned` inside it. While I got a sanity check from Janis that the approach works, I haven't had the time to test this so there are likely small things to catch there.

## executor
Executor is similarly implemented but not tested. The main thing remaining to implement here is handling transactions that fail because of a bad nonce. The approach i would take is resetting the bundler's nonce counter (there's an unused helper method for this that i added) and putting the unsigned transaction back into the channel for processing (will have to be signed again for new nonce).

# remaining work
- features
- [ ] add `events` and `actions` endpoints to api, to inspect current pending events and actions (nice to have, shouldn't be too much work)
- fixes
    - [ ] logic for handling invalid nonce on a sequencer. see the `todo!` in `searcher::run()`.
    - [ ] change the `healthz` api endpoint to `readyz`
- code cleanup
    - [ ] some of the errors can probably be cleaned up or improved
    - [ ] some function names can probably be renamed
    - [ ] move the event and action channel buffer sizes to config (currently just hard coded to 512)
- sequencer & sequencer_client-related items
    - [ ] the constructor for `sequence::action::Action` in the `sequencer` crate needs to be made public (i made this change locally but didn't commit because wasn't sure if there was context to this I was missing)
    - [ ] sequencer transaction types (`Unsigned`, `Signed`, `Action`, etc.) should be reexported from the `sequencer_client` so that composer (and other users of the `sequencer_client`) do not need to depend on the `sequencer` crate.
    - [ ] `SigningKey` should be exported from `sequencer_client` instead of requiring and extra dependency on `ed25519_consensus` for using `sequencer_client`
- tests
    - [ ] unit test for bundler's:
        - [ ] `get_nonce()`
        - [ ] `NewTx` handler (after moving out of )
    - [ ] unit test for executor's `SendSequencerSecondaryTx` handler
    - [ ] e2e tests for pulling a transaction from the geth node and submitting to the sequencer
    - [ ] set up k8s stuff in tests?
- general housekeeping
    - [ ] get rid of cli args and replace with env variables
    - [ ] containerization and adding to ci

PR #126, opened 2023-06-27
added astria-sequencer-utils crate with sequencer -> cometbft genesis…

sambukowski:
#67 new `astria-sequencer-utils` crate added with cli tool for parsing one json genesis file into another.


crates/astria-sequencer-utils/src/telemetry.rs
SuperFluffy, line 1: You can rm all of this tracing setup IMO

crates/astria-sequencer-utils/src/main.rs
```diff
@@ -0,0 +1,28 @@
...
+    },
+    genesis_parser::GenesisParser,
+    telemetry,
+};
+use tracing::info;
+
+#[tokio::main]
+async fn main() {
+    let config = Config::get();
+    telemetry::init(std::io::stdout).expect("failed to initialize telemetry");
```
SuperFluffy, line 14: Setting up telemetry is probably overkill for a small utility like this. I think just printing to stdout/stderr is sufficient.

crates/astria-sequencer-utils/src/genesis_parser.rs
```diff
@@ -0,0 +1,68 @@
...
+    instrument,
+};
+
+use crate::config::GenesisParserArgs;
+
+pub struct GenesisParser;
+
+impl GenesisParser {
+    #[instrument(skip_all)]
+    pub async fn propigate_data(data: GenesisParserArgs) -> Result<()> {
```
SuperFluffy, line 25:
```suggestion
    pub async fn propagate_data(data: GenesisParserArgs) -> Result<()> {
```

crates/astria-sequencer-utils/src/genesis_parser.rs
```diff
@@ -0,0 +1,68 @@
+use std::{
+    fs::File,
+    path::Path,
+};
+
+use anyhow::{
```
SuperFluffy, line 6: Let's use `eyre` instead of `anyhow` (and `WrapErr` with `my_result.wrap_err` instead of `Context` with `my_result.context`).

crates/astria-sequencer-utils/src/genesis_parser.rs
```diff
@@ -0,0 +1,68 @@
...
+            "loading genesis json data for propigation"
+        );
+        // load sequencer genesis data
+        let source_genesis_file_path = File::open(data.source_genesis_file)
+            .context("failed to open sequencer genesis file")?;
+        let source_genesis_data: Value = serde_json::from_reader(&source_genesis_file_path)
+            .context("failed deserializing sequencer genesis state from file")?;
+
+        // load cometbft genesis data
+        let destination_genesis_file_path = File::open(data.destination_genesis_file.clone())
```
SuperFluffy, line 38:
```suggestion
        let destination_genesis_file_path = File::open(&data.destination_genesis_file)
```
nbd for a small binary but the clone isn't needed

crates/astria-sequencer-utils/src/genesis_parser.rs
```diff
@@ -0,0 +1,68 @@
...
+        let dest_file = File::create(Path::new(data.destination_genesis_file.as_str()))
+            .context("failed to open destination genesis json file")?;
+        to_writer_pretty(dest_file, &destination_genesis_data)?;
+
+        Ok(())
+    }
+}
+
+/// Merges a source JSON Value into a destination JSON Value.
+fn merge_values(a: &mut Value, b: &Value) {
```
SuperFluffy, line 57: Slightly simplified version (no need to reborrow using `ref mut` anymore because with [match ergonomics](https://rust-lang.github.io/rfcs/2005-match-ergonomics.html) beingi merged for a few years now all that reborrowing is done for you (so the `a` is automatically a `&mut`, and the `b` a `&`).

Also, `serde_json::Map::entry` expects a `K: Into<String>`, and since `k: &String` and `impl From<&String> for String` you don't need to clone:

```rust
fn merge_json(a: &mut Value, b: &Value) {
    match (a, b) {
        (Value::Object(a), Value::Object(b)) => {
            for (k, v) in b {
                merge_json(a.entry(k).or_insert(Value::Null), v);
            }
        }
        (a, b) => *a = b.clone(),
    }
}
```

crates/astria-sequencer-utils/src/config.rs
```diff
@@ -0,0 +1,33 @@
+use clap::Parser;
+use serde::{
+    Deserialize,
+    Serialize,
+};
+
+#[derive(Debug, Deserialize, Parser, Serialize)]
```
SuperFluffy, line 7: Also no need for `Deserialize, Serialize` here. But a doc comment on this struct would be nice. `clap` takes that and creates a cli help msg from it.

SuperFluffy (commented):
Some comments. I agree that a single binary makes more sense then shoving all of that into astria-sequencer.

I believe I would prefer several single binaries over one binary with subcommands, and place that collection in `src/bin/`.

Also, can you please add a test for the functionality? I.e. what the final state should look like if you merged 2 jsons?

I also found a discussion re json merging here: https://stackoverflow.com/questions/47070876/how-can-i-merge-two-json-objects-with-rust

Can you add that as a comment in case people think of json-patch and wonder about how how keys with a `null` value are handled?

crates/astria-sequencer-utils/src/config.rs
```diff
@@ -0,0 +1,33 @@
...
+pub struct Config {
+    #[clap(subcommand)]
+    pub cmd: Command,
+}
+
+#[derive(Debug, Deserialize, Parser, Serialize)]
```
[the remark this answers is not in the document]
sambukowski: removed this

crates/astria-sequencer-utils/src/main.rs
```diff
@@ -0,0 +1,28 @@
...
+    config::{
+        Command,
+        Config,
+    },
+    genesis_parser::GenesisParser,
+    telemetry,
+};
+use tracing::info;
+
+#[tokio::main]
```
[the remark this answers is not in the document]
sambukowski: removed all async

crates/astria-sequencer-utils/src/genesis_parser.rs
sambukowski: fixed this

crates/astria-sequencer-utils/src/main.rs
sambukowski: i removed all the telemetry

crates/astria-sequencer-utils/src/config.rs
```diff
@@ -0,0 +1,33 @@
+use clap::Parser;
+use serde::{
+    Deserialize,
+    Serialize,
+};
+
+#[derive(Debug, Deserialize, Parser, Serialize)]
+pub struct Config {
+    #[clap(subcommand)]
```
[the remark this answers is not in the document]
sambukowski: made this a single binary. if we decide otherwise it's not that difficult to go back

crates/astria-sequencer-utils/src/config.rs
```diff
@@ -0,0 +1,33 @@
...
+
+#[derive(Debug, Deserialize, Parser, Serialize)]
+pub enum Command {
+    #[clap(name = "genesis-parse")]
+    GenesisParser(GenesisParserArgs),
+}
+
+#[derive(Debug, Deserialize, Parser, Serialize)]
+pub struct GenesisParserArgs {
+    #[clap(long, short = 's')]
```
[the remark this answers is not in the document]
sambukowski: removed the short

crates/astria-sequencer-utils/src/genesis_parser.rs
sambukowski: removed the clone and replaced with .as_str(). the dest path is used in a few places

crates/astria-sequencer-utils/src/config.rs
sambukowski: removed these

crates/astria-sequencer-utils/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+anyhow = "1"
+is-terminal = "0.4.7"
+clap = { workspace = true, features = ["derive"] }
+serde = { workspace = true, features = ["derive"] }
+serde_json = { workspace = true }
+tokio = { workspace = true, features = [
```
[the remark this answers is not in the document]
sambukowski: removed this

crates/astria-sequencer-utils/src/genesis_parser.rs
sambukowski: replaced the function with the code above but kept getting yelled at by the compiler so I've left the clone for now.

crates/astria-sequencer-utils/src/genesis_parser.rs
sambukowski: converted from anyhow to eyre

crates/astria-sequencer-utils/src/genesis_parser.rs
```diff
@@ -0,0 +1,68 @@
...
+            serde_json::from_reader(&destination_genesis_file_path)
+                .context("failed deserializing cometbft genesis state from file")?;
+
+        // merge sequencer genesis data into cometbft genesis data
+        merge_values(&mut destination_genesis_data, &source_genesis_data);
+
+        // write new state
+        let dest_file = File::create(Path::new(data.destination_genesis_file.as_str()))
+            .context("failed to open destination genesis json file")?;
+        to_writer_pretty(dest_file, &destination_genesis_data)?;
```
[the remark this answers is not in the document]
sambukowski: context added

crates/astria-sequencer-utils/README.md
```diff
@@ -0,0 +1,18 @@
+# astria-sequencer-utils
+
+## Requirements
+
+- rust 1.68
+
+## Usage
+
+Running requires two flags: `--source-genesis-file` and `--destination-genesis-file`.
```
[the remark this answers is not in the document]
sambukowski: I had something like that initially and @SuperFluffy requested what's here now. if we do need to add more functionality it's pretty easy to add back in.

crates/astria-sequencer-utils/src/genesis_parser.rs
```diff
@@ -0,0 +1,63 @@
...
+
+        Ok(())
+    }
+}
+
+/// Merges a source JSON Value into a destination JSON Value.
+// inpiration for this function came from: https://stackoverflow.com/questions/47070876/how-can-i-merge-two-json-objects-with-rust
```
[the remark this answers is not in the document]
sambukowski: fixed

joroshiba (approved):

crates/astria-sequencer-utils/src/genesis_parser.rs
```diff
@@ -0,0 +1,141 @@
...
+
+impl GenesisParser {
+    /// Propagates json data from one json file to another.
+    ///
+    /// # Errors
+    ///
+    /// An `eyre::Result` is returned if either source genesis files cannot be opened,
+    /// or if the destination genesis file cannot be saved.
+    pub fn propagate_data(data: Config) -> Result<()> {
+        println!("loading genesis json data for propigation:");
```
noot, line 27:
```suggestion
        println!("loading genesis json data for propagation:");
```

crates/astria-sequencer-utils/src/genesis_parser.rs
```diff
@@ -0,0 +1,141 @@
...
+
+pub struct GenesisParser;
+
+impl GenesisParser {
+    /// Propagates json data from one json file to another.
+    ///
+    /// # Errors
+    ///
+    /// An `eyre::Result` is returned if either source genesis files cannot be opened,
+    /// or if the destination genesis file cannot be saved.
```
noot, line 25: also if the file can be opened, but not deserialized

crates/astria-sequencer-utils/src/telemetry.rs
```diff
@@ -0,0 +1,60 @@
+use is_terminal::IsTerminal as _;
```
noot, line 1: it doesn't look like this file is used anywhere? you can delete

noot (approved):
looks good, some comments to be addressed before merging though!

crates/astria-sequencer-utils/src/config.rs
```diff
@@ -0,0 +1,17 @@
+use clap::Parser;
+
+#[derive(Debug, Parser)]
+pub struct Config {
```
SuperFluffy, line 4: If you add a doc comment above this type describing the app, then clap will automatically derive a nice help page from that.

crates/astria-sequencer-utils/src/config.rs
```diff
@@ -0,0 +1,17 @@
...
+    #[clap(long)]
+    pub source_genesis_file: String,
```
SuperFluffy, line 6: same for these. If you have `/// <what does this opt do>` above the field, clap will give you a description.

crates/astria-sequencer-utils/src/genesis_parser.rs
```diff
@@ -0,0 +1,141 @@
...
+            .wrap_err("failed to open cometbft genesis file")?;
+        let mut destination_genesis_data: Value =
+            serde_json::from_reader(&destination_genesis_file_path)
+                .wrap_err("failed deserializing cometbft genesis state from file")?;
+
+        // merge sequencer genesis data into cometbft genesis data
+        merge_json(&mut destination_genesis_data, &source_genesis_data);
+
+        // write new state
+        let dest_file = File::create(Path::new(data.destination_genesis_file.as_str()))
```
SuperFluffy, line 50: You can simplify this with:
`File::create(&data.destination_genesis_file)` because `File::create` takes a `AsRef<Path>` and you have `impl<'a> AsRef<Path> for &'a str`

crates/astria-sequencer-utils/src/genesis_parser.rs
```diff
@@ -0,0 +1,141 @@
...
+            data.destination_genesis_file
+        );
+        // load sequencer genesis data
+        let source_genesis_file_path = File::open(data.source_genesis_file)
+            .wrap_err("failed to open sequencer genesis file")?;
+        let source_genesis_data: Value = serde_json::from_reader(&source_genesis_file_path)
+            .wrap_err("failed deserializing sequencer genesis state from file")?;
+
+        // load cometbft genesis data
+        let destination_genesis_file_path = File::open(data.destination_genesis_file.as_str())
```
SuperFluffy, line 40: You will have 2 file handles open at the same time to the same file. 

Better read in this string in one go so that you only have one handle (this assumes that the input json will never be too large, which I think is reasonable:
```suggestion
        let dst_contents = std::fs::read_to_string(&data.destination_genesis_file).wrap_err("failed reading cometbft genesis file")?;
```

SuperFluffy (approved):
Looks good to me. Nit on the file handles existing at the same time (you are truncating/overriding a file while holding another handle).

Also `telemetry.rs` is redundant and can be deleted.

PR #127, opened 2023-06-28
feat(sequencer): implement `SequenceAction` type; refactor transactions to have multiple actions

### LINKED ISSUE #65: sequencer: implement "secondary" transaction type
implement an un-executed transaction type that is a tx destined for a rollup:
```
pub struct SecondaryTransaction {
    tx: Vec<u8>,
    chain_id: Vec<u8>,
    signature,
    public_key,
}
```

validation will basically just be validating the signature.

blocked by #64 (at least the signature / public key parts)

noot:
- refactor transaction type to contain multiple actions; now one transaction can do multiple actions atomically
- idea is that each action acts on a specific module, the only two of which we currently have are `accounts` and `sequence`
- update previous `accounts::Transaction` to be named `TransferAction`
- implement `SequenceAction` type which is an action which contains a chain ID (ie the destination rollup) and opaque bytes

closes #65


crates/astria-sequencer/src/transaction/unsigned.rs
```diff
@@ -53,21 +84,76 @@ impl Transaction {
     }
 }
 
+#[async_trait::async_trait]
+impl ActionHandler for Transaction {
+    fn check_stateless(&self) -> Result<()> {
+        for action in &self.actions {
+            match action {
+                Action::AccountsAction(_) | Action::SecondaryAction(_) => {}
```
joroshiba, line 92: Should we be calling the `check_stateless` for each action? to check?

crates/astria-sequencer/src/transaction/unsigned.rs
```diff
@@ -53,21 +84,76 @@ impl Transaction {
...
+        }
+        Ok(())
+    }
+
+    async fn check_stateful<S: StateReadExt + 'static>(
+        &self,
+        state: &S,
+        from: &Address,
+    ) -> Result<()> {
+        let curr_nonce = state.get_account_nonce(from).await?;
```
joroshiba, line 103: Should this be also checking the stateful properties on the actions?

If so is there some sort of pattern we should be using to enforce this? I feel like if each added action needs a new line in stateless, stateful, and execute someone might later forget to do all of that and it should fail something.

crates/astria-proto/proto/astria/sequencer/v1/transaction.proto
```diff
@@ -14,16 +14,26 @@ message SignedTransaction {
...
 // Note: all values must be set (ie. not `None`), otherwise it will
 // be considered invalid by the sequencer.
-message AccountsTransaction {
+message Transfer {
     bytes to = 1;
     astria.primitive.v1.Uint128 amount = 2;
-    uint32 nonce = 3;
+}
+
+message SecondaryAction {
```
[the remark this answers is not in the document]
noot: why does the secondary action need a nonce? the sequencer should be opaque to whatever data/transaction format the rollup is using. the nonce in `UnsignedTransaction` is sufficient for users to be able to order things how they want

and yeah, I wasn't totally sure about naming, how do `TransferAction` and `SequenceAction` sound?

crates/astria-sequencer/src/transaction/unsigned.rs
noot: yeah, I omitted it + the `check_stateless` above because the checks don't actually do anything right now. but added because it should be there.

do you mean some sort of compile-time check?

crates/astria-sequencer/src/transaction/action.rs
```diff
@@ -0,0 +1,55 @@
...
+    Action as ProtoAction,
+};
+use serde::{
+    Deserialize,
+    Serialize,
+};
+
+use crate::{
+    accounts::TransferAction,
+    sequence::Action as SequenceAction,
```
SuperFluffy, line 16: You can just refer to this as `sequence::Action`. No need to give it an alias.

crates/astria-proto/proto/astria/sequencer/v1/transaction.proto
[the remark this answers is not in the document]
joroshiba: Gotcha yeah I see what you mean on the secondary nonce, it feels like it should be needed but you get adequate control without it. 

New names seem good to me.

crates/astria-sequencer/src/transaction/unsigned.rs
joroshiba: I think this is where something like the manager setup in cosmos is "correct" but we can refactor later, with only two things going on it's not necessary but as our action types grow it makes sense to have a system for ensuring each action is called.

crates/astria-sequencer/src/transaction/action.rs
```diff
@@ -0,0 +1,55 @@
...
+    },
+    Action as ProtoAction,
+};
+use serde::{
+    Deserialize,
+    Serialize,
+};
+
+use crate::{
+    accounts::TransferAction,
```
SuperFluffy, line 15: Clarifying Q: will there be more actions on the accounts module?

crates/astria-sequencer/src/transaction/action.rs
```diff
@@ -0,0 +1,55 @@
...
+    sequence::Action as SequenceAction,
+};
+
+/// Represents an action on a specific module.
```
SuperFluffy, line 19: Nit: add an empty line in the markdown

crates/astria-sequencer/src/transaction/action.rs
```diff
@@ -0,0 +1,55 @@
...
+
+    pub(crate) fn try_from_proto(proto: &ProtoAction) -> Result<Self> {
+        Ok(
+            match proto
+                .value
+                .as_ref()
+                .ok_or_else(|| anyhow::anyhow!("missing value"))?
+            {
+                ProtoTransferAction(tx) => {
+                    Action::TransferAction(TransferAction::try_from_proto(tx)?)
```
SuperFluffy, line 49: Error context

crates/astria-sequencer/src/transaction/signed.rs
```diff
@@ -78,16 +71,8 @@ impl Signed {
         let Some(proto_transaction) = proto.transaction else {
             bail!("transaction is missing");
         };
+        let transaction = Unsigned::try_from_proto(&proto_transaction)?;
```
SuperFluffy, line 74: Error context

crates/astria-sequencer/src/transaction/signed.rs
```diff
@@ -112,29 +97,25 @@ impl Signed {
...
-impl ActionHandler for Signed {
+impl Signed {
     #[instrument]
-    fn check_stateless(&self) -> Result<()> {
+    pub(crate) fn check_stateless(&self) -> Result<()> {
         self.verify_signature()?;
-        match &self.transaction {
-            Unsigned::AccountsTransaction(_) => Ok(()),
-        }
+        self.transaction.check_stateless()?;
```
SuperFluffy, line 104: Error context

crates/astria-sequencer/src/transaction/unsigned.rs
```diff
@@ -54,21 +79,86 @@ impl Unsigned {
     }
 }
 
+#[async_trait::async_trait]
+impl ActionHandler for Unsigned {
+    fn check_stateless(&self) -> Result<()> {
+        for action in &self.actions {
+            match action {
+                Action::TransferAction(tx) => tx.check_stateless()?,
```
SuperFluffy, line 87: Error context for this and the call below to know what went wrong

crates/astria-sequencer/src/transaction/unsigned.rs
```diff
@@ -54,21 +79,86 @@ impl Unsigned {
...
+        state: &S,
+        from: &Address,
+    ) -> Result<()> {
+        let curr_nonce = state.get_account_nonce(from).await?;
+        ensure!(curr_nonce < self.nonce, "invalid nonce");
+
+        // do we need to make a StateDelta here so we can check the actions on the successive state?
+        for action in &self.actions {
+            match action {
+                Action::TransferAction(tx) => tx.check_stateful(state, from).await?,
```
SuperFluffy, line 105: Error context on this and below

crates/astria-sequencer/src/transaction/unsigned.rs
```diff
@@ -54,21 +79,86 @@ impl Unsigned {
...
+            }
+        }
+
+        Ok(())
+    }
+
+    #[instrument(
+        skip_all,
+        fields(
+            nonce = self.nonce.into_inner(),
```
SuperFluffy, line 116: Also print the address as a base64 field if applicable? I think it already implements Display?

crates/astria-sequencer/src/transaction/unsigned.rs
```diff
@@ -54,21 +79,86 @@ impl Unsigned {
...
+            .get_account_nonce(from)
+            .await
+            .context("failed getting `from` nonce")?;
+        state
+            .put_account_nonce(from, from_nonce + Nonce::from(1))
+            .context("failed updating `from` nonce")?;
+
+        for action in &self.actions {
+            match action {
+                Action::TransferAction(tx) => {
```
SuperFluffy, line 132: bunch of error contexts here and below

crates/astria-sequencer/src/transaction/action_handler.rs
```diff
@@ -5,9 +5,12 @@ use penumbra_storage::{
     StateWrite,
 };
 
+use crate::accounts::types::Address;
+
 #[async_trait]
 pub(crate) trait ActionHandler {
     fn check_stateless(&self) -> Result<()>;
```
SuperFluffy, line 12: If a lot (most?) actions don't do anything here anyways, you could provide a default impl for this:
```suggestion
    fn check_stateless(&self) -> Result<()> {
        Ok(())
    }
```
Then you can only impl that where you actually want custom behavior.

crates/astria-sequencer/src/sequence/action.rs
```diff
@@ -0,0 +1,71 @@
...
+    pub(crate) fn from_proto(proto: &ProtoSequenceAction) -> Self {
+        Self {
+            chain_id: proto.chain_id.clone(),
+            data: proto.data.clone(),
+        }
+    }
+}
+
+#[async_trait::async_trait]
+impl ActionHandler for Action {
```
SuperFluffy, line 54: As mentioned in the `ActionHandler` trait - you could just throw all of these default impls into the trait definition, i.e. make it look like so:
```rust
trait Action {
    fn check_stateless(&self) -> Result<()> {
        Ok(())
    }

    async fn check_stateful<S: StateReadExt + 'static>(
        &self,
        state: &S,
        from: &Address,
    ) -> Result<()> {
        Ok(())
    }

    #[instrument(skip_all)]
    async fn execute<S: StateWriteExt>(&self, state: &mut S, from: &Address) -> Result<()> {
        Ok(())
    }
}
```
Then you don't have to explicitly impl all these no-ops (unless you want to do that explicitly of course)

crates/astria-sequencer/src/transaction/action.rs
```diff
@@ -0,0 +1,55 @@
+use anyhow::Result;
+use astria_proto::sequencer::v1::{
+    action::Value::{
```
SuperFluffy, line 3: Since these enum variants only show up in function bodies below, you can just import them locally so you don't have to pollute your module namespace. I.e. smth like (changed the type alias to be on `Value` so that not every variant has to be renamed):
```rust
fn my_fn() {
    use astria_proto::sequencer::v1::action::Value as ProtoValue;
    match v {
        ProtoValue::SequenceAction ...
   }
}
```

SuperFluffy (approved):
Looks good to me.

Please add error contexts everywhere, and some text on what a `SequenceActions` is supposed to be

crates/astria-sequencer/src/transaction/action.rs
noot: not currently for in the near future, theoretically if we implement protocol-level account abstraction or native tokens there could be others

PR #131, opened 2023-06-30
maintainence: remove bech32 addresses

### LINKED ISSUE #97: Remove all bech32 addresses
Bech32 addresses are cosmos specific while we want to be agnostic with respect to proposer address.

CC @noot

noot:
- remove usages of bech32 addresses since they're cosmos-specific

closes #97

### DIFF
--- a/crates/astria-conductor/src/block_verifier.rs
+++ b/crates/astria-conductor/src/block_verifier.rs
@@ -14,11 +14,6 @@ use astria_sequencer_relayer::{
         CommitSig,
     },
 };
-use bech32::{
-    self,
-    ToBase32,
-    Variant,
-};
 use color_eyre::eyre::{
     self,
     bail,
@@ -60,8 +55,6 @@ use crate::tendermint::{
     TendermintClient,
     ValidatorSet,
 };
-
-const METRO_VALIDATOR_ADDRESS_PREFIX: &str = "metrovalcons";
 
 /// `BlockVerifier` is responsible for verifying the correctness of a block
 /// before executing it.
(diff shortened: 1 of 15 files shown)


joroshiba (commented):
Can we merge this not into main but into a new feature branch, so that conductor and sequencer-relayer remain compatible with metro until we can test everything works with retro? ie `feature/retro`

noot:
@joroshiba this is still compatible with metro, it just doesn't use bech32 addresses internally anymore. I just tested it with metro and it's fine!

crates/astria-conductor/src/block_verifier.rs
```diff
@@ -153,18 +149,21 @@ impl BlockVerifier {
...
-            .get_proposer()
-            .wrap_err("failed to get proposer from validator set")?
-            .address;
+        let expected_proposer_address = public_key_bytes_to_address(
+            validator_set
+                .get_proposer()
+                .wrap_err("failed to get proposer from validator set")?
+                .pub_key
+                .key
+                .0
```
SuperFluffy, line 158: I'm not usually too concerned with this, but this violates the law of demeter too badly.

Can you provide a function that takes the pub key and generates the final result from there?

crates/astria-conductor/src/block_verifier.rs
```diff
@@ -153,18 +149,21 @@ impl BlockVerifier {
...
 
         // check if the proposer address matches the sequencer block's proposer
-        let received_proposer_address = bech32::encode(
-            METRO_VALIDATOR_ADDRESS_PREFIX,
-            block.header.proposer_address.0.to_base32(),
-            Variant::Bech32,
-        )
-        .wrap_err("failed converting bytes to bech32 address")?;
+        let received_proposer_address =
+            AccountId::try_from(block.header.proposer_address.0.clone())
```
SuperFluffy, line 165: Same with this

crates/astria-conductor/src/block_verifier.rs
```diff
@@ -255,7 +260,10 @@ fn ensure_commit_has_quorum(
     let validator_map = validator_set
         .validators
         .iter()
-        .map(|v| (&v.address, v)) // address is in bech32
+        .filter_map(|v| {
+            let address = public_key_bytes_to_address(&v.pub_key.key.0).ok()?;
```
SuperFluffy, line 264: And here

crates/astria-conductor/src/block_verifier.rs
```diff
@@ -267,18 +275,14 @@ fn ensure_commit_has_quorum(
         }
 
         // verify validator exists in validator set
-        let validator_address = bech32::encode(
-            METRO_VALIDATOR_ADDRESS_PREFIX,
-            vote.validator_address.0.to_base32(),
-            Variant::Bech32,
-        )
-        .wrap_err("failed to encode validator address to bech32")?;
+        let validator_address = AccountId::try_from(vote.validator_address.0.clone())
```
SuperFluffy, line 278: And here

SuperFluffy (approved):
I do agree with the changes and welcome removal of bech32, but I'm not a fan of the deep field accesses. Can you refactor that a bit?

See this for the LoD I referenced: https://en.m.wikipedia.org/wiki/Law_of_Demeter

noot:
@SuperFluffy done!

SuperFluffy:
> @SuperFluffy done!

@noot  There are still a few instances where parent types reach very deep into their fields.

PR #137, opened 2023-07-05
sequencer-relayer, conductor: use rust sequencer node instead of metro

### LINKED ISSUE #21: use tendermint.rs conversion methods for headers
I noticed that we are converting to tendermint headers ourselves. We should rely on their upstream impls to save some work: https://docs.rs/tendermint/0.30.0/tendermint/block/header/struct.Header.html#impl-TryFrom%3CHeader%3E-for-Header:

https://github.com/astriaorg/sequencer-relayer/blob/442a1723a61ab0aec5e44626a1bb8c74b12f76f3/src/types.rs#L125-L126

### LINKED ISSUE #98: use tendermint-rs types for block commits

### LINKED ISSUE #99: update relayer to work with rust sequencer

noot:
- update sequencer-relayer and conductor to use "retro" + cometbft instead of metro
- remove all duplicate tendermint types and use the tendermint-rs ones
- TODO: need #139 to fix all the tests

closes #99
closes #98
closes #21

note: this breaks the `data_hash` check, as a sequencer transaction can have multiple actions, it quickly becomes very cumbersome to try to re-create the full transaction data on the conductor side to verify the merkle root. I decided to remove all the jank that was allowing only txs with 1 "sequence action"/"sequencer msg" through to rollups; in a follow up PR I'll re-add the data_hash verification.


crates/astria-conductor/src/executor.rs
```diff
@@ -68,15 +64,13 @@ pub(crate) async fn spawn(conf: &Config, alert_tx: AlertSender) -> Result<(JoinH
     Ok((join_handle, executor_tx))
 }
 
-// Given a string, convert to protobuf timestamp
-fn convert_str_to_prost_timestamp(value: &str) -> Result<ProstTimestamp> {
-    let time =
-        Time::parse_from_rfc3339(value).wrap_err("failed parsing string as rfc3339 datetime")?;
+// Given `Time`, convert to protobuf timestamp
+fn convert_str_to_prost_timestamp(value: Time) -> Result<ProstTimestamp> {
```
SuperFluffy, line 68: Incorrect function name: not converting strings anymore and this can't fail.

```suggestion
fn convert_tendermint_to_prost_timestamp(value: Time) {
```

crates/astria-conductor/src/executor.rs
```diff
@@ -209,7 +203,7 @@ impl<C: ExecutionClient> Executor<C> {
         // get transactions for our namespace
         let Some(txs) = block.rollup_txs.get(&self.namespace) else {
```
SuperFluffy, line 204: Suggestion to avoid the allocation/collection further down (see following commoents for the rest):
```suggestion
        let Some(txs) = block.rollup_txs.remove(&self.namespace) else {
```

crates/astria-conductor/src/executor.rs
```diff
@@ -218,34 +212,17 @@ impl<C: ExecutionClient> Executor<C> {
...
-            height = block.header.height,
+            height = block.header.height.value(),
             parent_block_hash = hex::encode(&prev_block_hash),
             "executing block with given parent block",
         );
 
-        // parse cosmos sequencer transactions into rollup transactions
-        // by converting them to SequencerMsgs and extracting the `data` field
         let txs = txs
             .iter()
```
SuperFluffy, line 221:
```suggestion
            .into_iter()
```

crates/astria-conductor/src/executor.rs
```diff
@@ -218,34 +212,17 @@ impl<C: ExecutionClient> Executor<C> {
...
-                        "ignoring cosmos tx with more than one sequencer message",
-                    );
-                    return None;
-                }
-                let Some(msg) = msgs.first() else {
-                    return None;
-                };
-                Some(msg.data.clone())
-            })
+            .map(|tx| tx.transaction.clone())
```
SuperFluffy, line 222: Together with the suggestions above (using the owning iterator we can move out of `tx` and don't need the hot clone):
```suggestion
            .map(|tx| tx.transaction)
```

crates/astria-sequencer-client/src/lib.rs
```diff
@@ -1,3 +1,29 @@
...
+            Nonce,
+        },
+    },
+    transaction::{
+        Action,
+        Signed as SignedTransaction,
+    },
+};
 pub use client::Client;
+pub use tendermint::block::Height;
```
SuperFluffy, line 18: Why the reexports here and below?

crates/astria-sequencer-relayer/Cargo.toml
```diff
@@ -29,7 +29,10 @@ tokio = { workspace = true, features = [ "macros", "rt-multi-thread" ] }
 astria-gossipnet = { path = "../astria-gossipnet" }
 astria-proto = { path = "../astria-proto" }
 astria-rs-cnc = { path = "../astria-rs-cnc" }
+astria-sequencer-client = { path = "../astria-sequencer-client" }
+
 base64-serde = "0.7.0"
+home = "0.5.5"
```
SuperFluffy, line 35: Don't use this in a service.

crates/astria-sequencer-relayer/src/config/mod.rs
```diff
@@ -84,7 +87,11 @@ impl Default for Config {
             gas_limit: crate::data_availability::DEFAULT_PFD_GAS_LIMIT,
             disable_writing: false,
             block_time: DEFAULT_BLOCK_TIME,
-            validator_key_file: DEFAULT_VALIDATOR_KEY_FILE.into(),
+            validator_key_file: Path::new(&home::home_dir().unwrap())
```
SuperFluffy, line 90: Don't use homedirs in services. Just have the service open the path and fail if it can't find it.

crates/astria-sequencer-relayer/src/types.rs
```diff
@@ -1,200 +1,261 @@
...
+            bail!("block has no data hash");
+        }
+
+        // we unwrap sequencer txs into rollup-specific data here,
+        // and namespace them correspondingly
+        let mut rollup_txs = HashMap::new();
+
+        for (index, tx) in b.data.iter().enumerate() {
+            debug!(
+                "parsing tx: {:?}",
```
SuperFluffy, line 177: don't format events
```suggestion
                index, bytes = general_purpose::STANDARD.encode(tx.as_slice()), "parsing data from tendermint block"

File: crates/astria-sequencer/src/sequence/action.rs [hunk 9]
```diff
@@ -11,8 +11,8 @@ use crate::transaction::action_handler::ActionHandler;
 /// which are bytes to be interpreted by the rollup.
 #[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
 pub struct Action {
-    chain_id: Vec<u8>,
-    data: Vec<u8>,
+    pub chain_id: Vec<u8>,
```
SuperFluffy, line 14: Why make this public

crates/astria-sequencer/src/transaction/signed.rs
```diff
@@ -32,7 +32,7 @@ use crate::{
 pub struct Signed {
     pub(crate) signature: Signature,
     pub(crate) public_key: VerificationKey,
-    pub(crate) transaction: Unsigned,
+    pub transaction: Unsigned,
```
SuperFluffy, line 35: Please don't make this public and instead provide an accessor.

This change violates means that I can violate the invariant noted in the doc comment.

crates/astria-conductor/src/block_verifier.rs
```diff
@@ -88,23 +70,18 @@ impl BlockVerifier {
...
-            .header
-            .height
-            .parse()
-            .wrap_err("failed to parse header height")?;
+        let height = data.data.header.height.value() as u32;
         let validator_set = self
-            .tendermint_client
-            .get_validator_set(height - 1)
+            .sequencer_client
+            .get_validator_set((height - 1).into())
```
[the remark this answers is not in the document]
noot: I think this should actually be the current height - the validator set at the current height is the one before the block is built, thus the proposer should be in that set. the next validator set is the validator set after the block is executed https://github.com/tendermint/tendermint/blob/main/spec/core/data_structures.md#block
will update!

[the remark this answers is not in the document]
noot: removed the - 1

crates/astria-sequencer-relayer/src/types.rs
```diff
@@ -1,200 +1,261 @@
...
-    pub evidence_hash: Option<Base64String>,
-    pub proposer_address: Base64String,
+/// ParsedSequencerBlockData represents a sequencer block's data
+/// to be submitted to the DA layer.
+///
+/// TODO: compression or a better serialization method?
+/// TODO: merkle proofs for each rollup's transactions
+#[derive(Serialize, Deserialize, Clone, Debug, Eq, PartialEq)]
+pub struct ParsedSequencerBlockData {
+    pub block_hash: Base64String,
```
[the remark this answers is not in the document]
noot: agreed, should i do that here or in follow-up?

crates/astria-sequencer-client/src/lib.rs
noot: removed the unused ones

crates/astria-sequencer-relayer/Cargo.toml
noot: i have no idea why but changing the validator key path seems to have made the default file un-findable for me :/ will remove though

SuperFluffy: Yeah, the default place to look for cometbft is a bit iffy. It makes sense if you run this as a user, but it's different for services.

In my opinion we should remove all defaults in these configs and provide all of it through env vars. That will a) cut down on complexity and b) means we have one source of truth (which is nice)

crates/astria-conductor/src/block_verifier.rs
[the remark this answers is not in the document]
SuperFluffy: Heh, good to know 😃

crates/astria-sequencer-relayer/src/types.rs
```diff
@@ -1,200 +1,261 @@
...
 }
 
-#[derive(Clone, Deserialize, Debug, PartialEq, Eq, Serialize)]
-pub struct BlockId {
-    pub hash: Base64String,
-    pub part_set_header: Parts,
+impl fmt::Display for Namespace {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        // FIXME: `hex::encode` does an extra allocation which could be removed
+        f.write_str(&hex::encode(self.0))
```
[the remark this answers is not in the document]
SuperFluffy: We'll probably figure this out when getting rid of Base64String, represent everything as raw bytes, and then refactor the code to decide on the representation in the interfaces to tendermint, Celestia, etc

crates/astria-sequencer/src/sequence/action.rs
SuperFluffy: Not necessarily... But you have to have a good reason to make it all pub.

For example the responses of the json rpc client I made all `pub`, because there aren't really any invariants to uphold (no methods that take them as arguments) and because I didn't care to hide implementation details (it's all Celestia responses anyway).

However, if your `Action` type contains restricted values (because other methods rely on that fact), then you should not allow its construction outside of dedicated methods.

Another reason is not allowing anyone outside your crate to create the type so that the type contents become an implementation details and you don't have to worry about breaking somebody else's code

crates/astria-sequencer-relayer-test/src/lib.rs
```diff
@@ -317,6 +317,7 @@ async fn wait_until_sequencer_is_available(namespace: &str) {
     let client = reqwest::Client::builder()
         .build()
         .expect("building a basic reqwest client should never fail");
+    // TODO
```
joroshiba, line 320: empty todo

crates/astria-sequencer-relayer/src/types.rs
```diff
@@ -1,200 +1,261 @@
...
-    pub last_commit_hash: Option<Base64String>,
-    pub data_hash: Option<Base64String>,
-    pub validators_hash: Base64String,
-    pub next_validators_hash: Base64String,
-    pub consensus_hash: Base64String,
-    pub app_hash: Base64String,
-    pub last_results_hash: Option<Base64String>,
-    pub evidence_hash: Option<Base64String>,
-    pub proposer_address: Base64String,
+/// ParsedSequencerBlockData represents a sequencer block's data
```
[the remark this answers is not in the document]
joroshiba: Maybe omit the `parsed` or `prepared` and go with `SequencerBlockData`?

crates/astria-conductor/Dockerfile
```diff
@@ -34,8 +34,11 @@ RUN cargo chef prepare --recipe-path recipe.json
 FROM chef as builder
 COPY --from=planner /build/recipe.json recipe.json
 
+RUN apt-get update \
+    && apt-get install -y \
+    librocksdb-dev
+    
 ARG TARGETPLATFORM
-ENV ROCKSDB_LIB_DIR=/usr/lib
```
joroshiba, line 38: I believe this was added to improve build times? @SuperFluffy knows context

crates/astria-sequencer-relayer/Dockerfile
```diff
@@ -34,8 +34,11 @@ RUN cargo chef prepare --recipe-path recipe.json
 FROM chef as builder
 COPY --from=planner /build/recipe.json recipe.json
 
+RUN apt-get update \
+    && apt-get install -y \
+    librocksdb-dev
+
 ARG TARGETPLATFORM
-ENV ROCKSDB_LIB_DIR=/usr/lib
```
joroshiba, line 38: same as conductor dockerfile

crates/astria-conductor/Dockerfile
noot: yeah i re-added this and now the image fails to build lol :(( i'll wait for comment

crates/astria-conductor/Dockerfile
```diff
@@ -34,8 +34,11 @@ RUN cargo chef prepare --recipe-path recipe.json
 FROM chef as builder
 COPY --from=planner /build/recipe.json recipe.json
 
+RUN apt-get update \
```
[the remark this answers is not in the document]
noot: removing this breaks the docker build, as it complains librocksdb can't be found :/ maybe #145 can help

crates/astria-conductor-test/src/lib.rs
```diff
@@ -317,6 +317,7 @@ async fn wait_until_sequencer_is_available(namespace: &str) {
     let client = reqwest::Client::builder()
         .build()
         .expect("building a basic reqwest client should never fail");
+    // TODO
```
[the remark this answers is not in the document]
noot: updated, it was to update the cosmos endpoint below (will do in  #148)

crates/astria-conductor/Dockerfile
SuperFluffy: See #149

crates/astria-sequencer-relayer/src/types.rs
```diff
@@ -1,200 +1,281 @@
-use std::convert::TryFrom;
+use std::{
```
noot, line 1: this file is basically sequencer_block.rs, i renamed it since the previous types.rs was deleted

crates/astria-sequencer-relayer/src/types.rs
```diff
@@ -1,200 +1,286 @@
...
-impl Header {
-    pub fn hash(&self) -> eyre::Result<TmHash> {
-        let tm_header = self
-            .to_tendermint_header()
-            .wrap_err("failed converting header to tendermint header")?;
-        Ok(tm_header.hash())
+    /// verify_data_hash verifies that the merkle root of the tree consisting of all the
+    /// transactions in the block matches the block's data hash.
+    ///
+    /// TODO: this breaks with the update to use Retro; need to update for merkle proofs
```
joroshiba, line 231: Can we make an issue for this?

joroshiba (approved):

crates/astria-sequencer-relayer/src/types.rs
noot: opened  #153

crates/astria-sequencer-relayer/src/types.rs
SuperFluffy, line 1: Bikeshedding: Not a fan of having a `types` module. I prefer if types are defined within the respective contexts where they make sense. So `sequencer_block` was not a bad name, imo.

SuperFluffy (approved):
looks good to me. obviously we need more testing, but I am happy to punt this into the future.

On a different note: :fireworks: :rocket: :partying_face: :tada: for getting rid of all the metro<>cosmos<>tendermint stuff.

Issue #15, opened 2023-05-02
conductor: queue for block ordering and execution


noot:
we need to execute blocks in order, without any gaps. however, this is not guaranteed by the current implementation.

we should implement some sort of queue that ensures blocks are executed in the correct sequence, and a block cannot be executed without its parent being the current head.

emhane:
https://github.com/astriaorg/astria/issues/332 makes issue obsolete

joroshiba:
secure websocket does not fully solve issue, we need lower level prevention of out of order execution for exceptional cases.

noot:
closing as duplicate/stale of #475 and #471 and #464

PR #150, opened 2023-07-11
feat: submit and fetch blobs atomically

### LINKED ISSUE #30: submit entire shared sequencer block to celestia in single transaction
Currently we submit a separate PFB (pay-for-bytes) transaction to Celestia for each rollup's transactions rather than a single transaction for the entire shared sequencer block.
The reason for this is that Celestia does not yet support batching data to multiple namespaces in a single transaction. However, this is likely to change in the near future as there is an existing PR to change this.
- [code for submit](https://github.com/celestiaorg/celestia-app/blob/c121c285005c0b5d47542fd955fc5be688905e1a/x/blob/payforblob.go#L17-L54)
- [code for commitment](https://github.com/celestiaorg/celestia-app/blob/c121c285005c0b5d47542fd955fc5be688905e1a/x/blob/types/payforblob.go#L168-L235)

This means a cross-rollup bundle wouldn't be atomic because there would be one Celestia tx for rollup A and one for rollup B. This has several implications/problems:
1. A shared sequencer block can be partially censored by a Celestia node.
2. A shared sequencer block can be spread over several Celestia nodes due to gas limits.
This affects the liveness guarantees that the shared sequencer provides to the rollups built on it and with it potentially the MEV market. Generally, the current situation is not desirable and needs to be updated once the changes are incorporated into Celestia.

When making this change, issue #31 should also be addressed.

SuperFluffy:
This commit makes submission of rollup transactions to the
data availability layer atomic by making use of the
`state.SubmitPayForBlob` celestia JSON RPC.

As part of this work the client periodically polling the sequencer
has been split out of the relayer into its own task. This means
that submitting blocks to celestia no longer blocks querying
the sequencer for new items.

closes #30


crates/astria-conductor/src/executor.rs
```diff
@@ -292,11 +289,14 @@ impl<C: ExecutionClient> Executor<C> {
     ///
     /// This function returns an error if:
     /// - the call to the execution service's FinalizeBlock function fails
-    #[instrument(ret, err, skip_all, fields(execution_block_hash = hex::encode(&execution_block_hash), %sequencer_block_hash))]
+    // TODO: display the sequencer block hash bytes nicely
+    // #[instrument(ret, err, skip_all, fields(execution_block_hash =
```
noot, line 293: remove?

crates/astria-conductor/src/reader.rs
```diff
@@ -150,65 +145,72 @@ impl Reader {
...
-        self.curr_block_height = self.celestia_client.get_latest_height().await?;
+        let curr_block_height = self
+            .celestia_client
+            .get_latest_height()
+            .await
+            .wrap_err("failed getting latest height from celestia")?;
+        ensure!(
+            curr_block_height > self.curr_block_height,
+            "no new celestia height"
+        );
```
noot, line 160: this returns an error? that seems a bit excessive, as this will make a warn log

crates/astria-conductor/src/reader.rs
```diff
@@ -150,65 +145,72 @@ impl Reader {
...
+            height.start = prev_height,
+            height.end = curr_block_height,
+            "checking celestia blocks for range of heights",
         );
 
+        let mut blocks = vec![];
         // check for any new sequencer blocks written from the previous to current block height
-        for height in prev_height..self.curr_block_height {
-            let res = self
+        'check_heights: for height in prev_height..=self.curr_block_height {
```
noot, line 170: what does the `=self.curr_block_height` do?

crates/astria-conductor/src/reader.rs
```diff
@@ -150,65 +145,72 @@ impl Reader {
...
+                        // this means someone submitted an invalid block to celestia;
+                        // we can ignore it
+                        warn!(error.msg = %e, error.cause_chain = ?e, "failed to get sequencer block from namespace data");
+                        continue 'get_sequencer_blocks;
                     }
+                };
+                if let Err(e) = self.block_verifier.validate_sequencer_block(&block).await {
+                    // this means someone submitted an invalid block to celestia;
+                    // we can ignore it
+                    warn!(error.msg = %e, error.cause_chain = ?e, "failed to validating sequencer block");
```
noot, line 210:
```suggestion
                    warn!(error.msg = %e, error.cause_chain = ?e, "failed to validate sequencer block");
```

crates/astria-sequencer-relayer/src/data_availability.rs
```diff
@@ -274,77 +257,43 @@ impl CelestiaClient {
     }
 
     /// submit_block submits a block to Celestia.
```
noot, line 259: update comment

crates/astria-sequencer-relayer/src/relayer.rs
```diff
@@ -97,62 +92,85 @@ impl Relayer {
...
-                "ignoring block: proposer address is not ours",
+    fn convert_block_response_to_sequencer_block_data(
+        &self,
+        res: BlockResponse,
+        current_height: Option<u64>,
+    ) -> eyre::Result<SequencerBlockData> {
+        if let Some(current_height) = current_height {
+            ensure!(
+                res.block.header.height.value() > current_height,
+                "sequencer block response had height below current height tracked in relayer"
```
noot, line 103:
```suggestion
                "sequencer block response had height below or equal to current height tracked in relayer"
```

crates/astria-sequencer-relayer/src/relayer.rs
```diff
@@ -97,62 +92,85 @@ impl Relayer {
...
+        let mut all_converted_blocks = Vec::with_capacity(block_responses.len());
+        for res in block_responses {
+            match self.convert_block_response_to_sequencer_block_data(
+                res,
+                new_state.current_sequencer_height,
+            ) {
+                Ok(converted) => {
+                    if let Err(error) = self.block_tx.send(converted.clone()) {
+                        warn!(?error, "failed sending sequencer block data to gossip task");
+                    }
```
noot, line 127: this will still block on writing to celestia right? this needs to be separated out so that it's gosisped as soon as a new block is received

crates/astria-sequencer-relayer/src/relayer.rs
```diff
@@ -97,62 +92,85 @@ impl Relayer {
...
+                    warn!(?error, "dropping block response");
+                }
             }
+        }
+        // get the max sequencer height from the valid blocks; the result of this op is `None` if
+        // the vector is empty.
+        let Some(max_sequencer_height) = all_converted_blocks
+            .iter()
+            .map(|block| block.header.height.value())
+            .max()
```
noot, line 141: this, combined with the `get_latest_block` for polling means blocks can potentially be missed if the polling time isn't fast enough.

crates/astria-sequencer-relayer/src/sequencer_poller.rs
```diff
@@ -0,0 +1,54 @@
...
+            poll_period: Duration::from_millis(cfg.block_time),
+            sequencer_blocks_tx,
+        })
+    }
+
+    pub(crate) async fn run(self) -> eyre::Result<()> {
+        let mut interval = interval(self.poll_period);
+        loop {
+            interval.tick().await;
+            match self.client.get_latest_block().await {
```
noot, line 39: we should probably change this not to be `get_latest_block` but to be "get all blocks since last height received" as blocks can potentially be missed

crates/astria-sequencer-relayer/src/sequencer_poller.rs
```diff
@@ -0,0 +1,54 @@
...
+                Ok(block) => self
+                    .sequencer_blocks_tx
+                    .send(block)
+                    .wrap_err("channel closed unexpectedtly")?,
+                Err(e) => {
+                    warn!(error.msg = %e, error.cause_chain = ?e, "failed getting latest block from sequencer")
+                }
+            }
+        }
+        // Return Ok to make the types align (see the method's doc comment why this is necessary).
```
noot, line 49: where is the doc comment? haha

crates/astria-conductor/src/reader.rs
SuperFluffy: You are right, this would be very noisy but actually normal behavior, so `warn!` is too heavy-handed. I was looking for a way to early return. Will change.

crates/astria-conductor/src/reader.rs
SuperFluffy: `..=` is to construct an inclusive range, i.e. `[x, y]` rathern than `[x, y)` as for `..`.

The old logic was to check `[old_current_block, new_current_block)` - so we never actually fetched the most up-to-date block.

On that note I just realize that I should be checking `prev_height + 1 ..= curr_block_height`

crates/astria-sequencer-relayer/src/relayer.rs
```diff
@@ -97,62 +92,85 @@ impl Relayer {
...
+            ensure!(
+                res.block.header.height.value() > current_height,
+                "sequencer block response had height below current height tracked in relayer"
             );
-            return Ok(new_state);
         }
+        ensure!(
+            res.block.header.proposer_address == self.validator.address,
+            "proposer recorded in sequencer block does not match internal validator"
+        );
```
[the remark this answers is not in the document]
SuperFluffy: ACK, this will be very noisy. I was reaching for an early return. I will change the logic to log this at the debug level and not overuse error handling.

crates/astria-sequencer-relayer/src/relayer.rs
SuperFluffy: Yeah, you are right. This PR splits reading-from-sequencer and submitting-blocks (which is writing-to-celestia + gossiping-blocks).

The moment we start draining the channel and processing each block we start gossiping, but once we submit to celestia gossiping is blocked again.

I will separate that.

crates/astria-sequencer-relayer/src/relayer.rs
SuperFluffy: Wouldn't we always miss blocks if we don't poll fast enough? Looking at the sequencer client there is I think no way to get a block at a particular height.

IMO we should probably get rid of the http client in favor of the websocket one and subscribe to [`EventType::NewBlock`](https://docs.rs/tendermint-rpc/0.32.0/tendermint_rpc/query/enum.EventType.html#variant.NewBlock)

There is still the danger of missing events, but at least we don't have to keep track internally of what the latest block was.

Hm, I am not sure if tower-abci supports websockets tho. :(

crates/astria-sequencer-relayer/src/relayer.rs
SuperFluffy: I have addressed this in a new commit. The following tasks are now split split such that they will not block one another:

1. receipt of new blocks from the poller (spawn new blocking task to convert to sequencer block data)
2. conversion of block responses to sequencer block data (runs in blocking tasks on a joinset)
3. queuing + gossip each block data
4. submitting the queued block data to celestia in a separate task.

crates/astria-sequencer-relayer/src/relayer.rs
noot: there is a function to get block by height: https://github.com/astriaorg/astria/blob/main/crates/astria-sequencer-client/src/client.rs#L71

yeah i agree, pushing would be preferred over polling

crates/astria-sequencer-relayer/src/data_availability.rs
```diff
@@ -355,128 +305,143 @@ impl CelestiaClient {
...
             })
             .collect::<Vec<_>>();
         Ok(sequencer_namespace_datas)
     }
 
-    /// get_sequencer_block returns the full SequencerBlock (with all rollup data) for the given
-    /// SequencerNamespaceData.
-    pub async fn get_sequencer_block(
+    /// get_sequencer_block returns the full SequencerBlock (with all rollup data) for the
+    pub async fn get_all_rollup_data_from_sequencer_namespace_data(
```
joroshiba, line 325: Am I correct in my reading that if there are N blocks per Celestia block, this function will be called N times?Which will result in many calls to celestia to retrieve the same data and filter it out? The number of calls to celestia grows by a factor of O(N)?

Can we make this be O(1) by not filtering out and throwing away the data we will later refetch and process?

crates/astria-sequencer-relayer/src/data_availability.rs
```diff
@@ -355,128 +305,143 @@ impl CelestiaClient {
...
+        let namespace_ids = namespace_data
+            .data
+            .rollup_namespaces
+            .iter()
+            .map(|ns| **ns)
+            .collect();
         let req = GetAllRequest {
             height,
-            namespace_ids: vec![*rollup_namespace],
+            namespace_ids,
```
joroshiba, line 341: It seems like we are grabbing the data for all namespaces instead of only for a given rollup?

If so that seems to negate the purpose of storing it in different namespaces anyways? Is this for validation purposes and to be fixed @noot?

noot: yeah right now we grab everything, just because the merkle proofs for fetching only the relevant namespace wasn't implemented yet. we can change this to just grab the relevant namespace and do verification in a separate issue, but out of scope for this PR

crates/astria-sequencer-relayer/src/data_availability.rs
joroshiba: I think this is also out of scope for this, like makes sense to fix alongside grabbing only local namespaced data.

crates/astria-conductor/src/executor.rs
SuperFluffy: Removed. All of this is covered in #165.

crates/astria-celestia-jsonrpc-client/Cargo.toml
```diff
@@ -9,10 +9,14 @@ edition = "2021"
...
 base64-serde = "0.7.0"
 http = "0.2.9"
-jsonrpsee = { version = "0.18.2", features = ["macros", "client"] }
+
+jsonrpsee = { workspace = true, features = ["macros", "client"] }
 serde = { workspace = true, features = ["derive"] }
 serde_json = { workspace = true, features = ["raw_value"] }
 
+[features]
+server = ["jsonrpsee/server"]
```
noot, line 18: what's the server feature for? i guess for testing?

SuperFluffy: Yeah, see the `dev-dependencies` entry for sequencer-relayer which uses it to mock a server

crates/astria-conductor/src/reader.rs
noot: reminder on this

crates/astria-sequencer-relayer/src/api.rs
```diff
@@ -103,17 +187,19 @@ impl IntoResponse for Readyz {
...
-        Self {
-            current_sequencer_height: relayer_status.current_sequencer_height,
-            current_data_availability_height: relayer_status.current_data_availability_height,
-        }
-    }
+async fn get_number_of_subscribers(tx: UnboundedSender<network::InfoQuery>) -> eyre::Result<u64> {
+    use eyre::Report;
+    use network::InfoQuery;
+    let (info_response_tx, info_response_rx) = oneshot::channel();
+    tx.send(InfoQuery::NumberOfPeers(info_response_tx))
```
noot, line 194: wow, this seems complex haha, it makes sense though. is this a common method for doing request/responses between tasks?

crates/astria-sequencer-relayer/src/data_availability.rs
noot: does reading from celestia have any cost? i was under the impression it didn't (other than processing time ofc)

noot: ideally this will only be called once, passing in the `namespace_data` which has already been verified as being the correct data. also this should probably be moved to conductor (or some shared crate if this is used in relayer tests?)

crates/astria-sequencer-relayer/src/data_availability.rs
noot: #175

crates/astria-sequencer-relayer/src/api.rs
SuperFluffy: I've seen and done it before when doing computations on a rayon threadpool triggered by requests that came in through the async (tokio) runtime. So sending the response channel to rayon was a way to show that it's computation is complete.

I was about to reach for permanent channels (one for sending requests to gossipnet, one to receive them from gossipnet). But how do you match which response was for which request? So then you have to make an envelope or add a tag or w/e, and store that in a map....

At first glance it feels complex because sending a channel back and forth is unusual, but the solution is much simpler.

crates/astria-sequencer-relayer/src/data_availability.rs
```diff
@@ -352,128 +309,143 @@ impl CelestiaClient {
...
-                }
-            })
-            .filter(|d| d.data.block_hash == block_hash);
+            .collect();
+        // retain rollup datas whose block hash matches the block hash of the namespaced data
+        rollup_datas
+            .retain(|_, rollup_data| namespace_data.data.block_hash == rollup_data.data.block_hash);
+
+        // retain rollup datas that can be verified
+        rollup_datas.retain(|namespace, rollup_data| {
```
noot, line 373: maybe add a `retain` before this that checks the public key which signed the rollup data is the same as `verification_key`? might save on some serialization/hashing

crates/astria-sequencer-relayer/src/data_availability.rs
SuperFluffy: This here exists only for conductor I think. It should be refactored and moved out of the relayer so that conductor doesn't import the whole thing.

Re cost: you mean in fees?

crates/astria-sequencer-relayer/src/data_availability.rs
```diff
@@ -352,128 +309,143 @@ impl CelestiaClient {
...
         };
-
-        // there should NOT be multiple datas with the same block hash and signer
-        if rollup_datas.next().is_some() {
-            bail!("multiple rollup datas with the same block hash and signer");
-        }
-
-        Ok(Some(rollup_data_for_block.data.rollup_txs))
+        let data = rollup_namespace_data
+            .to_signed(signing_key, verification_key)
```
noot, line 425: maybe out of scope, but why does this need the verification key? seems unnecessary (it can be derived from the signing key) is it more so we don't have to re-derive it?

crates/astria-sequencer-relayer/src/relayer.rs
```diff
@@ -1,50 +1,65 @@
...
-    block_tx: UnboundedSender<SequencerBlockData>,
+
+    data_availability: Option<CelestiaClient>,
+    validator: Validator,
+    gossip_block_tx: UnboundedSender<SequencerBlockData>,
     state_tx: watch::Sender<State>,
+    queued_blocks: Vec<SequencerBlockData>,
+    conversion_workers: task::JoinSet<eyre::Result<Option<SequencerBlockData>>>,
+    submission_task: Option<task::JoinHandle<eyre::Result<u64>>>,
+    sequencer_task: Option<task::JoinHandle<Result<block::Response, tendermint_rpc::Error>>>,
```
noot, line 49: can you add comments for what these are all for?

crates/astria-sequencer-relayer/src/data_availability.rs
SuperFluffy: Makes sense to change it. But I think I was just calling the provided method without thinking about it.

Make a follow-up issue?

crates/astria-sequencer-relayer/src/relayer.rs
```diff
@@ -84,75 +98,244 @@ impl Relayer {
...
+            Ok(block) => {
+                let current_height = self.state_tx.borrow().current_sequencer_height;
+                let validator = self.validator.clone();
+                // Start the costly conversion; note that the current height at
+                // the time of receipt matters. The internal state might have advanced
+                // past the height recorded in the block while it was converting, but
+                // that's ok.
+                self.conversion_workers.spawn_blocking(move || {
+                    convert_block_response_to_sequencer_block_data(block, current_height, validator)
+                });
```
noot, line 156: why is this costly?

crates/astria-sequencer-relayer/src/relayer.rs
```diff
@@ -161,21 +344,129 @@ impl Relayer {
...
+    #[instrument(name = "Relayer::run", skip_all)]
+    pub(crate) async fn run(mut self) -> eyre::Result<()> {
+        let wait_for_da = self.wait_for_data_availability_layer(5, Duration::from_secs(5), 2.0);
+        let wait_for_seq = self.wait_for_sequencer(5, Duration::from_secs(5), 2.0);
+        match tokio::try_join!(wait_for_da, wait_for_seq) {
+            Ok(((), ())) => {}
+            Err(err) => return Err(err).wrap_err("failed to start relayer"),
+        }
+        // .await
+        // .wrap_err("failed establishing connection to data availability layer")?;
```
noot, line 356: delete

crates/astria-sequencer-relayer/src/relayer.rs
```diff
@@ -84,75 +98,244 @@ impl Relayer {
...
                 }
-                Err(e) => warn!(error = ?e, "failed to submit block to DA layer"),
+                // Update the internal state if the block was admitted
+                let height = sequencer_block_data.header.height.value();
+                self.state_tx.send_if_modified(|state| {
+                    if Some(height) > state.current_sequencer_height {
+                        state.current_sequencer_height = Some(height);
+                        return true;
+                    }
+                    false
```
noot, line 197: is there ever a case where the height wouldn't be higher than the current? it seems guaranteed by `convert_block_response_to_sequencer_block_data`?

crates/astria-sequencer-relayer/src/relayer.rs
```diff
@@ -161,21 +344,129 @@ impl Relayer {
...
+        let stop_msg = loop {
+            select!(
+                // Query sequencer for the latest block if no task is in flight
+                _ = sequencer_interval.tick() => self.handle_sequencer_tick(),
+
+                // Handle the sequencer response by converting it
+                //
+                // NOTE: + wrapping the task in an async block makes this lazy;
+                //       + `unwrap`ping can't fail because this branch is disabled if `None`
+                res = async { self.sequencer_task.as_mut().unwrap().await }, if self.sequencer_task.is_some() => {
```
noot, line 375: what does the `async { self.sequencer_task.as_mut().unwrap().await }, if self.sequencer_task.is_some()` syntax do? does it only execute the future if `self.sequencer_task.is_some()`?

crates/astria-sequencer-relayer/src/relayer.rs
```diff
@@ -161,21 +344,129 @@ impl Relayer {
...
+                && !self.queued_blocks.is_empty()
+                && self.submission_task.is_none()
+            {
+                let client = self.data_availability.clone().expect(
+                    "this should not fail because the if condition of this block checked that a \
+                     client is present",
+                );
+                self.submission_task = Some(task::spawn(submit_blocks_to_data_availability_layer(
+                    client,
+                    self.queued_blocks.clone(),
```
noot, line 411: this seems pricey - can we just take all the values out of the vec or something?

crates/astria-sequencer-relayer/src/relayer.rs
SuperFluffy: You have theoretically unbounded computation inside this. In a loop you: decode protobuf, (re)construct ed25519 keypairs, hash your data (including serialization), verify signatures, and do a few more things. So in that sense this conversion is CPU bound compared to most of the other execution here which is i/o bound.

The rule of thumb is that around 10ms in terms of waiting between await points.

This is more educated guessing then precise science. I guess it would be good to make actual benchmarks for the conversion method.

crates/astria-sequencer-relayer/src/relayer.rs
```diff
@@ -84,75 +98,244 @@ impl Relayer {
...
+    #[instrument(skip_all)]
+    fn handle_sequencer_tick(&mut self) {
+        use tendermint_rpc::Client as _;
+        if self.sequencer_task.is_none() {
+            let client = self.sequencer.clone();
+            self.sequencer_task = Some(tokio::spawn(async move { client.latest_block().await }));
+        } else {
+            debug!("task polling sequencer is currently in flight; not scheduling a new task");
         }
+    }
```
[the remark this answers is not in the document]
SuperFluffy: Not putting it in its own task blocks the entire `loop { select! }` block until the next answer. If you are unlucky with the bias selection of the `select!` macro you will never fire the other branches.

With these `select!` statements you should always be acutely aware to not block the loop in the branch handlers.

crates/astria-sequencer/src/transaction/action.rs
```diff
@@ -28,6 +28,27 @@ pub enum Action {
...
+    #[must_use]
+    pub fn as_sequence(&self) -> Option<&sequence::Action> {
+        match self {
+            Self::SequenceAction(a) => Some(a),
+            _ => None,
+        }
+    }
+
+    #[must_use]
+    pub fn as_transfer(&self) -> Option<&Transfer> {
```
noot, line 40: is this used anywhere?

crates/astria-sequencer-relayer/src/relayer.rs
SuperFluffy: Task conversion happens in a joinset. Albeit unlikely, you are not guaranteed that the conversions finish in the order of their heights. For example, if 101 is very big and takes a long time to convert, but block 102 is very small and conversion finishes immediately, then you don't want to update the internal state after 101 is done converting.

On that note: what happens if we bundle sequencer blocks 100 and 102 and submit them to celestia, but 101 only finishes after. Is it ok to submit it out of order? Or should we ensure that there is never a gap?

crates/astria-sequencer-relayer/src/relayer.rs
```diff
@@ -161,21 +344,129 @@ impl Relayer {
...
+        // .await
+        // .wrap_err("failed establishing connection to data availability layer")?;
+        self.wait_for_sequencer(5, Duration::from_secs(5), 2.0)
+            .await
+            .wrap_err("failed establishing connection to the sequencer")?;
+
+        let mut sequencer_interval = interval(self.sequencer_poll_period);
+        // the first tick resolves immediately, so do it now because we don't
+        // want to poll right after start.
+        sequencer_interval.tick().await;
```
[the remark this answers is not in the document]
SuperFluffy: Huh, I did this because I was fixing the (mocked) gossip node subscribing to the "blocks" topic too late (so that sequencer-relayer did not have anywhere to send the data).

Now that relayer is waiting for both celestia + at least 1 subscriber to be present this might not be necessary anymore.

Thanks for raising this! I will have to look into it.

crates/astria-sequencer-relayer/src/relayer.rs
SuperFluffy: Yes, the `, if <condition>` syntax only conditionally enables this `select!` branch. Moving the `.as_mut().unwrap()` into an ad-hoc async block ensure that it is only unwrapped lazily (otherwise this might panic).

crates/astria-sequencer-relayer/src/relayer.rs
noot: i think submitting out of order is ok - the conductor should be able to handle it eventually (see #15)

noot (approved):
approving, please fix outstanding comments before merging! nice work on the test setup :D

crates/astria-conductor/src/reader.rs
SuperFluffy: Changed the early return to `Ok(None)` to address this.

While doing that noticed the conductor kills its event loop if processing failed. Fixed that.

crates/astria-sequencer-relayer/src/relayer.rs
joroshiba: I believe that right now submitting out of order does not work. 

This should be logic we have in conductor, but could break functionality until then.

crates/astria-sequencer-relayer/src/sequencer_poller.rs
SuperFluffy: Removed that entire module

crates/astria-conductor/src/reader.rs
```diff
@@ -150,65 +145,72 @@ impl Reader {
...
+                    .celestia_client
+                    .get_all_rollup_data_from_sequencer_namespace_data(height, &data)
+                    .await
+                    .wrap_err("failed to get rollup data")
+                {
+                    Ok(block) => block,
+                    Err(e) => {
+                        // this means someone submitted an invalid block to celestia;
+                        // we can ignore it
+                        warn!(error.msg = %e, error.cause_chain = ?e, "failed to get sequencer block from namespace data");
```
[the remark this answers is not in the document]
SuperFluffy: The DA client and logging have to be re engineered. Pushing to later https://github.com/astriaorg/astria/issues/179

crates/astria-sequencer-relayer/src/data_availability.rs
```diff
@@ -274,77 +258,43 @@ impl CelestiaClient {
...
-        // TODO: This could probably now be combined into one submission?
-        for (namespace, txs) in block.rollup_txs {
-            debug!(
-                "submitting rollup namespace data for namespace {}",
-                namespace
-            );
-            let rollup_namespace_data = RollupNamespaceData {
-                block_hash: block.block_hash.clone(),
-                rollup_txs: txs,
+        let num_expected_blobs = blocks.iter().map(|block| block.rollup_txs.len() + 1).sum();
```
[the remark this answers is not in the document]
SuperFluffy: https://github.com/astriaorg/astria/issues/180

crates/astria-sequencer-relayer/src/data_availability.rs
SuperFluffy: https://github.com/astriaorg/astria/issues/175

crates/astria-sequencer-relayer/src/relayer.rs
SuperFluffy: https://github.com/astriaorg/astria/issues/181

crates/astria-sequencer-relayer/src/data_availability.rs
SuperFluffy: I am leaving this to a followup PR because I think this might require some rethinking of our types. It's all a bit confusing right now.

crates/astria-sequencer-relayer/src/data_availability.rs
SuperFluffy: It's even worse then that - the verification key is literally part of the signingkey here: https://docs.rs/ed25519-consensus/latest/src/ed25519_consensus/signing_key.rs.html#16-21

The verification key is no longer passed around.

crates/astria-sequencer-relayer/src/relayer.rs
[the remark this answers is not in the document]
SuperFluffy: It turns out this was indeed no longer necessary.

joroshiba (approved):

PR #154, opened 2023-07-11
install ca-certificates in Dockerfile

steezeburger:
Got the following error when running the latest sequencer image. 
```
thread 'main' panicked at 'no CA certificates found', /usr/local/cargo/registry/src/index.crates.io-6f17d22bba15001f/hyper-rustls-0.24.0/src/config.rs:48:9
```

The Astria images need ca certificates. Let's install them in the reusable Dockerfile.

Part of https://github.com/astriaorg/dev-cluster/issues/38


SuperFluffy (approved):
I don't know what the best practices are: install the ca certs into the final layer or just copy them over like is done here?

Either way, I'm sure this is fine.

Please add the error emitted by the binaries to the PR descriptions.

noot (approved):

steezeburger:
> I don't know what the best practices are: install the ca certs into the final layer or just copy them over like is done here?
> 
> Either way, I'm sure this is fine.
> 
> Please add the error emitted by the binaries to the PR descriptions.

RE: best practices: I think it makes sense to copy over the binaries at the end. Otherwise, we would have to run `apt update` in the last layer again, which isn't necessary if we install in the chef layer where an `apt update` is already performed.

steezeburger:
This has become part of https://github.com/astriaorg/astria/pull/158

SuperFluffy:
> This has become part of #158

I commented in the other PR. I would prefer to merge the docker changes and the actual code changes in separate PRs.

PR #155, opened 2023-07-11
feat(gossipnet): implement reading libp2p private key from file

### LINKED ISSUE #108: implement reading/writing libp2p private key from disk
will allow for having persistent peer IDs across restarts.

see https://github.com/astriaorg/astria/blob/main/crates/astria-gossipnet/src/network.rs#L107

noot:
- implement reading libp2p private key from file
- update cli for relayer and conductor respectively

closes #108


joroshiba (approved):
Is there anything special we need to do to get the bootnode information? If so can we document?

noot:
@joroshiba yeah we have to generate keys and put them in a file. I can add some comments in the test

crates/astria-gossipnet/src/network.rs
```diff
@@ -95,10 +95,24 @@ impl NetworkBuilder {
         self
     }
 
+    /// Specify the keypair to use from a file.
+    /// The file should contain a hex-encoded 32-byte ed25519 secret key.
+    pub fn keypair_from_file<P: AsRef<std::path::Path>>(mut self, path: P) -> Result<Self> {
+        let key_string = std::fs::read_to_string(path).wrap_err("failed to read keypair file")?;
+        if key_string.len() < 64 {
```
SuperFluffy, line 102: Is the extra check here necessary? Wouldn't this fail if just calling `hex::decode(&key_string)` and then `Keepair::ed25519_from_bytes(bytes)`?

If you decide to keep it, prefer `color_eyre::eyre::ensure!`:
```suggestion
        ensure!(
            key_string.len() >= 64,
            "keypair file is too short"
        );
```

crates/astria-gossipnet/tests/network.rs
```diff
@@ -16,6 +16,20 @@ use tokio::{
 
 const TEST_TOPIC: &str = "test";
 
+#[tokio::test]
+async fn keypair_from_file() {
+    let key_string = "241ccaff3cc8681385b2cf92a82b49ab2e9c6410f7b6191322867c4299de42d8";
+    let path = std::env::temp_dir().join("test.key");
```
SuperFluffy, line 22: `std::env::temp_dir()` returns `/tmp`. In tests it's better to use [`tempfile::tempfile](`https://docs.rs/tempfile/latest/tempfile/fn.tempfile.html) which ensures that the file gets deleted by the OS after the file handle is closed (make sure to not drop the handle until after the test is done).

crates/astria-sequencer-relayer/src/config/mod.rs
```diff
@@ -186,6 +192,8 @@ astria-sequencer-relayer
                 validator_key_file: "/env/key".into(),
                 rpc_port: 5555,
                 p2p_port: 5555,
+                bootnodes: Vec::new(),
+                libp2p_private_key: None,
```
SuperFluffy, line 196:
```suggestion
                libp2p_private_key: Some("envprivatekey".into()),
```

crates/astria-sequencer-relayer/src/config/mod.rs
```diff
@@ -163,6 +167,8 @@ astria-sequencer-relayer
                 validator_key_file: "/cli/key".into(),
                 rpc_port: 9999,
                 p2p_port: 9999,
+                bootnodes: Vec::new(),
+                libp2p_private_key: None,
```
SuperFluffy, line 171:
```suggestion
                libp2p_private_key: Some("cliprivatekey"),
```

SuperFluffy (approved):
The general changes look good to me.

Please don't punt on the config tests (setting the private key path). I know we should get rid of all that in favor of only env vars. But while we have all of that it's good to keep the tests around.

I couldn't add review comments, but the `ALL_CLI_ARGS` const and the `fn set_all_env` helper function need to be updated to make that work.

crates/astria-gossipnet/src/network.rs
noot: i wanted to pass exactly 64 bytes into `hex::decode` because sometimes people have a newline/space at the end of the file which causes hex::decode to fail if that's written in

PR #159, opened 2023-07-13
fix(sequencer): don't use length-delimited proto serialization

noot:
golang doesn't have support for this :( required for https://github.com/astriaorg/go-ethereum/pull/24


SuperFluffy (approved):
I realize I never thought about that and just assumed protobuf is simply never length delimited.

Of course you can just add a delimiter.

Super interesting. Reading the prost code now.

PR #167, opened 2023-07-17
container: install dynamic libs and fix entrypoint

SuperFluffy:
This commit installs rocksdb in the final layer
of the generic containerfile because all binaries
are currently linked to librocksdb.

In addition, as it turns out build args are not expanded in `ENTRYPOINT` lines. To get a stable path that can be used as an argument to `ENTRYPOINT`, this commit crates a softlink to `/usr/bin/local/$TARGETBINARY` and uses that instead.

### DIFF
--- a/containerfiles/Dockerfile
+++ b/containerfiles/Dockerfile
@@ -111,5 +111,5 @@ RUN if [ "$TARGETPLATFORM" = "linux/arm64" ]; then \
     apt-get clean; \
     rm -rf /var/lib/apt/lists/*;
 COPY --from=builder /build/target/release/$TARGETBINARY /usr/local/bin/$TARGETBINARY
-RUN ln -s $TARGETBINARY /usr/local/bin/entrypoint
+RUN ln -s /usr/local/bin/$TARGETBINARY /usr/local/bin/entrypoint
 ENTRYPOINT ["/usr/local/bin/entrypoint"]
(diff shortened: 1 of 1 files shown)


containerfiles/Dockerfile
```diff
@@ -92,8 +92,24 @@ RUN if [ "$TARGETPLATFORM" = "linux/arm64" ]; then ARCH=aarch64; \
...
+      fi; \
+      apt update; \
+      apt install -y librocksdb-dev:amd64; \
+    fi; \
+    apt install -y wget ca-certificates; \
+    apt-get clean; \
+    rm -rf /var/lib/apt/lists/*;
 COPY --from=builder /build/target/release/$TARGETBINARY /usr/local/bin/$TARGETBINARY
-ENTRYPOINT ["/usr/local/bin/$TARGETBINARY"]
+RUN ln -s $TARGETBINARY /usr/local/bin/entrypoint
```
joroshiba, line 114: Should this be `/usr/local/bin/$TARGETBINARY`

SuperFluffy: eh, you are right

SuperFluffy: Note that that still works because it's looking in its relative dir.

I am still changing it since it's cleaner with an absolute path.

PR #171, opened 2023-07-17
add "pretty" logs to relayer

### LINKED ISSUE #128: Sequencer: Log contents of block production

noot:
logs look like this: 
```
2023-07-17T21:36:49.080710Z  INFO astria_sequencer_relayer::relayer: got block from sequencer height=177 tx_count=1
2023-07-17T21:36:49.081687Z  INFO astria_sequencer_relayer::relayer: submitting sequencer block to DA layer sequencer_block=177 proposer=account::Id(78AF5A8468EF8C1B52AFF27AD689FEAB10E4F5E3) namespaces_to_tx_count={61737472696173657175: 1}
2023-07-17T21:36:55.420648Z  INFO astria_sequencer_relayer::relayer: submitted sequencer block to DA layer sequencer_block=177 da_layer_block=18102 namespace_count=1
```

thoughts @joroshiba @jbowen93 ? should i merge the first 2 logs? 

closes  #128


crates/astria-sequencer-relayer/src/types.rs
```diff
@@ -75,6 +75,13 @@ impl Namespace {
     }
 }
 
+impl fmt::Debug for Namespace {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        // FIXME: `hex::encode` does an extra allocation which could be removed
+        f.write_str(&hex::encode(self.0))
```
SuperFluffy, line 81: This is breaking with rust convention and should not be done.

Also, this is the same as the `Display for Namespace` impl. Why not use that?

crates/astria-sequencer-relayer/src/relayer.rs
```diff
@@ -127,6 +127,18 @@ impl Relayer {
...
 
+        let mut namespaces_to_tx_count = std::collections::HashMap::new();
+        sequencer_block.rollup_txs.iter().for_each(|(ns, txs)| {
+            namespaces_to_tx_count.insert(*ns, txs.len());
+        });
+
+        info!(
+            sequencer_block = height,
+            proposer = ?sequencer_block.header.proposer_address,
+            namespaces_to_tx_count = ?namespaces_to_tx_count,
```
SuperFluffy, line 138: Using rust debug for tracing fields always ends up being messy.

Alternative suggestion:

1. create the map as above, but write the result of `serde_json::to_string` into the field;
2. write a serialization wrapper and serialize the blocks nicely directly using a `Serialize` impl. You can Almost directly use the example for `MyMap` here: https://serde.rs/impl-serialize.html#serializing-a-sequence-or-map

IMO the second option would be cleanest and avoid the extra allocation on every submission (which however would probably not be so bad).

crates/astria-sequencer-relayer/src/types.rs
noot: the hashmap requires debug, which causes the namespaces to be printed not as hex strings

SuperFluffy:
i made an example for my suggestions above: https://github.com/astriaorg/astria/tree/superfluffy/submissions_logging

noot:
@SuperFluffy feel free to push to here, was working on it but you beat me to it lol

crates/astria-sequencer-relayer/src/types.rs
SuperFluffy: I understand - this is fundamentally about tracing not being able to easily generate fields for  more complex structures. This should all be fixed with the valuable trait/crate. But it's not really moving anywhere.

For the time being you can work around it with either what I have done in https://github.com/astriaorg/astria/tree/superfluffy/submissions_logging (providing a wrapper to skip building the hashmap), or/in addition creating something like a `struct DisplayJson<a, T>(T)` and `impl<'a, T> Display for DisplayJson<'a, T> where T: Serialize` or something like that.

We should probably do something like the latter at some point because this issue will start popping up everywhere in the codebase at some point.

https://github.com/tokio-rs/tracing/discussions/1906

crates/astria-sequencer/src/app.rs
```diff
@@ -139,7 +139,12 @@ impl App {
             .await
             // TODO: explain why this is an invariant of the system
             .expect("block height should be set");
-        info!(?tx, ?height, "executed transaction");
+        info!(
+            ?tx,
+            ?height,
+            sender = tx.signer_address().to_string(),
```
SuperFluffy, line 145: No need for the extra to_string as it already impls Display:
```suggestion
            sender = %tx.signer_adress(),
```

crates/astria-sequencer-relayer/src/serde.rs
```diff
@@ -1,4 +1,42 @@
...
+        for (ns, txs) in self.0 {
+            map.serialize_entry(&ns, &txs.len())?;
+        }
+        map.end()
+    }
+}
+
+impl<'a> std::fmt::Display for NamespaceToTxCount<'a> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        // Note that this technically throws away the error, but that's
```
SuperFluffy, line 36: Please delete this before merging. It's absolutely wrong and I was too tired last night when I wrote that.

The actual comment should be:

This cannot fail because we are only serializing into a string (unless the system is OOM).

crates/astria-sequencer/src/app.rs
```diff
@@ -139,7 +139,12 @@ impl App {
             .await
             // TODO: explain why this is an invariant of the system
             .expect("block height should be set");
-        info!(?tx, ?height, "executed transaction");
+        info!(
+            ?tx,
+            ?height,
```
SuperFluffy, line 144: It's u64 I think, so don't add the `?`

crates/astria-sequencer/src/app.rs
```diff
@@ -139,7 +139,12 @@ impl App {
             .await
             // TODO: explain why this is an invariant of the system
             .expect("block height should be set");
-        info!(?tx, ?height, "executed transaction");
+        info!(
+            ?tx,
```
SuperFluffy, line 143: It's ok to leave this with `?`. But can you make a follow up issue that we replace this with a json wrapper?

SuperFluffy (approved):

joroshiba (approved):

PR #174, opened 2023-07-17
feat(sequencer): implement queries for previous heights

### LINKED ISSUE #130: sequencer: support queries for block numbers that are not latest
currently, the sequencer only supports queries on the latest state. however, it should handle queries for past states correctly.

noot:
- implement storage queries for heights that are non-latest
- note: latest height is defined as 0 for queries; this is what abci-cli passes in if height is unspecified (ie. latest is assumed)
- add block height to storage version map; this was needed because block height doesn't necessarily correspond to storage version. since we're using temp storage, the versions change on every restart (they get reset), and if we use persistent storage, height and storage version also may not line up (for example, in the case of blocks that are processed, but don't end up in the canonical chain)
- this map is not stored as part of the consensus state as it's irrelevant to that, and the height->version entries might actually change upon restart (since we are using temp storage)

closes #130


crates/astria-sequencer/src/state_ext.rs
```diff
@@ -41,6 +41,22 @@ pub(crate) trait StateReadExt: StateRead {
         Time::parse_from_rfc3339(&String::from_utf8_lossy(&bytes))
             .context("failed to parse timestamp from raw timestamp bytes")
     }
+
+    #[instrument(skip(self))]
+    async fn get_storage_version_by_height(&self, height: u64) -> Result<u64> {
+        let key: Vec<u8> = format!("storage_version/{height}").into();
+        let Some(bytes) = self
+            .nonconsensus_get_raw(key.as_slice())
```
SuperFluffy, line 49: 2 notes:

1. I tried to find this function in penumbra and it turns out they have already renamed it to [`nonverifiable_get_raw`](https://github.com/penumbra-zone/penumbra/commit/4eab19bc8defa1c98695e01d0b7a9f1f10b610a6). Should we bump our dependency?
2. Deref coercion should work here:
```suggestion
            .nonconsensus_get_raw(&key)
```

crates/astria-sequencer/src/state_ext.rs
```diff
@@ -41,6 +41,22 @@ pub(crate) trait StateReadExt: StateRead {
         Time::parse_from_rfc3339(&String::from_utf8_lossy(&bytes))
             .context("failed to parse timestamp from raw timestamp bytes")
     }
+
+    #[instrument(skip(self))]
+    async fn get_storage_version_by_height(&self, height: u64) -> Result<u64> {
+        let key: Vec<u8> = format!("storage_version/{height}").into();
```
SuperFluffy, line 47: Does it make sense to have a function for key construction to defend against accidental refactors/mismatches between this and `put_storage_version_by_height` below?

```rust
fn make_storage_version_by_height_key(height: u64) -> String {
    format!("storage_version/{height}")
}
```
Also you can avoid the casting with
```rust
self.nonconsensus_get_raw(format!("storage_version/{height}").as_bytes())
```

crates/astria-sequencer/src/app.rs
```diff
@@ -162,9 +162,22 @@ impl App {
...
+        // store the storage version indexed by block height
+        let height = state
+            .get_block_height()
+            .await
+            .expect("block height should be set");
+        state.put_storage_version_by_height(height, storage.latest_version().wrapping_add(1));
+        debug!(
+            ?height,
+            version = storage.latest_version().wrapping_add(1),
+            "stored storage version"
```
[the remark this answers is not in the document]
noot: i think `stored` is more accurate since it's storing a new key-value entry essentially, is `stored storage version for height` clearer perhaps?

crates/astria-sequencer/src/state_ext.rs
noot: yep I think we can bump our dep, will do in another PR

crates/astria-sequencer/src/app.rs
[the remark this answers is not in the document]
SuperFluffy: You are right, updated is bad.

I thought more about the lines of "committed to database". Somehow the alliteration makes it unclear?

Sorry, feel free to ignore.

crates/astria-sequencer/src/state_ext.rs
noot: ah turns out this wasn't in the v0.56 release lol

joroshiba (approved):

joroshiba:
Lets hold off on any merging until after janis work gets in and we can cut a new release

PR #176, opened 2023-07-18
bump penumbra version

noot:


SuperFluffy (approved):
Doesn't yet contain the rename, but this is good either way.

Wonder if there is some dependabot that we can use to notify us of updates.

steezeburger (approved):

sambukowski (approved):

PR #178, opened 2023-07-19
Add a local docker build command

joroshiba:


crates/astria-sequencer-relayer/justfile
```diff
@@ -1,3 +1,5 @@
+set fallback
```
SuperFluffy, line 1: this is for when you are not in the root of the repo?

SuperFluffy (approved):

crates/astria-sequencer-relayer/justfile
joroshiba: Yeah, specifically it says that  this justfile will fallback to any parent justfiles if commands don't exist.

So you can use the build command while in any folder of the repo.

PR #188, opened 2023-07-20
proto: commit generated code to repository

SuperFluffy:
Committing the compiled protobufs to the
repository servces two purposes:

1. it is easier to discover the compiled rust types by simply navigating to the files;
2. users who just want to use the rust crates need not install buf and protoc to compile the protobufs to rust on every build.

While it is convention to not commit generated
code to the repository, I believe the two points
above are good arguments why one should do this
in this case. Also, the protobuf definitions are
not expected to change frequently.

In addition, this commit pins the version of the
remote buf plugins to defend against accidental
breakage.


noot (commented):
can you also update the readme to mention buf isn't required to build anymore (only if protos are modified for development purposes)?

crates/astria-proto/src/proto/tonic/astria.execution.v1.rs
```diff
@@ -0,0 +1,39 @@
...
+pub struct InitStateRequest {
+}
+#[allow(clippy::derive_partial_eq_without_eq)]
+#[derive(Clone, PartialEq, ::prost::Message)]
+pub struct InitStateResponse {
+    #[prost(bytes="vec", tag="1")]
+    pub block_hash: ::prost::alloc::vec::Vec<u8>,
+}
+include!("astria.execution.v1.tonic.rs");
+// @@protoc_insertion_point(module)
```
steezeburger, line 39: newline nit

SuperFluffy: This is generated code.

SuperFluffy:
@noot 
> can you also update the readme to mention buf isn't required to build anymore (only if protos are modified for development purposes)?

Updated the README, removed the uncommented code (ty for noticing), and also removed the now unnecessary install step of `buf` and `protobuf-compiler` from the dockerfile.

crates/astria-proto/src/proto/tonic/astria.execution.v1.rs
steezeburger: lmao sorry, it's early! 😄

crates/astria-proto/README.md
```diff
@@ -1,6 +1,32 @@
 # Astria Proto
 
-This repo contains all the protobuf packages for Astria. 
+This repo contains all the protobuf packages for Astria. All rust code
+generated from the protobuf files in [`proto/`](`./proto/`) is commited
+to this repository and no extra tools are needed to encode to/decode from
+protobuf.
+
+Only when changing protobuf definitions (which is done by running running
```
steezeburger, line 8:
```suggestion
Only when changing protobuf definitions (which is done by running
```

crates/astria-proto/README.md
```diff
@@ -10,9 +36,11 @@ This repo contains all the protobuf packages for Astria.
 
 ## Adding a package
 
-* Create a new folder under `proto/astria/{new_package}`
-* write protos in new folder using package name `astria.{new_package}`
-* update `src/lib.rs` to include your new defintions in a module. Module names should reflect proto package names.
+* Create a new folder `proto/astria/<pkg-name>/<version>`;
```
steezeburger, line 39: Should this line and the line below end in semi-colons? Maybe period

crates/astria-proto/README.md
```diff
@@ -10,9 +36,11 @@ This repo contains all the protobuf packages for Astria.
...
+* write protos in this folder using the convention name `astria.<pkg-name>.<version>`;
+* update `src/proto/mod.rs` to include your new defintions in a module.
+  Follow the conventions of the other submodules;
```
steezeburger, line 42: Should this semicolon be a period? or a regular colon?

steezeburger (approved):
Saw a few issues with the README, but looks good overall

noot (approved):
can you update the readme in the repo root also to remove that buf is required? otherwise looks good!

SuperFluffy:
I noticed that `buf` has an internal protobuf compiler, so we can remove the extra step installing `protoc` in the github actions.

PR #190, opened 2023-07-21
sequencer-client: replace type wrapper by extension trait

itamarreif:
This commit replaces the newtype wrapper around `tendermint_rpc::HttpClient` with an extension trait `SequencerClientExt: tendermint_rpc::client::Client` that is implemented for both `HttpClient` and `WebsocketClient`. The actual impls are behind `http` and `websocket` feature gates, respectively.

In addition, this commit removes the dynamic errors in favor of typed errors that report on which underlying RPC failed, if deserialization of the abci query response payload failed, or if a sequencer query response other than the expected one was received.

To leave deserialization logic as an implementatian detail, the underlying error in case of failure is not accessible (outside of type erased calls to `std::error::Error::source`).


crates/astria-conductor/src/block_verifier.rs
```diff
@@ -46,22 +50,21 @@ use tracing::{
 /// `validate_signed_namespace_data` is used to validate the data received from the data
 /// availability layer. `validate_sequencer_block` is used to validate the blocks received from
 /// either the data availability layer or the gossip network.
-pub struct BlockVerifier {
-    sequencer_client: Client,
+#[async_trait]
+pub trait BlockVerifier {
```
SuperFluffy, line 54: I don't think this should be an extension trait. Just toss the http client in there and import the extension trait. Apart from that nothing in conductor should change.

crates/astria-sequencer-client/src/client.rs
```diff
@@ -17,95 +18,41 @@ use eyre::{
...
 };
 
 /// Default Tendermint base URL.
 pub const DEFAULT_TENDERMINT_BASE_URL: &str = "http://localhost:26657";
 
 /// Tendermint HTTP client which is used to interact with the Sequencer node.
-pub struct Client {
-    client: HttpClient,
+#[async_trait]
+pub trait Client {
```
SuperFluffy, line 33: Let's call this `ClientExt` to make clear that it is an extension trait. It should also be defined as `pub trait ClientExt: tendermint_rpc::Client`.

Look at `TryFutureExt` as an example of how that's done: https://docs.rs/futures-util/0.3.28/src/futures_util/future/try_future/mod.rs.html#137

crates/astria-sequencer-client/src/client.rs
```diff
@@ -17,95 +18,41 @@ use eyre::{
...
-        height: T,
-    ) -> eyre::Result<validators::Response> {
-        self.client
-            .validators(height.into(), tendermint_rpc::Paging::Default)
-            .await
-            .wrap_err("failed to get validator set")
-    }
-
+#[async_trait]
+impl Client for tendermint_rpc::HttpClient {
```
SuperFluffy, line 42: You can avoid all these explicit impls by making `ClientExt: Client` a supertrait of `tendermint_rpc::Client` and then providing default trait methods. That way all of these functions will immediately work for both, the http and the websocket clients.

SuperFluffy (changes requested):
Thank you for the PR. I like that we are removing one layer of abstraction around tendermint rpc's HttpClient/WebsocketClient.

Please make the extension trait a super trait of `tendermint_rpc::Client` (which is also a trait), and provide the methods as default trait methods. That way you will get all of the functionality for the http and the ws clients for free at the same time.

crates/astria-sequencer-client/src/lib.rs
```diff
@@ -4,6 +4,10 @@ pub use astria_sequencer::transaction::{
     Action,
     Signed as SignedTransaction,
 };
-pub use client::Client;
+pub use client::SequencerClientExt;
 /// Reexports
 pub use tendermint_rpc::endpoint::block::Response as BlockResponse;
```
SuperFluffy, line 9: IMO this reexport + type alias should be removed.

crates/astria-sequencer-client/src/client.rs
```diff
@@ -193,21 +122,20 @@ impl Client {
     /// # Errors
     ///
     /// - If calling the tendermint RPC endpoint fails.
-    pub async fn submit_transaction_commit(&self, tx: Signed) -> eyre::Result<tx_commit::Response> {
+    async fn submit_transaction_commit(&self, tx: Signed) -> eyre::Result<tx_commit::Response> {
```
SuperFluffy, line 125: `s/Signed/transaction::Signed`

SuperFluffy, line 125: I think this could just return the error of `broadcast_tx_commit` instead of wrapping it in `eyre`

crates/astria-sequencer-client/src/client.rs
```diff
@@ -177,10 +107,9 @@ impl Client {
     /// # Errors
     ///
     /// - If calling the tendermint RPC endpoint fails.
-    pub async fn submit_transaction_sync(&self, tx: Signed) -> eyre::Result<tx_sync::Response> {
+    async fn submit_transaction_sync(&self, tx: Signed) -> eyre::Result<tx_sync::Response> {
```
SuperFluffy, line 110: As for `submit_transaction_commit`, let's get rid of `eyre` here.

crates/astria-sequencer-client/src/client.rs
```diff
@@ -9,103 +9,38 @@ use astria_sequencer::{
...
-    },
-    Client as _,
-    HttpClient,
+use tendermint_rpc::endpoint::broadcast::{
+    tx_commit,
+    tx_sync,
 };
 
 /// Default Tendermint base URL.
 pub const DEFAULT_TENDERMINT_BASE_URL: &str = "http://localhost:26657";
```
SuperFluffy, line 26: IMO this should also be removed.

crates/astria-sequencer-client/src/client.rs
SuperFluffy, line 1: It's more a style question, but I think the entire `cilent` module should just be merged into `lib.rs`. It's IMO too small to have it be separate.

crates/astria-sequencer-client/Cargo.toml
```diff
@@ -6,14 +6,15 @@ edition = "2021"
...
+async-trait = { workspace = true }
 borsh = { workspace = true }
 ed25519-consensus = { workspace = true }
 eyre = { workspace = true }
 reqwest = { workspace = true, features = ["json"] }
 serde = { workspace = true, features = ["derive"] }
 serde_json = { workspace = true }
 tendermint = { workspace = true }
-tendermint-rpc = { workspace = true, features = ["http-client"]}
+tendermint-rpc = { workspace = true, features = ["http-client", "websocket-client"]}
```
SuperFluffy, line 17: You could also feature gate both of these by creating features like this:

```
[features]
http-client = ["tendermint-rpc/http-client"]
websocket-client = ["tendermint-rpc/websocket-client"]
```

You'd then just need to add `#[cfg(feature = "http-client")]` on the impl of `SequencerClientExt` for the http client, and similarly for the websocket client.

SuperFluffy (approved):
Looks good, thank you. Just a few style comments here and there.

crates/astria-conductor/src/block_verifier.rs
```diff
@@ -75,7 +78,7 @@ impl BlockVerifier {
         );
         let validator_set = self
             .sequencer_client
-            .get_validator_set(height)
+            .validators(height - 1, tendermint_rpc::Paging::Default)
```
noot, line 81: why is this height - 1? change back?

crates/astria-conductor/src/block_verifier.rs
```diff
@@ -47,21 +50,21 @@ use tracing::{
...
+            sequencer_client: HttpClient::new(sequencer_url)
                 .wrap_err("failed to construct sequencer client")?,
         })
     }
 
     /// validates `SignedNamespaceData` received from Celestia.
     /// This function verifies the block signature and checks that the data
     /// was signed by the expected proposer for this block height.
-    pub(crate) async fn validate_signed_namespace_data(
+    pub async fn validate_signed_namespace_data(
```
noot, line 67: why pub?

crates/astria-conductor/src/block_verifier.rs
```diff
@@ -121,7 +121,7 @@ impl BlockVerifier {
         // in the block is for the previous height
         let validator_set = self
             .sequencer_client
-            .get_validator_set(height - 1)
+            .validators(height - 1, tendermint_rpc::Paging::Default)
```
noot, line 124: same here

crates/astria-conductor/src/block_verifier.rs
SuperFluffy: This was probably an oversight undoing the `BlockVerifier as extension trait` change.

crates/astria-conductor/src/block_verifier.rs
SuperFluffy: @itamarreif this is changed from `height` to `height - 1`, whereas the other one is probably correct.

crates/astria-conductor/src/block_verifier.rs
SuperFluffy: This is probably correct (`height -1` to `height - 1`), whereas the other call above indeed changes `height - 1` to `height`.

joroshiba (approved):

noot (approved):

crates/astria-sequencer-client/Cargo.toml
SuperFluffy: Introduced features `http` and `websocket`

crates/astria-sequencer-client/src/lib.rs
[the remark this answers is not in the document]
SuperFluffy: Added a doc comment at the top of `lib.rs`

SuperFluffy:
Changes since the PR was approved:

1. added a crate level doc comment with a code example
2. added a job running doc tests

Issue #138, opened 2023-07-06
Refactor Execution API


joroshiba:
The current execution API has some naming issues, and isn't fully thought out. It should be redesigned. Working doc to be turned into spec in notion [here](https://www.notion.so/astria-org/Execution-API-Refresh-e9f9777311b0441db39be803291a0f3f?d=ccef5dfd5436481eb4720692cb247fff#6cc2196e59754f0abe3a78b91d47aa0b) 
```[tasklist]
### Tasks
- [ ] https://github.com/astriaorg/astria/issues/84
- [ ] https://github.com/astriaorg/astria/issues/205
- [ ] geth: implement new execution API server
- [ ] https://github.com/astriaorg/astria/issues/207
```

noot:
related / possible duplicate of #50

jbowen93:
Possibly related to https://github.com/astriaorg/astria/issues/84

PR #193, opened 2023-07-25
proto: execution API redesign

joroshiba:
- Redesign of  Execution API as v1alpha2.
- Documents how v1alpha2 API should work. 

#138


crates/astria-proto/proto/astria/execution/v1alpha2/execution.proto
```diff
@@ -0,0 +1,107 @@
+syntax = 'proto3';
+
+package astria.execution.v1alpha2;
+
+import "google/protobuf/timestamp.proto";
+
+// The set of information which deterministic driver may to know about a given executed Block
```
noot, line 7:
```suggestion
// The set of information which a deterministic driver may want to know about a given executed Block
```

maybe `driver of execution` or something like that as well?

crates/astria-proto/proto/astria/execution/v1alpha2/execution.proto
```diff
@@ -0,0 +1,107 @@
...
+message Block {
+  // The block number
+  uint32 number = 1;
+  // The hash of the block
+  bytes hash = 2;
+  // Timestamp on the block, standardized to google protobuf standard.
+  google.protobuf.Timestamp timestamp = 3;
+}
+
+// These fields should be indexed on most block chains, and can be used to identify a block.
```
noot, line 17:
```suggestion
// These fields can be used to identify a block.
```

I think every block chain can index using one of these, would be concerning if not lol

crates/astria-proto/proto/astria/execution/v1alpha2/execution.proto
```diff
@@ -0,0 +1,107 @@
...
+}
+message BatchGetBlocksResponse {
+  repeated Block blocks = 1;
+}
+
+
+// CreateBlockRequest contains all the information needed to create a new executed block.
+//
+// This information comes from previous execution blocks, as well as from sequencer blocks.
+message CreateBlockRequest {
```
noot, line 43:
```suggestion
// ExecuteBlockRequest contains all the information needed to create a new executed block.
//
// This information comes from previous execution blocks, as well as from sequencer blocks.
message ExecuteBlockRequest {
```

I slightly prefer this, thoughts?

crates/astria-proto/proto/astria/execution/v1alpha2/execution.proto
```diff
@@ -0,0 +1,107 @@
...
+  bytes prev_block_hash = 1;
+  // List of transactions to include in the new block
+  repeated bytes transactions = 2;
+  // The timestamp which should be used for new block
+  google.protobuf.Timestamp timestamp = 3;
+}
+// CreateBlockResponse is returned after calling CreateBlock
+message CreateBlockResponse {
+  // The executed block which was created.
+  Block block = 1;
```
noot, line 54: this returns the timestamp which was already specified in the request, do you think that's fine vs just returning hash and number here?

crates/astria-proto/proto/astria/execution/v1alpha2/execution.proto
```diff
@@ -0,0 +1,107 @@
...
+}
+
+// The CommitmentState holds the block at each stage of sequencer commitment level
+//
+// A Valid CommitmentState:
+// - Block numbers are such that soft+1 >= head >= soft >= firm.
```
noot, line 60:
```suggestion
// - Block numbers are such that head >= soft >= firm.
```

soft+1 can never be more than head? unless head is already softly committed (which shouldn't be the case with tendermint)

crates/astria-proto/proto/astria/execution/v1alpha2/execution.proto
```diff
@@ -0,0 +1,107 @@
...
+// - No blocks ever decrease in block number, only head may stay the same and have other changes
```
noot, line 61: you mean there can only be 1-block forks max basically?

crates/astria-proto/proto/astria/execution/v1alpha2/execution.proto
```diff
@@ -0,0 +1,107 @@
...
+// - The chain defined by head contains soft and firm blocks.
```
noot, line 62: basically every ancestor of head must be softly or firmly committed?

crates/astria-proto/proto/astria/execution/v1alpha2/execution.proto
```diff
@@ -0,0 +1,107 @@
...
+message CommitmentState {
+  // The head is the top of the executed chain
+  Block head = 1;
+  // Soft commitment is the executed block matching sequencer block with full consensus.
+  Block soft = 2;
+  // Firm commitment 
+  Block firm = 3;
```
noot, line 69: might be good to clarify what exactly the difference is between soft and firm as neither can be reverted in our current design

crates/astria-proto/proto/astria/execution/v1alpha2/execution.proto
```diff
@@ -0,0 +1,107 @@
...
+
+// There is only one CommitmentState object, so the request is empty.
+message GetCommitmentStateRequest{}
+message GetCommitmentStateResponse {
+  // The current CommitmentState
+  CommitmentState commitment_state = 1;
+}
+
+message UpdateCommitmentStateRequest {
+  // The CommitmentState to set, must include complete state and pass validation checks.
```
noot, line 80: pass validation checks where? on the execution side?

crates/astria-proto/proto/astria/execution/v1alpha2/execution.proto
```diff
@@ -0,0 +1,107 @@
...
+message GetBlockResponse {
+    Block block = 1;
+}
+
+message BatchGetBlocksRequest {
+  repeated BlockIdentifier identifiers = 1;
+}
+message BatchGetBlocksResponse {
+  repeated Block blocks = 1;
+}
```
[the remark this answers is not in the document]
joroshiba: The main use case is a range, ie on startup of conductor from a failed state fetching information back about blocks between firm and soft commitment.

Verb choice is based on AIPs across the board using the standard where appropriate: https://google.aip.dev/231 for batch get.

crates/astria-proto/proto/astria/execution/v1alpha2/execution.proto
joroshiba: Added comment on what firm means

crates/astria-proto/proto/astria/execution/v1alpha2/execution.proto
joroshiba: added clarification

crates/astria-proto/proto/astria/execution/v1alpha2/execution.proto
```diff
@@ -0,0 +1,107 @@
...
+
+// There is only one CommitmentState object, so the request is empty.
+message GetCommitmentStateRequest{}
+message GetCommitmentStateResponse {
+  // The current CommitmentState
+  CommitmentState commitment_state = 1;
+}
+
+message UpdateCommitmentStateRequest {
```
sambukowski, line 70: Do we always want to update head, safe, and final every time?
Would it be worth having individual `UpdateHeadCommit`, `UpdateSoftcommit`, and `UpdateFirmCommit` messages?

crates/astria-proto/proto/astria/execution/v1alpha2/execution.proto
joroshiba: All verbs chosen to be resource based: https://google.aip.dev/133. My thought here is that fundamentally we have two resources (Blocks, and a CommitmentState). Creating a block is what the API call does, executing is just a part of creating the new resource.

crates/astria-proto/proto/astria/execution/v1alpha2/execution.proto
```diff
@@ -0,0 +1,107 @@
...
+  // The CommitmentState to set, must include complete state and pass validation checks.
+  CommitmentState commitment_state = 1;
+}
+message UpdateCommitmentStateResponse {
+  // The result of the commitment state update.
+  CommitmentState commitment_state = 1;
```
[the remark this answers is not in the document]
joroshiba: Error responses can happen regardless of the response object: https://grpc.github.io/grpc/core/md_doc_statuscodes.html

crates/astria-proto/proto/astria/execution/v1alpha2/execution.proto
joroshiba: My initial design had it as three seperate commitments, but given that their validity is relative to each other I think it's cleaner to treat it as a single object.

crates/astria-proto/proto/astria/execution/v1alpha2/execution.proto
joroshiba: Cleaned up as non-important to this comment, but yeah validation (as always) should be done by the server upon receiving a request.

crates/astria-proto/proto/astria/execution/v1alpha2/execution.proto
joroshiba: Cleaned up response types to match AIP: https://google.aip.dev/133

specs/execution-api.md
```diff
@@ -0,0 +1,72 @@
...
+
+- No commitment can ever decrease in block number on the blockchain, if this is attempted return a `FAILED_PRECONDITION` error.
+- `HEAD` commitments MAY be to blocks which replace the previous `HEAD` with same block number.
+- `SOFT` and `FIRM` block MUST either increase in block number OR match current commitment state block.
+- `SOFT` and `FIRM` blocks MUST be members of the block chain starting from `HEAD`.
+- Block numbers in state MUST be such that `SOFT` + 1 >= `HEAD` >= `SOFT` >= `FIRM`, return a `FAILED_PRECONDITION` error if this is not true
+
+## Sequence Diagram
+
+![image](assets/execution_api_sequence.png)
```
sambukowski, line 74: I don't see an assets folder or image for this.

specs/execution-api.md
```diff
@@ -0,0 +1,74 @@
...
+### Startup
+
+Upon startup, conductor needs to know the state of commitments in the state machine and calls `GetCommitmentState`. If started on a fresh rollup these will all be the same block. If running against a state machine with previous block data, Conductor must also track the block hash of any blocks between commitments, it will call `BatchGetBlock` to get block information between commitments.
+
+### Execution & Commitments
+
+From the perspective of the sequencer, `HEAD` commitments are made everytime the sequencer creates a new block at height N. The transactions on this block may change up until sequencer block at height N+1 has been created, at this point the block at height N has a `SOFT` commitment. A soft commitment means that sequencer consensus has full agreement on the block. Once a block has been seen on DA it has a `FIRM` commitment, it has been made widely available across the DA network.
+
+When configuring conductor, you can configure the time at which blocks are executed in your rollup using the `execution_commitment_level` in the config file. If this is configured to a higher level of commitment, no action will be taken upon receiving lower commitments.
```
sambukowski, line 23: Is this config the ConductorConfig.toml?

specs/execution-api.md
```diff
@@ -0,0 +1,74 @@
...
+
+When configuring conductor, you can configure the time at which blocks are executed in your rollup using the `execution_commitment_level` in the config file. If this is configured to a higher level of commitment, no action will be taken upon receiving lower commitments. 
+
+`CreateBlock` is called to create a new execution chain block when the `execution_commitment_level` has been been reached for a given block. Upon receipt of a new block, the conudctor calls `UpdateCommitmentState` to update the commitment at the level of the `execution_commitment_level` and any level above it.
+
+If `execution_commitment_level` is set to `HEAD`, `UpdateCommitmentState` will be called with an update to the `SOFT` commitment block after receiving a sequencer block with `prev_hash` which has already been executed on the chain.
```
sambukowski, line 27: I'd prefer the different `execution_commitment_level` options in a list for visual clarity.

specs/execution-api.md
```diff
@@ -0,0 +1,74 @@
...
+- The API endpoint MUST fail atomically, returning either all requested resource or a `NOT_FOUND` error.
+- The returned objects MUST be in the same order as they were requested.
+
+### GetCommitmentState
+
+Returns the commitment state with execution `Block` information for each level of commitment.
+
+### UpdateCommitmentState
+
+`UpdateCommitmentState` replaces the
```
sambukowski, line 62: Was there supposed to me more here?

specs/execution-api.md
```diff
@@ -0,0 +1,74 @@
+# Execution API Specification
+
+## Overview
+
+The Execution API is the interface `Conductor` uses to drive deterministic execution of sequenced blocks in dependent chains. Inspired by other APIs, such as the Engine API and ABCI Consensus API, it is a chain agnostic mechanism intended to be very simple to implement. It is a gRPC API which any state machine can implement, and use conductor with to drive their execution and integrate with the Astria Sequencer.
+
+> Note: this documentation is for the v1alpha2 API which is currently being implemented, and some of the documentation is on how it should be implemented not how it is.
```
noot, line 7:
```suggestion
> Note: this documentation is for the v1alpha2 API which is currently being implemented, and some of the documentation is on how it should be implemented, not how it is currently.
```

minor nit

specs/execution-api.md
noot, line 23: `CreateBlockRequest` doesn't specify the commitment level so I assume will always set the latest block as `HEAD`. so if `execution_commitment_level = soft` for example, then the conductor will send a `CreateBlockRequest` on receiving a soft commit and then immediately set the just executed block to `SOFT`?

specs/execution-api.md
```diff
@@ -0,0 +1,74 @@
...
+
+When configuring conductor, you can configure the time at which blocks are executed in your rollup using the `execution_commitment_level` in the config file. If this is configured to a higher level of commitment, no action will be taken upon receiving lower commitments. 
+
+`CreateBlock` is called to create a new execution chain block when the `execution_commitment_level` has been been reached for a given block. Upon receipt of a new block, the conudctor calls `UpdateCommitmentState` to update the commitment at the level of the `execution_commitment_level` and any level above it.
+
+If `execution_commitment_level` is set to `HEAD`, `UpdateCommitmentState` will be called with an update to the `SOFT` commitment block after receiving a sequencer block with `prev_hash` which has already been executed on the chain.
+
+If `execution_commitment_level` is set to `SOFT` or lower, `UpdateCommitmentState` will be called to update the `FIRM` commitment block when it's corresponding sequencer block has been seen in the DA
```
noot, line 29: I find these a bit confusing, for the first, is it saying that the soft commitment is updated to the head block `N` when we receive its child `N+1`? and similar for the second?

specs/execution-api.md
```diff
@@ -0,0 +1,74 @@
...
+- `SOFT` Commitment -> `SAFE` Forkchoice
+- `FIRM` Commitment -> `FINAL` Forkchoice
+
+## Rollup Implementation Details
+
+### CreateBlock
+
+`CreateBlock` executes a set of given trasactions on the new chain. The following should be respected:
+
+- `prev_block_hash` MUST match hash of the `SOFT` commitment state block, return `FAILED_PRECONDITION` otherwise.
```
noot, line 42: wait why? shouldn't it also be allowed to match `HEAD`?

specs/execution-api.md
```diff
@@ -0,0 +1,74 @@
...
+### GetCommitmentState
+
+Returns the commitment state with execution `Block` information for each level of commitment.
+
+### UpdateCommitmentState
+
+`UpdateCommitmentState` replaces the 
+
+- No commitment can ever decrease in block number on the blockchain, if this is attempted return a `FAILED_PRECONDITION` error.
+- `HEAD` commitments MAY be to blocks which replace the previous `HEAD` with same block number.
```
noot, line 65: how can `HEAD` ever increase then?

crates/astria-proto/proto/astria/execution/v1alpha2/execution.proto
noot: okay, i see `head` can equal `soft` now. however, i find this a bit confusing since `soft+1 >= head` implies `head < soft+1` which is not always the case, `head == soft+1` can occur and probably often will

specs/execution-api.md
```diff
@@ -0,0 +1,74 @@
...
+- `SOFT` and `FIRM` block MUST either increase in block number OR match current commitment state block.
+- `SOFT` and `FIRM` blocks MUST be members of the block chain starting from `HEAD`.
+- Block numbers in state MUST be such that `SOFT` + 1 >= `HEAD` >= `SOFT` >= `FIRM`, return a `FAILED_PRECONDITION` error if this is not true
```
noot, line 68: same comment on `soft + 1 >= head` as above

noot (commented):
as a general comment, i'm wondering if `CreateBlock` and `UpdateCommitment` should be atomic/done together - is there ever a case where we want to execute a block but not also update the commitment?

specs/execution-api.md
```diff
@@ -0,0 +1,74 @@
+# Execution API Specification
+
+## Overview
+
+The Execution API is the interface `Conductor` uses to drive deterministic execution of sequenced blocks in dependent chains. Inspired by other APIs, such as the Engine API and ABCI Consensus API, it is a chain agnostic mechanism intended to be very simple to implement. It is a gRPC API which any state machine can implement, and use conductor with to drive their execution and integrate with the Astria Sequencer.
```
steezeburger, line 5:
```suggestion
The Execution API is the interface `Conductor` uses to drive deterministic execution of sequenced blocks in dependent chains. Inspired by other APIs, such as the Engine API and ABCI Consensus API, it is a chain agnostic mechanism intended to be very simple to implement. It is a gRPC API which any state machine can implement and use conductor with to drive their execution and integrate with the Astria Sequencer.
```
minor nit

specs/execution-api.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Conductor Usage
+
+### Startup
+
+Upon startup, conductor needs to know the state of commitments in the state machine and calls `GetCommitmentState`. If started on a fresh rollup these will all be the same block. If running against a state machine with previous block data, Conductor must also track the block hash of any blocks between commitments, it will call `BatchGetBlock` to get block information between commitments.
+
+### Execution & Commitments
+
+From the perspective of the sequencer, `HEAD` commitments are made everytime the sequencer creates a new block at height N. The transactions on this block may change up until sequencer block at height N+1 has been created, at this point the block at height N has a `SOFT` commitment. A soft commitment means that sequencer consensus has full agreement on the block. Once a block has been seen on DA it has a `FIRM` commitment, it has been made widely available across the DA network.
```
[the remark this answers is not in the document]
steezeburger:
```
A soft commitment means that sequencer consensus has full agreement on the block. Once a block has been seen on DA it has a `FIRM` commitment, it has been made widely available across the DA network.
```

Does `FIRM` commitment mean it has been made widely available across DA? I think an "i.e." would be clarifying here if so.

specs/execution-api.md
joroshiba: Another doc in the works on how to manage configuration.

specs/execution-api.md
```diff
@@ -0,0 +1,74 @@
...
+
+When configuring conductor, you can configure the time at which blocks are executed in your rollup using the `execution_commitment_level` in the config file. If this is configured to a higher level of commitment, no action will be taken upon receiving lower commitments. 
+
+`CreateBlock` is called to create a new execution chain block when the `execution_commitment_level` has been been reached for a given block. Upon receipt of a new block, the conudctor calls `UpdateCommitmentState` to update the commitment at the level of the `execution_commitment_level` and any level above it.
```
steezeburger, line 25:
```suggestion
`CreateBlock` is called to create a new execution chain block when the `execution_commitment_level` has been reached for a given block. Upon receipt of a new block, the conudctor calls `UpdateCommitmentState` to update the commitment at the level of the `execution_commitment_level` and any level above it.
```

specs/execution-api.md
steezeburger, line 29:
```suggestion
If `execution_commitment_level` is set to `SOFT` or lower, `UpdateCommitmentState` will be called to update the `FIRM` commitment block when its corresponding sequencer block has been seen in the DA.
```

crates/astria-proto/proto/astria/execution/v1alpha2/execution.proto
joroshiba: Basically the options are only soft + 1 or soft, I think this accurately represents that? Perhaps it would be better split out into a seperate validation list so it's `head >= soft >= firm` and `head block never be more than 1 block ahead of soft block`?

specs/execution-api.md
joroshiba: I will clarify in the `CreateBlock` section but as seen in the sequence diagram, CreateBlock creates a block but it does not mark it as the head, this is done in a follow up request. This is a departure from our previous API. This is done, in part to enable this different execution (it also follows a pattern similar to Engine API)

specs/execution-api.md
joroshiba: I still feel it's correct, but open to rewording? Head can never be further than 1 ahead of safe since that would enable greater than 1 block reorgs.

specs/execution-api.md
joroshiba: Key word may, this is basically saying "you can do a 1 block reorg". Let me know if you know if you have better idea on verbiage here.

specs/execution-api.md
joroshiba: You can when HEAD == SAFE. 

This is the whole, safe+1 thing. If block 50 was head, once 51 has been seen by conductor 50 is marked safe. I added some text further up to make this more clear.

specs/execution-api.md
joroshiba: Updated verbiage to be much more explicit.

crates/astria-proto/proto/astria/execution/v1alpha2/execution.proto
sambukowski: Does this make more sense?
`head >= (safe + 1 or safe) >= firm` with:
`head >= safe + 1` when `execution_commitment_level` == `head`
 and `head >= safe` when `execution_commitment_level` == `soft`

Gdakgdak22 (approved):
Gracias

specs/execution-api.md
```diff
@@ -0,0 +1,95 @@
...
+### Startup
+
+Upon startup, conductor needs to know the state of commitments in the state machine and calls `GetCommitmentState`. If started on a fresh rollup these will all be the same block. If running against a state machine with previous block data, Conductor must also track the block hash of any blocks between commitments, it will call `BatchGetBlock` to get block information between commitments.
+
+### Execution & Commitments
+
+From the perspective of the sequencer:
+- `HEAD` commitments are made every time the sequencer creates a new block at height N. Only the `HEAD` blocks can be reorged and they can be updated until the N+1 block has been created. Once the N+1 block is received, the block at height N is set to `SOFT`
+- `SOFT` commitment means that sequencer consensus has full agreement on the block.
+- `FIRM` commitment indicates that the block has been written, and has been propogated across the DA network.
```
steezeburger, line 24:
```suggestion
- `FIRM` commitment indicates that the block has been written and has been propagated across the DA network.
```
The comma suggestion is just a nit. You shouldn't need a comma when combining an independent and dependent clause.

sambukowski (approved):

joroshiba:
> as a general comment, i'm wondering if `CreateBlock` and `UpdateCommitmentState` should be atomic/done together - is there ever a case where we want to execute a block but not also update the commitment state?

@noot I don't think in the conductor use case right now there currently is, but I also don't think that changes that they are separate operations. Part of why to build this as resource oriented is to not limit ourselves by current use cases. I could, for instance, see the API used by a builder to ensure that blocks are constructed validly. They wouldn't want to update their commitment state because there is no commitment.

We also similar separations in other chain driving APIs: engine API has `newPayload` and `forkchoiceUpdated`, or ABCI having `FinalizeBlock` and `Commit`.

noot:
@joroshiba yep considering the other APIs that makes sense, cool!

noot (approved):

Issue #191, opened 2023-07-24
Separate public sequencer API from sequencer binary crate


SuperFluffy:
https://github.com/astriaorg/astria/pull/129 introduced the `sequencer-client` crate and made `sequencer` a dependency of it. This caused all crates that use the client to transitively pull in all of sequencer and its dependencies blowing up their compile times (especially due to rocksdb).

The types that are part of sequencer's public API should be split out into their own crate (for example `sequencer-rpc`), and made a dependency of sequencer, sequencer-client, etc.

Sequencer can subsequently be closed down again (making its now public types crate-private again).

noot:
closed by #311

PR #172, opened 2023-07-17
container: speedup docker builds by only building binary dependencies

SuperFluffy:
This commit adds the `--bin` arguments to the `cargo chef prepare` and `cargo chef cook` steps in order to speed up compilation by only compiling workspace members that are needed for building the target binary.

Note that at the time of this commit only speedups for `astria-sequencer` are noticable, because `astria-conductor` and `astria-sequencer-relayer` transitively pull in most of the other workspace members. This is tracked in issue #191.

Because cargo-chef version 0.1.61 contains a bug when preparing the initial recipe, this commit uses the argument `--bin crates/$TARGETBINARY` in the prep phase, but `--bin $TARGETBINARY` in the cook phase.

PR #198, opened 2023-07-27
ci: use BuildJet for faster GH Actions

joroshiba:
Introduces BuildJet to our GH actions, utilize 8vCPU for docker builds & 4vCPU for tests. This reduces time for all actions to under 5 minutes.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -56,7 +56,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: dtolnay/rust-toolchain@1.69.0
-      - uses: Swatinem/rust-cache@v2
+      - uses: astriaorg/buildjet-rust-cache@v2.5.1
       - name: Run doctests
         run: cargo test --doc --all-features
 
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -11,11 +11,11 @@ on:
 
 jobs:
   test:
-    runs-on: ubuntu-22.04
+    runs-on: buildjet-4vcpu-ubuntu-2204
     steps:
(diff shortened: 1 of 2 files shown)


jbowen93 (approved):

joroshiba:
Note: I've opened an PR to upstream the rust caching changes I've made. Was encountering performance issues with gh cache when in self-hosted runner, so created update to allow use of buildjet cache under the hood.

Upstream PR: https://github.com/Swatinem/rust-cache/pull/154

SuperFluffy:
I would like to hold off on merging this PR until we see the impact of #172, #200, and (ideally, but not necessarily) #191.

joroshiba:
> I would like to hold off on merging this PR until we see the impact of #172, #200, and (ideally, but not necessarily) #191.

It looks like as expected, caching reduces the performance improvement from bigger CPU, however we are still cutting build times in half.

.github/workflows/rust.yml
```diff
@@ -11,11 +11,11 @@ on:
 
 jobs:
   test:
-    runs-on: ubuntu-22.04
+    runs-on: buildjet-4vcpu-ubuntu-2204
     steps:
       - uses: actions/checkout@v3
       - uses: dtolnay/rust-toolchain@1.69.0
-      - uses: Swatinem/rust-cache@v2
+      - uses: joroshiba/rust-cache@v2.5.1
```
SuperFluffy, line 18: Not a blocker: this should ideally be under the astria org

SuperFluffy (approved):
Good quality of life improvement. You are right, makes no sense to delay

.github/workflows/rust.yml
joroshiba: done

PR #199, opened 2023-07-27
ci: auto label prs

joroshiba:
- Adds a workflow to apply PR labels to encourage small PRs
- Moves generated code into more clearly labelled directory
- Ensures generated code isn't included in line counts
- Auto labels prs based on config directories.


SuperFluffy:
Bit torn on this because I feel like it's just going to generate extra noise?

SuperFluffy (approved):
Let's try this.

steezeburger (approved):

.github/labeler.yml
```diff
@@ -0,0 +1,16 @@
+proto:
```
SuperFluffy, line 1: Could we add a label for the containerfiles (== docker) and name it `container`? The docker-related workflows might also fall under it (less sure about that).

crates/astria-proto/src/lib.rs
```diff
@@ -3,7 +3,7 @@
 #[rustfmt::skip]
 mod proto;
 
-pub use proto::tonic::{
+pub use proto::generated::{
```
SuperFluffy, line 6: Nit: fine with the rename, although it feels orthogonal to the rest of the PR

.gitattributes
```diff
@@ -0,0 +1 @@
+crates/astria-proto/src/proto/generated/** linguist-generated=true
```
SuperFluffy, line 1: Never saw this, cool

SuperFluffy (approved):
Looks good. The auto labelling is a nice touch because I often forget stuff.

crates/astria-proto/src/lib.rs
joroshiba: It is kinda orthogonal, but related. 

I exclude generated lines from line changes but it wasn't clear that's what was happening. This increases that clarity.

PR #200, opened 2023-07-27
ci: cache container image layers in ghcr

joroshiba:
Using the GHCR registry cache speeds up docker builds, cutting by about 1/3 with the default github runners. 

By my testing 30m -> 8m for longest build.


steezeburger (approved):

SuperFluffy (approved):
Very nice. This means that our use of `cargo-chef` finally makes sense.

PR #201, opened 2023-07-27
sequencer: tx nonce must match account nonce

joroshiba:
This is a breaking change.

The ethereum whitepaper states that on a transaction:

> nonce: A scalar value equal to the number of transactions sent by the sender

And later:

> the transaction nonce is valid (equivalent to the sender account’s current nonce);

As best I can tell cosmos chains treat [sequences similarly](https://github.com/astriaorg/metro/blob/d12705639f68d5e9d0ec6071748a0a0ec36ec7df/pkg/builder/builder.go#L116), and I don't believe we should differ from these implementations.

### DIFF
--- a/crates/astria-sequencer/src/transaction/unsigned.rs
+++ b/crates/astria-sequencer/src/transaction/unsigned.rs
@@ -116,7 +116,10 @@ impl ActionHandler for Unsigned {
         // Nonce should be equal to the number of executed transactions before this tx.
         // First tx has nonce 0.
         let curr_nonce = state.get_account_nonce(from).await?;
-        ensure!(curr_nonce == self.nonce, "invalid nonce, tx nonce must match account nonce");
+        ensure!(
+            curr_nonce == self.nonce,
+            "invalid nonce, tx nonce must match account nonce"
+        );
 
         // do we need to make a StateDelta here so we can check the actions on the successive state?
         for action in &self.actions {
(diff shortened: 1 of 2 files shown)


noot (approved):

crates/astria-sequencer/src/transaction/unsigned.rs
```diff
@@ -113,8 +113,10 @@ impl ActionHandler for Unsigned {
         state: &S,
         from: &Address,
     ) -> Result<()> {
+        // Nonce should be equal to the number of executed transactions before this tx.
+        // First tx has nonce 0.
         let curr_nonce = state.get_account_nonce(from).await?;
-        ensure!(curr_nonce < self.nonce, "invalid nonce");
+        ensure!(curr_nonce == self.nonce, "invalid nonce");
```
SuperFluffy, line 119:
```suggestion
        ensure!(curr_nonce == self.nonce, "transaction nonce must match stored account nonce");
```
Right now the reason gets lost:
```rust
fn run() -> eyre::Result<()> {
    let x = 5;
    let y = 6;
    eyre::ensure!(x == y, "invalid bla");
    Ok(())
}

fn main() {
    if let Err(e) = run() {
        println!("{e:?}");
        // invalid bla
        //
        // Location:
        //    src/main.rs:4:5
    }
}
```

SuperFluffy:
Marked as a bug not because there is a code problem but because it deviates from conventions.

crates/astria-sequencer/src/transaction/unsigned.rs
joroshiba: Yeah, I'm not sure how much this is intended behaivior or not. I know that when you submit an invalid sequence to a Cosmos chain you don't get any useful feedback, just that it's invalid.

SuperFluffy: Yeah, this is more for a user not having to dig into sequencer when this error happens. `invalid nonce` reads a bit like `error occured`. So like - yeah, great. But what exactly was the problem now?

itamarreif (approved):

PR #203, opened 2023-07-29
chore: remove `isTerminal` dependency and use `std::io::isTerminal`

### LINKED ISSUE #170: Replace is-terminal dependency by std::io::IsTerminal trait
Rust 1.70.0 includes the [`IsTerminal`](https://doc.rust-lang.org/stable/std/io/trait.IsTerminal.html) trait. We should be use that instead of the `is-terminal` dependency.

Minor downsides: this makes our minimal rust version 1.70.0.

jim4067:
resolves #170 

- add a `rust-toolchain` for MSRV.
- remove imports of `isTerminal` and replace with `std::io::isTerminal`.
- remove isTermnal in the `Cargo.toml` files.


dereksione (approved):

rust-toolchain
SuperFluffy, line 1: This is problematic with the way our gh actions are currently set up. Please remove.

SuperFluffy (changes requested):
Thank you for your contribution!

A lot of CI tests are currently failing because while this PR demands rust version of 1.70 stable some jobs either require nightly (running rustfmt) or have a pinned rust version of 1.69 (the docker builds).

Can you please also update the docker file and (instead of pinning the toolchain), add a `rust-version` entry to the `[package]` section of the affected binary crates?

jim4067:
I think for the rust workflow clippy and rustfmt components could be added.

Thanks for explaining why docker is also failing.
If explicitly specifying the rust version fixes this, where can I make the necessary changes?

SuperFluffy:
> I think for the rust workflow clippy and rustfmt components could be added.
> 
> Thanks for explaining why docker is also failing. If explicitly specifying the rust version fixes this, where can I make the necessary changes?

You need to update `.github/workflows/rust.yml` and `containerfiles/Dockerfile`.

SuperFluffy:
superseded by #271

PR #204, opened 2023-07-31
conductor: add block queue to executor

sambukowski:


crates/astria-conductor/src/executor.rs
```diff
@@ -114,6 +118,12 @@ struct Executor<C> {
     /// so that we can mark the block as final on the execution layer when
     /// we receive a finalized sequencer block.
     sequencer_hash_to_execution_hash: HashMap<Vec<u8>, Vec<u8>>,
+
+    /// most recently executed sequencer block hash
+    most_recently_executed_block_hash: Hash,
```
noot, line 123: shouldn't be a tendermint hash as it isn't a tendermint hash, it's a rollup block hash (which may not use the same algorithm) 

you can replace `execution_state` with this

crates/astria-conductor/src/executor.rs
```diff
@@ -114,6 +118,12 @@ struct Executor<C> {
...
+
+    /// block queue for blocks that have been recieved but their parent has not been executed yet
+    block_queue: PriorityQueue<SequencerBlockData, u64>,
```
noot, line 126: is the `u64` here the priority?

noot: also how does the priority work? is the lowest number popped first or no?

sambukowski: Yup. I'm just using the block height for this.

crates/astria-conductor/src/executor.rs
```diff
@@ -221,9 +250,15 @@ impl<C: ExecutionClient> Executor<C> {
...
-            .call_do_block(prev_block_hash, txs, Some(timestamp))
+            .call_do_block(prev_execution_block_hash, txs, Some(timestamp))
             .await?;
         self.execution_state = response.block_hash.clone();
+        self.most_recently_executed_block_hash =
+            if let Some(parent_block) = block.header.last_block_id {
+                parent_block.hash
+            } else {
+                Hash::default()
+            };
```
noot, line 261: shouldn't this be `self.most_recently_executed_block_hash = block.block_hash`?

crates/astria-conductor/src/executor.rs
sambukowski: and yes the lowest block in the queue will get popped first.
see the `try_execute_queue` function. the queue sorts highest to lowest so i'm actually popping the last item

crates/astria-conductor/src/executor.rs
```diff
@@ -196,6 +213,18 @@ impl<C: ExecutionClient> Executor<C> {
             );
             return Ok(Some(execution_hash.clone()));
         }
+        // if the block that we just recieved is not the next block in the sequence to be executed,
+        if let Some(parent_block) = block.header.last_block_id {
```
noot, line 217: I think we need to track the parent hash of the `most_recently_executed_block_hash` as well, since we are allowed to execute a block that is the same number as the most recently executed block (in the case of a 1-block fork for example)

crates/astria-conductor/src/executor.rs
sambukowski: yes.

crates/astria-conductor/src/executor.rs
noot: maybe instead of storing just `most_recently_executed_block_hash` store the whole most recently executed sequencer header?

crates/astria-conductor/src/executor.rs
```diff
@@ -306,6 +341,34 @@ impl<C: ExecutionClient> Executor<C> {
...
+    /// # Errors
+    ///
+    /// This function returns an error if:
+    /// - The call to execute_block fails
+    async fn try_execute_queue(&mut self) -> Result<()> {
+        while let Some((block, _)) = self.block_queue.clone().into_sorted_iter().last() {
+            if block
+                .header
+                .last_block_id
+                .expect("Could not unwrap last block")
```
noot, line 358: we should probably check this before putting it into the queue and remove the `expect` here

also, `last_block_id` should always be set except in the case of the genesis block, which we shouldn't be receiving in the conductor anyways

crates/astria-sequencer-relayer/src/types.rs
```diff
@@ -167,6 +168,12 @@ pub struct SequencerBlockData {
     pub rollup_txs: HashMap<Namespace, Vec<IndexedTransaction>>,
 }
 
+impl Hash for SequencerBlockData {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        self.block_hash.hash(state);
+    }
+}
```
noot, line 175: what is this? why is it re-hashing the block hash? I assume this is for the `PriorityQueue`, but should this just return `self.block_hash`?

crates/astria-conductor/src/executor.rs
sambukowski: Good point! My intention was to only keep track of the most recent safe block as the parent hash of all blocks that could be executed. I think i'm going to rename this and change this to reflect that

crates/astria-sequencer-relayer/src/types.rs
```diff
@@ -167,6 +168,12 @@ pub struct SequencerBlockData {
     pub rollup_txs: HashMap<Namespace, Vec<IndexedTransaction>>,
 }
 
+impl Hash for SequencerBlockData {
```
SuperFluffy, line 171: This violates the assumptions of the hash trait: https://doc.rust-lang.org/std/hash/trait.Hash.html#implementing-hash

This is also noted in the docs for priority queue:  https://docs.rs/priority-queue/latest/priority_queue/

crates/astria-conductor/src/executor.rs
```diff
@@ -114,6 +118,12 @@ struct Executor<C> {
     /// so that we can mark the block as final on the execution layer when
     /// we receive a finalized sequencer block.
     sequencer_hash_to_execution_hash: HashMap<Vec<u8>, Vec<u8>>,
+
+    /// most recently executed sequencer block hash
+    last_safe_block_hash: Hash,
+
+    /// block queue for blocks that have been recieved but their parent has not been executed yet
+    block_queue: PriorityQueue<SequencerBlockData, u64>,
```
SuperFluffy, line 126: Can we use the tendermint height instead of the u64 here?

SuperFluffy (changes requested):
Quick feedback because of time:

The implementation of `Hash` must be changed. Left references in the comment.

crates/astria-conductor/src/executor.rs
```diff
@@ -114,6 +118,12 @@ struct Executor<C> {
     /// so that we can mark the block as final on the execution layer when
     /// we receive a finalized sequencer block.
     sequencer_hash_to_execution_hash: HashMap<Vec<u8>, Vec<u8>>,
+
+    /// most recently executed sequencer block hash
+    last_safe_block_hash: Hash,
```
noot, line 123: this block isn't necessarily safe (unless the comment needs to be updated?)

crates/astria-conductor/src/executor.rs
```diff
@@ -206,24 +241,24 @@ impl<C: ExecutionClient> Executor<C> {
...
             .wrap_err("failed parsing str as protobuf timestamp")?;
 
+        // send the data in the sequencer block to the execution client
         let response = self
             .execution_rpc_client
-            .call_do_block(prev_block_hash, txs, Some(timestamp))
+            .call_do_block(prev_execution_block_hash, txs, Some(timestamp))
             .await?;
         self.execution_state = response.block_hash.clone();
+        self.last_safe_block_hash = Hash::try_from(block.block_hash.clone()).unwrap();
```
noot, line 261: should this be the block's parent hash now, since the just executed block may not be safe? this will probably get less confusing once #193 is done also

crates/astria-sequencer-relayer/src/types.rs
sambukowski: Updated.

crates/astria-conductor/src/executor.rs
sambukowski: Updated.

crates/astria-conductor/src/executor.rs
sambukowski: the intention is for this to be the most recent safe block. i'll update the comment

crates/astria-conductor/src/executor.rs
sambukowski: yes this should be the parent hash

crates/astria-conductor/src/executor.rs
```diff
@@ -196,6 +218,19 @@ impl<C: ExecutionClient> Executor<C> {
             );
             return Ok(Some(execution_hash.clone()));
         }
+        // if the block that we just recieved is not the next block in the sequence to be executed,
+        // add it to the block queue
+        if let Some(parent_block) = block.header.last_block_id {
```
joroshiba, line 223: Should we refactor a bit here? 

The `execute_block` logic feels pretty fat. We first early escape if not ready to execute, then process, and execute. There's an oddity to `execute_block` adding to the queue and `execute_queue` calling back here. Duplication of logic in validating block is ready to execute which if done incorrectly, could result in a the cycle of adding block back onto queue and then never executing. Done right, no issues but two places to mess up increases chances.

My thinking is that upon receiving a new block we should:
- process it down to what needs to be executed, fail fast log if issues here
- push the processed block to the queue (minimal data in the queue)

There can then be a separate async loop which checks the top of the queue and sees if it is ready to be executed, and executes. This can run pretty quickly, just iterating on the process of executing from queue.

cc @SuperFluffy for thoughts

crates/astria-sequencer-relayer/Cargo.toml
```diff
@@ -34,16 +35,18 @@ tendermint-config = { workspace = true }
...
 astria-gossipnet = { path = "../astria-gossipnet" }
 astria-proto = { path = "../astria-proto" }
 astria-sequencer = { path = "../astria-sequencer" }
 
 [dev-dependencies]
-astria-celestia-jsonrpc-client = { path = "../astria-celestia-jsonrpc-client", features = ["server"] }
-rand_core = { version = "0.6", features = [ "getrandom" ] }
+astria-celestia-jsonrpc-client = { path = "../astria-celestia-jsonrpc-client", features = [
+    "server",
+] }
```
joroshiba, line 48: nit: random formatting changes

SuperFluffy: @joroshiba #212

crates/astria-conductor/src/executor.rs
sambukowski: i moved the queue execution logic into `execute_block`. this is messy atm but I made #219 to clean up `execdute_block`.

crates/astria-conductor/src/executor.rs
sambukowski: i had to change this to the most recently executed sequencer block hash. i think we should be able to have this be more clear once #205 get's merged.

crates/astria-conductor/src/executor.rs
sambukowski: similar to my comment above, I had to leave this for now but will reevaluate after #205 is merged

crates/astria-conductor/src/executor.rs
sambukowski: made #220 to take a closer look as this

crates/astria-conductor/src/executor.rs
sambukowski: made #220 to take a closer look as this

crates/astria-conductor/src/executor.rs
```diff
@@ -114,6 +119,13 @@ struct Executor<C> {
     /// so that we can mark the block as final on the execution layer when
     /// we receive a finalized sequencer block.
     sequencer_hash_to_execution_hash: HashMap<Vec<u8>, Vec<u8>>,
+
+    /// most recently executed sequencer block hash
+    // TODO pending (GHI-205): reevaluate this after 205 gets merged
```
SuperFluffy, line 124: Please put the actual URL - easier to find.

crates/astria-conductor/src/executor.rs
```diff
@@ -114,6 +119,13 @@ struct Executor<C> {
...
+    last_executed_seq_block_hash: Hash,
+
+    /// block queue for blocks that have been recieved but their parent has not been executed yet
+    block_queue: PriorityQueue<SequencerBlockData, Height>,
```
SuperFluffy, line 128: What's the purpose of the priority queue here? If I understand this correctly, the goal is to execute the blocks by their height? Could you use just a `BTreeMap` instead that is self-ordering? Or an `IndexMap` followed by `sort_keys`?

I would prefer if we didn't have to derive/implement Hash on `SequencerBlockData` if there isn't a good reason to.

crates/astria-conductor/src/executor.rs
```diff
@@ -180,13 +194,18 @@ impl<C: ExecutionClient> Executor<C> {
...
     }
 
-    /// checks for relevant transactions in the SequencerBlock and attempts
-    /// to execute them via the execution service function DoBlock.
-    /// if there are relevant transactions that successfully execute,
-    /// it returns the resulting execution block hash.
-    /// if the block has already been executed, it returns the previously-computed
-    /// execution block hash.
-    /// if there are no relevant transactions in the SequencerBlock, it returns None.
+    /// This function takes a given sequencer block and sends it to the execution client.
```
SuperFluffy, line 197: Add a an empty doc comment under the first line. The first line is taken as a summary in the generated rust docs:
```suggestion
    /// This function takes a given sequencer block and sends it to the execution client.
    ///
```

crates/astria-conductor/src/executor.rs
```diff
@@ -180,13 +194,18 @@ impl<C: ExecutionClient> Executor<C> {
...
+    /// execution service's DoBlock function. If there are relevant transactions that
+    /// successfully execute, it returns the resulting execution block hash. If the block has
+    /// already been executed, it returns the previously-computed execution block hash. If there
+    /// are no relevant transactions in the SequencerBlock, it returns None.
+    ///
+    /// # Errors
+    ///
+    /// This function returns an error if:
+    /// - the call to the execution service's DoBlock function fails
+    // TODO (GHI-219): refactor execute_block
```
SuperFluffy, line 208: Please use the actual URL so that one doesn't have to go hunting for issue link.

crates/astria-conductor/src/executor.rs
SuperFluffy: I agree, this is a very complicated function and I don't quite grok what's going on here. Let's refactor it here instead of leaving it to a followup.

crates/astria-conductor/src/executor.rs
```diff
@@ -219,11 +252,64 @@ impl<C: ExecutionClient> Executor<C> {
...
 
+        // send the data in the sequencer block to the execution client
         let response = self
             .execution_rpc_client
-            .call_do_block(prev_block_hash, txs, Some(timestamp))
+            .call_do_block(prev_execution_block_hash, txs, Some(timestamp))
             .await?;
         self.execution_state = response.block_hash.clone();
+        // TODO pending (GHI-205): reevaluate this after 205 gets merged
+        self.last_executed_seq_block_hash = Hash::try_from(block.block_hash.clone()).unwrap();
```
SuperFluffy, line 262: why unwrap?

crates/astria-conductor/src/executor.rs
```diff
@@ -219,11 +252,64 @@ impl<C: ExecutionClient> Executor<C> {
...
+        // ======
+
+        // TOTO (GHI-219): queue execution funcitonality should be moved to its own function
+        if !self.block_queue.is_empty() {
+            // try executing all blocks in the block queue that have been recieved
+            while let Some((qblock, _)) = self.block_queue.clone().into_sorted_iter().last() {
+                if qblock
+                    .header
+                    .last_block_id
+                    .expect("Could not unwrap last block")
```
SuperFluffy, line 272: Why is `expect` here ok? Is it an invariant that `last_block_id` must be set?

crates/astria-sequencer-relayer/src/types.rs
```diff
@@ -167,6 +169,24 @@ pub struct SequencerBlockData {
...
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
```
SuperFluffy, line 173: If you handroll this (as opposed to deriving it), please provide a test that asserts that symmetry between `PartialEq` and `Hash` hold.

SuperFluffy (changes requested):
The `execute_block` function is getting quite fat and I have a hard time understanding it. I believe (but don't know) that `PriorityQueue` does not provide any advantage over just using a `BTreeMap`, but maybe that will become clearer if `execute_block` is refactored.

crates/astria-conductor/src/executor.rs
sambukowski: The `Hash::try_from` returns an Result. I guess I could an `if let` here instead.

crates/astria-conductor/src/executor.rs
sambukowski: Will do!

crates/astria-conductor/src/executor.rs
sambukowski: I'll look into swapping out the priority queue for one of those.

crates/astria-conductor/src/executor.rs
sambukowski: i made a `is_next_block()` function that checks if the id exists and must be passed before adding to the queue

sambukowski:
Closing this because #284 is replacing it.
This impl was broken when a few other PRs were merged into main.

PR #206, opened 2023-08-01
feat(sequencer): implement fees for actions

### LINKED ISSUE #66: sequencer: implement fee payments
transactions are currently free and don't cost anything. however, transactions should pay some fee. 

requirements:
- determine fee that `Transfer` and `Secondary` txs should take 
- `Transfer` should be fixed amount
- `Secondary` should be based on tx size
- implement checking that sender has enough balance to pay for the fee
- implement fee deduction at execution time

related to but not necessarily blocked by #64  and #65

noot:
- implement fees for each action
- note that the actual numbers are arbitrary right now
- question: should `sequence::Action.chain_id` be a fixed size (ie. hash)? if it's a variable size vector we also need to somehow add a fee for that, which seems not ideal.
- question: should `sequence::Action` also have a flat fee (added to the data size fee)? for example, someone can send a `sequence::Action` with `data.len() == 0` which will result in a free tx
- question: should a transaction itself have a fee? for example, someone can send empty transactions which have no actions in them, yet it will still take up a bit of block space.

closes #66


sambukowski (dismissed):

joroshiba:
Some thoughts on the opened questions:

> question: should sequence::Action.chain_id be a fixed size (ie. hash)? if it's a variable size vector we also need to somehow add a fee for that, which seems not ideal.

Yeah I think we should make this of fixed size, noting that there are specs for Celestia namespaces which we use this for as well. 

> question: should sequence::Action also have a flat fee (added to the data size fee)? for example, someone can send a sequence::Action with data.len() == 0 which will result in a free tx

Is there a good reason for this to be a valid tx?

> question: should a transaction itself have a fee? for example, someone can send empty transactions which have no actions in them, yet it will still take up a bit of block space.

Same as above, what is the valid reason for doing this? Feels purely adversarial?

crates/astria-sequencer/src/app.rs
joroshiba, line 1: Should we have tests to make sure txs fail if there is inadequate funding for the fee?

joroshiba (changes requested):

crates/astria-sequencer/src/accounts/types.rs
```diff
@@ -89,7 +89,7 @@ impl std::fmt::Display for Address {
     Ord,
     Debug,
 )]
-pub struct Balance(u128);
+pub struct Balance(pub(crate) u128);
```
itamarreif, line 92: why is the inner value `pub(crate)`?

noot: so that we can declare constants like `pub(crate) const TRANSFER_FEE: Balance = Balance(12);` and `pub(crate) const SEQUENCE_ACTION_FEE_PER_BYTE: Balance = Balance(1);`

noot:
@joroshiba updated to disallow empty `SequencerAction`s and transactions without actions, also placed a maximum length on `SequencerAction.chain_id` for now. also added a unit test for insufficient fees for tx

crates/astria-sequencer/src/sequence/action.rs
```diff
@@ -27,6 +27,8 @@ use crate::{
 /// Fee charged for a sequence `Action` per byte of `data` included.
 pub(crate) const SEQUENCE_ACTION_FEE_PER_BYTE: Balance = Balance(1);
 
+const MAX_CHAIN_ID_LENGTH: usize = 32;
```
joroshiba, line 30: Should we make this match the max celestia namespace? See [here](https://github.com/astriaorg/astria/blob/main/crates/astria-celestia-jsonrpc-client/src/rpc_impl/blob.rs#L42C54-L42C54)

crates/astria-sequencer/src/sequence/action.rs
```diff
@@ -88,6 +90,25 @@ impl ActionHandler for Action {
...
+            !self.chain_id.is_empty(),
+            "cannot have empty chain ID for sequence action",
+        );
+        ensure!(
+            self.chain_id.len() <= MAX_CHAIN_ID_LENGTH,
+            "chain ID cannot be longer than {} bytes",
+            MAX_CHAIN_ID_LENGTH,
+        );
+
+        // TODO: do we want to place a maximum on the size of the data?
```
joroshiba, line 104: I think this is fine to leave as open TODO, but lets make issue and tag with number here

crates/astria-sequencer/src/sequence/action.rs
noot: depends if we want to start using the chain_id here as the namespace directly, currently we're hashing the chain_id and taking the first 10 bytes. I wanted to keep them separate so that chain_id is human readable while namespace might not be, but open to changing it, it's somewhat arbitrary

crates/astria-sequencer/src/sequence/action.rs
noot: done

crates/astria-sequencer/src/accounts/types.rs
SuperFluffy: @itamarreif No, because trait methods are not currently usable in const expressions.

crates/astria-sequencer/src/app.rs
```diff
@@ -400,12 +404,48 @@ mod test {
...
+        let tx = Unsigned {
+            nonce: Nonce::from(0),
+            actions: vec![Action::TransferAction(Transfer::new(bob.clone(), value))],
+        };
+        let signed_tx = tx.into_signed(&keypair);
+        let bytes = signed_tx.to_proto().encode_to_vec();
+        let res = app
+            .deliver_tx(&bytes)
+            .await
+            .err()
```
SuperFluffy, line 442:
```suggestion
            .unwrap_err()
```
instead of `err().unwrap()`

crates/astria-sequencer/src/app.rs
```diff
@@ -400,12 +404,48 @@ mod test {
...
+            .unwrap()
+            .root_cause()
+            .to_string();
+        assert!(res.contains("insufficient funds"));
```
SuperFluffy, line 446: Now that we are matching on errors I think we should start transitioning parts of this lib to typed errors.

crates/astria-sequencer/src/app.rs
```diff
@@ -427,11 +467,14 @@ mod test {
         let alice_signing_key = SigningKey::from(alice_secret_bytes);
         let alice = Address::from_verification_key(&alice_signing_key.verification_key());
 
+        let data = b"hello world".to_vec();
+        let fee = SEQUENCE_ACTION_FEE_PER_BYTE * data.len() as u128;
```
SuperFluffy, line 471: Can you add a helper function that can calculate this explicitly?

crates/astria-sequencer/src/sequence/action.rs
```diff
@@ -50,4 +73,58 @@ impl Action {
...
+    async fn check_stateful<S: StateReadExt + 'static>(
+        &self,
+        state: &S,
+        from: &Address,
+    ) -> Result<()> {
+        let curr_balance = state
+            .get_account_balance(from)
+            .await
+            .context("failed getting `from` account balance")?;
+        let fee = SEQUENCE_ACTION_FEE_PER_BYTE * self.data.len() as u128;
```
SuperFluffy, line 86: I think a `fn calculate_sequencer_fee<T: AsRef<[u8]>>(data: T)` (or just `data: &[u8]`) + unit tests would be great for this.

crates/astria-sequencer/src/sequence/action.rs
```diff
@@ -1,10 +1,33 @@
...
+        types::{
+            Address,
+            Balance,
+        },
+    },
+    transaction::action_handler::ActionHandler,
+};
+
+/// Fee charged for a sequence `Action` per byte of `data` included.
+pub(crate) const SEQUENCE_ACTION_FEE_PER_BYTE: Balance = Balance(1);
```
SuperFluffy, line 28: See my comment re `calculate_sequencer_fee`: I think I'd prefer to make this non-pub(crate) and instead have a utility that calculates the fee.

Also, calling `Balance` a `Fee` is increasingly confusing.

crates/astria-sequencer/src/accounts/action.rs
```diff
@@ -73,7 +76,10 @@ impl ActionHandler for Transfer {
             .get_account_balance(from)
             .await
             .context("failed getting `from` account balance")?;
-        ensure!(curr_balance >= self.amount, "insufficient funds");
+        ensure!(
```
SuperFluffy, line 79: Similar to other comments I left - I think it would be nice to start using more typed errors.

crates/astria-sequencer/src/sequence/action.rs
```diff
@@ -50,4 +73,58 @@ impl Action {
...
+    }
+
+    #[instrument(
+        skip_all,
+        fields(
+            from = from.to_string(),
+        )
+    )]
+    async fn execute<S: StateWriteExt>(&self, state: &mut S, from: &Address) -> Result<()> {
+        let fee = SEQUENCE_ACTION_FEE_PER_BYTE * self.data.len() as u128;
```
SuperFluffy, line 120: Similar to my comment elsewhere: a utility method + unit tests for it would be preferable over calculating this inline.

crates/astria-sequencer/src/sequence/action.rs
```diff
@@ -50,4 +73,58 @@ impl Action {
...
+        Ok(())
+    }
+
+    fn check_stateless(&self) -> Result<()> {
+        ensure!(
+            !self.chain_id.is_empty(),
+            "cannot have empty chain ID for sequence action",
+        );
+        ensure!(
+            self.chain_id.len() <= MAX_CHAIN_ID_LENGTH,
```
SuperFluffy, line 99: IMO this would also make for a nice typed error.

SuperFluffy (approved):
Approving with comments.

In general surfacing the way fees and balances are calculated would be great instead of doing this implicitly/inline as part of the balance updates.

I would also forego convenience (using ops like `*`, `+` and `-`) in favour of checked operations.

And finally I think there are are more and more places where typed errors would be great (especially now that we are increasingly asserting on them).

I am not quite sure about test organizations. I am not a fan of the `app` module asserting on balances that are calculated inside `accounts`. It's getting insight into constants that it shouldn't necessarily get insight into (at least in the form of constants).

joroshiba (approved):

crates/astria-sequencer/src/accounts/action.rs
```diff
@@ -95,7 +101,7 @@ impl ActionHandler for Transfer {
             .await
             .context("failed getting `to` account balance")?;
         state
-            .put_account_balance(from, from_balance - self.amount)
+            .put_account_balance(from, from_balance - (self.amount + TRANSFER_FEE))
```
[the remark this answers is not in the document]
noot: like a function that just does this calculation `from_balance - (self.amount + TRANSFER_FEE)` ?

crates/astria-sequencer/src/app.rs
noot: agreed, i'll open an issue

crates/astria-sequencer/src/sequence/action.rs
noot: > Also, calling Balance a Fee is increasingly confusing.

what would be preferred? having just a uint? I think it's ok because the fee is always in terms of an account balance (ie deducting from it)

crates/astria-sequencer/src/accounts/action.rs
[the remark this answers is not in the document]
SuperFluffy: Yeah, exactly.

crates/astria-sequencer/src/sequence/action.rs
SuperFluffy: Maybe we should literally start using something like `token`?

crates/astria-sequencer/src/accounts/types.rs
```diff
@@ -143,13 +148,12 @@ impl std::ops::Sub<u128> for Balance {
...
 }
 
-impl std::ops::Mul<u128> for Balance {
-    type Output = Self;
-
-    fn mul(self, rhs: u128) -> Self::Output {
-        Self(self.0 * rhs)
-    }
-}
+// impl std::ops::Mul<u128> for Balance {
```
SuperFluffy, line 151: delete?

crates/astria-sequencer/src/sequence/action.rs
```diff
@@ -128,3 +126,7 @@ impl ActionHandler for Action {
         Ok(())
     }
 }
+
+pub(crate) fn calculate_fee(data: &[u8]) -> Option<Balance> {
```
SuperFluffy, line 130: Suggestion for tests:

1. empty data = empty fee?
2. one byte = one "balance"?
3. 10 bytes = 10 "balance"?

sambukowski (approved):

Issue #208, opened 2023-08-02
conductor: only read the data from DA for own rollup txs


joroshiba:
Blocked by #153 

Currently we we pull transaction data for all rollups to verify the sequencer blocks.

Once we have a better verification method implemented this should be updated to only pull rollup spec.

noot:
going to do this as part of #153

Issue #153, opened 2023-07-11
design and re-implement verification of transactions in a block


noot:
with #137, we are now using the rust sequencer. however, the transaction verification was removed to be done with this task.

Summary:
- when pulling transactions from DA in the conductor, we need to be able to verify that these transactions were truly the ones included in their respective block (ie no extra/missing txs). 
- the block header contains `data_hash`, which is a merkle root of all the transactions in a block.
- note that a sequencer transaction can contain multiple "actions", thus a single sequencer tx can act on multiple rollups.
- we want to separate and write each rollup's data to a different DA namespace, so the conductor doesn't need to pull the entire block's data to verify it.
- thus, we need to design a method which satisfies the above conditions, and ideally does not result in writing a lot of duplicate/extra data.

Possible solutions:
1. trivial solution is to force the conductor to pull the entire block's txs and reconstruct the merkle root to verify. not ideal.
2. write every tx that contains an action for "rollup A" to the namespace for rollup A and include a merkle proof of inclusion for each tx. not ideal, as it can result in duplicating data, for example in a tx contains data for rollup A and rollup B, then the tx will be written to 2 namepsaces. also too many merkle proofs + data there.
3. do the same duplication/sorting put on the block proposer side, inside the sequencer node, such that the txs that end up in a block are actually meta-txs, each of which contain all txs with data for a specific rollup. also can result in duplicate data, same case as above, but results in only 1 merkle proof per rollup instead of per tx.
4. do the sorting on the block proposer side, create a commitment to the batch of txs with data for a specific rollup, and put that as a special "transaction" type inside a block. then, the flow will be: 
a. conductor fetches batch of data for its rollup + the special "commitment transaction" + merkle proof of commitment tx inside `data_hash`
b. re-calculates and verifies the commitment using its batch of data 
c. verifies that the commitment was included inside the block's `data_hash`
this seems like a good solution, but we'll need to update the application validator logic to verify that the commitment txs are valid commitments before voting yes for that block.
5. we fork tendermint and add these commitments to the header structure, or we update the way `data_hash` is constructed - don't think we want to or need to go this route though

joroshiba:
I see this problem as being fundamentally similar to once Celestia has had to solve: how do you prove that a portion of the data is all of that portion of the data.

[...]

PR #209, opened 2023-08-02
feat(sequencer): implement generation of action tree commitment

noot:
first part of #153 (implementing solution 4. in that issue):
> 4. do the sorting on the block proposer side, create a commitment to the batch of txs with data for a specific rollup, and put that as a special "transaction" type inside a block. then, the flow will be:
> 
>     a. conductor fetches batch of data for its rollup + the special "commitment transaction" + merkle proof of commitment tx inside data_hash
>     b. re-calculates and verifies the commitment using its batch of data
>     c. verifies that the commitment was included inside the block's data_hash
this seems like a good solution, but we'll need to update the application validator logic to verify that the commitment txs are valid commitments before voting yes for that block.

- implement generation of "action tree" and put the merkle root of this tree as the first "transaction" in a block when proposing
- verify this when receiving a proposal
- remove `IndexedTransaction` in favour of just `Vec<u8>` as tx indices are no longer relevant for tx verification

next steps are #208 and then implementing the actual verification of the batch of rollup txs on the conductor side by verifying that it's included in the commitment transaction.


crates/astria-sequencer/src/proposal/commitment.rs
```diff
@@ -0,0 +1,81 @@
...
+
+/// Called when we receive a `PrepareProposal` or `ProcessProposal` consensus message.
+///
+/// In the case of `PrepareProposal`, we use this function to generate the `commitment_tx`
+/// to be placed at the start of the block.
+///
+/// In the case of `ProcessProposal`, we use this function to generate and verify the
+/// `commitment_tx` expected at the start of the block.
+///
+/// This function sorts the block's `secondary::Action`s contained within the transactions
```
SuperFluffy, line 21: The use of `secondary` is confusing because we don't use it anywhere else. Did you mean `sequence::Action`?

crates/astria-sequencer/src/service/consensus.rs
```diff
@@ -80,44 +83,40 @@ impl Consensus {
...
-                        // Error: https://github.com/dtolnay/anyhow/issues/66
-                        .map_err(|e| anyhow!(e))
-                        .context("failed to begin block")?,
+            ConsensusRequest::ProcessProposal(process_proposal) => {
+                ConsensusResponse::ProcessProposal(
+                    match Self::process_proposal(&process_proposal) {
+                        Ok(()) => response::ProcessProposal::Accept,
+                        Err(e) => {
+                            warn!(error = ?e, "rejecting proposal");
+                            response::ProcessProposal::Reject
```
SuperFluffy, line 97: Meh, sucks that there is no way to communicate why a proposal was rejected.

crates/astria-sequencer/src/service/consensus.rs
```diff
@@ -140,11 +139,42 @@ impl Consensus {
         Ok(response::InitChain::default())
     }
 
+    fn prepare_proposal(
```
SuperFluffy, line 142: Can you add comments what exactly this does? I see that it generates the tx commitment, but I don't understand at first glance why it's prepended.

crates/astria-sequencer/src/service/consensus.rs
```diff
@@ -189,11 +219,166 @@ impl Consensus {
...
+            hash: Hash::default(),
+            height: 1u32.into(),
+            next_validators_hash: Hash::default(),
+            time: Time::now(),
+            proposer_address: Id::from_str("0CDA3F47EF3C4906693B170EF650EB968C5F4B2C").unwrap(),
+        }
+    }
+
+    #[test]
+    fn prepare_proposal() {
```
SuperFluffy, line 281: Is this test useful? All it's doing is running `generate_transaction_commitment` in- and outside of `Consensus::prepare_proposal`, so it's only asserting that it's indeed following the individual steps.

Alternative suggestion: run a proposal through `prepare_proposal` and then through `process_proposal`, because these must always match.

SuperFluffy (approved):
It took me a bit to grok the general logic of this, but the implementation makes sense.

I have some comments re tests - I would prefer to have a sort of "roundtrip" test that asserts that `prepare_proposal` followed by `process_proposal` are sound.

Also, I would like to have a test where one can see that it indeed skips all non-sequence-action txs.

Also just some general comments re code organization. I think the `proposal` module shouldn't be lone-standing because it's so intimately tied to `consensus`.

Finally can you add a blurb on why we could get rid of `IndexedTransaction` to the PR/commit message?

crates/astria-sequencer/src/proposal/commitment.rs
```diff
@@ -0,0 +1,81 @@
...
+/// TODO: duplicate in `astria-sequencer-relayer/src/types.rs`
+fn get_namespace(bytes: &[u8]) -> [u8; 10] {
+    use sha2::Digest as _;
+    let mut hasher = sha2::Sha256::new();
+    hasher.update(bytes);
+    let result = hasher.finalize();
+    result[0..10].to_owned().try_into().unwrap()
+}
+
+/// Sorts the `sequence::Action`s within the transactions by their `chain_id`.
```
joroshiba, line 63: As @SuperFluffy notes this is best defined as grouping I think. (although it is also ordered)

crates/astria-sequencer/src/proposal/commitment.rs
```diff
@@ -0,0 +1,81 @@
...
+fn sort_txs_by_chain_id(txs: &[Signed]) -> BTreeMap<Vec<u8>, Vec<Vec<u8>>> {
+    let mut rollup_txs = BTreeMap::new();
+
+    for tx in txs.iter() {
+        tx.transaction().actions().iter().for_each(|action| {
+            if let Some(action) = action.as_sequence() {
+                let txs = rollup_txs
+                    .entry(action.chain_id().to_vec())
+                    .or_insert(vec![]);
+                txs.push(action.data().to_vec());
```
joroshiba, line 75:
```suggestion
                let chain_txs = rollup_txs
                    .entry(action.chain_id().to_vec())
                    .or_insert(vec![]);
                chain_txs.push(action.data().to_vec());
```

txs is overlapping with parent, a bit confusing

crates/astria-sequencer/src/proposal/commitment.rs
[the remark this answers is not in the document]
joroshiba: Consensus exists to process ABCI2 Consensus service requests. I think this is correct to have outside of that.

If we eventually used some driver API that is not ABCI, this is still relevant.

crates/astria-sequencer/src/proposal/commitment.rs
```diff
@@ -0,0 +1,81 @@
...
+pub(crate) fn generate_transaction_commitment(txs_bytes: &[Bytes]) -> Result<[u8; 32]> {
+    let txs = txs_bytes
+        .iter()
+        .map(|tx_bytes| Signed::try_from_slice(tx_bytes))
+        .collect::<Result<Vec<Signed>, Error>>()
+        .context("failed to deserialize transactions")?;
+
+    let chain_id_to_txs = sort_txs_by_chain_id(&txs);
+
+    let mut leaves: Vec<Vec<u8>> = vec![];
```
joroshiba, line 39: Could we get a comment above this about what the leaves are structurally (ie not u8, those bytes represent something)

crates/astria-sequencer/src/service/consensus.rs
joroshiba: This is by design of ABCI2 (https://github.com/cometbft/cometbft/blob/main/spec/abci/abci++_methods.md#processproposal). 

This is part of the decentralized consensus process, if running an honest validator there should be agreement and your local node will have context on why declined.

crates/astria-sequencer/src/service/consensus.rs
```diff
@@ -140,11 +139,42 @@ impl Consensus {
...
+
+        Ok(())
+    }
+
     #[instrument(skip(self))]
     async fn begin_block(
         &mut self,
         begin_block: request::BeginBlock,
-    ) -> Result<response::BeginBlock, BoxError> {
+    ) -> anyhow::Result<response::BeginBlock> {
```
joroshiba, line 177: These changes seem to be unrelated?

crates/astria-sequencer/src/service/consensus.rs
```diff
@@ -189,11 +219,166 @@ impl Consensus {
...
+        let txs = vec![tx_bytes.clone().into()];
+        let action_commitment = generate_transaction_commitment(&txs).unwrap();
+
+        let process_proposal =
+            new_process_proposal_request(vec![action_commitment.to_vec().into(), tx_bytes.into()]);
+        Consensus::process_proposal(&process_proposal).unwrap();
+    }
+
+    #[test]
+    fn process_proposal_fail() {
```
joroshiba, line 329: nit: should this be 3 tests?

crates/astria-sequencer/src/proposal/commitment.rs
joroshiba: Yeah in metro they were called secondary transactions, i think naming just got confused here.

crates/astria-sequencer/src/service/consensus.rs
joroshiba: I think we want to do a snapshot test here. That is we define the input and compare against an expected output which is saved.

If we change commitment generating it will break and either be on purpose or not, but it's always a breaking change. We need to know if the commitment generating logic changes though so it's a useful test.

We can wrap this manually or use a framework and have proper snapshots [insta for rust](https://docs.rs/insta/latest/insta/)). 

I think the roundtrip is also something we should test, because they do need to match.

joroshiba (changes requested):

joroshiba:
When this is all complete, I'd like to have a `spec` file about how it works and how to use it to verify data.

crates/astria-sequencer/src/service/consensus.rs
noot: I updated this test to also send the resulting proposal through `handle_process_proposal` as well and make sure it passes

crates/astria-sequencer/src/service/consensus.rs
noot: yeah it's just unifying the signature for all these methods

noot:
@joroshiba added a diagram w/ text to `specs/`, let me know what you think, I can also break the text out into the .md if that's easier to read

specs/assets/sequencer_inclusion_proofs.png
SuperFluffy, line 1: Not a fan of adding binary files to git, especially very text heavy ones.

This would be best as an actually typed markdown, with the graphs either as ASCII art or mermaid diagrams.

Also, putting this text as a module doc on `commitment` would be nice.

crates/astria-sequencer/src/service/consensus.rs
```diff
@@ -80,44 +83,40 @@ impl Consensus {
...
-                ConsensusResponse::PrepareProposal(response::PrepareProposal {
-                    txs: prepare_proposal.txs,
-                })
-            }
-            ConsensusRequest::ProcessProposal(_process_proposal) => {
-                // TODO: handle this
-                ConsensusResponse::ProcessProposal(response::ProcessProposal::Accept)
+                ConsensusResponse::PrepareProposal(
+                    handle_prepare_proposal(prepare_proposal)
+                        .context("failed to prepare proposal")?,
```
SuperFluffy, line 88: What happens if this returns with an error? I tried tracing the error back:

1. The loop inside `Consensus::run` creates a `warn!` event, and then sends the `Result<ConsensusResponse, BoxError>` back to the tower-abci server.
2. `tower-abci::Server` pushes the consensus `call` futures into a `FuturesUnordered`: https://docs.rs/tower-abci/latest/src/tower_abci/v037/server.rs.html#207
3. `tower-abci::Server` polls `FuturesUnordered` until it sees the consensus response https://docs.rs/tower-abci/latest/src/tower_abci/v037/server.rs.html#244
    a. the `response?.into()` lines tells me that it short circuits and just takes down the entire server if consensus encounters an issue?

This seems harsh. Reading `CometBFT` docs on [`PrepareProposal`](https://docs.cometbft.com/v0.37/spec/abci/abci++_methods#parameters-and-types) (scroll down a bit) I found this:
> If the Application considers that tx should not be proposed in this block, e.g., there are other transactions with higher priority, then it should not include it in ResponsePrepareProposal.txs. However, this will not remove tx from the mempool.

So wouldn't it be ok to just returns a `ConsensusResponse` with `txs` left empty?

noot: originally I thought this would be ok, as txs in the tendermint mempool go through `check_tx` first which will return an error code if the bytes fail to deserialized into `Signed`, so nothing passed from the mempool to the app should be un-deserializable. however I agree this does seem too dramatic in case somehow an error does occur; I updated it to ignore any tx bytes that aren't deserializable into `Signed`, so there is now no error returned.

crates/astria-sequencer/src/service/consensus.rs
```diff
@@ -189,11 +185,213 @@ impl Consensus {
...
+        generate_sequence_actions_commitment(&process_proposal.txs[1..]);
+    ensure!(
+        received_action_commitment == expected_action_commitment,
+        "transaction commitment does not match expected",
+    );
+
+    Ok(())
+}
+
+#[cfg(test)]
```
joroshiba, line 238: I still think we need a snapshot test here. Concern would be that we change the commitment logic accidentally or intentionally, and the output commitment for the same input changes. 

That is either a breaking bug or a major breaking change, right?

specs/sequencer-inclusion-proofs.md
joroshiba, line 1: I think the image is fine, may be good as many images, but I also think the text should be in markdown. This could be 4 images with three blocks of text in the spec file?

crates/astria-sequencer/src/proposal/commitment.rs
```diff
@@ -0,0 +1,190 @@
...
+/// This structure can be referred to as the "action tree".
+///
+/// The leaf, which contains a commitment to every action with the same `chain_id`, is currently
+/// implemented as ( `namespace(chain_id)` || root of merkle tree of the `sequence::Action`s ).
+/// This is somewhat arbitrary, but could be useful for proof of an action within the action tree.
+pub(crate) fn generate_sequence_actions_commitment(txs_bytes: &[Bytes]) -> [u8; 32] {
+    // ignore any transactions that are not deserializable
+    let txs = txs_bytes
+        .iter()
+        .filter_map(|tx_bytes| Signed::try_from_slice(tx_bytes).ok())
```
SuperFluffy, line 29: Now I am worried about us silently dropping transactions that could not be read. In a first step, can you log them?

But more to the point, I am reading this https://docs.cometbft.com/v0.37/spec/abci/abci++_methods#prepareproposal

And it notes that applications are allowed to remove txs from a proposal and instead put an updated list into `ResponsePrepareProposal.txs`. Should we follow this strategy here?

I am not totally sure if this is the intended use, but it seems reasonable?

SuperFluffy (changes requested):
Worried about the implications of silently dropping txs that were not deserializable (or in this case, deserializable + verifiable). See my review comment about returning an updated `ResponsePrepareProposal` with the bad txs dropped.

crates/astria-sequencer/src/service/consensus.rs
noot: you mean like with test vectors (ie. these txs result in this commitment?)  yeah I can add this

SuperFluffy: @noot I believe @joroshiba refers to something https://insta.rs/ / https://crates.io/crates/insta

joroshiba: I'm fine with just having the commitment inline (depending on size), but yeah I think proper snapshots are better.

Insta looks nice. The nice thing is it's generalizable across other tests and less magic value "where this come from" inputs in the tests, outputs in snapshots. If you see a diff while reviewing it's very obviously a change to think twice about.

noot:
@joroshiba @SuperFluffy updated the spec doc, feel free to take another look!

crates/astria-sequencer/src/service/consensus.rs
noot: put a basic test vector unit test for now, but happy to change to `insta` also. does feel a bit overkill for the one test but I see how it's nice, less magic value for sure

joroshiba (approved):

crates/astria-sequencer/src/proposal/commitment.rs
```diff
@@ -0,0 +1,220 @@
...
+        leaves.iter().enumerate().for_each(|(i, leaf)| {
+            if i == 0 {
+                return;
+            }
+            assert!(leaf > &leaves[i - 1]);
+        });
+    }
+
+    #[test]
+    fn generate_sequence_actions_commitment_snapshot() {
```
SuperFluffy, line 191: Is there are a more descriptive name for this test?

SuperFluffy (approved):
Please fix the structured fields in the debug event. Otherwise looks good. Thank you for the PR

crates/astria-sequencer/src/proposal/commitment.rs
noot: I'll think of one + add a comment

Issue #27, opened 2023-05-19
move shared "read" and types code from relayer/conductor to utils crate


noot:
currently, there is a decent amount of code shared between the relayer and conductor, mainly by means of the conductor importing the relayer code. however, this shared code should live in some utils crate used by both. this will allow them to not depend on each other, and also facilitate testing within the conductor.

noot:
will do this after #99

joroshiba:
@SuperFluffy is this related to your current work?

SuperFluffy:
@joroshiba yeah, that's all the same issue essentially

noot:
closing as duplicate of #86

PR #211, opened 2023-08-03
feat(conductor): only read rollup data from our revelant namespace

### LINKED ISSUE #208: conductor: only read the data from DA for own rollup txs

noot:
- as title says, update conductor to only read rollup data from our revelant namespace
- related to #153 but tx verification is not implemented yet, this is just one PR to prep for it

closes #208


crates/astria-conductor/src/executor.rs
```diff
@@ -198,14 +214,6 @@ impl<C: ExecutionClient> Executor<C> {
         }
 
         // get transactions for our namespace
```
joroshiba, line 216: nit: remove the text about the block

crates/astria-sequencer-relayer/src/data_availability.rs
```diff
@@ -337,6 +337,66 @@ impl CelestiaClient {
...
+
+        let req = GetAllRequest {
+            height,
+            namespace_ids: vec![*rollup_namespace],
+        };
+
+        let rsp = match self.client.blob_get_all(req).await {
+            Ok(rsp) => rsp,
+            Err(e) => {
+                if let ErrorKind::Rpc(astria_celestia_jsonrpc_client::JsonRpseeError::Call(inner)) =
```
joroshiba, line 364: is there a clean way to shorten this line?

crates/astria-sequencer-relayer/src/data_availability.rs
```diff
@@ -381,27 +441,53 @@ impl CelestiaClient {
...
+                warn!("failed to deserialize rollup namespace data");
+                None
+            }
+        })
+        .collect();
 
-        // retain rollup datas that can be verified
-        rollup_datas.retain(|namespace, rollup_data| {
+    // retain rollup datas whose block hash matches the block hash of the namespaced data
+    rollup_datas.retain(|_, rollup_data| block_hash == rollup_data.data.block_hash);
```
joroshiba, line 487: Should we log something about the data which shouldn't be retained?

joroshiba (approved):
Nice clean up!

crates/astria-sequencer-relayer/src/data_availability.rs
```diff
@@ -337,6 +337,66 @@ impl CelestiaClient {
...
+        // filter out blobs; we should only be left with either zero or one rollup datas
+        let rollup_datas = filter_and_convert_rollup_data_blobs(
+            rsp.blobs,
+            &namespace_data.data.block_hash,
+            &verification_key,
+        );
+
+        // this should *not* happen; the only case where it would happen is if the sequencer-relayer
+        // posts multiple blobs with the same rollup ID for the same block (could this happen?)
+        if rollup_datas.len() > 1 {
```
SuperFluffy, line 385: If this happens something very strange is going on with celestia.

If our assumptions around what it returns are broken, we should either:

1. return the entire hashmap and let the downstream processing handle it
2. or alternatively return a big fat error here.

crates/astria-sequencer-relayer/src/data_availability.rs
```diff
@@ -337,6 +337,66 @@ impl CelestiaClient {
...
+                "found more than one rollup data for the given block hash; returning the first one"
+            );
+        }
+
+        // this case can happen if someone posts a blob to the namespace with invalid data
+        if rollup_datas.is_empty() {
+            return Ok(None);
+        }
+
+        Ok(rollup_datas.into_iter().next().map(|data| data.1.data))
```
SuperFluffy, line 397: Re my comment above: assuming something bad like this happens, I would want to understand what namespaced data was matched.

SuperFluffy (changes requested):
In general looks fine to me, but I am blocking with `request changes` because I would like to discuss `get_rollup_data` and how it treats (potentially wrong?) celestia responses.

crates/astria-sequencer-relayer/src/data_availability.rs
noot: I don't think it's necessary, there might be a lot of these considering we write 15 sequencer blocks for 1 celestia block

crates/astria-sequencer-relayer/src/data_availability.rs
```diff
@@ -337,6 +337,66 @@ impl CelestiaClient {
...
     }
 
+    /// Returns the rollup data for a given rollup namespace at a given height, if it exists.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if:
+    /// + the verification key could not be constructed from the data stored in `namespace_data`;
+    /// + the RPC to fetch the blobs failed.
+    pub async fn get_rollup_data(
```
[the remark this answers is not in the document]
noot: agreed, i have #27 open for this. do you think this should go directly into `conductor` or into a shared crate so both conductor and relayer can use it for testing?

crates/astria-conductor/src/executor.rs
```diff
@@ -95,6 +96,7 @@ struct Executor<C> {
     execution_rpc_client: C,
 
     /// Namespace ID
+    /// TODO: remove?
```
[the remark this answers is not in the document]
noot: will remove comment

crates/astria-sequencer-relayer/src/data_availability.rs
SuperFluffy: @noot can you use the utility landed with #236 to make this a bit neater?

crates/astria-sequencer-relayer/src/data_availability.rs
```diff
@@ -382,12 +385,10 @@ impl CelestiaClient {
...
         // posts multiple blobs with the same rollup ID for the same block (could this happen?)
-        if rollup_datas.len() > 1 {
-            warn!(
-                num_rollup_datas = rollup_datas.len(),
-                "found more than one rollup data for the given block hash; returning the first one"
-            );
-        }
+        ensure!(
+            rollup_datas.len() <= 1,
+            "shpuld not have more than one rollup data for the given block hash",
```
SuperFluffy, line 390:
```suggestion
            "should not have more than one rollup data for the given block hash",
```

SuperFluffy (approved):
Noticed a typo and the complex error match can be simplified, see comment here https://github.com/astriaorg/astria/pull/211#discussion_r1289227742

PR #212, opened 2023-08-04
ci: check toml formatting

SuperFluffy:
This commit adds a github workflow job to
standardize toml formatting across the
monorepo. Because contributors use different
editors (some with and some without toml
formatting built in), this should help reduce
churn.

https://github.com/tamasfe/taplo version 0.8.1
is downloaded and put into /usr/local/bin and run
with as `taplo format --check`, failing if some
toml files were not formatted.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -79,6 +79,18 @@ jobs:
         # or `cargo +nightly-2023-07-07 fmt` for specifying the exactly release.
         run: cargo fmt --all -- --check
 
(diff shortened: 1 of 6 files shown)


joroshiba (approved):
Can we add something to README about formatting and tools we use for it?

SuperFluffy:
@joroshiba 
> Can we add something to README about formatting and tools we use for it?
Added a section on formatting to the root README.

noot (approved):

PR #217, opened 2023-08-04
sequencer-client: re-export tendermint Client

SuperFluffy:
When using `HttpClient` and `WebsocketClient` one frequently wants to also use other trait methods provided by the tendermint `Client` trait. This reexports it so one does not have to explicitly add `tendermint-rpc` as a dependency.


joroshiba (commented):
Should this also include a Client file?

SuperFluffy:
@joroshiba 
> Should this also include a Client file?
Can you explain what you mean by that?

noot (approved):

joroshiba (approved):

PR #223, opened 2023-08-04
Configuration standard spec

joroshiba:
- documents configuration standard
- updates astria-composer as an example of implementing the standard

Follow up work all listed on #240 tasklist

### DIFF
--- a/crates/astria-composer/README.md
+++ b/crates/astria-composer/README.md
@@ -1,16 +1,22 @@
 # Astria Composer
 
-Composer sits between a rollup execution node and the Astria Shared Sequencer to submit transactions. As currently implmented, it follows a websocket on a Geth Node for new transactions, wraps and signs the bytes into a sequencer transaction. Composer currently supports gathering transactions from a single rollup. In the future it may support multiple rollups.
+Composer sits between a rollup execution node and the Astria Shared Sequencer to
+submit transactions. As currently implmented, it follows a websocket on a Geth
+Node for new transactions, wraps and signs the bytes into a sequencer
+transaction. Composer currently supports gathering transactions from a single
+rollup. In the future it may support multiple rollups.
 
 ## Running Composer
 
 ### Dependencies
 
-We use [just](https://just.systems/man/en/chapter_4.html) for convenient project specific commands.
+We use [just](https://just.systems/man/en/chapter_4.html) for convenient project
+specific commands.
 
 ### Configuration
 
-Composer is configured via environment variables. An example configuration can be seen in `local.env.example`.
+Composer is configured via environment variables. An example configuration can
+be seen in `local.env.example`.
 
 To copy a configuration to your `.env` file run:
 
@@ -25,7 +31,8 @@ just copy-env
 
 ### Running locally
 
-After creating a `.env` file either manually or by copying as above, `just` will load it and run locally:
+After creating a `.env` file either manually or by copying as above, `just` will
+load it and run locally:
 
 ```bash
 just run
--- a/crates/astria-composer/README.md
+++ b/crates/astria-composer/README.md
@@ -1,22 +1,20 @@
 # Astria Composer
 
-Composer sits between a rollup execution node and the Astria Shared Sequencer to
-submit transactions. As currently implmented, it follows a websocket on a Geth
-Node for new transactions, wraps and signs the bytes into a sequencer
-transaction. Composer currently supports gathering transactions from a single
-rollup. In the future it may support multiple rollups.
+Composer sits between a rollup execution node and the Astria Shared Sequencer to submit
+transactions. As currently implemented, it follows a websocket on a Geth Node for new transactions,
+wraps and signs the bytes into a sequencer transaction. Composer currently supports gathering
+transactions from a single rollup. In the future it may support multiple rollups.
 
 ## Running Composer
 
 ### Dependencies
 
-We use [just](https://just.systems/man/en/chapter_4.html) for convenient project
-specific commands.
+We use [just](https://just.systems/man/en/chapter_4.html) for convenient project specific commands.
 
 ### Configuration
 
-Composer is configured via environment variables. An example configuration can
-be seen in `local.env.example`.
+Composer is configured via environment variables. An example configuration can be seen in
+`local.env.example`.
 
 To copy a configuration to your `.env` file run:
 
@@ -31,8 +29,8 @@ just copy-env
 
 ### Running locally
 
-After creating a `.env` file either manually or by copying as above, `just` will
-load it and run locally:
+After creating a `.env` file either manually or by copying as above, `just` will load it and run
+locally:
 
 ```bash
 just run
--- a/crates/astria-composer/README.md
+++ b/crates/astria-composer/README.md
@@ -1,20 +1,22 @@
 # Astria Composer
 
-Composer sits between a rollup execution node and the Astria Shared Sequencer to submit
-transactions. As currently implemented, it follows a websocket on a Geth Node for new transactions,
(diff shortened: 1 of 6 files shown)


crates/astria-composer/README.md
```diff
@@ -0,0 +1,27 @@
+# Astria Composer
+
+Composer sits between a rollup execution node and the Astria Shared Sequencer to submit transactions. As currently implmented, it follows a websocket on a Geth Node for new transactions, wraps and signs the bytes into a sequencer transaction.
```
noot, line 3: specify the relationship (n-1, n-n) between the execution nodes and composer? from what I understand there can be many composers and each composer can handle many rollup nodes

crates/astria-composer/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Running Composer
+
+### Dependencies
+
+We use [just](https://just.systems/man/en/chapter_4.html) for convenient project specific commands
```
noot, line 9:
```suggestion
We use [just](https://just.systems/man/en/chapter_4.html) for convenient project specific commands.
```

extreme nit

crates/astria-composer/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+### Configuration
+
+Composer is configured via environment variables. An example configuration can be seen in `local.env.example`.
+
+To copy a configuration to your `.env` file run, environment will default to local:
+
+```bash
+just copy-env <ENVIRONMENT>
```
noot, line 18: can you run `just copy-env` without `<ENVIRONMENT>` also?

crates/astria-composer/local.env.example
```diff
@@ -8,7 +8,7 @@ ASTRIA_COMPOSER_API_LISTEN_ADDR="0.0.0.0:0"
 ASTRIA_COMPOSER_SEQUENCER_URL="http://127.0.0.1:26657"
 
 # Execution chain ID that we want to connect to
-ASTRIA_COMPOSER_CHAIN_ID="912559"
+ASTRIA_COMPOSER_CHAIN_ID="astria"
```
noot, line 11: is our EVM named astria now? haha

crates/astria-composer/justfile
```diff
@@ -0,0 +1,12 @@
...
+
+set dotenv-load
+set fallback
+
+default_env := 'local'
+copy-env type=default_env:
+  cp {{ type }}.env.example .env
+
+run:
+  cargo run
```
noot, line 12: should this also include copying the example env (if .env does not exist)?

specs/config-standard.md
```diff
@@ -0,0 +1,42 @@
...
+  - `.env` files are gitignored in repo
+  - Tooling
+    - README provides information on running locally, including copying of example env file
+    - Each service will have a `justfile` to maintain ease of running
+      - Loads `.env` via `set dotenv-load`
+      - `just copy-env {type}` command to ease copying `.env.example` files
+        - default to `local`
+      - `just run` as a wrapper to `cargo run` with loaded environment
+  - Shared Configuration (ie Genesis)
+    - configure via passed in path to `ASTRIA_<SERVICE>_<FILENAME>_PATH`
```
noot, line 36:
```suggestion
    - configure via passed in path to `ASTRIA_<SERVICE>_<FILENAME>_PATH`
```

what's `FILENAME` here? should it just be `GENESIS`?

specs/config-standard.md
```diff
@@ -0,0 +1,42 @@
...
+    - may have example files in repo, pointed to by matching `{ENV}.env.example`
+      - `local.genesis.json`
+- CLI Configuration
+  - Configured via environment variable OR CLI arguments w/ sensible defaults as appropriate
+    - defaults < environment variable < cli in terms of precedence
+  - Is not a wrapper to start a service
```
noot, line 42: what do you mean by this?

crates/astria-composer/local.env.example
joroshiba: It needed to be a string instead of numerical network I'd, used for the sequencer not the RPC. 

The rollup helm charts actually default to "astriachain"

crates/astria-composer/justfile
joroshiba: We could but I like the explicit need to copy the environment. 

It's a two liner vs a one liner for first run, but makes configuration an explicit step.

specs/config-standard.md
joroshiba: Cosmos CLI has commands to generate config, as well as the "start" command. 

If we have a CLI for generating scripts it should not also start the service is my view here.

specs/config-standard.md
```diff
@@ -0,0 +1,42 @@
+# Configuration Standard
+
+## Purpose
+
+Define how configuration should be implemented across all services within Astria services.
+
+## Background
+
+When running a service, it is generally [best practice to store configuration in environment variables](https://12factor.net/config), without defaults. This makes it difficult to commit environment specific config, and ensures seperation of configuration and code. It makes it easier to integrate with KMS for sercrets, and container orchestration systems are configured to enable this easily. Not having defaults ensures configuration is done with intention on a given environment.
```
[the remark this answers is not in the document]
joroshiba: [Key management system](https://en.wikipedia.org/wiki/Key_management)

specs/config-standard.md
```diff
@@ -0,0 +1,43 @@
+# Configuration Standard
+
+## Purpose
+
+Define how configuration should be implemented across all services within Astria services.
+
+## Background
+
+When running a service, it is generally [best practice to store configuration in environment variables](https://12factor.net/config), without defaults. This makes it difficult to commit environment specific config, and ensures separation of configuration and code. It makes it easier to integrate with KMS for secrets, and container orchestration systems are configured to enable this easily. Not having defaults ensures configuration is done with intention on a given environment.
```
SuperFluffy, line 9: Very-very-nit: from my old latex days I prefer to use more linebreaks linebreaks so that adding a word or fixing a typo doesn't highlight the entire line.

specs/config-standard.md
```diff
@@ -0,0 +1,43 @@
...
+
+For local development, however, this is often inconvenient. CLI args and defaults are often easier to manage while running on your local machine. Additionally, blockchains often have genesis files which need to be coordinated across a diverse set of decentralized actors. Having files for this type of configuration is often best, and the configurations may be managed via command line tools.
+
+This standard is intended to balance the best practice, with the needs of developers for ease, and some shared configuration.
+
+## Astria Standard
+
+> Note that this standard is not implemented across repo, there are tagged issues to update services to use the standard [here](https://github.com/astriaorg/astria/issues/240).
+
+- Configuration in Rust managed via [Figment](https://docs.rs/figment/latest/figment/)
```
SuperFluffy, line 19: Let's not enforce the use of a particular crate - to make reporting better we should probably be move away from figment and roll our own library on top of serde because a) we don't use most of figment, b) we have the issue of short circuiting if a config is missing so that only one gets reported, and c) the log only contains the config field but not the full name of the missing env var.

So for example if we can't find `ASTRIA_SERVICE_FOO` and `ASTRIA_SERVICE_BAR`, we are getting the very unhelpful message that `foo` is missing.

SuperFluffy (approved):
Agree with the spec. Only some minor comments.

specs/config-standard.md
joroshiba: We can update the spec when we change that. I'd like to have some standardization for now, just an established pattern this will make the rust code very similar across binaries. When we change, we can update with an established easy to swap everything pattern as well.

noot (approved):

Issue #214, opened 2023-08-04
Composer: Infrastructure for arbitrary number of collectors to stream data to composer


dereksione:
Right now, we can only collecting pending tx data from one rollup client (eth), but we want to allow any arbitrary number of rollups to be added to the pipeline, and thus the composer to be able to easily spawn up new collectors that can start streaming data to the composer pipeline.

# Subtasks

- [x] Add a generic provider wrapper type that isn't tied to eth provider implementations
- [x] Collectors should be generic over provider
- [x] Arbitrary number of collectors should be streaming data to the same outgoing channel 
- [ ] Testing suite that ensure that there are no backpressure bottlenecks

PR #226, opened 2023-08-07
Composer: architecture refactor

dereksione:
This PR: 

- Splits Searcher into three components:
  - Collector (collects transactions from multiple rollups and funnels them into one stream)
  - Bundler (bundles multiple transactions into a bundle that can be turned into one sequencer tx using MEV strategies)
  - Executor (executes bundles - passing them into block builders or to the sequencer directly)
  
 Also addresses #246


dereksione:
Tiny bit of work left in this refactor, but the core logic is ready to review

crates/astria-composer/src/bundler.rs
```diff
@@ -0,0 +1,59 @@
...
+            executor_sender: stream,
+        };
+        (new_strategy, sink)
+    }
+
+    async fn bundle_transactions(
+        rollup_txs: Vec<RollupTxExt>,
+    ) -> Result<Vec<RollupTxExt>, eyre::Error> {
+        // NOTE: Replace this with arbitrary strategy logic
+        let bundle = NoStrategy::execute(rollup_txs).await?;
```
noot, line 34: should this be a generic for this function? like `async fn bundle_transactions<S: Strategy>` and then `S::execute()`? since I'm assuming this is the part we want to be switched out by users

noot: actually this function is private anyways so this might be unnecessary, but just curious about how you see this eventually being extended for different strategies

crates/astria-composer/src/collector.rs
```diff
@@ -0,0 +1,38 @@
+use crate::ds::{RollupChainId, RollupTx, RollupTxExt, StreamingClient};
+use color_eyre::eyre::{self};
+use tokio::sync::mpsc as tokio_mpsc;
+
+pub(crate) struct Collector {
+    digestion_channel: tokio_mpsc::UnboundedSender<RollupTxExt>,
+}
+
+impl Collector {
```
noot, line 9: can you add a comment explaining what this type does?

crates/astria-composer/src/ds/types.rs
```diff
@@ -0,0 +1,9 @@
+use ethers::types::Transaction as EthersTx;
+
+#[non_exhaustive]
+pub enum RollupTx {
+    EthersTx(EthersTx),
+}
+
+pub(crate) type RollupChainId = String;
+pub type RollupTxExt = (RollupTx, RollupChainId);
```
SuperFluffy, line 9: Please only use aliases if they help legibility.

Also this should not be pub

crates/astria-composer/src/searcher.rs
```diff
@@ -1,358 +1,72 @@
...
-    submission_tasks: JoinSet<eyre::Result<()>>,
-}
+use color_eyre::eyre::{self, Context};
+use tokio::sync::watch;
 
-#[derive(Debug, Default)]
-pub(crate) struct Status {
-    eth_connected: bool,
+#[derive(Default)]
+pub struct SearcherStatus {
```
SuperFluffy, line 12: Please don't repeat the module name in the type name.

Also: not pub

crates/astria-composer/src/ds/wire_format.rs
```diff
@@ -0,0 +1,11 @@
+use ethers::types::Transaction as EthersTx;
```
SuperFluffy, line 1: Why aliases?

SuperFluffy (changes requested):
Please add more information about the reasoning that went into the architecture of this RP and how to read it. On a first glance I cannot see a) what was simply shifted around to to new modules, and b) what constitutes new functionality.

There are a lot of new types with generic type parameters and I don't know how they are supposed to be used.

Looking at the PR description, can you split this refactor into 3 different refactors to make it easier to understand what's going on?

crates/astria-composer/src/collector.rs
```diff
@@ -0,0 +1,38 @@
...
+    pub(crate) fn new() -> (Self, tokio_mpsc::UnboundedReceiver<RollupTxExt>) {
+        let (stream, sink) = tokio_mpsc::unbounded_channel::<RollupTxExt>();
+        let new_collector = Collector {
+            digestion_channel: stream,
+        };
+
+        (new_collector, sink)
+    }
+
+    pub(crate) async fn add_provider<P>(
```
itamarreif, line 19: this could also use a documentation comment

crates/astria-composer/src/collector.rs
```diff
@@ -0,0 +1,38 @@
...
+use color_eyre::eyre::{self};
+use tokio::sync::mpsc as tokio_mpsc;
+
+pub(crate) struct Collector {
+    digestion_channel: tokio_mpsc::UnboundedSender<RollupTxExt>,
+}
+
+impl Collector {
+    pub(crate) fn new() -> (Self, tokio_mpsc::UnboundedReceiver<RollupTxExt>) {
+        let (stream, sink) = tokio_mpsc::unbounded_channel::<RollupTxExt>();
```
itamarreif, line 11: why name them `stream`, `sink`? i've seen `tx` and `rx` used everywhere

crates/astria-composer/src/collector.rs
```diff
@@ -0,0 +1,38 @@
...
+        let new_collector = Collector {
+            digestion_channel: stream,
+        };
+
+        (new_collector, sink)
```
itamarreif, line 16: should we be returning the channel reader from the constructor? i think it would be nicer to have a `Collector.subscribe()` method for this and simply returning the `Collector` from the constructor.

deferring to @SuperFluffy

crates/astria-composer/src/bundler.rs
```diff
@@ -0,0 +1,59 @@
...
+impl Bundler {
+    pub(crate) fn new(
+        collector_receiver: UnboundedReceiver<RollupTxExt>,
+    ) -> (Self, UnboundedReceiver<Vec<RollupTxExt>>) {
+        let (stream, sink) = tokio_mpsc::unbounded_channel::<Vec<RollupTxExt>>();
+        let new_strategy = Self {
+            collector_receiver,
+            executor_sender: stream,
+        };
+        (new_strategy, sink)
```
itamarreif, line 27: same comment as for `Collector` - i feel like we should probably only return the object that is constructed from the constructor and add a `subscribe()` method for getting a reader to its channel

crates/astria-composer/src/bundler.rs
itamarreif, line 34: I dont think that this is the correct way to make this replaceable by others' strategies - they should instead replace the entire bundler by their own implementation for a strategy

crates/astria-composer/src/bundler.rs
```diff
@@ -0,0 +1,59 @@
...
+        rollup_txs: Vec<RollupTxExt>,
+    ) -> Result<Vec<RollupTxExt>, eyre::Error> {
+        // NOTE: Replace this with arbitrary strategy logic
+        let bundle = NoStrategy::execute(rollup_txs).await?;
+        Ok(bundle)
+    }
+
+    pub(crate) async fn start(&mut self) -> Result<(), eyre::Error> {
+        while let Some(rollup_tx) = self.collector_receiver.recv().await {
+            let sender = self.executor_sender.clone();
```
itamarreif, line 40: rust question (?) : is this cloned inside the loop so that each `rollup_tx` that's picked up from the collector's channel is processed in its own task?

crates/astria-composer/src/searcher.rs
```diff
@@ -1,358 +1,72 @@
...
-            // For now, each transaction is transmitted from a new account with nonce 0
-            let sequencer_key = ed25519_consensus::SigningKey::new(rand::thread_rng());
-            let nonce = Nonce::from(0);
-
-            // Pack into sequencer tx
-            let data = rollup_tx.rlp().to_vec();
-            let chain_id = chain_id.into_bytes();
-            let seq_action = SequencerAction::SequenceAction(SequenceAction::new(chain_id, data));
-            let unsigned_tx = UnsignedSequencerTx::new_with_actions(nonce, vec![seq_action]);
+    // FIXME: How do we start up multiple providers if we are reading from Config?
```
itamarreif, line 32: config needs to be changed in a later pr to allow for multiple rollups

crates/astria-composer/src/searcher.rs
```diff
@@ -1,358 +1,72 @@
...
+    async fn setup_providers_from_config(
+        cfg: &Config,
+    ) -> Result<Vec<(Box<dyn StreamingClient<Error = eyre::Error>>, RollupChainId)>, eyre::Error>
```
itamarreif, line 35: i dont think we need to be generic over providers at this point. this should be a later change.

crates/astria-composer/src/searcher.rs
```diff
@@ -1,358 +1,72 @@
...
+        let mut executor = Executor::new(&cfg.sequencer_url, bundler_recv_channel).await?;
+        // Set status flag to true since sequencer has started successfully if this code is executed
+        status.sequencer_connected = true;
 
-        // FIXME: ensure that we can get here
-        #[allow(unreachable_code)]
-        Ok(())
-    }
+        let (status, _) = watch::channel(status);
+        tokio::task::spawn(async move {
```
itamarreif, line 60: we should separate the constructor from the `run` functions

crates/astria-composer/src/strategy.rs
```diff
@@ -0,0 +1,22 @@
+use crate::ds::RollupTxExt;
+use async_trait::async_trait;
+
+/// The Strategy trait defines arbitrary MEV strategies executed on a list of received
+/// rollup transactions
+#[async_trait]
+pub(crate) trait Strategy{
```
itamarreif, line 7: I dont see the need for this right now - this should be a separate, self-contained pr for abstracting over the strategy

crates/astria-composer/src/ds/mod.rs
itamarreif, line 1: While I like the idea of separating the data structure implementations from the application logic, I'm personally not a fan of having a single `data structures` module for the entire crate - i think these should be moved into the specific modules to which they are relevant.

crates/astria-composer/src/searcher.rs
```diff
@@ -1,358 +1,72 @@
...
-        };
-        let wait_for_eth = self.wait_for_eth(5, Duration::from_secs(5), 2.0);
-        let wait_for_seq = self.wait_for_sequencer(5, Duration::from_secs(5), 2.0);
-        match tokio::try_join!(wait_for_eth, wait_for_seq) {
-            Ok(((), ())) => {}
-            Err(err) => return Err(err).wrap_err("failed to start searcher"),
+        let mut status = SearcherStatus::default();
+        for (provider, chain_id) in Self::setup_providers_from_config(cfg).await? {
+            collector.add_provider(provider, chain_id).await?;
+            status.active_providers += 1;
```
itamarreif, line 51: we need to call `provider.setup()` (formerly `wait_for_eth`) before incrementing the active providers count.

crates/astria-composer/src/searcher.rs
```diff
@@ -1,358 +1,72 @@
...
-            Middleware as _,
-            StreamExt as _,
-        };
-        let wait_for_eth = self.wait_for_eth(5, Duration::from_secs(5), 2.0);
-        let wait_for_seq = self.wait_for_sequencer(5, Duration::from_secs(5), 2.0);
-        match tokio::try_join!(wait_for_eth, wait_for_seq) {
-            Ok(((), ())) => {}
-            Err(err) => return Err(err).wrap_err("failed to start searcher"),
+        let mut status = SearcherStatus::default();
+        for (provider, chain_id) in Self::setup_providers_from_config(cfg).await? {
```
itamarreif, line 49: let's stick to a single provider for now - moving to multiple providers should be a separate pr

crates/astria-composer/src/ds/eth_provider.rs
```diff
@@ -0,0 +1,92 @@
...
+    ///
+    /// # Errors
+    ///
+    /// An error is returned if calling eth failed after `n_retries + 1` times.
+    #[instrument(skip_all, fields(
+        retries.max_number = n_retries,
+        retries.initial_delay = %format_duration(delay),
+        retries.exponential_factor = factor,
+    ))]
+    pub(crate) async fn setup(
```
itamarreif, line 50: why change the name of this from `wait_for_eth`? also, this should be called before incrementing the `active_providers` count in `searcher::Status` (or before setting it to `true` in the single provider case).

crates/astria-composer/src/ds/eth_provider.rs
```diff
@@ -0,0 +1,92 @@
+use std::time::Duration;
+
+use async_trait::async_trait;
+use humantime::format_duration;
+use tracing::{debug, info, instrument, warn};
+use color_eyre::eyre::{self, Context};
+use tokio::sync::mpsc as tokio_mpsc;
+
+use ethers::providers::{
+    Middleware, Provider as EthersProvider, ProviderError as EthersProviderError, StreamExt, Ws,
```
itamarreif, line 10: why do we need the type aliases here, instead of simply `Provider` and `ProviderError`? this module only implements our provider for eth

crates/astria-composer/src/ds/streaming_client.rs
```diff
@@ -0,0 +1,46 @@
...
+pub(crate) trait StreamingClient {
+    type Error;
+    async fn start_stream(
+        &self,
+        chain_id: RollupChainId,
+    ) -> Result<tokio_mpsc::UnboundedReceiver<RollupTxExt>, Self::Error>;
+}
+
+#[async_trait]
+impl<C> StreamingClient for EthersProvider<C>
```
itamarreif, line 20: why are we implementing this for the `EthersProvider` type rather than our own `EthProvider`?

crates/astria-composer/src/ds/eth_provider.rs
```diff
@@ -0,0 +1,92 @@
...
+            .wrap_err(
+                "failed to retrieve net version from eth after seferal retries",
+            )?;
+        info!(version, rpc = "net_version", "RPC was successful");
+        Ok(())
+    }
+}
+
+#[async_trait]
+impl StreamingClient for EthProvider {
```
itamarreif, line 83: i don't understand why the trait is implemented for `EthersProvider` instead of only for `EthProvider`?

crates/astria-composer/src/ds/eth_provider.rs
```diff
@@ -0,0 +1,92 @@
...
+use super::{streaming_client::StreamingClient, RollupChainId, RollupTxExt};
+
+/// A thin wrapper around [`Provider<Ws>`] to add timeouts.
+pub(crate) struct EthProvider {
+    inner: EthersProvider<Ws>,
+}
+
+// TODO(https://github.com/astriaorg/astria/issues/216): add timeouts for
+// `subscribe_full_pendings_txs` (more complex because it's a stream).
+impl EthProvider {
```
itamarreif, line 22: this name is confusing since there's an `EthersProvider` and an `EthProvider`

crates/astria-composer/src/executor.rs
```diff
@@ -0,0 +1,89 @@
...
+        let unsigned_sequencer_tx = UnsignedSequencerTx::new_with_actions(nonce, actions);
+
+        Ok(unsigned_sequencer_tx.into_signed(&sequencer_key))
+    }
+
+    async fn sign_and_submit_bundle(
+        &mut self,
+        bundle: Vec<RollupTxExt>,
+    ) -> Result<(), eyre::Error> {
+        let signed = Self::sign_bundle(bundle).await?;
```
itamarreif, line 70: this should be run in a blocking task, as serializing and signing over a potentially unbounded amount of data can become costly and would block other tasks that run on `tokio`'s non-blocking threadpool

itamarreif (changes requested):

crates/astria-composer/src/collector.rs
SuperFluffy: @itamarreif can't subscribe to an unbounded mpsc channel unfortunately, that's for watch or broadcast channels.

You indeed need to pass around the sender half of the channel and clone it where needed.

However, I concur that it would be nicer to pass in the rx part of the channel into the constructor rather than return the sender half from it. Just to stay consistent with your rest of the crate.

crates/astria-composer/src/bundler.rs
dereksione: Yeah this would make sense, but I think we should allow for cases where multiple strategies can be chained together. Plus, I don't think making this generic wouldn't actually reduce the amount of code that need to be replaced

crates/astria-composer/src/strategy.rs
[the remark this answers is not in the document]
dereksione: I'm not sure the reasoning behind this - the bundler we are shipping with is always a no-op. If I remove this, it only changes maybe like 5 lines of code that will have to be a second PR. There is no further functionality that will be added to this

crates/astria-composer/src/ds/wire_format.rs
dereksione: The reasoning behind this is that every blockchain will have a Transaction type, but we will eventually support other Eth libraries. Plus typing `etheres::types::Transaction` everywhere to disambiguate is annoying. But I can remove it if you feel strongly

crates/astria-composer/src/ds/types.rs
```diff
@@ -0,0 +1,9 @@
+use ethers::types::Transaction as EthersTx;
+
+#[non_exhaustive]
+pub enum RollupTx {
+    EthersTx(EthersTx),
+}
+
+pub(crate) type RollupChainId = String;
```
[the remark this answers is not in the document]
dereksione: Not sure why this needs to be a new type - can you elaborate?

crates/astria-composer/src/bundler.rs
dereksione: I'm not sure why? They should have to replace the bundler since that is just infra wiring - I would really rather make the strategies a separate trait that can be extended

crates/astria-composer/src/bundler.rs
dereksione: Yes, you need a mutable reference, which needs ownership

crates/astria-composer/src/searcher.rs
dereksione: I started out this PR with the intent of adding multiple providers, i.e. #214 after talking to @joroshiba and this is the only real new functionality added to the crate with this PR. But we can split it up if that is easier. Would defer to what others think

crates/astria-composer/src/strategy.rs
dereksione: I can split it up into another PR, but I'm not sure if that's really necessary? It will just delete 5 lines of code max

crates/astria-composer/src/ds/mod.rs
dereksione: The data structures crate is only for types that are used by more than one module - I'll take a look at if there was anything shoved there that shouldn't be

crates/astria-composer/src/ds/eth_provider.rs
dereksione: I implemented it for EthersProvider because there is a test provider that I will use for unit testing

crates/astria-composer/src/ds/mod.rs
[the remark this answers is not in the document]
dereksione: Will do

crates/astria-composer/src/ds/streaming_client.rs
dereksione: See above

crates/astria-composer/src/ds/types.rs
dereksione: Can remove if you think it doesn't

crates/astria-composer/src/executor.rs
dereksione: Not sure what you mean. This is a single blocking task. The executor is running in a single task that signs one bundle at a time

crates/astria-composer/src/composer.rs
```diff
@@ -1,80 +1,37 @@
...
 }
 
 impl Composer {
-    /// Constructs a new Searcher service from config.
-    ///
-    /// # Errors
-    ///
-    /// An error is returned if the searcher fails to be initialized.
-    /// See `[Searcher::from_config]` for its error scenarios.
-    pub async fn from_config(cfg: &Config) -> eyre::Result<Self> {
```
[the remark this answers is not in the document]
dereksione: Good point. Will change

dereksione:
UPDATE: synced with @joroshiba and splitting bundler + strategy into a different PR

crates/astria-composer/src/ds/eth_provider.rs
dereksione: LOL, I'll try to do better but here Ethers means the library and Eth is the blockchain

crates/astria-composer/src/ds/types.rs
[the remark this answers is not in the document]
itamarreif: because not every string should be a valid rollup chain id

crates/astria-composer/src/api.rs
itamarreif, line 1: the linter undoes this diff on my machine - not sure if i am the problem or not but flagging this.

crates/astria-composer/src/searcher/mod.rs
```diff
@@ -0,0 +1,123 @@
...
+    executor: Executor,
+    status: watch::Sender<Status>,
+}
+
+impl Searcher {
+    pub(super) fn subscribe(&self) -> watch::Receiver<Status> {
+        self.status.subscribe()
+    }
+
+    pub(super) fn setup_searcher(cfg: &Config) -> Result<Self, eyre::Error> {
```
itamarreif, line 59: can we go back to calling this `from_config`?

PR #235, opened 2023-08-08
use cargo resolver 2, fix used features

SuperFluffy:
This commit changes the cargo workspace to use resolver version 2. Furthermore, it reduces the tokio features to the ones actually needed in composer, and fixes sequencer-relayer's tendermint-rpc features (which surfaced because of resolver 2).

Unfortunately, there is no good way to enforce (or check) that the `tokio` dependency only uses its minimal set of necessary features. In fact, tokio could be specified just like `tokio = { workspace = true }`. Due to feature unification all necessary features are set through the `astria-sequencer` dependency.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,4 @@
 [workspace]
-
 members = [
   "crates/astria-celestia-jsonrpc-client",
   "crates/astria-composer",
@@ -12,6 +11,7 @@ members = [
   "crates/astria-sequencer-utils",
   "crates/astria-test-utils",
 ]
+resolver = "2"
 
 [workspace.dependencies]
 async-trait = "0.1.52"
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -12,6 +11,7 @@ members = [
   "crates/astria-sequencer-utils",
   "crates/astria-test-utils",
 ]
+resolver = "2"
```
joroshiba, line 14: Should we set `edition = 2021` instead? https://doc.rust-lang.org/edition-guide/rust-2021/default-cargo-resolver.html

SuperFluffy: See further down in that link:

> The resolver is a global setting for a workspace, and the setting is ignored in dependencies. The setting is only honored for the top-level package of the workspace. If you are using a virtual workspace, you will still need to explicitly set the resolver field in the [workspace] definition if you want to opt-in to the new resolver.

noot (approved):

joroshiba (approved):

PR #238, opened 2023-08-08
Updating crates with images to match released docker image tags

joroshiba:


noot (approved):

SuperFluffy (approved):
I am fine with doing this for now to match the images, but I think we should adopt a proper release strategy.

PR #239, opened 2023-08-09
feat(composer): read sequencer private key from env and track nonce

### LINKED ISSUE #215: composer: support funded accounts
The initial implementation in #192 sets a fixed Nonce of `0` and generates a random key for each transaction that gets submitted to the sequencer. This will break with #206 because sequencer will require funded accounts to accept transactions.

noot:
- needed so i can merge #206
- reads sequencer private key from env; decided to do this instead of reading from a file since we already have a .env file, so felt a bit duplicative
- also track nonce for this account. note this PR doesn't add handling if a tx is rejected, so if any tx is rejected for some reason the nonce may get out of whack. will open an issue fix this by either resetting the nonce every sequencer block, or re-submitting transactions with the nonce that was rejected

closes  #215

### DIFF
--- a/crates/astria-composer/Cargo.toml
+++ b/crates/astria-composer/Cargo.toml
@@ -13,6 +13,7 @@ ethers = { version = "2.0.7", features = ["ws"] }
 async-trait = { workspace = true }
 axum = { workspace = true }
 backon = { workspace = true }
+borsh = { workspace = true }
 color-eyre = { workspace = true }
 ed25519-consensus = { workspace = true }
 figment = { workspace = true, features = ["env"] }
--- a/crates/astria-composer/Cargo.toml
+++ b/crates/astria-composer/Cargo.toml
@@ -14,7 +14,6 @@ secrecy = "0.8"
 async-trait = { workspace = true }
 axum = { workspace = true }
 backon = { workspace = true }
-borsh = { workspace = true }
 color-eyre = { workspace = true }
 ed25519-consensus = { workspace = true }
 figment = { workspace = true, features = ["env"] }
@@ -45,6 +44,7 @@ features = ["http"]
 impl-serde = "0.4.0"
 tokio-test = "0.4.2"
 
+borsh = { workspace = true }
 figment = { workspace = true, features = ["test"] }
 jsonrpsee = { workspace = true, features = ["macros", "server"] }
 once_cell = { workspace = true }
--- a/crates/astria-composer/Cargo.toml
+++ b/crates/astria-composer/Cargo.toml
@@ -9,6 +9,7 @@ edition = "2021"
 astria-sequencer = { path = "../astria-sequencer" }
 
 ethers = { version = "2.0.7", features = ["ws"] }
+secrecy = "0.8"
 
 async-trait = { workspace = true }
 axum = { workspace = true }
--- a/crates/astria-composer/env.example
+++ b/crates/astria-composer/env.example
@@ -14,4 +14,5 @@ ASTRIA_COMPOSER_CHAIN_ID="912559"
 ASTRIA_COMPOSER_EXECUTION_URL="ws://127.0.0.1:8545"
 
 # Private key for the sequencer account used for signing transactions
+# Must be a hex-encoded 32-byte array (64-character hex string)
 ASTRIA_COMPOSER_PRIVATE_KEY="2bd806c97f0e00af1a1fc3328fa763a9269723c8db8fac4f93af71db186d6e90"
--- a/crates/astria-composer/env.example
+++ b/crates/astria-composer/env.example
@@ -12,3 +12,6 @@ ASTRIA_COMPOSER_CHAIN_ID="912559"
 
 # Address of the RPC server for execution
 ASTRIA_COMPOSER_EXECUTION_URL="ws://127.0.0.1:8545"
+
+# Private key for the sequencer account used for signing transactions
+ASTRIA_COMPOSER_PRIVATE_KEY="2bd806c97f0e00af1a1fc3328fa763a9269723c8db8fac4f93af71db186d6e90"
--- a/crates/astria-composer/env.example
+++ b/crates/astria-composer/env.example
@@ -14,5 +14,4 @@ ASTRIA_COMPOSER_CHAIN_ID="912559"
 ASTRIA_COMPOSER_EXECUTION_URL="ws://127.0.0.1:8545"
 
 # Private key for the sequencer account used for signing transactions
-# Note that this corresponds to the funded "Alice" test key in sequencer
 ASTRIA_COMPOSER_PRIVATE_KEY="2bd806c97f0e00af1a1fc3328fa763a9269723c8db8fac4f93af71db186d6e90"
--- a/crates/astria-composer/env.example
+++ b/crates/astria-composer/env.example
@@ -14,4 +14,5 @@ ASTRIA_COMPOSER_CHAIN_ID="912559"
 ASTRIA_COMPOSER_EXECUTION_URL="ws://127.0.0.1:8545"
 
 # Private key for the sequencer account used for signing transactions
+# Note that this corresponds to the funded "Alice" test key in sequencer
 ASTRIA_COMPOSER_PRIVATE_KEY="2bd806c97f0e00af1a1fc3328fa763a9269723c8db8fac4f93af71db186d6e90"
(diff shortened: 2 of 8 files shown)


crates/astria-composer/src/searcher.rs
```diff
@@ -46,6 +58,10 @@ pub(super) struct Searcher {
     // The client for submitting wrapped and signed pending eth transactions to the astria
     // sequencer.
     sequencer_client: SequencerClient,
+    // Private key used to sign sequencer transactions
+    sequencer_key: SigningKey,
+    // Nonce of the sequencer account we sign with
+    sequencer_nonce: Arc<Mutex<Nonce>>,
```
SuperFluffy, line 64: This would be better as an `Arc<AtomicU32>` (both to not depend on internal sequencer types, but also because the Mutex is heavy and IMO unnecessary).

crates/astria-composer/src/config.rs
```diff
@@ -40,6 +40,9 @@ pub struct Config {
 
     /// Address of the RPC server for execution
     pub execution_url: String,
+
+    /// Private key for the sequencer account used for signing transactions
+    pub private_key: String,
```
SuperFluffy, line 45: Please use `zeroize` and `ZeroizeOnDrop` derived on `Config` to override this data when config is dropped.

Also, please add a `#[serde(serialize_with = "serialize_scrambled")]` or something like that that hides the value of the string (maybe except for the first few chars).

crates/astria-composer/tests/blackbox/helper/mock_sequencer.rs
```diff
@@ -18,6 +25,12 @@ use wiremock::{
 pub async fn start() -> MockServer {
     let server = MockServer::start().await;
     mount_abci_info_mock(&server).await;
+    mount_abci_query_mock(
```
SuperFluffy, line 28: Do you expect this  query to be called more than once? Can you add a `.expect(1)` to it?

crates/astria-composer/env.example
```diff
@@ -12,3 +12,7 @@ ASTRIA_COMPOSER_CHAIN_ID="912559"
 
 # Address of the RPC server for execution
 ASTRIA_COMPOSER_EXECUTION_URL="ws://127.0.0.1:8545"
+
+# Private key for the sequencer account used for signing transactions
+# Note that this corresponds to the funded "Alice" test key in sequencer
```
SuperFluffy, line 17: Please remove the bit about sequencer.

crates/astria-composer/src/searcher.rs
```diff
@@ -1,7 +1,13 @@
-use std::time::Duration;
+use std::{
+    sync::Arc,
+    time::Duration,
+};
 
 use astria_sequencer::{
-    accounts::types::Nonce,
+    accounts::types::{
```
SuperFluffy, line 7: Nit: not a fan of the increasing coupling between sequencer and composer here. I am preparing a PR to reduce this.

crates/astria-composer/Cargo.toml
```diff
@@ -13,6 +13,7 @@ ethers = { version = "2.0.7", features = ["ws"] }
 async-trait = { workspace = true }
 axum = { workspace = true }
 backon = { workspace = true }
+borsh = { workspace = true }
```
SuperFluffy, line 16: Should this be a dev dependency?

crates/astria-composer/src/searcher.rs
```diff
@@ -147,9 +163,25 @@ impl Searcher {
...
+        let private_key_bytes: [u8; 32] = hex::decode(&cfg.private_key)
+            .wrap_err("failed to decode private key bytes from hex string")?
+            .try_into()
+            .map_err(|_| eyre!("invalid private key length; must be 32 bytes"))?;
+        let sequencer_key =
+            SigningKey::try_from(private_key_bytes).wrap_err("failed to parse sequencer key")?;
+        let address = Address::from_verification_key(&sequencer_key.verification_key());
+        let starting_nonce = sequencer_client
+            .inner
+            .get_nonce(&address, None)
```
SuperFluffy, line 176: Please add a timeout to this feature. You can use `backon`.

crates/astria-composer/src/searcher.rs
```diff
@@ -147,9 +163,25 @@ impl Searcher {
...
+            .await
+            .wrap_err("failed to get nonce")?;
```
SuperFluffy, line 178:
```suggestion
            .wrap_err("failed to query cometbft/sequencer for nonce")?;
```

SuperFluffy (changes requested):
Biggest ask is to ensure that the signing key isn't accidentally leaked. To do this, please use https://docs.rs/secrecy and https://docs.rs/zeroize (the former I think exports the latter), and wrap the private key in `secrecy::SecretString`. Then, provide a `serde(serialize_with = "...")` fn that maybe prints the first 3 or 4 chars of the secret and overrides the rest with xxx or ### or smth like that.

crates/astria-composer/src/searcher.rs
joroshiba: We should think about what types in sequencer are things that we and others might want to use externally and create a crate for these. Things like nonce, transactions and signing them are types of things external tools might build around.

crates/astria-composer/src/searcher.rs
noot: I moved this to after `wait_for_seq` in `Searcher::run()`, makes more sense to fetch nonce there

crates/astria-composer/env.example
```diff
@@ -12,3 +12,6 @@ ASTRIA_COMPOSER_CHAIN_ID="912559"
 
 # Address of the RPC server for execution
 ASTRIA_COMPOSER_EXECUTION_URL="ws://127.0.0.1:8545"
+
+# Private key for the sequencer account used for signing transactions
+ASTRIA_COMPOSER_PRIVATE_KEY="2bd806c97f0e00af1a1fc3328fa763a9269723c8db8fac4f93af71db186d6e90"
```
joroshiba, line 17: Does this private key correspond to one of the accounts in the local test genesis (would be good to have a funded account already here)?

joroshiba (commented):
I think we need to add logic in here to support refetching the nonce if a tx submission fails. Otherwise we will run into issues if running two composer nodes with same private key, or if a tx submission fails for some other reason.

SuperFluffy:
> I think we need to add logic in here to support refetching the nonce if a tx submission fails. Otherwise we will run into issues if running two composer nodes with same private key, or if a tx submission fails for some other reason.

I agree that that's necessary, but I think it's also fair to leave this to a follow-up PR (because I think the proper design of that feature would blow up this PR).

crates/astria-composer/src/searcher.rs
```diff
@@ -216,6 +250,18 @@ impl Searcher {
             Ok(((), ())) => {}
             Err(err) => return Err(err).wrap_err("failed to start searcher"),
         }
+
+        // set initial sequencer nonce
+        let address = Address::from_verification_key(&self.sequencer_key.verification_key());
+        let starting_nonce = self
```
SuperFluffy, line 256: Nit (to be left for a future PR because this does not have to be addressed here): getting the nonce from the sequencer is such a strong statement that it can be used in lieu of the `abci_info` query that is used in `wait_for_sequencer`.

SuperFluffy (approved):
Thank you, this is great, I like that we take care to not leak secrets.

Note my suggestions on the `serialize_private_key` suggestion which I think is not a blocker, but probably good practice to anticipate what might cause confusion when people are confused by their logs.

crates/astria-composer/env.example
```diff
@@ -12,3 +12,6 @@ ASTRIA_COMPOSER_CHAIN_ID="912559"
 
 # Address of the RPC server for execution
 ASTRIA_COMPOSER_EXECUTION_URL="ws://127.0.0.1:8545"
+
+# Private key for the sequencer account used for signing transactions
```
SuperFluffy, line 16: As an afterthought: can you add that this is supposed to be a hex encoded 32 byte array (meaning that the key has to be 64 chars long)?

crates/astria-composer/env.example
noot: yeah it corresponds to the `alice` account

noot:
@joroshiba yeah I think that should be in a different PR, and apparently #226 addresses the issue also

crates/astria-composer/src/searcher.rs
noot: agreed, i can open an issue

PR #247, opened 2023-08-10
sequencer: decouple abci account queries

SuperFluffy:
This is the first of 2 commits to address  https://github.com/astriaorg/astria/issues/191. This commit refactors sequencer to return protobuf types when it receives a tendermint/cometbft `abci_query` request defined in `astria-proto`.

## Background

At the moment, the sequencer ABCI API return types are only defined in sequencer's Rust code. This goes against the idea that `astria-proto` be the only source of truth for astria's public API. Furthermore, the wire format of these types is defined by `borsh`, not protobuf.

An additional reason to refactor is that other astria services either directly or indirectly (by importing `astria-sequencer-client`) depend on `astria-sequencer`, which
1. quantitatively leads to long compile times and bad developer experience (because when building e.g. `composer` all of `sequencer` is compiled along with it), and
2. qualitatively increases coupling which means that refactors of `astria-sequencer` might immediately force refactors of all its users.

## Open issues for followup commits

This commit does not completely decouple sequencer and its RPC. `SequencerClientExt::submit_transaction_sync` and `SequencerClientExt::submit_transaction_commit` are left untouched, and hence transaction construction still remains tightly coupled with sequencer (this includes types like `Balance`, `Nonce`, and of course the signed and unsigned transactions).

The refactoring of this logic will be left to a follow-up commit, which will finish the decoupling and allow to make all sequencer types crate-private again.

### 1. Is astria-proto the right crate for this? Should we rename it?

The main reason to keep the native types close to the protobuf-derived types are orphan rules and cyclic dependencies:

1. assume we have crates `astria-proto` and `astria-native` that each define a `Response` type;
2.  `astria-native` cannot implement methods like `astria-proto::Response::to_native` due to coherence/orphan rules;
3. but dual methods like `astria-native::Response::to_proto -> astria_proto::Response` and `astria-proto::Response::to_native -> astria_native::Response` are not possible because this would cause a cyclic dependency.

Solution 1: `native` and `proto` are submodules of one common repository.
Solution 2: there are no conversion methods in `astria_proto`; instead `astria_native` defines traits like `IntoNative` and implements  the conversions in terms of those.
Solution 3: we don't implement any methods on the protobuf-derived types, and instead rely entirely on methods like `astria-native::Response::to_proto` and `astria-native::Response::from_proto`.

Solution 2 adds complexity by introducing single-use traits (I don't see a lot of cases where we really desire to abstract over the conversion). Solution 3 is simple, but the code would be more verbose. Instead of having things like `proto_type.to_native()` we would have `NativeType::from_proto(proto_type);`. This however does not constitute a huge problem.

### 2. Request counterparts for the proto responses and grpc

Should the `*Response` protobuf types already come with `*Request` counterparts? Should this PR already contain grpc definitions? I am leaning toward leaving this to a future PR where these will actually find use. YAGNI.

### 3. Should the native RPC types be versioned?

Right now the native RPC types are under `astria_proto::native::sequencer` - does it make sense to keep them that way or should they be versioned similar to the protobuf types? If their types are kept public (as they are right now), then probably yes. An alternative is to keep their implementation details hidden and only provide getters.

### New protobuf types

Taking inspiration from #193, this commit defines the protobuf messages `astria.sequencer.v1alpha1.sequencer.BalanceResponse` and `astria.sequencer.v1alpha1.sequencer.NonceResponse`. The intention is that these will be returned by the sequencer. This commit foregoes `NonceRequest` and `BalanceRequest` because they will not find immediate use, but they could be defined in follow-up PRs together with grpc service definitions.

### Native RPC types

This commit introduces "native" astria types, which are types converted form the protobuf types that are more ergonomic to use. For example, a field `balance: Option<U128>` is convered to a `balance: u128` (mapping `None` to `0`), and `account: Vec<u8>` is converted to an `account: Address`, where `Address([u8; 20])`.

### Reasoning behind putting the native types into `astria-proto`

The RPC types themselves could have been put into a `astria-rpc` crate instead, with conversion methods from/to proto. Since it is also convenient to implement conversion method on the proto types to get to the native types, these have to stay in `astria-proto` (due to orphan rules). Therefore it makes sense to put the native and the proto types side-by-side because they are so tightly coupled.

### Protobuf types are no longer exported at the root

To not confuse native RPC types and generated RPC types, `astria_proto::generated::*` are no longer re-exported at the root. Instead, only the `generic` and `native` modules are exported. This is a breaking change.

### Update and refactor of  sequencer

The `tendermint::abci::response::Query` types returned by the info service now contain the `BalanceResponse` and `NonceResponse` types defined above. Furthermore, in case the abci queries encounter errors, the query will be returned with a non-zero error code and with its `log` and `info` fields populated. This PR attempts to standardize on the application specific status codes by introducing a `sequencer::service::info::AbciCode` type that contains several associated types.,

Finally, `sequencer::accounts::query::Request`, `sequencer::acounts::query::QueryHandler` and `sequencer::service::info::Query` were removed. `sequencer::service::info::Info` was reworked to use a `matchit::Router` and type erased abci query handler. This was inspired by axum's `Router` implementation and lowers coupling between the info service and the query handling logic in the accounts module.

### Update of sequencer-client

Sequencer client was updated to correctly interpret the returned bytes as protobuf types.


crates/astria-sequencer-client/src/lib.rs
```diff
@@ -248,60 +197,54 @@ impl ErrorKind {
...
-    ) -> Result<Balance, Error> {
-        use borsh::BorshDeserialize as _;
-        use sequencer::accounts::query::Response;
-        let height = height.map(Into::into);
+    /// - If the bytes contained in the abci query response cannot be read as an
+    ///   `astria.sequencer.v1alpha1.BalanceResponse`.
+    async fn get_balance(&self, address: [u8; 20], height: u32) -> Result<BalanceResponse, Error> {
+        use proto::Message as _;
+
+        let prefix = b"accounts/balance/";
```
noot, line 215: make this a const (or put it in `sequencer` query module?)

crates/astria-sequencer-client/src/lib.rs
```diff
@@ -248,60 +197,54 @@ impl ErrorKind {
...
+        let mut path = vec![0u8; prefix.len() + 20 * 2];
+        path.extend_from_slice(prefix);
```
noot, line 217: doesn't this append the prefix to the vector? shouldn't it be using `copy_from_slice` since you want to put the prefix into the start of `path`?

crates/astria-sequencer-client/src/lib.rs
```diff
@@ -248,60 +197,54 @@ impl ErrorKind {
...
+        faster_hex::hex_encode(&address, &mut path).expect(
```
noot, line 218: what's the benefit of the `faster_hex` library? also should `&mut path` be `&mut path[prefix.len()..]` since the address goes after the prefix?

crates/astria-sequencer-client/src/lib.rs
```diff
@@ -311,33 +254,44 @@ pub trait SequencerClientExt: Client {
...
+    async fn get_nonce(&self, address: [u8; 20], height: u32) -> Result<NonceResponse, Error> {
+        use proto::Message as _;
+
+        let prefix = b"accounts/nonce/";
+        let mut path = vec![0u8; prefix.len() + 20 * 2];
+        path.extend_from_slice(prefix);
+        faster_hex::hex_encode(&address, &mut path).expect(
+            "this is a bug: the buffer holding the abci query path should have the correct \
+             precalculated size",
+        );
```
noot, line 268: same comments about the path as above

crates/astria-sequencer-client/src/tests/http.rs
```diff
@@ -120,57 +129,69 @@ fn create_signed_transaction() -> transaction::Signed {
...
+        server,
+        client,
+    } = MockSequencer::start().await;
+
+    let expected_response = NonceResponse {
+        account: ALICE_ADDRESS.to_vec(),
+        height: 10,
+        nonce: 1,
+    };
+    let _guard =
```
noot, line 152: what's the purpose of declaring `_guard`?

crates/astria-sequencer/src/service/info.rs
```diff
@@ -36,28 +32,119 @@ use tendermint::{
...
-    accounts::query::QueryHandler,
-    state_ext::StateReadExt,
-};
+#[derive(Clone, Copy, Debug)]
+pub(crate) struct AbciCode(u32);
+impl AbciCode {
+    pub(crate) const INTERNAL_ERROR: Self = Self(3);
+    pub(crate) const INVALID_PARAMETER: Self = Self(2);
+    pub(crate) const OK: Self = Self(0);
+    pub(crate) const UNKNOWN_PATH: Self = Self(1);
```
noot, line 45: nit but can you order these :p also are these from somewhere? if so can you add a comment about that

SuperFluffy: I did order them, but rustfmt decided that it preferred them in alphabetic order. 😭

crates/astria-sequencer/src/service/info.rs
```diff
@@ -72,126 +159,119 @@ impl Service<InfoRequest> for Info {
...
-                .await
-                .context("failed handling query request")?,
-        )),
-        // this was removed after v0.34
-        InfoRequest::SetOption(_) => Ok(InfoResponse::SetOption(SetOption {
-            code: Code::default(),
-            log: "SetOption is not supported".to_string(),
-            info: "SetOption is not supported".to_string(),
-        })),
+struct BoxedAbciQueryHandler(Box<dyn ErasedAbciQueryHandler>);
```
noot, line 169: what is this type for (and ErasedAbciQueryHandler?)

crates/astria-sequencer/src/service/info.rs
```diff
@@ -36,28 +32,119 @@ use tendermint::{
...
+                code: Code::default(),
+                log: "SetOption is not supported".to_string(),
+                info: "SetOption is not supported".to_string(),
+            })),
         }
     }
+
+    /// Handles `abci_query` RPCs.
+    async fn handle_abci_query(self, request: request::Query) -> response::Query {
+        let (handler, params) = match self.query_router.at(&request.path) {
```
noot, line 124: where do the `params` come from here? are they just parsed from the path basically?

crates/astria-sequencer/src/service/info.rs
noot: omg :| that's annoying lol

crates/astria-sequencer/src/service/info.rs
SuperFluffy: Yes, fundamentally that is the case.

I am going to write down my reasoning as to why we ended up with these sevaral layers of complexity:

1. I want to register one handler per matched path, so that we can quickly associate `path -> handler`;
2. first problem: every handler is a `async fn -> ret`, which is sugar for `fn -> impl Fut<Out=ret>`. So because every handler returns an anonymous type, each handler itself is an anonymous type;
3. there are 2 ways around this:
    a. change the function signature to `fn -> Box<dyn Fut<Out=ret>>`
    b. box the entire function, i.e. have a a `Box<dyn fn -> impl Fut=ret>`
4. I chose option b. because that way the function signature of the handler does not have to change; `trait AbciQueryHandler` just makes this all nicer because I can avoid wirting all the `Clone + Send + etc` stuff line.
5. Now, `struct Info` is `Clone`, so `matchit::Router<T>` must be clone, but `AbciQueryHandler: Clone`, so it's not typesafe, meaning, I can't make a `Box<dyn AbciQueryHandler>` from it.
6. This means we need to introduce another `trait ErasedAbciQueryHandler` for it that is not clone. But you see it has `ErasedAbciQueryHandler::clone_box`, so I can implement `Clone for BoxedAbciQueryHandler` in terms of it.
7.  Finally, I have to have a way to go from a handler (that implements `AbciQueryHandler`) to something that implements `ErasedAbciQueryAndler`. This is where `MakeErasedAbciQueryHandler` comes in as a wrapper around the handler.

crates/astria-sequencer-client/src/tests/http.rs
SuperFluffy: It's to provide a cleaner way of tracking which mock failed. Since `register_abci_query_response` uses `Mock::mount_as_scoped`, the `expect(1)` condition is checked when `_guard` goes out of scope. If it were mounted using just `Mock::mount`, the condition would be checked only once the entire server goes out of scope.

So doing it this way you will know that the condition associated with this `_guard` is the culprit.

The underscore is just because `_guard`'s entire purpose is to be checked on drop, and rust's dead code analysis would complain that `guard` is never used.

crates/astria-sequencer-client/src/lib.rs
SuperFluffy: Good find, thank you so much! I fixed one bug and introduced another.

I added a `make_path_from_prefix_and_address` helper function and added a test to ensure we get the expected result.

crates/astria-sequencer-client/src/lib.rs
SuperFluffy: I removed `faster_hex` because it's orthogonal to this PR. In general it's better to use it over `hex` because it's a lot faster.

Re the path: tysm! I messed up while making the tests work, and it turns out the mock tests didn't catch thits. Added a helper function to the construct to make sure it lines up with what we expect.

In general the mock tests in this crate should probably be replaced by integration tests against an actual sequencer (+ cometbft).

crates/astria-sequencer-client/src/lib.rs
SuperFluffy: I made the prefixes const. I guess we could expose them somewhere, but I think it's fine if they remain private for now.

Not sure which `sequencer` query module you mean? If `sequencer::accounts::query` then I think no, because it's the responsibility of `sequencer::service::info` to route the requests to the relevant helpers.

crates/astria-sequencer-client/src/lib.rs
SuperFluffy: ty for noticing this! see the other comment re fix

crates/astria-sequencer/src/service/info.rs
SuperFluffy: I forgot to add: no, these are application defined, right? So we could do some bikeshedding on the naming haha

SuperFluffy:
Heads-up: rebasing on top of recent master made more changes necessary. In the meantime I have moved `Address` construction to `astria-proto` and refactor sequencer accordingly. I now need to change composer. Going to reopen the PR after I am done with that.

crates/astria-proto/src/lib.rs
```diff
@@ -1,12 +1,15 @@
...
 #[rustfmt::skip]
 mod proto;
 
-pub use proto::generated::{
-    execution,
-    primitive,
-    sequencer,
-};
+#[cfg(feature = "native")]
+pub mod native;
```
SuperFluffy, line 13: I am torn on nesting this here. I wonder if we should introduce a `astria-rpc` or `astria-api` crate and make the protos a submodule of it.

I believe there are good reasons to have the native types and their protobuf counterparts live in the same crate (mainly to avoid issues around orphan rules), but having the "native" types underneath something called `-proto` seems confusing?

crates/astria-proto/src/lib.rs
```diff
@@ -1,12 +1,15 @@
...
+    EncodeError,
+    Message,
+};
+
 #[allow(warnings)]
 #[allow(unreachable_pub)]
 #[rustfmt::skip]
 mod proto;
 
-pub use proto::generated::{
```
SuperFluffy, line 6: Note here that with this commit the protos are no longer exported at the root level but only accessible through `generated`.

crates/astria-composer/src/searcher.rs
```diff
@@ -197,6 +189,12 @@ impl Searcher {
 
     /// Serializes and signs a sequencer tx from a rollup tx.
     fn handle_pending_tx(&mut self, rollup_tx: Transaction) {
+        use astria_sequencer::{
+            accounts::types::Nonce,
+            sequence,
+            transaction::action,
+        };
```
noot, line 196: are you planning to eventually break these types out as well?

crates/astria-proto/proto/astria/sequencer/v1alpha1/account.proto
```diff
@@ -0,0 +1,21 @@
...
+
+syntax = "proto3";
+
+package astria.sequencer.v1alpha1;
+
+import "astria/primitive/v1/types.proto";
+
+// A response containing the balance of an account.
+message BalanceResponse {
+  bytes account = 1;
```
noot, line 11: do we need to return account here since account is already in the request? same for `NonceResponse`

crates/astria-proto/src/lib.rs
noot: agreed, we should probably rename this crate and put `proto` as a module inside it

crates/astria-proto/src/native/sequencer.rs
```diff
@@ -0,0 +1,326 @@
...
+
+use crate::generated::sequencer::v1alpha1;
+
+pub const ADDRESS_LEN: usize = 20;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct Address([u8; ADDRESS_LEN]);
+
+impl Address {
+    pub const LEN: usize = ADDRESS_LEN;
```
noot, line 14:
```suggestion
    pub const LEN: usize = 20;
```

super nit but then you can remove `ADDRESS_LEN`?

crates/astria-proto/src/native/sequencer.rs
```diff
@@ -0,0 +1,326 @@
...
+    pub fn from_proto(proto: v1alpha1::BalanceResponse) -> Result<Self, IncorrectAddressLength> {
+        let v1alpha1::BalanceResponse {
+            account,
+            height,
+            balance,
+        } = proto;
+        Ok(Self {
+            account: Address::try_from_slice(&account)?,
+            height,
+            balance: balance.map_or(0, Into::into),
```
noot, line 168: should this error if balance is `None`?

crates/astria-sequencer/src/accounts/query.rs
```diff
@@ -1,84 +1,180 @@
...
-    BalanceQuery(Address),
-    NonceQuery(Address),
+pub(crate) async fn balance_request(
+    storage: Storage,
+    request: request::Query,
+    params: Vec<(String, String)>,
+) -> response::Query {
+    use astria_proto::{
+        native::sequencer::BalanceResponse,
+        // sequencer::v1alpha1::BalanceResponse,
```
noot, line 28: remove

crates/astria-sequencer/src/accounts/query.rs
```diff
@@ -1,84 +1,180 @@
...
+            log: "path did not contain path parameter".into(),
+            ..response::Query::default()
+        });
+    };
+    let address = hex::decode(address)
+        .map_err(|err| response::Query {
+            code: AbciCode::INVALID_PARAMETER.into(),
+            info: format!("{}", AbciCode::INVALID_PARAMETER),
+            log: format!(
+                "account public key could not be constructed from provided paratemer: {err:?}"
```
noot, line 154:
```suggestion
                "address could not be constructed from provided paratemer: {err:?}"
```

crates/astria-sequencer/src/accounts/query.rs
```diff
@@ -1,84 +1,180 @@
...
+            ),
+            ..response::Query::default()
+        })
+        .and_then(|addr| {
+            Address::try_from_slice(&addr).map_err(|err| response::Query {
+                code: AbciCode::INVALID_PARAMETER.into(),
+                info: format!("{}", AbciCode::INVALID_PARAMETER),
+                log: format!(
+                    "account public key could not be constructed from provided paratemer: {err:?}"
```
noot, line 163:
```suggestion
                    "address could not be constructed from provided parameter: {err:?}"
```

crates/astria-proto/src/lib.rs
SuperFluffy: I like `astria-core` better, I have a dislike for crates and modules named `*-types`.

I would however leave the renaming to a followup-pr to not blow up this PR (and to not mix renaming + new business logic).

And then we can do better bikeshedding on names. 😂

What do you think?

crates/astria-sequencer/src/crypto.rs
```diff
@@ -1,39 +0,0 @@
-pub(crate) use ed25519_consensus::{
-    Signature,
-    SigningKey,
-    VerificationKey,
-};
-
-#[cfg(test)]
-mod test {
```
noot, line 8: it would be nice to keep this somewhere as it shows how the default accounts were created, but i agree it isn't a test and should probably not live here. thoughts?

crates/astria-composer/src/searcher.rs
SuperFluffy: Yes, that will be done in a follow-up so that this PR doesn't explode.

crates/astria-sequencer/src/service/info.rs
noot: true, we could prefix them with something, up to you lol

crates/astria-proto/src/lib.rs
```diff
@@ -1,12 +1,15 @@
...
-    execution,
-    primitive,
-    sequencer,
-};
+#[cfg(feature = "native")]
```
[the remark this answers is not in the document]
SuperFluffy: Yes, exactly. This does pull in a lot of dependencies and it's probably what you want by default, but my thinking is that if I can cleanly feature-gate this to disable it there is no harm done.

crates/astria-sequencer/src/crypto.rs
SuperFluffy: I will bring these back as doc tests or maybe as an example.

I'll think on it over night 😃

crates/astria-sequencer/src/service/info.rs
noot: wow a lot of steps, thanks for explaining!

noot:
> 1. Is astria-proto the right crate for this? Should we rename it?

I think solution 1 and 3 you outlined here are both fine, I don't have a strong preference for one or the other. if we do keep the native types in what is currently `astria-proto`, we should rename the crate and put the protos in their own module. part of me thinks it would be nice to have only protos in a crate and do solution 3, but I don't feel super strongly about it.

> 2. Request counterparts for the proto responses and grpc

yeah if these aren't necessary right now I'd leave it as is

> 3. Should the native RPC types be versioned?

yes, probably. it's unlikely that stuff like `Address` will change but I guess it might. do you think we need versioning now, or only once we're doing a proper 1.0 (or public testnet) release?

crates/astria-sequencer/src/accounts/query.rs
SuperFluffy: used the suggested change and streamlined the response construction.

SuperFluffy:
> > 3. Should the native RPC types be versioned?
> 
> yes, probably. it's unlikely that stuff like `Address` will change but I guess it might. do you think we need versioning now, or only once we're doing a proper 1.0 (or public testnet) release?

I moved the native sequencer types under `astria_proto::native::sequencer::v1alpha1` to be in sync with the protobufs.

crates/astria-proto/src/native/sequencer.rs
joroshiba: I think this is also kinda correct here? A balance of none doesn't really make sense.

joroshiba (approved):

crates/astria-composer/src/searcher.rs
```diff
@@ -9,18 +9,7 @@ use std::{
...
-        Nonce,
-    },
-    sequence::Action as SequenceAction,
-    transaction::{
-        action::Action as SequencerAction,
-        Signed as SignedSequencerTx,
-        Unsigned as UnsignedSequencerTx,
-    },
-};
+use astria_sequencer::transaction;
```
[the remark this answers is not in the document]
SuperFluffy: I am not sure what you mean by scoping?

crates/astria-proto/proto/astria/sequencer/v1alpha1/account.proto
```diff
@@ -0,0 +1,19 @@
+// Requests for account information on and responses from the astria sequencer.
```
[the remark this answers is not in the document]
SuperFluffy: Yeah, it looks like there is now way to provide the equivalent to a rust module comment. I will remove.

crates/astria-sequencer/src/service/info.rs
SuperFluffy: @joroshiba I have moved the router impl to a submodule under `service::info` and explained the implementation there.

PR #248, opened 2023-08-10
feat(relayer): add `retain` public key check before deserialization retain checks

### LINKED ISSUE #182: Ensure that verification keys match before reading rollup data
> maybe add a `retain` before this that checks the public key which signed the rollup data is the same as `verification_key`? might save on some serialization/hashing

_Originally posted by @noot in https://github.com/astriaorg/astria/pull/150#discussion_r1267089828_

noot:
- adds a `retain` that checks the public key which signed the rollup data is the same as `verification_key` to save on deserialization when fetching rollup data
closes #182


SuperFluffy:
Also, please add more info in the PR description about why this is necessary.

joroshiba (approved):

PR #249, opened 2023-08-11
Remove the size labeler, annoying message duplication

joroshiba:


SuperFluffy (approved):
Heh, yeah. That was a bit excessive :-)

PR #255, opened 2023-08-11
sequencer: use penumbra's `Storage` instead of `TempStorage`

### LINKED ISSUE #254: Use Storage instead of TempStorage

steezeburger:
This PR changes the storage type we are using in `sequencer`. It is being changed from `penumbra_storage::TempStorage` to `penumbra_storage::Storage`. I also added a config value to define the path of the penumbra database file.

Resolves #254

### DIFF
--- a/crates/astria-sequencer/src/config.rs
+++ b/crates/astria-sequencer/src/config.rs
@@ -1,3 +1,5 @@
+use std::path::PathBuf;
+
 use clap::Parser;
 use serde::{
     Deserialize,
@@ -16,7 +18,7 @@ pub struct Config {
     pub(crate) genesis_file: String,
     /// The path to penumbra storage db.
     #[arg(long)]
-    pub(crate) db_datadir: String,
+    pub(crate) db_filepath: PathBuf,
 }
 
 impl Config {
--- a/crates/astria-sequencer/src/config.rs
+++ b/crates/astria-sequencer/src/config.rs
@@ -14,6 +14,9 @@ pub struct Config {
     /// The path to the json encoded genesis file with a list of accounts.
     #[arg(long)]
     pub(crate) genesis_file: String,
+    /// The path to penumbra storage db.
+    #[arg(long)]
+    pub(crate) db_filepath: String,
 }
 
 impl Config {
--- a/crates/astria-sequencer/src/config.rs
+++ b/crates/astria-sequencer/src/config.rs
@@ -16,7 +16,7 @@ pub struct Config {
     pub(crate) genesis_file: String,
     /// The path to penumbra storage db.
     #[arg(long)]
-    pub(crate) db_filepath: String,
+    pub(crate) db_datadir: String,
 }
 
 impl Config {
--- a/crates/astria-sequencer/src/sequencer.rs
+++ b/crates/astria-sequencer/src/sequencer.rs
@@ -29,8 +29,8 @@ impl Sequencer {
     pub async fn run_until_stopped(config: Config) -> Result<()> {
         let genesis_state =
             GenesisState::from_path(config.genesis_file).context("failed reading genesis state")?;
-        let db_filepath = PathBuf::from(config.db_datadir);
-        let storage = penumbra_storage::Storage::load(db_filepath.clone())
+        let db_path = PathBuf::from(config.db_datadir);
+        let storage = penumbra_storage::Storage::load(db_path.clone())
             .await
             .context("failed to load storage backing chain state")?;
         let snapshot = storage.latest_snapshot();
--- a/crates/astria-sequencer/src/sequencer.rs
+++ b/crates/astria-sequencer/src/sequencer.rs
@@ -1,5 +1,3 @@
-use std::path::PathBuf;
-
 use anyhow::{
     anyhow,
     Context as _,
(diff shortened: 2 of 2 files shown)


crates/astria-sequencer/src/config.rs
```diff
@@ -14,6 +14,9 @@ pub struct Config {
     /// The path to the json encoded genesis file with a list of accounts.
     #[arg(long)]
     pub(crate) genesis_file: String,
+    /// The path to penumbra storage db.
+    #[arg(long)]
+    pub(crate) db_filepath: String,
```
noot, line 19: this is often called `datadir`, rename?

noot (approved):

crates/astria-sequencer/src/config.rs
steezeburger: I copied the name from https://github.com/penumbra-zone/penumbra/blob/43768e74c1e4a6bd8280dd3faeb487b791b49fc0/crates/storage/src/storage/temp.rs#L24

but looking at it deeper, I think that even though it's named `penumbra.db`, it is actually a folder. I'll update it!

SuperFluffy: @steezeburger the original terminology was correct. `Storage::load` opens a binary file at the supplied path: https://rustdoc.penumbra.zone/main/src/penumbra_storage/storage.rs.html#46-130

Please rename back because calling this a directory is not correct.

crates/astria-sequencer/src/sequencer.rs
```diff
@@ -28,9 +30,10 @@ impl Sequencer {
     pub async fn run_until_stopped(config: Config) -> Result<()> {
         let genesis_state =
             GenesisState::from_path(config.genesis_file).context("failed reading genesis state")?;
-        let storage = penumbra_storage::TempStorage::new()
+        let db_path = PathBuf::from(config.db_datadir);
+        let storage = penumbra_storage::Storage::load(db_path.clone())
```
SuperFluffy, line 34: Can you check for the existence of a file and report whether a db will be created or opened? This isn't ideal because it should happen inside the `Storage::load` function, but it is what it is.

SuperFluffy (changes requested):
Please revert the name of the config var, see the penumbra code I linked. Dir is correct for the tempstorage, but not for the (non-temp-) storage.

joroshiba:
Should we have a more complex setting which allows to run with temp storage if desired for local/testing environments?

SuperFluffy:
> Should we have a more complex setting which allows to run with temp storage if desired for local/testing environments?

IMO that's not necessary because of penumbra storage works. Just give it `ASTRIA_SEQUENCER_DB_FILEPATH=/tmp/mysequencer.db"`

crates/astria-sequencer/src/config.rs
```diff
@@ -14,6 +16,9 @@ pub struct Config {
     /// The path to the json encoded genesis file with a list of accounts.
     #[arg(long)]
     pub(crate) genesis_file: String,
+    /// The path to penumbra storage db.
+    #[arg(long)]
+    pub(crate) db_filepath: PathBuf,
```
SuperFluffy, line 21: bikeshed (feel free to ignore): `db_filepath` is kinda excessive, `db_path` is probably fine.

steezeburger:
> because of penumbra storage works.

Penumbra's `TempStorage` uses `tempfile::TempDir` which deletes the directory as soon as the variable goes out of scope, but data in `/tmp` directories is generally deleted when the computer restarts for Mac. I think this is OS dependent.

crates/astria-sequencer/src/sequencer.rs
```diff
@@ -28,9 +28,20 @@ impl Sequencer {
     pub async fn run_until_stopped(config: Config) -> Result<()> {
         let genesis_state =
             GenesisState::from_path(config.genesis_file).context("failed reading genesis state")?;
-        let storage = penumbra_storage::TempStorage::new()
+        if config.db_filepath.try_exists()? {
```
SuperFluffy, line 31: If you use `try_exists`, please provide error context, .e. `try_exists().context("failed checking for existence of db storage file")?`;

However, I think it's fine to just do `exists()` here because if there is some weirdness with the file then `Storage::load` will probably fail.

crates/astria-sequencer/src/sequencer.rs
```diff
@@ -28,9 +28,20 @@ impl Sequencer {
...
+            info!(
+                path = config.db_filepath.display().to_string(),
```
SuperFluffy, line 33: No need to allocate:
```suggestion
                path = %config.db_filepath.display(),
```

SuperFluffy (approved):
Looks fine to me. Please address the minor comments before merging.

crates/astria-sequencer/src/sequencer.rs
steezeburger: The [docs](https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.exists) say `exists` is error prone and to consider `try_exists`. Is that fine in this instance?

SuperFluffy:
> > because of penumbra storage works.
> 
> Penumbra's `TempStorage` uses `tempfile::TempDir` which deletes the directory as soon as the variable goes out of scope, but data in `/tmp` directories is generally deleted when the computer restarts for Mac. I think this is OS dependent.
> 
> This matters for local development because of the new persistent volumes in dev-cluster.

Yeah, but that's a requirement driven by local development that can be easily fixed by using a just command or a script to delete the tempdir, or by doing `ASTRIA_SEQUENCER_DB_FILE_PATH=/tmp/$(rand -hex 10)` or something like that.

I want to be careful with running diverging codepaths between prod and local test if it's just convenience.

crates/astria-sequencer/src/sequencer.rs
SuperFluffy: My thinking is that we are just doing some light reporting/guessing about what we think penumbra is going to do. I am not sure in how far the error conditions listed in the doc are relevant to us. I am wondering if they would be helpful or confusing (given that at that code point we have no insight into penumbra).

steezeburger:
I do like the idea of a just command to clean up local persistent storage. I'll add that.

PR #256, opened 2023-08-14
feat(sequencer, relayer): generate and write merkle inclusion proofs as part of `RollupNamespaceData`

### LINKED ISSUE #180: fix confusing naming in data availability module
Also some naming confusion here, for follow up: `rollup_txs` on `SequencerBlockData` is more a list of rollup tx bundles, but on `RollupNamespaceData` is actually a list of rollup txs.

_Originally posted by @joroshiba in https://github.com/astriaorg/astria/pull/150#discussion_r1262794790_

noot:
- implement `MerkleTree` and `InclusionProof` types (compatible with tendermint's merkle tree algo) inside a new `astria-sequencer-validation` crate (happy to bikeshed on the naming of this)
- implement generation of merkle tree inside relayer, also generate inclusion proof for each `RollupNamespaceData` which is written to DA
- this proves that the rollup data was included inside the sequencer block's "action tree root commitment"
- next PR: include inclusion proof of the commitment tx inside the header `data_hash` as part of `SequencerNamespaceData`

related to #153

closes  #180


crates/astria-sequencer-relayer/src/data_availability.rs
```diff
@@ -9,6 +12,13 @@ use astria_celestia_jsonrpc_client::{
     Client,
     ErrorKind,
 };
+use astria_sequencer::proposal::{
```
SuperFluffy, line 15: Early feedback on this: we should not increase coupling between sequencer and sequencer-relayer further, but split this out into a free-standing crate, and instead have something like `astria-sequencer-core`.

Also `InclusionProof` and `MerkleTree` are used in RPCs/as part of a public API, so should we make these into protobufs?

noot: yes agreed, going to put all this shared stuff into its own crate

crates/astria-sequencer-types/src/sequencer_block_data.rs
```diff
@@ -42,8 +49,8 @@ pub struct SequencerBlockData {
     pub(crate) header: Header,
     /// This field should be set for every block with height > 1.
     pub(crate) last_commit: Option<Commit>,
-    /// namespace -> rollup txs
-    pub(crate) rollup_txs: HashMap<Namespace, Vec<Vec<u8>>>,
+    /// namespace -> rollup data (chain ID and transactions)
+    pub(crate) rollup_data: HashMap<Namespace, RollupData>,
```
noot, line 53: debating just changing this into `HashMap<Vec<u8>, Vec<Vec<u8>>>` where the key is the `chain_id` and the value is the transactions, and then only doing chain_id -> namespace when we're constructing the `RollupNamespaceData`

Cargo.toml
```diff
@@ -8,6 +8,8 @@ members = [
   "crates/astria-sequencer",
   "crates/astria-sequencer-client",
   "crates/astria-sequencer-relayer",
+  "crates/astria-sequencer-validation",
+  "crates/astria-sequencer-types",
```
joroshiba, line 12: Should the types here have been added in a different PR seems like might warrant a seperate small quick PR?

crates/astria-sequencer-validation/src/proof.rs
```diff
@@ -0,0 +1,190 @@
...
+        let digest = *sha2::digest::Output::<Sha256>::from_slice(root_hash.as_bytes());
+        let ct_root = RootHash::<Sha256>::new(digest, self.num_leaves);
+        ct_root
+            .verify_inclusion(&self.value, self.idx, &self.inclusion_proof)
+            .map_err(|e| eyre::eyre!("failed to verify inclusion: {}", e))
+    }
+}
+
+#[cfg(test)]
+mod test {
```
joroshiba, line 127: Seems like another good spot for a snapshot test

joroshiba (approved):

Cargo.toml
noot: yeah, it doesn't seem to have any ill effect right now but will add

crates/astria-sequencer-types/src/sequencer_block_data.rs
noot: I'm leaning towards separating it out b/c this type should be generic/usable for any DA layer while `Namespace` here is celestia-specific, it also makes it easier/cleaner to implement some of the verification logic. I'll open a separate PR for it and see how you feel

noot: actually it doesn't really matter, I can leave as-is. I think we need to open up a bigger discussion about what we're gossiping around and the format of it, as that also matters for verification.

crates/astria-sequencer-validation/src/proof.rs
```diff
@@ -0,0 +1,214 @@
...
+            num_leaves: self.0.len(),
+            inclusion_proof: self.0.prove_inclusion(idx),
+        })
+    }
+}
+
+/// A merkle proof of inclusion.
+#[derive(Debug)]
+#[allow(clippy::module_name_repetitions)]
+pub struct InclusionProof {
```
SuperFluffy, line 62: Am I right that the main purpose of this type is to allow sending proofs over the wire? Should the value-to-be-proven indeed be serialized along with the proof?

crates/astria-sequencer-validation/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+
+[dependencies]
+ct-merkle = "0.1"
+
+bytes = { workspace = true }
+eyre = { workspace = true }
+hex = { workspace = true, features = ["serde"] }
+serde = { workspace = true, features = ["derive"] }
+sha2 = { workspace = true }
+tendermint = { workspace = true }
```
SuperFluffy, line 14: Noted in the crate. I think this should not have a direct dep on tendermint. Alternatively, the methods that do use tendermint hashes should have a `*tendermint*` or something like that in their name because it's very surprising to find them there.

I do think however that we don't want such a direct dependency.

crates/astria-sequencer-validation/src/proof.rs
```diff
@@ -0,0 +1,214 @@
...
+            .into_iter()
+            .fold(CtMerkleTree::new(), |mut tree, leaf| {
+                tree.push(leaf);
+                tree
+            });
+        MerkleTree(tree)
+    }
+
+    #[must_use]
+    pub fn root(&self) -> Hash {
```
SuperFluffy, line 32: It's very surprising that this returns tendermint types because the crate description does not mention it.

Can this return `[u8; 32]` instead?

crates/astria-sequencer-validation/src/proof.rs
```diff
@@ -0,0 +1,214 @@
...
+use tendermint::Hash;
+
+/// A wrapper around [`ct_merkle::CtMerkleTree`], which uses sha256 as the hashing algorithm
+/// and Vec<u8> as the leaf type.
+#[derive(Debug, Default)]
+pub struct MerkleTree(CtMerkleTree<Sha256, Vec<u8>>);
+
+impl MerkleTree {
+    #[must_use]
+    pub fn from_leaves(leaves: Vec<Vec<u8>>) -> Self {
```
SuperFluffy, line 21: Please provide comments on public interfaces.

crates/astria-sequencer-validation/src/lib.rs
SuperFluffy, line 1: Please provide a short doc comment on top of this documenting what this crate does.

crates/astria-sequencer-validation/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "astria-sequencer-validation"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+ct-merkle = "0.1"
```
SuperFluffy, line 7: This is imported from crates.io, but a git-dep of `ct-merkle` is a workspace dep.

SuperFluffy (changes requested):
I agree with the decision to not expose `CtMerkle` or any of its sibling types directly but provide newtype wrappers around it. This way we can swap it out later if we want (for one, I don't like that it's not maintained in a year).

Before we merge this we should fix the workspace vs direct dependency (I think the git-dep in the workspace might be overlooked)? And also for (de)serialization we should probably use the ct-merkle impl instead of hand-rolling it.

I am torn on using the tendermint `Hash` type directly. On the one hand we have updated conductor to use it. On the other I strongly lean toward tendermint being an implementation detail of `sequencer`. Our other services should not care or know of it.

crates/astria-sequencer-validation/src/proof.rs
```diff
@@ -0,0 +1,214 @@
...
+    value: Vec<u8>,
+    // leaf index of value to be proven
+    idx: usize,
+    // total number of leaves in the tree
+    num_leaves: usize,
+    // the merkle proof itself
+    inclusion_proof: CtInclusionProof<Sha256>,
+}
+
+impl Serialize for InclusionProof {
```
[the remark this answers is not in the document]
noot: yes, i did this in the follow up but forgot to update here, this is updated now

crates/astria-sequencer-validation/src/proof.rs
noot: I actually removed the value from here in the follow up, will update here also. yes the type basically only has Serialize, Deserialize, and verify functions

crates/astria-sequencer-validation/src/proof.rs
noot: yes, updated to no longer depend on tendermint. it's now just a dev dependency since I have the compatibility unit tests

crates/astria-sequencer-validation/Cargo.toml
noot: yeah the one in the workspace is a patch

SuperFluffy: Cool and sorry, didn't notice.

crates/astria-sequencer-validation/src/lib.rs
```diff
@@ -1,3 +1,8 @@
+/// This crate contains types and logic for constructing merkle trees and generating proofs of
```
SuperFluffy, line 1: Change this to `//!` so you get a doc comment on the crate/library.

SuperFluffy (approved):
Looks good, ty for the changes. Note that you want `//!` instead of `///` to document a module.

crates/astria-sequencer/Cargo.toml
```diff
@@ -6,9 +6,11 @@ rust-version = "1.70.0"
 
 [dependencies]
 astria-proto = { path = "../astria-proto" }
+astria-sequencer-validation = { path = "../astria-sequencer-validation" }
 telemetry = { package = "astria-telemetry", path = "../astria-telemetry" }
 
 anyhow = "1"
+digest = "0.10"
```
SuperFluffy, line 13: remove? It's only used in the validation crate?

noot: nice catch!

PR #258, opened 2023-08-15
refactor: move shared types from `astria_sequencer_relayer` to their own crate

noot:
- temporary stopgap PR before types can be properly moved to astria_proto (cc @SuperFluffy )
- move types defined in `astria_sequencer_relayer` that were used in many places into their own crate `astria_sequencer_types`
- types moved are `Namespace`, `SequencerBlockData`, `NamespaceToTxCount` helper type and `default_header()` for testing
- need this as part of #153 as `astria-sequencer` requires the `Namespace` type which would create a circular dependency, I think this also just makes sense as it would eventually be nice to not have conductor import `astria_sequencer_relayer` at all 

partial impl of #27


crates/astria-sequencer-types/src/sequencer_block_data.rs
```diff
@@ -0,0 +1,170 @@
...
+use crate::namespace::Namespace;
+
+/// `SequencerBlockData` represents a sequencer block's data
+/// to be submitted to the DA layer.
+///
+/// TODO: compression or a better serialization method?
+/// TODO: merkle proofs for each rollup's transactions
+#[derive(Serialize, Deserialize, Clone, Debug, Eq, PartialEq)]
+pub struct SequencerBlockData {
+    #[serde(with = "crate::serde::Base64Standard")]
```
SuperFluffy, line 35: None of these should be pub since this type must uphold invariants?

crates/astria-sequencer-types/src/sequencer_block_data.rs
```diff
@@ -0,0 +1,170 @@
...
+    pub rollup_txs: HashMap<Namespace, Vec<Vec<u8>>>,
+}
+
+impl SequencerBlockData {
+    /// Converts the `SequencerBlockData` into bytes using json.
+    ///
+    /// # Errors
+    ///
+    /// - if the data cannot be serialized into json
+    pub fn to_bytes(&self) -> eyre::Result<Vec<u8>> {
```
SuperFluffy, line 50: Remove this. Serialization should be explicit

crates/astria-sequencer-types/src/sequencer_block_data.rs
```diff
@@ -0,0 +1,170 @@
...
+    ///
+    /// # Errors
+    ///
+    /// - if the block has no data hash
+    /// - if a transaction in the block cannot be parsed
+    pub fn from_tendermint_block(b: Block) -> eyre::Result<Self> {
+        use astria_sequencer::transaction::Signed;
+
+        if b.header.data_hash.is_none() {
+            bail!("block has no data hash");
```
SuperFluffy, line 80: Let's type out the errors

crates/astria-sequencer-types/src/sequencer_block_data.rs
```diff
@@ -0,0 +1,170 @@
...
+
+    /// verifies that the merkle root of the tree consisting of all the
+    /// transactions in the block matches the block's data hash.
+    ///
+    /// TODO: this breaks with the update to use Retro; need to update for merkle proofs
+    ///
+    /// # Errors
+    ///
+    /// - unimplemented
+    pub fn verify_data_hash(&self) -> eyre::Result<()> {
```
SuperFluffy, line 123: Remove

crates/astria-sequencer-types/src/sequencer_block_data.rs
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    /// verifies that the merkle root of the tree consisting of the block header
+    /// matches the block's hash.
+    ///
+    /// # Errors
+    ///
+    /// - if the block hash calculated from the header does not match the block hash stored
+    ///  in the sequencer block
+    pub fn verify_block_hash(&self) -> eyre::Result<()> {
```
SuperFluffy, line 134: IMO the type should not be representable if this invariant does not hold.

crates/astria-sequencer-types/src/namespace.rs
```diff
@@ -0,0 +1,134 @@
...
+        Self(inner)
+    }
+
+    /// Creates a namespace from a 10-byte hex-encoded string.
+    ///
+    /// # Errors
+    ///
+    /// - if the string cannot be deocded as hex
+    /// - if the string does not contain 10 bytes
+    pub fn from_string(s: &str) -> eyre::Result<Self> {
```
SuperFluffy, line 54: Remove. This is mainly a representation/serialization concern.

crates/astria-sequencer-types/src/lib.rs
```diff
@@ -0,0 +1,15 @@
...
+pub(crate) mod serde;
+pub mod test_utils;
+
+pub use namespace::{
+    Namespace,
+    DEFAULT_NAMESPACE,
+};
+pub use sequencer_block_data::SequencerBlockData;
+
+pub use crate::serde::{
```
SuperFluffy, line 12: IMO these shouldn't be reexported at the top level.

crates/astria-sequencer-types/src/namespace.rs
```diff
@@ -0,0 +1,134 @@
...
+    }
+}
+
+impl Serialize for Namespace {
+    fn serialize<S: serde::Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+        serializer.serialize_str(&hex::encode(self.0))
+    }
+}
+
+impl<'de> Deserialize<'de> for Namespace {
```
SuperFluffy, line 94: This can be implemented in terms of https://docs.rs/hex/latest/hex/serde/fn.deserialize.html

SuperFluffy (changes requested):
When breaking out modules into separate crates please take that as an opportunity to refactor and pay down technical debt. I have commented on a few instances that can be improved on IMO.

Can we also give this crate a clearer name? This crate is mainly for interacting with sequencer-relayer? Can this be called `sequencer-relayer-rpc`?

crates/astria-sequencer-types/src/namespace.rs
noot: changed to private since it's used in this file for deserialization only

crates/astria-sequencer-types/src/sequencer_block_data.rs
```diff
@@ -0,0 +1,170 @@
...
+        // TODO: don't use json, use our own serializer (or protobuf for now?)
+        serde_json::to_vec(self).wrap_err("failed serializing signed namespace data to json")
+    }
+
+    /// Converts json-encoded bytes into a `SequencerBlockData`.
+    ///
+    /// # Errors
+    ///
+    /// - if the data cannot be deserialized from json
+    /// - if the block hash cannot be verified
```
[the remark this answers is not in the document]
SuperFluffy: My gripe with `from_bytes` is that we are mixing representations. We are using both protobuf and json, and something like `from_bytes` hides this, making it a guessing game what the actual representation is.

So to partially address the issue you can either impl a custom deserializer for for `SequencerBlockData` (which is IMO the correct full solution), or you make use of https://serde.rs/container-attrs.html#try_from as a quick fix where you delegate (de)serialization to a crate-private `priv::SequencerBlockData`, and then construct your real public type with `SequencerBlockData::try_from(priv::SequencerBlockData)` (which the serde attribute does).

SuperFluffy (approved):
After syncing with @noot approving to unblock work on #256, #153 with the actual full refactor by #265.

The changes in this commit should make that step easier.

Issue #161, opened 2023-07-13
relayer: gossip newly proposed blocks; only write finalized blocks


noot:
as @joroshiba said:
> relayer should probably gossip the most recently seen block no matter what. So long as its exceedingly rare to receive a bad block, the latency hit would be a full block time for soft confirmations.
> relayer should only ever write to DA data from 1 block below the current max height. This is final final not maybe final.

currently, the relayer gossips and writes to DA newly proposed blocks, however a newly proposed block is not guaranteed to be final. thus, it should only be written once it's final, ie. there is a commit for it.

relates to #50 on the conductor side, but different issues still.

emhane:
cometbft rpc call for latest block returns finalised block from store
2 /3 majority confirmed
https://github.com/astriaorg/cometbft/blob/342404c9011137cb0e9a4abfd9fa1fc2f9ca889e/consensus/state.go#L1626-L1631
then block stored 
https://github.com/astriaorg/cometbft/blob/342404c9011137cb0e9a4abfd9fa1fc2f9ca889e/consensus/state.go#L1700-L1707

as relayer is a sidecar to the sequencer (running cometbft client) receives blocks on localhost, blocks are guaranteed to come in sequential order with respect to block height.

PR #260, opened 2023-08-15
Finalize sequencer blocks and write to DA

emhane:
Tracking https://github.com/astriaorg/astria/issues/161

Introduces the `QueuedBlocks` struct which has one queue for finalised blocks and another for blocks pending finalisation. Block finalisation is driven by the enqueue operation. When a new block is received from the network finality is computed and timed out pending finality blocks are discarded.

The time out `MAX_QUEUE_TIME` is set to 10s as a placeholder, better informed suggestion as to what a sufficient time out is are appreciated. Since the previous queue `Vec<SequencerBlockData>` didn't appear to have bounds to how many blocks can be queued, I omitted this from `QueuedBlocks` for now.


crates/astria-sequencer-relayer/src/queued_blocks.rs
```diff
@@ -0,0 +1,83 @@
...
+};
+use tracing::warn;
+
+use crate::types::SequencerBlockData;
+
+// the max time to wait for a block to finalize
+const MAX_QUEUE_TIME: Duration = Duration::from_secs(10);
+
+#[derive(Default)]
+pub(crate) struct QueuedBlocks {
```
SuperFluffy, line 21: Commented below why I think this is problematic. My suggestion is to replace this by a binary heap like like so:
```rust
struct QueuedBlocks {
    pending: HashMap<Vec<u8>, SequencerBlockData>>,
    heap: BinaryHeap<HeapValue>,
    expiration_duration: Duration,
}

struct HeapValue {
    key: Vec<u8>,
    instant: Instant,
}

}
impl PartialOrd for HeapValue {
    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
        Some(self.cmp(other))
    }
}
impl Ord for HeapValue {
    fn cmp(&self, other: &Self) -> Ordering {
        self.instant.cmp(&other.instant).reverse()
    }
}
impl<K> PartialEq for HeapValue<K> {
    fn eq(&self, other: &Self) -> bool {
        self.instant.eq(&other.instant)
    }
}
impl<K> Eq for HeapValue<K> {}

impl QueuedBlocks {  
    fn discard(&mut self) {
        let deadline = Instant::now() - self.expiration_duration;
        while let Some(HeapValue { instant, .. }) = self.heap.peek() {
            if *instant > deadline { return; }
            let key = self.heap.pop().unwrap().key;
            self.hash_map.remove(&key);
        }
    }
```

I haven't checked the code and it's missing insertions etc.

crates/astria-sequencer-relayer/src/queued_blocks.rs
```diff
@@ -0,0 +1,83 @@
...
+                        Base64Display::new(block_id, &STANDARD)
+                    );
+                    false
+                }
+            });
+    }
+
+    #[must_use]
+    pub(crate) fn drain_finalized(&mut self) -> Vec<SequencerBlockData> {
+        self.finalized.drain(..).collect()
```
SuperFluffy, line 77: I haven't checked this, but my gut feeling is that simply cloning the vector followed by a call to `Vec::clear` will be cheaper over using an iterator + collect.

crates/astria-sequencer-relayer/src/queued_blocks.rs
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    pub(crate) fn finalized_is_empty(&self) -> bool {
```
SuperFluffy, line 80:
```suggestion
    pub(crate) fn has_finalized(&self) -> bool {
```

SuperFluffy (changes requested):
Thank you for the PR, this seems very useful. I left a couple of suggestions for using a more appropriate data structure.

Also, because this is fairly complex functionality, please provide both unit tests for `QueuedBlocks` and black box tests.

The black box test would probably look like so:

1. send two blocks
2. wait almost up to the timeout
3. send the finalizer for one block
4. wait past the timeout
5. send the finalizer block for the other block
6. assert that the mocked sequencer receives one but not the other.

I know(? not 100% sure) sequencer-relayer attempts to send something to sequencer immediately if no submission task is in flight, so if choreographing this is too tough I am happy to only send 1 block and wait past the timeout to ensure nothing ever reaches sequencer.

crates/astria-sequencer-relayer/src/queued_blocks.rs
```diff
@@ -0,0 +1,83 @@
...
+                .insert(new_block.block_hash.clone(), (new_block, now));
+        }
+
+        self.discard_timed_out(now);
+    }
+
+    fn discard_timed_out(&mut self, current_time: Instant) {
+        // discards blocks that are taking too long to finalize
+        self.pending_finalization
+            .retain(|block_id, (_, insert_time)| {
```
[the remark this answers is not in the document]
emhane: If I order after insert time then that would slow down finding a parent block
```rust
if let Some((parent, _)) = self.pending_finalization.remove(&parent_id) {
                self.finalized.push(parent);
            }
```
of course, I can store an additional queue `insert_time -> block_id`. however since I am iterating through all entires already once, let me do the expiry check then. rn this means cloning all the block hashes which are `Vec<u8>` (heap allocations), so isn't much cheaper then iterating through all the entries again afaik. nonetheless I was meaning to mention that in an issue, that we should at least use `SmallVec<u8>` for the block hashes.

SuperFluffy (changes requested):
This PR looks very useful to put into sequencer-relayer. I have a couple of requests before we merge it:

1. please explain its functionality as doc comments in the code;
2. try to avoid implementing too many single-use interfaces (`TryFrom`, `From`, etc). Prefer inherent methods as they are easier to understand and find.
3. please don't use `()` (or any other error that does not adhere to `std::error::Error`), with the exception of top level error types like `eyre::Report` or `anhow::Error`

crates/astria-sequencer-relayer/Cargo.toml
```diff
@@ -41,18 +41,23 @@ astria-sequencer = { path = "../astria-sequencer" }
...
-astria-sequencer = { path = "../astria-sequencer" }
+
+astria-celestia-jsonrpc-client = { path = "../astria-celestia-jsonrpc-client", features = [
+  "server",
+] }
+astria-sequencer-types = { path = "../astria-sequencer-types", features = [ "relayer-test" ] }
+
+[features]
+default = [ "alpha" ]
+alpha = []
```
noot, line 63: what is this feature for?

crates/astria-sequencer-relayer/src/config/mod.rs
```diff
@@ -14,7 +14,11 @@ use serde::{
 
 mod cli;
 
-const DEFAULT_BLOCK_TIME: u64 = 1000;
+/// Max time in ms to wait for a block to finalize after it is received from the sequencer.
```
noot, line 17: what happens if it does not finalize in that time? also this appears to only be used in tests, can you move it to that module?

crates/astria-sequencer-relayer/src/finalization_pipeline/head_candidate.rs
```diff
@@ -0,0 +1,81 @@
+use astria_sequencer_types::SequencerBlockData;
+use tendermint::Hash;
+use tendermint_rpc::endpoint::block;
+
+#[derive(Clone, Default, Debug)]
+pub(crate) enum HeadCandidate {
+    ProposedByValidator(SequencerBlockData),
+    CometBft(CometBftHeadCandidate),
```
noot, line 8: can you add a comment here briefly explaining each of these?

crates/astria-sequencer-relayer/src/finalization_pipeline/head_candidate.rs
```diff
@@ -0,0 +1,81 @@
...
+    pub(crate) fn new_from_cometbft(block: block::Response) -> Self {
+        Self::CometBft(block.into())
+    }
+
+    pub(crate) fn block_hash(&self) -> Hash {
+        use HeadCandidate::*;
+        match self {
+            ProposedByValidator(block) => block.block_hash(),
+            CometBft(block) => block.block_hash,
+            _ => unreachable!(),
```
noot, line 40: please remove the `Default` and the `unreachable`s since `HeadCandidate::Default` isn't used anywhere

crates/astria-sequencer-relayer/src/finalization_pipeline/head_candidate.rs
```diff
@@ -0,0 +1,81 @@
...
+    pub(crate) fn new_from_validator(block: SequencerBlockData) -> Self {
+        Self::ProposedByValidator(block)
+    }
+
+    pub(crate) fn new_from_cometbft(block: block::Response) -> Self {
+        Self::CometBft(block.into())
+    }
+
+    pub(crate) fn block_hash(&self) -> Hash {
+        use HeadCandidate::*;
```
noot, line 36: is this needed?

crates/astria-sequencer-relayer/src/finalization_pipeline/head_candidate.rs
```diff
@@ -0,0 +1,81 @@
...
+            CometBft(_) => Err(()),
+            _ => unreachable!(),
+        }
+    }
+}
+
+impl HeadCandidate {
+    pub(crate) fn new_from_validator(block: SequencerBlockData) -> Self {
+        Self::ProposedByValidator(block)
+    }
```
noot, line 29: where is it getting this data from? the relayer only gets data in the form of `block::Response`s as it only gets blocks via polling sequencer(cometbft). might be nicer to have the types for this and `new_from_cometbft` be the same, maybe change `new_from_cometbft` to also take `SequencerBlockData`?

crates/astria-sequencer-relayer/src/finalization_pipeline/head_candidate.rs
```diff
@@ -0,0 +1,81 @@
...
+#[derive(Clone, Copy, Debug)]
+pub(crate) struct CometBftHeadCandidate {
+    block_hash: Hash,
+    parent_block_hash: Option<Hash>,
+    height: u64,
+}
+
+impl From<block::Response> for CometBftHeadCandidate {
+    fn from(res: block::Response) -> Self {
+        let block_hash = res.block.header.hash();
```
noot, line 72: prefer `res.block_id.hash`, as this merkleizes the header again

crates/astria-sequencer-relayer/src/finalization_pipeline/mod.rs
```diff
@@ -0,0 +1,194 @@
...
+pub(crate) struct Head {
+    block: PipelineItem,
+}
+
+// pipeline can handle forks 1 height deep
+#[derive(Default)]
+pub(crate) struct FinalizationPipeline {
+    pub(crate) chain_head: Head,
+    pending: HashMap<Hash, PipelineItem>,
+    validator_finalized: Vec<SequencerBlockData>,
```
noot, line 26: can you add comments for each of these?

crates/astria-sequencer-relayer/src/finalization_pipeline/mod.rs
```diff
@@ -0,0 +1,194 @@
...
+
+impl FinalizationPipeline {
+    pub(crate) fn submit(&mut self, new_block: HeadCandidate) {
+        let new_block: PipelineItem = new_block.into();
+
+        match new_block.parent_block_hash() {
+            Some(new_block_parent) => {
+                debug_assert!(new_block.height() > self.chain_head.block.height());
+
+                let steps = new_block.height() - self.chain_head.block.height();
```
noot, line 37: if the new_block is below or at the chain head height, this does nothing - there's a chance we receive two blocks with the same height (in the case where the first one fails to finalize for whatever reason) that case should be handled here

crates/astria-sequencer-relayer/src/finalization_pipeline/mod.rs
```diff
@@ -0,0 +1,194 @@
...
+                    );
+                    for competing_block in pending_at_prev_height.into_values() {
+                        debug_assert!(
+                            competing_block.height() == self.chain_head.block.height() + 1
+                        );
+
+                        if competing_block.block_hash() == new_block_parent {
+                            let old_head = mem::replace(
+                                &mut self.chain_head,
+                                competing_block.canonize().unwrap(),
```
noot, line 55: `unwrap()`? seems unsafe, pls either handle err or explain how it's infallible using `expect`

crates/astria-sequencer-relayer/src/finalization_pipeline/mod.rs
noot, line 26: can rename `validator_finalized` to just `finalized` also

crates/astria-sequencer-relayer/src/finalization_pipeline/mod.rs
```diff
@@ -0,0 +1,194 @@
...
+// head of the sequencer chain as observed on cometbft.
+#[derive(Default)]
+pub(crate) struct Head {
+    block: PipelineItem,
+}
+
+// pipeline can handle forks 1 height deep
+#[derive(Default)]
+pub(crate) struct FinalizationPipeline {
+    pub(crate) chain_head: Head,
```
noot, line 24: why make this pub over having a `fn new(chain_head: Head)`? would prefer the latter

crates/astria-sequencer-relayer/src/finalization_pipeline/head_candidate.rs
```diff
@@ -0,0 +1,81 @@
+use astria_sequencer_types::SequencerBlockData;
+use tendermint::Hash;
+use tendermint_rpc::endpoint::block;
+
+#[derive(Clone, Default, Debug)]
+pub(crate) enum HeadCandidate {
```
noot, line 6: what's the purpose of this enum?

crates/astria-sequencer-relayer/src/finalization_pipeline/mod.rs
```diff
@@ -0,0 +1,194 @@
...
+pub(crate) use pipeline_item::PipelineItem;
+
+// head of the sequencer chain as observed on cometbft.
+#[derive(Default)]
+pub(crate) struct Head {
+    block: PipelineItem,
+}
+
+// pipeline can handle forks 1 height deep
+#[derive(Default)]
```
noot, line 22: remove the `Default`

crates/astria-sequencer-relayer/src/finalization_pipeline/mod.rs
```diff
@@ -0,0 +1,194 @@
...
+use tendermint::Hash;
+
+pub(crate) mod head_candidate;
+pub(crate) mod pipeline_item;
+
+pub(crate) use head_candidate::HeadCandidate;
+pub(crate) use pipeline_item::PipelineItem;
+
+// head of the sequencer chain as observed on cometbft.
+#[derive(Default)]
```
noot, line 16: remove the `Default`, there isn't a natural default for this

crates/astria-sequencer-relayer/src/finalization_pipeline/pipeline_item.rs
```diff
@@ -0,0 +1,87 @@
+use astria_sequencer_types::SequencerBlockData;
+use tendermint::Hash;
+
+use super::{
+    Head,
+    HeadCandidate,
+};
+
+#[derive(Clone, Copy, Default, Debug)]
+enum State {
```
noot, line 10: please add comment explaining this

noot: also remove `Default`, don't think there's a natural default for this

crates/astria-sequencer-relayer/src/finalization_pipeline/pipeline_item.rs
```diff
@@ -0,0 +1,87 @@
...
+
+impl TryInto<SequencerBlockData> for PipelineItem {
+    type Error = ();
+
+    fn try_into(self) -> Result<SequencerBlockData, Self::Error> {
+        self.block.try_into()
+    }
+}
+
+impl PipelineItem {
```
noot, line 39: I'm unclear on the purpose of `PipelineItem`, since blocks are already sorted within the `FinalizationPipeline` as finalized or not, this seems unnecessary

crates/astria-sequencer-relayer/src/relayer.rs
```diff
@@ -56,13 +60,13 @@ pub struct Relayer {
     // A watch channel to track the state of the relayer. Used by the API service.
     state_tx: watch::Sender<State>,
 
-    // Sequencer blocks that have been received but not yet submitted to the data availability
-    // layer (for example, because a submit RPC was currently in flight) .
-    queued_blocks: Vec<SequencerBlockData>,
+    // Sequencer blocks that have been received but not yet finalized. Only finalized blocks are
+    // submitted to the data availability layer.
+    finalization_pipeline: FinalizationPipeline,
```
noot, line 65: the previous comment still stands right?
>     // (for example, because a submit RPC was currently in flight)

where do blocks that are finalized, but waiting for the current `submit_blob` RPC to finish, go? are they still in the `finalization_pipeline`?

crates/astria-sequencer-relayer/src/relayer.rs
```diff
@@ -127,12 +131,12 @@ impl Relayer {
...
         Ok(Self {
             sequencer,
-            sequencer_poll_period: Duration::from_millis(cfg.block_time),
+            sequencer_block_time_ms: Duration::from_millis(cfg.block_time),
             data_availability,
             validator,
             gossip_block_tx,
             state_tx,
-            queued_blocks: Vec::new(),
+            finalization_pipeline: FinalizationPipeline::default(),
```
noot, line 139: seems strange to have the default, would prefer explicitly setting the `head` here

crates/astria-sequencer-relayer/src/relayer.rs
```diff
@@ -455,26 +467,34 @@ impl Relayer {
     }
 }
 
+enum ConversionError {
+    HeightAlreadyCanonical,
```
noot, line 471: maybe rename to `HeightTooLow`?

crates/astria-sequencer-types/src/sequencer_block_data.rs
```diff
@@ -82,10 +98,12 @@ impl SequencerBlockData {
         let Some(data_hash) = header.data_hash else {
             bail!(Error::MissingDataHash);
         };
-        action_tree_root_inclusion_proof
-            .verify(action_tree_root.as_bytes(), data_hash)
-            .wrap_err("failed to verify action tree root inclusion proof")?;
-        // TODO(https://github.com/astriaorg/astria/issues/270): also verify last_commit_hash
+        if !cfg!(feature = "relayer-test") {
```
noot, line 101: pls remove, we should not modify the code for tests 

you can see the tests in this file for how to construct this correctly

noot (changes requested):
overall, this is a good start, the finalization pipeline itself makes sense. I think some of the types/enums used for it can be removed. as well, please remove `Default`s and test-related code, and add more comments!

crates/astria-sequencer-relayer/Cargo.toml
emhane: remains from smthg I tried out

crates/astria-sequencer-relayer/src/config/mod.rs
emhane: obsolete

crates/astria-sequencer-relayer/src/finalization_pipeline/head_candidate.rs
emhane: default of `HeadCandidate` is used when default of `FinalizationPipeline` is used in tests

crates/astria-sequencer-relayer/src/finalization_pipeline/head_candidate.rs
emhane: not necessarily, thought it made the code in the finalisation pipeline easier to read. are you strongly opposed to it instead of calling `block.block.block_hash()` ?

crates/astria-sequencer-relayer/src/finalization_pipeline/head_candidate.rs
emhane: you mean that this is an expensive computation?

crates/astria-sequencer-relayer/src/finalization_pipeline/mod.rs
emhane: yeah sure, excuse me I have poorly documented the whole thing

crates/astria-sequencer-relayer/src/finalization_pipeline/mod.rs
emhane: the `chain_head.height()` here refers to the head of the canonical chain, i.e. the head is finalised. non-finalised blocks are head candidates.

emhane: is it better to use `soft_block` for head and `head` for head candidates like in specs for conductor? https://github.com/emhane/astria/blob/da-finalized-blocks/specs/conductor.md

crates/astria-sequencer-relayer/src/finalization_pipeline/head_candidate.rs
emhane: to be able to make the separation between blocks proposed by this validator which will be submitted to da upon finalisation and blocks that are simply added to the pipeline to verify canonicity

crates/astria-sequencer-relayer/src/finalization_pipeline/mod.rs
emhane: using it for tests! still objections?

crates/astria-sequencer-relayer/src/finalization_pipeline/mod.rs
emhane: I believe it is in fact `Fork` since requirements for this code was that blocks coming from sequencer always point to [forks of] the canonical chain and always arrive in sequential order relative to height

crates/astria-sequencer-relayer/src/finalization_pipeline/pipeline_item.rs
emhane: I believe it is in fact `Fork` since requirements for this code was that blocks coming from sequencer always point to [forks of] the canonical chain and always arrive in sequential order relative to height

crates/astria-sequencer-relayer/src/finalization_pipeline/pipeline_item.rs
emhane: it is used for destructing blocks in the finalisation pipeline into `Head` and `SequencerBlockData`

crates/astria-sequencer-relayer/src/relayer.rs
emhane: the previous comment stands, will add it back, thanks!
finalised blocks stay in the pipeline until it's drained
https://github.com/emhane/astria/blob/4abcbadf354e0b4d4352a027abc00570d70c3d85/crates/astria-sequencer-relayer/src/relayer.rs#L450

crates/astria-sequencer-relayer/src/relayer.rs
emhane: sure, what is the genesis block hash?

emhane: or what else would be explicitly set it to at this point if not default? another option is wrapping the head in `Option<T>`, what do you think?

crates/astria-sequencer-types/src/sequencer_block_data.rs
emhane: sure, added this like a means to keep up with the rapidly changing main branch. needed some control parameter to stay constant so I could keep chaining the functionality adding code in this branch. will fix this as last thing in this pr.

emhane:
thanks for the review @noot. I think main themes I can condense it into are
(i) change to conductor spec terminology `head` for fork heads and `soft` for canonical head
(ii) remove use of default to init where no natural default exists

to address the second point I need to answer, what do should the head init to? or should I wrap it in option?

crates/astria-sequencer-relayer/src/finalization_pipeline/head_candidate.rs
emhane: oh yeah this is needed! matching on the enum variant to get the block hash in the pipeline code is to granular to have a good overview of the pipeline code

emhane:
discoveries in cometbft code made issue obsolete

PR #262, opened 2023-08-15
ci: add timeout, improve caching

joroshiba:
- Adds a timeout for test which has been flakily long running
- Moves clippy to buildjet
- Makes caches all use the cache closest to where running (buildjet only for buildjet cached things)


.github/workflows/rust.yml
```diff
@@ -15,13 +15,16 @@ jobs:
...
+      - uses: Swatinem/rust-cache@v2.6.1
+        with:
+          cache-provider: "buildjet"
       - name: Install buf cli
         uses: bufbuild/buf-setup-action@v1
         with:
           version: "1.17.0"
           github_token: ${{ secrets.GITHUB_TOKEN }}
       - name: Run tests (excluding RPC client, which needs kubernetes)
+        timeout-minutes: 20
```
SuperFluffy, line 27: Just in case, separate the actual test from the building phase so that cold caches don't eat into the test budget, so add an extra step with the `--no-run` flag:
```yaml
      - name: Build tests (excluding RPC client, which needs kubernetes)
        run: |
          cargo test --no-run --release --workspace \
          --all-features \
          --all-targets \
          --exclude astria-celestia-jsonrpc-client
      - name: Run tests (excluding RPC client, which needs kubernetes)
        timeout-minutes: 20
        run: |
          cargo test --release --workspace \
          --all-features \
          --all-targets \
          --exclude astria-celestia-jsonrpc-client
```

SuperFluffy (approved):
I agree with the changes, but I think splitting the test build and test execution might be split up.

Maybe the timeout can then also be reduced to 10 minutes?

Also, should the PR title be `ci: add testing timeouts` to be a bit more specific?

.github/workflows/rust.yml
```diff
@@ -33,7 +36,9 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: dtolnay/rust-toolchain@1.69.0
-      - uses: astriaorg/buildjet-rust-cache@v2.5.1
+      - uses: Swatinem/rust-cache@v2.6.1
+        with:
+          cache-provider: "github"
```
SuperFluffy, line 41: Nice

noot (approved):

specs/composer.md
```diff
@@ -0,0 +1,4 @@
+# Astria Composer
+
+The Astria Composer connects to a rollup mempool and the sequencer to wrap
+rollup transactions in sequencer transactions, and pay for submission.
```
noot, line 4:
```suggestion
rollup transactions into sequencer transactions, and pay for submission.
```

specs/sequencer.md
```diff
@@ -0,0 +1,5 @@
+# Astria Sequencer
+
+The Astria sequencer is an ABCI application built in Rust, using Tower ABCI. It
```
noot, line 3: maybe link to tower-abci?

specs/composer.md
SuperFluffy, line 1: Are the spec docs here by accident? They seem very unrelated.

joroshiba: Yep... will remove belong to a seperate not yet open PR

Issue #267, opened 2023-08-15
Add repo PR template


joroshiba:
Add a PR template to encourage uniformity across our PRs and make it easier to gain context information required to review code.

PR #269, opened 2023-08-15
ci: introduce pr template

joroshiba:
## Summary

Adds a pull request template to the repo.

## Background

Want to improve quality of PR messages, which are by default used for squash/merge commit messages, and make it easier for reviewers to get information they need on PRs.

## Changes

- PR template added

## Testing

N/A, but I used the template here. PR templates only used when exist in default branch.

## Related Issues

closes https://github.com/astriaorg/astria/issues/267


.github/pull_request_template.md
```diff
@@ -0,0 +1,15 @@
...
+
+## Breaking Changelist
+<!-- Delete section if none, bulleted list of breaking changes -->
+- 
+
+## Issue number/link
+<!-- Link to related issues -->
+
+## Test plan
+<!-- How will this be tested -->
```
noot, line 15: I think tests for a feature/fix of a PR should go into the PR, maybe reword this to "how was this tested" or combine it with changelist?

SuperFluffy: Agree with @noot. Every PR should be accompanied by tests (or good arguments why tests don't apply), which is probably what is meant by this section.

But it should be rephrased so we make it clear that this must not be punted to a future PR.

.github/pull_request_template.md
```diff
@@ -0,0 +1,15 @@
...
+## Changelist
+<!-- Bulleted list of changes made -->
+- 
+
+## Breaking Changelist
+<!-- Delete section if none, bulleted list of breaking changes -->
+- 
+
+## Issue number/link
+<!-- Link to related issues -->
```
SuperFluffy, line 12: Should we also mention the github feature on the right hand side (see screenshot)? The text next to that feature is a bit confusing and should say "related issues" or something like that, but I like to make use of it because it's neat.

image

.github/pull_request_template.md
```diff
@@ -0,0 +1,15 @@
+<!--- Provide a general summary of your changes in the Title above -->
+
+## Changelist
+<!-- Bulleted list of changes made -->
+- 
+
+## Breaking Changelist
+<!-- Delete section if none, bulleted list of breaking changes -->
```
SuperFluffy, line 8: IMO this shouldn't be deleted so that reviewers are also reminded to keep this in mind.
```suggestion
<!-- Bulleted list of breaking changes; write "none" if no breaking changes -->
```

SuperFluffy (commented):
This template is good for answering "what?" but what I frequently miss is "why?"

Having a list of changes is good, but frequently becomes too technical ("added type"). Answering "why?" helps reviewers embed a PR in the wider context and also helps future readers understand why certain changes have been made - and it will ultimately help writing changelogs that are more than a myopic rundown of technical changes that lack context.

 "why this feature?" "what is improved with this PR?" "why was this a bug?" "how was the bug observed?"

.github/pull_request_template.md
```diff
@@ -0,0 +1,15 @@
...
+- 
+
+## Issue number/link
```
SuperFluffy, line 11:
```suggestion
## Related issues
```

SuperFluffy (approved):

SuperFluffy (approved):

Issue #240, opened 2023-08-09
Configuration Standards


joroshiba:
Standardize our configuration across services & binaries.

```[tasklist]
### Tasks
- [ ] https://github.com/astriaorg/astria/issues/241
- [ ] https://github.com/astriaorg/astria/issues/242
- [ ] https://github.com/astriaorg/astria/issues/243
- [ ] https://github.com/astriaorg/astria/issues/244
- [ ] https://github.com/astriaorg/astria/issues/245
```

PR #271, opened 2023-08-16
Unify telemetry by replacing telemetry modules by crate

### LINKED ISSUE #170: Replace is-terminal dependency by std::io::IsTerminal trait

SuperFluffy:
## Summary (why this PR?)

This patch replaces all bespoke telemetry modules by one central `astria-telemetry` crate. The different tracing initializers were almost exactly the same, with some minor differences. The `astria-telemetry` crate ensures that tracing is initialized in the same comprehensible way accross all services.

## Changelist

+ introduce the `astria-telemetry` crate;
+ replace all bespoke `telemetry` modules in `astria-{sequencer,sequencer-relayer,composer,conductor}` by the new crate;
+ update `astria-sequencer` and `astria-conductor` to read the tracing directives from the config and the `RUST_LOG` env var (this ensures that this patch is not a breaking change);
+ bump the msrv to `1.70.0` for all services because `astria-telemetry` uses `std::io::IsTerminal` instead of the `is-terminal` crate;
+ bump github actions and pin the docker file to the msrv.

## Related issues

+ the config changes of `astria-conductor` and `astria-sequencer` partially relate to https://github.com/astriaorg/astria/issues/240
+ closes #170 

## Testing

None: can't really be tested, wasn't tested before, and does not constitute n actual code change in `astria-composer`, `astria-sequencer-relayer` that `astria-telemetry` is a verbatim copy of.


noot (approved):
really nice, was thinking about how telemetry was duplicated everywhere, nice update!

joroshiba (approved):

PR #272, opened 2023-08-16
sequencer: use tendermint hash type instead of opaque bytes

sambukowski:
Update `SequencerBlockData`, `SequencerNamespaceData`, and `RollupNamespaceData` to use a `tendermint::Hash` instead of `Vec<u8>`. This cuts down on the amount of type conversions needed when validating and comparing hashes when trying to execute blocks. This also simplifies and fixes the (de)serialization of all types mentioned.

### DIFF
--- a/crates/astria-conductor/src/reader.rs
+++ b/crates/astria-conductor/src/reader.rs
@@ -232,7 +232,7 @@ impl Reader {
                 if let Err(e) = self
                     .block_verifier
                     .validate_rollup_data(
-                        data.data.block_hash.as_bytes(),
+                        data.data.block_hash,
                         &data.data.header,
                         &data.data.last_commit,
                         &rollup_data,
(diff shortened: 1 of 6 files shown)


crates/astria-conductor/src/reader.rs
```diff
@@ -232,7 +232,7 @@ impl Reader {
                 if let Err(e) = self
                     .block_verifier
                     .validate_rollup_data(
-                        &data.data.block_hash,
+                        data.data.block_hash.as_bytes(),
```
SuperFluffy, line 235: Does it make sense to redefine this function to take the hash instead of a byte slice now that all types have been updated?

SuperFluffy (approved):
Good fix, thank you.

Left a comment for the block verifier method to maybe take a hash instead of a byte slice.

Also, please provide a short summary in the PR (and hence commits message) as to why this patch makes sense (in this case fundamentally to fix (de) serialization of `SequencerBlockData`).

crates/astria-conductor/src/reader.rs
sambukowski: Yes. This was an easy change.

joroshiba (approved):

noot (approved):

PR #273, opened 2023-08-16
specs: adding stub specs to be filled in for components

joroshiba:


joroshiba:
Getting real docs instead of stubs.

PR #275, opened 2023-08-16
ci: markdown formatting

### LINKED ISSUE #274: markdown formatting
Markdown files in the repo don't have any universal formatting, can be frustrating to open in code editor.

Add a linter, and update all markdown to be readable in text editors

joroshiba:
## Summary

Adds a markdown linter, enforces via CI, and brings all existing markdown up to standard.

## Background

We had markdown files with a lot of different formatting preferences.

## Changes

- Use `markdownlint-cli2` github action for ci tests
- updated all markdown to be compliant
- README updates
  - section on formatting
  - some content which was stale and formatted incorrectly removed

## Testing

- run markdownlint locally + the added ci jobs

## Related Issues

closes #274


README.md
```diff
@@ -47,158 +62,69 @@ just docker-build astria-sequencer debug
...
-```
-
-Then, start cometbft:
-```sh
-cometbft init
-# set block time to 15s 
-sed -i'.bak' 's/timeout_commit = "1s"/timeout_commit = "15s"/g' ~/.cometbft/config/config.toml
-cometbft start
-```
+The entire stack consists of many different binaries. It's recommended to use the
```
joroshiba, line 65: Note this is actual content change. The directions were no longer accurate and didn't pass formatting rules in many places.

README.md
```diff
@@ -47,158 +62,69 @@ just docker-build astria-sequencer debug
...
+Note that the `astria-proto` generates its code by running tests (and verifying
+that nothing changed). In order for its tests to run you also need
+[Buf](https://buf.build/docs/installation/) installed.
 
-```sh
-./target/release/astria-sequencer-relayer --validator-key-file=$HOME/.cometbft/config/priv_validator_key.json --disable-writing
-```
-
-#### Build and start astria go-ethereum
+## Formatting
```
joroshiba, line 81: Formatting section has been updated.

README.md
```diff
@@ -47,158 +62,69 @@ just docker-build astria-sequencer debug
...
+### Markdown
 
-To run unit tests:
 ```sh
-cargo test
-```
+# Install for macOS w/ homebrew
+brew install markdownlint-cli2
+# Install globally as node package
+npm install markdownlint-cli2 --global
```
SuperFluffy, line 115: It's IMO bad practice to install executable globally from untrusted sources (like npm) as it requires root on Linux systems (unless you use something like volta that changes `{prefix}`). Let's not suggest it here and instead link to https://github.com/DavidAnson/markdownlint-cli2#install

Also because we do it elsewhere, please add that it can be installed from the arch `extra` repos with `sudo pacman -S markdownlint-cli2`

.github/workflows/lint.yml
```diff
@@ -0,0 +1,40 @@
...
+jobs:
+  markdown:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v3
+      - uses: DavidAnson/markdownlint-cli2-action@v11
+        with:
+          globs: '**/*.md'
+
+  rust:
```
SuperFluffy, line 17: I don't know how I feel about pulling this out of the rust-specific workflow. I think it should stay in `rust.yml`. Same with `taplo` being mainly used to enforce rust/crate specific config formatting.

SuperFluffy (approved):
I generally agree with markdown formatting, but I have reservations about moving the crate/rust specific linting out of `rust.yml`

.github/pull_request_template.md
joroshiba, line 1: This file is omitted from linting, I updated to make for more concise squash merge messages.

.github/workflows/reusable-run-checker.yml
```diff
@@ -13,11 +13,14 @@ on:
         description: If lint for protos need to run need to be run, will be 'true'
         value: ${{ jobs.changes.outputs.lint_workflow == 'true' || jobs.changes.outputs.proto == 'true' }}
       run_lint_rust:
-        description: If lint for rust need to run need to be run, will be 'true'
+        description: If lint for rust needs to run need to be run, will be 'true'
```
steezeburger, line 16:
```suggestion
        description: If lint for rust needs to be run, will be 'true'
```

steezeburger: Looks like this was from a copy/paste from the previous `run_lint_proto` and is copied a couple of times below as well.

README.md
```diff
@@ -47,158 +62,69 @@ just docker-build astria-sequencer debug
...
-0a6996ccaca77a1d48633da20062ae051e11c75f3f561dfd2ac01b0c9c874662
-```
-
-Alternatively, in the go-ethereum step, you can add genesis account allocations inside `go-ethereum/genesis.json` (you will need to re-init geth if you've already initialized it).
-
-Navigate to the `localhost:8545` network. Then, you can send transactions as normal. If you check the logs, you can see them move through the stack from sequencer -> relayer -> DA -> conductor -> geth.
-
-#### Clean up local environment
+# Install for macOS
+$ brew install taplo
```
steezeburger, line 101: I think we should remove these `$`. It seems the rule was not correctly triggered because of the comments between the commands.

https://github.com/DavidAnson/markdownlint/blob/main/doc/md014.md

crates/astria-conductor/README.md
```diff
@@ -2,11 +2,11 @@
 
 Coordinates blocks between the data availability layer and the execution layer.
 
-### Running for development
+## Running for development
 
 * create `ConductorConfig.toml` in the repo root and populate accordingly, e.g.
 
-```
+```bash
```
steezeburger, line 9: There are a few instances of ```` ```bash ````, but we mostly use ```` ```sh ````

Do we want to make sure we're consistent with these? It doesn't look like there is a rule to enforce this specifically.

steezeburger (changes requested):

README.md
noot: should we move it somewhere after it's updated? or just remove lol

noot: I can work on updating it if that's helpful, I think it would be

noot (approved):
are we just using the default settings from `markdownlint-cli2` or did we configure them somewhere? doesn't look like we did?

crates/astria-conductor/README.md
noot: yeah we should use ````sh`

sambukowski (approved):

joroshiba:
> are we just using the default settings from `markdownlint-cli2` or did we configure them somewhere? doesn't look like we did?

Default settings with no additional configuration.

steezeburger (approved):

README.md
joroshiba: I don't think we want to actually have this documented long term. We should instead work on making dev cluster the easiest way to test code.

Running microservice clusters all on bare metal same machine is a long way from the production environment, and more likely to test locally say "it works" when it won't in production.

Admittedly we have some work to do on dev-cluster to make this the reality, so I'm torn on documenting it right now.

Issue #276, opened 2023-08-16
composer: collector config should support multiple rollups


itamarreif:
In order to support reading transactions from multiple rollups (#214), the config needs to be changed to parse `chain_id, rpc` tuples for multiple rollups.

A `Rollup` config struct should be added to encompass this tuple and a vector of the tuples should be read from an environment variable as a string and parsed as a regex.

The environment variable should be specified in the format `foo::ws://example.one,bar::https://example.two`

1. Chain ID for the rollup
2. `::` separator between `chain_id` and `rpc`
3. RPC URL for rollup node
4. comma separator between tuples

itamarreif:
this is addressed by #277, #278, #282

SuperFluffy:
Closed by #277

PR #277, opened 2023-08-17
composer: ingest transactions from multiple rollups

SuperFluffy:
## Summary

This patch enables composer to read from multiple rollups that are compatible with the
geth implementation of the ethereum execution API (due to its use of the `eth_subscribe` JSONRPC).

This PR was the original of 3 PRs that that a) added a new config to parse a list of rollups, b) implemented the logic to ingest transactions from different rollups, and c) test the logic. These PRs were subsequently merged into this patch.

## Background

It's a core requirement that composer be able to submit transactions from different rollups to the sequencer using its private key.

## Changes

- parse `ASTRIA_COMPOSER_ROLLUPS` env var that takes a list of the form `<chain_id>::<url>,<chain_id>::<url>`.
- define the `Collector` type to read new pending transactions from an EVM rollup supporting the geth implementation of the ethereum execution API (through `eth_subscribe`).
- Replace the the new pending transaction stream in searcher by an mpsc channel that receives from a set of collector.
- start one collector per chain ID.
- Each collector tags its transaction with its unique chain ID.

## Testing

- The parsing logic comes with many tests for various scenarios.
- The blackbox tests were extended to read from 2 different collectors implemented in terms of two separate mock servers.

## Breaking changes

+ env vars `ASTRIA_COMPOSER_CHAIN_ID` and `ASTRIA_COMPOSER_EXECUTION_URL` are removed;
+ `ASTRIA_COMPOSER_ROLLUPS` must be set.

## Related issues

Original sibling PRs:
- https://github.com/astriaorg/astria/pull/277
- https://github.com/astriaorg/astria/pull/278 (the present PR)
- https://github.com/astriaorg/astria/pull/282

Closes https://github.com/astriaorg/astria/issues/214, https://github.com/astriaorg/astria/issues/276


crates/astria-composer/src/searcher/mod.rs
```diff
@@ -150,16 +154,29 @@ impl Searcher {
     /// + failed to connect to the eth RPC server;
     /// + failed to construct a sequencer clinet
     pub(super) async fn from_config(cfg: &Config) -> eyre::Result<Self> {
+        use rollup::Rollup;
+        let mut rollups = cfg
```
itamarreif, line 158: I feel like this should happen in `Config::from_env()`. The `Config` object should have a `Vec<Rollup>` field, which is what the `Searcher` would use in its constructor.

This way `Config` is validated when it is created, and there isn't a `Config` object that's passed around with an invalid value until it reaches the `Searcher`.

crates/astria-composer/src/searcher/mod.rs
```diff
@@ -150,16 +154,29 @@ impl Searcher {
...
+            .rollups
+            .split(',')
+            .filter(|s| !s.is_empty())
+            .map(|s| Rollup::parse(s).map(Rollup::into_parts))
+            .collect::<Result<HashMap<_, _>, _>>()
+            .wrap_err("failed parsing provided <chain_id>::<url> pairs as rollups")?;
+        ensure!(
```
itamarreif, line 165: We should have a test for this

crates/astria-composer/src/searcher/rollup.rs
```diff
@@ -0,0 +1,135 @@
+//! Parsing strings of the form `<chain_id>::<url>`
+
+use std::fmt;
+
+use once_cell::sync::Lazy;
+use regex::Regex;
+
+#[derive(Debug)]
+pub(super) struct Rollup {
+    chain_id: String,
```
itamarreif, line 10: in a later PR `chain_id` should be made into a non-`String` type to enforce that it is a valid chain id format, similar to what is done in [`sequencer-types`](https://github.com/astriaorg/astria/blob/main/crates/astria-sequencer-types/src/namespace.rs#L25).

or maybe just use that type here?

a `chain_id` here gets mapped directly into a namespace, anyways, as far as im aware

crates/astria-composer/src/config.rs
```diff
@@ -41,11 +41,8 @@ pub struct Config {
     /// Address of the RPC server for the sequencer chain
     pub sequencer_url: String,
 
-    /// Chain ID that we want to connect to
-    pub chain_id: String,
-
-    /// Address of the RPC server for execution
-    pub execution_url: String,
+    /// A list of <chain_id>::<url> pairs
+    pub rollups: String,
```
itamarreif, line 45: Should be a `Vec<Rollup>` imo, see comment on `Searcher::from_config()`

itamarreif (changes requested):

crates/astria-composer/src/searcher/mod.rs
SuperFluffy: This is going to be removed in #278. This serves as a reminder.

crates/astria-composer/src/searcher/rollup.rs
SuperFluffy: Yeah, agree. But that's orthogonal to this PR and requires some more thinking around sequencer's public interfaces/API

crates/astria-composer/src/searcher/mod.rs
SuperFluffy: Will do

SuperFluffy: Because `Config` is a public interface and because all of its fields are public this would require making `Rollup` public, which I don't want to do.

I am going to leave argument as `String`. Another reason to not parse it at config construction is to not worry about serializing config: it is good practice to leave the arguments as-is (or ensure they are serialized to exactly the same form as the input) to give visual feedback to the user that the config was received as provided.

The `SocketAddr` impl ensures that the latter is the case.

crates/astria-composer/src/config.rs
SuperFluffy: Similar argument to above: for hygiene reasons I prefer to leave `Rollup` private to the `searcher` module and not make it crate-private, which prevents this.

itamarreif (approved):

crates/astria-composer/src/searcher/rollup.rs
```diff
@@ -0,0 +1,135 @@
...
+    fn parse_single_rollup_valid() {
+        let rollups = expect_parse_rollups("chain1::http://some.url");
+        assert_eq!(rollups.len(), 1, "\nparsed: {rollups:#?}");
+        assert_eq!(rollups[0].chain_id, "chain1");
+        assert_eq!(rollups[0].url, "http://some.url");
+    }
+
+    #[test]
+    fn parse_mixed_case_chain_id_is_lowercased() {
+        let rollups = expect_parse_rollups("ChAiN1::http://some.url");
```
noot, line 98: we should document that rollup chain IDs are not case sensitive

noot (approved):

joroshiba (approved):
I know this PR was pre-template, would appreciate if we apply it, but at minimum call out the breaking change here.

noot:
does this close #276?

Issue #246, opened 2023-08-09
composer: handle txs rejected by sequencer


noot:
currently composer does not handle txs that fail `check_tx` when submitted to the sequencer, it just logs a warning. after #239 is added one of the potential cases that should be handled is invalid nonce, or if the tx is rejected for some other reason that causes our internally-tracked nonce to get out of sync. 

for example, say we submit txs with nonce 4, 5, and 6, but tx with nonce 4 is rejected. then, composer will try to use nonce 7 for the next submission, but we actually need to resubmit a tx with nonce 4 for 5,6,7 to get included. 

a potential solution here is to make tx signing and submission (and nonce incrementing) atomic/synchronous as currently signing and submission are split (but there is probably a reason they are split up @SuperFluffy ?)

noot:
can be closed - signing + submission was put into one task

PR #278, opened 2023-08-17
composer: ingest transactions from multiple rollups

SuperFluffy:
## Summary

This is second PR of three PRs to support reading from multiple rollups in composer. 

This patch defines the `Collector` type as a long running tasks that read transactions from a different rollup each and submit them to the searcher together with a chain ID. Because the searcher has to keep track of the nonce, it still converts and signs the different transactions when converting them to sequencer blocks.

## Background

This patch enables https://github.com/astriaorg/astria/issues/214

## Changes

- Define the `Collector` type, which reads new pending transactions from an EVM rollup supporting the ethereum execution API.
- Replace the the new pending transaction stream in searcher by an mpsc channel that receives from a set of collector.
- Spin up one collector per `<chain_id>::<url>` pair provided through config.
- Each collector tags its transaction with its unique chain ID.

## Testing

The blackbox tests were extended to read from 2 different collectors implemented in terms of two separate mock servers.

## Related Issues

Sibling PRs:

- https://github.com/astriaorg/astria/pull/277
- https://github.com/astriaorg/astria/pull/278 (the present PR)
- https://github.com/astriaorg/astria/pull/282

- https://github.com/astriaorg/astria/issues/214 (will be closed after issue 277 is merged)


crates/astria-composer/src/searcher/mod.rs
```diff
@@ -255,17 +258,13 @@ impl Searcher {
 
     /// Runs the Searcher
     pub(super) async fn run(mut self) -> eyre::Result<()> {
-        use ethers::providers::{
-            Middleware as _,
-            StreamExt as _,
-        };
-        let wait_for_eth = self.wait_for_eth(5, Duration::from_secs(5), 2.0);
+        self.spawn_executors();
```
itamarreif, line 261: `spawn_collectors`

crates/astria-composer/src/searcher/mod.rs
```diff
@@ -255,17 +258,13 @@ impl Searcher {
...
+        let wait_for_executors = self.wait_for_executors();
```
itamarreif, line 262: `wait_for_collectors`

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -0,0 +1,221 @@
...
+            is_connected: false,
+        }
+    }
+
+    pub(super) fn is_connected(&self) -> bool {
+        self.is_connected
+    }
+}
+
+impl Executor {
```
itamarreif, line 65: rename to `Collector`

itamarreif (changes requested):

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -0,0 +1,221 @@
...
+    debug,
+    info,
+    instrument,
+    warn,
+};
+
+/// A wrapper around an [`ethers::types::Transaction`] that includes the chain ID.
+///
+/// Used to send new transactions to the searcher.
+pub(super) struct Transaction {
```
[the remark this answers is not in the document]
SuperFluffy: IMO not necessary as the searcher only refers to this as `collector::Transaction` now, which I think is as clean as you get. Calling it `RollupTransaction` in addition doesn't help a future reader.

SuperFluffy:
Removed all mention of `executor` in favour of `collector`

crates/astria-composer/src/searcher/collector.rs
```diff
@@ -0,0 +1,221 @@
...
+    debug,
+    info,
+    instrument,
+    warn,
+};
+
+/// A wrapper around an [`ethers::types::Transaction`] that includes the chain ID.
+///
+/// Used to send new transactions to the searcher.
+pub(super) struct Transaction {
```
dereksione, line 30: Nit: Can we add a `new` function to this to build descriptively?

crates/astria-composer/src/searcher/collector.rs
```diff
@@ -0,0 +1,221 @@
...
+
+        let mut tx_stream = client
+            .inner
+            .subscribe_full_pending_txs()
+            .await
+            .wrap_err("failed to subscribe eth client to full pending transactions")?;
+        while let Some(tx) = tx_stream.next().await {
+            match searcher_channel
+                .send_timeout(
+                    Transaction {
```
dereksione, line 113: Nit: Would prefer to use a builder pattern here instead

crates/astria-composer/src/searcher/collector.rs
```diff
@@ -0,0 +1,221 @@
...
+    ///
+    /// # Errors
+    ///
+    /// An error is returned if calling eth failed after `n_retries + 1` times.
+    #[instrument(skip_all, fields(
+    retries.max_number = n_retries,
+    retries.initial_delay = %format_duration(delay),
+    retries.exponential_factor = factor,
+))]
+    async fn wait_for_evm(
```
dereksione, line 153: Nit: Can we break off the connecting parts into a separate module? I would prefer to have a clean collector file that only implements core functionality

crates/astria-composer/src/searcher/collector.rs
```diff
@@ -0,0 +1,221 @@
...
+    }
+}
+
+/// A thin wrapper around [`Provider<Ws>`] to add timeouts.
+///
+/// Currently only provides a timeout around for `get_net_version`.
+/// TODO(https://github.com/astriaorg/astria/issues/216): add timeouts for
+/// `subscribe_full_pendings_txs` (more complex because it's a stream).
+#[derive(Clone, Debug)]
+struct EthClient {
```
dereksione, line 201: Same as above

crates/astria-composer/src/searcher/collector.rs
```diff
@@ -0,0 +1,221 @@
...
+
+    pub(super) fn is_connected(&self) -> bool {
+        self.is_connected
+    }
+}
+
+impl Collector {
+    pub(super) async fn new(
+        chain_id: String,
+        url: String,
```
dereksione, line 68: Imo, the collector should accept a client instead of initializing it (Single responsibility)

dereksione (changes requested):

crates/astria-composer/src/searcher/collector.rs
SuperFluffy: I don't see the need for that. This type is constructed only in one place.

itamarreif:
> Imo, the collector should accept a client instead of initializing it (Single responsibility) 

I disagree - a `Collector` should be responsible for initializing a client, pulling txs from it and forwarding them for bundling. For working with multiple rollups we should have multiple `Collector`s.

crates/astria-composer/src/searcher/collector.rs
itamarreif: I disagree - a `Collector` should be responsible for initializing a client, pulling txs from it and forwarding them for bundling. I don't see why we need an additional abstraction layer between the client and the `Collector`. If we want to use a different client then there should be a different `Collector` implementation around it.

For working with multiple rollups we should have multiple Collectors.

crates/astria-composer/src/searcher/collector.rs
itamarreif: I think the client should belong with the `Collector` implementation.

crates/astria-composer/src/searcher/collector.rs
dereksione: Personally, I would add a layer of abstraction so that collectors being generic is a simple change in the future. Plus, it's not an abstraction at all at this point, you are just initializing somewhere else. Eventually, the collector will need to accept a generic type of a node that it can connect to, which I believe should be a trait. But it's more of a design decision at the end of the day

dereksione: Different collector implementation per client doesn't make sense to me because why would we want people to copy paste wiring code?

itamarreif: I think it's a bit early to address this, and making this change before we have a different client to implement this for adds unneeded complexity. I think that when we deal with a different rollup node we're inevitably going to have to change whatever abstraction we made anyways, so no need to do this here.

crates/astria-composer/src/searcher/mod.rs
```diff
@@ -79,41 +77,13 @@ pub(super) struct Searcher {
 
 #[derive(Debug, Default)]
 pub(crate) struct Status {
-    eth_connected: bool,
+    all_collectors_connected: bool,
```
itamarreif, line 80: thoughts on having this be a count of connected collectors rather than a boolean?

my thinking is that a composer is still healthy if only 1 of n collectors is working - but i'm not sure about the distinction between how `healthz` and `readyz` are used?

my rational here is along the lines that you wouldnt want a single rollup failing to disable your composer because then it would impact the other rollups as well.

crates/astria-composer/src/searcher/collector.rs
SuperFluffy: This is a small module. Making it more granular doesn't help legibility.

crates/astria-composer/src/searcher/collector.rs
SuperFluffy: This is out of scope. This PR is only concerned with reading from different rollups that serve the same RPC.

We should think about generalization once we need it (for example once we create a crate to implement bespoke composers).

crates/astria-composer/src/searcher/collector.rs
SuperFluffy: Answered above

crates/astria-composer/src/searcher/mod.rs
SuperFluffy: Right now this only flips to true once all requested providers respond.

I think this is reasonable for readyz because not having this indicates a problem.

I agree that for an actual status api request or for healthz this should look different.

We can also implement a config to require one or all rollups to be present.

But the current implementation is a direct extension of the previous functionality, so I think this would be out of scope.

Happy to get follow-up PRs though or to discuss how to do this.

itamarreif: sgtm - will make an issue about this

crates/astria-composer/src/searcher/collector.rs
itamarreif, line 153: can we rename this to `wait_for_eth_client`?

itamarreif (changes requested):

crates/astria-composer/src/searcher/collector.rs
SuperFluffy: This function isn't waiting for the client. It's waiting for the server to respond with a certain message.

itamarreif (approved):

dereksione (approved):

crates/astria-composer/src/searcher/collector.rs
```diff
@@ -0,0 +1,221 @@
...
+                        inner: tx,
+                    },
+                    Duration::from_millis(500),
+                )
+                .await
+            {
+                Ok(()) => {}
+                Err(SendTimeoutError::Timeout(_tx)) => {
+                    // TODO(superfluffy): what parts of tx should we log? just its hash?
+                    warn!("timed out sending new transaction to searcher after 500ms; dropping tx");
```
joroshiba, line 124: We should probably log the chainid + the hash. That will uniquely identify the tx.

crates/astria-composer/src/searcher/mod.rs
```diff
@@ -154,24 +124,52 @@ impl Searcher {
...
+
+        let mut create_collectors = rollups
+            .into_iter()
+            .map(|(chain_id, url)| {
+                let task_name = chain_id.clone();
+                tokio::spawn(Collector::new(chain_id, url, tx_sender.clone()))
+                    .map(move |result| (task_name, result))
+            })
+            .collect::<futures::stream::FuturesUnordered<_>>();
+        // TODO(superfluffy): allow aborting this using `futures::stream::AbortHandle`
```
joroshiba, line 150: TODO in code without linked issue

crates/astria-composer/src/searcher/collector.rs
```diff
@@ -0,0 +1,221 @@
...
+    instrument,
+    warn,
+};
+
+/// A wrapper around an [`ethers::types::Transaction`] that includes the chain ID.
+///
+/// Used to send new transactions to the searcher.
+pub(super) struct Transaction {
+    pub(super) chain_id: String,
+    pub(super) inner: ethers::types::Transaction,
```
noot, line 32: this is eth-specific, do we plan to make this generic later?

crates/astria-composer/src/searcher/collector.rs
```diff
@@ -0,0 +1,221 @@
...
+  to connect to eth"
+        );
+        let backoff = ExponentialBuilder::default()
+            .with_min_delay(delay)
+            .with_factor(factor)
+            .with_max_times(n_retries);
+        let version = (|| {
+            let client = self.client.clone(); // This is using `get_net_version` because that's what ethers'
+            // Middleware is // implementing. Maybe the `net_listening` RPC would be better, but
+            // ethers // does not have that.
```
noot, line 174: nit but comment is a bit jank here

crates/astria-composer/src/searcher/mod.rs
```diff
@@ -154,24 +124,52 @@ impl Searcher {
...
+
+        let (tx_sender, new_transactions) = mpsc::channel(256);
+
+        let mut create_collectors = rollups
+            .into_iter()
+            .map(|(chain_id, url)| {
+                let task_name = chain_id.clone();
+                tokio::spawn(Collector::new(chain_id, url, tx_sender.clone()))
+                    .map(move |result| (task_name, result))
+            })
```
noot, line 148: why start all the collectors in tasks instead of just synchronously?

noot (approved):

crates/astria-composer/src/searcher/collector.rs
SuperFluffy: ty, overlooked that!

crates/astria-composer/src/searcher/collector.rs
SuperFluffy: Yeah, I think that's the plan ultimately. But not here. :)

crates/astria-composer/src/searcher/mod.rs
SuperFluffy: Because then we would have to wait for the sum of their startup times. This way it only takes as long as the slowest. It's just better. :-)

crates/astria-composer/src/searcher/collector.rs
SuperFluffy: Logging the hash. `chain_id` was already logged as part of the instrumented scope.

crates/astria-composer/src/searcher/mod.rs
SuperFluffy: Added https://github.com/astriaorg/astria/issues/287

Also noticed a TODO with a missing link underneath. Added https://github.com/astriaorg/astria/issues/246 there as well.

PR #280, opened 2023-08-17
Broad restructuring of our test files

joroshiba:


.github/workflows/release.yml
```diff
@@ -0,0 +1,37 @@
+name: Release
```
noot, line 1: what does this workflow do?

noot (approved):

SuperFluffy (approved):
I haven't seen CI split up this way for rust projects yet. Interested to see how working with this will be like.

.github/workflows/release.yml
joroshiba: Yeah, right now that's the only thing we have that really releases anything, heh, but anything release related could be in here.

IE releasing crates to cargo.

PR #281, opened 2023-08-17
feat(sequencer): faucet module

### LINKED ISSUE #257: sequencer/devnet: implement faucet module
To get funds to users for the devnet, we can add a simple faucet module to the sequencer. This is *not* a feature for mainnet or even testnet, but nice for our initial devnet.

Features:
- an action type that will distribute funds to the specified recipient
- tracks how much has been received by each recipient address in a certain time period
- rate-limit the amount each account can receive per time period (day?)
- ideally can be easily configured to be on/off

noot:
## Summary

- implement faucet module for sequencer

## Background

- we need a way to distribute funds for the devnet, this is probably the simplest solution. users just send a `faucet::Request` action to the sequencer, which will send them funds
- there is a daily limit which resets daily at UTC midnight

## Changes

- faucet module + action type was implemented
- the faucet module is under a feature flag, I think this makes sense as if we want to distribute binaries/images for a specific sequencer network, we can configure at compilation if the faucet is on or off

## Testing

- unit tests

## Breaking Changelist

- none

## Related Issues

closes #257


crates/astria-sequencer/src/faucet/action.rs
```diff
@@ -0,0 +1,41 @@
+use crate::account::types::Balance;
+use astria_proto::{
+    native::sequencer::v1alpha1::Address,
+};
+
+/// Represents a request for funds from the faucet.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Request {
```
SuperFluffy, line 8: Early feedback: after the (un)signed transactions are moved out of astria-sequencer it must no longer export any types except for its config and the `Sequencer` type as an entrypoint to the service.

noot: yeah agreed, however all the different actions for each module needs to be moved to a different crate as well, would rather not do that here. if we know what crate we're moving them to, i can make a PR to move them out

crates/astria-sequencer/src/faucet/action.rs
```diff
@@ -0,0 +1,294 @@
...
+        &self,
+        state: &mut S,
+        _: Address,
+    ) -> Result<()> {
+        let to_balance = state
+            .get_account_balance(self.to)
+            .await
+            .context("failed getting `to` account balance")?;
+        state
+            .put_account_balance(self.to, to_balance + self.amount)
```
SuperFluffy, line 101: Please only use checked arithmetic.

crates/astria-sequencer/src/faucet/action.rs
```diff
@@ -0,0 +1,294 @@
...
+
+        let now = SystemTime::now();
+        if now > u64_to_system_time(info.reset_time) {
+            // we're past the reset time, so reset the amount remaining
+            let reset_time = calculate_reset_time(now)?;
+            state
+                .put_account_info(
+                    self.to,
+                    AccountInfo {
+                        amount_remaining: FAUCET_LIMIT_PER_DAY - self.amount,
```
SuperFluffy, line 117: checked arithmetic

crates/astria-sequencer/src/faucet/action.rs
```diff
@@ -0,0 +1,294 @@
...
+// converts a u64 unix timestamp to a SystemTime
+fn u64_to_system_time(seconds: u64) -> SystemTime {
+    SystemTime::UNIX_EPOCH + std::time::Duration::from_secs(seconds)
+}
+
+// calculates the unix timestamp of the next midnight (UTC time) after the given time
+fn calculate_reset_time(time: SystemTime) -> Result<u64> {
+    const SECONDS_PER_DAY: u64 = 24 * 60 * 60;
+
+    let now_in_seconds = time.duration_since(SystemTime::UNIX_EPOCH)?.as_secs();
```
SuperFluffy, line 148: context

crates/astria-sequencer/src/faucet/action.rs
```diff
@@ -0,0 +1,294 @@
...
+
+        let info = state
+            .get_account_info(self.to)
+            .await
+            .context("failed getting `to` account info")?;
+
+        let now = SystemTime::now();
+        if now > u64_to_system_time(info.reset_time) {
+            // we're past the reset time, so reset the amount remaining
+            let reset_time = calculate_reset_time(now)?;
```
SuperFluffy, line 112: context

crates/astria-sequencer/src/faucet/state_ext.rs
```diff
@@ -0,0 +1,76 @@
...
+    pub(crate) amount_remaining: Balance,
+    // unix timestamp where the amount remaining will be reset
+    pub(crate) reset_time: u64,
+}
+
+impl Default for AccountInfo {
+    fn default() -> Self {
+        Self {
+            amount_remaining: crate::faucet::action::FAUCET_LIMIT_PER_DAY,
+            reset_time: 0,
```
SuperFluffy, line 42: Shouldn't the reset time be the time since midnight?

crates/astria-sequencer/Cargo.toml
```diff
@@ -32,3 +32,7 @@ tendermint-proto = { workspace = true }
 tendermint = { workspace = true }
 tokio = { workspace = true, features = ["rt", "tracing"] }
 tracing = { workspace = true }
+
+[features]
+default = ["faucet"]
```
SuperFluffy, line 37: Can we make this a non-default feature? I prefer that one explicitly opt into using a faucet always.

crates/astria-sequencer/Cargo.toml
```diff
@@ -32,3 +32,7 @@ tendermint-proto = { workspace = true }
...
+faucet = []
```
SuperFluffy, line 38: At first I questioned whether a cfg feature is a good idea, but then I realized you don't want that in production.

I now realize that I would very much like some extra protection around this:

1. require in sequencer's `main` that a `DEVNET=1` env var be set (in other words, we must opt in not only with the feature flag, but also with an explicit config to activate the faucet).
2. in `main`, fire up a task that prints something like `THIS SEQUENCER HAS A FAUCET ACTIVE` every 30 seconds.

SuperFluffy (changes requested):
Please document the `faucet` module at its root to understand how it works.

The reset time handling is tricky. I understand why system time is necessary, at the same time I don't like that this is using system time in an async context. I contemplated if we can spawn a task for the remaining time until midnight to reset the the remaining amount. But that does not seem right (and is unnecessarily complex)

I am wondering if the extra complexity is even necessary for dev net? Couldn't we just permit amounts be issued without using complicated reset logic?

I saw a few fallible ops that could use more context. Also, please only use checked arithmetic when calculating balances.

I am a big fan of the exhaustive tests! Wish they wouldn't need to work around system time, but I can't come up with a better alternative.

noot:
closing in favour of mint module

Issue #250, opened 2023-08-11
feat(conductor): add `execution_commit_level` to configuration


sambukowski:
add the ability to configure the conductor to parse incoming sequencer block data differently based on the `commit_level`

the commit levels correspond to when data gets sent to the rollup.

related to #205  #207  #220  #230

sambukowski:
pr #372

sambukowski:
This issue will also cover the tasks required for completing #345

Issue #230, opened 2023-08-07
conductor: either execute all blocks or offer setting - add `execute_commit_level` setting


joroshiba:
Currently Conductor only executes blocks which have transactions. This is an opinionated take on what constitutes a block to the rollup, rollups can have empty blocks. We should either allow the rollup to make that decision when `execute_block` is called or offer configuration for conductor which allows this behavior to be changed.

sambukowski:
related to: #205 #207

sambukowski:
Closing this as #250 and #285 cover the same scope.

Issue #220, opened 2023-08-04
conductor: reevaluate block queue parent/executed block relations after new v1alpha2 api is implemented


sambukowski:
investigate how the block queue looks at block relationships once #205 has been implemented
specifically look at `Executor::last_executed_seq_block_hash`

sambukowski:
the current implementation of the block queue in #284 does keep this in mind. so the refactor here is likely to be limited.

joroshiba:
Duplicate of #285 + #230

PR #284, opened 2023-08-18
Sb/executor queue

### LINKED ISSUE #15: conductor: queue for block ordering and execution

sambukowski:
## Summary

Adding in a block queue to the executor to handle incoming blocks from the sequencer that may come out of order.

## Background

Previous attempt at implementation here #204 
This impl was initially insufficient in some ways and was also broken when a few other PRs were merged into main.
This PR was also implemented with #207, #220, #230, and #250 in mind.

## Changes

- Added `executor::queue::Queue` type
- Added queue to `executor::Executor`
- Updated `executor::Executor::execute_block()` to incorporate insertion and retrieval of blocks from the queue 

## Testing

See `test_block_queue()` test in `executor::queue`

## Related Issues

#204 

closes #15

### DIFF
--- a/crates/astria-conductor/src/executor/mod.rs
+++ b/crates/astria-conductor/src/executor/mod.rs
@@ -41,6 +41,9 @@ use crate::{
     types::SequencerBlockSubset,
 };
 
+mod queue;
+use crate::executor::queue::Queue;
+
 pub(crate) type JoinHandle = task::JoinHandle<Result<()>>;
 
 /// The channel for sending commands to the executor task.
@@ -117,6 +120,9 @@ struct Executor<C> {
     /// so that we can mark the block as final on the execution layer when
     /// we receive a finalized sequencer block.
     sequencer_hash_to_execution_hash: HashMap<Hash, Vec<u8>>,
+
+    /// The block queue for incoming blocks from gossip
+    block_queue: Queue,
 }
 
 impl<C: ExecutionClient> Executor<C> {
@@ -136,6 +142,7 @@ impl<C: ExecutionClient> Executor<C> {
                 alert_tx,
                 execution_state,
                 sequencer_hash_to_execution_hash: HashMap::new(),
+                block_queue: Queue::new(),
             },
             cmd_tx,
         ))
@@ -214,33 +221,53 @@ impl<C: ExecutionClient> Executor<C> {
             return Ok(Some(execution_hash.clone()));
         }
 
-        let prev_block_hash = self.execution_state.clone();
-        info!(
-            height = block.header.height.value(),
-            parent_block_hash = hex::encode(&prev_block_hash),
-            "executing block with given parent block",
-        );
+        // let prev_block_hash = self.execution_state.clone();
+        // info!(
+        //     height = block.header.height.value(),
+        //     parent_block_hash = hex::encode(&prev_block_hash),
+        //     "executing block with given parent block",
+        // );
 
-        let timestamp = convert_tendermint_to_prost_timestamp(block.header.time)
-            .wrap_err("failed parsing str as protobuf timestamp")?;
+        // let timestamp = convert_tendermint_to_prost_timestamp(block.header.time)
+        //     .wrap_err("failed parsing str as protobuf timestamp")?;
 
-        let response = self
-            .execution_rpc_client
-            .call_do_block(prev_block_hash, block.rollup_transactions, Some(timestamp))
-            .await?;
-        self.execution_state = response.block_hash.clone();
+        self.block_queue.insert(block.clone());
+        let queued_blocks = self.block_queue.get_blocks();
 
-        // store block hash returned by execution client, as we need it to finalize the block later
-        info!(
-            sequencer_block_hash = ?block.block_hash,
-            sequencer_block_height = block.header.height.value(),
-            execution_block_hash = hex::encode(&response.block_hash),
-            "executed sequencer block",
-        );
-        self.sequencer_hash_to_execution_hash
-            .insert(block.block_hash, response.block_hash.clone());
+        let mut final_response = Ok(None);
 
-        Ok(Some(response.block_hash))
+        if let Some(blocks) = queued_blocks {
+            for block in blocks {
+                let prev_block_hash = self.execution_state.clone();
+                info!(
+                    height = block.header.height.value(),
+                    parent_block_hash = hex::encode(&prev_block_hash),
+                    "executing block with given parent block",
+                );
+                let timestamp = convert_tendermint_to_prost_timestamp(block.header.time)
+                    .wrap_err("failed parsing str as protobuf timestamp")?;
+                let response = self
+                    .execution_rpc_client
+                    .call_do_block(prev_block_hash, block.rollup_transactions, Some(timestamp))
+                    .await?;
+                self.execution_state = response.block_hash.clone();
+
+                // store block hash returned by execution client, as we need it to finalize the
+                // block later
+                info!(
+                    sequencer_block_hash = ?block.block_hash,
+                    sequencer_block_height = block.header.height.value(),
+                    execution_block_hash = hex::encode(&response.block_hash),
+                    "executed sequencer block",
+                );
+                self.sequencer_hash_to_execution_hash
+                    .insert(block.block_hash, response.block_hash.clone());
+
+                final_response = Ok(Some(response.block_hash));
+            }
+        }
+
+        final_response
     }
 
     async fn handle_block_received_from_data_availability(
--- a/crates/astria-conductor/src/executor/mod.rs
+++ b/crates/astria-conductor/src/executor/mod.rs
@@ -232,9 +232,9 @@ impl<C: ExecutionClient> Executor<C> {
         //     .wrap_err("failed parsing str as protobuf timestamp")?;
 
         self.block_queue.insert(block.clone());
-        let queued_blocks = self.block_queue.get_blocks();
+        let queued_blocks = self.block_queue.pop_blocks();
 
-        let mut final_response = Ok(None);
+        let mut final_response = Ok(Some(self.execution_state.clone()));
 
         if let Some(blocks) = queued_blocks {
             for block in blocks {
--- a/crates/astria-conductor/src/executor/queue.rs
+++ b/crates/astria-conductor/src/executor/queue.rs
@@ -3,35 +3,31 @@ use std::collections::{
     HashMap,
 };
 
-// use astria_sequencer_types::SequencerBlockData;
 use tendermint::{
     block::Height,
     hash::Hash,
 };
-use tracing::{
-    info,
-    warn,
-};
+use tracing::info;
 
 use crate::types::SequencerBlockSubset;
 
-enum ExecutorQueueParentStatus {
-    ParentPending(Box<SequencerBlockSubset>),
-    ParentSoft,
-    UnknownParent,
-}
-
-/// A queue for the SequencerBlockData type that holds blocks that are
+/// A queue for the SequencerBlockSubset type that holds blocks that are
 /// pending or not yet ready for execution.
+///
+/// This Queue handles all the fork choice logic for incoming Sequencer blocks
+/// from gossip. It is responsible for determining which blocks are safe to
+/// pass on to execution, and holds on to the other data until it is safe to
+/// execute, or deletes it if it is no longer needed or becomes stale or invalid.
+#[derive(Debug, Clone)]
 pub(super) struct Queue {
     head_height: Height,
-    most_recent_soft_hash: Hash, // ??? not sure if we still need this
+    most_recent_soft_hash: Hash,
 
-    // the collection of all pending blocks. the lowest height in this map is
-    // used the head blocks
+    // The collection of all pending blocks. the blocks in this map at Height ==
+    // Queue.head_height are the head blocks
     pending_blocks: HashMap<Height, HashMap<Hash, SequencerBlockSubset>>,
-    // all blocks in order by height that can be considered safe because they
-    // have a parent
+    // All blocks in order by height that can be considered safe because they
+    // have a child block
     soft_blocks: BTreeMap<Height, SequencerBlockSubset>,
 }
 
@@ -47,14 +43,13 @@ impl Queue {
 
     /// Inserts a new block into the ExecutorQueue.
     ///
-    /// Returns None if the block was added to the queue.
-    /// Returns Some(Hash) if the block was already present in the queue. The
-    /// Hash will be the hash of the block that was already present in the queue.
-    // TODO: add error handling
+    /// This is the only way to add data to the queue. When inserting blocks,
+    /// the internal state of the queue will also be updated to properly order
+    /// and arrange all blocks in the queue, base on the tendermin/CometBFT fork
+    /// choice rules.
     pub(super) fn insert(&mut self, block: SequencerBlockSubset) {
         // if the block is already in the queue, return its hash
         if self.is_block_present(&block) {
-            // TODO: do this for all other tracing prints
             info!(
                 block.height = %block.height(),
                 block.hash = %block.block_hash(),
@@ -62,34 +57,85 @@ impl Queue {
             );
         }
 
-        if block.header().height < self.head_height() {
+        // if the block is stale, ignore it
+        if block.header().height < self.head_height {
             info!(
                 block.height = %block.height(),
                 "block is stale and will not be added to the queue"
             );
         }
 
-        match self.check_if_parent_present(block.clone()) {
-            Some(ExecutorQueueParentStatus::ParentPending(parent_block)) => {
-                self.insert_and_update_pending_blocks(block.clone(), *parent_block);
-            }
-            Some(ExecutorQueueParentStatus::ParentSoft) => {
-                self.insert_and_update_soft_queue(block.clone());
-            }
-            // if the block has no parent, add it to the pending blocks without
-            // updating other data
-            Some(ExecutorQueueParentStatus::UnknownParent) => {
-                self.insert_to_pending_blocks(block.clone());
-            }
-            None => {
-                warn!("block doesn't have a parent, discarding");
-            }
-        }
+        // if the block is at the head height OR in the future, just add it to
+        // the pending blocks
+        self.insert_to_pending_blocks(block.clone());
+
+        self.update_internal_state();
+
         info!(
             block.height = %block.height(),
             block.hash = %block.block_hash(),
             "block added to queue"
         );
+    }
+
+    /// Removes and returns all "soft" and "Head" blocks in the queue, inorder
+    /// from oldest to newest.
+    ///
+    /// This function returns an `Option<Vec<SequencerBlockData>>`. A `Some`
+    /// value contains a vector of `SequencerBlockData` that are ready to be
+    /// passed on to execution.
+    /// A `None` value indicates that there are no blocks in the queue that are
+    /// ready to be passed on. A `None` value does not mean there are no blocks
+    /// in the queue.
+    pub(super) fn pop_blocks(&mut self) -> Option<Vec<SequencerBlockSubset>> {
+        let mut output_blocks: Vec<SequencerBlockSubset> = vec![];
+
+        let soft_blocks = self.pop_soft_blocks();
+        if let Some(mut soft_blocks) = soft_blocks {
+            output_blocks.append(soft_blocks.as_mut());
+        }
+        if let Some(mut head_blocks) = self.pop_head_blocks() {
+            output_blocks.append(head_blocks.as_mut());
+        }
+
+        if !output_blocks.is_empty() {
+            Some(output_blocks)
+        } else {
+            None
+        }
+    }
+
+    // TODO: this will return all the blocks in the soft queue that are already "safe"
+    pub(super) fn pop_soft_blocks(&mut self) -> Option<Vec<SequencerBlockSubset>> {
+        let mut soft_blocks: Vec<SequencerBlockSubset> =
+            self.soft_blocks.values().cloned().collect();
+        if !soft_blocks.is_empty() {
+            soft_blocks.sort();
+            self.soft_blocks.clear();
+            let highest_soft_block = soft_blocks[soft_blocks.len() - 1].clone();
+            self.head_height = highest_soft_block.height().increment();
+            self.remove_data_blow_height(self.head_height);
+            Some(soft_blocks)
+        } else {
+            None
+        }
+    }
+
+    fn pop_head_blocks(&mut self) -> Option<Vec<SequencerBlockSubset>> {
+        if let Some(head_blocks) = self.pending_blocks.get_mut(&self.head_height) {
+            let mut output_blocks: Vec<SequencerBlockSubset> = vec![];
+            let tmp_blocks = head_blocks.clone();
+            let mut blocks: Vec<&SequencerBlockSubset> = tmp_blocks.values().collect();
+            blocks.sort();
+            for block in blocks {
+                output_blocks.push(block.clone());
+            }
+            let most_recent_height = output_blocks[output_blocks.len() - 1].height();
+            self.head_height = most_recent_height.increment();
+            self.remove_data_blow_height(self.head_height);
+            return Some(output_blocks);
+        }
+        None
     }
 
     // check to see if the block is already present in the queue
@@ -113,33 +159,6 @@ impl Queue {
         false
     }
 
-    // check if the parent of the incoming block is present in the queue and
-    // return that parent block if it is
-    fn check_if_parent_present(
-        &mut self,
-        block: SequencerBlockSubset,
-    ) -> Option<ExecutorQueueParentStatus> {
-        let parent_height = block.parent_height();
-        let parent_hash = block.parent_hash()?;
-
-        // check if parent in pending data
-        if let Some(pending_blocks) = self.pending_blocks.get(&parent_height) {
-            if let Some(parent_block) = pending_blocks.get(&parent_hash) {
-                return Some(ExecutorQueueParentStatus::ParentPending(Box::new(
-                    parent_block.clone(),
-                )));
-            }
-        }
-        // check if parent in soft data
-        if let Some(soft_block) = self.soft_blocks.get(&parent_height) {
-            if soft_block.block_hash() == parent_hash {
-                return Some(ExecutorQueueParentStatus::ParentSoft);
-            }
-        }
-
-        Some(ExecutorQueueParentStatus::UnknownParent)
-    }
-
     fn is_block_a_parent(&mut self, block: SequencerBlockSubset) -> bool {
         let block_hash = block.block_hash();
         if let Some(child_blocks) = self.pending_blocks.get(&block.child_height()) {
@@ -155,24 +174,6 @@ impl Queue {
         false
     }
 
-    fn update_head_height(&mut self, height: Height) {
-        self.head_height = height;
-    }
-
-    fn update_most_recent_soft_hash(&mut self, hash: Hash) {
-        self.most_recent_soft_hash = hash;
-    }
-
-    fn head_height(&mut self) -> Height {
-        self.head_height
-    }
-
-    // fn head_height_plus_one(&mut self) -> Height {
-    //     Height::try_from(self.head_height.value() + 1).expect("could not convert u64 to Height")
-    // }
-
-    // a basic insert into the pending blocks
-    // TODO: update for error handling
     fn insert_to_pending_blocks(&mut self, block: SequencerBlockSubset) {
         let height = block.height();
         let block_hash = block.block_hash();
@@ -185,64 +186,6 @@ impl Queue {
             let mut new_map = HashMap::new();
             new_map.insert(block_hash, block);
             self.pending_blocks.insert(height, new_map);
-        }
-
-        // TODO: add a call to update the queue
-    }
-
-    // TODO: update description
-    fn insert_and_update_pending_blocks(
-        &mut self,
-        block: SequencerBlockSubset,
-        parent_block: SequencerBlockSubset,
-    ) -> Hash {
-        let height = block.height();
-        let block_hash = block.block_hash();
-        let parent_height = parent_block.height();
-        let parent_hash = parent_block.block_hash();
-
-        // remove parent data from pending blocks
-        if let Some(pending_blocks) = self.pending_blocks.get_mut(&parent_height) {
-            pending_blocks.remove(&parent_hash);
-        }
-        // add parent block to the soft queue
-        self.soft_blocks.insert(parent_height, parent_block);
-        self.update_most_recent_soft_hash(parent_hash);
-
-        // remove all other data below the new incoming block
-        self.remove_data_blow_height(height);
-        // remove all blocks in the head queue that don't have the most recent soft block as their
-        // parent
-        self.clean_head_blocks();
-
-        // check if the new block is a parent of any of the pending blocks
-        if self.is_block_a_parent(block.clone()) {
-            self.remove_data_blow_height(block.child_height());
-            // TODO: update this to used the update and insert to soft blocks function
-            self.soft_blocks.insert(height, block.clone());
-            self.update_most_recent_soft_hash(block_hash);
-            self.clean_head_blocks();
-        } else {
-            // add the new block to the pending blocks
-            self.insert_to_pending_blocks(block);
-        }
-
-        block_hash
-    }
-
-    // remove all blocks in the head queue that don't have the most recent soft
-    // block as their parent
-    fn clean_head_blocks(&mut self) {
-        if let Some(head_blocks) = self.pending_blocks.get_mut(&self.head_height) {
-            let tmp_blocks = head_blocks.clone();
-            let blocks = tmp_blocks.values();
-            for block in blocks {
-                if let Some(parent_hash) = block.parent_hash() {
-                    if parent_hash != self.most_recent_soft_hash {
-                        head_blocks.remove(&block.block_hash());
-                    }
-                }
-            }
         }
     }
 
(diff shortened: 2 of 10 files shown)


crates/astria-conductor/src/executor/queue.rs
```diff
@@ -0,0 +1,504 @@
...
+            let mut new_map = HashMap::new();
+            new_map.insert(block_hash, block);
+            self.pending_blocks.insert(height, new_map);
+        }
+    }
+
+    // remove all data in the queue below a given height. this does not remove
+    // data from the soft queue, only the pending queue, and updates the head height.
+    // TODO: add error handling
+    fn remove_data_blow_height(&mut self, height: Height) {
```
joroshiba, line 195:
```suggestion
    fn remove_data_below_height(&mut self, height: Height) {
```

Not sure if the `blow` was intentional

crates/astria-conductor/src/executor/queue.rs
```diff
@@ -0,0 +1,504 @@
...
+
+    /// This function organizes the internal state of the queue based on the
+    /// tendermint/CometBTF fork choice rules.
+    ///
+    /// Once a block is added to the pending_blocks in the queue, this function
+    /// is called. It walks the panding blocks from lowest to highest height,
+    /// checking to see if there is a continues chain of blocks. For every block
+    /// that is a decendant of the most recent "soft" block, and has a direct
+    /// decendant, that block gets added to the `soft_blocks` BTreeMap and the
+    /// head height is updated.
```
joroshiba, line 215:
```suggestion
    /// is called. It walks the pending blocks from lowest to highest height,
    /// checking to see if there is a continuous chain of blocks. For every block
    /// that is a descendant of the most recent "soft" block, and has a direct
    /// descendant, that block gets added to the `soft_blocks` BTreeMap and the
    /// head height is updated.
```

nit: sp

crates/astria-conductor/src/executor/queue.rs
```diff
@@ -0,0 +1,504 @@
...
+    /// after, and including, the most recent "soft" block that has a direct child
```
joroshiba, line 216: Is this supposed to be somewhere in the above paragraph?

crates/astria-conductor/src/executor/queue.rs
```diff
@@ -0,0 +1,504 @@
...
+            };
+            block.header.last_block_id = Some(block_id);
+
+            blocks.push(block);
+        }
+        blocks
+    }
+
+    #[tokio::test]
+    async fn test_block_queue() {
```
joroshiba, line 376: split into multiple tests instead of one large test?

crates/astria-conductor/src/types.rs
```diff
@@ -1,22 +1,56 @@
...
+    }
+}
+
+impl PartialOrd for SequencerBlockSubset {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+// TODO: write a text to check this for the hash invariant
```
joroshiba, line 40: Should probably be done in this PR?

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -189,14 +196,21 @@ impl<C: ExecutionClient> Executor<C> {
...
+    /// the block has already been executed, the previously-computed execution block
+    /// hash is returned. If there are no relevant transactions in the SequencerBlock, it returns
+    /// None. If multiple blocks end up being executed (because several blocks were in
+    /// the queue waiting for a missing block to arrive), only the execution
+    /// hash of the last block is returned. The execution hash of all executed
+    /// blocks are stored in the `sequencer_hash_to_execution_hash` map.
     async fn execute_block(&mut self, block: SequencerBlockSubset) -> Result<Option<Vec<u8>>> {
+        // TODO: (GHI 285 - https://github.com/astriaorg/astria/issues/285): decide how empty blocks should be handled. either do nothing,
+        // or execute an empty block.
+        // if there are no transactions in the block, return None
```
noot, line 213: I don't see a need for executing empty blocks - I think we can ignore them, unless someone sees a good use case for empty rollup blocks?

crates/astria-conductor/src/executor/queue.rs
sambukowski: this was a typo. it's been removed

crates/astria-conductor/src/types.rs
sambukowski: tests added!

crates/astria-conductor/src/executor/queue.rs
```diff
@@ -0,0 +1,672 @@
...
+        }
+
+        if !output_blocks.is_empty() {
+            Some(output_blocks)
+        } else {
+            None
+        }
+    }
+
+    // TODO: this will return all the blocks in the soft queue that are already "safe"
```
joroshiba, line 108: What's the todo here?

specs/conductor.md
```diff
@@ -41,11 +41,46 @@ The responsibilities of each module are as follows:
...
+   consensus agreement will be sent to the execution layer and will not be
+   reorged. Internally this means that the queue will hold all blocks at the
+   head height, but only return blocks that have a child. Thus all blocks sent
+   to the execution layer are always soft. Blocks are marked as FIRM when they are seen in DA. 
+3. `FIRM`: The FIRM setting indicates that only blocks that have been written
+   and propagated across the DA network will be sent to the execution layer. All
+   blocks sent will always be FIRM.
+
+
+
```
joroshiba, line 83: nit: excess new lines

specs/conductor.md
```diff
@@ -41,11 +41,46 @@ The responsibilities of each module are as follows:
...
  - Filters out transactions by their rollup namespace and sends them to the rollup for execution (`soft` commits)
  - Catalogs and matches the block hashes received from the gossip network and the DA by rollup namespace to send `firm` commits to the rollup
  - Blocks are sent to the execution layer using [Astria’s GRPC Execution client interface](https://buf.build/astria/astria/docs/main:astria.execution.v1)
      - Rollups utilizing the Conductor must implement this interface
 
+#### The Execution Queue
+The Queue within the Executor is responsible for verifying the ordering of
+blocks received from the Sequencer. The basic flow for the blocks through the
+Executor and into and out of the Queue, is as follows:
+**From the Sequencer:**
```
joroshiba, line 55: nit: if should be on new line, need two line breaks

specs/conductor.md
```diff
@@ -41,11 +41,46 @@ The responsibilities of each module are as follows:
...
+1. Blocks are received from the Sequencer and are validated.
```
joroshiba, line 56: nit: should be a new line before list

crates/astria-conductor/src/executor/queue.rs
```diff
@@ -0,0 +1,672 @@
...
+    /// and arrange all blocks in the queue, based on the Tendermint/CometBFT fork
+    /// choice rules.
+    pub(super) fn insert(&mut self, block: SequencerBlockSubset) {
+        // if the block is already in the queue, return its hash
+        if self.is_block_present(&block) {
+            info!(
+                block.height = %block.height(),
+                block.hash = %block.block_hash(),
+                "block is already present in the queue"
+            );
```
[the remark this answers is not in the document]
sambukowski: Yup. Fixed.

crates/astria-conductor/src/executor/queue.rs
```diff
@@ -0,0 +1,672 @@
...
+                block.height = %block.height(),
+                "block is stale and will not be added to the queue"
+            );
+        }
+
+        // if the block is at the head height OR in the future, just add it to
+        // the pending blocks
+        self.insert_to_pending_blocks(block.clone());
+
+        self.update_internal_state();
```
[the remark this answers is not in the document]
sambukowski: Good catch. `update_internal_state` has been moved to inside `insert_to_pending_blocks`

crates/astria-conductor/src/executor/queue.rs
sambukowski: Changes this to a doc comment.

specs/conductor.md
sambukowski: I removed the bold heading and reworded things a little and added the new lines.

crates/astria-conductor/src/executor/queue.rs
```diff
@@ -0,0 +1,672 @@
...
+                "block is already present in the queue"
+            );
+        }
+
+        // if the block is stale, ignore it
+        if block.header().height < self.head_height {
+            info!(
+                block.height = %block.height(),
+                "block is stale and will not be added to the queue"
+            );
```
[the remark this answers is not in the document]
sambukowski: added a return here as well.

crates/astria-conductor/src/executor/queue.rs
```diff
@@ -0,0 +1,683 @@
...
+    ///
+    /// Once a block is added to the pending_blocks in the queue, this function
+    /// is called. It walks the pending blocks from lowest to highest height,
+    /// checking to see if there is a continuous chain of blocks. For every block
+    /// that is a descendant of the most recent "soft" block, and has a direct
+    /// descendant, that block gets added to the `soft_blocks` BTreeMap and the
+    /// head height is updated.
+    fn update_internal_state(&mut self) {
+        // check if the block added connects blocks in the pending queue
+        let mut heights: Vec<Height> = self.pending_blocks.keys().cloned().collect();
```
SuperFluffy, line 234: Prefer `copied` over `collect`

crates/astria-conductor/src/executor/queue.rs
SuperFluffy, line 1: Please add a module doc comment that explains this module. Especially the difference between `pending` and `soft` blocks. Do I understand correctly that `soft` blocks are also `pending` blocks, but that also have children?

crates/astria-conductor/src/executor/queue.rs
```diff
@@ -0,0 +1,683 @@
...
+    /// This function organizes the internal state of the queue based on the
+    /// tendermint/CometBTF fork choice rules.
+    ///
+    /// Once a block is added to the pending_blocks in the queue, this function
+    /// is called. It walks the pending blocks from lowest to highest height,
+    /// checking to see if there is a continuous chain of blocks. For every block
+    /// that is a descendant of the most recent "soft" block, and has a direct
+    /// descendant, that block gets added to the `soft_blocks` BTreeMap and the
+    /// head height is updated.
+    fn update_internal_state(&mut self) {
```
SuperFluffy, line 232: This function performs reallocation of a two dimensional data structure on potentially every iteration of a loop.

Furthermore, adding new blocks to the `pending_blocks` queue probably happens very frequently, so this function does a lot of unnecessary work.

You can fix this by making use of a `loop {}` instead of iterating over the heights directly (and thus holding on to a reference of into `self`, making it impossible to also mutate it.

First I think it's necessary to change `pending_blocks` from a `HashMap` to a `BTreeMap` so that it's self-ordering (which IMO makes sense because you almost always collect + sort it anyway). Then you can make use of the fact that you are effectively using `head_height` as the iteration index.

Then use `loop {}`, using `head_height` to index into `self.pending_blocks`, checking if you have a parent, mutating the inner state, etc. Finally, increment `head_height` and define a stopping condition (for example, `head_height > self.pending_blocks.last_key_value.0`).

crates/astria-conductor/src/executor/queue.rs
```diff
@@ -0,0 +1,683 @@
...
+            // if the first height in the queue is not the head height just stop reorg
+            } else {
+                break;
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+mod test {
```
SuperFluffy, line 261: Can you write tests for this queue in isolation to demonstrate usage and edge cases? The tests can be smaller and more focused and don't need a mocked executor.

SuperFluffy (changes requested):

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -214,6 +231,37 @@ impl<C: ExecutionClient> Executor<C> {
             return Ok(Some(execution_hash.clone()));
         }
 
+        let mut final_response = Ok(Some(self.execution_state.clone()));
+        // TODO: (GHI 250 - https://github.com/astriaorg/astria/issues/250):
+        // if blocks are coming from DA, do we still need them to pass through
+        // the queue? if not, we can skip the insert and pop_blocks calls
```
noot, line 237: yeah blocks from DA should still go through the queue as there could be gaps/out of order

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -214,6 +231,37 @@ impl<C: ExecutionClient> Executor<C> {
...
+        // add a match statement here to either `pop_blocks` (returns soft and
+        // head) or `pop_soft_blocks` (just soft) based on the `execution_commit_level` setting
+        let queued_blocks = self.block_queue.pop_blocks();
+
+        if let Some(blocks) = queued_blocks {
+            // execute all the blocks returned from the queue
+            for block in blocks {
+                final_response = self.execute_single_block(block).await;
+            }
+        }
```
noot, line 251: this ignores/overwrites the error if one of these blocks fails to execute, please check err after every call

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -214,6 +231,37 @@ impl<C: ExecutionClient> Executor<C> {
             return Ok(Some(execution_hash.clone()));
         }
 
+        let mut final_response = Ok(Some(self.execution_state.clone()));
+        // TODO: (GHI 250 - https://github.com/astriaorg/astria/issues/250):
+        // if blocks are coming from DA, do we still need them to pass through
+        // the queue? if not, we can skip the insert and pop_blocks calls
+        if let Ok(None) = self.block_queue.insert(block.clone()) {
```
noot, line 238: `block.clone()` D: any way we can avoid this?

noot: also can you add a comment explaining this line?

crates/astria-conductor/src/executor/queue.rs
```diff
@@ -0,0 +1,683 @@
...
+};
+use tracing::info;
+
+use crate::types::SequencerBlockSubset;
+
+/// A queue for the SequencerBlockSubset type that holds blocks that are
+/// pending or not yet ready for execution.
+///
+/// This Queue handles all the fork choice logic for incoming Sequencer blocks
+/// from gossip. It is responsible for determining which blocks are safe to
```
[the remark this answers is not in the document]
noot: > which blocks are safe to

safe? or just ready to be executed?

crates/astria-conductor/src/executor/queue.rs
```diff
@@ -0,0 +1,683 @@
...
+    }
+
+    /// Inserts a new block into the ExecutorQueue.
+    ///
+    /// This is the only way to add data to the queue. When inserting blocks,
+    /// the internal state of the queue will also be updated to properly order
+    /// and arrange all blocks in the queue, based on the Tendermint/CometBFT fork
+    /// choice rules.
+    pub(super) fn insert(&mut self, block: SequencerBlockSubset) -> Result<Option<Hash>> {
+        // if the block is already in the queue, return its hash
```
noot, line 52: why? i'd return `None` here instead, so to differentiate from the case where it's added

crates/astria-conductor/src/executor/queue.rs
```diff
@@ -0,0 +1,683 @@
...
+        if self.is_block_present(&block) {
+            info!(
```
noot, line 54: change to `debug`?

crates/astria-conductor/src/executor/queue.rs
```diff
@@ -0,0 +1,683 @@
...
+                block.height = %block.height(),
+                block.hash = %block.block_hash(),
+                "block is already present in the queue"
+            );
+            return Ok(Some(block.block_hash()));
+        }
+
+        // if the block is stale, ignore it
+        if block.header().height < self.head_height {
```
noot, line 63: the naming here is a bit confusing - only blocks higher than the `head_height` are added to the queue? can you rename `head_height` to `base_height` or `safe_height` something? also from what i understand, all blocks in the queue should be a descendant of this "root" block (the highest safe block), is there a check for that?

crates/astria-conductor/src/executor/queue.rs
```diff
@@ -0,0 +1,683 @@
...
+
+    /// Return all the blocks at the head height
+    fn pop_head_blocks(&mut self) -> Option<Vec<SequencerBlockSubset>> {
+        if let Some(head_blocks) = self.pending_blocks.get_mut(&self.head_height) {
+            let mut output_blocks: Vec<SequencerBlockSubset> = vec![];
+            let tmp_blocks = head_blocks.clone();
+            let mut blocks: Vec<&SequencerBlockSubset> = tmp_blocks.values().collect();
+            blocks.sort();
+            for block in blocks {
+                output_blocks.push(block.clone());
```
noot, line 142: why `clone()`? you should have ownership of the blocks?

crates/astria-conductor/src/executor/queue.rs
```diff
@@ -0,0 +1,683 @@
...
+        } else {
+            None
+        }
+    }
+
+    /// Return all the blocks at the head height
+    fn pop_head_blocks(&mut self) -> Option<Vec<SequencerBlockSubset>> {
+        if let Some(head_blocks) = self.pending_blocks.get_mut(&self.head_height) {
+            let mut output_blocks: Vec<SequencerBlockSubset> = vec![];
+            let tmp_blocks = head_blocks.clone();
```
noot, line 138: please remove the `clone()`, take the values out of the hash map and re-insert them later if needed

crates/astria-conductor/src/executor/queue.rs
```diff
@@ -0,0 +1,683 @@
...
+        }
+        self.update_internal_state();
+    }
+
+    // remove all data in the queue below a given height. this does not remove
+    // data from the soft queue, only the pending queue, and updates the head height.
+    // TODO: add error handling
+    fn remove_data_below_height(&mut self, height: Height) {
+        // remove all data below the new incoming block from the pending data
+        let tmp_pending = self.pending_blocks.clone();
```
noot, line 213: no clone pls, just remove the values you want to get rid of directly

crates/astria-conductor/src/executor/queue.rs
```diff
@@ -0,0 +1,683 @@
...
+    /// Safely insert a block into the pending blocks. This function will insert
+    /// the block into existing maps or create a new one if needed.
+    fn insert_to_pending_blocks(&mut self, block: SequencerBlockSubset) {
+        let height = block.height();
+        let block_hash = block.block_hash();
+
+        if let Some(pending_blocks) = self.pending_blocks.get_mut(&height) {
+            let mut new_map = pending_blocks.clone();
+            new_map.insert(block_hash, block);
+            self.pending_blocks.insert(height, new_map);
```
noot, line 199: remove and re-insert the map so you don't have to clone: https://doc.rust-lang.org/stable/std/collections/struct.HashMap.html#method.remove

crates/astria-conductor/src/executor/queue.rs
```diff
@@ -0,0 +1,683 @@
...
+pub(super) struct Queue {
+    head_height: Height,
+    most_recent_soft_hash: Hash,
+
+    // The collection of all pending blocks. the blocks in this map at Height ==
+    // Queue.head_height are the head blocks
+    pending_blocks: HashMap<Height, HashMap<Hash, SequencerBlockSubset>>,
+    // All blocks in order by height that can be considered safe because they
+    // have a child block
+    soft_blocks: BTreeMap<Height, SequencerBlockSubset>,
```
noot, line 32: at a high level, i don't understand the difference/need to differentiate between soft and pending here (or what they are even)?

crates/astria-conductor/src/executor/queue.rs
```diff
@@ -0,0 +1,683 @@
...
+            // if the very first height in the pending blocks is the head height
+            if height == self.head_height {
+                if let Some(pending_blocks) = self.pending_blocks.clone().get(&height) {
+                    // walk the pending blocks at that height and check if any of them are a parent
+                    for block in pending_blocks.values() {
+                        if self.is_block_a_parent(block.clone()) {
+                            self.soft_blocks.insert(height, block.clone());
+                            self.most_recent_soft_hash = block.block_hash();
+                            self.head_height = height.increment();
+                            self.remove_data_below_height(self.head_height);
```
noot, line 247: why are blocks dropped here? what if they need to be executed still?

crates/astria-conductor/src/executor/queue.rs
```diff
@@ -0,0 +1,683 @@
...
+    /// descendant, that block gets added to the `soft_blocks` BTreeMap and the
+    /// head height is updated.
+    fn update_internal_state(&mut self) {
+        // check if the block added connects blocks in the pending queue
+        let mut heights: Vec<Height> = self.pending_blocks.keys().cloned().collect();
+        heights.sort();
+        // walk the pending blocks starting from the head height
+        for height in heights {
+            // if the very first height in the pending blocks is the head height
+            if height == self.head_height {
```
noot, line 239: readability:
```suggestion
            if height != self.head_height {
                continue
            }
            // code here
```

crates/astria-conductor/src/executor/queue.rs
```diff
@@ -0,0 +1,683 @@
...
+                if let Some(pending_blocks) = self.pending_blocks.clone().get(&height) {
+                    // walk the pending blocks at that height and check if any of them are a parent
+                    for block in pending_blocks.values() {
+                        if self.is_block_a_parent(block.clone()) {
+                            self.soft_blocks.insert(height, block.clone());
+                            self.most_recent_soft_hash = block.block_hash();
```
noot, line 245: this seems weird; since `pending_blocks` is a hash map it just picks some random hash out of the blocks at this height for the `most_recent_soft_hash`?

crates/astria-conductor/src/executor/queue.rs
noot: ah okay, i see this only affects the "pending" map, can you update the function name to reflect that?

crates/astria-conductor/src/executor/queue.rs
noot: yes i would like this, can you test for cases such as:
1. adding a block 1 height above the committed height but the parent hash is not the committed block (should not get added)
2. adding a block N+2 where the committed height is N, then committing N+1 which is not a parent of N+2 (N+2 should be removed)
3. committing a block at some height should remove all other blocks at that height and below

by "committing" above i mean finalizing, but i'm not sure if that corresponds to "head/soft" in this queue or not

crates/astria-conductor/src/types.rs
```diff
@@ -1,23 +1,57 @@
...
+impl std::hash::Hash for SequencerBlockSubset {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        self.block_hash.hash(state);
+        self.header.hash().hash(state);
+
+        let mut transactions = self.rollup_transactions.clone();
+        transactions.sort();
```
noot, line 48: why are these sorted? i don't think they should be

crates/astria-conductor/src/types.rs
```diff
@@ -41,4 +75,152 @@ impl SequencerBlockSubset {
             rollup_transactions: our_rollup_data.transactions,
         })
     }
+
+    /// Return the block hash.
+    pub fn block_hash(&self) -> Hash {
```
noot, line 80: why are these all pub?

crates/astria-conductor/src/types.rs
```diff
@@ -41,4 +75,152 @@ impl SequencerBlockSubset {
...
+    pub fn height(&self) -> Height {
+        self.header().height
+    }
+
+    /// Get the height of the block's parent.
+    ///
+    /// # Panics
+    ///
+    /// This function will panic if the block height is less than or equal to 1.
+    /// Only the genesis block has a height of 1, and all other blocks must have
```
noot, line 100: genesis block height is 0

crates/astria-sequencer-types/src/sequencer_block_data.rs
```diff
@@ -276,6 +277,48 @@ impl SequencerBlockData {
...
+        );
+        Height::try_from(self.header().height.value() - 1)
+            .expect("should have been able to decrement tendermint height")
+    }
+
+    /// Get the height of the block's child, or the next block.
+    #[must_use]
+    pub fn child_height(&self) -> Height {
+        self.height().increment()
+    }
```
noot, line 308: what's the purpose of this (and parent_height) over just getting the height and adding/subtracing in the caller?

crates/astria-sequencer-types/src/sequencer_block_data.rs
```diff
@@ -276,6 +277,48 @@ impl SequencerBlockData {
...
+    /// Will return `Some(Hash)` if the block has a parent hash.
+    /// Will return `None` if the block does not have a parent hash. This is the case for the
+    /// genesis block.
+    #[must_use]
+    pub fn parent_hash(&self) -> Option<Hash> {
+        if let Some(parent_hash) = self.header().last_block_id {
+            return Some(parent_hash.hash);
+        }
+        None
+    }
```
noot, line 321: same here, why not just use `header().last_block_id` in the caller?

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -214,6 +231,37 @@ impl<C: ExecutionClient> Executor<C> {
...
+    /// Send a block to the execution layer for execution.
+    ///
+    /// This function takes a SequencerBlockSubset and its associated data to
+    /// build the inputs for a call to `DoBlock` on the execution layer.
+    async fn execute_single_block(
+        &mut self,
+        block: SequencerBlockSubset,
+    ) -> Result<Option<Vec<u8>>> {
+        // get the previous execution state to pass to do block
         let prev_block_hash = self.execution_state.clone();
```
noot, line 265: this is going to error if 2 blocks at the same height are executed one after the other; from what i understand `self.block_queue.pop_blocks()` can return multiple blocks at the same height, this should handle that

noot (changes requested):
at a high level, i don't understand the need for `soft` and `pending` maps in the queue, i think it would be much simpler to just have one map of height -> vec of pending blocks, and have an interface like:

`insert_block(block: SequencerBlockSubset) -> Option<Vec<SequencerBlockSubset>>` which inserts a block, and if inserting this block makes other blocks ready for execution, returns those blocks

`finalize_block(block_hash: Hash)` which marks a block "final" and prunes any blocks that are clearly not a descendant of it

the queue itself doesn't need to differentiate between commitment levels, so i think removing that will make the code a lot simpler.

at a code level, please get rid of clones, especially of any block data as that's potentially a lot being cloned!

crates/astria-conductor/src/types.rs
sambukowski: fixed!

crates/astria-conductor/src/executor/queue.rs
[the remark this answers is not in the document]
sambukowski: The queue will be used for everything (both gossip or da), and it is just for blocks that are ready to be executed.

crates/astria-conductor/src/executor/queue.rs
```diff
@@ -0,0 +1,683 @@
...
+            returned_soft_blocks.sort();
+            self.soft_blocks.clear();
+            let highest_soft_block = returned_soft_blocks[returned_soft_blocks.len() - 1].clone();
+            self.head_height = highest_soft_block.height().increment();
+            self.remove_data_below_height(self.head_height);
+            Some(returned_soft_blocks)
+        } else {
+            None
+        }
+    }
```
emhane, line 132:
```suggestion
    pub(super) fn pop_soft_blocks(&mut self) -> Option<Vec<SequencerBlockSubset>> {
        if !self.soft_blocks.is_empty() {
            let returned_soft_blocks = std::mem::replace(&mut self.soft_blocks, BTreeMap::new());
            let mut returned_soft_blocks: Vec<SequencerBlockSubset> =
                returned_soft_blocks.into_values().collect();
            returned_soft_blocks.sort();
            let highest_soft_block = &returned_soft_blocks[returned_soft_blocks.len() - 1];
            self.head_height = highest_soft_block.height().increment();
            self.remove_data_below_height(self.head_height);
            Some(returned_soft_blocks)
        } else {
            None
        }
    }
```
this is probably the most efficient way to avoid allocating memory on the heap with clone. also, good to check `!self.soft_blocks.is_empty()` as early as possible to return early.

crates/astria-conductor/src/executor/queue.rs
```diff
@@ -0,0 +1,683 @@
...
+/// from gossip. It is responsible for determining which blocks are safe to
+/// pass on to execution, and holds on to the other data until it is safe to
+/// execute, or deletes it if it is no longer needed or becomes stale or invalid.
+#[derive(Debug, Clone)]
+pub(super) struct Queue {
+    head_height: Height,
+    most_recent_soft_hash: Hash,
+
+    // The collection of all pending blocks. the blocks in this map at Height ==
+    // Queue.head_height are the head blocks
```
emhane, line 28: head candidates is probably a more precise name to use through out the code. then we have that, a chain can have many head candidates and only one head, a head is chosen among head candidates.

crates/astria-conductor/src/types.rs
```diff
@@ -41,4 +75,152 @@ impl SequencerBlockSubset {
...
+    pub fn parent_height(&self) -> Height {
+        assert!(
+            self.height().value() > 1,
+            "block height must be greater than 1"
+        );
+        Height::try_from(self.header().height.value() - 1)
+            .expect("should have been able to decriment tendermint height")
+    }
+
+    /// Get the height of the block's child, or the next block.
```
emhane, line 112:
```suggestion
    /// Get the next height relative to this block.
```
I thought this was a little confusing at first. "or the next block" intends the case when a block doesn't have a child, is that right?

crates/astria-conductor/src/executor/queue.rs
emhane: I am assuming this is being thought about in a thread safe way, and hence why atomically updating the entry for the `height` key in pending blocks is valuable. since hash map isn't thread safe by itself, I don't think it's necessary at this stage. I suggest avoiding the extra heap allocation with clone:
```suggestion
pending_blocks.insert(block_hash, block);
```

crates/astria-conductor/src/executor/queue.rs
```diff
@@ -0,0 +1,683 @@
...
+            new_map.insert(block_hash, block);
+            self.pending_blocks.insert(height, new_map);
+        }
+        self.update_internal_state();
+    }
+
+    // remove all data in the queue below a given height. this does not remove
+    // data from the soft queue, only the pending queue, and updates the head height.
+    // TODO: add error handling
+    fn remove_data_below_height(&mut self, height: Height) {
```
emhane, line 211: rn this is always called with `self.head_height` passed as param, I think we can just load `self.head_height` inside the scope of the function.

crates/astria-conductor/src/executor/queue.rs
emhane, line 247: this is the only place that `self.head_height` should have to be incremented and as a follow, the only place where `self.remove_data_below_height()` should have to be called. this is the only place that can move a block from the pending queue to soft queue so the only occurrence that can advance `head_height`.

crates/astria-conductor/src/executor/queue.rs
```diff
@@ -0,0 +1,683 @@
...
+                            break;
+                        }
+                    }
+                }
+            // if the first height in the queue is not the head height just stop reorg
+            } else {
+                break;
+            }
+        }
+    }
```
emhane, line 257:
```suggestion
    fn update_internal_state(&mut self) {
        // check if the block added connects blocks in the pending queue
        loop {
            let head_height = self.head_height;

            let Some(head_candidates) = self.pending_blocks.get(&head_height) else {
                return; // if the head height is not in the queue yet just stop reorg
            };
            // walk the pending blocks at that height and check if any of them are a parent
            let mut new_soft_block_hash = None;
            for block in head_candidates.values() {
                if self.is_block_a_parent(block) {
                    new_soft_block_hash = Some(block.block_hash());
                    break;
                }
            }
            if let Some(block_hash) = new_soft_block_hash {
                let head_candidates = self.pending_blocks.get_mut(&self.head_height).unwrap();
                let block = head_candidates.remove(&block_hash).unwrap().clone();
                self.soft_blocks.insert(self.head_height, block);
                self.most_recent_soft_hash = block_hash;
                self.head_height = self.head_height.increment();
                self.remove_data_below_height(self.head_height);
            }

            if head_height == self.head_height {
                // head is at height of chain
                return;
            }
        }
    }
```
removes the soft block from pending blocks

crates/astria-conductor/src/executor/queue.rs
```diff
@@ -0,0 +1,683 @@
...
+        async fn call_init_state(&mut self) -> Result<InitStateResponse> {
+            let hasher = sha2::Sha256::new();
+            Ok(InitStateResponse {
+                block_hash: hasher.finalize().to_vec(),
+            })
+        }
+    }
+
+    /// Return the number of blocks in the queue
+    fn queue_len(queue: Queue) -> usize {
```
emhane, line 335:
```suggestion
    fn queue_len(queue: &Queue) -> usize {
```
passing references as params saves a lot of clones

crates/astria-conductor/src/executor/queue.rs
```diff
@@ -0,0 +1,683 @@
...
+                return true;
+            }
+        }
+
+        false
+    }
+
+    /// Check if there is another block in the pending blocks at a lower height
+    /// that points to this block as its parent.
+    fn is_block_a_parent(&mut self, block: SequencerBlockSubset) -> bool {
```
emhane, line 175:
```suggestion
    fn is_block_a_parent(&self, block: SequencerBlockSubset) -> bool {
```

emhane:
nice! I really like how you solved this so simply by checking chain links one height at a time. from my understanding of the documented scope, competing forks shouldn't be acknowledged, but unordered arrival over gossipnet should be accommodated for. that is, the new soft head is decided on a fcfs basis, rather than for example 'longest chain' basis.

a soft head is chosen on a fcfs basis given that it satisfies the conditions: (i) has a direct descendant (ii) is direct descendant of `most_recent_soft_hash`. I noticed the second condition isn't checked anywhere rn, i would add it into `update_internal_state`.
```rust
if self.is_block_a_parent(block) && block.parent_hash() == Some(self.most_recent_soft_hash) {..}
```

asides that, working more with references as params will save a lot of clones.

crates/astria-conductor/src/executor/queue.rs
emhane: hm, I think I am porting this from somewhere else, let's stick to the terminology originating from geth terminology, `head` and `soft` for canonical head

crates/astria-conductor/src/executor/queue.rs
emhane: also here, better call `head_candidates` `soft_candidates`

crates/astria-conductor/src/executor/queue.rs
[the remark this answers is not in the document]
sambukowski: updated this to a reference.

crates/astria-sequencer-types/src/sequencer_block_data.rs
```diff
@@ -265,6 +266,48 @@ impl SequencerBlockData {
...
+
+    /// Get the height of the block's parent.
+    ///
+    /// # Panics
+    ///
+    /// This function will panic if the block height is less than or equal to 1.
+    /// Only the genesis block has a height of 1, and all other blocks must have
+    /// a larger height.
+    #[must_use]
+    pub fn parent_height(&self) -> Height {
```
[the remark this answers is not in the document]
sambukowski: i removed a few of these functions from SequencerBlockData. They were there originally because we weren't using SequencerBlockSubset yet and I forgot to removed them once that was added.

crates/astria-conductor/src/executor/queue.rs
```diff
@@ -0,0 +1,683 @@
...
+        }
+    }
+
+    /// Inserts a new block into the ExecutorQueue.
+    ///
+    /// This is the only way to add data to the queue. When inserting blocks,
+    /// the internal state of the queue will also be updated to properly order
+    /// and arrange all blocks in the queue, based on the Tendermint/CometBFT fork
+    /// choice rules.
+    pub(super) fn insert(&mut self, block: SequencerBlockSubset) -> Result<Option<Hash>> {
```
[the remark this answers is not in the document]
sambukowski: Pulled out the `Result`, just using the `Option` now.

crates/astria-conductor/src/executor/queue.rs
```diff
@@ -0,0 +1,683 @@
...
+    /// Removes and returns all "soft" and "Head" blocks in the queue, inorder
+    /// from oldest to newest.
+    ///
+    /// This function returns an `Option<Vec<SequencerBlockData>>`. A `Some`
+    /// value contains a vector of `SequencerBlockData` that are ready to be
+    /// passed on to execution.
+    /// A `None` value indicates that there are no blocks in the queue that are
+    /// ready to be passed on. A `None` value does not mean there are no blocks
+    /// in the queue.
+    pub(super) fn pop_blocks(&mut self) -> Option<Vec<SequencerBlockSubset>> {
```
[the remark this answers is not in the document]
sambukowski: Renamed `pop` to `drain` and also updated the functions to use the Iterator return.

crates/astria-conductor/src/executor/queue.rs
```diff
@@ -0,0 +1,683 @@
...
+    /// value contains a vector of `SequencerBlockData` that are ready to be
+    /// passed on to execution.
+    /// A `None` value indicates that there are no blocks in the queue that are
+    /// ready to be passed on. A `None` value does not mean there are no blocks
+    /// in the queue.
+    pub(super) fn pop_soft_blocks(&mut self) -> Option<Vec<SequencerBlockSubset>> {
+        let mut returned_soft_blocks: Vec<SequencerBlockSubset> =
+            self.soft_blocks.values().cloned().collect();
+        if !returned_soft_blocks.is_empty() {
+            returned_soft_blocks.sort();
```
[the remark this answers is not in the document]
sambukowski: Removed.

crates/astria-conductor/src/executor/queue.rs
```diff
@@ -0,0 +1,683 @@
...
+/// A queue for the SequencerBlockSubset type that holds blocks that are
+/// pending or not yet ready for execution.
+///
+/// This Queue handles all the fork choice logic for incoming Sequencer blocks
+/// from gossip. It is responsible for determining which blocks are safe to
+/// pass on to execution, and holds on to the other data until it is safe to
+/// execute, or deletes it if it is no longer needed or becomes stale or invalid.
+#[derive(Debug, Clone)]
+pub(super) struct Queue {
+    head_height: Height,
```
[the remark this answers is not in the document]
sambukowski: Comment added.

crates/astria-conductor/src/executor/queue.rs
```diff
@@ -0,0 +1,683 @@
...
+    most_recent_soft_hash: Hash,
```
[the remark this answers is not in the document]
sambukowski: Comment added.

crates/astria-conductor/src/executor/queue.rs
```diff
@@ -0,0 +1,683 @@
...
+    /// ready to be passed on. A `None` value does not mean there are no blocks
+    /// in the queue.
+    pub(super) fn pop_soft_blocks(&mut self) -> Option<Vec<SequencerBlockSubset>> {
+        let mut returned_soft_blocks: Vec<SequencerBlockSubset> =
+            self.soft_blocks.values().cloned().collect();
+        if !returned_soft_blocks.is_empty() {
+            returned_soft_blocks.sort();
+            self.soft_blocks.clear();
+            let highest_soft_block = returned_soft_blocks[returned_soft_blocks.len() - 1].clone();
+            self.head_height = highest_soft_block.height().increment();
```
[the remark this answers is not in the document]
sambukowski: Turns out this wasn't needed here. Looking over things again, the head height was always getting set to latest soft height + 1. So the increment here was just setting head_height to the same value.

crates/astria-conductor/src/executor/queue.rs
```diff
@@ -0,0 +1,683 @@
...
+        if let Some(head_blocks) = self.pending_blocks.get_mut(&self.head_height) {
+            let mut output_blocks: Vec<SequencerBlockSubset> = vec![];
+            let tmp_blocks = head_blocks.clone();
+            let mut blocks: Vec<&SequencerBlockSubset> = tmp_blocks.values().collect();
+            blocks.sort();
+            for block in blocks {
+                output_blocks.push(block.clone());
+            }
+            let most_recent_height = output_blocks[output_blocks.len() - 1].height();
+            self.head_height = most_recent_height.increment();
```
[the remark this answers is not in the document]
sambukowski: This is no longer needed in `pop_soft_blocks` and was removed there. I'll add more context for this in some comments.

crates/astria-conductor/src/types.rs
```diff
@@ -1,23 +1,57 @@
...
+    },
     Hash,
 };
 
 /// `SequencerBlockSubset` is a subset of a SequencerBlock that contains
 /// information required for transaction data verification, and the transactions
 /// for one specific rollup.
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Eq, PartialEq)]
 pub struct SequencerBlockSubset {
```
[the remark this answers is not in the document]
sambukowski: Everything in the queue is `pub(super)` and I updated everything that touches `SequencerBlockSubset` to be `pub(crate)`.

crates/astria-conductor/src/executor/queue.rs
sambukowski: Completely changed this function so don't need to worry about this now.

crates/astria-conductor/src/executor/queue.rs
sambukowski: Module comment added.

sambukowski:
Closing this. Recent changes and refactors made in the Sequencer and DA sync implementations cover everything needed here.
Sequencer sync pr: https://github.com/astriaorg/astria/pull/455
DA Sync pr: https://github.com/astriaorg/astria/pull/575

PR #286, opened 2023-08-18
feat(sequencer): implement proving of action tree root inside `Header::data_hash`

noot:
## Summary

- implement generation of merkle inclusion proof of the action tree commitment (first "tx" in a sequencer block) inside the block header's `data_hash` (aka transactions root)

## Background

- required for validation of data inside conductor

## Changes

- update `SequencerBlockData` and `SequencerNamespaceData` to contain `action_tree_root` and `action_tree_root_inclusion_proof`
- `SequencerBlockData::from_tendermint_block` now generates the inclusion proof
- `SequencerBlockData::new()` checks the inclusion proof

## Testing

- unit tests

## Related Issues

#153 (does not close yet, follow up will close it)


crates/astria-sequencer-types/src/sequencer_block_data.rs
```diff
@@ -140,15 +176,22 @@ impl SequencerBlockData {
...
 
+        let action_tree_root =
+            Hash::try_from(b.data[0].clone()).wrap_err("failed to parse action tree root")?;
+
         // we unwrap sequencer txs into rollup-specific data here,
         // and namespace them correspondingly
         let mut rollup_data = HashMap::new();
 
-        for (index, tx) in b.data.iter().enumerate() {
+        for (index, tx) in b.data[1..].iter().enumerate() {
```
SuperFluffy, line 194: Can you add a note here why the first one is skipped? Maybe just reference the construction?

crates/astria-sequencer-validation/src/proof.rs
```diff
@@ -70,55 +62,27 @@ pub struct InclusionProof {
...
+impl Eq for InclusionProof {}
 
 impl InclusionProof {
     /// Verify that the proof is valid for the given root hash.
     ///
     /// # Errors
     ///
     /// - if the proof is invalid
-    pub fn verify(&self, root_hash: Hash) -> eyre::Result<()> {
+    pub fn verify(&self, value: &[u8], root_hash: Hash) -> eyre::Result<()> {
```
SuperFluffy, line 81: Does this still match the doc comment? What is value?

This method name is confusing to me. Is there proof verified? Is the data verified? Can you add a doc test that demonstrates how this should be used?

SuperFluffy (changes requested):
This is a soft request for changes.

I feel like a lot of going on here and it's not clear from the public interfaces how this should be used.

Can you add an example doc comment / test on how to construct a proof and how to use it to verify it (or some other data?).

Also, the relatively innocently named `from_tendermint` constructor is doing a lot more now. Can you please update its doc?

noot:
> Also, the relatively innocently named from_tendermint constructor is doing a lot more now 

yeah, this was basically so that `SequencerNamespaceData` can only be constructed if valid. I can update the doc

> I feel like a lot of going on here and it's not clear from the public interfaces how this should be used.

this PR is the one that actually uses the proofs created: https://github.com/astriaorg/astria/pull/290 I can merge these two if that's clearer but I figured they can be separate PRs

> Can you add an example doc comment / test on how to construct a proof and how to use it to verify it (or some other data?). 

yes will add an example

crates/astria-sequencer-validation/src/proof.rs
noot: updated the comment slightly - it's verifying a merkle proof of inclusion, so proving that the `value` is a leaf inside the merkle tree where the root is `root_hash`. not sure how much info I should add, since I thought the type was somewhat explanatory haha

can add a doc comment though

noot:
@SuperFluffy usage example added, feel free to review this again!

crates/astria-conductor/src/types.rs
```diff
@@ -22,7 +22,12 @@ impl SequencerBlockSubset {
         data: SequencerBlockData,
         namespace: Namespace,
     ) -> Self {
-        let (block_hash, header, _, mut rollup_txs) = data.into_values();
+        // we don't need to verify the action tree root here,
+        // as [`SequencerBlockData`] would not be constructable
+        // if it was invalid
+
```
joroshiba, line 28: nit: remove empty line between code + comment?

crates/astria-sequencer-types/src/sequencer_block_data.rs
```diff
@@ -63,15 +72,31 @@ impl SequencerBlockData {
...
     ) -> eyre::Result<Self> {
+        // perform data validations to ensure only valid [`SequencerBlockData`]
+        // can be constructed
+        let Some(data_hash) = header.data_hash else {
+            bail!(Error::MissingDataHash);
+        };
+        action_tree_root_inclusion_proof
+            .verify(action_tree_root.as_bytes(), data_hash)
+            .wrap_err("failed to verify action tree root inclusion proof")?;
+        // TODO: also verify last_commit_hash (#270)
```
joroshiba, line 86: nit: full link preferable

crates/astria-sequencer-types/src/sequencer_block_data.rs
```diff
@@ -97,12 +122,23 @@ impl SequencerBlockData {
 
     #[allow(clippy::type_complexity)]
     #[must_use]
-    pub fn into_values(self) -> (Hash, Header, Option<Commit>, HashMap<Namespace, RollupData>) {
+    pub fn into_values(
+        self,
```
joroshiba, line 126: Is this formatting correct? Feels weird to push the self to new line because the return is

crates/astria-sequencer-validation/Cargo.toml
```diff
@@ -4,7 +4,7 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-ct-merkle = "0.1"
+ct-merkle = { version = "0.1", features = ["serde"] }
```
joroshiba, line 7: Should this be using the workspace version?

SuperFluffy, line 7: `ct-merkle` is imported here, but there is a `ct-merkle` in the workspace cargo toml that refers to a git repo under the astria org.

crates/astria-sequencer-types/src/sequencer_block_data.rs
```diff
@@ -42,14 +46,19 @@ pub struct RollupData {
...
 pub struct SequencerBlockData {
     pub(crate) block_hash: Hash,
     pub(crate) header: Header,
     /// This field should be set for every block with height > 1.
     pub(crate) last_commit: Option<Commit>,
     /// namespace -> rollup data (chain ID and transactions)
     pub(crate) rollup_data: HashMap<Namespace, RollupData>,
+    /// The root of the action tree for this block.
+    pub(crate) action_tree_root: Hash,
+    /// The inclusion proof that the action tree root in included
```
SuperFluffy, line 59: typo:
```suggestion
    /// The inclusion proof that the action tree root is included
```

crates/astria-sequencer-types/src/sequencer_block_data.rs
SuperFluffy: Please provide a full link to the issue or PR, like so:
```suggestion
        // TODO(https://github.com/astriaorg/astria/issues/270): also verify last_commit_hash
```

crates/astria-sequencer-types/src/sequencer_block_data.rs
```diff
@@ -97,12 +122,23 @@ impl SequencerBlockData {
 
     #[allow(clippy::type_complexity)]
     #[must_use]
-    pub fn into_values(self) -> (Hash, Header, Option<Commit>, HashMap<Namespace, RollupData>) {
+    pub fn into_values(
```
SuperFluffy, line 125: This is a code smell. We should differentiate between a "raw" `SequencerBlockData` as a (de)serialization format (where it's ok to make all fields pub), and "verified" `SequencerBlockData` that upholds invariants (where you want to prevent anyone pulling the rug from under you). Permitting `clippy::type_complexity` proves that point.

My suggestion: keep the `SequencerBlockData` as is, but introduce a `RawSequencerBlockData` that has exactly the same fields, but comes with a `RawSequencerBlockData::verify(self) -> eyre::Result<SequencerBlockData>`. This way you can keep using `SequencerBlockData` everywhere as is.

Next, scrap `SequencerBlockData::into_values` and introduce `SequencerBlockData::into_raw`. This way users can do the following instead of destructuring tuples:

```rust
fn do_something(data: SequencerBlockData) {
    let RawSequencerBlockData {
        block_hash,
        ...
    } = data.into_raw();
    further_processing(block_has);
}
```

We then have an easier path forward to move this into `astria-proto`.

crates/astria-sequencer-types/src/sequencer_block_data.rs
```diff
@@ -136,19 +172,36 @@ impl SequencerBlockData {
     /// # Errors
     ///
     /// - if the block has no data hash
+    /// - if the block has no transactions
```
SuperFluffy, line 175: This list is great when eventually moving to typed errors, which we should really do...

crates/astria-sequencer-types/src/sequencer_block_data.rs
```diff
@@ -170,11 +223,25 @@ impl SequencerBlockData {
...
+        let tx_tree = MerkleTree::from_leaves(b.data);
+        let calculated_data_hash = tx_tree.root();
+        ensure!(
+            // this should never happen for a correctly-constructed block
+            calculated_data_hash == data_hash,
+            "action tree root does not match the first transaction in the block",
+        );
+        let action_tree_root_inclusion_proof = tx_tree
+            .prove_inclusion(0) // action tree root is always the first tx in a block
+            .expect("failed to generate inclusion proof for action tree root");
```
SuperFluffy, line 236: Looks like this whould have been a `wrap_err`

SuperFluffy (changes requested):
Saw a few things that should be fixed prior to accepting.

~~Orthogonal to this PR I had a chance to read the wrapper of `CtMerkle`. I was wondering why you chose to implement this as a newtype wrapper instead of using an extension trait (or simply utility functions)?~~

~~The difficulty with wrappers is often that one ends up reimplementing more and more of the methods of the wrapped type, without providing proper documentation. That could be avoided by simply using the type directly.~~

EDIT: After reading #256 and thinking it through I agree with not directly exposing ctmerkle and its types.

crates/astria-sequencer-types/src/sequencer_block_data.rs
noot: yeah I didn't really like this either but wasn't sure if I should just make another type with all pub fields. will update!

crates/astria-sequencer-types/src/sequencer_block_data.rs
```diff
@@ -95,15 +118,16 @@ impl SequencerBlockData {
...
-    #[allow(clippy::type_complexity)]
     #[must_use]
-    pub fn into_values(self) -> (Hash, Header, Option<Commit>, HashMap<Namespace, RollupData>) {
-        (
-            self.block_hash,
-            self.header,
-            self.last_commit,
-            self.rollup_data,
-        )
+    pub fn into_raw(self) -> RawSequencerBlockData {
```
SuperFluffy, line 122: Please add a comment on top of this function.

Nit: destructuring `self ` looks slightly nicer (less `self.`)
```rust
    let Self {
        block_hash,
        header,
        // and the rest
    } = self;
    RawSequencerBlockData {
        block_hash,
        header,
        // and the rest
    }
```

crates/astria-sequencer-types/src/sequencer_block_data.rs
```diff
@@ -42,14 +46,19 @@ pub struct RollupData {
 
 /// `SequencerBlockData` represents a sequencer block's data
 /// to be submitted to the DA layer.
-#[derive(Serialize, Deserialize, Clone, Debug, Eq, PartialEq)]
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq)]
 pub struct SequencerBlockData {
     pub(crate) block_hash: Hash,
```
SuperFluffy, line 51: Do you still need all of this to be `pub(crate)`?

crates/astria-sequencer-types/src/sequencer_block_data.rs
```diff
@@ -42,14 +46,19 @@ pub struct RollupData {
 
 /// `SequencerBlockData` represents a sequencer block's data
 /// to be submitted to the DA layer.
-#[derive(Serialize, Deserialize, Clone, Debug, Eq, PartialEq)]
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq)]
```
SuperFluffy, line 49: serde provides this attribute: https://serde.rs/container-attrs.html#try_from

Can you implement a `TryFrom<RawSequencerBlockData>` (I'd prefer to not impl a single-use `TryFrom` and instead use a custom deserializer, but in this case it's probably fine).

This way you can still deserialize, but in addition you get verification. Right now you still serialize directly into a `SequencerBlockData` and have no assurance that it's legitimate (invariants upheld, etc).

Also see https://serde.rs/container-attrs.html#into for the inverse

SuperFluffy (approved):
Looks good to me. Please address the comment re deserialization so that there is no backdoor into creating types where invariants are violated (in this case inclusion proofs).

crates/astria-sequencer/Cargo.toml
```diff
@@ -10,6 +10,7 @@ astria-sequencer-validation = { path = "../astria-sequencer-validation" }
 telemetry = { package = "astria-telemetry", path = "../astria-telemetry" }
 
 anyhow = "1"
+digest = "0.10"
```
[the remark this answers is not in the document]
noot: woops haha

crates/astria-sequencer-types/src/sequencer_block_data.rs
noot: no :D

crates/astria-sequencer-types/src/sequencer_block_data.rs
noot: nice, added!

PR #290, opened 2023-08-21
feat(conductor): implement `RollupNamespaceData` verification inside conductor

### LINKED ISSUE #153: design and re-implement verification of transactions in a block

noot:
## Summary

implement `RollupNamespaceData` verification inside conductor

## Background

required for validation of rollup data read from DA inside conductor

## Changes

- implement verification of the `action_tree_root_inclusion_proof` inside conductor, proving that the action tree root is correct and was included in a block (proves that `action_tree_root` is in `header.data_hash`)
- implement verification of the rollup data inside `action_tree_root`, proving that the rollup data read from DA was correct (no txs included incorrectly or txs missing and ordering is correct)

## Testing

- unit tests 
- running entire stack (todo)

## Breaking Changelist

- requires the relayer to implement writing inclusion proofs (which will be merged before this)

## Related Issues

closes  #153


crates/astria-conductor/src/block_verifier.rs
```diff
@@ -103,18 +104,64 @@ impl BlockVerifier {
...
+        let rollup_data_tree = MerkleTree::from_leaves(rollup_data.rollup_txs.clone());
+        let rollup_data_root = rollup_data_tree.root();
+        let mut leaf = rollup_data.chain_id.clone();
+        leaf.append(&mut rollup_data_root.to_vec());
+
+        rollup_data
+            .inclusion_proof
+            .verify(&leaf, sequencer_namespace_data.action_tree_root)
+            .wrap_err("failed to verify rollup data inclusion proof")?;
+        Ok(())
```
noot, line 130: gonna turn this into a method of `RollupNamespaceData`

crates/astria-conductor/src/executor.rs
```diff
@@ -149,16 +149,19 @@ impl<C: ExecutionClient> Executor<C> {
...
-                        .execute_block(SequencerBlockSubset::from_sequencer_block_data(
-                            *block,
-                            self.namespace,
-                        ))
-                        .await
-                    {
+
+                    let Some(subset) =
+                        SequencerBlockSubset::from_sequencer_block_data(*block, self.namespace)
+                    else {
```
joroshiba, line 159: What is this else on?

crates/astria-conductor/src/block_verifier.rs
```diff
@@ -191,6 +256,14 @@ impl BlockVerifier {
                     bail!("last commit hash in header does not match calculated last commit hash");
                 }
 
+                // get validator set for the previous height, as the commit contained
+                // in the block is for the previous height
+                let validator_set = self
+                    .sequencer_client
+                    .validators(height, tendermint_rpc::Paging::Default)
```
steezeburger, line 263: Should this be `last_commit.height`? Or `height - 1`?

crates/astria-conductor/src/block_verifier.rs
```diff
@@ -45,23 +45,55 @@ use tracing::{
     warn,
 };
 
+#[async_trait::async_trait]
+pub(crate) trait SequencerClient {
+    async fn validators(
```
SuperFluffy, line 50: Is this functionality that should live in conductor only, or should we extend the `astria-sequencer-client`'s extension trait?

crates/astria-conductor/src/block_verifier.rs
```diff
@@ -414,12 +495,136 @@ fn get_proposer(validator_set: &ValidatorSet) -> eyre::Result<Validator> {
...
+        block::Commit,
+        validator,
+    };
 
     use super::*;
 
+    struct MockSequencerClient;
+
+    #[async_trait::async_trait]
+    impl SequencerClient for MockSequencerClient {
```
SuperFluffy, line 521: OK, I understand now why the sequencer client trait was introduced above. I am in general not a big fan of mocking at the generics level because you don't up testing the actual implementation but a shadow of the implementation.

We should implement proper black box tests on conductor instead.

SuperFluffy (changes requested):
We shouldn't mold the types and generics of the implementation around using mocked types that impl similar interfaces in testing, because this inevitably doesn't quite test the service but something else.

Because `astria-conductor` is undertested I am fine with not providing exhaustive blackbox tests for this feature. We should however fix that next.

Please remove the generic type params and reconsider the `TendermintHttpClient` newtype wrapper.

crates/astria-conductor/src/block_verifier.rs
```diff
@@ -128,38 +201,30 @@ impl BlockVerifier {
     /// - the signature is valid
     /// - the root of the markle tree of all the header fields matches the block's block_hash
```
steezeburger, line 202:
```suggestion
    /// - the root of the merkle tree of all the header fields matches the block's block_hash
```

crates/astria-conductor/src/executor.rs
noot: it's `let Some(..) else` syntax: https://doc.rust-lang.org/rust-by-example/flow_control/let_else.html

crates/astria-conductor/src/block_verifier.rs
noot: yeah I agree, however implementing entire black box tests in this PR seemed a bit overkill since I only wanted to test the `validate_..` functions implemented on `BlockVerifier`, and they only use one function on `SequncerClient`. alternatively I can separate the code that uses `SequencerClient` vs just validates so that can be unit tested without the generic

crates/astria-conductor/src/block_verifier.rs
noot: we can - either one is fine imo. currently, we only use this here though

crates/astria-conductor/src/block_verifier.rs
```diff
@@ -45,23 +45,55 @@ use tracing::{
     warn,
 };
 
+#[async_trait::async_trait]
+pub(crate) trait SequencerClient {
```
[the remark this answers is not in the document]
noot: i got rid of it!

crates/astria-conductor/src/block_verifier.rs
```diff
@@ -45,23 +45,55 @@ use tracing::{
...
 /// `BlockVerifier` is responsible for verifying the correctness of a block
 /// before executing it.
 /// It has two public functions: `validate_signed_namespace_data` and `validate_sequencer_block`.
 ///
 /// `validate_signed_namespace_data` is used to validate the data received from the data
 /// availability layer. `validate_sequencer_block` is used to validate the blocks received from
 /// either the data availability layer or the gossip network.
-pub struct BlockVerifier {
-    sequencer_client: HttpClient,
+pub(crate) struct BlockVerifier<C: SequencerClient> {
```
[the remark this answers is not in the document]
noot: removed

crates/astria-sequencer-relayer/src/data_availability.rs
```diff
@@ -168,6 +168,33 @@ pub struct RollupNamespaceData {
...
+
+    pub fn verify_inclusion_proof(&self, root_hash: [u8; 32]) -> eyre::Result<()> {
+        let rollup_data_tree = MerkleTree::from_leaves(self.rollup_txs.clone());
+        let rollup_data_root = rollup_data_tree.root();
+        let mut leaf = self.chain_id.clone();
+        leaf.append(&mut rollup_data_root.to_vec());
+
+        self.inclusion_proof
+            .verify(&leaf, root_hash)
+            .wrap_err("failed to verify rollup data inclusion proof")
```
SuperFluffy, line 194: Nit: you can just return the result of `self.inclusion_proof.verify(&leaf, root_hash)`, no need to wrap this in another error. When used you'd otherwise end up with something like this:
```rust
my_rollup_data.verify_inclusion_proof(some_hash)
    .wrap_err("failed to verify inclusion proof")?;
```
And then the error would be e.g.:
```
Error occured:
0: failed to verify inclusion proof

1. failed to verify rollup data inclusion proof

2. <actual errors>
```

crates/astria-conductor/src/types.rs
```diff
@@ -34,7 +34,10 @@ impl SequencerBlockSubset {
             ..
         } = data.into_raw();
 
-        let our_rollup_data = rollup_data.remove(&namespace)?;
+        let Some(our_rollup_data) = rollup_data.remove(&namespace) else {
```
SuperFluffy, line 37: Using `?` as it was before is idiomatic. No need to use the `let...else` construct.

crates/astria-conductor/src/block_verifier.rs
```diff
@@ -414,12 +478,136 @@ fn get_proposer(validator_set: &ValidatorSet) -> eyre::Result<Validator> {
...
+        block::Commit,
+        validator,
+    };
 
     use super::*;
 
+    fn make_test_validator_set(height: u32) -> ValidatorSet {
+        use ed25519_consensus::SigningKey;
+
+        let key_bytes = STANDARD.decode("XXhEZctO2gj3JhjHV7214N53zEl3s05eD9cWIJJHAVrPP2L5ARCQ0PSkNA85f3avKwpBsaDD6Q6VJlhN6ZHKjQ==").unwrap();
```
SuperFluffy, line 504: Are the magic bytes here important? Could we get away with just generating the key randomly?

crates/astria-conductor/src/block_verifier.rs
```diff
@@ -414,12 +478,136 @@ fn get_proposer(validator_set: &ValidatorSet) -> eyre::Result<Validator> {
...
+        let signing_key = SigningKey::try_from(key_bytes[0..32].to_vec().as_slice()).unwrap();
```
SuperFluffy, line 505:
```suggestion
        let signing_key = SigningKey::try_from(&key_bytes)
```

crates/astria-conductor/src/block_verifier.rs
```diff
@@ -414,12 +478,136 @@ fn get_proposer(validator_set: &ValidatorSet) -> eyre::Result<Validator> {
...
+        let pub_key = tendermint::public_key::PublicKey::Ed25519(
+            signing_key.verification_key().as_ref().try_into().unwrap(),
```
SuperFluffy, line 507:
```suggestion
        let pub_key = PublicKey::from_raw_ed25519(signing_key.verification_key().as_ref()).unwrap();
```

SuperFluffy (approved):
Just minor nits.

PR #291, opened 2023-08-21
fix(sequencer): fix `deliver_tx` to accept the action tree root

noot:
## Summary

my previous PR #209 actually broke sequencer+cometbft since `deliver_tx` is called on every tx, after `prepare_proposal`. since it was trying to deserialize every tx into `Signed`, it was failing on the action tree commitment (as it's just 32 bytes at the start of the block)

## Changes

- update the sequencer `App` to handle the first tx in a block properly, right now it just checks that it's 32 bytes. this is fine because the content of these 32 bytes is checked in `process_proposal`.

## Testing

ran it with cometbft and now works

## Related Issues

related to #153


SuperFluffy:
Can we create a very simple test that calls these methods in order?

noot:
@SuperFluffy do you mean a unit test that that calls the `Consensus` service functions the way cometbft would? I can add that

SuperFluffy:
@noot yeah. Ideally we'd have a cometbft mock that drives our app, but just calling these functions in order seems quicker.

crates/astria-sequencer-types/src/sequencer_block_data.rs
```diff
@@ -138,6 +138,7 @@ impl SequencerBlockData {
     /// - if the block has no data hash
     /// - if a transaction in the block cannot be parsed
     pub fn from_tendermint_block(b: Block) -> eyre::Result<Self> {
+        // TODO: this crate should not import astria_sequencer
```
joroshiba, line 141: nit: issue pls 🙏

crates/astria-sequencer/src/app.rs
```diff
@@ -99,11 +110,18 @@ impl App {
...
             .expect("components should not retain copies of shared state");
 
+        self.has_block_just_begun = true;
         self.apply(state_tx)
     }
 
     #[instrument(name = "App:deliver_tx", skip(self))]
     pub(crate) async fn deliver_tx(&mut self, tx: &[u8]) -> Result<Vec<abci::Event>> {
+        if self.has_block_just_begun {
+            ensure!(tx.len() == 32);
```
joroshiba, line 120: is there any other validation which can be done on this tx?

joroshiba (approved):
Agreed on the need for better testing, but don't see as blocking.

crates/astria-sequencer-types/src/sequencer_block_data.rs
noot: yes  #305 will update comment

crates/astria-sequencer/src/app.rs
noot: unfortunately not really, however the validators already verify this value in `ProcessProposal` by reconstructing the action tree so this is fine

crates/astria-sequencer/src/service/consensus.rs
```diff
@@ -408,4 +408,109 @@ mod test {
...
+                votes: vec![],
+            },
+            byzantine_validators: vec![],
+        };
+        consensus_service
+            .handle_request(ConsensusRequest::BeginBlock(begin_block))
+            .await
+            .unwrap();
+
+        for tx in txs {
```
SuperFluffy, line 494: this loop would have failed previously, right?

SuperFluffy (approved):
Looks good, thank you for the test!

crates/astria-sequencer/src/service/consensus.rs
noot: yes exactly

Issue #293, opened 2023-08-21
composer: refetch nonce on transaction failure due to invalid nonce


itamarreif:
We currently do not handle sequencer transactions failing due to incorrect nonce. This should be addressed so that rollup transactions don't get lost due to failed submission.

The channel communication to `Executor` that will be implemented in #292 will allow the `Executor` to submit transactions sequentially. This will allow the `Executor` reason about the nonce of one transaction at a time, rather than multiple pending submissions.

With these changes, nonces can be managed by having an `Option<Nonce>` in the `Executor` upon constructing an `Unsigned` from a `Vec<Action>` :
1. If nonce is `None`, fetch latest nonce from the sequencer
2. Return current nonce and increment for next time

When a `Signed` transaction submission fails due to an invalid nonce, reset the nonce to `None` and reprocess the `Vec<Action>`. This will result in the correct latest nonce being fetched from the sequencer, constructing a new `Unsigned`, signing it and submitting.

emhane:
nice issue! adding the several related issues into a task list in the root issue would make it easier to follow along.

to me it then seems like [this local value of the sequencer's nonce](https://github.com/astriaorg/astria/blob/a889ca6af6adbea28788cd4f438fe3f4f4be3be4/crates/astria-composer/src/searcher/mod.rs#L70-L71) could be removed in favour of [fetching the global value of the nonce](https://github.com/astriaorg/astria/blob/a889ca6af6adbea28788cd4f438fe3f4f4be3be4/crates/astria-composer/src/searcher/mod.rs#L276-L282)? that way, in the future, transitioning to a relationship between sequencer and composer that isn't 1-1 would be a piece of cake. for example if two node operators on two different rollups each run their own composer, but use the same sequencer. the local value otherwise always risks being out of sync.

emhane:
oh no, hang on, the [sequencer key](https://github.com/astriaorg/astria/blob/a889ca6af6adbea28788cd4f438fe3f4f4be3be4/crates/astria-composer/src/searcher/mod.rs#L68-L69) is for a sequencer account, right. then it's still efficient to use the local tracking nonce, so long as not many composers share a sequencer account, which I think we can assume they won't.

Issue #252, opened 2023-08-11
composer: use `get_nonce` as check to see if sequencer is ready instead of `abci_info`


noot:
currently, we use `abci_info` as a status check to see if the sequencer node is ready. however, we also call `get_nonce` directly after this to set the initial nonce. we can just use `get_nonce` as a status check instead and get rid of `abci_info`.

PR #294, opened 2023-08-22
fix(composer): refactor executor logic and initialize nonce when pinging sequencer

### LINKED ISSUE #292: composer: separate bundle construction from sequencer transaction building
Currently, rollup transactions are serialized and turned into a `Vec<Action>` in the [same function](https://github.com/astriaorg/astria/blob/main/crates/astria-composer/src/searcher/mod.rs#L209) that nonce is attached to create an `Unsigned`.

This creates a problem when transactions fail to be submitted due to incorrect nonce. While transaction failures due to incorrect nonce are not currently handled, separating these two functionalities will make the change much simpler in the future.

Instead of simply fetching a new nonce and attaching it to the `Vec<Action>` in [`handle_pending_tx`](https://github.com/astriaorg/astria/blob/main/crates/astria-composer/src/searcher/mod.rs#L209) do the following:

1. Move nonce and signing logic out from `handle_pending_txs`
    1a. Rename `handle_pending_txs` to `bundle_pending_txs` to clarify what it is doing.
    1b. Output a `Vec<Action>` to separate the interaction with the sequencer from bundle construction
2. Create an `Executor` struct that handles sequencer interaction
    2a. Run `wait_for_sequencer` in the `Executor`'s `run_until_stopped` function
    2b. For each `Vec<Action` do the following:
        - Retrieve nonce to create an `Unsigned`
        - Sign `Unsigned` to create a `Signed` tx
        - Submit `Signed` tx to sequencer.
3. Instead of running [`handle_signed_tx`](https://github.com/astriaorg/astria/blob/main/crates/astria-composer/src/searcher/mod.rs#L239) tasks in the `Searcher` object, pass the `Vec<Action>` to the `Executor` via a channel.

itamarreif:
## Summary

This is the first PR out of 3 to address nonce management in the Composer. This one specifically implements #292 and #252. A PR implementing #293 will follow and depend on this one.

## Background

Currently, sequencer transactions that fail due to invalid nonce are not handled, resulting in rollup transactions.

## Changes
(from #292 )

- [x] Move nonce and signing logic out from handle_pending_txs
  - [x] Rename handle_pending_txs to bundle_pending_txs to clarify what it is doing.
  - [x] Output a Vec<Action> to separate the interaction with the sequencer from bundle construction
- [x] Create an Executor struct that handles sequencer interaction
  - [x] Run wait_for_sequencer in the Executor's run_until_stopped function
  - [x] For each Vec<Action do the following:
    - Retrieve nonce to create an Unsigned
    - Sign Unsigned to create a Signed tx
    - Submit Signed tx to sequencer.
- [x] Instead of running [handle_signed_tx](https://github.com/astriaorg/astria/blob/main/crates/astria-composer/src/searcher/mod.rs#L239) tasks in the Searcher object, pass the Vec<Action> to the Executor via a channel.

(from #252)
- [x] Initialize `Executor`'s nonce in `Executor::wait_for_sequencer()`

(minor cleanup)
- [x] Change naming in `Collector` to reflect that the implementation is Geth-specific, rather than generic over EVM clients (due to dependency on websockets endpoint).

## Testing

- [x] Updated the existing tests to also verify the nonce is initialized properly.

## Related Issues

closes #292, #252


crates/astria-composer/src/searcher/collector.rs
```diff
@@ -213,12 +213,11 @@ impl Collector {
 /// TODO(https://github.com/astriaorg/astria/issues/216): add timeouts for
 /// `subscribe_full_pendings_txs` (more complex because it's a stream).
 #[derive(Clone, Debug)]
-struct EthClient {
+struct GethClient {
```
noot, line 216: why is this geth now? what geth-specific functionality are we using?

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -0,0 +1,321 @@
...
+
+use crate::Config;
+
+pub(super) type StatusReceiver = watch::Receiver<Status>;
+pub(super) type Sender = mpsc::Sender<Vec<Action>>;
+
+pub(super) fn spawn(cfg: &Config) -> eyre::Result<(Sender, StatusReceiver)> {
+    info!("Spawning Executor subtask for Searcher");
+    // create channel for sending bundles to executor
+    let (executor_tx, executor_rx) = mpsc::channel(256);
```
noot, line 55: why 256?

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -0,0 +1,321 @@
...
+            .map_err(|_| eyre!("invalid private key length; must be 32 bytes"))?;
+        let sequencer_key =
+            SigningKey::try_from(private_key_bytes).wrap_err("failed to parse sequencer key")?;
+        private_key_bytes.zeroize();
+
+        // create address from signing key
+        let sequencer_address = Address::from_verification_key(sequencer_key.verification_key());
+
+        // create channel for status reporting
+        let (status, _) = watch::channel(Status::new());
```
noot, line 152: where's the receiver? the receiver has to be used otherwise sending on this channel will error due to all receivers being dropped

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -0,0 +1,321 @@
...
+        self.status.subscribe()
+    }
+
+    /// Return the current nonce
+    fn nonce(&self) -> u32 {
+        self.nonce.load(Ordering::Relaxed)
+    }
+
+    /// Gets the next nonce to sign over
+    fn get_next_nonce(&mut self) -> u32 {
```
noot, line 175: name this something like `get_and_increment_nonce` - bad to have a function called `get` that actually mutates stuff

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -0,0 +1,321 @@
...
+        // get current nonce and calculate next one
+        let curr_nonce = self.nonce();
+        let next_nonce = curr_nonce + 1;
+        // save next nonce
+        self.nonce.store(next_nonce, Ordering::Relaxed);
+        curr_nonce
```
noot, line 181: replace this with `fetch_add`: https://doc.rust-lang.org/stable/std/sync/atomic/struct.AtomicU32.html#method.fetch_add

```suggestion
self.nonce.fetch_add(1, Ordering::Relaxed)
```

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -0,0 +1,321 @@
...
+    }
+
+    /// Creates an `Unsigned` from `Vec<Action>` using the current nonce.
+    /// If the current nonce is not stored, fetches the latest nonce from the sequencer node.
```
noot, line 185: is the nonce ever not stored?

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -0,0 +1,321 @@
...
+    fn make_unsigned_tx(&mut self, actions: Vec<Action>) -> UnsignedTransaction {
+        // get current nonce and increment nonce
+        let curr_nonce = self.get_next_nonce();
+        UnsignedTransaction {
+            nonce: curr_nonce,
+            actions,
+        }
+    }
+
+    /// TODO
```
noot, line 195: is this done?

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -0,0 +1,321 @@
...
+
+        // for each bundle received, create unsigned tx, sign tx, then submit tx
+        // Vec<Action> -> Unsigned -> Signed -> Submit
+        while let Some(bundle) = self.executor_rx.recv().await {
+            // create unsigned tx
+            let unsigned_tx = self.make_unsigned_tx(bundle);
+            // sign tx
+            let signed_tx = unsigned_tx.into_signed(&self.sequencer_key);
+            // submit tx
+            self.submit_tx(signed_tx).await?;
```
noot, line 231: this will shut down the loop if there's an error, is that what we want?

crates/astria-composer/src/searcher/collector.rs
itamarreif: it depends on the websockets endpoint, which is geth specific. nothing changed, just renaming for clarity

crates/astria-composer/src/searcher/executor.rs
itamarreif: my bad, forgot to update comment

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -0,0 +1,333 @@
...
+    error,
+    info,
+    instrument,
+    warn,
+};
+
+use crate::Config;
+
+pub(super) type StatusReceiver = watch::Receiver<Status>;
+pub(super) type Sender = mpsc::Sender<Vec<Action>>;
```
dereksione, line 41: Would prefer a more specific name for this - `Sender` is too generic - would prefer a `Vec<Action>` be renamed to something like  a `bundle` (just for ex.) and this to be `BundleSender`

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -0,0 +1,333 @@
...
+                error!("executor task exited unexpectedly");
+            }
+            Ok(Err(e)) => {
+                error!(
+                    error.message = %e,
+                    error.cause_chain = ?e,
+                    "executor task failed unexpectedly with error",
+                );
+            }
+            Err(e) => {
```
dereksione, line 68: I feel like this case and the case above can be combined. If there is meaningful distinction between the two, they don't seemed to be clear from how the branches are handled

crates/astria-composer/src/searcher/mod.rs
```diff
@@ -376,48 +270,27 @@ impl Searcher {
...
-            let client = self.sequencer_client.clone();
-            async move { client.abci_info().await }
-        })
-        .retry(&backoff)
-        .notify(|err, dur| warn!(error.msg = %err, retry_in = %format_duration(dur), "failed getting abci info; retrying"))
+    async fn wait_for_executor(&self) -> eyre::Result<()> {
+        // wait to receive executor status
+        let mut status = self.executor_status.clone();
+        async move {
+            match status.wait_for(executor::Status::is_connected).await {
```
dereksione, line 277: I don't think we need a match statement for this - just upwards propagate the error and then return and `Ok(())`

crates/astria-composer/src/searcher/executor.rs
itamarreif: They are for handling different errors. This one is for the executor task panicking and the other is for it returning an error - since they're different code paths I feel like they should be reported differently.

crates/astria-composer/src/searcher/executor.rs
itamarreif: this follows the naming scheme used in the conductor: https://github.com/astriaorg/astria/blob/937f710d23cdde80a2768309f12105d46e55129b/crates/astria-conductor/src/executor.rs#L45

externally to this module this appears as `executor::Sender`

SuperFluffy (commented):
I realize I never submitted by review

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -0,0 +1,322 @@
...
+    debug,
+    error,
+    info,
+    instrument,
+    warn,
+};
+
+use crate::Config;
+
+pub(super) type StatusReceiver = watch::Receiver<Status>;
```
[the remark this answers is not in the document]
itamarreif: just matching the conductor: https://github.com/astriaorg/astria/blob/937f710d23cdde80a2768309f12105d46e55129b/crates/astria-conductor/src/executor.rs#L45

crates/astria-composer/src/searcher/executor.rs
SuperFluffy: Type aliases should be uses sparsely and only when they improve code clarity. A good example is `type Result<T> = Result<T, CrateSpecificError>`. For things like `Sender<ExecutorCommand>` they don't really improve clarify because a) this type is only used in 2 or so places, and b) I have to look up the alias to understand what's inside. The overhead isn't worth it. We should change conductor accordingly.

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -0,0 +1,322 @@
...
+    status: watch::Sender<Status>,
+    // Channel for receving bundles to pack, sign, and submit
+    executor_rx: mpsc::Receiver<Vec<Action>>,
+    // The client for submitting wrapped and signed pending eth transactions to the astria
+    // sequencer.
+    sequencer_client: SequencerClient,
+    // Private key used to sign sequencer transactions
+    sequencer_key: SigningKey,
+    // Nonce of the sequencer account we sign with
+    nonce: Arc<AtomicU32>,
```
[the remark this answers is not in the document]
itamarreif: my bad, this was a relic of how multiple txs were submitted in parallel previously. fixed.

crates/astria-composer/src/searcher/mod.rs
```diff
@@ -226,6 +245,39 @@ impl Searcher {
         }
     }
 
+    fn spawn_executor(&mut self) -> tokio::task::JoinHandle<()> {
+        // spawn executor task
+        let handle = tokio::spawn(
```
dereksione, line 250: Nit: would prefer to be consistent between `tokio::spawn` or `tokio::task::spawn` and not use both

dereksione (approved):

crates/astria-composer/src/searcher/mod.rs
itamarreif: good catch - thanks

crates/astria-composer/src/searcher/collector.rs
SuperFluffy, line 1: Generally fine with the naming changes, but they technically have nothing to do with the described objective of the PR. Please try to keep orthogonal changes in their own PRs.

crates/astria-composer/src/searcher/mod.rs
```diff
@@ -45,30 +20,29 @@ use tokio::{
...
 use tracing::{
-    debug,
-    instrument,
+    error,
     warn,
 };
 
-use crate::Config;
+use crate::{
+    searcher::executor::Executor,
```
SuperFluffy, line 28: executor is a submodule of searcher. Import it with `use executor::Executor`

crates/astria-composer/src/searcher/mod.rs
```diff
@@ -45,30 +20,29 @@ use tokio::{
...
+    Config,
+};
 
 mod collector;
+mod executor;
 mod rollup;
 
 use collector::Collector;
 
+const BUNDLE_CHANNEL_SIZE: usize = 256;
```
SuperFluffy, line 38: Why define a const here that's used in one place?

crates/astria-composer/src/searcher/mod.rs
```diff
@@ -181,31 +133,26 @@ impl Searcher {
...
-        // create signing key for sequencer txs
-        let mut private_key_bytes: [u8; 32] = hex::decode(cfg.private_key.expose_secret())
-            .wrap_err("failed to decode private key bytes from hex string")?
-            .try_into()
-            .map_err(|_| eyre!("invalid private key length; must be 32 bytes"))?;
-        let sequencer_key =
-            SigningKey::try_from(private_key_bytes).wrap_err("failed to parse sequencer key")?;
-        private_key_bytes.zeroize();
+        // create channel for sending bundles to executor
+        let (executor_tx, executor_rx) = mpsc::channel(BUNDLE_CHANNEL_SIZE);
```
SuperFluffy, line 139: IMO remove the const here, just put the number here.

crates/astria-composer/src/searcher/mod.rs
```diff
@@ -216,83 +163,47 @@ impl Searcher {
...
         } = tx;
 
-        let sequencer_key = self.sequencer_key.clone();
-        let nonce = self.sequencer_nonce.clone();
-
+        // rollup transaction data serialization is a heavy compute task, so it is spawned
+        // on tokio's blocking threadpool
         self.conversion_tasks.spawn_blocking(move || {
-            // Pack into sequencer tx
+            // Pack into a vector of sequencer actions
```
SuperFluffy, line 175: IMO remove comment. Kinda self evident

crates/astria-composer/src/searcher/mod.rs
```diff
@@ -316,6 +227,8 @@ impl Searcher {
                         ),
                     }
                 }
+
+                _ = &mut exectuor_handle => { todo!("handle executor failure"); }
```
SuperFluffy, line 231: Please fix this

crates/astria-composer/src/searcher/mod.rs
```diff
@@ -216,83 +163,47 @@ impl Searcher {
...
                 // serialize and sign sequencer tx for incoming pending rollup txs
-                Some(rollup_tx) = self.new_transactions.recv() => self.handle_pending_tx(rollup_tx),
+                Some(rollup_tx) = self.new_transactions.recv() => self.bundle_pending_tx(rollup_tx),
 
                 // submit signed sequencer txs to sequencer
                 Some(join_result) = self.conversion_tasks.join_next(), if !self.conversion_tasks.is_empty() => {
                     match join_result {
-                        Ok(Ok(signed_tx)) => self.handle_signed_tx(signed_tx),
-                        Ok(Err(e)) => warn!(error.message = %e, error.cause_chain = ?e, "failed to sign sequencer transaction"),
+                        Ok(actions) => self.executor_tx.send(actions).await?,
```
SuperFluffy, line 206: This is a) breaking the loop, but also b) not adding error context. Is that intentional

crates/astria-composer/src/searcher/mod.rs
```diff
@@ -332,6 +245,39 @@ impl Searcher {
         }
     }
 
+    fn spawn_executor(&mut self) -> tokio::task::JoinHandle<()> {
+        // spawn executor task
+        let handle = tokio::task::spawn(
```
SuperFluffy, line 250: This should be done directly inside `searcher::run`

crates/astria-composer/src/searcher/mod.rs
```diff
@@ -332,6 +245,39 @@ impl Searcher {
         }
     }
 
+    fn spawn_executor(&mut self) -> tokio::task::JoinHandle<()> {
```
SuperFluffy, line 248: Remove

crates/astria-composer/src/searcher/mod.rs
```diff
@@ -332,6 +245,39 @@ impl Searcher {
...
+        // spawn executor task
+        let handle = tokio::task::spawn(
+            self.executor
+                .take()
+                .expect("executor should only be run once")
+                .run_until_stopped(),
+        );
+
+        // return handle to task that logs why executor exits
+        tokio::task::spawn(async move {
```
SuperFluffy, line 258: Don't spawn a second task here

SuperFluffy (changes requested):
Please address the `todo!()` macro: the `spawn` method can be removed - just spawn directly inside the searcher and await the handle. If the handle returns early, shut down searcher

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -0,0 +1,296 @@
...
+                // FIXME: currently this will fail both when there is an issue with the nonce and
+                // when unable to reach the sequencer. As there is currently no error returned by
+                // the sequencer for invalid nonces, there is nothing to handle. This should be
+                // changed after #364 is merged in a followup PR to handle nonce failues.
+                error!(
+                    error.message = %e,
+                    error.cause_chain = ?e,
+                    "failed submitting the bundle to the sequencer; bailing",
+                );
+                break;
```
SuperFluffy, line 190: Note that I added this. Upon further inspection some of the errors are critical,  preventing the service from going on.

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -0,0 +1,299 @@
...
+            .wrap_err("failed retrieving initial nonce from sequencer")?;
+
+        while let Some(bundle) = self.executor_rx.recv().await {
+            if let Err(e) = self.process_bundle(bundle.clone()).await {
+                // FIXME: currently this will fail both when there is an issue with the nonce and
+                // when unable to reach the sequencer. As there is currently no error returned by
+                // the sequencer for invalid nonces, there is nothing to handle. This should be
+                // changed after #364 is merged in a followup PR to handle nonce failues.
+                error!(
+                    error.message = %e,
```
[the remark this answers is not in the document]
SuperFluffy: Added a break statement below

SuperFluffy (approved):
Pushed the changes I requested after syncing with @itamarreif

crates/astria-composer/src/searcher/mod.rs
itamarreif: https://github.com/astriaorg/astria/pull/294#discussion_r1323202939

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -0,0 +1,296 @@
...
+        })
+        .retry(&backoff)
+        .notify(|err, dur|
+            warn!(error.msg = %err, retry_in = %format_duration(dur), "failed getting nonce for {:?}; retrying", self.address))
+        .await
+        .wrap_err(
+            "failed to retrieve initial nonce from sequencer after several retries",
+        )?;
+
+        // update stored nonce
```
SuperFluffy, line 240: can we remove this obvious comments? :)

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -0,0 +1,296 @@
...
+        self.nonce = Some(nonce_response.nonce);
+        info!(
+            nonce_response.nonce,
+            "retrieved initial nonce from sequencer successfully"
+        );
+
+        // update status to connected
```
SuperFluffy, line 247: same

crates/astria-composer/src/searcher/mod.rs
SuperFluffy: This was extracted from waiting for the collectors. I think (but am not sure) that it was necessary to do it there explicitly to avoid any sort of implicit `+ 'a` bound being added to the returned future.

I agree that in this case the entire `async move` is probably not necessary

crates/astria-composer/src/searcher/mod.rs
```diff
@@ -181,32 +127,26 @@ impl Searcher {
...
 
-        // create signing key for sequencer txs
-        let mut private_key_bytes: [u8; 32] = hex::decode(cfg.private_key.expose_secret())
-            .wrap_err("failed to decode private key bytes from hex string")?
-            .try_into()
-            .map_err(|_| eyre!("invalid private key length; must be 32 bytes"))?;
-        let sequencer_key =
-            SigningKey::try_from(private_key_bytes).wrap_err("failed to parse sequencer key")?;
-        private_key_bytes.zeroize();
+        // create channel for sending bundles to executor
```
SuperFluffy, line 132: obvious comments are obvious :)

crates/astria-composer/src/searcher/mod.rs
```diff
@@ -181,32 +127,26 @@ impl Searcher {
...
+        let (executor_tx, executor_rx) = mpsc::channel(256);
+        let executor = Executor::new(&cfg.sequencer_url, &cfg.private_key, executor_rx)
+            .wrap_err("executor construction from config failed")?;
+
+        // create channel for receiving executor status
```
SuperFluffy, line 137: please rm

crates/astria-composer/src/searcher/mod.rs
```diff
@@ -216,111 +156,96 @@ impl Searcher {
...
-            let curr_nonce = nonce.fetch_add(1, Ordering::Relaxed);
-            let unsigned_tx = UnsignedTransaction {
-                nonce: curr_nonce,
-                actions: vec![Action::Sequence(SequenceAction {
-                    chain_id: chain_id.into_bytes(),
-                    data,
-                })],
-            };
-            Ok(unsigned_tx.into_signed(&sequencer_key))
+            vec![seq_action]
```
SuperFluffy, line 175: creating a single-item vec seems extreme?

crates/astria-composer/src/searcher/mod.rs
```diff
@@ -216,111 +156,96 @@ impl Searcher {
...
             chain_id,
             inner: rollup_tx,
         } = tx;
 
-        let sequencer_key = self.sequencer_key.clone();
-        let nonce = self.sequencer_nonce.clone();
-
+        // rollup transaction data serialization is a heavy compute task, so it is spawned
+        // on tokio's blocking threadpool
         self.conversion_tasks.spawn_blocking(move || {
```
SuperFluffy, line 167: I believe this is no longer necessary

crates/astria-composer/src/searcher/mod.rs
```diff
@@ -216,111 +156,96 @@ impl Searcher {
...
                 // serialize and sign sequencer tx for incoming pending rollup txs
-                Some(rollup_tx) = self.new_transactions.recv() => self.handle_pending_tx(rollup_tx),
+                Some(rollup_tx) = self.new_transactions.recv() => self.bundle_pending_tx(rollup_tx),
 
                 // submit signed sequencer txs to sequencer
                 Some(join_result) = self.conversion_tasks.join_next(), if !self.conversion_tasks.is_empty() => {
                     match join_result {
-                        Ok(Ok(signed_tx)) => self.handle_signed_tx(signed_tx),
-                        Ok(Err(e)) => warn!(error.message = %e, error.cause_chain = ?e, "failed to sign sequencer transaction"),
+                        Ok(bundle) => self.handle_bundle_execution(bundle).await,
```
SuperFluffy, line 215: This now effectively blocks the entire loop

crates/astria-composer/src/searcher/mod.rs
itamarreif: the payload of a sequencer transaction is a vector of actions. can make the vector here or in the executor, but doing so here is more correct since the executor should just receive a payload rather than construct it.

crates/astria-composer/src/searcher/mod.rs
itamarreif: i disagree - since collectors are ingesting transactions concurrrently there is no reason serialization shouldnt happen concurrently. this will eventually need to be fixed, see issue #409 for an in-depth explanation.

crates/astria-composer/src/searcher/mod.rs
itamarreif: see #409

PR #295, opened 2023-08-22
feat(composer): adding more docs

dereksione:
## Summary

Adding more doc comments to underdocumented parts of the composer

## Background

This is a result of the documentation workshop in the Colorado onsite

## Changes

- Adding more doc comments

### DIFF
--- a/crates/astria-composer/src/composer.rs
+++ b/crates/astria-composer/src/composer.rs
@@ -20,14 +20,16 @@ use crate::{
     Config,
 };
 
-/// Composer is the component in the Astria ecosystem that is responsible for
-/// submitting transactions to the Shared Sequencer. It usually does this by
-/// running a searcher module that establishes connections to individual
-/// rollup nodes and then bundling one or more of these rollup transactions into
-/// a sequencer transaction before submitting these to the sequencer.
+/// Composer is a service in the Astria ecosystem responsible for submitting
+/// transactions to the Astria Shared Sequencer.
 pub struct Composer {
+    /// ApiServer is used for monitoring status of the Composer service.
     api_server: ApiServer,
+    /// Searcher establishes connections to individual rollup nodes, receiving
+    /// pending transactions from them and wraps them as sequencer transactions
+    /// for submission.
     searcher: Searcher,
+}
 }
 
 impl Composer {
--- a/crates/astria-composer/src/composer.rs
+++ b/crates/astria-composer/src/composer.rs
@@ -20,6 +20,11 @@ use crate::{
     Config,
 };
 
+/// Composer is the component in the Astria ecosystem that is responsible for
+/// submitting transactions to the Shared Sequencer. It usually does this by
+/// running a searcher module that establishes connections to individual
+/// rollup nodes and then bundling one or more of these rollup transactions into
+/// a sequencer transaction before submitting these to the sequencer.
 pub struct Composer {
     api_server: ApiServer,
     searcher: Searcher,
@@ -54,6 +59,7 @@ impl Composer {
         })
     }
 
+    /// Returns the socket address the the api server is served over
     pub fn local_addr(&self) -> SocketAddr {
         self.api_server.local_addr()
     }
--- a/crates/astria-composer/src/composer.rs
+++ b/crates/astria-composer/src/composer.rs
@@ -30,7 +30,6 @@ pub struct Composer {
     /// for submission.
     searcher: Searcher,
 }
-}
 
 impl Composer {
     /// Constructs a new Searcher service from config.
--- a/crates/astria-composer/src/composer.rs
+++ b/crates/astria-composer/src/composer.rs
@@ -61,7 +61,7 @@ impl Composer {
         })
     }
 
-    /// Returns the socket address the the api server is served over
+    /// Returns the socket address the api server is served over
     pub fn local_addr(&self) -> SocketAddr {
         self.api_server.local_addr()
     }
--- a/crates/astria-composer/src/composer.rs
+++ b/crates/astria-composer/src/composer.rs
@@ -20,8 +20,8 @@ use crate::{
     Config,
 };
 
-/// Composer is a service in the Astria ecosystem responsible for submitting
-/// transactions to the Astria Shared Sequencer.
+/// Composer is a service responsible for submitting transactions to the Astria
+/// Shared Sequencer.
 pub struct Composer {
     /// ApiServer is used for monitoring status of the Composer service.
     api_server: ApiServer,
--- a/crates/astria-composer/src/lib.rs
+++ b/crates/astria-composer/src/lib.rs
@@ -1,7 +1,7 @@
 //! Astria's composer submits EVM rollup transactions to astria's sequencer.
 //!
 //! At the moment composer can read from EVM rollups only, and submits each EVM
-//! transaction (regardless of where it's collected from) as one sequencer transaction. 
+//! transaction (regardless of where it's collected from) as one sequencer transaction.
 //! The submission nonces are based on a custom generated private key specific to the composer.
 //! Each sequencer transaction is submitted with a new nonce which is one more than the nonce
 //! used in the last submission.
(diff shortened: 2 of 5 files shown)


crates/astria-composer/src/composer.rs
```diff
@@ -20,6 +20,11 @@ use crate::{
...
 };
 
+/// Composer is the component in the Astria ecosystem that is responsible for
+/// submitting transactions to the Shared Sequencer. It usually does this by
+/// running a searcher module that establishes connections to individual
+/// rollup nodes and then bundling one or more of these rollup transactions into
+/// a sequencer transaction before submitting these to the sequencer.
 pub struct Composer {
     api_server: ApiServer,
     searcher: Searcher,
```
joroshiba, line 30: Should we have a high level comment on the service, and then comment on the components which describe what happens today vs what will eventually happen? ie:

```suggestion
/// Composer is a service in the Astria ecosystem responsible for submitting
/// transactions to the Astria Shared Sequencer.
pub struct Composer {
    /// ApiServer is used for monitoring status of the Composer service.
    api_server: ApiServer,
    /// Searcher establishes connections to individual rollup nodes, receiving
    /// pending transactions from them and wraps them as sequencer transactions
    /// for submission.
    searcher: Searcher,
}
```

crates/astria-composer/src/composer.rs
```diff
@@ -54,6 +59,7 @@ impl Composer {
         })
     }
 
+    /// Returns the socket address the the api server is served over
```
joroshiba, line 62:
```suggestion
    /// Returns the socket address the api server is served over
```

crates/astria-composer/src/searcher/collector.rs
```diff
@@ -32,6 +32,10 @@ pub(super) struct Transaction {
     pub(super) inner: ethers::types::Transaction,
 }
 
+/// A Collector is a sub-actor in the Searcher module that interfaces with individual rollups.
+/// It is responsible for fetching pending transactions submitted to the rollup nodes and then
+/// passing them downstream for the searcher to process. Thus, a searcher can have multiple
+/// collectors running at the same time funneling data from multiple rollup nodes.
```
joroshiba, line 38: Cleaned up language a bit here, but main things:
- collector doesn't fetch right now, future iterations could but collect is the most accurate.
- each collector operates on a specific rollup node. 

```suggestion
/// Collector is a sub-actor in the Searcher module that interfaces with
/// individual rollups. It collects transactions submitted to a rollup node and
/// passes them downstream for the searcher to process.
/// 
/// A searcher may have multiple collectors running at the same time funneling
/// data from multiple rollup nodes.
```

crates/astria-composer/src/searcher/mod.rs
```diff
@@ -52,10 +52,13 @@ mod rollup;
 
 use collector::Collector;
 
-/// the astria seqeuencer.
+/// A Searcher collates transactions from multiple rollups and bundles them into
+/// Astria sequencer transactions using arbitrary strategies and then passed on to the 
+/// shared sequencer. These sequencer transactions have the highly desirable quality of
+/// atomic inclusion, i.e. if they are submitted to the sequencer, all of them are going
+/// to be executed in the same Astria block.
```
joroshiba, line 59: Want to avoid marketing language here ie `highly desirable quality`.

We also don't currently have arbitrary strategies.

joroshiba (changes requested):
Mainly verbiage stuff here, but on the whole, more than anything else. I want to make sure our doc strings are accurate to how things work. The comments don't match the behavior is a bug IMO.

Put a couple suggestions here, but I'm not strongly opinionated on my language. Just want to keep succinct and  such want to keep more technical than marketing speak + technically accurate.

crates/astria-composer/src/composer.rs
```diff
@@ -20,10 +20,17 @@ use crate::{
...
+/// transactions to the Astria Shared Sequencer.
 pub struct Composer {
+    /// ApiServer is used for monitoring status of the Composer service.
     api_server: ApiServer,
+    /// Searcher establishes connections to individual rollup nodes, receiving
+    /// pending transactions from them and wraps them as sequencer transactions
+    /// for submission.
     searcher: Searcher,
 }
+}
```
joroshiba, line 33: extra curly in here.

crates/astria-composer/src/lib.rs
```diff
@@ -1,8 +1,10 @@
 //! Astria's composer submits EVM rollup transactions to astria's sequencer.
 //!
-//! At the moment composer can read from one EVM rollup only, and submits each EVM
-//! transaction as one sequencer transaction. It also does not support using a specific
-//! account/signing key and instead generates a random account for each submission.
+//! At the moment composer can read from EVM rollups only, and submits each EVM
+//! transaction (regardless of where it's collected from) as one sequencer transaction.
+//! The submission nonces are based on a custom generated private key specific to the composer.
+//! Each sequencer transaction is submitted with a new nonce which is one more than the nonce
+//! used in the last submission.
```
SuperFluffy, line 7: maybe previous submission?

crates/astria-composer/src/composer.rs
```diff
@@ -20,8 +20,14 @@ use crate::{
     Config,
 };
 
+/// Composer is a service in the Astria ecosystem responsible for submitting
```
SuperFluffy, line 23: I think we don't need to repeat `Astria` just quite that often. :-)

```suggestion
/// Composer submits transactions to the astria shared sequencer.
```

crates/astria-composer/src/composer.rs
```diff
@@ -20,8 +20,14 @@ use crate::{
...
+/// transactions to the Astria Shared Sequencer.
 pub struct Composer {
+    /// ApiServer is used for monitoring status of the Composer service.
```
SuperFluffy, line 26: Nit: I have a preference for using the field name, not the type.

crates/astria-composer/src/searcher/collector.rs
```diff
@@ -32,6 +32,12 @@ pub(super) struct Transaction {
     pub(super) inner: ethers::types::Transaction,
 }
 
+/// Collector is a sub-actor in the Searcher module that interfaces with
```
SuperFluffy, line 35: Please leave an empty doc comment that summarizes this type, e.g.:
```suggestion
/// Collects transactions submitted to a rollup node and passes them downstream for further processing.
///
/// Collector is a sub-actor in the Searcher module that interfaces with
```

crates/astria-composer/src/composer.rs
dereksione:
```
/// Composer is a service responsible for submitting transactions to the Astria
/// Shared Sequencer.
```

Is this ok? I want to make clear that the Composer is a separate service that runs on its own. Maybe redundant though, open to thoughts

crates/astria-composer/src/searcher/collector.rs
dereksione: Yeah, that makes sense

crates/astria-composer/src/searcher/collector.rs
```diff
@@ -206,6 +217,7 @@ struct EthClient {
 }
 
 impl EthClient {
```
itamarreif, line 219: I would note that this is a Geth client rather than a generic Ethereum client

crates/astria-composer/src/lib.rs
```diff
@@ -1,8 +1,10 @@
 //! Astria's composer submits EVM rollup transactions to astria's sequencer.
 //!
-//! At the moment composer can read from one EVM rollup only, and submits each EVM
-//! transaction as one sequencer transaction. It also does not support using a specific
-//! account/signing key and instead generates a random account for each submission.
+//! At the moment composer can read from EVM rollups only, and submits each EVM
```
itamarreif, line 3: from Geth nodes only

itamarreif (approved):
looks good, would just clarify about Geth

joroshiba (approved):

Issue #310, opened 2023-08-25
Verifying signed transactions is not reliable because protobuf serialization is not deterministic


SuperFluffy:
The protobuf `SignedTransaction` message needs to be changed to contain the raw byte stream encoding an unsigned transaction so that the signature and verification key can be applied to the raw bytes:
```protobuf
// Proposal
message SignedTransactionNew {
  bytes signature = 1;
  bytes verification_key = 2;
  bytes transaction = 3;
}
// Current
message SignedTransactionOld {
  bytes signature = 1;
  byte verification_key = 2;
  UnsignedTransaction transaction = 3;
}
```
The problem with the current scheme is that when a signed transaction is received over the wire and decoded into a rust native `SignedTransaction`, the raw bytes that made up the inner `UnsignedTransaction` are lost and there is no way to get them back. This is because protobuf serialization is not deterministic, and can change between languages and even version of the parser library. See https://gist.github.com/kchristidis/39c8b310fd9da43d515c4394c3cd9510.

Currently, in order to guarantee that the signed transaction indeed matches the signature and verification key, we have a flow like this:
```
signed tx bytes
-> unverified SignedTransactionOld
-> encode inner unsigned transaction to unsigned tx bytes (!!)
-> apply verification key and signature to unsigned tx bytes (??)
-> verified SignedTransaction
```
The problem is the encoding step marked with `(!!)`: there is no guarantee that the bytes of that step match the bytes of the original transaction that was send over the wire, and hence that this passes the verification check marked with `(??)`. Currently we don't encounter issues because all signed tx construction happens in services that are all locked to the same version of `prost`, and so they do produce the same byte-output.

With the proposed `SignedTransactionNew` the flow would be changed:
```
signed tx bytes
-> unverified SignedTransactionNew
-> apply verification key and signature to inner tx bytes
-> decode inner tx bytes to `UnsignedTransaction`
-> verified SignedTransaction
```

noot:
proposed solution:

update `SignedTransaction` to contain a `google.protobuf.Any` instead of an `UnsignedTransaction`:

```
message SignedTransaction {
  bytes signature = 1;
  bytes public_key = 2;
  google.protobuf.Any transaction = 3;
}
```

then, the signature can be verified over the bytes in the `Any` instead of an `UnsignedTransaction` which needs to be reserialized.

SuperFluffy:
Agree, that's the right way of moving forward.

Issue #228, opened 2023-08-07
sequencer: implement `check_tx` error codes


noot:
`check_tx` currently returns 0 for ok and 1 for error. tendermint defines any nonzero value as error. we can return different error codes depending on the error. these don't strictly need to match cosmos but might be nice if they do.

ShivangRawat30:
Hey I am currently learning rust, as it is a good first issue can you please assign this to me.

noot:
done by #311

PR #311, opened 2023-08-28
sequencer: move all public interfaces to `astria-proto` and lock it down

SuperFluffy:
## Summary

This patch makes all of `astria-sequencer` private except for its config and service entry point. All its public interfaces are moved to `astria-proto` and all other crates are implemented in terms of it.

## Background

`astria-sequencer` was a direct (or indirect, via `astria-sequencer-client`) dependency of several services, leading to a tightly coupled dependency graph and long compile times (primarily due to sequencer pulling in rocksdb as a dependency). It is best practice to not make binary crates dependencies of other crates. Furthermore, `astria-proto` serves as the single point of truth for public APIs (which should include communication between astria services).

## Changes

- Remove all transaction (signed, unsigned) and action (sequence, transfer) types from sequencer.
- Make all of `astria-sequencer` private except for its config and `Sequencer` type entrypoint.
- Define `SignedTransaction`, `UnsignedTransaction`, `SequenceAction`, `TransferAction` in `astria_proto::native::sequencer:v1alpha1`.
- Define transformations from the raw generated protobuf types to the above "native" astria types.
- Reimplement all services in terms of these native types.
- Refactor functions and methods as appropriate.

## Testing

The changes were not explicitly tested, but all tests were updated and ensured to still pass.

## Breaking Changelist

- All services now communicate via the protobuf encoded types.
- It should be assumed that no services older than this patch can communicate with services from this patch onward.

## Related Issues

Followup to https://github.com/astriaorg/astria/pull/247

closes
- https://github.com/astriaorg/astria/issues/86
- https://github.com/astriaorg/astria/issues/191


crates/astria-sequencer/src/transaction/mod.rs
SuperFluffy, line 1: Note on this module: there are no logic changes here, just definitions moved out of `unsigned.rs` and the trait reimplemented for the external `UnsignedTransaction` type

crates/astria-sequencer/src/service/mod.rs
```diff
@@ -7,3 +7,39 @@ pub(crate) use consensus::Consensus;
 pub(crate) use info::Info;
 pub(crate) use mempool::Mempool;
 pub(crate) use snapshot::Snapshot;
+use tendermint::abci::Code;
+
+#[derive(Clone, Copy, Debug)]
+pub(crate) struct AbciCode(u32);
+
+#[rustfmt::skip]
+impl AbciCode {
```
SuperFluffy, line 16: Moved this to its own block so that rustfmt does not reorder the associated consts.

crates/astria-sequencer/src/accounts/state_ext.rs
```diff
@@ -2,26 +2,29 @@ use anyhow::{
...
     instrument,
 };
 
-use crate::accounts::types::{
-    Balance,
-    Nonce,
-};
+/// Newtype wrapper to read and write a u32 from rocksdb.
+#[derive(BorshSerialize, BorshDeserialize, Debug)]
+struct Nonce(u32);
```
SuperFluffy, line 23: `Nonce` and `Balance` are now private to this module and just wrappers to ensure that they are correctly read from and written to rocksdb.

crates/astria-sequencer-validation/src/utils.rs
```diff
@@ -2,17 +2,23 @@ use std::collections::BTreeMap;
...
+}
+
 /// Groups the `sequence::Action`s within the transactions by their `chain_id`.
 /// The `BTreeMap` key is the chain ID and value is the transactions.
 #[must_use]
 pub fn generate_action_tree_leaves(
-    chain_id_to_txs: BTreeMap<Vec<u8>, Vec<Vec<u8>>>,
-) -> Vec<Vec<u8>> {
-    let mut leaves: Vec<Vec<u8>> = vec![];
+    chain_id_to_txs: BTreeMap<[u8; 32], Vec<Vec<u8>>>,
```
SuperFluffy, line 16: This makes use of the fact that chain ID is now a fixed size array. Note that `astria-proto` is not made a dependency of `astria-sequencer-validation` so the 32 bytes are not called `CHAIN_ID_LENGTH`.

crates/astria-sequencer-validation/src/proof.rs
```diff
@@ -41,13 +41,10 @@ impl MerkleTree {
...
 
-    /// Returns the root hash of the merkle tree.
+    /// Returns the root hash of the merkle tree as a fixed sized array of 32 bytes.
     #[must_use]
     pub fn root(&self) -> [u8; 32] {
-        sha2::digest::Output::<Sha256>::from_slice(self.0.root().as_bytes())
-            .as_slice()
-            .try_into()
-            .expect("sha256 output is always 32 bytes")
+        (*self.0.root().as_bytes()).into()
```
SuperFluffy, line 47: The above was unecessary. Clippy complained about this method not having a `# Panics`. No idea why this wasn't triggered earlier.

crates/astria-sequencer-validation/src/proof.rs
```diff
@@ -26,12 +26,12 @@ use sha2::Sha256;
 /// inclusion_proof.verify(&value, root).unwrap();
 /// ```
 #[derive(Debug, Default)]
-pub struct MerkleTree(CtMerkleTree<Sha256, Vec<u8>>);
+pub struct MerkleTree<T: AsRef<[u8]>>(CtMerkleTree<Sha256, T>);
```
SuperFluffy, line 29: We are now using merkle trees with both `Vec<u8>` and `[u8; u32]`. It turns out that `CtMerkleTree` supports both, but unfortunately it has the trait bound on its type definitions. Maybe we should fork it and remove it.

crates/astria-sequencer-types/src/namespace.rs
```diff
@@ -41,7 +41,7 @@ impl Namespace {
     /// returns an 10-byte namespace given a byte slice by hashing
     /// the bytes with sha256 and returning the first 10 bytes.
     #[must_use]
-    pub fn from_slice(bytes: &[u8]) -> Namespace {
+    pub fn with_hashed_bytes(bytes: &[u8]) -> Namespace {
```
SuperFluffy, line 44: Strictly speaking this does not have anything to do with the rest, but it bugged me because `from_slice` is much more innocent than what actually happens. Also this brings naming in line with `ChainId::with_hashed_bytes`

crates/astria-sequencer-types/src/sequencer_block_data.rs
```diff
@@ -38,9 +38,10 @@ pub enum Error {
 }
 
 /// Rollup data that relayer/conductor need to know.
-#[derive(Default, Serialize, Deserialize, Clone, Debug, Eq, PartialEq)]
```
SuperFluffy, line 41: Removed this because `ChainId` has no natural default: chain ID is constructed by hashing some bytes using sha256, and a zero-hash is a valid sha256 hash.

crates/astria-conductor/src/types.rs
```diff
@@ -22,7 +22,7 @@ impl SequencerBlockSubset {
     pub(crate) fn from_sequencer_block_data(
         data: SequencerBlockData,
         namespace: Namespace,
-    ) -> Self {
+    ) -> Option<Self> {
```
SuperFluffy, line 25: The change in this signature followes from the removal of the `Default` impl for `RollupData`

crates/astria-conductor/src/executor.rs
```diff
@@ -152,13 +152,13 @@ impl<C: ExecutionClient> Executor<C> {
...
                         block_height: block.header().height.value(),
                     })?;
-                    if let Err(e) = self
-                        .execute_block(SequencerBlockSubset::from_sequencer_block_data(
-                            *block,
-                            self.namespace,
-                        ))
-                        .await
-                    {
+                    let Some(block_subset) =
```
SuperFluffy, line 155: Note the comment on `from_sequencer_block_data`. This change does not affect the execution of this block because `execute_block` checks again if the rollup data inside the block was empty (and we now mapped the case of "no rollup under the name, so use a default with no data" to `None`)

crates/astria-composer/tests/blackbox/composer.rs
```diff
@@ -61,22 +60,33 @@ async fn tx_from_two_rollups_are_received_by_sequencer() {
 /// sequence action is for the given `expected_chain_id`.
 async fn mount_broadcast_tx_sync_mock(
     server: &MockServer,
-    expected_chain_id: &'static str,
+    expected_chain_name: &'static str,
```
SuperFluffy, line 63: Changed this to better disambiguate between the chain name and its ID which was derived from the name.

crates/astria-proto/src/native/sequencer/v1alpha1.rs
```diff
@@ -3,18 +3,720 @@ use std::{
...
+impl Display for IncorrectChainIdLength {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "expected 32 bytes, got {}", self.received)
+    }
+}
+
+impl Error for IncorrectChainIdLength {}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct ChainId(pub [u8; CHAIN_ID_LEN]);
```
SuperFluffy, line 646: I am a bit torn on using a newtype wrapper here. It makes constructing chain IDs more straightforward by using `ChainId::with_hashed_bytes("my_test_chain")`, but at the expense of introducing yet another type. So the signature of `sequence_action` could instead be:
```rust
struct SequenceAction {
    chain_id: [u8; 32],
    data: Vec<u8>,
}
```
which is a bit more obvious. Then we'd provide utilities like `chain_id::with_hashed_bytes -> [u8; 32]` and `chain_id::try_from_slice -> [u8; 32]`, which are also a biy ugly.

crates/astria-proto/src/native/sequencer/v1alpha1.rs
SuperFluffy, line 1: I have not provided doc comments on all methods in here because the PR was getting very lengthy and the function signatures were self explanatory.

crates/astria-sequencer/src/transaction/mod.rs
```diff
@@ -1,9 +1,131 @@
...
+
+    #[instrument(
+        skip_all,
+        fields(
+            nonce = self.nonce,
+            from = from.to_string(),
+        )
+    )]
+    async fn execute<S: StateWriteExt>(&self, state: &mut S, from: Address) -> anyhow::Result<()> {
+        // TODO: make a new StateDelta so this is atomic / can be rolled back in case of error
```
sambukowski, line 101: Is there an issue for this?

sambukowski (approved):

crates/astria-sequencer/src/transaction/mod.rs
SuperFluffy: This came over when I moved the code around. @noot: is this still up to date?

Maybe we can have a small PR that adds links to issues (not this PR though)

crates/astria-sequencer/src/service/mempool.rs
```diff
@@ -40,37 +37,55 @@ impl Service<MempoolRequest> for Mempool {
...
+                info: "failed decoding bytes as a protobuf SignedTransaction".into(),
+                ..response::CheckTx::default()
+            };
+        }
+    };
+    let signed_tx = SignedTransaction::try_from_raw(raw_signed_tx).unwrap();
+    // if the tx passes the check, status code 0 is returned.
+    // TODO: status codes for various errors
+    // TODO: offload `check_stateless` using `deliver_tx_bytes` mechanism
+    //       and a worker task similar to penumbra
```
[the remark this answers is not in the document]
SuperFluffy: As above (copied the text):

This came over when I moved the code around. @noot: is this still up to date?

Maybe we can have a small PR that adds links to issues (not this PR though)

crates/astria-composer/src/searcher/collector.rs
```diff
@@ -10,6 +10,7 @@ use ethers::providers::{
     Ws,
 };
 use humantime::format_duration;
+use proto::native::sequencer::v1alpha1::ChainId;
```
noot, line 13: I'm wondering if we should rename `proto` to `types` or something, and turns the protos into just a module within that crate. it's a bit confusing to have the native types inside a crate called `proto`

crates/astria-composer/src/searcher/collector.rs
```diff
@@ -44,7 +45,8 @@ pub(super) struct Transaction {
 pub(super) struct Collector {
     // Chain ID to identify in the astria sequencer block which rollup a serialized sequencer
     // action belongs to.
-    chain_id: String,
+    chain_id: ChainId,
+    chain_name: String,
```
noot, line 49: why do we need to store this?

crates/astria-composer/src/searcher/collector.rs
```diff
@@ -73,16 +75,18 @@ impl Status {
...
-        chain_id: String,
+        chain_name: String,
         url: String,
         searcher_channel: Sender<Transaction>,
     ) -> eyre::Result<Self> {
         let client = EthClient::connect(&url)
             .await
             .wrap_err("failed connecting to eth")?;
         let (status, _) = watch::channel(Status::new());
+        let chain_id = ChainId::with_hashed_bytes(&chain_name);
```
noot, line 86: not sure how I feel about this, since the `chain_id` was supposed to just be some plaintext identifier of a chain, with a maximum length of 32. it's weird to have a chain_name and chain_id imo, can get confusing

crates/astria-conductor/src/executor.rs
```diff
@@ -152,13 +152,13 @@ impl<C: ExecutionClient> Executor<C> {
...
+                        SequencerBlockSubset::from_sequencer_block_data(*block, self.namespace)
+                    else {
+                        info!(namespace = %self.namespace, "block did not contain data for namespace; skiping");
```
noot, line 158:
```suggestion
                        info!(namespace = %self.namespace, "block did not contain data for namespace; skipping");
```

crates/astria-proto/src/native/sequencer/v1alpha1.rs
```diff
@@ -3,18 +3,721 @@ use std::{
     fmt::Display,
 };
 
-use crate::generated::sequencer::v1alpha1;
+use ed25519_consensus::{
```
noot, line 6: is there a reason to have all of these types in 1 file? it's rather long lol

crates/astria-proto/src/native/sequencer/v1alpha1.rs
```diff
@@ -3,18 +3,721 @@ use std::{
...
+        let n_raw_actions = actions.len();
+        let actions: Vec<_> = actions
+            .into_iter()
+            .filter_map(|raw_act| {
+                // Silently drop unset actions.
+                let converted = Action::try_from_raw(raw_act);
+                if let Err(e) = &converted {
+                    if e.is_unset() {
+                        return None;
+                    }
```
noot, line 262: I think this should be an error, is there any non-error case where this would happen?

crates/astria-proto/src/native/sequencer/v1alpha1.rs
noot: I think a newtype is fine, but idk if I like the `with_hashed_bytes`, since we now have a plaintext string we input, then hash to a `chain_id`, then hash again for a `Namespace`, which seems overkill. would prefer just having a `ChainId::new()` which takes some bytes and pads them to 32 bytes if needed

crates/astria-sequencer-relayer/src/data_availability.rs
```diff
@@ -160,7 +164,8 @@ impl NamespaceData for SequencerNamespaceData {}
 #[derive(Serialize, Deserialize, Debug)]
 pub struct RollupNamespaceData {
     pub(crate) block_hash: Hash,
-    pub(crate) chain_id: Vec<u8>,
+    #[serde(with = "sequencer_types::serde::chain_id")]
```
noot, line 167: why put this here instead of implementing on the `ChainId` type?

crates/astria-sequencer-validation/src/proof.rs
noot: where are we using it with `[u8; u32]` for leaves?

noot: I see `[u8; 64]` (maybe you meant that)

crates/astria-sequencer/src/proposal/commitment.rs
```diff
@@ -26,18 +26,32 @@ use crate::transaction::Signed;
...
-        .filter_map(|tx_bytes| match Signed::try_from_slice(&tx_bytes) {
-            Ok(tx) => Some((tx, tx_bytes)),
-            Err(err) => {
-                tracing::debug!(tx_bytes = hex::encode(tx_bytes), error = ?err, "failed to deserialize tx bytes");
-                None
-            }
+        .filter_map(|bytes| {
+            raw::SignedTransaction::decode(&*bytes)
+            .map_err(|err| {
+                info!(error = ?err, "failed to deserialize bytes as a signed transaction");
```
noot, line 38: info seems high potentially

crates/astria-sequencer/src/proposal/commitment.rs
```diff
@@ -178,37 +195,37 @@ mod test {
...
         };
 
         let signed_tx = tx.into_signed(&signing_key);
-        let tx_bytes = signed_tx.to_proto().encode_to_vec();
+        let tx_bytes = signed_tx.into_raw().encode_to_vec();
         let txs = vec![tx_bytes.into()];
-        let (action_commitment, _) = generate_sequence_actions_commitment(txs);
+        let (actual, _) = generate_sequence_actions_commitment(txs);
 
-        let expected_commitment: [u8; 32] = [
```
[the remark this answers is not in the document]
noot: don't think that's possible unless we change the encoding of transactions within the sequencer to be independent of the wire format (but idk if that would even make sense to do)

crates/astria-sequencer/src/service/mempool.rs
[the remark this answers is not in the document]
noot: I only wrote the first todo which is https://github.com/astriaorg/astria/issues/228 

idk about the second todo

crates/astria-sequencer/src/transaction/mod.rs
```diff
@@ -1,9 +1,131 @@
...
+
+pub(crate) async fn execute<S: StateWriteExt>(
+    tx: &SignedTransaction,
+    state: &mut S,
+) -> anyhow::Result<()> {
+    let signer_address = Address::from_verification_key(tx.verification_key());
+    tx.unsigned_transaction()
+        .execute(state, signer_address)
+        .await
+}
```
noot, line 43: why not implement these on `SignedTransaction` (like it was before)?

crates/astria-sequencer/src/transaction/mod.rs
noot: https://github.com/astriaorg/astria/issues/317

noot (changes requested):
looks good overall, going to block on the chain_id thing since it's getting a bit overcomplicated imo (see PR comments)

crates/astria-sequencer/src/proposal/commitment.rs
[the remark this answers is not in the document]
SuperFluffy: Does it make sense to use borsh then for a specific subset of our public interfaces? They promise determinism and provide implementations in go, python, javascript/typescript, and obviously rust that seem reasonably up to date.

crates/astria-sequencer-validation/src/proof.rs
SuperFluffy: Yeah, I meant 64 bytes. Thanks for spotting!

crates/astria-composer/src/searcher/collector.rs
SuperFluffy: Yeah, definitely. The naming is not optimal. We should rename `astria-proto` and make `proto` a submodule of that. I don't know what to be honest. I am leaning toward `astria-api` or `astria-rpc` or `astria-public`. Or even just `astria`. Not in this PR obviously, but agree.

crates/astria-composer/src/searcher/collector.rs
SuperFluffy: Mainly to give `Collector` an identity. This will be nice once we have it report its state (in a health check) and when constructing its scope when calling its `run` function (or whatever it's called).

crates/astria-proto/src/native/sequencer/v1alpha1.rs
SuperFluffy: Do you mean `ed25519_consensus` or this module in general? It's not so uncommon to have relatively large rust files as long as it makes sense that they are grouped together, which I think is the case here.

crates/astria-sequencer-relayer/src/data_availability.rs
SuperFluffy: I did not want to serde as yet another dependency to astria-proto for something that (at the moment) is sequencer-relayer specific. It will probably end up there, but confining this change to just sequencer-relayer seemed like the right choice.

crates/astria-sequencer/src/proposal/commitment.rs
SuperFluffy: Do you mean it should be `warn!` or `debug!`?

crates/astria-sequencer/src/transaction/mod.rs
SuperFluffy: Because `ActionHandler::execute` takes a `signer_address`, but `SignedTransaction` is the thing that a) constructs the signer address and then b) calls `<UnsignedTransaction as ActionHandler>::execute` with it.

So implementing `ActionHandler for SignedTransaction` does not make sense.

On the other hand doing `impl SignedTransaction { fn execute(...) -> ... }` the way it was before doesn't work either because of coherence (`SignedTransaction` is an external type).

The only change I think might make sense is to change `ActionHandler` entirely to the definition below and `impl ActionHandler for SignedTransaction` and remove the impl for `UnsignedTransaction`:
```rust
trait ActionHandler {
    async fn execute<S: StateWriteExt>(&self, state: &mut S) -> Result<()> {
        todo!()
    }
}
```
But I didn't want to redesign it just that much

crates/astria-proto/src/native/sequencer/v1alpha1.rs
SuperFluffy: Chances are an end user directly constructs a raw protobuf with `actions: vec![Some(SequenceTx), None, None, Some(TransferTx)]` or, in go, populating the list with `nil` by accident. I felt like rejecting the entire tx when you sent a valid protobuf is too harsh if just treating "empty" actions as being absent.

crates/astria-composer/src/searcher/collector.rs
SuperFluffy: After thinking it through I agree. There is something off about demanding a fixed length.

At the same time I think a maximum length of 32 is too strict. Should we change this type to a `string` instead, and then internally hash it?

crates/astria-sequencer-relayer/src/data_availability.rs
```diff
@@ -160,6 +160,7 @@ impl NamespaceData for SequencerNamespaceData {}
 #[derive(Serialize, Deserialize, Debug)]
 pub struct RollupNamespaceData {
     pub(crate) block_hash: Hash,
+    #[serde(with = "hex::serde")]
```
SuperFluffy, line 163: Note that this was serialize as a json list of integers. Makes sense to change the deser format if everything else is being changed as well.

crates/astria-sequencer-relayer/src/data_availability.rs
```diff
@@ -588,9 +589,10 @@ fn assemble_blobs_from_sequencer_block_data(
             .map_err(|e| eyre::eyre!(e))
             .context("failed to generate inclusion proof")?;
 
+        let namespace = Namespace::from_slice(&chain_id);
```
SuperFluffy, line 592: Just shifted this above to avoid an unncessary clone.

crates/astria-sequencer/src/transaction/mod.rs
SuperFluffy: Added this to the PR

crates/astria-conductor/src/types.rs
SuperFluffy: Note that `RollupData` is again using a `Vec<u8>` as a chain ID. But I am not going to re-add the `Default` impl because it does not make sense to have it.

crates/astria-sequencer/src/service/mempool.rs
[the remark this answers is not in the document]
SuperFluffy: I created https://github.com/astriaorg/astria/issues/319 and added it to the TODO

crates/astria-sequencer-relayer/src/data_availability.rs
SuperFluffy: Resolved by removing `ChainId` for now.

SuperFluffy:
@noot removed `ChainId`. It did not fit this PR very well, so I was happy to remove it. Also responded to the other comments.

crates/astria-proto/src/native/sequencer/v1alpha1.rs
noot: sorry yeah, i meant the whole file. okay, makes sense

crates/astria-proto/src/native/sequencer/v1alpha1.rs
```diff
@@ -3,18 +3,592 @@ use std::{
...
+    pub fn try_from_raw(proto: raw::SignedTransaction) -> Result<Self, SignedTransactionError> {
+        use crate::Message as _;
+        let raw::SignedTransaction {
+            signature,
+            public_key,
+            transaction,
+        } = proto;
+        let signature =
+            Signature::try_from(&*signature).map_err(SignedTransactionError::signature)?;
+        let verification_key = VerificationKey::try_from(&*public_key)
```
noot, line 145: why the `&*public_key` instead of `public_key.as_slice()`?

crates/astria-proto/src/native/sequencer/v1alpha1.rs
noot: > actions: vec![Some(SequenceTx), None, None, Some(TransferTx)]

this wouldn't be a valid transaction/set of actions though, since a `None` or `nil` action doesn't make sense. also allowing `None` would result in different transactions if someone makes a `raw`, converts it to native, and converts it back to `raw` which seems bad

crates/astria-proto/src/native/sequencer/v1alpha1.rs
```diff
@@ -60,14 +634,16 @@ impl From<[u8; ADDRESS_LEN]> for Address {
 
 impl Display for Address {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        use hex::ToHex as _;
-        f.write_str(&self.encode_hex::<String>())
+        for byte in self.0 {
+            write!(f, "{byte:02x}")?;
+        }
```
noot, line 639: is this better than using `encode_hex`?

crates/astria-sequencer/src/proposal/commitment.rs
noot: probably `debug!` here

crates/astria-sequencer/src/transaction/mod.rs
noot: that makes sense, I didn't necessarily mean `impl ActionHandler for SignedTransaction` but you could make these methods on `SignedTransaction`?

noot (approved):
looks good, approving for now as not to block, but still a few comments that should be addressed before merging!

crates/astria-proto/src/native/sequencer/v1alpha1.rs
SuperFluffy: Yeah, that's indeed a problem which lead me to open #310. I am going to change this to return early with an error because my change is out of scope for this PR. Thank you for raising this!

crates/astria-sequencer/src/proposal/commitment.rs
SuperFluffy: Changed it to debug

crates/astria-sequencer/src/transaction/mod.rs
SuperFluffy: Unfortunately you can only implement inherent implementations in the crate that owns the original definition (so in this case, `astria-proto`). But doing that there does not make sense because this is sequencer-private functionality.

I could make a single-use trait just to implement these methods on `SignedTransaction`. Or I could make a newtype wrapper for `SignedTransaction`. But both seem overkill.

crates/astria-proto/src/native/sequencer/v1alpha1.rs
SuperFluffy: No reason other than it being more concise and . Unfortunately I cannot do `Signature::try_from(&signature)` because to make use of deref coercion because rust does not do deref coercion when matching traits (so the fact that `Vec<T>` derefs to`[T]` does not help). So I have to use`&*` to do that manually.

crates/astria-proto/src/native/sequencer/v1alpha1.rs
SuperFluffy: @noot forgot to answer: this avoids an allocation. The hex crate only has methods that allocate to a string, whereas this writes directly to the formatter without an extra allocation.

PR #328, opened 2023-09-05
initial commit of cometbft test utility

SuperFluffy:
## Summary
Added a crate to launch cometbft as a subprocess.

## Background
PR https://github.com/astriaorg/astria/pull/327 runs tests against the output of cometbft. It manually adds magic hex strings that are then processed with functions it defines. This is not optimal because there is no easy way to reproduce the strings nor can the code be run against actual responses from cometbft.

This utility allows launching cometbft as a subprocess in tests with its rpc server bound to a random port `0`. The launcher returns successfully if it was able to find the TCP port on which cometbft is serving requests.

## Changes
* Added a `cometbft-launcher` crate
* Added a step in the rust test job to install cometbft v0.37.2 

## Testing
* For now only the functionality of the crate is tested using a blackbox test.

## Related Issues
This is useful in PR https://github.com/astriaorg/astria/pull/327.

### DIFF
--- a/crates/test/cometbft-launcher/src/lib.rs
+++ b/crates/test/cometbft-launcher/src/lib.rs
@@ -5,7 +5,7 @@
 //! # tokio_test::block_on(async {
 //! use cometbft_launcher::CometBft;
 //! let cometbft = CometBft::builder()
-//!     .proxy_app("http://127.0.0.1:26657")
+//!     .proxy_app("tcp://127.0.0.1:26657")
 //!     .launch()
 //!     .await
 //!     .unwrap();
--- a/crates/test/cometbft-launcher/src/lib.rs
+++ b/crates/test/cometbft-launcher/src/lib.rs
@@ -1,0 +1,318 @@
+//! Launch cometbft as a subprocess for running tests.
+//!
+//! # Examples
+//! ```no_run
+//! # tokio_test::block_on(async {
+//! use cometbft_launcher::CometBft;
+//! let cometbft = CometBft::builder()
+//!     .proxy_app("http://127.0.0.1:26657")
+//!     .launch()
+//!     .await
+//!     .unwrap();
+//! # })
+
+use std::{
+    ffi::{
+        OsStr,
+        OsString,
+    },
+    net::SocketAddr,
+    path::Path,
+    process::Stdio,
+    time::Duration,
+};
+
+use anyhow::{
+    ensure,
+    Context as _,
+};
+use tempfile::TempDir;
+use tokio::process::{
+    Child,
+    Command,
+};
+
+/// CometBFT running in a subprocess.
+pub struct CometBft {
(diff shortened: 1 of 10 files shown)


crates/test/cometbft-launcher/src/lib.rs
```diff
@@ -0,0 +1,336 @@
+//! Launch cometbft as a subprocess for running tests.
+//!
+//! # Examples
+//! ```no_run
+//! # tokio_test::block_on(async {
+//! use cometbft_launcher::CometBft;
+//! let cometbft = CometBft::builder()
+//!     .proxy_app("tcp://127.0.0.1:26657")
```
noot, line 8:
```suggestion
//!     .proxy_app("tcp://127.0.0.1:26658")
```

the app default port is at 26658

crates/test/cometbft-launcher/src/lib.rs
```diff
@@ -0,0 +1,336 @@
...
+    for fd in listpidinfo::<ListFDs>(pid, info.pbi_nfiles as usize)
+        .map_err(Error::msg)
+        .context("failed listing file descriptors associated with pid")?
+    {
+        if let ProcFDType::Socket = fd.proc_fdtype.into() {
+            let socket = pidfdinfo::<SocketFDInfo>(pid, fd.proc_fd)
+                .map_err(Error::msg)
+                .context("failed getting info on socket fd")?;
+            if let SocketInfoKind::Tcp = socket.psi.soi_kind.into() {
+                // unsafe ok because union is discriminated by soi_kind
```
noot, line 206: what does this mean exactly?

noot (approved):
this looks good! can you open an issue describing which tests we could write/update to use this when it's merged? (if there are others than the last_commit_hash test)

crates/test/cometbft-launcher/src/lib.rs
SuperFluffy: These are all C structs read into Rust. C does not have a language native concept of a tagged union, so you need to manually keep track of what you store in your union. In Rust the equivalent to an `enum` is actually something like this:
```rust
struct MyTaggedUnion {
    tag: u8,
    the_union: MyUnion,
}

union MyUnion {
    Number(u64),
    Str(String),
}

const THE_NUMBER: u8: 0;
const THE_STRING: u8: 1;

# an instance of my tagged union
const demo: MyTaggedUnion = MyTaggedUnion {
    tag: THE_NUMBER,
    the_union: MyUnion::Number(42),
};
```

So I am discriminating what's in `MyTaggedUnion` by the `tag` field.

SuperFluffy:
Status on this: I need to actually use this lib in a test so it does not just sit there and bitrot. Moving to draft until then

SuperFluffy:
This is potentially useful, but is sitting idle for close to a year now.

Mothballing this for the time being. Will revisit at a later date.

PR #329, opened 2023-09-06
Update Conductor config management

### LINKED ISSUE #243: update configuration of astria-conductor to config standard

dereksione:
## Summary
Updating how the conductor does configuration management to use the specs defined in the configuration specs

## Background
Addressing the tech debt of different crates using different configuration setups and hierarchies

## Changes
- Removed setting up config using a TOML file
- Removed setting up config using Cli overrides (not 100% sure on this one, open to hearing thoughts)
- Added Justfile
- Updated readme with just instructions
- Added example env

## Testing
All written tests pass, and additional tests to test for proper figment functionality + config derivation has been added.

## Breaking Changelist
- All config management should be done using `.env` files now, as outlined in the spec - cli overrides and TOML file support is removed

closes #243


crates/astria-conductor/README.md
```diff
@@ -1,29 +1,66 @@
...
+We use [just](https://just.systems/man/en/chapter_4.html) for convenient project
+specific commands.
 
-* run `cargo run`
-* alternatively, you can do `cargo build && ./target/debug/conductor`.
-* to connect directly to a node via p2p, you can use the `--bootnodes` flag,
-  e.g. `--bootnodes=/ip4/127.0.0.1/tcp/33900` or
-  `--bootnodes=/ip4/127.0.0.1/tcp/34471/p2p/12D3KooWDCHwgGetpJuHknJqv2dNbYpe3LqgH8BKrsYHV9ALpAj8`.
+### Configuration
+Composer is configured via environment variables. An example configuration can
```
steezeburger, line 13:
```suggestion
Conductor is configured via environment variables. An example configuration can
```

crates/astria-conductor/local.env.example
```diff
@@ -0,0 +1,14 @@
...
+ASTRIA_CONDUCTOR_TENDERMINT_URL = "http://localhost:1318"
+
+# Chain ID
+ASTRIA_CONDUCTOR_CHAIN_ID = "ethereum"
+
+# Execution RPC URL
+ASTRIA_CONDUCTOR_EXECUTION_RPC_URL = "http://localhost:50051"
+
+# disable block finalization
+ASTRIA_CONDUCTOR_DISABLE_FINALIZATION=false
```
steezeburger, line 14: nit: new line for eof

crates/astria-conductor/src/bin/astria-conductor.rs
```diff
@@ -41,17 +30,16 @@ async fn main() -> Result<()> {
...
-    let args = Cli::parse();
+    let conf = match config::get() {
+        Ok(cfg) => cfg,
+        Err(e) => {
+            eprintln!("failed to read configuration: {e}");
+            std::process::exit(2);
+        }
+    };
+
     // hierarchical config. cli args override Envars which override toml config values
```
steezeburger, line 41: This comment is no longer valid if we get rid of cli args. It could be edited and moved up though.

steezeburger (changes requested):

steezeburger:
I'm not sure we want to get rid of the cli args. CLI args are mentioned in the [config standard ](https://github.com/astriaorg/astria/blob/b72db74959e17b73ecd3ef97c9aafb0362c87541/specs/config-standard.md#L57), and we use them in dev-cluster. 

We could update the dev-cluster to use all envars, but I do think the flexibility that cli overrides allow is nice.

dereksione:
> I'm not sure we want to get rid of the cli args. CLI args are mentioned in the [config standard ](https://github.com/astriaorg/astria/blob/b72db74959e17b73ecd3ef97c9aafb0362c87541/specs/config-standard.md#L57), and we use them in dev-cluster.
> 
> We could update the dev-cluster to use all envars, but I do think the flexibility that cli overrides allow is nice.

@joroshiba ?

SuperFluffy:
The standard doc mentions this for CLI tools. Services are intended to have a single point of entry + single source of truth.

crates/astria-conductor/Cargo.toml
```diff
@@ -9,10 +9,9 @@ sync_wrapper = "0.1.2"
 
 async-trait = { workspace = true }
 base64 = { workspace = true }
-clap = { workspace = true, features = ["derive"] }
 color-eyre = { workspace = true }
 ed25519-consensus = { workspace = true }
-figment = { workspace = true, features = ["toml", "env"] }
+figment = { workspace = true, features = ["test", "env"] }
```
SuperFluffy, line 14: Only activate the `test` feature in tests, i.e. `dev-dependencies`

crates/astria-conductor/justfile
```diff
@@ -0,0 +1,18 @@
+alias r := run
```
SuperFluffy, line 1: Nit: I wouldn't complicate it with extra abbreviations.

crates/astria-conductor/local.env.example
SuperFluffy, line 1: Remove the spaces around `=` since these are intended to be env files.

crates/astria-conductor/src/bin/astria-conductor.rs
```diff
@@ -41,17 +30,16 @@ async fn main() -> Result<()> {
 }
 
 async fn run() -> Result<()> {
-    let args = Cli::parse();
+    let conf = match config::get() {
+        Ok(cfg) => cfg,
+        Err(e) => {
+            eprintln!("failed to read configuration: {e}");
```
SuperFluffy, line 36: Return the result of `config::get` + some error context here and then let `main` a) report the error and b) call`std::process::exit`

crates/astria-conductor/src/bin/astria-conductor.rs
```diff
@@ -41,17 +30,16 @@ async fn main() -> Result<()> {
...
-    let conf: Config = Figment::new()
-        .merge(Toml::file("ConductorConfig.toml"))
-        .merge(Env::prefixed("RUST_").split("_").only(&["log"]))
-        .merge(Env::prefixed("ASTRIA_"))
-        .merge(Serialized::defaults(args))
-        .extract()?;
-
-    telemetry::init(std::io::stdout, conf.log.as_deref().unwrap_or("info"))
-        .expect("failed to initialize telemetry");
+    telemetry::init(std::io::stdout, &conf.log).expect("failed to initialize telemetry");
```
SuperFluffy, line 42: Since we are in a `run()` function that returns back to `main` we should just `wrap_err("failed to initialize telemtry")?` if we touch this line.

crates/astria-conductor/src/config.rs
SuperFluffy, line 1: Remove all defaults in this file. Instead use the defaults defined here for the `local.env.example`

SuperFluffy (changes requested):
The `local.env` file is currently broken because POSIX shell does not understand spaces around assignments.

Please remove all defaults in `config`. The `local.env.example` is kind of understood as the "defaults config".

joroshiba:
> I'm not sure we want to get rid of the cli args. CLI args are mentioned in the [config standard ](https://github.com/astriaorg/astria/blob/b72db74959e17b73ecd3ef97c9aafb0362c87541/specs/config-standard.md#L57), and we use them in dev-cluster.
> 
> We could update the dev-cluster to use all envars, but I do think the flexibility that cli overrides allow is nice.

All env vars is the standard for services, CLI args are mentioned in the context of CLI tools, where services aren't that.

The idea is a forcing function to running with best practices. We don't need to enable people to use non-best practices. CLI tools (and defaults) are mostly for local dev, we have added helper methods to the `justfile` with dot file loading to make it easy to run locally still.

crates/astria-conductor/local.env.example
```diff
@@ -0,0 +1,14 @@
+# Data Availability service url (Celestia node in this case)
+ASTRIA_CONDUCTOR_CELESTIA_NODE_URL = "http://localhost:26659"
+
+# Tendermint URL
+ASTRIA_CONDUCTOR_TENDERMINT_URL = "http://localhost:1318"
```
noot, line 5:
```suggestion
ASTRIA_CONDUCTOR_TENDERMINT_URL = "http://localhost:26657"
```

crates/astria-conductor/src/bin/astria-conductor.rs
```diff
@@ -41,17 +30,16 @@ async fn main() -> Result<()> {
 }
 
 async fn run() -> Result<()> {
-    let args = Cli::parse();
+    let conf = match config::get() {
+        Ok(cfg) => cfg,
+        Err(e) => {
+            eprintln!("failed to read configuration: {e}");
+            std::process::exit(2);
```
noot, line 37: why 2?

crates/astria-conductor/justfile
dereksione: I don't feel strongly about this at all, but isn't it a very common pattern to alias most used commands?

crates/astria-conductor/src/bin/astria-conductor.rs
dereksione: I've always coded failing to read essential config for CLI tools with exit status 2, but open to suggestions

crates/astria-conductor/justfile
SuperFluffy: Well, "common" here is probably by less than 10 people and only 3 times.

It's IMO more confusing to have abbreviations for very rarely used tools

crates/astria-conductor/src/bin/astria-conductor.rs
dereksione: Changing to 78 after reading  https://freedesktop.org/software/systemd/man/systemd.exec.html#Process%20Exit%20Codes

crates/astria-conductor/src/config.rs
```diff
@@ -65,25 +69,48 @@ where
     ))
 }
 
-// NOTE - using default fns instead of defaults in Cli because defaults
-//   in Cli always override values from a Config file, which we don't want.
```
joroshiba, line 69: I think rust lint complaining about this.

joroshiba (approved):
Other than test failure, LGTM

crates/astria-conductor/src/bin/astria-conductor.rs
```diff
@@ -35,25 +27,24 @@ use tracing::{
     instrument,
 };
 
+const EXIT_CONFIG_CODE: i32 = 78;
```
SuperFluffy, line 30: Can you link the systemd man page for this and maybe call it `EX_CONFIG` so other folks immediately know where this comes from?

We should probably also create a tracking issue to address this in all our services and implement this somehow: https://doc.rust-lang.org/std/process/trait.Termination.html

crates/astria-conductor/src/bin/astria-conductor.rs
```diff
@@ -35,25 +27,24 @@ use tracing::{
...
+
 #[tokio::main]
 async fn main() -> Result<()> {
-    run().await?;
+    if let Err(e) = run().await {
+        eprintln!("Exited with error: {e}");
```
SuperFluffy, line 35:
```suggestion
        eprintln!("Exited with error:\n{e:?}");
```
Please print the entire eyre error backtrace.

crates/astria-conductor/src/bin/astria-conductor.rs
```diff
@@ -35,25 +27,24 @@ use tracing::{
...
+        // FIXME: might have to bubble up exit codes, since we might need 
+        //        to exit with other exit codes if something else fails
+        std::process::exit(EXIT_CONFIG_CODE);
```
SuperFluffy, line 38: Yeah, looks like thisis ok for now but will need to be fixed. Can you create an issue to track this and provide the url of the issue in the `FIXME(https://github.com/...)` above?

crates/astria-conductor/local.env.example
```diff
@@ -0,0 +1,20 @@
+# Log Level
+ASTRIA_CONDUCTOR_LOG="info"
```
SuperFluffy, line 2: Maybe `"astria_conductor=info"`?

crates/astria-conductor/local.env.example
SuperFluffy, line 1: This file also serves as documentation, so it would be great if there was a bit more info similar to composer and sequencer relayer. Can you take them as inspiration and add a few more explanations?

You can skip the p2p/gossipnet bits because they are going to be removed soon.

crates/astria-conductor/src/bin/astria-conductor.rs
```diff
@@ -35,25 +27,24 @@ use tracing::{
...
-        .merge(Serialized::defaults(args))
-        .extract()?;
-
-    telemetry::init(std::io::stdout, conf.log.as_deref().unwrap_or("info"))
-        .expect("failed to initialize telemetry");
+    let conf = config::get().wrap_err("failed to read config")?;
+
+    telemetry::init(std::io::stdout, &conf.log).wrap_err("failed to initialize telemetry")?;
 
     info!("Using chain ID {}", conf.chain_id);
```
SuperFluffy, line 49: While we are changing configs anyway, can you serialize the config as json and print all of it here instead of just a few bits over several lines? Have a look at how the other services are doing it. It's good practice to tell people immediately what the config is at service startup.

SuperFluffy (approved):
Approving because I don't see blockers, but please address the remaining comments in this review.

crates/astria-conductor/src/bin/astria-conductor.rs
```diff
@@ -35,30 +27,28 @@ use tracing::{
...
-
-    info!("Using chain ID {}", conf.chain_id);
-    info!("Using Celestia node at {}", conf.celestia_node_url);
-    info!("Using execution node at {}", conf.execution_rpc_url);
-    info!("Using Tendermint node at {}", conf.tendermint_url);
+    let conf = config::get().wrap_err("failed to read config")?;
+
+    telemetry::init(std::io::stdout, &conf.log).wrap_err("failed to initialize telemetry")?;
+
+    info!("Available config is: \n{}", serde_json::json!(conf));
```
SuperFluffy, line 51: Please use structured logging

steezeburger (approved):

PR #335, opened 2023-09-06
Conductor: More structured logging

### LINKED ISSUE #14: Replace logging by tracing, remove alert actor

steezeburger:
## Summary
This PR adds more structured logs to the Conductor. It also instruments several functions and adds an `executor_span` and `reader_span`.

## Background
We want better logging to improve traceability, debugging, monitoring, and analytics about the platform.

## Changes
- adds structured information to `error!` calls
- instruments driver's `new` function
- instruments actors' `spawn` functions with custom spans
  - calls actors' `run` functions in these custom parent spans
- instruments actor's `run` functions

## Testing
How are these changes tested?

## Breaking Changelist
N/A

## Related Issues
https://github.com/astriaorg/astria/issues/14

closes #14


crates/astria-conductor/src/bin/astria-conductor.rs
```diff
@@ -118,7 +118,12 @@ async fn run() -> Result<()> {
             _ = stop_rx.changed() => {
                 info!("shutting down conductor");
                 if let Some(e) = driver_tx.send(DriverCommand::Shutdown).err() {
-                    error!("error sending Shutdown command to driver: {}", e);
+                    error!(
+                        actor_name = "driver",
+                        error.msg = %e,
+                        error.cause = ?e,
```
noot, line 124: will this end up printing the most recent error twice (since it prints it as both display + debug)?

noot (approved):

noot:
also does this close #14?

SuperFluffy:
Especially for repeated fields you can make use spans. Their fields are automatically added to the events that are emitted within them so you don't need to repeat e.g. `actor_name = "foo"`.

You can get that for free by adding instrumenting your futures with the `#[instrument]` macro and defining your fields as required. See here for more: https://docs.rs/tracing/latest/tracing/attr.instrument.html

Note: this is only necessary for the future running at the highest level (here: `run`), and its span will get propagated to all futures (and functions) running within it. The only exception is futures that are `spawn`ed on the executor, which need to be instrumented explicitly. See the doc comment on the `Instrument::instrument` trait method here: https://docs.rs/tracing/latest/tracing/trait.Instrument.html#method.instrument

steezeburger:
> also does this close #14?

Not quite! I want to improve this structured tracing with `spans` to close #14

sambukowski (approved):

crates/astria-conductor/src/bin/astria-conductor.rs
steezeburger: I think not for this case because `SendError` has different implementations of display and debug, but I think this would cause duplicate logs for the other places I am using this.

steezeburger: @SuperFluffy Do you have any insight on when to use this syntax? What should it look like for other errors?

sambukowski (approved):

SuperFluffy (approved):
Looks much cleaner.

Re the debug errors: it's not perfect. We'd need to implement Value for the errors somehow so that we get an actual json output instead of actual rust debug prints. :/

PR #336, opened 2023-09-07
sequencer-relayer: update config to astria standard

### LINKED ISSUE #244: update configuration of astria-sequencer-relayer to config standard

SuperFluffy:
## Summary
This patch updates the sequencer-relayer to Astria's config standard for services.

## Background
Introduced in https://github.com/astriaorg/astria/pull/223 the Astria config spec requires services to have a single point of entry and single source of truth when reading config. This patch updates it to that spec.

## Changes
- Remove all config defaults
- Remove CLI config and config parser
- create a `local.env.example` documenting config options
- copy composer's justfile
- remove outdated notes on testing, docker
- update README
- rename smoke -> blackbox tests (because the tests were touched anyways)

## Testing
- there have not been changes to sequencer-relayer's logic
- config creation in blackbox tests has been updated (because config no longer implements Default)
- tests have been added to ensure that `local.env.example` stays in sync with the config type.

## Breaking Changelist
- Because command line options and defaults have been removed it should be assumed that older versions of sequencer-relayer are no longer compatible with the current version of sequencer-relayer (as relates to deployment in k8s).

## Related Issues
Related tracking issue: https://github.com/astriaorg/astria/issues/240
Original spec: https://github.com/astriaorg/astria/issues/223

closes https://github.com/astriaorg/astria/issues/244


crates/astria-sequencer-relayer/README.md
```diff
@@ -1,59 +1,37 @@
...
+We use [just](https://just.systems/man/en/chapter_4.html) for convenient project
+specific commands.
 
-- Rust 1.66
-- kind 0.18.0 (for tests)
-- kubectl 1.26.3 (for tests)
+### Configuration
 
-## Building
+Composer is configured via environment variables. An example configuration can
```
noot, line 17:
```suggestion
Sequencer-relayer is configured via environment variables. An example configuration can
```

crates/astria-sequencer-relayer/local.env.example
```diff
@@ -0,0 +1,45 @@
+# Address of sequencer/cometbft/tendermint to request new blocks.
+# 127.0.0.1:26657 is the default socket address at which cometbft
+# serves RPCs.
+ASTRIA_SEQUENCER_RELAYER_SEQUENCER_ENDPOINT="http://127.0.0.1:26657"
+
+# Address of the celestia node. 127.0.0.1:26658 is the default socket
+# address at which celestia node servces RPCs.
+ASTRIA_SEQUENCER_RELAYER_CELESTIA_ENDPOINT="http://127.0.0.1:26658"
```
noot, line 8: this conflicts with the sequencer app port :// (this is the port ABCI applications use by default). we should probably change the default celestia endpoint?

crates/astria-sequencer-relayer/local.env.example
```diff
@@ -0,0 +1,45 @@
...
+# The port that sequencer relayer will bind on 127.0.0.1 to serve RPCs.
+ASTRIA_SEQUENCER_RELAYER_RPC_PORT=2450
+
+# The port that sequencer relayer will bind to join the gossipnet p2p network.
+ASTRIA_SEQUENCER_RELAYER_P2P_PORT=33900
+
+# A list of bootnodes that sequencer relayer will use to join the gossipnet p2p network.
+# Optional: if left empty sequencer relayer will be a single node in the gossipnet,
+# waiting for peers to join it.
+ASTRIA_SEQUENCER_RELAYER_BOOTNODES=/ip4/1.2.3.4/tcp/1234
```
noot, line 38: should this just be empty for now?

crates/astria-sequencer-relayer/local.env.example
```diff
@@ -0,0 +1,45 @@
...
+
+# The path to the libp2p private key.
+# Optional: if left empty sequencer-relayer will only join permissionless gossipnets.
```
noot, line 41: you can remove this line, the private key is used as authentication on the network (ie. its peer ID, message authentication), not to do with what network it joins

noot (approved):
looks good!

crates/astria-sequencer-relayer/local.env.example
SuperFluffy: Right, so 26658 is even wrong. It's 26657 actually, but that clashes with cometbft's port at `:26657`.

To be honest, all of this is kind of an example and not a recommendation for actual ports to use. It should still be on the person deploying these services to wire them up correctly.

crates/astria-sequencer-relayer/local.env.example
SuperFluffy: This is the difficulty with examples as docs. :( We don't really intend for people to just run these nodes on their systems (it takes way too much work to do all of that by hand). So that's why this is called `.example` not just `local.env` or `default.env`.

I actually wanted to provide a full list of multiaddrs there so that people understand the syntax, i.e. `ASTRIA_SEQUENCER_RELAYER_ BOOTNODES=/ip4/1.2.3.4/tcp/1234,/ip4/1.2.3.4/tcp/1234`.

joroshiba: I think we should probably leave this as blank.

It doesn't really matter once we get rid of gossipnet though. It's going bye bye.

joroshiba (approved):

crates/astria-sequencer-relayer/local.env.example
SuperFluffy: Fair, removed.

Issue #308, opened 2023-08-23
commitment and verification of namespaces in `SequencerNamespaceData`


noot:
`SequencerNamespaceData`, which is written to DA by the relayer, is as follows:

```
pub struct SequencerNamespaceData {
    pub block_hash: Hash,
    pub header: Header,
    pub last_commit: Option<Commit>,
    pub rollup_namespaces: Vec<Namespace>,
}
```

however, `rollup_namespaces` is not committed to during the consensus process, so the relayer can omit a namespace entirely without it being detectable by the conductor.

there should be a commitment of the rollup chain IDs in a block during the consensus process, allowing for verification that there was no data for a rollup completed omitted.

PR #338, opened 2023-09-07
feat(sequencer): implement generating and verifying chain ID commitment

noot:
## Summary
proposers now make a commitment to the `sequence::Action` `chain_id`s inside the block. this will allow for conductors to verify that the sequencer-relayer has not omitted any namespaces in the `SequencerBlockData` they write. (punishment if they do is TBD)

## Background
- required so that we can verify that the sequencer-relayer did not totally omit writing a rollup's data for a block

## Changes
- proposers now make a commitment to the `sequence::Action` `chain_id`s inside the block inside `prepare_proposal` (placed as the second tx in a block, directly after the current commitment to `sequence::Action`s)
- other validators verify this commitment in `process_proposal`

## Testing
unit tests

## Breaking Changelist
- proposed blocks must now contain the chain ID commitment as the second tx in the block

## Related Issues

part 1 of #308 (follow up will implement this verification in conductor)


crates/astria-sequencer/src/proposal/commitment.rs
```diff
@@ -22,7 +22,7 @@ use proto::native::sequencer::v1alpha1::SignedTransaction;
 /// This is somewhat arbitrary, but could be useful for proof of an action within the action tree.
 pub(crate) fn generate_sequence_actions_commitment(
     txs_bytes: Vec<Bytes>,
-) -> ([u8; 32], Vec<Bytes>) {
+) -> ([u8; 32], [u8; 32], Vec<Bytes>) {
```
joroshiba, line 25: It feels like this may have outgrown a tuple. It's unclear what the function of each set of u8 and bytes is here.

Can deduce from usages, but I think a struct return might be better for understanding from the function signature itself standpoint.

crates/astria-sequencer/src/proposal/commitment.rs
```diff
@@ -3,6 +3,24 @@ use std::collections::BTreeMap;
 use bytes::Bytes;
 use proto::native::sequencer::v1alpha1::SignedTransaction;
 
+/// Wrapper for values returned by [`generate_sequence_actions_commitment`].
+pub(crate) struct GenerateCommitmentsResult {
```
SuperFluffy, line 7: Maybe leave out the `Result` suffix to not involve memories of the Result type. Just `GeneratedCommitments` is fine

Also, do we need to make this a public interface at some point given that it's returned in the ABCI interface below?

SuperFluffy (approved):
Looks good to me. Should these methods ever be made public or do you want to keep them all implementation details?


Unrelated to this PR:

Still wondering about `chain_id` and if we should treat it as a human readable string or the hash of it.

Composer treats it as human readable. Treating it as a Vec of opaque bytes gives us neither human-readability, nor a guaranteed size on the stack (if we were to hash).

crates/astria-sequencer/src/proposal/commitment.rs
noot: don't need to make it public, it's not returned publicly, this struct is just used to populate the txs returned

dereksione (approved):
LGTM 🥇

PR #343, opened 2023-09-08
conductor: added `skip_empty_blocks` flag, updated executor to toggle empty block execution based on flag

### LINKED ISSUE #285: conductor: empty block execution setting
Currently the Conductor does nothing when it receives empty blocks.
More specifically, the `executor::Executor::execute_block()` returns `None` when it receives an empty block and never makes a call to `DoBlock`.

Is this the long term desired functionality?

Depending on the use case, having an option to either ignore or execute empty blocks may be useful.

sambukowski:
## Summary
Added the option to choose wether or not Conductor will send empty blocks to the execution layer.

## Background
Historically we were ignoring all empty blocks. It was determined that this is not something many users like.
Thus, adding in the option to execute empty block or not was required.

## Changes
- Added `skip_empty_blocks` flag. Defaults to false.
- Updated cli.rs and config.rs to add the flag
- Updated `execute_block()` to accommodate the flag
- Added new tests to Executor to verify the functionality difference between `skip_empty_blocks` settings

## Testing
Run `cargo test`

## Related Issues
closes #285

### DIFF
--- a/crates/astria-conductor/src/cli.rs
+++ b/crates/astria-conductor/src/cli.rs
@@ -39,4 +39,8 @@ pub struct Cli {
 
     #[arg(long = "disable-finalization")]
     pub disable_finalization: bool,
+
+    /// Flag to skip or execute empty blocks
+    #[arg(long = "skip-empty-blocks")]
+    pub skip_empty_blocks: bool,
 }
--- a/crates/astria-conductor/src/executor.rs
+++ b/crates/astria-conductor/src/executor.rs
@@ -497,7 +497,7 @@ mod test {
                 .get(&executor.execution_state)
                 .is_none()
         );
-        assert_eq!(previous_execution_state, executor.execution_state,);
+        assert_eq!(previous_execution_state, executor.execution_state);
         // should be empty because nothing was executed
         assert!(executor.sequencer_hash_to_execution_hash.is_empty());
     }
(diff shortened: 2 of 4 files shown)


crates/astria-conductor/src/cli.rs
```diff
@@ -39,4 +39,8 @@ pub struct Cli {
 
     #[arg(long = "disable-finalization")]
     pub disable_finalization: bool,
+
+    /// Flag to skip or execute empty blocks
+    #[arg(long = "skip-empty-blocks")]
+    pub skip_empty_blocks: bool,
```
noot, line 45: the default here is false i guess?

crates/astria-conductor/src/executor.rs
```diff
@@ -460,4 +501,39 @@ mod test {
...
+        // should be added to finalized blocks
+        assert!(!finalized_blocks.lock().await.is_empty());
+        assert!(
+            finalized_blocks
+                .lock()
+                .await
+                .get(&executor.execution_state)
+                .is_some()
+        );
+        assert_eq!(expected_execution_state, executor.execution_state,);
```
noot, line 535:
```suggestion
        assert_eq!(expected_execution_state, executor.execution_state);
```

noot (approved):
looks good!

crates/astria-conductor/src/cli.rs
sambukowski: yeah default is false.

sambukowski:
Closing this because #329 got merged and changed how config works.
The pr has been replaced by #373

PR #349, opened 2023-09-11
chore: fork penumbra storage, tower-trace

SuperFluffy:
## Summary
This patch forks penumbra-storage and penumbra-tower-trace to astria-storage and astria-tower-trace, respectively.

## Background
Penumbra's crates are written for penumbra at their own pace and primarily with their own ecosystem in mind. Contributing PRs to penumbra has potential ripple effects to their ecosystem of crates, needing larger changes. Since we want to update our dependencies at our pace (including tendermint), it is easier to fork these two limited crates and add our changes on top of them.

Astria is already using a fork of tendermint at https://github.com/astriaorg/tendermint-rs/tree/from-str-fix as a `[patch]` in `Cargo.toml` because penumbra is still on tendermint 0.32. Forking `penumbra-tower-trace` allows us to update tendermint to upstream independently.

Furthermore, `penumbra-storage` contains penumbra-specific strings which (as in their `metrics` module), which wouuld need to be renamed for astria purposes.

## Changes
- copied/forked `penumbra-{storage,tower-trace}` to `astria-{storage,tower-trace} 0.1.0` at tag `v0.59.0`, sha `a9ce1e4`
- removed all mention of `penumbra` from the crate
- put storage metrics behind a feature gate
- made all tests pass by lazily initializing a global tracing subscriber
- changed `astria-sequencer` to use `penumbra-storage`, `astria-tower-trace`
- changed uses of `nonconsensus_*` methods to `nonverifiable_*` methods (this was a breaking change in penumbra-storage 0.57)
- replace all shared dependencies by workspace dependencies
- remove all unused dependencies
- apply pedantic clippy to new crates
- vendor rocksdb in container builds to ensure the correct versions are built (thus no longer installing it from the debian bookworm repos)

## Testing
No actual code changes took place, all astria tests still pass

## Breaking Changelist
- astria-sequencer now depends on rust-rocksdb v0.21.0, which pulls in rocksdb 8.3.2. This might be a breaking change for persistent rocksdb storage (although as far as I know rocksdb guarantees compatibility)
- all other changes do not constitute observable breaking changes for astria services/binaries.


SuperFluffy:
Honestly, I don't think we should even to keep this synchronized.

If you look at their storage crate (the metrics module) this is seen as a deep dependency of their stack and not really to be used publicly.

noot:
@SuperFluffy I think it's worth messaging them about how we can keep these synced - are there any other breaking changes expected except for tendermint-rs updates? I can't think of any, so if that's the only change, they are going to make that update eventually. either way I think forking like this is fine for now

SuperFluffy:
@noot Talked to @joroshiba. Don't feel very convinced about this PR myself.

Penumbra is tracking the move to tendermint-0.33 in https://github.com/penumbra-zone/penumbra/pull/2961

erwanor:
Hey there! Is there a way we can convince you not to fork? We just merged our `tendermint-rs@0.33` integration, it includes a restructuring of `tower-trace` and `tower-actor` to make it easy and stable to support multiple versions (`v034` and `v037`) of cometbft side-by-side. This was done with y’all in mind (we use 34)!

I also added a “metrics” feature to `penumbra_storage`, it can be disabled by specifying `default-features = false` in your cargo files, see [#3007](https://github.com/penumbra-zone/penumbra/pull/3007).

We built the different parts of the penumbra stack to be generic building blocks to create blockchains in rust. The current design is the third iteration on this idea and is getting mature. So for us, supporting a more diverse set of use-cases is definitely a priority, and something we want to collaborate with you on!

I think the best way forward is to connect directly on discord ([invite here](https://discord.gg/deDRuCKMjZ)) so that we have direct channels with each other. All our processes are public, and our goal is to make it very legible what our roadmap is and what technical priorities drive this process.

SuperFluffy:
@erwanor yeah, I think we are actually not at all convinced that our fork is the right way to go about things.

I had a look at your open PRs after I created this when I saw your work on integrating tendermint 0.33. I should have done that first and synced with you. 😅

Thank you for giving a ping on it being merged. I'll integrate it ASAP!

erwanor:
> @erwanor yeah, I think we are actually not at all convinced that our fork is the right way to go about things.
> 
> I had a look at your open PRs after I created this when I saw your work on integrating tendermint 0.33. I should have done that first and synced with you. 😅
> 
> Thank you for giving a ping on it being merged. I'll integrate it ASAP!

Oh no worries, generally speaking we try to keep up with the tip of tendermint-rs. And if there are other things you need or want to chat about then really please do feel free to either open a ticket or ping me on discord, I think @noot is already there too.

PR #352, opened 2023-09-11
Conductor: add v1alpha2 rpc interface to conductor execution client

steezeburger:
## Summary
Non destructively adds v1alpha2 rpc functionality to Conductor's `ExecutionRpcClient`.

> Note that the v1alpha1 api will be removed in the PR for #207 

## Background
This PR was created in preparation of Conductor's switch to the new [Execution API v1alpha2.](https://buf.build/astria/astria/docs/main:astria.execution.v1alpha2)

## Changes
- adds v1alpha2 function definitions to `ExecutionClient` trait
- adds v1alpha2 function implementations to `ExecutionRpcClient` struct

## Testing
No tests yet as there is no implementation done here.

## Related Issues
https://github.com/astriaorg/astria/issues/205

closes https://github.com/astriaorg/astria/issues/205


SuperFluffy (commented):
I don't think the service should support both APIs at the same time using two different clients.

steezeburger:
> I don't think the service should support both APIs at the same time using two different clients.

It won't. I'm working on another branch that captures the work fr #207 that will merge into this one. It will remove the alpha1 api. This PR only capturing the work for #205 not #207

crates/astria-conductor/src/execution_client.rs
```diff
@@ -23,12 +37,37 @@ pub(crate) trait ExecutionClient: crate::private::Sealed {
     async fn call_finalize_block(&mut self, block_hash: Vec<u8>) -> Result<()>;
 
     async fn call_init_state(&mut self) -> Result<InitStateResponse>;
+
+    // v1alpha2
```
noot, line 41: let's separate these all into a different `ExecutionClientV1Alpha2` trait (and rename this existing trait)

sambukowski (approved):

noot (approved):
looks good!

PR #354, opened 2023-09-11
Conductor: implement v1alpha2 functionality in conductor

### LINKED ISSUE #207: conductor: update executor to utilize execution api v1alpha2

steezeburger:
## Summary
This PR refactors the conductor to use the functionality from the new execution rpc client v1alpha2.
It is dependent on the work done in #352.

## Background
v1alpha1 was not sufficient for our needs, so v1alpha2 was created

## Changes
- [x] refactor `Executor::new` to call `GetCommitmentState` rpc 
- [x] `ExecuteBlock` rpc to execute blocks for sequencer block fetched from DA
- [x] use `UpdateCommitmentState` rpc to set `FIRM` for sequencer block fetched from DA
- [x] call `ExecuteBlock` rpc for sequencer block received from gossipnet (this logic shouldn't need to change much for the websocket implementation that will be replacing gossipnet soon)
- [x] use `UpdateCommitmentState` rpc to set `SOFT` or `FIRM` for blocks received from gossipnet

## Testing
- Updated existing tests so they pass. Added test for scenario where an already executed block is rx from DA

## Breaking Changelist
- no more `do_block`. use `execute_block`
- no more `finalize_block`. use `update_commitment_state`

## Related Issues
- #205
- #207 

closes #207


crates/astria-conductor/src/executor.rs
```diff
@@ -115,11 +125,54 @@ struct Executor<C> {
...
+
+                // TODO - get sequencer hashes for these blocks. best way to do this?
+
+                // TODO - rebuild the hash that tracks sequencer hashes to execution hashes
+
+                // FIXME - what do i actually want to set the execution_state to in this case?
+                firm.hash
+            }
+            Ordering::Greater => {
+                // FIXME - what should we do in this case?
```
joroshiba, line 164: This is an invalid state, probably a panic?

crates/astria-conductor/src/executor.rs
```diff
@@ -115,11 +125,54 @@ struct Executor<C> {
...
+                debug!(
+                    "batch_get_blocks_response: {:#?}",
+                    batch_get_blocks_response
+                );
+
+                // TODO - get sequencer hashes for these blocks. best way to do this?
+
+                // TODO - rebuild the hash that tracks sequencer hashes to execution hashes
+
+                // FIXME - what do i actually want to set the execution_state to in this case?
```
joroshiba, line 160: We might want to update the execution state in general to be a commitment state? If that doesn't make sense, the soft commitment is what we are executing on top of moving forward.

crates/astria-conductor/src/executor.rs
```diff
@@ -226,21 +279,53 @@ impl<C: ExecutionClientV1Alpha1> Executor<C> {
...
+        &mut self,
+        block: SequencerBlockSubset,
+        execution_block_hash: Vec<u8>,
+    ) -> Result<()> {
+        let sequencer_block_height = block.header.height.value();
+        let sequencer_block_timestamp = convert_tendermint_to_prost_timestamp(block.header.time)
+            .wrap_err("failed parsing str as protobuf timestamp")?;
+        self.execution_rpc_client
+            .call_update_commitment_state(CommitmentState {
+                soft: None,
```
joroshiba, line 315: This isn't a patch, always need to set the full resource.

crates/astria-conductor/src/executor.rs
```diff
@@ -268,49 +355,21 @@ impl<C: ExecutionClientV1Alpha1> Executor<C> {
...
                 else {
                     // no txs for our namespace, nothing to do
-                    debug!("execute_block returned None; skipping finalize_block");
+                    debug!("execute_block returned None; skipping call_update_commitment_state");
                     return Ok(());
                 };
 
-                // finalize the block after it's been executed
-                self.finalize_block(execution_block_hash, sequencer_block_hash)
+                self.update_commitment_state_firm(block, execution_block_hash)
```
joroshiba, line 367: We need to update the full commitment here. Execution no longer updates the soft commitment, it's a separate action on a different resource.

crates/astria-conductor/src/executor.rs
```diff
@@ -226,21 +279,53 @@ impl<C: ExecutionClientV1Alpha1> Executor<C> {
...
+        Ok(Some(response.hash))
+    }
+
+    /// Updates the block's commitment state to firm
+    ///
+    /// # Arguments
+    ///
+    /// * `block` - the sequencer block that we know is firm
+    /// * `execution_block_hash` - the execution block hash corresponding to the sequencer block
+    async fn update_commitment_state_firm(
```
joroshiba, line 305: I think we should make this just `update_commitment_state` and have it take both the soft and firm commitment blocks as args?

There's no API to update one without the other anymore.

crates/astria-conductor/src/executor.rs
```diff
@@ -156,12 +162,28 @@ impl<C: ExecutionClientV1Alpha1> Executor<C> {
...
+                    match self.execute_block(block_subset.clone()).await {
+                        Ok(Some(execution_block_hash)) => {
+                            // after execution, set SOFT as this block. FIRM stays the same.
+                            let timestamp =
+                                convert_tendermint_to_prost_timestamp(block_subset.header.time)?;
+                            let firm = self.commitment_state.firm.clone().unwrap();
+                            let soft = Block {
+                                number: height as u32,
+                                hash: execution_block_hash.clone(),
+                                parent_block_hash: firm.hash.clone(),
```
sambukowski, line 174: I believe this should be `parent_block_hash: self.commitment_state.soft.unwrap().hash.clone()` or something like that. This blocks parent is the previous executed block, not necessarily the `firm` block.

crates/astria-conductor/src/executor.rs
```diff
@@ -219,7 +241,7 @@ impl<C: ExecutionClientV1Alpha1> Executor<C> {
             return Ok(Some(execution_hash.clone()));
         }
 
-        let prev_block_hash = self.execution_state.clone();
+        let prev_block_hash = self.commitment_state.firm.clone().unwrap().hash;
```
sambukowski, line 244: This should be the most recent soft block hash. `execute_block()` is primarily called to execute a block at the tip of the chain, so the most recent block will be the most recent soft block, but may happen to be firm.
In the case that the chain is finalize all the way to the tip, `firm` and `soft` will be the same, but I don't think that is that likely to happen very often.

crates/astria-conductor/src/executor.rs
```diff
@@ -259,8 +299,19 @@ impl<C: ExecutionClientV1Alpha1> Executor<C> {
...
-                self.finalize_block(execution_block_hash, sequencer_block_hash)
+                // this case means block has already been executed.
+                // setting execution chain's FIRM and SOFT as this block
+                let executed_block = Block {
+                    number: block.header.height.value() as u32,
+                    hash: execution_block_hash.clone(),
+                    parent_block_hash: self.commitment_state.firm.clone().unwrap().hash,
+                    timestamp: Some(convert_tendermint_to_prost_timestamp(block.header.time)?),
+                };
+                self.update_commitment_state(executed_block.clone(), executed_block)
```
sambukowski, line 310: The second `execute_block` should be replaced with `self.commitment_state.soft.unwrap().hash`.
It's likely that the height of the `soft` commit will be ahead of the `firm` commit.

crates/astria-conductor/src/executor.rs
```diff
@@ -273,49 +324,32 @@ impl<C: ExecutionClientV1Alpha1> Executor<C> {
...
                     return Ok(());
                 };
 
-                // finalize the block after it's been executed
-                self.finalize_block(execution_block_hash, sequencer_block_hash)
-                    .await?;
+                // after execution, set SOFT as this block. FIRM stays the same.
+                let timestamp = convert_tendermint_to_prost_timestamp(block.header.time)?;
+                let firm = self.commitment_state.firm.clone().unwrap();
+                let soft = Block {
```
sambukowski, line 339: I think the `firm` and `soft` commits here need to be reversed.
Updating the commit state to finalize a block we just executed should set `firm` to that block and `soft` should stay at whatever it was unless `firm` and `soft` become the same block.

crates/astria-conductor/src/executor.rs
```diff
@@ -400,7 +504,7 @@ mod test {
             .await
             .unwrap();
 
-        let expected_exection_hash = hash(&executor.execution_state);
+        let expected_execution_hash = hash(&executor.commitment_state.firm.clone().unwrap().hash);
```
sambukowski, line 507: I think this should be `soft` not `firm`.

sambukowski (approved):

crates/astria-conductor/src/executor.rs
```diff
@@ -231,21 +244,34 @@ impl<C: ExecutionClientV1Alpha1> Executor<C> {
...
-            execution_block_hash = hex::encode(&response.block_hash),
+            execution_block_hash = hex::encode(&response.hash),
             "executed sequencer block",
         );
         self.sequencer_hash_to_execution_hash
-            .insert(block.block_hash, response.block_hash.clone());
+            .insert(block.block_hash, response.hash.clone());
 
-        Ok(Some(response.block_hash))
+        Ok(Some(response.hash))
```
joroshiba, line 260: I feel like we should update `execute_block` to return the response (a block) here. Right now we are having to rebuild the block instead of just using the return type when updating the state commitment.

crates/astria-conductor/src/executor.rs
```diff
@@ -273,47 +301,56 @@ impl<C: ExecutionClientV1Alpha1> Executor<C> {
...
-            .remove(&sequencer_block_hash);
+            .remove(&block.block_hash);
+        Ok(())
+    }
+
+    /// Sets SOFT on execution chain to executed block.
+    async fn handle_executed_gossip_block(
+        &mut self,
+        block: SequencerBlockSubset,
+        execution_block_hash: Vec<u8>,
```
joroshiba, line 344: per above comment, I believe we could pass the block directly in here.

crates/astria-conductor/src/executor.rs
```diff
@@ -273,47 +301,56 @@ impl<C: ExecutionClientV1Alpha1> Executor<C> {
...
                     // no txs for our namespace, nothing to do
-                    debug!("execute_block returned None; skipping finalize_block");
+                    debug!("execute_block returned None; skipping call_update_commitment_state");
                     return Ok(());
                 };
 
-                // finalize the block after it's been executed
-                self.finalize_block(execution_block_hash, sequencer_block_hash)
-                    .await?;
+                self.handle_executed_da_block(block, execution_block_hash)
```
joroshiba, line 313: I don't think this case is handling updating the soft commitment, because in this case we executed & have firm commitment we need to update both state blocks.

joroshiba (changes requested):

crates/astria-conductor/src/executor.rs
steezeburger: I don't think that's correct.
See: https://github.com/astriaorg/astria/pull/354#discussion_r1327953266

crates/astria-conductor/src/executor.rs
```diff
@@ -272,50 +311,26 @@ impl<C: ExecutionClientV1Alpha1> Executor<C> {
...
+                    debug!("execute_block returned None; skipping call_update_commitment_state");
                     return Ok(());
                 };
-
-                // finalize the block after it's been executed
-                self.finalize_block(execution_block_hash, sequencer_block_hash)
-                    .await?;
+                self.update_commitment_state(
+                    executed_block,
+                    self.commitment_state.soft.clone().unwrap(),
```
joroshiba, line 325: I think we have the actions here flipped, if we just executed update both, otherwise only firm.

joroshiba (changes requested):

crates/astria-conductor/src/executor.rs
sambukowski: This statement should be swapped with the one on line 323. See comment below.

sambukowski (approved):

crates/astria-conductor/src/executor.rs
```diff
@@ -320,11 +323,9 @@ impl<C: ExecutionClientV1Alpha1 + ExecutionClientV1Alpha2> Executor<C> {
                     debug!("execute_block returned None; skipping call_update_commitment_state");
                     return Ok(());
                 };
-                self.update_commitment_state(
-                    executed_block,
-                    self.commitment_state.soft.clone().unwrap(),
-                )
-                .await?;
+                //
```
joroshiba, line 326: nit: fix comment

PR #355, opened 2023-09-12
proto: extract sequencer and sequencer-relayer types are protobuf

SuperFluffy:
## Summary
This patch defines protobuf versions of various types that are currently only defined in Rust and exchange as json-serialized text. 

## Background
`astria-proto` is intended to be the single source of truth for the wire format of data that is exchanged between services. At the moment `SequencerBlockData`, `SequencerNamespaceData`, `RollupNamespaceData` and `InclusionProof` are only defined in Rust. Some older versions of these exist as protobuf but were never used in the public interfaces.

## Changes
- Define a sequencer `v1alpha2` API for sequencer block, inclusion proof, and celestia blobs (called `CelestiaItem` and `CelestiaHeader`).

## Testing
- N/A; currently these are just protobuf definitions.

## Related Issues

https://github.com/astriaorg/astria/pull/12 originally defined sequencer blocks and transactions as protobuf, but these were never actually implemented.

https://github.com/astriaorg/astria/issues/86 is the reason this effort was started.

### DIFF
--- a/crates/astria-proto/proto/astria/sequencer/v1alpha2/block.proto
+++ b/crates/astria-proto/proto/astria/sequencer/v1alpha2/block.proto
@@ -1,0 +1,35 @@
+syntax = "proto3";
+
+package astria.sequencer.v1alpha2;
+
+import "tendermint/types/types.proto";
+import "astria/sequencer/v1alpha2/validation.proto";
+
+// `RollupTransactions` are a sequence of opaque bytes together with the human
+// readable `chain_id` of the rollup they belong to.
+//
+// The binary encoding is understood as an implementation detail of the
+// services sending and receiving the transactions.
+message RollupTransactions {
+    // The human readable name of the rollup these transactions belong to.
+    string chain_id = 1;
+    // The serialized opaque bytes of the rollup transactions.
+    repeated bytes transactions = 2;
+}
+
+// `SequencerBlock` is constructed from a tendermint/cometbft block by
+// converting its opaque `data` bytes into sequencer specific types.
+message SequencerBlock {
+    // The hash of the sequencer block. Must be 32 bytes.
+    bytes block_hash = 1;
+    // The original cometbft header that was the input to this sequencer block.
+    tendermint.types.Header header = 2;
+    // The commit/set of signatures that commited this block.
+    tendermint.types.Commit last_commit = 3;
+    // The collection of rollup transactions that were included in this block.
+    RollupTransactions rollup_transactions = 4;
+    // The root of the action tree of this block. Must be 32 bytes.
+    bytes action_tree_root = 5;
+    // The proof that the action tree root was included in `header.data_hash`.
+    astria.sequencer.v1alpha2.InclusionProof action_tree_inclusion_proof = 6;
+}
--- a/crates/astria-proto/proto/astria/sequencer/v1alpha2/celestia.proto
+++ b/crates/astria-proto/proto/astria/sequencer/v1alpha2/celestia.proto
@@ -1,0 +1,50 @@
+syntax = "proto3";
+
+package astria.sequencer.v1alpha2;
+
+import "tendermint/types/types.proto";
+import "astria/sequencer/v1alpha2/validation.proto";
+
+// An "item blob" of rollup transactions submitted to celestia.
+//
+// The transactions contained in the item belong to a rollup identified
+// by `chain_id`, and were included in the sequencer block identified
+// by `sequencer_block_hash`.
+message CelestiaItem {
+    // The hash of the sequencer block. Must be 32 bytes.
+    bytes sequencer_block_hash = 1;
+    // The human readable chain ID identifiying the rollup these transactions belong to.
+    string chain_id = 2;
+    // A list of opaque bytes that are serialized rollup transactions.
+    repeated bytes transactions = 3;
+    // The proof that the action tree root was included in
+    // `sequencer_block_header.data_hash` (to be found in `CelestiaHeader`).
+    // The inclusion of these transactions in the original sequencer block
+    // can be verified using the action tree root stored in `CelestiaHeader`.
+    astria.sequencer.v1alpha2.InclusionProof action_tree_inclusion_proof = 4;
+}
+
+// The "header blob" that is submitted to celestia under a sequencer-specific
+// celestia namespace.
+//
+// It is created by splitting up `SequencerBlockData` into a header
+// and a list of rollup blobs, where one blob contains a sequence of
+// rollup transactions for one specific rollup.
+//
+// The original sequencer block is identified by its `block_hash`.
+message CelestiaHeader {
+    // The hash of the sequencer block. Must be 32 bytes.
+    bytes sequencer_block_hash = 1;
+    // The original cometbft header that was the input to this sequencer block.
+    tendermint.types.Header sequencer_block_header = 2;
+    // The commit/set of signatures that commited this block.
+    tendermint.types.Commit sequencer_bock_last_commit = 3;
+    // The namespaces under which rollup transactions belonging to the sequencer
+    // block identified by `sequencer_block_hash` where submitted to celestia.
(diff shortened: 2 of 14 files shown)


crates/astria-proto/proto/astria/sequencer/v1alpha2/celestia.proto
```diff
@@ -0,0 +1,50 @@
...
+
+import "tendermint/types/types.proto";
+import "astria/sequencer/v1alpha2/validation.proto";
+
+// An "item blob" of rollup transactions submitted to celestia.
+//
+// The transactions contained in the item belong to a rollup identified
+// by `chain_id`, and were included in the sequencer block identified
+// by `sequencer_block_hash`.
+message CelestiaItem {
```
noot, line 13: does this correspond to the `RollupNamespaceData` type? let's make the names match up

crates/astria-proto/proto/astria/sequencer/v1alpha2/celestia.proto
```diff
@@ -0,0 +1,50 @@
...
+
+// The "header blob" that is submitted to celestia under a sequencer-specific
+// celestia namespace.
+//
+// It is created by splitting up `SequencerBlockData` into a header
+// and a list of rollup blobs, where one blob contains a sequence of
+// rollup transactions for one specific rollup.
+//
+// The original sequencer block is identified by its `block_hash`.
+message CelestiaHeader {
```
noot, line 35: same here for `SequencerNamespaceData`

crates/astria-proto/proto/astria/sequencer/v1alpha2/validation.proto
```diff
@@ -0,0 +1,14 @@
...
+package astria.sequencer.v1alpha2;
+
+// A merkle proof of inclusion of a leaf at `index` in a tree
+// of size `num_leaves`.
+message InclusionProof {
+    // leaf index of value to be proven
+    uint64 index = 1;
+    // total number of leaves in the tree
+    uint64 num_leaves = 2;
+    // the merkle proof itself; must be 32 bytes.
```
noot, line 12: merkle proof is probably longer than 32 bytes, should just be `bytes`

crates/astria-proto/proto/astria/sequencer/v1alpha2/celestia.proto
SuperFluffy: It does, but I want to move away from relatively generic sounding names like `RollupNamespaceData` and use celestia specific naming instead (because these are at the end of the day only ending up on celestia).

I am not a fan of `CelestiaItem`, but I would like something more specific than `RollupNamespaceData`

crates/astria-proto/proto/astria/sequencer/v1alpha2/celestia.proto
SuperFluffy: See comment above. I would like this to be better describe what is inside + make it clear that this is a celestia-specific thing. `SequencerNamespaceData` doesn't feel very "semantic" if that makes sense?

But yeah, happy to change naming.

crates/astria-proto/proto/astria/sequencer/v1alpha2/validation.proto
SuperFluffy: True, I misread the ctmerkle inclusion proof.

noot (changes requested):
is this PR just adding the new proto types? can you also delete the duplicate/unused protos inside astria/sequencer/v1alpha1/block.proto and data.proto? i don't think the protos in either of those files are used

SuperFluffy:
> is this PR just adding the new proto types? can you also delete the duplicate/unused protos inside astria/sequencer/v1alpha1/block.proto and data.proto? i don't think the protos in either of those files are used

It is, but I wanted to move this to a `v1alpha2` instead to show the evolution of the API. Not a big fan of just removing or changing protos. :-/

crates/astria-proto/proto/astria/sequencer/v1alpha2/celestia.proto
noot: true, in my head `(Something)NamespaceData` is celestia-specific since namespaces are for celestia, but I understand the reasoning. maybe `CelestiaRollupData` and `CelestiaSequencerData` or something?

emhane: replacing 'Celestia' with 'DataAvailability' in naming would adhere to the design of keeping the choice of data availability provider open to the user

crates/astria-proto/proto/astria/sequencer/v1alpha2/celestia.proto
SuperFluffy: Renamed

crates/astria-proto/proto/astria/sequencer/v1alpha2/celestia.proto
SuperFluffy: @noot I renamed the "items" to `CelestiaRollupData` and the head to `CelestiaSequencerData`.

@emhane Yes, the design vision is to let users choose their preferred rollup. Unfortunately we still need to define rollup-specific datastructures to interact with them. The goal of this PR is to make this clear.

noot:
@SuperFluffy I'm quite sure the protos currently in astria/sequencer/v1alpha1/block.proto and data.proto are not used anywhere at all, so you can remove them safely as it doesn't actually affect v1alpha1

crates/astria-proto/proto/astria/sequencer/v1alpha2/validation.proto
noot: i don't see it updated?

noot (approved):
can you merge main in? diff looks like it has unrelated changes. otherwise looks good!

crates/astria-proto/proto/astria/sequencer/v1alpha2/block.proto
```diff
@@ -0,0 +1,35 @@
...
+    // The hash of the sequencer block. Must be 32 bytes.
+    bytes block_hash = 1;
+    // The original cometbft header that was the input to this sequencer block.
+    tendermint.types.Header header = 2;
+    // The commit/set of signatures that commited this block.
+    tendermint.types.Commit last_commit = 3;
+    // The collection of rollup transactions that were included in this block.
+    RollupTransactions rollup_transactions = 4;
+    // The root of the action tree of this block.
+    bytes action_tree_root = 5;
```
noot, line 32: you can re-add the "must be 32 bytes" here

crates/astria-proto/proto/astria/sequencer/v1alpha2/celestia.proto
```diff
@@ -0,0 +1,48 @@
...
+// (this `CelestiaSequencerData`), and a list of `CelestiaRollupData` items.
+//
+// The original sequencer block is identified by its `block_hash`.
+message CelestiaSequencerData {
+    // The hash of the sequencer block. Must be 32 bytes.
+    bytes sequencer_block_hash = 1;
+    // The original cometbft header that was the input to this sequencer block.
+    tendermint.types.Header sequencer_block_header = 2;
+    // The commit/set of signatures that commited this block.
+    tendermint.types.Commit sequencer_bock_last_commit = 3;
```
noot, line 39:
```suggestion
    tendermint.types.Commit sequencer_block_last_commit = 3;
```

crates/astria-proto/proto/astria/sequencer/v1alpha2/validation.proto
SuperFluffy: I have removed `32 bytes` and added that the merkle proof is derived from an RFC 6962 compliant merkle tree.

SuperFluffy:
Something broke github. After squashing and rebasing on top of recent main this PR is no longer in sync with its branch. Opening a new PR.

Issue #227, opened 2023-08-07
sequencer: investigate and fix `init_chain`


noot:
currently, tendermint/cometbft is not calling `init_chain` as the first ABCI call as expected, so right now we're calling it manually when the sequencer is started. need to investigate why cometbft isn't calling it and fix it so that it is.

also `init_chain` should return the app hash to cometbft.

PR #359, opened 2023-09-12
feat(sequencer): implement basic validator set updates

### LINKED ISSUE #68: sequencer: basic PoA implementation
the sequencer node should have a basic PoA `component` type that is able to rotate the validator set. note that cometbft already has the ability to have any amount of validators at genesis, so this component would deal with changing the validator set.

open question:
- when/how do we change the validator set? is it a multisig of the existing validators, or is there a "sudo" key that is allowed to authorize validator set changes?
- do we change the validator set at the next block (or some other defined block height) once a change has been proposed and accepted? or do we implement an epoch-based system where all changes happen at a fixed height?

noot:
## Summary
 implement basic validator set updates using a sudo key set at genesis 

for a follow up:
- allow this sudo key to be updated as well
- #227

## Background
we need a way to update validators for devnet

## Changes
- implement `authority` module, contains one action `validator::Update` (from the tendermint type) which must be sent by the sudo key. set the `power` to 0 to remove the validator.
- module also contains `AuthorityComponent` which sets the sudo key and validator set at genesis
- validator set updates are stored in the non-consensus storage while the block txs are executing, then the batch of validator updates are returned during `end_block`

## Testing
unit tests (todo)
ran the sequencer node (todo)

## Breaking Changelist
- sequencer genesis file must contain `authority_sudo_key`

## Related Issues

closes #68

### DIFF
--- a/crates/astria-sequencer/src/app.rs
+++ b/crates/astria-sequencer/src/app.rs
@@ -24,6 +24,7 @@ use tracing::{
 
 use crate::{
     accounts::component::AccountsComponent,
+    authority::component::AuthorityComponent,
     component::Component,
     genesis::GenesisState,
     state_ext::{
@@ -76,7 +77,11 @@ impl App {
     }
 
     #[instrument(name = "App:init_chain", skip(self))]
-    pub(crate) async fn init_chain(&mut self, genesis_state: GenesisState) -> Result<()> {
+    pub(crate) async fn init_chain(
+        &mut self,
+        genesis_state: GenesisState,
+        genesis_validators: Vec<tendermint::validator::Update>,
+    ) -> Result<()> {
         let mut state_tx = self
             .state
             .try_begin_transaction()
@@ -86,9 +91,8 @@ impl App {
 
         // call init_chain on all components
         AccountsComponent::init_chain(&mut state_tx, &genesis_state).await?;
+        AuthorityComponent::init_chain(&mut state_tx, &(genesis_state, genesis_validators)).await?;
         state_tx.apply();
-
-        // TODO: call commit and return the app hash?
         Ok(())
     }
 
@@ -342,8 +346,9 @@ mod test {
         let mut app = App::new(snapshot);
         let genesis_state = GenesisState {
             accounts: default_genesis_accounts(),
+            authority_sudo_key: Address::from([0; 20]),
         };
-        app.init_chain(genesis_state).await.unwrap();
+        app.init_chain(genesis_state, vec![]).await.unwrap();
         assert_eq!(app.state.get_block_height().await.unwrap(), 0);
         for Account {
             address,
@@ -366,8 +371,9 @@ mod test {
         let mut app = App::new(snapshot);
         let genesis_state = GenesisState {
             accounts: vec![],
+            authority_sudo_key: Address::from([0; 20]),
         };
-        app.init_chain(genesis_state).await.unwrap();
+        app.init_chain(genesis_state, vec![]).await.unwrap();
 
         let mut begin_block = abci::request::BeginBlock {
             header: default_header(),
@@ -397,8 +403,9 @@ mod test {
         let mut app = App::new(snapshot);
         let genesis_state = GenesisState {
             accounts: default_genesis_accounts(),
+            authority_sudo_key: Address::from([0; 20]),
         };
-        app.init_chain(genesis_state).await.unwrap();
+        app.init_chain(genesis_state, vec![]).await.unwrap();
 
         // transfer funds from Alice to Bob
         // this secret key corresponds to ALICE_ADDRESS
@@ -449,8 +456,9 @@ mod test {
         let mut app = App::new(snapshot);
         let genesis_state = GenesisState {
             accounts: default_genesis_accounts(),
+            authority_sudo_key: Address::from([0; 20]),
         };
-        app.init_chain(genesis_state).await.unwrap();
+        app.init_chain(genesis_state, vec![]).await.unwrap();
 
         // create a new key; will have 0 balance
         let keypair = SigningKey::new(OsRng);
@@ -487,8 +495,9 @@ mod test {
         let mut app = App::new(snapshot);
         let genesis_state = GenesisState {
             accounts: default_genesis_accounts(),
+            authority_sudo_key: Address::from([0; 20]),
         };
-        app.init_chain(genesis_state).await.unwrap();
+        app.init_chain(genesis_state, vec![]).await.unwrap();
 
         // this secret key corresponds to ALICE_ADDRESS
         let alice_secret_bytes: [u8; 32] =
@@ -534,9 +543,10 @@ mod test {
         let mut app = App::new(snapshot);
         let genesis_state = GenesisState {
             accounts: default_genesis_accounts(),
+            authority_sudo_key: Address::from([0; 20]),
         };
 
-        app.init_chain(genesis_state).await.unwrap();
+        app.init_chain(genesis_state, vec![]).await.unwrap();
         assert_eq!(app.state.get_block_height().await.unwrap(), 0);
         for Account {
             address,
--- a/crates/astria-sequencer/src/app.rs
+++ b/crates/astria-sequencer/src/app.rs
@@ -24,7 +24,10 @@ use tracing::{
 
 use crate::{
     accounts::component::AccountsComponent,
-    authority::component::AuthorityComponent,
+    authority::{
+        component::AuthorityComponent,
+        state_ext::StateReadExt as _,
+    },
     component::Component,
     genesis::GenesisState,
     state_ext::{
@@ -180,15 +183,29 @@ impl App {
     pub(crate) async fn end_block(
         &mut self,
         end_block: &abci::request::EndBlock,
-    ) -> Vec<abci::Event> {
+    ) -> abci::response::EndBlock {
         let state_tx = StateDelta::new(self.state.clone());
         let mut arc_state_tx = Arc::new(state_tx);
 
         // call end_block on all components
         AccountsComponent::end_block(&mut arc_state_tx, end_block).await;
+        AuthorityComponent::end_block(&mut arc_state_tx, end_block).await;
+
+        // gather and return validator updates
+        let validator_updates = self
+            .state
+            .get_validator_updates()
+            .await
+            .expect("failed getting validator updates");
+
         let state_tx = Arc::try_unwrap(arc_state_tx)
             .expect("components should not retain copies of shared state");
-        self.apply(state_tx)
+        let events = self.apply(state_tx);
+        abci::response::EndBlock {
+            validator_updates: validator_updates.0,
+            events,
+            ..Default::default()
+        }
     }
 
     #[instrument(name = "App:commit", skip(self))]
--- a/crates/astria-sequencer/src/app.rs
+++ b/crates/astria-sequencer/src/app.rs
@@ -310,7 +310,6 @@ mod test {
         Address::from_array(arr)
     }
 
-    // generated with test `generate_default_keys()`
     const ALICE_ADDRESS: &str = "1c0c490f1b5528d8173c5de46d131160e4b2c0c3";
     const BOB_ADDRESS: &str = "34fec43c7fcab9aef3b3cf8aba855e41ee69ca3a";
     const CAROL_ADDRESS: &str = "60709e2d391864b732b4f0f51e387abb76743871";
--- a/crates/astria-sequencer/src/authority/action.rs
+++ b/crates/astria-sequencer/src/authority/action.rs
@@ -1,12 +1,17 @@
-use anyhow::Result;
+use anyhow::{
+    ensure,
+    Context as _,
+    Result,
+};
 use ed25519_consensus::VerificationKey;
 use proto::native::sequencer::v1alpha1::Address;
 use tracing::instrument;
 
 use crate::{
-    accounts::state_ext::{
+    authority::state_ext::{
         StateReadExt,
         StateWriteExt,
+        Validator,
     },
     transaction::action_handler::ActionHandler,
 };
@@ -25,14 +30,36 @@ impl ActionHandler for ValidatorUpdate {
         from: Address,
     ) -> Result<()> {
         // ensure signer is the valid `sudo` key in state
+        let sudo_address = state.get_sudo_address().await?;
+        ensure!(sudo_address == from, "signer is not the sudo key");
+
         // ensure validator to be updated is in the set
+        let validator_set = state.get_validator_set().await?;
+        ensure!(
+            validator_set
+                .0
+                .iter()
+                .any(|v| v.public_key == self.public_key.to_bytes()),
+            "validator to be updated is not in the set"
+        );
         Ok(())
     }
 
     #[instrument(skip_all)]
-    async fn execute<S: StateWriteExt>(&self, state: &mut S, from: Address) -> Result<()> {
+    async fn execute<S: StateWriteExt>(&self, state: &mut S, _: Address) -> Result<()> {
         // add validator update in non-consensus state
         // to be used in end_block
+        let mut validator_updates = state
+            .get_validator_updates()
+            .await
+            .context("failed getting validator updates")?;
+        validator_updates.0.push(Validator {
+            public_key: self.public_key.to_bytes(),
+            voting_power: self.voting_power,
+        });
+
+        state.put_validator_updates(validator_updates)?;
+
         Ok(())
     }
 }
--- a/crates/astria-sequencer/src/authority/action.rs
+++ b/crates/astria-sequencer/src/authority/action.rs
@@ -3,7 +3,6 @@ use anyhow::{
     Context as _,
     Result,
 };
-use ed25519_consensus::VerificationKey;
 use proto::native::sequencer::v1alpha1::Address;
 use tracing::instrument;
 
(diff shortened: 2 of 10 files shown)


crates/astria-sequencer/src/authority/component.rs
```diff
@@ -0,0 +1,75 @@
...
+    type AppState = (GenesisState, Vec<validator::Update>);
+
+    #[instrument(name = "AuthorityComponent:init_chain", skip(state))]
+    async fn init_chain<S: StateWriteExt>(mut state: S, app_state: &Self::AppState) -> Result<()> {
+        // set sudo key and initial validator set
+        state
+            .put_sudo_address(app_state.0.authority_sudo_key)
+            .context("failed to set sudo key")?;
+        state
+            .put_validator_set(ValidatorSet(app_state.1.clone()))
```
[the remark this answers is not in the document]
sambukowski: I don't think it has to be? `AppState` is defined in this crate and the type is specified in this impl, so probably not hard to make it a struct instead of a tuple.

sambukowski (approved):
nit: I think it would be worth addressing @joroshiba's comments on the `.0`s and `.1`s but otherwise looks good!

crates/astria-proto/src/native/sequencer/v1alpha1.rs
```diff
@@ -323,6 +325,7 @@ impl Action {
         let kind = match self {
             Action::Sequence(act) => Value::SequenceAction(act.to_raw()),
             Action::Transfer(act) => Value::TransferAction(act.to_raw()),
+            Action::ValidatorUpdate(act) => Value::ValidatorUpdateAction(act.clone().into()),
```
SuperFluffy, line 328: I believe this clone can be avoided, it looks like `tendermint::validator::Update` should impl `Copy` because it's fields are both Copy, but they forgot to add that.

crates/astria-sequencer/src/app.rs
```diff
@@ -85,6 +99,14 @@ impl App {
...
         // call init_chain on all components
         AccountsComponent::init_chain(&mut state_tx, &genesis_state).await?;
+        AuthorityComponent::init_chain(
+            &mut state_tx,
+            &AuthorityComponentAppState {
+                authority_sudo_key: genesis_state.authority_sudo_key,
+                genesis_validators,
+            },
+        )
+        .await?;
```
SuperFluffy, line 109: `context`

crates/astria-sequencer/src/app.rs
```diff
@@ -85,6 +99,14 @@ impl App {
 
         // call init_chain on all components
         AccountsComponent::init_chain(&mut state_tx, &genesis_state).await?;
```
SuperFluffy, line 101: (context here as well, although it wasn't there before)

crates/astria-sequencer/src/app.rs
```diff
@@ -174,15 +184,33 @@ impl App {
...
+        let mut state_tx = Arc::try_unwrap(arc_state_tx)
             .expect("components should not retain copies of shared state");
-        self.apply(state_tx)
+
+        // clear validator updates
+        state_tx.clear_validator_updates();
+
+        let events = self.apply(state_tx);
+        abci::response::EndBlock {
+            validator_updates: validator_updates.0,
```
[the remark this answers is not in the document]
SuperFluffy: Can we avoid making the inner vec `pub(crate)` and instead provide a `ValidatorSet::into_inner`?

crates/astria-sequencer/src/app.rs
```diff
@@ -523,6 +556,120 @@ mod test {
...
 
+    #[tokio::test]
+    async fn app_deliver_tx_validator_update() {
+        let alice_secret_bytes: [u8; 32] =
+            hex::decode("2bd806c97f0e00af1a1fc3328fa763a9269723c8db8fac4f93af71db186d6e90")
+                .unwrap()
+                .try_into()
+                .unwrap();
+        let alice_signing_key = SigningKey::from(alice_secret_bytes);
+        let alice = Address::from_verification_key(alice_signing_key.verification_key());
```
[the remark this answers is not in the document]
noot: agreed, done

crates/astria-proto/src/native/sequencer/v1alpha1.rs
noot: hm, should i upstream that to tendermint-rs then?

sambukowski (approved):

crates/astria-sequencer/src/authority/action.rs
```diff
@@ -0,0 +1,41 @@
...
+
+#[async_trait::async_trait]
+impl ActionHandler for tendermint::validator::Update {
+    async fn check_stateful<S: StateReadExt + 'static>(
+        &self,
+        state: &S,
+        from: Address,
+    ) -> Result<()> {
+        // ensure signer is the valid `sudo` key in state
+        let sudo_address = state.get_sudo_address().await?;
```
SuperFluffy, line 25: context

crates/astria-sequencer/src/authority/action.rs
```diff
@@ -0,0 +1,41 @@
...
+
+    #[instrument(skip_all)]
+    async fn execute<S: StateWriteExt>(&self, state: &mut S, _: Address) -> Result<()> {
+        // add validator update in non-consensus state to be used in end_block
+        let mut validator_updates = state
+            .get_validator_updates()
+            .await
+            .context("failed getting validator updates")?;
+        validator_updates.0.push(self.clone());
+        state.put_validator_updates(validator_updates)?;
```
SuperFluffy, line 38: context

crates/astria-sequencer/src/authority/component.rs
```diff
@@ -0,0 +1,79 @@
...
+    #[instrument(name = "AuthorityComponent:end_block", skip(state))]
+    async fn end_block<S: StateWriteExt + StateReadExt + 'static>(
+        state: &mut Arc<S>,
+        _end_block: &EndBlock,
+    ) {
+        // update validator set
+        let validator_updates = state
+            .get_validator_updates()
+            .await
+            .expect("failed getting validator updates");
```
SuperFluffy, line 65: Should this really crash the entire app?

crates/astria-sequencer/src/authority/component.rs
```diff
@@ -0,0 +1,79 @@
...
+
+        let mut current_set = state
+            .get_validator_set()
+            .await
+            .expect("failed getting validator set");
```
SuperFluffy, line 70: crash?

crates/astria-sequencer/src/authority/component.rs
```diff
@@ -0,0 +1,79 @@
...
+        current_set.apply_updates(validator_updates);
+
+        // this is safe because we are the only ones with a reference to the state
+        let state = Arc::get_mut(state).unwrap();
+        state
+            .put_validator_set(current_set)
+            .expect("failed putting validator set");
```
SuperFluffy, line 77: crash?

crates/astria-sequencer/src/authority/component.rs
```diff
@@ -0,0 +1,79 @@
...
+            .expect("failed getting validator updates");
+
+        let mut current_set = state
+            .get_validator_set()
+            .await
+            .expect("failed getting validator set");
+        current_set.apply_updates(validator_updates);
+
+        // this is safe because we are the only ones with a reference to the state
+        let state = Arc::get_mut(state).unwrap();
```
SuperFluffy, line 74: I prefer to just put that into the `expect` message so you also have it in the logs:
```suggestion
        let state = Arc::get_mut(state).expect("this must be the only reference to the state; this is a bug");
```
Because rust has `unsafe` blocks I would try to avoid talking about `safe` in other context (panics are "safe" in that they don'y lead to undefined behavior).

crates/astria-sequencer/src/authority/state_ext.rs
```diff
@@ -0,0 +1,155 @@
...
+pub(crate) struct ValidatorSet(pub(crate) Vec<tendermint::validator::Update>);
+
+impl ValidatorSet {
+    pub(crate) fn apply_updates(&mut self, validator_updates: ValidatorSet) {
+        for update in validator_updates.0 {
+            if self
+                .0
+                .iter()
+                .map(|val| val.pub_key)
+                .collect::<Vec<tendermint::public_key::PublicKey>>()
```
SuperFluffy, line 44: If I read this code correctly, this allocation is not necessary, just take the inner `find` and apply it directly, no need for the outer iter + map + collect + contains:
```rust
if let Some(validator) = self.0.iter_mut().find(|validator| validator.pub_key == update.pub_key) {
    validator.power = update.power;
} else {
    self.0.push(update);
}
```

Although I feel like using a hashmap here is probably more desirable (see the comment on the type def above)

crates/astria-sequencer/src/authority/state_ext.rs
```diff
@@ -0,0 +1,155 @@
...
+
+impl<T: StateRead> StateReadExt for T {}
+
+#[async_trait]
+pub(crate) trait StateWriteExt: StateWrite {
+    #[instrument(skip(self))]
+    fn put_sudo_address(&mut self, address: Address) -> Result<()> {
+        self.put_raw(
+            SUDO_STORAGE_KEY.to_string(),
+            SudoAddress(address.0).try_to_vec()?,
```
SuperFluffy, line 126: context

crates/astria-sequencer/src/authority/state_ext.rs
```diff
@@ -0,0 +1,155 @@
...
+        );
+        Ok(())
+    }
+
+    #[instrument(skip(self))]
+    fn put_validator_set(&mut self, validator_set: ValidatorSet) -> Result<()> {
+        self.put_raw(
+            VALIDATOR_SET_STORAGE_KEY.to_string(),
+            serde_json::to_vec(&validator_set)?,
```
SuperFluffy, line 135: context

crates/astria-sequencer/src/authority/state_ext.rs
```diff
@@ -0,0 +1,155 @@
...
+        );
+        Ok(())
+    }
+
+    #[instrument(skip(self))]
+    fn put_validator_updates(&mut self, validator_updates: ValidatorSet) -> Result<()> {
+        self.nonconsensus_put_raw(
+            VALIDATOR_UPDATES_KEY.to_vec(),
+            serde_json::to_vec(&validator_updates)?,
```
SuperFluffy, line 144: context

SuperFluffy (changes requested):
Requesting changes because I saw a lot of `expect` in places that seemed like an oversight.

Also some suggestions on the definition of `ValidatorSet`. I think it's worth thinking about modules as little crates with a strict pub-priv separation (just in this case `pub(crate)` vs not-`pub(crate)`). I feel like many instances of `.0` access to the inner vec would be better served by just defining inherent methods like `ValidatorSet::len` that delegate to the inner `self.0.len())`.

crates/astria-sequencer/src/authority/component.rs
noot: unfortunately, yeah it should. any error during block execution like this will result in the execution ending early, and will cause a resulting different app state which won't match other nodes that did not error, causing a consensus mismatch. there isn't really any way to recover from something like this

crates/astria-sequencer/src/authority/state_ext.rs
```diff
@@ -0,0 +1,155 @@
...
+            }
+        }
+
+        // delete validators with power 0
+        self.0.retain(|v| v.power.value() != 0);
+    }
+}
+
+const SUDO_STORAGE_KEY: &str = "sudo";
+const VALIDATOR_SET_STORAGE_KEY: &str = "valset";
```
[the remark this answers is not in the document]
noot: not really, just saving a small amount of bytes on disk

crates/astria-sequencer/src/authority/state_ext.rs
```diff
@@ -0,0 +1,155 @@
...
+#[derive(BorshSerialize, BorshDeserialize, Debug)]
+struct SudoAddress([u8; ADDRESS_LEN]);
+
+/// Newtype wrapper to read and write a validator set from rocksdb.
+///
+/// Note: this is stored only in the nonconsensus state, thus is not part
+/// of the application state, so it's fine for it to be serialized/deserialized
+/// in a non-deterministic way (ie. with serde-json).
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+pub(crate) struct ValidatorSet(pub(crate) Vec<tendermint::validator::Update>);
```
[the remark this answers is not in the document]
noot: updated to use a `HashMap` and also have the inner field private, however I now have:
```
    #[allow(dead_code)]
    pub(crate) fn len(&self) -> usize {
        self.0.len()
    }

    #[allow(dead_code)]
    pub(crate) fn get(&self, pub_key: &[u8]) -> Option<&tendermint::validator::Update> {
        self.0.get(pub_key)
    }
```

these functions are used in tests, but nowhere in the non-test code :/ not sure how to reconcile this without making the inner field crate public

sambukowski (approved):

steezeburger (approved):

crates/astria-sequencer/src/authority/component.rs
SuperFluffy: I guess what I am asking is why this warrants a crash instead of bubbling up the error just taking down tower-abci instead of doing it here with a panic? This would require changing `App::end_block` to return a `Result`. It's actually surprising that some of `App`'s methods are returning errors, while others are outright panicking.

crates/astria-sequencer/src/authority/state_ext.rs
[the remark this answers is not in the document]
SuperFluffy: Can we use the full name? 🙏

crates/astria-sequencer/src/authority/state_ext.rs
```diff
@@ -0,0 +1,175 @@
...
+impl ValidatorSet {
+    pub(crate) fn new_from_updates(updates: Vec<tendermint::validator::Update>) -> Self {
+        let mut validator_set = HashMap::new();
+        for update in updates {
+            validator_set.insert(update.pub_key.to_hex(), update);
+        }
+        Self(validator_set)
+    }
+
+    #[allow(dead_code)]
```
SuperFluffy, line 46: For private functions it's actually ok to do:
```rust
#[cfg(test)]
pub(crate) fn len(&self) -> usize {
```
It's only problematic in integration tests

crates/astria-sequencer/src/authority/state_ext.rs
```diff
@@ -0,0 +1,175 @@
...
+    pub(crate) fn len(&self) -> usize {
+        self.0.len()
+    }
+
+    #[allow(dead_code)]
```
SuperFluffy, line 51: Same with this

crates/astria-sequencer/src/authority/state_ext.rs
```diff
@@ -0,0 +1,175 @@
...
+///
+/// Contains a map of hex-encoded public keys to validator updates.
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+pub(crate) struct ValidatorSet(HashMap<String, tendermint::validator::Update>);
+
+impl ValidatorSet {
+    pub(crate) fn new_from_updates(updates: Vec<tendermint::validator::Update>) -> Self {
+        let mut validator_set = HashMap::new();
+        for update in updates {
+            validator_set.insert(update.pub_key.to_hex(), update);
```
SuperFluffy, line 41: You can express this as:
```rust
let validator_set = updates
    .into_iter()
    .map(|update| (update.pub_key.to_hex(), update))
    .collect();
```

crates/astria-sequencer/src/genesis.rs
```diff
@@ -5,9 +5,20 @@ use serde::{
 };
 
 /// The genesis state for the application.
-#[derive(Debug, Deserialize, Default)]
+#[derive(Debug, Deserialize)]
 pub(crate) struct GenesisState {
     pub(crate) accounts: Vec<Account>,
+    #[serde(deserialize_with = "deserialize_address")]
+    pub(crate) authority_sudo_key: Address,
```
SuperFluffy, line 12: Does a 0 address make sense as a default? A hash of `0`s is valid, so an address of `0`s is valid.

crates/astria-sequencer/src/genesis.rs
```diff
@@ -5,9 +5,20 @@ use serde::{
...
+}
+
+impl Default for GenesisState {
```
SuperFluffy, line 15: IMO this should be removed. Is this only for tests?

crates/astria-sequencer/src/genesis.rs
SuperFluffy: Alternatively, `Option<Address>`?

SuperFluffy (approved):
Looks good. Just a few comments.

For the open question:


This looks like the kind of thing that you want to have feature gated indeed.

IMO what does make sense is to have 2 versions of the authority module: the feature gated one that impls the full functionality (which very noisily informs us that it's active). And then the standard one so that the code in `transaction` and other parts of the code is not affected, but that only does no-ops.

crates/astria-sequencer/src/authority/component.rs
noot: @SuperFluffy updated `end_block` to return a `Result`, so the error now bubbles up and is returned to tower-abci

crates/astria-sequencer/src/authority/state_ext.rs
noot: oh wow, i did not know this, thank you!!

noot:
@SuperFluffy yeah, I think this eventually needs to be feature-gated, not sure if it's necessary now but will definitely be needed once we have a staking component (as only one or another can be enabled at once). what do you see the benefit of the no-op module being?

PR #364, opened 2023-09-13
sequencer: return error code if submitted transaction contained invalid nonce

itamarreif:
## Summary
The sequencer does not currently return error responses for any of the ABCI Consensus methods failures. This adds a quick though not ideal response for invalid nonces.

## Background
Composer requires an error response to identify when transaction submission failed due to invalid nonce so that it can refetch the correct nonce and resubmit the transaction.

## Changes
- The consensus service downcasts the error returned by `app.deliver_tx()` to check if it failed due to invalid nonce, returning the appropriate error response
- Added an ABCI code for invalid nonce
- The `ActionHandler` for `UnsignedTransaction` returns an `InvalidNonce` object with the nonce that was invalid upon error instead of a string.

## Testing
- Regression test for `app.deliver_tx()` to ensure that invalid nonce errors are captured.

## Related Issues
- Required for #293


crates/astria-sequencer/src/transaction/mod.rs
```diff
@@ -42,6 +44,19 @@ pub(crate) async fn execute<S: StateWriteExt>(
         .await
 }
 
+#[derive(Debug)]
+pub(crate) struct InvalidNonce(pub(crate) u32);
```
SuperFluffy, line 48: Add an `impl std::error::Error for InvalidNonce` here. No need to impl any of its methods, just use the default ones.

crates/astria-sequencer/src/transaction/mod.rs
```diff
@@ -42,6 +44,19 @@ pub(crate) async fn execute<S: StateWriteExt>(
...
+
+impl Display for InvalidNonce {
```
SuperFluffy, line 50: Import `std::fmt` above, and then refer to all the types in here as `fmt::{Display, Formatter, Result}` for consistency.

crates/astria-sequencer/src/service/consensus.rs
```diff
@@ -160,20 +159,28 @@ impl Consensus {
...
+                let code = if let Some(_e) = e.downcast_ref::<InvalidNonce>() {
+                    tracing::warn!("{}", e);
+                    AbciCode::INVALID_NONCE
+                } else {
+                    tracing::warn!(error = ?e, "deliver_tx failed");
+                    AbciCode::INTERNAL_ERROR
+                };
+                response::DeliverTx {
+                    code: code.into(),
+                    info: code.info().unwrap_or("invalid ABCI code").to_string(),
```
SuperFluffy, line 178: Use it's Display impl here: `code.to_string()`.

SuperFluffy (approved):
Looks great, just a few comments to address before merge.

crates/astria-sequencer/src/app.rs
```diff
@@ -523,6 +523,70 @@ mod test {
         );
     }
 
+    #[tokio::test]
+    async fn app_deliver_tx_invalid_nonce() {
```
SuperFluffy, line 527: It's unfortunate that the test for the thing you want to test is so far away from it.

If we define request handlers as freestanding functions rather than inherent methods on types we can test the handles independently and inject the necessary state as needed.

That's how axum is designed

joroshiba: Can we crate an issue around what we would need to do to fix this?

crates/astria-sequencer/src/service/mod.rs
```diff
@@ -18,6 +18,7 @@ impl AbciCode {
     pub(crate) const UNKNOWN_PATH: Self = Self(1);
     pub(crate) const INVALID_PARAMETER: Self = Self(2);
     pub(crate) const INTERNAL_ERROR: Self = Self(3);
+    pub(crate) const INVALID_NONCE: Self = Self(4);
```
joroshiba, line 21: Should these be defined outside the sequencer crate for use in other binaries (ie composer presumably needs to know what the code means?)

joroshiba (approved):
We should have a high level issue for going through consensus methods to make sure we return errors and get the error codes setup.

crates/astria-sequencer/src/transaction/mod.rs
```diff
@@ -68,11 +85,9 @@ impl ActionHandler for UnsignedTransaction {
         // Nonce should be equal to the number of executed transactions before this tx.
         // First tx has nonce 0.
         let curr_nonce = state.get_account_nonce(from).await?;
-        ensure!(
-            curr_nonce == self.nonce,
-            "invalid nonce, tx nonce must match account nonce"
-        );
+        ensure!(curr_nonce == self.nonce, InvalidNonce(self.nonce));
 
+        // do we need to make a StateDelta here so we can check the actions on the successive state?
```
noot, line 90: remove line

crates/astria-sequencer/src/app.rs
noot: you should be able to write a test on `UnsignedTransaction::check_stateful` directly instead of here, that would be clearer imo.

crates/astria-sequencer/src/app.rs
```diff
@@ -523,6 +523,70 @@ mod test {
...
+            ],
+        };
+
+        // sign tx
+        let signed_tx = tx.into_signed(&alice_signing_key);
+        let bytes = signed_tx.into_raw().encode_to_vec();
+
+        // Act
+        // pass tx to deliver_tx
+        let response = app.deliver_tx(&bytes).await;
```
noot, line 570: if you want to keep this unit test, i'd move this code into its own function (as it's essentially the same as the other tests) that takes an `UnsignedTransaction` and returns the response

crates/astria-sequencer/src/app.rs
SuperFluffy: @noot the test needs to be written against `App::deliver_tx` because it needs to protect against adding extra error layers, silently breaking the downcast_ref never activating.

crates/astria-sequencer/src/service/mod.rs
itamarreif: see #369

crates/astria-sequencer/src/transaction/mod.rs
emhane: use the derive macro thiserror, [here is an example ](https://github.com/ralexstokes/ethereum-consensus/blob/02a70ad7f9b8a1089011f6c27b77ef1854d86a2d/ethereum-consensus/src/state_transition/error.rs#L10-L57), or derive-more, [here is an example deriving `Display`](https://github.com/emhane/discv5/blob/4750490760931b3ba89afa44594f7187fd74a7b5/src/rpc/notification.rs#L16-L25) and [here is an example deriving `From`](https://github.com/emhane/discv5/blob/4750490760931b3ba89afa44594f7187fd74a7b5/src/error.rs#L6-L64) but `Error` is also available from that crate. disable default features and only bring the features into scope that you use, [like this](https://github.com/emhane/discv5/blob/4750490760931b3ba89afa44594f7187fd74a7b5/Cargo.toml#L41).

emhane: in general, in rust, "Errors must describe themselves through the [Display](https://doc.rust-lang.org/std/fmt/trait.Display.html) and [Debug](https://doc.rust-lang.org/std/fmt/trait.Debug.html) traits. Error messages are typically concise lowercase sentences without trailing punctuation" (source:[ std library docs](https://doc.rust-lang.org/std/error/trait.Error.html))

itamarreif: I agree about `thiserror`, but feels like overkill just for this. When we implement error handling across the sequencer we should probably do that though

crates/astria-sequencer/src/app.rs
SuperFluffy: @joroshiba https://github.com/astriaorg/astria/issues/408

PR #365, opened 2023-09-13
feat(conductor): implement verification of chain IDs commitment

### LINKED ISSUE #308: commitment and verification of namespaces in `SequencerNamespaceData`

noot:
## Summary
previous PR implemented generating the chain IDs commitment in `prepare_proposal` and verifying it in `process_proposal`. this PR finishes the job by ensuring the chain IDs commitment is verified in conductor.

## Background
required so that we can ensure that the sequencer-relayer actually wrote data for every rollup ID in a block, previously it would have been able to omit data for a whole rollup without detection. we still need to implement slashing (eventually) for this - what happens right now is that the conductor will fail to verify a block if the commitment does not match up, even if it was able to obtain the rollup data for its own chain ID.

## Changes
-  ensuring the chain IDs commitment is verified in conductor
- update the sequencer app and `SequencerBlockData::from_tendermint_block` to handle a block with 2 commitments at the start of the txs correctly; this should have been done in my previous PR, apologies
- update ChainId to be max 32 bytes

## Testing
unit tests

## Related Issues

closes #308


crates/astria-conductor/Cargo.toml
```diff
@@ -35,11 +35,12 @@ astria-sequencer-client = { path = "../astria-sequencer-client", features = [
 ] }
 astria-sequencer-relayer = { path = "../astria-sequencer-relayer" }
 astria-sequencer-types = { path = "../astria-sequencer-types" }
+astria-sequencer-validation = { path = "../astria-sequencer-validation" }
```
SuperFluffy, line 38: I have started writing these like so:
```toml
sequencer-validation = { package = "astria-sequencer-validation", path = "../astria-sequencer-validation" }
```
So that in the rust code you have the slightly shorter `sequencer_validation::FooBar`. You can see this here with the telemetry crate.

I think that's neater, but I should have probably brought it up for discussion.

crates/astria-sequencer-types/src/sequencer_block_data.rs
```diff
@@ -43,9 +43,17 @@ pub enum Error {
...
 impl ChainId {
-    #[must_use]
-    pub fn new(inner: Vec<u8>) -> Self {
-        Self(inner)
+    /// Creates a new `ChainId` from the given bytes.
+    ///
+    /// # Errors
+    ///
+    /// - if the given bytes are longer than 32 bytes
+    pub fn new(inner: Vec<u8>) -> eyre::Result<Self> {
```
SuperFluffy, line 51: In PR #376 eyre is removed in favor of typing the errors. Can you do that here as well and merge after that PR?

crates/astria-sequencer-types/src/sequencer_block_data.rs
```diff
@@ -237,22 +253,30 @@ impl SequencerBlockData {
             bail!(Error::MissingDataHash);
         };
 
-        if b.data.is_empty() {
-            bail!("block has no transactions; ie action tree root is missing");
+        if b.data.len() < 2 {
```
SuperFluffy, line 256: Same here - PR #376 changes this slightly.

SuperFluffy (approved):
Change looks good. My only asks is to merge this after #376 goes through - and adapt the changes to `astria-sequencet-types` accordingly.

crates/astria-sequencer-types/src/sequencer_block_data.rs
noot: yep, will wait until that's merged!

SuperFluffy:
@noot PR #375 and #376 have been merged, unblocking (but also breaking :-( ) this. If you want I can fix this PR.

sambukowski (approved):
Other than @SuperFluffy's comments, lgtm

noot:
@SuperFluffy no worries, I can fix it later today!

Issue #332, opened 2023-09-06
remove gossipnet from conductor


joroshiba:
Based on discussion in #325, where we have opted for option 1.

We want to strip out the gossipnet, and instead rely on a full node of the sequencer to read blocks. This is similar to what we are doing in `sequencer-relayer` to read new blocks off of proposing chains. We should do this using a [cometbft websocket subscription](https://docs.cometbft.com/v0.37/core/subscription).

Note: we may need to handle network retry logic in here in case a packet goes missing, and ws connection loses a block.

emhane:
before updating events cometbft validates the block. this means blocks received over the web socket always point to the canonical chain, that is their parent block is the soft block.

apply block tries to validate block and last publishes events
https://github.com/astriaorg/cometbft/blob/342404c9011137cb0e9a4abfd9fa1fc2f9ca889e/state/execution.go#L182-L264

validation checks our equivalent to head block's parent block hash with our equivalent to soft block hash
https://github.com/astriaorg/cometbft/blob/342404c9011137cb0e9a4abfd9fa1fc2f9ca889e/state/validation.go#L45-L51

the state where last block hash is taken from is block applied before this one
https://github.com/astriaorg/cometbft/blob/342404c9011137cb0e9a4abfd9fa1fc2f9ca889e/state/execution.go#L234-L238

so event publisher publishes validated block
https://github.com/astriaorg/cometbft/blob/342404c9011137cb0e9a4abfd9fa1fc2f9ca889e/state/execution.go#L259-L261

this constrains the way to handle incoming blocks.

it means, assuming we trust the validator we are connected to via web socket, all blocks that are received at head height point to the soft block. further this implies, if a child to any of the head blocks is received it will certainly finalise the soft block. also by single slot finality it implies that if head block N+1 is lost but we see its child N+2 and grandchild N+3, we can deduce that the parent block hash in N+2 is the final head.

also we know that any blocks that will become part of the canonical chain, will pass conversion from tendermint block response to `SequencerBlockData` which verifies commit to the rollup data, because all shared-sequencer validators do the same checks on the data before prevoting the block. by this we know that all blocks in the canonical chain will have valid commitment to rollup data, hence with the above example, if N+2 successfully passes the conversion to `SequencerBlockData` we know when N+3 comes that the parent block of N+2 had valid commitment to rollup data.

https://github.com/astriaorg/astria/blob/main/specs/sequencer-app.md#processproposal

emhane:
[...]

Issue #325, opened 2023-09-01
conductor block verification; to gossip or not to gossip?


noot:
# Problem

Conductor needs to be able to validate that the data it receives was actually validly proposed and/or committed by the sequencer. To do this, it needs to be able to track the validator set of the sequencer chain.

Currently, the conductor requires an RPC endpoint of a sequencer node, which is how it obtains this data. However this requires all conductor/rollup nodes to also run a sequencer node, or trust a remote node (not ideal).

This also ties into whether we want to keep the gossipnet at all - right now it is redundant since conductor must be run alongisde a sequencer node.

# Solutions

1. Conductor must have an RPC endpoint of a sequencer node from which it can obtain validator info. This means all rollup nodes must then also run a sequencer node or use a trusted node. If we go with this option we can remove the gossipnet. This is arguably the simplest.
2. Remove the need for conductor to have an RPC endpoint of a sequencer node. Instead, we can implement a "validator set sync protocol" using the p2p layer. The conductor will come bundled with the genesis state of the sequencer chain. Conductor/relayer nodes will implement a p2p request-response protocol used by conductor nodes to sync the validator set from genesis in a trust-minimized way (basically a light client protocol, but just following the sequencer chain's validator set).
3.  Remove the need for conductor to have an RPC endpoint of a sequencer node. Instead, we regularly post the set of validator set changes to DA. As above, conductor nodes will come bundled with the sequencer genesis state, and can sync to the current validator set using this data. Note: we need to post the set of validator set changes since genesis regularly, since DA does not guarantee storage. This means we will be posting more and more data every time the validator set changes.

# Tradeoffs

Option 1 is the simplest to implement engineering-wise, we basically already have it. However, it puts burden on the end-user of a rollup on Astria to run or otherwise obtain a sequencer node. This might be less painful if cometbft light clients exist, but it seems there are none used in production currently?

Option 2 requires us to keep the p2p layer and also add more protocols to it. Sync protocols are non-trivial to implement (cometbft's fast finality makes it easier though). Also keeping the p2p layer will add a lot of engineering work/debugging time down the line - we will need to implement connecting gating, peer scoring, debugging sync, making sure our infra works with another p2p network, etc. 

[...]

PR #366, opened 2023-09-14
Remove gossipnet from relayer

### LINKED ISSUE #333: remove gossipnet from relayer
Based on discussion in https://github.com/astriaorg/astria/issues/325, where we have opted for option 1.

We no longer need to push blocks to gossipnet, should remove setup, configuration, and push to gossipnet from the workflow.

emhane:
## Summary
remove gossipnet from relayer

## Background
relayer has in the past submitted blocks to gossipnet for conductor to receive them earlier than celestia. conductor will receive blocks via RPC instead of p2p from one trusted connection to a sequencer node for the near future.

## Changes
- remove gossip net from relayer

## Testing
removing mock conductor from relayer integration tests, this should not alter the way celestia interacts with tests

## Breaking Changelist
- relayer integration test `slow_celestia_leads_to_bundled_blobs` stops working though removing code that submits a block from gossip net doesn't effect the way a block is submitted to celestia. I am positive it has to do with the test framework, specifically how it interleaves time, which is out of scope of this pr to fix.

## Related Issues
#332, #325

closes #333


SuperFluffy:
I fixed the issue testing the batched celestia blobs.

You are right that it has to do with how time is handled in tokio tests: it auto-advances to the next tick, but that happens faster than the mock response + jsonrpc + deserialization, causing all sorts of issues.

crates/astria-sequencer-relayer/src/relayer.rs
```diff
@@ -266,6 +244,7 @@ impl Relayer {
                 return;
             }
         };
+        println!("submission res {:?}", submission_result);
```
noot, line 247: remove

SuperFluffy:
@emhane If you give permission I will push my patches fixing the blackbox tests to this branch so that CI passes (Monday or Tuesday).

emhane:
> I fixed the issue testing the batched celestia blobs.
> 
> You are right that it has to do with how time is handled in tokio tests: it auto-advances to the next tick, but that happens faster than the mock response + jsonrpc + deserialization, causing all sorts of issues.

if you advance time by the tick interval, naturally this happens. not quite sure how you mean the issue is how tokio handles time in tests? the time is controlled manually in these tests.

if you push the patch to a branch I can check it out.

SuperFluffy:
@emhane see PR #384 which fixes the tests

SuperFluffy (changes requested):
The config must be changed to reflect that writing to celestia is now the only functionality of sequencer-relayer.

Please fix the failing tests before merge. They are fixed in #384

emhane:
> @emhane see PR #384 which fixes the tests

your open pr implements the same issue. I implemented it because it was assigned to, shame with duplicate work, so would have preferred if you would have just assigned yourself.

SuperFluffy:
@emhane 
> > @emhane see PR #384 which fixes the tests
> 
> your open pr implements the same issue. I implemented it because it was assigned to me, shame with duplicate work, so would have preferred if you would have just assigned yourself.
I needed to investigate why the tests were failing to merge this PR, which is the brunt of #384, and because you noted this:

 >  I am positive it has to do with the test framework, specifically how it interleaves time, which is out of scope of this pr to fix.

PR #371, opened 2023-09-15
conductor: infallibly convert verification key to tendermint public key

SuperFluffy:
## Summary
Use an infallible conversion method to convert an ed25519 verification key to a tendermint public key.

## Background
The conversion was unnecessary complex. This makes use of the available tendermint API.

## Changes
- Replace `TryFrom` by `From`, simplifying the code.
- remove some type aliases, turning e.g. `AccountId` into `account::Id` (same for `block` and `chain`, for consistency)

## Testing
Not necessary: tests still pass and this change does not implement new logic.


noot (approved):
nice!

steezeburger (approved):

sambukowski (approved):

PR #373, opened 2023-09-15
conductor: add execute empty blocks setting

### LINKED ISSUE #285: conductor: empty block execution setting

sambukowski:
## Summary
Currently empty blocks are ignored in Conductor. As regular block times are considered more ideal, a setting for users to either execute or ignore empty blocks will be added.

## Changes
- [x] add `disable_empty_block_execution` setting to `Config`
- [x] add `DISABLE_EMPTY_BLOCK_EXECTION` env var to local.env.example
- [x] update Executor `execute_block()` to perform differently based on the `disable_empty_block_execution` setting

## Testing
- [x] add test to see that changing the `disable_empty_block_execution` setting alters execution

## Related Issues
this pr replaces #343 because #329 got merged and changed how configuration works

closes #285


crates/astria-conductor/src/executor.rs
```diff
@@ -393,16 +393,32 @@ mod test {
         }
     }
 
+    fn test_config() -> Config {
```
steezeburger, line 396:
```suggestion
    fn get_test_config() -> Config {
```

crates/astria-conductor/src/executor.rs
```diff
@@ -413,9 +429,10 @@ mod test {
     }
 
     #[tokio::test]
-    async fn execute_block_without_relevant_txs() {
-        let chain_id = ChainId::new(b"test".to_vec());
-        let (mut executor, _) = Executor::new(MockExecutionClient::new(), chain_id)
+    async fn skip_block_without_relevant_txs() {
+        let mut conf = test_config();
+        conf.execute_empty_blocks = false; // empty block will be executed
```
steezeburger, line 434:
```suggestion
        conf.execute_empty_blocks = false;
```

crates/astria-conductor/src/executor.rs
```diff
@@ -458,13 +476,15 @@ mod test {
     }
 
     #[tokio::test]
-    async fn handle_block_received_from_data_availability_no_relevant_transactions() {
-        let chain_id = ChainId::new(b"test".to_vec());
+    async fn handle_block_received_from_data_availability_not_yet_executed_no_relevant_transactions_skip_block()
+     {
+        let mut conf = test_config();
+        conf.execute_empty_blocks = false; // empty block will be executed
```
steezeburger, line 482:
```suggestion
        conf.execute_empty_blocks = false;
```

crates/astria-conductor/src/executor.rs
```diff
@@ -487,4 +507,37 @@ mod test {
...
+
+        let block: SequencerBlockSubset = get_test_block_subset();
+        let expected_execution_state = hash(&executor.execution_state);
+
+        executor
+            .handle_block_received_from_data_availability(block)
+            .await
+            .unwrap();
+
+        // should not have executed or finalized the block
```
steezeburger, line 530:
```suggestion

```

crates/astria-conductor/src/executor.rs
```diff
@@ -487,4 +507,37 @@ mod test {
...
+        assert!(!finalized_blocks.lock().await.is_empty());
```
steezeburger, line 531: I don't think this assertion is necessary. If `finalized_blocks` was empty, the next assertion would also fail.

steezeburger (changes requested):

crates/astria-conductor/local.env.example
```diff
@@ -22,3 +22,6 @@ ASTRIA_CONDUCTOR_DISABLE_FINALIZATION=false
 
 # libp2p Port
 ASTRIA_CONDUCTOR_LIBP2P_PORT=2451
+
+# choose whether or not to execute empty blocks
+ASTRIA_CONDUCTOR_EXECUTE_EMPTY_BLOCKS=true
```
steezeburger, line 27: We have `ASTRIA_CONDUCTOR_DISABLE_FINALIZATION` above.
Should this be something like `ASTRIA_CONDUCTOR_ENABLE_EMPTY_BLOCK_EXECUTION` to keep them more consistent?

I think it would be most ideal to have `ENABLE_FINALIZATION` and `ENABLE_EMPTY_BLOCK_EXECUTION`, but I don't think it's a blocker or anything.

joroshiba: I think that boolean values should normally be false. If at some point in the future we walk back getting rid of clis, and add defaults back in this makes the most sense, setting the flag would be a deviation from normal. `true` values are essentially not the default (this also mimics boolean behaivior in many languages, uninitialized booleans are false).

In this instance I think that means we should make this a `disable`, I like `ASTRIA_CONDUCTOR_DISABLE_EMPTY_BLOCK_EXECUTION`.

noot: agreed w `ASTRIA_CONDUCTOR_DISABLE_EMPTY_BLOCK_EXECUTION` if we want to execute empty blocks by default

sambukowski: updated to `ASTRIA_CONDUCTOR_DISABLE_EMPTY_BLOCK_EXECUTION`

crates/astria-conductor/src/executor.rs
```diff
@@ -487,4 +505,34 @@ mod test {
...
+            .unwrap();
+
+        assert!(
+            finalized_blocks
+                .lock()
+                .await
+                .get(&executor.execution_state)
+                .is_some()
+        );
+        assert_eq!(expected_execution_state, executor.execution_state,);
```
noot, line 534:
```suggestion
        assert_eq!(expected_execution_state, executor.execution_state);
```

crates/astria-conductor/src/executor.rs
```diff
@@ -487,4 +505,34 @@ mod test {
...
+        // should be empty because nothing was executed
+        assert!(executor.sequencer_hash_to_execution_hash.is_empty());
```
noot, line 536: is this correct? from the test name + previous lines i would expect something to be executed here

joroshiba (approved):

crates/astria-conductor/src/executor.rs
sambukowski: The comment here was wrong, but the test is correct. The block is getting executed and finalized so it should not appear in the `sequencer_hash_to_exection_hash` map.

crates/astria-conductor/src/executor.rs
```diff
@@ -487,4 +505,34 @@ mod test {
         // should be empty because nothing was executed
         assert!(executor.sequencer_hash_to_execution_hash.is_empty());
     }
+
+    #[tokio::test]
+    async fn handle_block_received_from_data_availability_not_yet_executed_no_relevant_transactions_execute_block()
```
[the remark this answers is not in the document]
sambukowski: Updated the test names to make them more clear.

crates/astria-conductor/src/executor.rs
```diff
@@ -55,11 +55,7 @@ pub(crate) async fn spawn(conf: &Config) -> Result<(JoinHandle, Sender)> {
         "Spawning executor task."
     );
     let execution_rpc_client = ExecutionRpcClient::new(&conf.execution_rpc_url).await?;
```
SuperFluffy, line 57: Please add error context (even though it's not part of this PR)

crates/astria-conductor/src/executor.rs
```diff
@@ -117,21 +113,26 @@ struct Executor<C> {
...
     /// we receive a finalized sequencer block.
     sequencer_hash_to_execution_hash: HashMap<Hash, Vec<u8>>,
+
+    /// Chose to execute empty blocks or not
+    disable_empty_block_execution: bool,
 }
 
 impl<C: ExecutionClient> Executor<C> {
-    async fn new(mut execution_rpc_client: C, chain_id: ChainId) -> Result<(Self, Sender)> {
+    async fn new(mut execution_rpc_client: C, conf: &Config) -> Result<(Self, Sender)> {
```
SuperFluffy, line 122: IMO it's nicer to not pass the conf around as a "grab back". I prefer the former version (we are not doing a good job of adhering to this in other parts of the code base either).

crates/astria-conductor/src/executor.rs
```diff
@@ -117,21 +113,26 @@ struct Executor<C> {
...
         let init_state_response = execution_rpc_client.call_init_state().await?;
         let execution_state = init_state_response.block_hash;
         Ok((
             Self {
                 cmd_rx,
                 execution_rpc_client,
-                chain_id: chain_id.clone(),
-                namespace: Namespace::from_slice(chain_id.as_ref()),
+                chain_id: ChainId::new(conf.chain_id.as_bytes().to_vec())
+                    .expect("failed to create ChainId"),
```
SuperFluffy, line 131: Please don't just panic but return an error with context

SuperFluffy (approved):
General logic looks fine, but please address the comments before merging

crates/astria-conductor/src/executor.rs
sambukowski: Put this back to how it was before.

steezeburger (approved):

SuperFluffy (approved):
Few more nits, nothing major

PR #375, opened 2023-09-15
sequencer-validation: type errors

SuperFluffy:
## Summary
Replace all untyped errors in sequencer-validation by typed errors.

## Background
`sequencer-validation` has to be made a dependency of `sequencer-client`, which only uses typed errors. Therefore `sequencer-validation` must have typed errors (because otherwise `sequencer-client` would be "infected" by eyre, which would require changing all of sequencer-client to also use eyre :-/ ).

## Changes
- Replace `eyre` reports by typed errors
- Fix `MerkleTree::prove_inclusion` to actually return an error: the docs said it would error, but it actually panicked if `index` was out of bounds.
- Add eyre error context to sequencer-relayer: because the error in it wasn't properly wrapped, changing the sequencer-validation types caused it to no longer compile

## Testing
Added a test to check for out of bounds inclusion proof index. The other error type is just a wrapper.

## Related Issues
This is done as part of #332, which now requires extra trait methods in the `SequencerClientExt` and `SequencerSubscriptionClientExt` extension traits.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -595,7 +595,6 @@ version = "0.1.0"
 dependencies = [
  "bytes",
  "ct-merkle",
- "eyre",
  "hex",
  "serde",
  "serde_json",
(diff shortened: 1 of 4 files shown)


noot (approved):
looks good! should we move the errors to their own module or do you think it's fine for now (since there's just 2)?

SuperFluffy:
@noot I like to keep errors close to their use so that the logic is one coherent unit especially if it's so few lines and also only used here.

Makes sense especially for bigger errors that are used throughout a codebase.

dereksione (approved):

PR #376, opened 2023-09-17
sequencer-types: type all errors

SuperFluffy:
## Summary
Replace all implicit eyre errors by context specific typed errors

## Background
Same reasoning as for https://github.com/astriaorg/astria/pull/375 as part of the work for https://github.com/astriaorg/astria/issues/332:

`sequencer-types` has to be made a dependency of `sequencer-client`, which only uses typed errors. Therefore `sequencer-types` must have typed errors (because otherwise `sequencer-client` would be "infected" by eyre, which would require changing all of sequencer-client to also use eyre.

## Changes
- Introduced one error variant per distinct eyre error.

## Testing
N/A. Nothing changed on the code level. Testing that the errors are actually triggered could be done in a follow-up PR, but big parts of the code will be replaced by the work in https://github.com/astriaorg/astria/pull/355 and follow-up PRs to it, moving the types exchanged over RPCs to the astria-proto crate.

## Breaking Changelist
- No observable breaking changes in the astria services. Breaking changes in the library crates because their public API changed (errors), but we are not considering them right now

## Related Issues
https://github.com/astriaorg/astria/pull/12
https://github.com/astriaorg/astria/issues/86
https://github.com/astriaorg/astria/issues/366
https://github.com/astriaorg/astria/issues/332


crates/astria-sequencer-types/src/sequencer_block_data.rs
```diff
@@ -1,19 +1,23 @@
...
+use sequencer_validation::{
+    InclusionProof,
+    MerkleTree,
 };
+// use eyre::{
+//     bail,
+//     ensure,
+//     eyre,
+//     WrapErr as _,
+// };
```
noot, line 20: remove

noot (approved):
nice!

crates/astria-sequencer-types/src/sequencer_block_data.rs
SuperFluffy: Done, ty for noticing

joroshiba (approved):

Issue #374, opened 2023-09-15
Implement std Error for ct-merkle errors


SuperFluffy:
The errors inside the ct-merkle library used in `astria-sequencer-validation` should implement `std::error::Error`.

noot:
i'm looking into rewriting this lib, let's put this on hold for now

PR #377, opened 2023-09-17
update ct-merkle dependency

noot:
## Summary
my PR for the ct-merkle lib got merged so can remove the patch.


SuperFluffy (approved):
Nice.

Have you seen https://github.com/astriaorg/astria/issues/374 ?

Wonder if we should just fork it and pull it into the monorepo

noot:
@SuperFluffy yeah, I'm considering rewriting that lib anyways, it's pretty unoptimized for our use-case right now.

PR #382, opened 2023-09-18
sequencer-client: add extension trait to subscriber to new sequencer blocks

SuperFluffy:
## Summary
Add a `SequencerSubscriptionClientExt` trait with a `subscribe_new_blocks` method. 

## Background
Removing gossipnet from conductor as part of https://github.com/astriaorg/astria/issues/332 requires it to read new blocks directly from cometbft / sequencer. This patch extends the sequencer-client crate to provide that functionality using a convenience function so that conductor need not know how to convert tendermint new block events to sequencer blocks.

## Changes
- Add a `SequencerSubscriptionClientExt` trait to the sequencer-client crate.
- Reorganize the sequencer-client crate to export items under the appropriate features.
- warn users if they set neither the `http` nor `websocket` features.

## Testing
Testing this client is out of scope for this PR. We are relying on the underlying tendermint-rpc WebSocketClient being correctly implemented. It would be great to extend wiremock or a similar to provide a quick way to test websocket APIs.

An alternative is to derive a jsonrpc server with jsonrpsee, but that's a lot more work.

## Breaking Changelist
None, only functionality was added crates depending on sequencer-client were not touched.

## Related Issues
Part of https://github.com/astriaorg/astria/issues/332

https://github.com/astriaorg/astria/pull/375 and https://github.com/astriaorg/astria/pull/376 are precursor PRs unblocking this patch.


crates/astria-sequencer-client/src/extension_trait.rs
```diff
@@ -0,0 +1,458 @@
...
+}
+
+#[async_trait]
+pub trait SequencerSubscriptionClientExt: SubscriptionClient {
+    /// Subscribe to sequencer blocks from cometbft, returning a stream.
+    ///
+    /// This trait method calls the cometbft `/subscribe` endpoint with a
+    /// `tm.event = 'NewBlock'` argument, and then attempts to convert each
+    /// cometbft block to a `SequencerBlockData` type.
+    async fn subscribe_new_blocks(&self) -> Result<NewBlocksStream, SubscriptionFailed> {
```
noot, line 272:
```suggestion
    async fn subscribe_new_block_data(&self) -> Result<NewBlocksStream, SubscriptionFailed> {
```

crates/astria-sequencer-client/src/extension_trait.rs
```diff
@@ -0,0 +1,458 @@
...
+    fn poll_next(
+        mut self: Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Option<Self::Item>> {
+        self.inner.as_mut().poll_next(cx)
+    }
+}
+
+#[async_trait]
+pub trait SequencerSubscriptionClientExt: SubscriptionClient {
```
noot, line 266: does this need to be feature-gated?

crates/astria-sequencer-client/src/extension_trait.rs
```diff
@@ -0,0 +1,458 @@
...
+    },
+    event::EventData,
+    Client,
+    SubscriptionClient,
+};
+
+#[cfg(feature = "http")]
+impl SequencerClientExt for HttpClient {}
+#[cfg(feature = "websocket")]
+const _: () = {
```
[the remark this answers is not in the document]
SuperFluffy: It allows me to group all these impls in one block. `{}` is not available at the module level, so this is not allowed:
```rust
# not allowed
#[cfg(feature = "websocket")]
{
    impl Foo for Bar {}
    impl Foo for Baz {}
}
```
This would work, but it's very repetitive:
```rust
# so much noise
#[cfg(feature = "websocket")]
impl Foo for Bar {}
#[cfg(feature = "websocket")]
impl Foo for Baz {}
```
Or this one, but there is a module in the namespace which has no right being there.
```rust
#[cfg(feature = "websocket")]
mod __impls {
    impl Foo for Bar {}
    impl Foo for Baz {}
}

noot (approved):

crates/astria-sequencer-client/src/extension_trait.rs
SuperFluffy: Yeah this makes sense. I just had the thought that we might want to change `SequenderBlockData` to `SequencerBlock`? Or is that confusing?

sambukowski (approved):

PR #385, opened 2023-09-18
fix(sequencer): fix init_chain and update astria-sequencer-utils

### LINKED ISSUE #227: sequencer: investigate and fix `init_chain`

noot:
## Summary
cometbft was previously not calling `init_chain` as it should have due to the `Info` response being wrong.

## Background
need the sequencer to work properly :) `init_chain` contains genesis validators amongst other info, which we need to store in  #359

the issue was that previously the `Info` response returned:
```
last_block_height: Height::default(),
```
however Height::default() is 1, so cometbft thought the app had already been initialized. if height is 0, then cometbft sends the init_chain message.

## Changes
- fix ABCI `Info` response, init_chain is now called properly
- init_chain passes in the `app_state` from the cometbft genesis file, so we don't need to manually parse the app genesis anymore
- update `astria-sequencer-utils` and readmes to properly generate the cometbft genesis file including the genesis app state

## Testing
instructions in readme

## Related Issues

closes #227


crates/astria-sequencer/src/service/consensus.rs
```diff
@@ -131,28 +131,32 @@ impl Consensus {
             bail!("database already initialized");
         }
 
-        let genesis_state: GenesisState = serde_json::from_slice(&init_chain.app_state_bytes)
-            .expect("can parse app_state in genesis file");
+        let genesis_state: GenesisState = serde_json::from_str(
```
noot, line 134: i need to update to use from_bytes

SuperFluffy: IMO not obfuscating that this is json is a good thing.

crates/astria-sequencer/src/service/info/mod.rs
```diff
@@ -70,11 +67,27 @@ impl Info {
...
+                let app_hash = crate::app_hash::AppHash::from(
+                    self.storage
+                        .latest_snapshot()
+                        .root_hash()
+                        .await
+                        .context("failed to get app hash")?,
+                );
+
                 let response = InfoResponse::Info(response::Info {
                     version: "0.1.0".to_string(),
```
SuperFluffy, line 85: You can use `env!("CARGO_PKG_VERSION")` to compile the version of sequencer into this response. See here for available cargo env vars: https://doc.rust-lang.org/cargo/reference/environment-variables.html

You can make use of the https://crates.io/crates/vergen crate to also grab the sha commit, which would be even more useful (because our versions right now are pretty meaningless).

crates/astria-sequencer/src/service/consensus.rs
```diff
@@ -131,28 +131,32 @@ impl Consensus {
...
+            &String::from_utf8(init_chain.app_state_bytes.into())
+                .context("failed to convert app_state_bytes to string")?,
+        )
+        .expect("can parse app_state in genesis file");
```
SuperFluffy, line 138: Let's return this as an error.

crates/astria-sequencer-utils/src/genesis_parser.rs
```diff
@@ -58,23 +62,19 @@ impl GenesisParser {
 
 /// Merges a source JSON Value into a destination JSON Value context:
 // https://stackoverflow.com/questions/47070876/how-can-i-merge-two-json-objects-with-rust
```
SuperFluffy, line 64: this comment probably does not make sense anymore

crates/astria-sequencer-utils/src/genesis_parser.rs
```diff
@@ -58,23 +62,19 @@ impl GenesisParser {
...
-fn merge_json(a: &mut Value, b: &Value) {
-    match (a, b) {
-        (Value::Object(a), Value::Object(b)) => {
-            for (k, v) in b {
-                merge_json(a.entry(k).or_insert(Value::Null), v);
-            }
+fn insert_app_state(dst: &mut Value, app_state: &Value) {
+    match dst {
```
SuperFluffy, line 66: This is cleaner:
```rust
let Value::Object(dst) = dst else {
    panic!(...);
};
dst.insert(...);
```

SuperFluffy (approved):
Just a few comments here and there..

Can you explain why exactly the previous info response was wrong? Why did cometbft not like it and hence not call init_chain? Would be great in the background section for learning.

crates/astria-sequencer/src/service/info/mod.rs
noot: nice, will do!

crates/astria-sequencer-utils/src/genesis_parser.rs
noot: removed

noot:
@SuperFluffy the exact issue was this, previously the Info response returned:
```
last_block_height: Height::default(),
```

however `Height::default()` is `1`, so cometbft thought the app had already been initialized. if height is 0, then cometbft sends the `init_chain` message.

SuperFluffy:
@noot can you add that to the PR/commit message? 🙏

joroshiba (approved):

Issue #386, opened 2023-09-18
Conductor config tests may fail when extraneous envars are defined in .env


steezeburger:
There is a test in the Conductor code base `ensure_example_env_is_in_sync` that can improperly fail when the tests are ran via `just`.

Our `justfile` uses `set dotenv-load` which loads envars from a `.env` file before running the defined command. If an extra envar exists in .env that is not present in Conductor's `Config` type (maybe because the coder added it for work on another feature branch), then the `ensure_example_env_is_in_sync` test will fail, even though a user's local .env should never affect the tests.

Figment's Jail isn't truly sandboxed, so when `Config::from_environment` is called, it sees all envars loaded from `just` and from `jail` that start with `ASTRIA_CONDUCTOR_`. 

We could refactor `from_environment` and `populate_environment_from_example` to take a prefix that is prepended to `ASTRIA_CONDUCTOR_`. This way, only the correctly prefixed envars are loaded and

steezeburger:
Might be a better idea to pull config functionality out into its own crate, e.g. `astria-config`

PR #387, opened 2023-09-18
Conductor: fix config loading from environment to have more control

steezeburger:
## Summary
This PR refactors how we load values from the environment so that we have more control on which envars are used.
We can now define the `envar_prefix` we want as well as a `test_envar_prefix` so that we have full control over which envar names should be part of the config.

## Background
A user's locally loaded environment should not affect test outcomes.

## Changes
- added `envar_prefix` argument to `from_environment`
- added `test_envar_prefix` to `populate_environment_from_example`

## Testing
- I fixed the test that was erroneously failing previously.

## Breaking Changelist
- You must now pass in an `envar_prefix` to `Config::from_environment`

closes https://github.com/astriaorg/astria/issues/386

### DIFF
--- a/crates/astria-conductor/src/config.rs
+++ b/crates/astria-conductor/src/config.rs
@@ -9,7 +9,7 @@ use serde::{
 };
 
 pub fn get() -> Result<Config, figment::Error> {
-    Config::from_environment(None)
+    Config::from_environment("ASTRIA_CONDUCTOR_")
 }
 
 /// The global configuration for the driver and its components.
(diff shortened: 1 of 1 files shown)


sambukowski (approved):

crates/astria-conductor/src/config.rs
```diff
@@ -66,10 +66,14 @@ where
 }
 
 impl Config {
-    fn from_environment() -> Result<Config, figment::Error> {
+    fn from_environment(prefix: Option<&str>) -> Result<Config, figment::Error> {
```
noot, line 69: it appears `None` is the only value passed in here (excluding tests), maybe just remove the param and the `Some` case?

SuperFluffy (changes requested):
IMO the code can be simplified by not using the option but instead changing `Config::from_environment` to take a `&str`, and then defining a hardcoded string to use in `fn get`.

Also, I think it would be great if we expanded this to the other crates.

joroshiba (approved):
Generally LGTM, although I agree with comments that we should set a const and always take a string instead of using `None` value.

Assuming that is taken care of, approved.

steezeburger:
> IMO the code can be simplified by not using the option but instead changing `Config::from_environment` to take a `&str`, and then defining a hardcoded string to use in `fn get`.
> 
> Also, I think it would be great if we expanded this to the other crates.

Totally! I made an issue for pulling this out into its own crate https://github.com/astriaorg/astria/issues/398

SuperFluffy (approved):
Looking good, ty for the changes

Issue #32, opened 2023-05-19
refactor services into rich library types and minimalist binary wrappers


SuperFluffy:
At the moment `astria-conductor` and `astria-sequencer-relayer` have overly complex binary definitions. Their binary entrypoints should be as minimal as possible, while their backing libraries should only expose:
1. getting the config
2. setting up telemetry
3. running the service (`Conductor` and `SequencerRelayer, respectively)

The primary goals are:
1. code clarity (i.e. making it immediately clear where to main entrypoint to the services are)
2. decoupling (binary crates must not import items from other binary crates)
3. running integration tests against the same codepaths that are used when running the services as binaries

```[tasklist]
### Tasks
- [ ] https://github.com/astriaorg/astria/issues/60
- [ ] https://github.com/astriaorg/astria/issues/90
```

PR #330, opened 2023-09-06
Updating sequencer config implementation

dereksione:
## Summary
Updating how the sequencer does configuration management to use the specs defined in the configuration specs

## Background
Addressing the tech debt of different crates using different configuration setups and hierarchies

## Changes
- Removed manual env parsing in favor of figment
- Removed clap parsing
- Added justfile
- Added example .env
- Added common TESTNET commands to justfile
- Combined README and TESTNET file

## Testing
Added new figment tests for the config setup

## Breaking Changelist
- All config management should be done using .env files now, as outlined in the spec - cli overrides support is removed

## Related Issues
Link any issues that are related, prefer full github links.

closes #245

PR #392, opened 2023-09-19
conductor: refactor with a single `Conductor` type entrypoint.

SuperFluffy:
## Summary
Move most logic out of the binary into a `Conductor` type.

## Background
Astria services are intended to have a single entry point with a thin binary. Refactoring them this way allows to run actual blackbox against the service as if it were started from the compiled artifact.

## Changes
- Create a `Conductor` type with inherent methods `new` and `run_until_stopped`
- Rename `bin/astria-condicutor.rs` as `main.rs` to make it clear that this is a single-binary crate.
- Rewrite `main.rs` in terms of the `Conductor` type

## Testing
Tests still run. This change will make it possible to run blackbox tests against conductor in the future.

## Breaking Changelist
- None of the changes here should be visible to an outside observer.

## Related Issues

Tracking issue: https://github.com/astriaorg/astria/issues/32

closes https://github.com/astriaorg/astria/issues/92


noot:
closes #95 also?

crates/astria-conductor/src/conductor.rs
```diff
@@ -0,0 +1,173 @@
...
+    reload_rx: watch::Receiver<()>,
+    stop_rx: watch::Receiver<()>,
+}
+
+fn spawn_signal_handler() -> SignalReceiver {
+    let (stop_tx, stop_rx) = watch::channel(());
+    let (reload_tx, reload_rx) = watch::channel(());
+    tokio::spawn(async move {
+        let mut sighup = signal(SignalKind::hangup()).expect(
+            "setting a SIGHUP listener should always work on linux; is this running on linux?",
```
noot, line 143: does this work on unix or just linux?

crates/astria-conductor/src/main.rs
```diff
@@ -0,0 +1,56 @@
...
+    config,
+};
+use tracing::{
+    error,
+    info,
+};
+
+// Following the BSD convention for failing to read config
+// See here: https://freedesktop.org/software/systemd/man/systemd.exec.html#Process%20Exit%20Codes
+const EX_CONFIG: u8 = 78;
```
noot, line 14: cool, are we going to add this to the other crates too?

noot (approved):
nice!!

crates/astria-conductor/src/conductor.rs
SuperFluffy: Good catch, this is a unix thing, not a Linux thing. Will change

crates/astria-conductor/src/main.rs
steezeburger: I could do this as part of https://github.com/astriaorg/astria/issues/398

steezeburger (approved):

crates/astria-conductor/src/main.rs
SuperFluffy: @steezeburger Awesome, sync with @dereksione who has been doing some config work in #330

Issue #31, opened 2023-05-19
flatten `SequencerBlock` protos


itamarreif:
# Current situation
The current `SequencerBlock` proto definition has a nested array of transactions:
```protobuf
// data.proto
// IndexedTransaction represents a sequencer transaction along with the index
// it was originally in the sequencer block.
message IndexedTransaction {
    uint64 block_index = 1; // TODO: this is usize - how to define for variable size?
    bytes transaction = 2;
}

// block.proto
message NamespacedIndexedTransactions {
    bytes namespace = 1;
    repeated IndexedTransaction txs = 2;
}

// SequencerBlock
message SequencerBlock {
    bytes block_hash = 1;
    tendermint.types.Header header = 2;
    repeated IndexedTransaction sequencer_transactions = 3;
    repeated NamespacedIndexedTransactions rollup_txs = 4;
}
```

# The problem
The problem with this, as described by Janis:
"From my experience deeply nested repeated types (so fields with a `repeated` prefix and things like `map`) are tricky from a performance-standpoint, because rust (or any language) can't amortize memory allocation away.So if you have soomething like `Vec<Vec<T>>` and lots of small inner vectors (say with 20 elements on average), you will end up with 4 memory allocations per inner vector (4 -> 8 -> 16 -> 32). Let's say you have 100 elements in the outer vector, that's 406 memory allocations in total, 400 memory allocations for the inner vecs + 6 allocations for the outer vec (`4 -> 8 -> 16 -> 32 -> 64 -> 128`).If you find a way to flatten this data structure so that you just have a 1d vector with 400 elements in total, you can bring this down from 406 to 8 (`128 -> 256 -> 512`)."

# Proposed solutions
As discussed, there are several possible solutions to this:
1. Flatten the `namespace` into each `IndexedTransaction`, i.e. have:
    ```protobuf
    // block.proto
    message NamespacedIndexedTransaction {
        bytes namespace = 1;
        IndexedTransaction tx = 2;
    }
    
    // SequencerBlock
    message SequencerBlock {
        bytes block_hash = 1;
        tendermint.types.Header header = 2;
        repeated IndexedTransaction sequencer_transactions = 3;
        repeated NamespacedIndexedTransaction rollup_txs = 4;
    }
    ```
    where `SequencerBlock.rollup_txs` is the flattened vector of all the rollup txs.
[...]

Issue #20, opened 2023-03-11
sequencer: use serialization for data that isn't json


noot:
Currently, we're just using json for serializing data before writing to Celestia. We should probably use something like protobuf at least for now, and eventually look into compression formats.

itamarreif:
issue #12 adds proto definitions to the `astria-proto` crate

still need to do:
- [ ] convert from proto types into the sanitized types we have in the  crate (e.g. conversions from protobuf `bytes` to `Base64String`)
- [ ] replace JSON deserialization with protobuf in `sequencer-relayer`
- [ ] update `conductor` and `sequencer-relayer` to use proto defs
- [ ] get rid of duplicate `tendermint-rs` types (some, e.g. `Header`, were copied over into [`types.rs`](https://github.com/astriaorg/astria/blob/main/crates/astria-sequencer-relayer/src/types.rs)) and use [`tendermint-proto`](https://docs.rs/tendermint-proto/latest/tendermint_proto/) instead

itamarreif:
related documentation tasks:
- [ ] comments explaining what `block_hash` and `block_height` refer to in `SequencerNamespaceData` and `RollupNamespaceData`
    - [ ] in rust code
    - [ ] in proto definitions
- [ ] comments explaining what `block_index` refers to in `IndexedTransaction` and what it's used for
    - [ ] in rust code
    - [ ] in proto definitions
- [ ] relationship between namespace data objects and the block object
    - [ ] comments on the rust types
    - [ ] comments on the proto definitions
    - [ ] hopefully a data flow diagram for the sequencer-relayer

noot:
blocked by #258

noot:
closed by #395

PR #395, opened 2023-09-19
feat: redefine sequencer blocks, celestia blobs as protobuf

SuperFluffy:
## Summary
This patch defines protobuf versions of various types that are currently only defined in Rust and exchange as json-serialized text. 

## Background
`astria-proto` is intended to be the single source of truth for the wire format of data that is exchanged between services. At the moment `SequencerBlockData`, `SequencerNamespaceData`, `RollupNamespaceData` and `InclusionProof` are only defined in Rust. Some older versions of these exist as protobuf but were never used in the public interfaces.

## Changes
- Replace `SequencerBlockData` by a protobuf `SequencerBlock` (rewriting this protobuf type is technically a breaking change, but it was never used, so in practice it is not)
- Replace `SequencerNamespaceData` and `RollupNamespaceData` by protobuf `CelestiaSequencerBlob` and `CelestiaRollupBlob`.
- Introduce other protobuf types like `RollupTransactions`, `Proof` to decode/encode all required objects
- Write idiomatic rust types for all generated protobuf representations to uphold invariants

## Testing
- all tests have been updated and still pass
- testing this is in a devnet to check that the data still flows is open

## Related Issues

https://github.com/astriaorg/astria/pull/12 originally defined sequencer blocks and transactions as protobuf, but these were never actually implemented.

https://github.com/astriaorg/astria/issues/86 is the reason this effort was started.


SuperFluffy:
Opened a new PR because something broke https://github.com/astriaorg/astria/pull/355. No changes with respect to the most recent review.

crates/astria-proto/proto/astria/sequencer/v1alpha2/celestia.proto
```diff
@@ -0,0 +1,48 @@
...
+// by `sequencer_block_hash`.
+message CelestiaRollupData {
+    // The hash of the sequencer block. Must be 32 bytes.
+    bytes sequencer_block_hash = 1;
+    // The human readable chain ID identifiying the rollup these transactions belong to.
+    string chain_id = 2;
+    // A list of opaque bytes that are serialized rollup transactions.
+    repeated bytes transactions = 3;
+    // The proof that the action tree root was included in
+    // `sequencer_block_header.data_hash` (to be found in `CelestiaHeader`).
```
noot, line 21:
```suggestion
    // `sequencer_block_header.data_hash` (to be found in `CelestiaSequencerData`).
```

noot (approved):
looks good! are you planning to do #20 (and move the types to astria-proto) in a follow up?

also is  #31 still relevant or can it be closed?

SuperFluffy:
@noot yeah, #20 should be the immediate follow-up to this PR

#31 will probably be outdated with this change.

I have dismissed your review because I think this PR needs more discussion in light of #401 (to ensure this does not accidentally sneak in).

noot:
we can put #401 on hold for now - don't think we need to change the verification stuff yet, probably going to close #401 and open another broader issue. so this should be good to go in!

noot (approved):

crates/astria-proto/proto/astria/sequencer/v1alpha1/block.proto
joroshiba, line 1: are these files supposed to be in v1alpha2? Where are the v1alpha2 generated files coming from?

SuperFluffy: forgot to push the cleanup

SuperFluffy:
The proto lint will fail because this is a breaking proto buf change (which is ok, because the proto definitions were never used)

.github/workflows/test.yml
```diff
@@ -88,6 +88,7 @@ jobs:
           cargo test --release --workspace \
           --all-features \
           --all-targets \
+          -- --include-ignored
```
SuperFluffy, line 91: I set all optimism tests as `#[ignore]` because I did not have the required dependencies installed.

crates/astria-proto/proto/astria/sequencer/v1alpha1/transaction.proto
```diff
@@ -51,10 +51,10 @@ message TransferAction {
 // `SequenceAction` represents a transaction destined for another
 // chain, ordered by the sequencer.
 //
-// It contains the chain ID of the destination chain, and the
+// It contains the rollup ID of the destination chain, and the
 // opaque transaction data.
 message SequenceAction {
-    bytes chain_id = 1;
+    bytes rollup_id = 1;
```
SuperFluffy, line 57: Renamed this field to be in line with renaming chain ID -> rollup ID.

This is not a breaking protobuf change (field names don't matter)

crates/astria-proto/src/lib.rs
```diff
@@ -1,3 +1,8 @@
+#[cfg(not(target_pointer_width = "64"))]
```
SuperFluffy, line 1: I did not want to worry about casting usize to u64 and back. Astria Merkle is written entirely using `usize` so it works on any architecture. However, moving in between architectures is more tricky. So 64 bit it is.

crates/astria-proto/proto/astria/sequencer/v1alpha1/block.proto
```diff
@@ -3,21 +3,41 @@ syntax = "proto3";
...
-    bytes block_hash = 1;
-    tendermint.types.Header header = 2;
-    repeated IndexedTransaction sequencer_transactions = 3;
-    // FIXME: the current nested array layout results in bad allocation behavior on deserialization
-    // see https://github.com/astriaorg/astria/issues/31
-    repeated NamespacedIndexedTransactions rollup_transactions = 4;
+    // The original CometBFT header that was the input to this sequencer block.
+    tendermint.types.Header header = 1;
+    // The collection of rollup transactions that were included in this block.
+    repeated RollupTransactions rollup_transactions = 2;
```
SuperFluffy, line 26: Rather than using a map here this is now a list of `(Id, [bytes])` tuples. This means that any language will get these in the correct order.

crates/astria-proto/src/native/sequencer/v1alpha1.rs
SuperFluffy, line 1: Note on this file: we will split this in the future because it's getting too unwieldy. It should be done in a separate PR however to not mix feature changes and simply moving code around

crates/astria-proto/src/native/sequencer/v1alpha1.rs
```diff
@@ -1001,115 +1006,125 @@ impl Display for Address {
     }
 }
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, PartialOrd, Ord)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[cfg_attr(feature = "serde", serde(transparent))]
-pub struct ChainId {
+pub struct RollupId {
```
SuperFluffy, line 1012: @noot @joroshiba I wanted to especially flag this: I renamed chain ID to rollup ID everywhere.

crates/astria-proto/src/native/sequencer/v1alpha1.rs
```diff
@@ -1251,6 +1266,1158 @@ impl Display for IncorrectAddressLength {
...
+            rollup_transactions_proof,
+            rollup_ids_proof,
+        }
+    }
+
+    /// Turn the sequencer block into a [`CelestiaSequencerBlob`] and its associated list of
+    /// [`CelestiaRollupBlob`]s.
+    #[must_use]
+    #[allow(clippy::missing_panics_doc)] // the proofs are guaranteed to exist; revisit if refactoring
+    pub fn to_celestia_blobs(&self) -> (CelestiaSequencerBlob, Vec<CelestiaRollupBlob>) {
```
SuperFluffy, line 1573: This method used to be a free standing method in `astria-celestia-client`. Because all these types are defined here it's natural to leave this here.

crates/astria-proto/src/native/sequencer/v1alpha1.rs
```diff
@@ -1251,6 +1266,1158 @@ impl Display for IncorrectAddressLength {
...
+        }
+        (head, tail)
+    }
+
+    /// Converts from a [`tendermint::Block`].
+    ///
+    /// # Errors
+    /// TODO(https://github.com/astriaorg/astria/issues/612)
+    #[allow(clippy::missing_panics_doc)] // the panic sources are checked before hand; revisit if refactoring
+    pub fn try_from_cometbft(block: tendermint::Block) -> Result<Self, SequencerBlockError> {
```
SuperFluffy, line 1613: This was moved from `astria-sequencer-types` and enriched by checking more invariants.

crates/astria-proto/src/native/sequencer/v1alpha1.rs
```diff
@@ -1251,6 +1266,1158 @@ impl Display for IncorrectAddressLength {
...
+    let mut tree = merkle::Tree::new();
+    for (rollup_id, txs) in rollup_ids_to_txs {
+        let root = merkle::Tree::from_leaves(txs).root();
+        tree.build_leaf().write(rollup_id.as_ref()).write(&root);
+    }
+    tree
+}
+
+// TODO: This can all be done in-place once https://github.com/rust-lang/rust/issues/80552 is stabilized.
+pub fn group_sequence_actions_in_signed_transaction_transactions_by_rollup_id(
```
SuperFluffy, line 2391: This was moved/adapted from astria-sequencer. Because other parts of the code are interacting with it (mainly through tests) it made sense to me to put it here, also because it's closely relatted to `derive_merkle_from_rollup_txes`

crates/astria-conductor/src/data_availability/block_verifier.rs
```diff
@@ -423,32 +409,18 @@ mod test {
...
-        let rollup_namespace_data = RollupNamespaceData {
-            block_hash,
-            chain_id,
-            rollup_txs: vec![test_tx],
-            inclusion_proof: action_tree.construct_proof(0).unwrap(),
-        };
-
-        validate_sequencer_namespace_data(&validator_set, &commit, &sequencer_namespace_data)
-            .unwrap();
-        rollup_namespace_data
```
SuperFluffy, line 449: This check no longer made sense because the method to fetch the blobs in `astria-celestia-client` now also ensures that only blobs that are related to the sequencer blob are returned.

crates/astria-conductor/src/conductor.rs
```diff
@@ -99,12 +95,26 @@ impl Conductor {
...
         let (executor_tx, sync_start_block_height) = {
             let (block_tx, block_rx) = mpsc::unbounded_channel();
             let (shutdown_tx, shutdown_rx) = oneshot::channel();
 
-            let executor = make_executor(&cfg, block_rx, shutdown_rx)
+            let hook = make_optimism_hook(&cfg)
+                .await
+                .wrap_err("failed constructing optimism hook")?;
+
+            let executor = Executor::builder()
```
joroshiba, line 110: Nice

crates/astria-proto/proto/astria/sequencer/v1alpha1/celestia.proto
```diff
@@ -0,0 +1,53 @@
...
+    // The rollup IDs for which `CelestiaRollupBlob`s were submitted to celestia.
+    // Corresponds to the `astria.sequencer.v1alpha1.RollupTransactions.rollup_id` field
+    // and is extracted from `astria.sequencer.v1alpha.SequencerBlock.rollup_transactions`.
+    repeated bytes rollup_ids = 2;
+    // The Merkle Tree Hash of the rollup transactions. Corresponds to
+    // `MHT(astria.sequencer.v1alpha.SequencerBlock.rollup_transactions)`, the Merkle
+    // Tree Hash deriveed from the rollup transactions.
+    // Always 32 bytes.
+    bytes rollup_transactions_root = 3;
+    // The proof that the rollup transactions are included in sequencer block.
```
noot, line 47:
```suggestion
    // The proof that the rollup transactions are included in sequencer block header.
```

crates/astria-proto/proto/astria/sequencer/v1alpha1/celestia.proto
```diff
@@ -0,0 +1,53 @@
...
+    // Corresponds to `astria.sequencer.v1alpha.SequencerBlock.rollup_transactions_proof`.
+    astria.sequencer.v1alpha1.Proof rollup_transactions_proof = 4;
+    // The proof that the rollup IDs are included in sequencer block.
+    // Corresponds to `astria.sequencer.v1alpha.SequencerBlock.rollup_ids_proof`.
+    astria.sequencer.v1alpha1.Proof rollup_ids_proof = 5;
```
noot, line 52: no `rollup_ids_root`?

crates/astria-proto/src/native/sequencer/v1alpha1.rs
noot: linking #602

crates/astria-proto/proto/astria/sequencer/v1alpha1/celestia.proto
SuperFluffy: Can be reconstructed ad hoc from the list of IDs.

In fact one needs to do this always anyways to ensure the stored is correct. But then we can just skip it.

See the comment and the constructor of the idiomatic native type.

crates/astria-proto/src/native/sequencer/v1alpha1.rs
```diff
@@ -1251,6 +1266,1158 @@ impl Display for IncorrectAddressLength {
...
+/// converting its opaque `data` bytes into sequencer specific types.
+#[derive(Clone, Debug, PartialEq)]
+pub struct SequencerBlock {
+    /// The result of hashing `header`. Guaranteed to not be `None` as compared to
+    /// the cometbft/tendermint-rs return type.
+    block_hash: [u8; 32],
+    /// The original `CometBFT` header that was the input to this sequencer block.
+    header: tendermint::block::header::Header,
+    /// The collection of rollup transactions that were included in this block.
+    rollup_transactions: IndexMap<RollupId, Vec<Vec<u8>>>,
```
noot, line 1489: how does `IndexMap`'s ordering work compared to `BTreeMap`? assuming it's different, that's a breaking change

crates/astria-proto/src/native/sequencer/v1alpha1.rs
```diff
@@ -1251,6 +1266,1158 @@ impl Display for IncorrectAddressLength {
...
+    // The proof that the rollup transactions are included in the `CometBFT` block this
+    // sequencer block is derived form. This proof together with
+    // `Sha256(MTH(rollup_transactions))` must match `header.data_hash`.
```
noot, line 1492: what is `MTH`? merkle tree hash?

crates/astria-proto/src/native/sequencer/v1alpha1.rs
```diff
@@ -1251,6 +1266,1158 @@ impl Display for IncorrectAddressLength {
...
+    // `MTH(rollup_transactions)` is the Merkle Tree Hash derived from the
+    // rollup transactions.
+    rollup_transactions_proof: merkle::Proof,
+    // The proof that the rollup IDs listed in `rollup_transactions` are included
+    // in the `CometBFT` block this sequencer block is derived form. This proof together
+    // with `Sha256(MTH(rollup_ids))` must match `header.data_hash`.
+    // `MTH(rollup_ids)` is the Merkle Tree Hash derived from the rollup IDs listed in
+    // the rollup transactions.
+    rollup_ids_proof: merkle::Proof,
```
noot, line 1501: same question as on the proto - how come the proofs are here but the commitments aren't? is it because you plan to always reconstruct the `rollup_transactions_root` from `rollup_transactions`?

crates/astria-proto/proto/astria/sequencer/v1alpha1/block.proto
```diff
@@ -3,21 +3,41 @@ syntax = "proto3";
...
+    // The proof that the rollup IDs listed in `rollup_transactions` are included
+    // in the CometBFT block this sequencer block is derived form.
+    //
+    // This proof is used to verify that the relayer that posts to celestia
+    // includes all rollup IDs and does not censor any.
+    //
+    // This proof together with `Sha256(MTH(rollup_ids))` must match `header.data_hash`.
+    // `MTH(rollup_ids)` is the Merkle Tree Hash derived from the rollup IDs listed in
+    // the rollup transactions.
+    astria.sequencer.v1alpha1.Proof rollup_ids_proof = 4;
```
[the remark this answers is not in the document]
noot: ok, i see you are planning to just reconstruct the commitments everywhere

crates/astria-proto/src/native/sequencer/v1alpha1.rs
SuperFluffy: This new grouping mechanism is used inside sequencer as well. You can see that the snapshot test is unaffected.

When the indexmap is constructed it is ordered by its keys after the fact. But this is the same order as that used for the btreemap.

crates/astria-proto/src/native/sequencer/v1alpha1.rs
SuperFluffy: Yeah, but this was always done anyway to ensure these data types uphold invariants and that the stored root and the data it is constructed from is correct.

But it need not be written to the wire format.

If it ever becomes necessary to give repeated access to this root we can add it to the idiomatic types as is done for the cometbft header block hash.

crates/astria-proto/src/native/sequencer/v1alpha1.rs
```diff
@@ -1251,6 +1266,1158 @@ impl Display for IncorrectAddressLength {
...
+        Ok(Self {
+            block_hash,
+            header,
+            rollup_transactions,
+            rollup_transactions_proof,
+            rollup_ids_proof,
+        })
+    }
+
+    /// Converts from the raw decoded protobuf representatin of this type.
```
noot, line 1702:
```suggestion
    /// Converts from the raw decoded protobuf representation of this type.
```

crates/astria-proto/src/native/sequencer/v1alpha1.rs
```diff
@@ -1251,6 +1266,1158 @@ impl Display for IncorrectAddressLength {
...
+            rollup_transactions,
+            rollup_transactions_proof,
+            rollup_ids_proof,
+        } = raw;
+
+        let rollup_transactions_proof = 'proof: {
+            let Some(rollup_transactions_proof) = rollup_transactions_proof else {
+                break 'proof Err(SequencerBlockError::FieldNotSet(
+                    "rollup_transactions_proof",
+                ));
```
noot, line 1725: why the `'proof` syntax over just returning the error?

SuperFluffy: There are two steps in this block: one ensures the protobuf value was set, the next one actually does the conversion.

These two steps belong together logically and I didn't want to declare variables at the function level namespace.

I also think this is neat.

crates/astria-proto/src/native/sequencer/v1alpha1.rs
noot, line 2391:
```suggestion
pub fn group_sequence_actions_in_signed_transactions_by_rollup_id(
```

crates/astria-proto/src/native/sequencer/v1alpha1.rs
```diff
@@ -1251,6 +1266,1158 @@ impl Display for IncorrectAddressLength {
...
+         provided: {0}"
+    )]
+    SequencerBlockHash(usize),
+}
+
+/// A shadow of [`CelestiaRollupBlob`] with public access to all its fields.
+///
+/// At the moment there are no invariants upheld by [`CelestiaRollupBlob`] so
+/// they can be converted directly into one another. This can change in the future.
+pub struct UncheckedCelestiaRollupBlob {
```
joroshiba, line 1840: Why the "unchecked" verbiage here? There is nothing checked when converting? It's just pub access vs not?

crates/astria-sequencer-types/src/tendermint.rs
```diff
@@ -1,24 +1 @@
-// Calculates the `last_commit_hash` given a Tendermint [`Commit`].
```
noot, line 1: delete this file?

crates/astria-conductor/src/conductor.rs
```diff
@@ -99,12 +95,26 @@ impl Conductor {
 
         let signals = spawn_signal_handler();
 
+        let rollup_id =
+            proto::native::sequencer::v1alpha1::RollupId::from_unhashed_bytes(&cfg.chain_id);
```
noot, line 99: rename the config value also to `rollup_id`

crates/astria-proto/proto/astria/sequencer/v1alpha1/block.proto
SuperFluffy: The order is important for constructing the Merkle Hash Tree of the roll-up transactions.

The order in this sequencer block is the canonical order used by the sequencer.

Maps vs lists aside, I'm not even sure that we can rely on an order on bytes in other languages. So this is cleaner.

crates/astria-conductor/src/conductor.rs
```diff
@@ -168,11 +178,14 @@ impl Conductor {
                     .await
                     .wrap_err("failed to get sequencer namespace")?
             };
-            info!(
-                celestia_namespace = %Base64Display::new(sequencer_namespace.as_bytes(), &STANDARD),
-                sequencer_chain_id = %cfg.chain_id,
-                "celestia namespace derived from sequencer chain id",
-            );
+            // XXX: This log is very misleading. The field called `sequencer_chain_id` is actually
```
noot, line 181: yeah, do you wanna delete/update it?

crates/astria-celestia-client/src/client.rs
```diff
@@ -213,95 +233,111 @@ pub enum BlobAssemblyError {
...
-/// Filters out blobs that cannot be deserialized to `RollupNamespaceData`,
-/// whose block hash does not match that of `sequencer_data` or that
-/// have the wrong namespace.
-fn filter_and_convert_rollup_data_blobs(
-    blobs: &[Blob],
+/// Attempts to convert the bytes stored in the celestia blobs to [`CelestiaRollupBlob`].
+///
+/// Drops a blob under the following conditions:
+/// + the blob's namespace does not match the provided [`Namespace`]
+/// + cannot be decode/convert to [`CelestiaRollupBlob`]
```
noot, line 270:
```suggestion
/// + cannot decode/convert to [`CelestiaRollupBlob`]
```

crates/astria-celestia-client/src/client.rs
```diff
@@ -213,95 +233,111 @@ pub enum BlobAssemblyError {
...
+/// + block hash does not match that of [`CcelestiaSequencerBlob`]
+/// + the proof, ID, and transactions recorded in the blob cannot be verified against the seuencer
```
noot, line 272:
```suggestion
/// + the proof, ID, and transactions recorded in the blob cannot be verified against the sequencer
```

noot (approved):

crates/astria-proto/src/native/sequencer/v1alpha1.rs
SuperFluffy: BtreeMap is balancing while it's being constructed. That gives nlogn overhead. Insertion the indexmap is linear and the final ordering has the same complexity.

Access on the other hand is O(1) where it is log n for the btreemap.

The indexmap is cache efficient, the btreemap is not.

So for our purpose the index map is strictly better.

crates/astria-proto/src/native/sequencer/v1alpha1.rs
noot:
```
break 'proof Err(SequencerBlockError::FieldNotSet(
   "rollup_transactions_proof",
));
```

this is logically the same as:

```
return Err(SequencerBlockError::FieldNotSet(
   "rollup_transactions_proof",
));
```

right?  just checking for my own sanity lol

crates/astria-proto/src/native/sequencer/v1alpha1.rs
SuperFluffy: Yeah, thats the case here.

Honestly no reason other than being symmetric with the sequencer blob.

I'm not a big fan of the unchecked versions. But yes, it's to give pub access to the fields to avoid allocations.

joroshiba (approved):

crates/astria-conductor/src/conductor.rs
SuperFluffy: I wasn't sure about this. It's m technically not the ID but just the name. The ID are the 32 bytes and I don't want to cause confusion between these two.

Is rollup_name good? Also CC @joroshiba

PR #397, opened 2023-09-19
fix(ci): conditional workflows fix, fix docker image loading

joroshiba:
## Summary
Updates conditional workflows so they don't always run, and some docker image loading issues

## Background
Some of our tests should only run conditionally, but were running everytime. Additionally docker images pull was experiencing issues with QEMU loading due to docker not being logged in yet.

## Changes
- updates conditionals to evaluate against strings, because outputs are converted to strings
- moves docker login higher up for docker workflows.

## Testing
Many testing commits in this PR to evaluate what was going on and get fixed. Will manually test again after merge.


steezeburger (dismissed):

noot (dismissed):

.github/workflows/release.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     environment: BUF
     needs: run_checker
-    if: ${{ needs.run_checker.outputs.run_release_proto }} && github.repository_owner == 'astriaorg'
+    if: ${{ needs.run_checker.outputs.run_release_proto && github.repository_owner == 'astriaorg' }}
```
SuperFluffy, line 16: Wow, that's a good catch. I don't quite get the semantics of this. Does `&&` not work outside of `${{ }}`?

SuperFluffy (dismissed):

.github/workflows/release.yml
joroshiba: This was actually not the issue, but updates fix.

.github/workflows/release.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     environment: BUF
     needs: run_checker
-    if: ${{ needs.run_checker.outputs.run_release_proto }} && github.repository_owner == 'astriaorg'
+    if: needs.run_checker.outputs.run_release_proto && github.repository_owner == 'astriaorg'
```
SuperFluffy, line 16: The first conditional need not be `== 'true'` like in the other `if`s of this patch?

EDIT: I think this might. In a similar && condition you have == further down

.github/workflows/reusable-docker-build.yml
```diff
@@ -25,23 +25,23 @@ jobs:
...
         with:
           registry: ${{ env.REGISTRY }}
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
-      - name: Login to Docker Hub
-        uses: docker/login-action@v2
-        with:
-          username: ${{ secrets.DOCKER_USER }}
-          password: ${{ secrets.DOCKER_TOKEN }}
+      - name: Set up QEMU
```
SuperFluffy, line 39: Orthogonal to this PR: I believe this is no longer necessary

steezeburger (approved):

SuperFluffy (approved):
Looks good. I think one == needs to be added

PR #400, opened 2023-09-19
Feature/conductor execution api v1alpha2 integration

steezeburger:
## Summary
This PR is the integration PR for Execution API v1alpha2. The logic has been reviewed and approved, but there may be some Rust improvements to be made.

## Background
v1alpha1 was not sufficient for our needs, so v1alpha2 was created

## Changes
- removes v1alpha1 api
- adds v1alpha2 api
- implements functionality to use v1alpha2

## Testing
- fixes existing tests
- added new test for scenario where we get an already executed block from da

## Breaking Changelist
- v1alpha1 api methods removed

## Related Issues
https://github.com/astriaorg/astria/issues/205
https://github.com/astriaorg/astria/issues/207

closes https://github.com/astriaorg/astria/issues/205
closes https://github.com/astriaorg/astria/issues/207


crates/astria-conductor/src/execution_client.rs
```diff
@@ -13,16 +17,26 @@ use tracing::info;
 
 #[async_trait::async_trait]
 pub(crate) trait ExecutionClient: crate::private::Sealed {
```
noot, line 19: since this trait is `pub(crate)` you can remove the `Sealed` trait

SuperFluffy: @noot this is already done in the single-entry-point PR

crates/astria-conductor/src/executor.rs
```diff
@@ -156,12 +161,23 @@ impl<C: ExecutionClient> Executor<C> {
                         );
                         continue;
                     };
-                    if let Err(e) = self.execute_block(block_subset).await {
-                        error!(
-                            height = height,
-                            error = ?e,
-                            "failed to execute block"
-                        );
+                    match self.execute_block(block_subset.clone()).await {
```
noot, line 164:
```suggestion
                    match self.execute_block(block_subset).await {
```

why are we cloning now? looks unnecessary

crates/astria-conductor/src/executor.rs
```diff
@@ -156,12 +161,23 @@ impl<C: ExecutionClient> Executor<C> {
...
+                        Ok(Some(executed_block)) => {
+                            // set soft but leave firm the same
+                            self.update_commitment_state(
+                                self.commitment_state.firm.clone().unwrap(),
+                                executed_block,
+                            )
+                            .await?;
```
noot, line 171:
```suggestion
                            .await.wrap_err("failed to update commitment state")?;
```

noot: actually, i think returning an error here stops the whole event loop and thus the entire conductor - is that what we want, or should we just log here? the error below is just logged, so maybe we do that?

crates/astria-conductor/src/executor.rs
```diff
@@ -210,16 +226,19 @@ impl<C: ExecutionClient> Executor<C> {
...
 
-        if let Some(execution_hash) = self.sequencer_hash_to_execution_hash.get(&block.block_hash) {
+        if let Some(execution_block) = self
+            .sequencer_hash_to_execution_block
+            .get(&block.block_hash)
+        {
             debug!(
                 height = block.header.height.value(),
-                execution_hash = hex::encode(execution_hash),
+                execution_hash = hex::encode(execution_block.clone().hash),
```
noot, line 235: can you just clone the hash instead of the block?

crates/astria-conductor/src/executor.rs
```diff
@@ -210,16 +226,19 @@ impl<C: ExecutionClient> Executor<C> {
...
                 "block already executed"
             );
-            return Ok(Some(execution_hash.clone()));
+            return Ok(Some(execution_block.clone()));
         }
 
-        let prev_block_hash = self.execution_state.clone();
+        let prev_block_hash = self.commitment_state.soft.clone().unwrap().hash;
```
noot, line 241: no `unwrap` pls

crates/astria-conductor/src/executor.rs
```diff
@@ -229,23 +248,37 @@ impl<C: ExecutionClient> Executor<C> {
...
+    async fn update_commitment_state(&mut self, firm: Block, soft: Block) -> Result<()> {
+        let commitment_state = self
+            .execution_rpc_client
+            .call_update_commitment_state(CommitmentState {
+                firm: Some(firm),
+                soft: Some(soft),
+            })
+            .await?;
+        self.commitment_state = commitment_state;
+        Ok(())
```
noot, line 281:
```suggestion
       self.commitment_state = self
            .execution_rpc_client
            .call_update_commitment_state(CommitmentState {
                firm: Some(firm),
                soft: Some(soft),
            })
            .await?
```

nit

crates/astria-conductor/src/executor.rs
```diff
@@ -254,13 +287,21 @@ impl<C: ExecutionClient> Executor<C> {
...
-                    .await?;
+            Some(executed_block) => {
+                // this case means block has already been executed.
+                // setting execution chain's FIRM.
+                self.update_commitment_state(
+                    executed_block,
+                    self.commitment_state.soft.clone().unwrap(),
+                )
+                .await?;
+                // remove the sequencer block hash from the map, as it's been executed
```
noot, line 302:
```suggestion
                // remove the sequencer block hash from the map, as it's been firmly committed
```

crates/astria-conductor/src/executor.rs
```diff
@@ -272,50 +313,25 @@ impl<C: ExecutionClient> Executor<C> {
...
-                    debug!("execute_block returned None; skipping finalize_block");
+                    debug!("execute_block returned None; skipping call_update_commitment_state");
                     return Ok(());
                 };
-
-                // finalize the block after it's been executed
-                self.finalize_block(execution_block_hash, sequencer_block_hash)
+                // when we execute a block received from da, nothing else has been executed on top
+                // of it, so we set FIRM and SOFT to this executed block
+                self.update_commitment_state(executed_block.clone(), executed_block)
```
noot, line 327: this looks weird - i also noticed we are usually only updating `soft` OR `firm`, not both at the same time (except for here), so i'd add functions `update_soft_commitment` and `update_firm_commitment` to make it clear which is being updated

crates/astria-conductor/src/execution_client.rs
SuperFluffy, line 1: High level comment on this module: right now this is basically defining a newtype wrapper around `ExecutionServiceClient` + a trait `ExecutionClient` that is then impl'd on the wrapper. The wrapper itself carries no further state.

I think this PR touches so many bits that it's justified to either:

1. impl the trait methods as inherent methods on the wrapper (without th trait)
2. or impl `ExecutionClient` as an extension trait on `ExecutionServiceClient` (without th wrapper)

crates/astria-conductor/src/executor.rs
SuperFluffy: I think cloning in general is unnecessary. tendermint's `Hash` impls `AsRef<[u8]>` and `hex::encode` is defined for `T: AsRef<[u8]>`

steezeburger: `execution_block.hash` is a `Vec<u8>` here

SuperFluffy: Doesn't change the argument tho: https://doc.rust-lang.org/std/vec/struct.Vec.html#impl-AsRef%3C%5BT%5D%3E-for-Vec%3CT,+A%3E :-)

crates/astria-conductor/src/executor.rs
steezeburger: updated! to avoid nesting, i refactored the match into some `if lets`. please lmk if this is better @noot

crates/astria-conductor/src/execution_client.rs
steezeburger: updated! i went with option 2. lmk if this is what you intended! @SuperFluffy

joroshiba:
High level LGTM, reviewed all the previous PRs into this feature branch.

I think we should block merging this on ensuring it works with [astriaorg/go-ethereum#](https://github.com/astriaorg/go-ethereum/pull/29). Hopefully this "just works", but we should do some testing to make sure. May require changes to both. 

I want to avoid ending up in a state where where we can't cut a new conductor release because there isn't a geth image that works with it (especially with other changes currently going in).

crates/astria-conductor/src/executor.rs
```diff
@@ -210,16 +228,25 @@ impl<C: ExecutionClient> Executor<C> {
...
-            return Ok(Some(execution_hash.clone()));
+            return Ok(Some(execution_block.clone()));
         }
 
-        let prev_block_hash = self.execution_state.clone();
+        let prev_block_hash = if let Some(soft_commitment) = self.commitment_state.soft.clone() {
+            soft_commitment.hash
+        } else {
+            warn!("could not get previous block. soft commitment is None");
+            return Ok(None);
```
noot, line 247: is there ever a case this could happen? seems concerning if there is, as nothing can get executed

crates/astria-conductor/src/executor.rs
```diff
@@ -229,23 +256,64 @@ impl<C: ExecutionClient> Executor<C> {
...
+        Ok(Some(executed_block))
+    }
+
+    /// Updates the commitment state on the execution layer.
+    /// Updates the local commitment_state with the new values.
+    async fn update_commitment_state(&mut self, commitment_state: CommitmentState) -> Result<()> {
+        let new_commitment_state = self
+            .execution_rpc_client
+            .call_update_commitment_state(commitment_state)
+            .await?;
```
noot, line 284: wrap_err

crates/astria-conductor/src/executor.rs
steezeburger: I don't think there is a case it would ever occur. That type comes from the proto definitions. I think I misunderstood your comment and added this complex handling instead of just using `wrap_err` haha. Should I just use `wrap_err` here?

noot (approved):
looks good! like @joroshiba said we should probably test w/ geth before merging

crates/astria-conductor/src/executor.rs
noot: maybe return an error here - looking at the code, it seems this case shouldn't ever occur, so would be developer error if it does

steezeburger: It should not ever happen no. What's a better way I can handle the `Option` here?

crates/astria-conductor/src/execution_client.rs
```diff
@@ -1,88 +1,128 @@
...
 use prost_types::Timestamp;
 use tonic::transport::Channel;
 use tracing::info;
 
+/// Creates a new RPC Client for Execution
+///
+/// # Arguments
+///
+/// * `address` - The address of the RPC server that we want to communicate with.
+pub(crate) async fn new(address: &str) -> Result<ExecutionServiceClient<Channel>> {
```
SuperFluffy, line 23: Small nit: I think I would personally just use `ExecutionServiceClient::connect` at the callsite and forego this extra wrapper.

crates/astria-conductor/src/execution_client.rs
```diff
@@ -1,88 +1,128 @@
...
+    let client = ExecutionServiceClient::connect(address.to_owned()).await?;
+    info!("Connected to execution service at {}", address);
+    Ok(client)
+}
+
 #[async_trait::async_trait]
 pub(crate) trait ExecutionClient {
```
SuperFluffy, line 30: Small nit: might call this `ExecutionClientExt` to make it clear this is an extension trait, but no big deal. Many other crates in the ecosystem would also just call this `Client` without `Ext`.

crates/astria-conductor/src/execution_client.rs
```diff
@@ -1,88 +1,128 @@
...
-        })
+    /// * `identifiers` - List of block identifiers describes which blocks we want to get
+    async fn call_batch_get_blocks(
+        &mut self,
+        identifiers: Vec<BlockIdentifier>,
+    ) -> Result<BatchGetBlocksResponse> {
+        let request = BatchGetBlocksRequest {
+            identifiers,
+        };
+        let response = self.batch_get_blocks(request).await?.into_inner();
```
SuperFluffy, line 67: error context, i.e. `await.wrap_err(...)`

crates/astria-conductor/src/execution_client.rs
```diff
@@ -1,88 +1,128 @@
...
-    ) -> Result<DoBlockResponse> {
-        let request = DoBlockRequest {
+    ) -> Result<Block> {
+        let request = ExecuteBlockRequest {
             prev_block_hash,
             transactions,
             timestamp,
         };
-        let response = self.client.do_block(request).await?.into_inner();
+        let response = self.execute_block(request).await?.into_inner();
```
SuperFluffy, line 89: error context

crates/astria-conductor/src/execution_client.rs
```diff
@@ -1,88 +1,128 @@
...
+    /// # Arguments
+    ///
+    /// * `identifier` - The identifier describes the block we want to get
+    async fn call_get_block(&mut self, identifier: BlockIdentifier) -> Result<Block> {
+        let request = GetBlockRequest {
+            identifier: Some(identifier),
         };
-        self.client.finalize_block(request).await?;
-        Ok(())
+        let response = self.get_block(request).await?.into_inner();
```
SuperFluffy, line 102: error context :)

crates/astria-conductor/src/execution_client.rs
```diff
@@ -1,88 +1,128 @@
...
+        Ok(response)
+    }
+
+    /// Calls remote procedure GetCommitmentState
+    async fn call_get_commitment_state(&mut self) -> Result<CommitmentState> {
+        let request = GetCommitmentStateRequest {};
+        let response = self.get_commitment_state(request).await?.into_inner();
```
SuperFluffy, line 109: error context

crates/astria-conductor/src/execution_client.rs
```diff
@@ -1,88 +1,128 @@
...
+    ///
+    /// * `commitment_state` - The CommitmentState to set, must include complete state.
+    async fn call_update_commitment_state(
+        &mut self,
+        commitment_state: CommitmentState,
+    ) -> Result<CommitmentState> {
+        let request = UpdateCommitmentStateRequest {
+            commitment_state: Some(commitment_state),
+        };
+        let response = self.update_commitment_state(request).await?.into_inner();
```
SuperFluffy, line 125: you guessed it: error context :P

crates/astria-conductor/src/executor.rs
```diff
@@ -104,34 +107,35 @@ struct Executor<C> {
...
+    sequencer_hash_to_execution_block: HashMap<Hash, Block>,
 }
 
 impl<C: ExecutionClient> Executor<C> {
     async fn new(mut execution_rpc_client: C, chain_id: ChainId) -> Result<(Self, Sender)> {
         let (cmd_tx, cmd_rx) = mpsc::unbounded_channel();
-        let init_state_response = execution_rpc_client.call_init_state().await?;
-        let execution_state = init_state_response.block_hash;
+
+        let commitment_state = execution_rpc_client.call_get_commitment_state().await?;
```
SuperFluffy, line 129: error context

crates/astria-conductor/src/executor.rs
SuperFluffy: It looks like executor is handling too many of the protobuf idiosyncracies like `soft` and `hard` being `Option`. If it basically never makes sense for these fields to be unset, then it's better to have one extra step that converts from the `protobuf-CommitmentState` to an `idiotmatic-CommitmentState` (failing if one of the fields is not set).

This way you don't have to worry about all of that `None` business.

sambukowski (approved):

joroshiba:
#528 merged instead, contained conflict updates.

PR #405, opened 2023-09-20
feat(sequencer): implement mempool nonce check

### LINKED ISSUE #361: sequencer: add stateful nonce check in `check_tx`
currently, we have a `check_stateful` method on txs which is called before execution. `check_tx` (called to check tx before putting in mempool) only calls `check_stateless`. however, we do want to perform a stateful check as well, namely the nonce check (nonce is not below current). we should add a check for this. it already occurs in `check_stateful` but we should pull it out to a separate function.

noot:
## Summary
implement nonce check before putting tx in mempool 

## Background
`check_tx` (called to check tx before putting in mempool) previously only called `check_stateless`. however, we do want to perform a stateful check as well, namely the nonce check (nonce is not below current).

## Changes
- add check in `check_tx` that tx nonce is not below current account nonce, as that would instantly invalidate it

## Related Issues

closes #361


crates/astria-sequencer/src/service/mempool.rs
```diff
@@ -79,6 +98,18 @@ fn handle_check_tx(req: request::CheckTx) -> response::CheckTx {
     // TODO(https://github.com/astriaorg/astria/issues/228): status codes for various errors
     // TODO(https://github.com/astriaorg/astria/issues/319): offload `check_stateless` using `deliver_tx_bytes` mechanism
     //       and a worker task similar to penumbra
+    match transaction::check_nonce_mempool(&signed_tx, &state).await {
+        Ok(_) => (),
```
SuperFluffy, line 102: Replace with `if let Err(e) = etcetc.await {` if you re throwing away the ok.

crates/astria-sequencer/src/service/mempool.rs
```diff
@@ -79,6 +98,18 @@ fn handle_check_tx(req: request::CheckTx) -> response::CheckTx {
...
+        Err(e) => {
+            return response::CheckTx {
+                code: AbciCode::INVALID_PARAMETER.into(),
```
SuperFluffy, line 105: Should this be an `INVALID_NONCE` introduced in #364 ?

crates/astria-sequencer/src/service/mempool.rs
```diff
@@ -79,6 +98,18 @@ fn handle_check_tx(req: request::CheckTx) -> response::CheckTx {
...
+                info: "failed verifying nonce of SignedTransaction".into(),
```
SuperFluffy, line 106: I think we can leave out language (and implementation) specific names/types etc
```rust
                info: "failed verifying nonce of transaction".into(),
```

crates/astria-sequencer/src/transaction/mod.rs
```diff
@@ -18,6 +18,19 @@ use crate::accounts::state_ext::{
...
 };
 
+pub(crate) async fn check_nonce_mempool<S: StateReadExt + 'static>(
+    tx: &SignedTransaction,
+    state: &S,
+) -> anyhow::Result<()> {
+    let signer_address = Address::from_verification_key(tx.verification_key());
+    let curr_nonce = state.get_account_nonce(signer_address).await?;
+    ensure!(
+        tx.unsigned_transaction().nonce < curr_nonce,
```
SuperFluffy, line 28: Is this sufficient? Or would `==` be too strict?

crates/astria-sequencer/src/transaction/mod.rs
```diff
@@ -18,6 +18,19 @@ use crate::accounts::state_ext::{
     StateWriteExt,
 };
 
+pub(crate) async fn check_nonce_mempool<S: StateReadExt + 'static>(
+    tx: &SignedTransaction,
+    state: &S,
+) -> anyhow::Result<()> {
+    let signer_address = Address::from_verification_key(tx.verification_key());
+    let curr_nonce = state.get_account_nonce(signer_address).await?;
```
SuperFluffy, line 26: context

crates/astria-sequencer/src/transaction/mod.rs
```diff
@@ -18,6 +18,19 @@ use crate::accounts::state_ext::{
...
+    ensure!(
+        tx.unsigned_transaction().nonce < curr_nonce,
+        "nonce already used by account"
+    );
+    Ok(())
+}
+
 pub(crate) fn check_stateless(tx: &SignedTransaction) -> anyhow::Result<()> {
     tx.unsigned_transaction().check_stateless()
```
SuperFluffy, line 35: should still add context, even if that's the only fn

crates/astria-sequencer/src/service/mempool.rs
noot: yes! thanks

PR #413, opened 2023-09-21
feat(ci): Cut new releases, and autocut moving forward

joroshiba:
## Summary
Adds reusable workflow which ties image releases to `Cargo.toml` files, and creates releases on merge with tags. 

## Background
Right now the release workflow is "create a tag". We updated the `Cargo.toml` to match these versions, but there was nothing enforcing it. The release flow here notifies that a release will be cut, and stops us from accidentally duplicating. It also will create a tag only after merge, and a release to go with it (which will spark docker build).

## Changes
- Update cargo.toml files to cut new releases.
- Added actions to cut releases.

## Testing
Github actions, so manually and some can only be tested after merge.


WafflesVonMaple (approved):
👍

steezeburger (approved):
pretty slick!

Issue #369, opened 2023-09-14
sequencer: define ABCI code responses as protos


itamarreif:
ABCI codes are currently defined as a Rust module in the `astria-sequencer` crate. As these responses are required by services interacting with the sequencer (such as `composer` and `conductor`) they should be treated as a public interface and defined in `astria-proto` as protobufs.

PR #414, opened 2023-09-21
move out abci codes to astria-sequencer-types crate

itamarreif:
## Summary
This moves out the `AbciCode` type from `astria-sequencer` to `astria-sequencer-types` so that the entire sequencer crate does not need to be added as a dependency in order to handle sequencer responses.

## Background
The `Composer` needs to access these types in order to handle failed transaction submission.

## Changes
- Remove `AbciCode` from `astria-sequencer` and add it to `astria-sequencer-types`
- Fix the failed import statements that result

## Related Issues
This is required for [#293](https://github.com/astriaorg/astria/issues/293) and partially addresses [#369](https://github.com/astriaorg/astria/issues/369) (though more work needs to be done to move this out to protos and implement rust types on top)


noot (approved):
does this close #369?

crates/astria-sequencer-types/src/abci_codes.rs
SuperFluffy, line 1: I would rename this module `abci_code` to match the type `AbciCode` (convention)

SuperFluffy (approved):
Looks good and makes sense.

PR #415, opened 2023-09-21
fix(sequencer): fix config

noot:
## Summary

`ASTRIA_SEQUENCER` should be `ASTRIA_SEQUENCER_` - also small doc updates

## Testing
i ran it

### DIFF
--- a/crates/astria-sequencer/README.md
+++ b/crates/astria-sequencer/README.md
@@ -69,10 +69,8 @@ make install_abci
 In astria-sequencer/:
 
 ```sh
-cargo build
-../../target/debug/astria-sequencer
+just run
 ```
-
 
 ### Query the app for info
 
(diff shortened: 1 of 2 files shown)


SuperFluffy (approved):
We should have a test for this.

sambukowski (approved):

crates/astria-sequencer/src/service/mempool.rs
```diff
@@ -74,7 +74,18 @@ fn handle_check_tx(req: request::CheckTx) -> response::CheckTx {
             };
         }
     };
-    let signed_tx = SignedTransaction::try_from_raw(raw_signed_tx).unwrap();
```
SuperFluffy, line 77: This is really orthogonal to the rest...

noot: should i add to a different PR? sorry i was just running some tests and also found this :p

SuperFluffy: Yeah, that would be much cleaner. 🙏

noot: https://github.com/astriaorg/astria/pull/418

PR #418, opened 2023-09-21
fix(sequencer): no unwrap in `check_tx`

noot:
## Summary
remove unwrap, caused sequencer to panic on invalid proto SignedTransaction

## Testing
i ran this https://github.com/astriaorg/go-sequencer-client


crates/astria-sequencer/src/service/mod.rs
```diff
@@ -19,6 +19,7 @@ impl AbciCode {
     pub(crate) const INVALID_PARAMETER: Self = Self(2);
     pub(crate) const INTERNAL_ERROR: Self = Self(3);
     pub(crate) const INVALID_NONCE: Self = Self(4);
+    pub(crate) const INVALID_SIGNATURE: Self = Self(5);
```
SuperFluffy, line 22: Does this capture all errors that can occur above?

This should have a clearer mapping from error code to abci code before I'm afraid the error log might not match the code.

Maybe leave it as an internal error for now?

crates/astria-sequencer/src/service/mempool.rs
```diff
@@ -74,7 +74,18 @@ fn handle_check_tx(req: request::CheckTx) -> response::CheckTx {
             };
         }
     };
-    let signed_tx = SignedTransaction::try_from_raw(raw_signed_tx).unwrap();
+    let signed_tx = match SignedTransaction::try_from_raw(raw_signed_tx) {
+        Ok(tx) => tx,
+        Err(e) => {
+            return response::CheckTx {
+                code: AbciCode::INVALID_SIGNATURE.into(),
+                info: "failed to verify signed transaction".into(),
```
SuperFluffy, line 82: This is set from the abci code description.

I believe thats the intended use of the codes.

More context is then provided through the logs

SuperFluffy (approved):
Approval for the fix but I have comments on the response.

Should there ever be more errors transforming the raw block the code/info will inadvertently become out of whack

crates/astria-sequencer/src/service/mod.rs
noot: yeah i guess there are other errors that could occur like the fields being the wrong length. i'll update

crates/astria-sequencer/src/service/mempool.rs
noot: updated to return the error as well

PR #424, opened 2023-09-22
fix tendermint block to SequencerBlockData conversion

noot:
## Summary
fix tendermint block to SequencerBlockData conversion by fixing `data_hash` calculation, txs need to be hashed before `data_hash` is calculated

## Background 
our data_hash calculation was wrong, which caused the type conversion to always fail. also moved a test which was previously in relayer which tested this to astria-sequencer-types


crates/astria-sequencer-types/src/sequencer_block_data.rs
```diff
@@ -328,7 +328,8 @@ impl SequencerBlockData {
         }
 
         // generate the action tree root proof of inclusion in `Header::data_hash`
-        let tx_tree = MerkleTree::from_leaves(b.data);
+        let hashed_txs = b.data.iter().map(|tx| sha256_hash(tx)).collect::<Vec<_>>();
```
SuperFluffy: In a follow-up the merkle tree should be generalized to take `[u8; 32]` to avoid all the (now unnecessary) allocations.

SuperFluffy (approved):
Please provide a more extensive PR description for why this is necessary (background, etc).

Is this because this does not match how cometbft is constructing the proof?

If so, don't we have a test for that? Why did it not catch it?

If not, can we create a test?

crates/astria-sequencer-types/src/sequencer_block_data.rs
```diff
@@ -514,4 +527,24 @@ mod test {
...
     }
+
+    #[test]
+    fn calculate_data_hash() {
+        use base64::{
+            engine::general_purpose::STANDARD,
+            Engine as _,
+        };
+
+        // data_hash is calculated from the txs in a block, where the leaves of the merkle tree are
```
SuperFluffy, line 538: Please add the steps you took too generate this hash and which versions so that somebody else could generate this (or a similar) magic number

noot:
@SuperFluffy there was a test before in relayer, but it was not moved to sequencer-types. basically our `data_hash` calculation was wrong, which caused the type conversion to always fail

SuperFluffy (commented):
More comments on the commit after the approval was submitted

SuperFluffy:
> @SuperFluffy there was a test before in relayer, but it was not moved to sequencer-types. basically our `data_hash` calculation was wrong, which caused the type conversion to always fail

Ok, I understand - but can you add that to the PR text so that it gets immortalized in the commit tree?

crates/astria-sequencer-types/src/sequencer_block_data.rs
noot: added!

joroshiba (approved):

PR #426, opened 2023-09-22
fix(conductor): continue to execution when block subset empty

joroshiba:
## Summary
Allow empty blocks to make it to executor

## Background
We added a setting for executor to execute empty blocks in astriaorg/astria#373, however we didn't fix that run short circuits before getting there

## Changes
- update to allow blocks with no transactions be converted
- add an error if problem converting to block subset

## Testing
Manually tested with dev-cluster

### DIFF
--- a/crates/astria-conductor/src/executor.rs
+++ b/crates/astria-conductor/src/executor.rs
@@ -162,14 +162,9 @@ impl<C: ExecutionClient> Executor<C> {
                     block,
                 } => {
                     let height = block.header().height.value();
-                    let Some(block_subset) =
-                        SequencerBlockSubset::from_sequencer_block_data(*block, &self.chain_id)
-                    else {
-                        error!(
-                            "there was an issue deriving rollup block subset from sequencer block"
-                        );
-                        continue;
-                    };
+                    let block_subset =
+                        SequencerBlockSubset::from_sequencer_block_data(*block, &self.chain_id);
+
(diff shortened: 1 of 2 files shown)


crates/astria-conductor/src/types.rs
```diff
@@ -34,7 +34,11 @@ impl SequencerBlockSubset {
             ..
         } = data.into_raw();
 
-        let rollup_transactions = rollup_data.remove(chain_id)?;
+        let rollup_transactions = match rollup_data.remove(chain_id) {
+            Some(txs) => txs,
+            None => vec![],
+        };
```
noot, line 40:
```suggestion
        let rollup_transactions = match rollup_data.remove(chain_id).unwrap_or_default();
```

crates/astria-conductor/src/executor.rs
```diff
@@ -165,9 +165,8 @@ impl<C: ExecutionClient> Executor<C> {
                     let Some(block_subset) =
                         SequencerBlockSubset::from_sequencer_block_data(*block, &self.chain_id)
                     else {
-                        info!(
-                            namespace = %self.namespace,
-                            "block did not contain data for namespace; skipping"
+                        error!(
+                            "there was an issue deriving rollup block subset from sequencer block"
                         );
                         continue;
```
noot, line 171: `from_sequencer_block_data` does not need to return an option anymore, it always returns `Some`

noot (approved):

PR #428, opened 2023-09-24
wip: Proc-macro for generating config

### LINKED ISSUE #398: Pull out services' configs to its own crate `astria-config`
After doing the work for https://github.com/astriaorg/astria/pull/387, we realized it's a good idea to pull out configuration into its own crate `astria-config`

### LINKED ISSUE #416: config: ensure correct prefix is set
#415 fixes the sequencer prefix being incorrectly set. This should ideally be protected by a test.

dereksione:
## Summary
Introducing a new crate `astria-config` that offers an easy to use proc-macro to convert a struct into a Configuration struct that expects env variable based configs as outlined in the spec

## Background

We have a lot of duplicate boilerplate config setup across composer, conductor, sequencer, sequencer-relayer

## Changes
- Delete all config setup code from the services in favor of using the macro
- Introduce the proc-macro crate astria-config

## Testing

- `cargo expand` when using the macro returns equivalent code to what exists currently
- The old tests are passing as intended 

## Follow up

This basically is a drop in replacement for the current state of config management. However it does not add additional functionality for better error handling and custom deserializing behavior as has been discussed as desirable on engineering team calls. This effort needs to be further scoped out and is left for future PRs

closes #398
closes #416


crates/astria-config/src/lib.rs
```diff
@@ -0,0 +1,133 @@
...
+        impl #struct_name {
+            fn from_environment(env_prefix: &str) -> Result<Config, figment::Error> {
+                figment::Figment::new()
+                    .merge(figment::providers::Env::prefixed("RUST_").split("_").only(&["log"]))
+                    .merge(figment::providers::Env::prefixed(env_prefix))
+                    .extract()
+            }
+        }
+
+        #[cfg(test)]
```
SuperFluffy, line 86: Test should probably never be made part of generated code. Should one of these fail, it's difficult for users to find the problem.

crates/astria-config/src/lib.rs
```diff
@@ -0,0 +1,133 @@
...
+    //     })
+    //     .collect::<Vec<_>>();
+
+    let code_gen_ast = quote! {
+        const ENV_PREFIX: &str = #prefix;
+
+        use serde::{Serialize, Deserialize};
+
+        // Adding serde dependencies on top of the old struct
+        #[derive(Debug, Deserialize, Serialize)]
```
SuperFluffy, line 69: It's better to derive these explicitly, and then instead define a `trait Config: Deserialize + Serialize` that requires these bounds. At the moment this is very magic and surprising to users. How do you discover this?

SuperFluffy (commented):
I initially tried a similar approach but rejected it because it had several issues:

1. test discovery is bad: if one of the tests fail a third party will have a hard time tracking down the error. In general, tests should never be done in generated code because debugging them is not a fun experience.
2. the implicit derivation of `Serialize` and `Deserialize` is surprising. Deriving these traits should probably be left to the user, but then required by creating a derive macro that requires them as bounds. Alternatively one could define another receiver for a struct `Config`, let's call it `PrivConfig`, for which you instead derive the traits.
3. the big issue of which config options are missing and which env vars are not set is not addressed.

Most of the functionality of this proc macro can be implemented with something like this:

```rust
trait Config: Serialize + Deserialize {
    const PREFIX: &'static str;

    fn from_environment() -> Result<Self, figment::Error>
       figment::Figment::new()
            .merge(figment::providers::Env::prefixed("RUST_").split("_").only(&["log"]))
            .merge(figment::providers::Env::prefixed(Self::PREFIX))
            .extract()
}

impl Config for MyConfig {
    const PREFIX: &'static str = "ASTRIA_SERVICE_";
}
```

crates/astria-config/src/lib.rs
```diff
@@ -0,0 +1,133 @@
...
+        .collect::<Vec<_>>()
+        .contains(&"log".to_string())
+    else {
+        panic!("config struct does not contain a log element, which is necessary");
+    };
+
+    // Check if the macro itself has a valid attribute
+    let attr_str = attr.to_string().to_lowercase();
+    match attr_str.as_str() {
+        "conductor" | "composer" | "sequencer" | "sequencer_relayer" => {}
```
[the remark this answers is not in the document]
dereksione: I don't necessarily agree with this - I think this config setup only works for these services because we had tests for these services - if we add services in the future that make use of this macro, I think we should follow a process in which the config is set up vanilla -> and then migrated to this macro to see if it behaves the same and changing the macro as desired. Restricting is like this would be my preference.

crates/astria-config/src/lib.rs
dereksione: I agree with this statement but disagree about these specific tests because they are sanity checks. If these are failing, then they did not follow the spec. I don't think these specific tests being super visible adds to that. I was going to write documentation that basically pasted these so that people can refer to them from the cargo docs. But I think having these tests everywhere is unnecessary clutter.

crates/astria-config/src/lib.rs
dereksione: Yeah that's a great point, I'll change.

crates/astria-config/src/lib.rs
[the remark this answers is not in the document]
dereksione: Also another point to note is that this only follows the config spec for services, not for user facing cli tools for instance. Making this agnostic would be confusing IMO

crates/astria-config/src/lib.rs
dereksione: Update: Had a brief conversation with @SuperFluffy about this, and we came up with a tentative in between solution

crates/astria-config/src/lib.rs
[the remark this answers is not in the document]
dereksione: Update: Spoke with @SuperFluffy briefly, also came up with an in-between solution that doesn't compromise readability, but also is more generalized than this implementation

Cargo.toml
```diff
@@ -69,6 +70,9 @@ tracing-subscriber = "0.3"
 uuid = "1.3.1"
 which = "4.4.0"
 wiremock = "0.5"
+quote = "1.0.33"
+syn = { version = "2.0.37", features = ["extra-traits"] }
```
noot, line 74: nit but make this alphabetical pls

crates/astria-conductor/src/config.rs
```diff
@@ -1,19 +1,9 @@
...
-    Config::from_environment("ASTRIA_CONDUCTOR_")
-}
+use astria_config::astria_config;
 
 /// The global configuration for the driver and its components.
-#[derive(Serialize, Deserialize)]
-#[serde(deny_unknown_fields)]
+#[astria_config(conductor)]
+// #[derive(Debug, Deserialize, Serialize)]
+// #[serde(deny_unknown_fields)]
```
noot, line 6: remove

crates/astria-config/src/lib.rs
```diff
@@ -0,0 +1,133 @@
...
+    let test_prefix = format!("TESTTEST_{}", prefix);
+
+    // let struct_field_ast = struct_fields
+    //     .iter()
+    //     .map(|(id, ty, vis)| {
+    //         quote! {
+    //             #vis #id: #ty
+    //         }
+    //     })
+    //     .collect::<Vec<_>>();
```
noot, line 61: remove

noot (approved):
nice!

crates/astria-utils/src/config.rs
```diff
@@ -0,0 +1,17 @@
...
+    providers::Env as FigmentEnv,
+    Figment,
+};
+use serde::{
+    Deserialize,
+    Serialize,
+};
+
+pub trait AstriaConfig<'a>: Serialize + Deserialize<'a> {
+    fn from_environment(env_prefix: &str) -> Result<Self, figment::Error> {
```
SuperFluffy, line 11: nit: for compatibility you can make this this return a custom `Error` with `figment::Error` being left as an implementation detail, i.e.

```rust
pub struct Error {
    inner: figment::Error,
}

impl Error for Error {
    fn source(&self) -> Option<&dyn std::error::Error + 'static> {
        Some(&self.inner)
    }
}
```

crates/astria-composer/src/config.rs
```diff
@@ -15,22 +12,10 @@ use serde::{
...
-///
-/// An error is returned if the config could not be read.
-pub fn get() -> Result<Config, figment::Error> {
-    Config::from_environment()
-}
-
 /// The high-level config for creating an astria-composer service.
-#[derive(Debug, Deserialize, Serialize)]
+#[derive(Serialize, Deserialize)]
 #[serde(deny_unknown_fields)]
```
SuperFluffy, line 17: The tests in `astria-utils` passing rely on `deny_unknown_fields` being present, right?

It's pretty surprising that that's the case (and honestly, I should have added that `deny_unknown_fields` in the first place). Is there a way to not rely on that?

crates/astria-utils/src/config.rs
```diff
@@ -0,0 +1,17 @@
+use figment::{
+    providers::Env as FigmentEnv,
+    Figment,
+};
+use serde::{
+    Deserialize,
+    Serialize,
+};
+
+pub trait AstriaConfig<'a>: Serialize + Deserialize<'a> {
```
SuperFluffy, line 10: Can you make the prefix an associated const of this trait?

You can have:
```rust
pub trait AstriaConfig<'a>: Serialize + Deserialize<'a> {
    const PREFIX: &'static str;
```
And then in the macro you can set that prefix from the proc macro (or preferably derive macro).

crates/astria-sequencer/src/config.rs
```diff
@@ -27,60 +18,26 @@ pub struct Config {
...
-                    !(RE_END.is_match(key) || RE_START.is_match(val)),
-                    "env vars must not contain spaces in assignment\n{line}"
-                );
-                let prefixed_key = format!("{test_envar_prefix}_{key}");
-                jail.set_env(prefixed_key, val);
-            }
-        }
-    }
+    const EXAMPLE_ENV: &str = include_str!("../local.env.example");
+    const ENV_PREFIX: &str = "ASTRIA_SEQUENCER_";
```
SuperFluffy, line 31: This is pretty boilerplatey - see my comment on the `AstriaConfig` trait using an associated const in the trait that sets this for you.

crates/astria-utils/src/config.rs
SuperFluffy, line 10: Bikeshed: maybe just call this trait `Config`?

crates/astria-composer/src/lib.rs
```diff
@@ -19,7 +19,7 @@
 //! };
 //! # use tracing::info;
 //! # tokio_test::block_on(async {
-//! let cfg = config::get().expect("failed to read configuration");
+//! let cfg = Config::get().expect("failed to read configuration");
```
sambukowski, line 22: I got a doc test fail on this line.
I fixed it by changing the `config` import on line 17 to `config::Config` and that fixed it.

crates/astria-utils/src/config.rs
dereksione: Since the individual structs are called `Config` I wanted to avoid calling this `Config` - plus the name `AstriaConfig` invokes an Astria specific config structure, which I like

crates/astria-utils/src/config.rs
dereksione: Happy to open up a future issue for this one, but I don't believe this is urgent for functionality

crates/astria-config-derive/src/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+        return Err(SynError::new(
+            item_ast.span(),
+            "config struct does not contain a log element, which is necessary",
+        ));
+    };
+
+    Ok(struct_name)
+}
+
+#[proc_macro_attribute]
```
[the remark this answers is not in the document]
dereksione: Happy to do this at a future date, but I think the current implementation is sufficient for desired functionality

crates/astria-composer/src/config.rs
SuperFluffy: Yeah, it's just that the initial implementation was not good. We (== I) should not have enforced this at runtime, only at test-time.

SuperFluffy:
@dereksione pushed a few changes. Please review.

Approving this PR otherwise

SuperFluffy (dismissed):

SuperFluffy:
Need to update the doc

SuperFluffy (changes requested):
Please don't panic in the proc macro. See my previous comments.

SuperFluffy:
To push this through I created a draft PR here: https://github.com/astriaorg/astria/pull/484

Changes with respect to this PR:
1. proc macro removed
2. added a `config::get` utility
3. implemented the config trait by hand (just 2 extra lines per file)
4. gave better names to the test functions
5. added `track_caller` to the tests
6. cleaned up all the cargo.tomls as they had dev-dependencies that were no longer needed
7. added more doc comments
8. imported `astria-config` as `config` everywhere, renamed `AstriaConfig` to `Config` since we are trying to keep the `Astria` prefix to a minimum

Cargo.toml
```diff
@@ -64,3 +66,14 @@ tracing = "0.1"
 tracing-subscriber = "0.3"
 which = "4.4.0"
 wiremock = "0.5"
+
+[patch.crates-io]
```
SuperFluffy, line 70: Needs a rebase on main to rm all of this

crates/astria-config/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+[dependencies]
+serde = { workspace = true }
+figment = { workspace = true, features = ["env"] }
+once_cell = { workspace = true, optional = true }
+regex = { workspace = true, optional = true }
+astria-config-attribute = { path = "../astria-config-attribute", optional = true }
+
+[features]
+attribute = ["dep:astria-config-attribute"]
+config-tests = ["figment/test", "dep:once_cell", "dep:regex"]
```
SuperFluffy, line 17: Can we just call this `tests`?

crates/astria-config/src/config_tests.rs
```diff
@@ -0,0 +1,92 @@
...
+    let test_prefix = format!("TESTTEST_{}", C::PREFIX);
+
+    Jail::expect_with(|jail| {
+        populate_environment_from_example(jail, "TESTTEST", example_env);
+        C::get_with_prefix(test_prefix.as_str(), _internal::Internal).unwrap();
+        Ok(())
+    });
+}
+
+pub fn config_should_reject_unknown_var<'a, C>(example_env: &str)
```
SuperFluffy, line 79: Add `[track_caller]`

crates/astria-config/src/config_tests.rs
```diff
@@ -0,0 +1,92 @@
...
+use figment::Jail;
+use once_cell::sync::Lazy;
+use regex::Regex;
+
+use crate::{
+    Config,
+    _internal,
+};
+
+#[track_caller]
```
SuperFluffy, line 46: Rm line; should be on the pub test below

crates/astria-composer/src/main.rs
```diff
@@ -1,13 +1,14 @@
 use astria_composer::{
-    config,
     telemetry,
     Composer,
+    Config,
 };
+use config::Config as _;
```
SuperFluffy, line 6: Very nit: IMO it would nice to have a `fn get::<C: AstriaConfig>` that delegates to the trait so one need not import this. Kind of like `str::parse` which delegates to `FromStr` so one need not import the trait.

crates/astria-config-attribute/src/lib.rs
```diff
@@ -0,0 +1,69 @@
...
+        .map(|x| {
+            x.ident.clone().ok_or(syn::Error::new(
+                item_top_level_span,
+                "Missing field identifier in struct",
+            ))
+        })
+        .try_collect()?;
+
+    // Check if log is a valid member of the config, otherwise throw compile error
+    let true = struct_fields
```
SuperFluffy, line 35: replace this by `if struct_fields...contains()` instead of having this odd `let pat = ... else` construct.

crates/astria-config-attribute/src/lib.rs
```diff
@@ -0,0 +1,69 @@
...
+        .iter()
+        .map(|x| x.to_string().to_lowercase())
```
SuperFluffy, line 37: I think you don't want to lowercase this since `"Log" != "log"`

crates/astria-config-attribute/src/lib.rs
```diff
@@ -0,0 +1,69 @@
...
+        .collect::<Vec<_>>()
```
SuperFluffy, line 38: No need to collect, use `Iterator::find(|ident| ident == "log")` (irrespective of if you turn `ident` into a string or leave it as `Ident`).

crates/astria-config/src/lib.rs
```diff
@@ -0,0 +1,65 @@
...
+};
+
+pub trait Config: Serialize + DeserializeOwned {
+    const PREFIX: &'static str;
+
+    fn get() -> Result<Self, figment::Error> {
+        Self::get_with_prefix(Self::PREFIX, _internal::Internal)
+    }
+
+    fn get_with_prefix(
```
SuperFluffy, line 48: Let's slap a `#[doc(hidden)]` on this

crates/astria-sequencer/src/Cargo.toml
SuperFluffy, line 1: delete this file

SuperFluffy (approved):
Some oddities in the proc macro that I think should be changed.

Also a spurious `Cargo.toml` in `astria-sequencer/src/` ( think I might have added that in my branch).

Also pleasw ensure that the patch entry gets removed from the workspace Cargo.toml after you rebase on top of main or merge it in.

Rest are nits.


Looks good otherwise, thank you for the patch.

SuperFluffy:
Closing in favor of #537

PR #429, opened 2023-09-25
feat(sequencer): remove byzantine validators in begin_block

### LINKED ISSUE #360: sequencer: handle byzantine validators
we need to decide how to deal with byzantine validators: https://github.com/tendermint/tendermint/blob/v0.34.x/spec/abci/abci.md#beginblock

option: just kick them out for now, as we don't have slashing implemented, only PoA

noot:
## Summary
 remove byzantine validators in `begin_block`

## Background
needed to decide how to handle byzantine validators; for now, since we only have a PoA implementation, we can just remove them 

## Changes
- updated `ValidatorSet` HashMap key to be `account::Id` 
- update `AuthorityComponent::begin_block` to remove byzantine validators from the validator set

## Testing
unit tests

## Related Issues

closes  #360


joroshiba (approved):

crates/astria-sequencer/src/authority/component.rs
```diff
@@ -48,14 +48,26 @@ impl Component for AuthorityComponent {
         Ok(())
     }
 
-    #[instrument(name = "AuthorityComponent:begin_block", skip(_state))]
-    async fn begin_block<S: StateWriteExt + 'static>(
-        _state: &mut Arc<S>,
-        _begin_block: &BeginBlock,
-    ) {
+    #[instrument(name = "AuthorityComponent::begin_block", skip(state))]
+    async fn begin_block<S: StateWriteExt + 'static>(state: &mut Arc<S>, begin_block: &BeginBlock) {
```
SuperFluffy, line 52: Should probably return an error and let tower-abci take down the app gracefully

crates/astria-sequencer/src/authority/state_ext.rs
```diff
@@ -50,31 +54,33 @@ impl ValidatorSet {
...
     /// If the power of a validator is set to 0, remove it from the set.
     /// Otherwise, update the validator's power.
     pub(crate) fn apply_updates(&mut self, validator_updates: ValidatorSet) {
-        for (pub_key, update) in validator_updates.0 {
+        for (address, update) in validator_updates.0 {
             match update.power.value() {
-                0 => self.0.remove(&pub_key),
-                _ => self.0.insert(pub_key, update),
+                0 => self.0.remove(&address),
+                _ => self.0.insert(address, update),
```
SuperFluffy, line 78: Does it make sense to use the `remove` and `push_update` methods you defined above?

SuperFluffy (approved):
Looks good. So this just makes use of cometbft telling us which validators are Byzantine?

Also, is there a chance of this removing all validators? If so, what would be the implications?

crates/astria-sequencer/src/authority/state_ext.rs
noot: I think this is okay since `push_update` recalculates the address which isn't needed

PR #431, opened 2023-09-26
feat(sequencer): implement sudo key changes

### LINKED ISSUE #402: sequencer PoA: allow changing sudo key
follow up to #353

the holder of the sudo key should be able to send a tx which changes the sudo key to some other address.

noot:
## Summary
the PoA implementation has a "sudo key" which is allowed to make validator set changes. this key is set at genesis. this PR implements functionality which allows the sudo key to set a new sudo key.

## Background
needed for devnet

## Changes
- implement `SudoKeyChangeAction` in `Authority` module, which sets the sudo address to a different address
- this action must be signed by the current sudo address

## Testing
unit tests

## Related Issues

closes #402


sambukowski (approved):

steezeburger (approved):

SuperFluffy (approved):
The implementation looks good.

This is implicitly defined by the code, but can you add an explanation to the `SudoAddressChangeAction` handler in sequencer (and also to the PR test) where the address change flow is described?

Basically, if I am reading this PR (or the code in `authority`) I would like to be assured that noone can forge the `from` address and set `sudo_address` to whatever they want.

Also, we have a test for the happy path. Can you also add a test where the action handler returns an error instead (if the `from` address *does not match* the current sudo address)?w

PR #432, opened 2023-09-26
Use plain strings as chain IDs

SuperFluffy:
## Summary
Removed `ChainId` type in favor of a plain old String. Strings are hashed before constructing commitments.

## Background
https://github.com/astriaorg/astria/pull/395 introduces plain strings as chain IDs in its protobuf definitions. This patch is a step to make the refactor easier.

### Open question
1. The `SequenceAction` protobuf message still takes bytes as the chain ID. It should also be changed to take a `String`
2. Sending data to the sequencer costs tokens, but the chain ID is currently not priced in. Should it be added?
3. Pricing in the chain ID encourages shorter names, but potentially also namesquatting. Should we instead commit the sha256 hash of the chain ID?

## Changes
- Replace the `ChainId` type by a plain `String`
- Hash the `String` using sha256 before constructing merkle trees and roots as commitments

## Testing
All present tests should still pass after this change.

## Breaking Changelist
- This is a breaking change because chain IDs are no longer plain bytes but utf8 strings
- The action tree and chain ID commitments change because chain IDs are hashed before constructing trees.

## Related Issues
This is a follow-up to https://github.com/astriaorg/astria/pull/346

Closes https://github.com/astriaorg/astria/pull/433


crates/astria-sequencer-relayer/src/data_availability.rs
```diff
@@ -183,7 +183,7 @@ impl RollupNamespaceData {
     pub fn verify_inclusion_proof(&self, root_hash: [u8; 32]) -> eyre::Result<()> {
         let rollup_data_tree = MerkleTree::from_leaves(self.rollup_txs.clone());
         let rollup_data_root = rollup_data_tree.root();
-        let mut leaf = self.chain_id.as_ref().to_vec();
+        let mut leaf = utils::sha256_hash(self.chain_id.as_bytes()).to_vec();
```
noot, line 186: does this PR change the format of the `sequence::Action` merkle tree? can you update the doc in `specs/sequencer-inclusion-proofs.md` accordingly?

crates/astria-sequencer-types/src/sequencer_block_data.rs
```diff
@@ -317,12 +294,14 @@ impl SequencerBlockData {
                 if let Some(action) = action.as_sequence() {
                     // TODO(https://github.com/astriaorg/astria/issues/318): intern
                     // these namespaces so they don't get rebuild on every iteration.
-                    rollup_data
-                        .entry(ChainId(action.chain_id.clone()))
-                        .and_modify(|data: &mut Vec<Vec<u8>>| {
-                            data.push(action.data.clone());
-                        })
-                        .or_insert_with(|| vec![action.data.clone()]);
+                    // FIXME(https://): change sequence actions to carry string chain IDs
```
noot, line 297: add issue number?

crates/astria-sequencer-types/src/sequencer_block_data.rs
```diff
@@ -317,12 +294,14 @@ impl SequencerBlockData {
...
+                    let chain_id = String::from_utf8_lossy(&action.chain_id);
```
noot, line 298: is it better to us `from_utf8_lossy` over `from_utf8`?

noot (approved):
looks fine, please update specs/sequencer-inclusion-proofs.md before merging!

crates/astria-sequencer-types/src/sequencer_block_data.rs
SuperFluffy: The problem is that we don't know what's inside `action.chain_id`. Right now we define it as `bytes` (in the protobuf definition), so it could be bytes that are not utf8.

SuperFluffy:
@noot I have pushed a small modification where the `sha256(chain_id)` is submitted to the DA rather than the `chain_id` string itself. The commit is reversible.

sambukowski (approved):
lgtm, just need to fix build tests

SuperFluffy:
Closing this in favor of #436

PR #434, opened 2023-09-26
fix(sequencer): Nonces can be equal

joroshiba:
## Summary
Updated the `check_mempool_nonce` to allow all valid nonces through.

## Background
The nonce ensure function was written backwards as an if instead of as an ensure

## Changes
- changed < to >= for nonce validations

### DIFF
--- a/crates/astria-sequencer/src/transaction/mod.rs
+++ b/crates/astria-sequencer/src/transaction/mod.rs
@@ -30,7 +30,7 @@ pub(crate) async fn check_nonce_mempool<S: StateReadExt + 'static>(
         .await
         .context("failed to get account nonce")?;
     ensure!(
-        tx.unsigned_transaction().nonce < curr_nonce,
+        tx.unsigned_transaction().nonce <= curr_nonce,
         "nonce already used by account"
     );
     Ok(())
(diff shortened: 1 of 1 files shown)


itamarreif (approved):

crates/astria-sequencer/src/transaction/mod.rs
```diff
@@ -30,7 +30,7 @@ pub(crate) async fn check_nonce_mempool<S: StateReadExt + 'static>(
         .await
         .context("failed to get account nonce")?;
     ensure!(
-        tx.unsigned_transaction().nonce < curr_nonce,
+        tx.unsigned_transaction().nonce <= curr_nonce,
```
noot, line 33:
```suggestion
        tx.unsigned_transaction().nonce > curr_nonce,
```

i think this is the correct check

noot: sorry i was totally thinking this was an `if` instead of `ensure` when writing this :'(

noot (approved):

crates/astria-sequencer/src/transaction/mod.rs
```diff
@@ -30,7 +30,7 @@ pub(crate) async fn check_nonce_mempool<S: StateReadExt + 'static>(
         .await
         .context("failed to get account nonce")?;
     ensure!(
-        tx.unsigned_transaction().nonce < curr_nonce,
+        tx.unsigned_transaction().nonce >= curr_nonce,
```
SuperFluffy, line 33: Do we have tests for this?

SuperFluffy (approved):

PR #435, opened 2023-09-26
feat(sequencer): implement minting module

### LINKED ISSUE #358: sequencer: implement mint module
replaces #257

Features:
- an action type that will mint funds to the specified recipient
- mint action must be signed by a specific key

noot:
## Summary 
implement minting module which allows the sudo PoA key to mint funds at will

## Background
needed for devnet for a faucet

## Changes
- implement `MintAction` which mints funds to some account, can only be signed by the sudo address
- feature gate this under the `mint` feature for both `astria-proto` and `astria-sequencer`
- add warning log which prints on sequencer startup if mint feature is enabled

## Testing
unit tests

## Related Issues

closes #358


crates/astria-proto/proto/astria/sequencer/v1alpha1/transaction.proto
```diff
@@ -28,6 +28,7 @@ message Action {
         TransferAction transfer_action = 1;
         SequenceAction sequence_action = 2;
         tendermint.abci.ValidatorUpdate validator_update_action = 3;
+        MintAction mint_action = 5;
```
sambukowski, line 31: Why is this 5 and not 4?

noot: 4 is used here: https://github.com/astriaorg/astria/pull/431

noot: so I just did this in anticipation haha

sambukowski (approved):

steezeburger (approved):

crates/astria-proto/buf.gen.yaml
```diff
@@ -8,6 +8,8 @@ plugins:
      - extern_path=.tendermint.types=::tendermint-proto::types
      - extern_path=.tendermint.abci=::tendermint-proto::abci
      - type_attribute=.astria.primitive.v1.Uint128=#[derive(Copy)]
+     - type_attribute=.astria.sequencer.v1alpha1.MintAction=#[cfg(feature = "mint")]
+     - field_attribute=.astria.sequencer.v1alpha1.Action.value.mint_action=#[cfg(feature = "mint")]
```
SuperFluffy, line 12: I don't know if feature gating this here is necessary or advisable. I can see why, but at the same time I wonder if this is going to lead to strange errors.

crates/astria-proto/src/native/sequencer/v1alpha1.rs
```diff
@@ -304,6 +304,8 @@ pub enum Action {
     Transfer(TransferAction),
     ValidatorUpdate(tendermint::validator::Update),
     SudoAddressChange(SudoAddressChangeAction),
+    #[cfg(feature = "mint")]
```
SuperFluffy, line 307: As above. This feature gate is going to infect every crate that somehow handles these actions.

On the one hand I like that this is unrepresentable, but on the other I think the handling of these mint actions should be entirely confined to the sequencer handlers

crates/astria-sequencer/src/main.rs
```diff
@@ -24,6 +24,13 @@ async fn main() -> ExitCode {
         config = serde_json::to_string(&config).expect("serializing to a string cannot fail"),
         "initializing sequencer"
     );
+
+    #[cfg(feature = "mint")]
+    {
+        tracing::warn!("MINT FEATURE IS ENABLED!");
```
SuperFluffy, line 30: Can you spin up a task that prints this every 5 seconds?

Also, it should also be written to stderr (`eprintln!`)

crates/astria-proto/buf.gen.yaml
noot: it won't compile otherwise :( due to the feature gating in sequencer, i also need to do feathre gating here, otherwise will complain due to missing enum match arms

crates/astria-proto/src/native/sequencer/v1alpha1.rs
noot: see above - should i just ignore enum cases that are not enabled in the sequencer then?

SuperFluffy: Yes, I think so. I would actually leave two mint components in: the deactivated one that just logs if we received a mint request (and maybe from where), and the feature gated one that has the actual functionality.

crates/astria-sequencer/src/main.rs
noot: why stderr if this isn't an error?

noot: updated though!

SuperFluffy: stderr is not only used for errors but also other information like debug, config, etc. stdout is the intended behaviour of a program (think counting the lines of a program) and stderr for a summary of input flags, files, debug info etc

crates/astria-sequencer/src/main.rs
```diff
@@ -24,6 +24,17 @@ async fn main() -> ExitCode {
         config = serde_json::to_string(&config).expect("serializing to a string cannot fail"),
         "initializing sequencer"
     );
+
+    #[cfg(feature = "mint")]
+    tokio::spawn(async {
+        let duration = std::time::Duration::from_secs(5);
+        loop {
+            eprintln!("MINT FEATURE IS ENABLED!");
```
SuperFluffy, line 32: Please return `warn!` :-D

I wanted logging to stderr in case tracing is disabled, but also `warn!` if tracing is enabled but we end up using a a tracing subscriber that sends directly to e.g. opentelemetry.

crates/astria-sequencer/src/transaction/mod.rs
```diff
@@ -96,6 +98,12 @@ impl ActionHandler for UnsignedTransaction {
                 Action::SudoAddressChange(act) => act
                     .check_stateless()
                     .context("stateless check failed for SudoAddressChangeAction")?,
+                #[cfg(feature = "mint")]
+                Action::Mint(act) => act
+                    .check_stateless()
+                    .context("stateless check failed for MintAction")?,
+                #[cfg(not(feature = "mint"))]
+                _ => bail!("unsupported action type: {:?}", action),
```
SuperFluffy, line 106: is this going to crash the sequencer? Sorry, don't remember if action handler errors are printed or bubbled up.

SuperFluffy (approved):
At the risk of seeming paranoid:

1. please bring back `warn!` in addition to `eprintln!` (see the comment)
2. should we add the following env vars?
    a. `ASTRIA_SEQUENCER_ENABLE_MINT`
    b. `ASTRIA_SEQUENCER_NOT_PRODUCTION`

and if the mint module is active you must then set `ASTRIA_SEQUENCER_ENABLE_MINT=1 && ASTRIA_SEQUENCER_NOT_PRODUCTION` because it will refuse to start up otherwise?

crates/astria-sequencer/src/main.rs
noot: updated

crates/astria-sequencer/src/main.rs
noot: updated!

crates/astria-sequencer/src/transaction/mod.rs
noot: no, it won't crash the sequencer - this is used in `deliver_tx` which just warns if a check fails.

noot:
@SuperFluffy I think having 2 env vars along with the feature is a bit overkill, maybe one env var is good for now? should I add in this PR?

SuperFluffy:
@noot 
> @SuperFluffy I think having 2 env vars along with the feature is a bit overkill, maybe one env var is good for now? should I add in this PR?

Yeah, let's add one of them to really make sure. 😊

PR #436, opened 2023-09-27
feat: require chain_id be 32 bytes

SuperFluffy:
## Summary
Make chain IDs 32 bytes and construct them from sha256-hashed names.

## Background
At the moment chain IDs sent as part of sequence actions can be arbitrarily long. The only place where an upper limit on chain IDs is enforced is in `SequencerBlockData`, which however is constructed in sequencer relayer and after cometbft/sequencer have come to a consensus over new data.

Using a chain ID of 32 bytes as opposed to a vec is desirable for the these reasons:
+ it allows it to be `Copy` so that unnecessary vec-allocations are avoided
+ it minimizes the likelihood of rollup namespace clashes because it suggests to use sha256 (although it cannot prevent it)
+ it allows to deterministically calculate the overhead for each transaction allowing to take it into account when calculating tx submission fees

## Changes
- Make `ChainId` a think wrapper over `[u8; 32]` (away from `Vec<u8>`)
- Move `ChainId` to `astria-proto`
- Reimplement all services in terms of it (sequencer now enforces the length through `ChainId` being a part of `SequenceAction`, whereas before it was just a byte vector)
- Adjust all tests

## Testing
All tests still pass or have been adjusted. The snapshot test for calculating the merkle tree hash over the actions had to be updated.

## Breaking Changelist
This breaks all public APIs because chain IDs are now expected to be 32 bytes (whereas they were arbitrary length before).


noot (approved):
looks good to me!

PR #443, opened 2023-09-28
proto: define sequencer v1alpha1

SuperFluffy:
This patch extracts Rust types that are used in communication between sequencer, sequencer-relayer, conductor, and celestia (as the data availability layer). The resulting protobuf types are intended to be the only way that services exchange data, making the current json-based wire format obsolete.

The former protobuf definitions defined in sequencer.v1alpha1 are overriden. This does not constitute a breaking change because these types were never used.

## Summary

## Background

## Changes

## Testing
How are these changes tested?

## Breaking Changelist

## Related Issues

closes


SuperFluffy:
Implemented by #395

Issue #439, opened 2023-09-28
conductor: sync full node from sequencer/celestia


joroshiba:
Conductor works right now only on a new rollup node, new transactions come in and can be executed. We want to update the architecture such that on startup, it will sync from a configured starting block for sequencer + DA  to sync from history before starting to follow the head block.

**User Journey**
User deploys a rollup locally during development, generates data which is pushed to live sequencer network + celestia network. Data is preserved by these networks, and they want to be able to start a cloud hosted node. They deploy their rollup on their cloud provider, tell it where to start syncing from and get all their local test/starting block data by syncing.

sambukowski:
I stated a design doc in Notion for planning the implementation of this:
https://www.notion.so/astria-org/Conductor-Sync-Design-09f56f2b2ab04a0c8a5d13e83e279719

PR #445, opened 2023-09-28
conductor: add genesis sequencer block height to config and env vars

sambukowski:
## Summary
Add an env var that specifies which sequencer block (using the block height) the genesis block of the rollup was in.

## Background
This is a requirement for adding sync functionality to the Conductor.

## Changes
- Added `ASTRIA_CONDUCTOR_GENESIS_SEQUENCER_BLOCK_HEIGHT` env var and `genesis_sequencer_block_height` to Conductor config.

## Testing
This should pass all existing tests.

## Related Issues
#439


joroshiba (approved):
Might want to bike shed the config name later but this is fine for now.

steezeburger (approved):

crates/astria-conductor/local.env.example
```diff
@@ -33,3 +33,6 @@ ASTRIA_CONDUCTOR_SEQUENCER_URL="ws://127.0.0.1:26657"
 # true: skips empty blocks
 # false: executes empty blocks
 ASTRIA_CONDUCTOR_DISABLE_EMPTY_BLOCK_EXECUTION=false
+
+# the sequencer block height that the rollup genesis block was in
```
noot, line 37:
```suggestion
# the sequencer block height that the rollup's first block was in
```

crates/astria-conductor/local.env.example
```diff
@@ -33,3 +33,6 @@ ASTRIA_CONDUCTOR_SEQUENCER_URL="ws://127.0.0.1:26657"
...
+ASTRIA_CONDUCTOR_GENESIS_SEQUENCER_BLOCK_HEIGHT=0
```
noot, line 38:
```suggestion
ASTRIA_CONDUCTOR_INIITIAL_SEQUENCER_BLOCK_HEIGHT=0
```

crates/astria-conductor/local.env.example
noot: the genesis block of the rollup is determined by the genesis state/file - the first block that uses the sequencer would be block 1

PR #450, opened 2023-09-29
fix(ci): Run all tests for every non-pr, remove glob check

joroshiba:
## Summary
Removed a glob check to try and omit running tests for simple markdown changes, and make all tests run when not a pull_request event.

## Background
Tests failed, we bypassed tests which should run.

## Changes
- removed md glob ommission
- run all tests on main.

## Testing
Lots of iterations and commits to test what runs when on this PR.


SuperFluffy:
The tests are not running on this pr? Is that expected?

joroshiba:
> The tests are not running on this pr? Is that expected?

Yes, that's why there are lots of commits where I pushed changes to other files to make sure they work. See, for example this run: where I confirmed things would run: https://github.com/astriaorg/astria/pull/450/checks?sha=ae26ab28c0d67381e6d45b31c4a3d79c3ea88dc7

SuperFluffy (approved):

PR #451, opened 2023-09-29
conductor: restart websocket connection

joroshiba:
## Summary
Recreates websocket connection instead of destroying driver when connection is closed.

## Background
Websockets generally have timeouts, we need to seamlessly spin re-attempt to connect.

## Changes
- move creation of sequencer client into a function instead of having be part of Driver startup
- create client when run starts, create new client if the connection closes.

## Testing
Manually tested within dev-cluster.

## Issues
closes astriaorg/astria#451


sambukowski (approved):

crates/astria-conductor/src/driver.rs
```diff
@@ -168,8 +162,14 @@ impl Driver {
                     if let Some(block) = new_block {
                         self.handle_new_block(block).await
                     } else {
-                        warn!("sequencer new-block subscription closed unexpectedly; shutting down driver");
-                        break;
+                        warn!("sequencer new-block subscription closed; restarting");
+                        match self.get_new_block_stream().await {
+                            Ok(res) => new_blocks = res,
+                            Err(_) => {
```
noot, line 168: log this error as part of the `warn`

crates/astria-conductor/src/driver.rs
```diff
@@ -168,8 +162,14 @@ impl Driver {
...
+                                warn!("sequencer new-block restart failed; shutting down driver");
```
noot, line 169:
```suggestion
                                warn!("sequencer new-block subscription restart failed; shutting down driver");
```

crates/astria-conductor/src/driver.rs
```diff
@@ -185,6 +185,18 @@ impl Driver {
...
     }
 
+    async fn get_new_block_stream(&self) -> Result<NewBlocksStream> {
+        use sequencer_client::SequencerSubscriptionClientExt as _;
+
+        let sequencer_client = SequencerClient::new(&self.sequencer_ws_url)
+            .await
+            .wrap_err("failed constructing a cometbft websocket client to read off sequencer")?;
+
+        let block_stream = sequencer_client.client.subscribe_new_block_data().await?;
```
noot, line 195: wrap_err

noot (approved):

SuperFluffy:
This is addressed as part of #455. Note that the present PR does not address a failure of the underlying websocket connection: since `WebSocketClient` is just a thin wrapper of an mpsc sender to communicate with the websocket driver that is managing the communication with the remote host, the entire TCP connection needs to be reestablished because the mpsc channel is dropped after the driver goes down.

PR #454, opened 2023-10-02
feat(sequencer-relayer): use websocket client for sequencer instead of polling

### LINKED ISSUE #279: sequencer-relayer: use cometbft websocket
Right now sequencer relayer runs on a polling tick to grab new blocks from the cometbft instance. This requires we poll the sequencer client, which introduces complexity to tests, and doesn't guarantee we get new blocks ASAP.

We can update this to use a [websocket connection subscription](https://docs.cometbft.com/v0.37/core/subscription) for cometbft, and instead have a subscription push event to the relayer.

noot:
## Summary
 use the tendermint websocket client for sequencer inside sequencer-relayer instead of polling.

## Background
polling is annoying since we need to configure the block time.

## Changes
- update relayer to use the tendermint `WebSocketClient` and start a `NewBlock` subscription

## Testing
relayer blackbox tests (wip need to update)

## Related Issues

closes #279


noot:
closing due to being really stale and we moved to the sequencer API

Issue #452, opened 2023-09-29
conductor: websocket closure restart


joroshiba:
When the websocket connection to sequencer closes, we completely kill the driver which shuts down conductor entirely. Instead we should attempt to recreate a connection, because timeout closures are normal with ingress.

Issue #393, opened 2023-09-19
Replace http client by websocket client


SuperFluffy:
PR #332 introduces a websocket client to stream new blocks from cometbft/sequencer.

The stil existing http client used to verify blocks should be replaced by the same websocket client to reuse the open tcp socket.

Doing this will require refactoring block verifier and driver to accept websocket client types instead of opening sockets themselves.

PR #455, opened 2023-10-02
conductor: refactor and implement full node sync from sequencer

SuperFluffy:
## Summary
This patch adds sync functionality to conductor. On every startup conductor will get all blocks up to the latest height from sequencer and forward them to the executor before submitting blocks it got from its subscription.

## Background
Because conductor is thought of as stateless sidecar to a rollup node there needs to be a mechanism to bring the rollup node to the latest state. This patch adds that functionality by fetching all sequencer blocks up to the latest sequencer height for the target chain ID and replaying ("executing") them against the rollup node.

Because the prior implementation of conductor was strictly sequential (blocking the async loop) this patch required a refactor of conductor into fully async components.

## Changes
- Replace all http clients by websocket clients.
- Add an object pool of websocket clients with a managing task that restarts the websocket connection on failure.
- Remove the `Driver` type.
- Simplify Executor by not being generic over some client
- Introduce the `data_availability::Reader` and `sequencer::Reader` types that are independently reading off celestia and sequencer, respectively
- Refactor `data_availability::Reader` from a for-loop blocking the actor's event loop into a chain of futures that are driven by its event loop
- Perform sequencer namespace data verification in one step:
    - the block verifier need not get a validator set for each rollup blob retrieved from celestia; it is enough to run its inclusion proof + data against the root in the sequencer data.
    - since the sequencer namespace data is verified and rollup processing does not continue without it this is enough
- Add sync functionality to `sequencer::Reader`: fetch all sequencer blocks between a configured initial height and the latest height and submit them to the rollup node for execution before submitting newer blocks.
- Refactor the `Conductor` entry point to track its constituent tasks in a `JoinMap`.

## Testing
These tests have to be tested by running conductor in devnet. Conductor does not currently have black box tests. These must be added in a follow PR.

The executor mock tests have been uncommented because executor should not be generic. These too will be replaced by blackbox tests.

## Breaking Changelist
- the `ASTRIA_CONDUCTOR_TENDERMINT_URL` has been removed. All communication with cometbft/sequencer happens over the URL configured in `ASTRIA_CONDUCTOR_SEQUENCER_URL`.
- no other breaking changes should be observable.

## Related Issues
closes https://github.com/astriaorg/astria/issues/381
closes https://github.com/astriaorg/astria/issues/393
closes https://github.com/astriaorg/astria/issues/439
closes https://github.com/astriaorg/astria/issues/451
closes https://github.com/astriaorg/astria/issues/452


crates/astria-conductor/src/data_availability.rs
```diff
@@ -0,0 +1,382 @@
...
+    }
+
+    fn get_latest_height(&mut self) {
+        let client = self.celestia_client.inner().clone();
+        self.get_latest_height = Some(tokio::spawn(async move {
+            Ok(client.header_network_head().await?.height())
+        }))
+    }
+
+    fn get_sequencer_datas(&mut self, latest_height_res: Result<eyre::Result<u64>, JoinError>) {
```
joroshiba, line 153: I think this function name is a bit confusing, it matches `get_sequencer_datas` `JoinSet` but it only sometimes does things related, and then only instantiates jobs on that set?

When reading from top I was concerned based on naming we were double fetching, maybe something like `start_data_availability_fetch`?

crates/astria-conductor/src/data_availability.rs
```diff
@@ -0,0 +1,382 @@
...
+            );
+            return;
+        }
+        let first_new_height = self.current_block_height + 1;
+        info!(
+            height.start = first_new_height,
+            height.end = self.current_block_height,
+            "spawning tasks to fetch sequencer blocks for different celestia heights",
+        );
+        for height in first_new_height..=self.current_block_height {
```
joroshiba, line 182: Shouldn't this be going to `latest_height` instead of `self.current_block_height`?

crates/astria-conductor/src/data_availability.rs
```diff
@@ -0,0 +1,382 @@
...
+        info!(da_height = current_block_height, "creating Reader");
+
+        Ok(Self {
+            executor_tx,
+            celestia_client,
+            celestia_poll_interval,
+            current_block_height,
+            get_latest_height: None,
+            get_sequencer_datas: JoinMap::new(),
+            process_sequencer_datas: JoinMap::new(),
```
joroshiba, line 103: nit: this is data_availability data and we need to process it to make sure it comes from the sequencer, the output of processing is sequencer data

crates/astria-conductor/src/data_availability.rs
```diff
@@ -0,0 +1,382 @@
...
+            }
+            Ok(Err(e)) => {
+                warn!(error.message = %e, error.cause = ?e, "task processing sequencer data returned with an error");
+                return Ok(());
+            }
+            Ok(Ok(subsets)) => subsets,
+        };
+        self.executor_tx
+            .send(executor::ExecutorCommand::FromCelestia(subsets))
+            .wrap_err("failed sending processed sequencer subsets: executor channel is closed")
```
joroshiba, line 265: I believe we have a problematic race condition bug here in 2 different instances:

1. If blocks are received out of order from DA, they will be executed out of order potentially.
2. On startup, if there are blocks being synced (in sequencer_reader loop), and this fires off some new data, again executing out of order.

The first one we can call a bug and hope doesn't happen, create an issue around the second I believe is pretty problematic.

crates/astria-conductor/src/sequencer.rs
```diff
@@ -0,0 +1,340 @@
...
+                return Err(e).wrap_err("first sequencer block returned from subscription was bad");
+            }
+            Some(Ok(block)) => {
+                let height = block.header().height.value();
+                pending_blocks.push_back(futures::future::ready(Ok(block)));
+                height
+            }
+        };
+
+        let latest_height: u32 = latest_height.try_into().wrap_err(
```
joroshiba, line 123: ditto u32 vs u64 from config comment

crates/astria-conductor/src/sequencer.rs
```diff
@@ -0,0 +1,340 @@
...
+                        }
+                    };
+                    break 'reader_loop ret;
+                }
+
+                // Drain the stream of heights to sync the sequencer reader.
+                // The stream is fused, so after it exhausted this match arm will be deactivated.
+                // The condition on the length of priority blocks relies on the pool being currently set
+                // to 50. This ensures that no more than 20 requests to the sequencer are active at the same time.
+                // Leaving some objects in the pool is important, so that the match arm for resolving priority blocks
```
joroshiba, line 165: I think this comment is out of date, mentions priority blocks

crates/astria-conductor/src/sequencer.rs
```diff
@@ -0,0 +1,340 @@
...
+                // below can reschedule blocks.
+                Some(res) = sync_heights.next(), if !sync_heights.is_done() && sync_blocks.len() < 20 => {
+                    if let Err(e) = sync_block_at_height(&mut sync_blocks, res) {
+                        error!(error.message = %e, error.cause = ?e, "syncing heights failed; exiting reader");
```
joroshiba, line 169: if there is an error grabbing a client it seems like this will trigger full reader loop exit?

crates/astria-conductor/src/sequencer.rs
```diff
@@ -0,0 +1,340 @@
...
+                    schedule_for_forwarding_or_resubscribe(
+                        &mut resubscribe,
+                        &mut pending_blocks,
+                        self.pool.clone(),
+                        new_block,
+                    );
+                }
+
+                // Regular pending blocks will be submitted to the executor in the order they were received.
+                // The condition on priority_blocks ensures that blocks from the sync process are sent first.
```
joroshiba, line 206: priority blocks mention again

crates/astria-conductor/src/sequencer.rs
```diff
@@ -0,0 +1,340 @@
...
+        let latest_height: u32 = latest_height.try_into().wrap_err(
+            "failed converting the cometbft height to u32, but this should always work",
+        )?;
+
+        info!(
+            height.initial = self.initial_sequencer_block_height,
+            height.latest = latest_height,
+            "syncing sequencer between configured initial and latest retrieved height"
+        );
+        let mut sync_heights = {
```
joroshiba, line 132: To make sure I understand correctly, we now only sync on startup? I believe in the previous version we looked over if a block came via the subscription out of order we would also start a sync on the missing blocks?

Is a sync job potentially something we can trigger via an inbound command? This would allow the executor to have an execution queue and start a sync if something was missing?

crates/astria-conductor/src/sequencer.rs
```diff
@@ -0,0 +1,215 @@
...
+    /// The channel used to send messages to the executor task.
+    executor_tx: executor::Sender,
+
+    pool: deadpool::managed::Pool<crate::client_provider::ClientProvider>,
+
+    shutdown: oneshot::Receiver<()>,
+}
+
+impl Reader {
+    #[instrument(name = "driver", skip_all)]
```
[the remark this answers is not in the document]
SuperFluffy: removed instrumentation: `sequencer::Reader::new` is not async and can't fail.

crates/astria-conductor/src/sequencer.rs
```diff
@@ -0,0 +1,215 @@
...
+            .latest_sequencer_block()
+            .await
+            .wrap_err("failed to issue an initial request for the latest block")?;
+
+        let mut sync_heights = {
+            let pool = self.pool.clone();
+            Box::pin(
+                futures::stream::iter(1..latest_block.header().height.value()).then(
+                    move |height| {
+                        let height: u32 = height.try_into().expect(
```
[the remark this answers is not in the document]
SuperFluffy: I moved that height construction out of the block and have `sequencer::Reader::run_until_stopped` return an error if `latest_height` cannot be turned into a `u32`., but the question is still relevant:

CometBft block heights are internally recorded as i64. `tendermint-rs` decided to store them as a wrapper around `u64` here: https://docs.rs/tendermint/latest/src/tendermint/block/height.rs.html#17

However, in practice only the range `0..u32::MAX` can be represented and so any `u64` that cannot be made into a `i64` is rejected: https://docs.rs/tendermint/latest/src/tendermint/block/height.rs.html#38

So only `u32` are valid.

crates/astria-conductor/src/sequencer.rs
```diff
@@ -0,0 +1,215 @@
...
+            .await
+            .wrap_err("failed getting client from pool to fetch initial latest block")?
+            .latest_sequencer_block()
+            .await
+            .wrap_err("failed to issue an initial request for the latest block")?;
+
+        let mut sync_heights = {
+            let pool = self.pool.clone();
+            Box::pin(
+                futures::stream::iter(1..latest_block.header().height.value()).then(
```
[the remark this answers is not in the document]
SuperFluffy: Done

crates/astria-conductor/src/conductor.rs
```diff
@@ -10,39 +18,114 @@ use tokio::{
...
+    tasks: JoinMap<&'static str, eyre::Result<()>>,
+    /// channels to the long-running tasks to shut them down gracefully
+    shutdown_channels: HashMap<&'static str, oneshot::Sender<()>>,
+    sequencer_client_pool: deadpool::managed::Pool<ClientProvider>,
 }
 
 impl Conductor {
+    const DATA_AVAILABILITY: &str = "data_availability";
+    const EXECUTOR: &str = "executor";
+    const SEQUENCER: &str = "sequencer";
```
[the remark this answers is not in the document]
SuperFluffy: Not in this case because I only want these as keys into the `JoinMap` and for logging events. Using an enum would would just make this more complex (and either require implementin/deriving a bunch of traits or always going `enum variant -> str`).

crates/astria-conductor/src/conductor.rs
```diff
@@ -10,39 +18,115 @@ use tokio::{
...
+
     pub async fn new(cfg: Config) -> eyre::Result<Self> {
+        let mut tasks = JoinMap::new();
+        let mut shutdown_channels = HashMap::new();
+
         let signals = spawn_signal_handler();
         // spawn our driver
```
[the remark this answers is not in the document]
SuperFluffy: Removed the comment because it was stating the obvious.

crates/astria-conductor/src/sequencer.rs
SuperFluffy: Comment above, reproducing it here:

> CometBft block heights are internally recorded as i64. `tendermint-rs` decided to store them as a wrapper around `u64` here: https://docs.rs/tendermint/latest/src/tendermint/block/height.rs.html#17
>
> However, in practice only the range `0..u32::MAX` can be represented and so any `u64` that cannot be made into a `i64` is rejected: https://docs.rs/tendermint/latest/src/tendermint/block/height.rs.html#38
>
> So only `u32` are valid.

crates/astria-conductor/src/sequencer.rs
SuperFluffy: Yes, exactly. Assume the driver connection failed, then `pool.get().await` will essentially block on `WebSocketClient::new(url).await` in `ClientProvider` resolving, so the steps are:

1. call `pool.get().await`
2. create a one-shot channel, the sender half is sent to `ClientProvider` 
3. in `get()`: immediately block on `rx.await` to get a client from `ClientProvider`
4. in ClientProvider: if `client = None`, push the `tx` half into `pending_requests`
5. wait for the reconnection to resolve
6. if successful: drain `pending_requests`, sending a new client to each requester
7. if failed: exit the loop, dropping `pending_requests`, which drops all `tx`s closing the channel
8. in `get()`: `recv.await` will only fail if the tx has been dropped

crates/astria-conductor/src/sequencer.rs
SuperFluffy: sync on startup: Yes, that is correct.

previous version: Looking at current `main`, I don't think we had that, at least not for the sequencer part.

trigger sync via command: I think we can leave this squarely inside the `sequencer::Reader` actor so that executor need not be concerned with it: we can add logic to the sequencer reader to track the current height and require it be contiguous. If we detect a hole, we can explicitly fetch that block and pause forwarding to the executor until the hole is plugged.

Follow-up PR?

crates/astria-conductor/src/config.rs
```diff
@@ -43,7 +40,7 @@ pub struct Config {
     pub disable_empty_block_execution: bool,
 
     /// The Sequencer block height that the rollup genesis block was in
-    pub initial_sequencer_block_height: u64,
+    pub initial_sequencer_block_height: u32,
```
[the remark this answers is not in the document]
SuperFluffy: Commented elsewhere, but it's because only `0..u32::MAX` is valid: https://docs.rs/tendermint/latest/src/tendermint/block/height.rs.html#35-44

crates/astria-conductor/src/conductor.rs
```diff
@@ -10,39 +18,115 @@ use tokio::{
...
+        tasks.spawn(Self::SEQUENCER, sequencer_reader.run_until_stopped());
+        shutdown_channels.insert(Self::SEQUENCER, sequencer_shutdown_tx);
+
+        if !cfg.disable_finalization {
+            let (shutdown_tx, shutdown_rx) = oneshot::channel();
+            let block_verifier = BlockVerifier::new(sequencer_client_pool.clone());
+            let data_availability_reader = data_availability::Reader::new(
+                &cfg.celestia_node_url,
+                &cfg.celestia_bearer_token,
+                std::time::Duration::from_secs(3),
```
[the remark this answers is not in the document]
SuperFluffy: Yeah, it should be, but I would leave this to a follow-up because this was copied from the current impl.

crates/astria-conductor/local.env.example
```diff
@@ -24,8 +20,9 @@ ASTRIA_CONDUCTOR_EXECUTION_RPC_URL="http://127.0.0.1:50051"
 # disable block finalization
 ASTRIA_CONDUCTOR_DISABLE_FINALIZATION=false
 
-# The URL to a fully trusted CometBFT/Sequencer to subscribe to new blocks
-# over websocket using cometbft's RPC.
+# The URL to a fully trusted CometBFT/Sequencer to issue cometbft RPCs. Example
+# RPCs are subscribing to new blocks, fetching blocks at a given level, or
+# retrieving validators.
 # 127.0.0.1:26657 is the default socket address in comebft's `rpc.laddr` setting.
 ASTRIA_CONDUCTOR_SEQUENCER_URL="ws://127.0.0.1:26657"
```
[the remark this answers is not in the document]
SuperFluffy: good catch, ty

crates/astria-conductor/src/data_availability.rs
SuperFluffy: Ad 1: yeah, we can fix that by enforcing heights being contiguous. Follow-up?

Ad 2: I have pushed a commit that tries to address this: `sequencer::Reader` gets a oneshot channel over which it notifies conductor if sync is complete. Only once sync is complete will conductor execute `data_availability::Reader::run_until_stopped`. Do you thinks this addresses the issue?

SuperFluffy:
Update: I have significantly simplified `sequencer::Reader` by moving the sync process into its own future.

crates/astria-conductor/src/sequencer.rs
SuperFluffy: Removed and replaced by `sync`.

crates/astria-conductor/src/data_availability.rs
SuperFluffy: The naming comes from `SequencerNamespaceData`, so the thinking is that we are processing multiple `sequencer_(namespace_)datas`. The naming is very unfortunate.

crates/astria-conductor/src/conductor.rs
```diff
@@ -10,39 +18,115 @@ use tokio::{
...
+            cfg.disable_empty_block_execution,
+            executor_rx,
+            executor_shutdown_rx,
+        )
+        .await
+        .wrap_err("failed to construct executor")?;
+
+        tasks.spawn(Self::EXECUTOR, executor.run_until_stopped());
+        shutdown_channels.insert(Self::EXECUTOR, executor_shutdown_tx);
+
```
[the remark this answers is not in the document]
SuperFluffy: I have put this and the other bits into blocks with a small comment to make it clear that the variables only live as long as the block. I hope that makes it clearer.

crates/astria-conductor/src/conductor.rs
```diff
@@ -10,39 +18,115 @@ use tokio::{
...
+            sequencer_shutdown_rx,
+            executor_tx.clone(),
+        )
+        .await
+        .wrap_err("failed initializing driver")?;
+
+        tasks.spawn(Self::SEQUENCER, sequencer_reader.run_until_stopped());
+        shutdown_channels.insert(Self::SEQUENCER, sequencer_shutdown_tx);
+
+        if !cfg.disable_finalization {
```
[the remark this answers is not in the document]
SuperFluffy: Not sure what comment you thought of, but I noted that the DA reader is constructed at that point and and only ran after sync is done.

crates/astria-conductor/src/conductor.rs
```diff
@@ -10,39 +18,115 @@ use tokio::{
...
+        let (sequencer_shutdown_tx, sequencer_shutdown_rx) = oneshot::channel();
+        let sequencer_reader = sequencer::Reader::new(
+            cfg.initial_sequencer_block_height,
+            sequencer_client_pool.clone(),
+            sequencer_shutdown_rx,
+            executor_tx.clone(),
+        )
+        .await
+        .wrap_err("failed initializing driver")?;
+
```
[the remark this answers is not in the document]
SuperFluffy: Removed. Also, the reader instantiation can no longer fail.

crates/astria-conductor/src/data_availability.rs
```diff
@@ -0,0 +1,382 @@
...
+
+    get_latest_height: Option<JoinHandle<eyre::Result<u64>>>,
+
+    /// A map of in-flight queries to celestia for new sequencer blocks
+    get_sequencer_datas:
+        JoinMap<u64, eyre::Result<Vec<SignedNamespaceData<SequencerNamespaceData>>>>,
+
+    process_sequencer_datas: JoinMap<u64, eyre::Result<Vec<SequencerBlockSubset>>>,
+
+    shutdown: oneshot::Receiver<()>,
```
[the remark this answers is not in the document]
SuperFluffy: Added

crates/astria-conductor/src/conductor.rs
```diff
@@ -10,122 +18,239 @@ use tokio::{
...
                 }
 
-
-                res = async { reader.as_mut().unwrap().await }, if reader.is_some() => {
+                Some((name, res)) = tasks.join_next() => {
                     match res {
-                        Ok(Ok(())) => error!("reader task exited unexpectedly"),
-                        Ok(Err(e)) => error!(error.msg = %e, error.cause = ?e, "reader exited with error"),
-                        Err(e) => error!(error.msg = %e, error.cause = ?e, "reader task failed"),
+                        Ok(Ok(())) => error!(tak.name = name, "task exited unexpectedly, shutting down"),
```
sambukowski, line 205: nit: typo: tak -> task

crates/astria-conductor/src/executor.rs
```diff
@@ -327,247 +305,243 @@ impl<C: ExecutionClient> Executor<C> {
...
-                .await
-                .get(&executor.execution_state)
-                .is_some()
-        );
-        assert_eq!(expected_execution_state, executor.execution_state);
-        // should be empty because block was executed and finalized, which deletes it from the map
-        assert!(executor.sequencer_hash_to_execution_hash.is_empty());
-    }
-}
+// #[cfg(test)]
```
sambukowski, line 308: a new issue should be created to fix and add these tests back in

crates/astria-conductor/src/client_provider.rs
```diff
@@ -0,0 +1,134 @@
...
+        .build()
+        .wrap_err("failed to create sequencer client pool")
+}
+
+#[derive(Debug, thiserror::Error)]
+pub(crate) enum Error {
+    #[error("the client provider failed to reconnect and is permanently closed")]
+    Failed,
+    #[error("the channel over which to receive a client was closed unexpectedly")]
+    ClientChannelDroped,
```
noot, line 39:
```suggestion
    ClientChannelDropped,
```

crates/astria-conductor/src/client_provider.rs
```diff
@@ -0,0 +1,134 @@
...
+        oneshot,
+    },
+    task::JoinHandle,
+};
+use tracing::warn;
+
+type ClientRx = mpsc::UnboundedReceiver<oneshot::Sender<Result<WebSocketClient, Error>>>;
+type ClientTx = mpsc::UnboundedSender<oneshot::Sender<Result<WebSocketClient, Error>>>;
+
+pub(super) async fn start_pool(url: &str) -> eyre::Result<Pool<ClientProvider>> {
```
noot, line 24: what's the purpose of using a pool over just having each component have their own `ClientProvider`?

SuperFluffy: You mean having one open websocket connection per component?

The object pool with the provider solves two things:

1. You only have one websocket connection to the sequencer (that's just good practice and the reason to use websocket in the first place).
2. You ensure that there is a cap on the maximum number of requests to sequencer alive at the same time.

I hope this answers the question?

crates/astria-conductor/src/conductor.rs
```diff
@@ -10,122 +18,239 @@ use tokio::{
...
-                            error.msg = %e,
-                            error.cause = ?e,
-                            "error sending GetNewBlocks command to driver"
-                        );
-                        break;
-                    }
-                }
 
-                res = &mut driver => {
+                res = &mut sync_done, if !sync_done.is_terminated() => {
```
noot, line 189: would there be a case where `sync_done` is ready but `sync_done.is_terminated()`? reading the docs, the future is supposed to be terminated when it returns `Poll::Ready` right? https://docs.rs/futures/latest/futures/future/trait.FusedFuture.html

crates/astria-conductor/src/conductor.rs
```diff
@@ -10,122 +18,239 @@ use tokio::{
...
-                    match res {
-                        Ok(Ok(())) => error!("executor task exited unexpectedly"),
-                        Ok(Err(e)) => error!(error.msg = %e, error.cause = ?e, "executor exited with error"),
-                        Err(e) => error!(error.msg = %e, error.cause = ?e, "executor task failed"),
+                    if let Some(data_availability_reader) = data_availability_reader.take() {
+                        info!("starting data availability reader");
+                        tasks.spawn(
+                            Self::DATA_AVAILABILITY,
+                            data_availability_reader.run_until_stopped(),
+                        );
```
noot, line 199: the sync occurs entirely from the sequencer node instead of from DA? guess this makes sense, since the sequencer nodes should hold all block data (currently). i wonder if syncing from a celestia node could also make sense eventually

crates/astria-conductor/src/conductor.rs
```diff
@@ -10,122 +18,239 @@ use tokio::{
...
+        if !tasks.is_empty() {
+            warn!(
+                number = tasks.len(),
+                "aborting tasks that haven't shutdown yet"
+            );
+            Rc::get_mut(&mut tasks)
+                .expect("only one Rc to the conductor tasks should exist; this is a bug")
+                .shutdown()
+                .await;
+        }
```
noot, line 253: what's the purpose of the `Rc` and this section over just waiting 5 seconds?

crates/astria-conductor/src/data_availability.rs
```diff
@@ -0,0 +1,382 @@
...
+    warn,
+    Instrument,
+};
+
+use crate::{
+    block_verifier::{
+        self,
+        BlockVerifier,
+    },
+    // config::Config,
```
noot, line 40: remove

crates/astria-conductor/src/data_availability.rs
```diff
@@ -0,0 +1,382 @@
...
+                }
+            )
+        }
+        Ok(())
+    }
+
+    fn get_latest_height(&mut self) {
+        let client = self.celestia_client.inner().clone();
+        self.get_latest_height = Some(tokio::spawn(async move {
+            Ok(client.header_network_head().await?.height())
```
noot, line 149: error context?

crates/astria-conductor/src/client_provider.rs
joroshiba: This pool also solves for reconnect as well right? If each component had their own websocket connection (bad), they would also need their own reconnect logic.

crates/astria-conductor/src/data_availability.rs
noot: if `let first_new_height = self.current_block_height + 1;` this doesn't execute?

crates/astria-conductor/src/conductor.rs
joroshiba: Yeah sync right now occurs solely from sequencer, I agree that there is a use case for DA sync (quite possibly more useful) in the future. The reason to not have this right now is that we haven't gotten everything configured to run against prod celestia DA, and rollups run by 3rd parties can't connect to our custom celestia network right now.

This is done largely to avoid inheriting instability from prelaunch Celestia.

crates/astria-conductor/src/data_availability.rs
noot: can you add a comment explaining `latest_height_res` is the celestia height you're fetching at

crates/astria-conductor/src/conductor.rs
SuperFluffy: The if-block is evaluated before the future is polled. So the arm is only deactivated if and only if there are no more futures in the stream.

joroshiba (approved):

crates/astria-conductor/src/data_availability.rs
```diff
@@ -0,0 +1,382 @@
...
+                    client.get_sequencer_namespace_data(height).await
+                });
+            }
+        }
+    }
+
+    #[instrument(skip(self, sequencer_data_res))]
+    fn process_sequencer_datas(
+        &mut self,
+        height: u64,
```
noot, line 200:
```suggestion
        da_height: u64,
```
or add comment

crates/astria-conductor/src/conductor.rs
SuperFluffy: Rc: see at the top of the block where it's defined. It's to make the future 'static.

The 5 seconds are to let the tasks wind down and exit before being killed. Just a number to be honest. We should probably fine-tune it.

crates/astria-conductor/src/client_provider.rs
SuperFluffy: Not the pool but the ClientProvider.

But yes, if each actor got its own provider, we'd have one independent task per actor each, with each attempting to fix their own websocket connection at probably the same time.

crates/astria-conductor/src/data_availability.rs
SuperFluffy: Yeah, that range would be considered empty.

crates/astria-conductor/src/data_availability.rs
```diff
@@ -0,0 +1,382 @@
...
+            );
+        } else {
+            let verifier = block_verifier.clone();
+            verification_tasks.spawn(
+                block_hash,
+                async move {
+                    verifier
+                        .validate_signed_namespace_data(&data)
+                        .await
+                        .map(|()| data)
```
noot, line 375: why the `map`?

crates/astria-conductor/src/data_availability.rs
```diff
@@ -0,0 +1,382 @@
...
+
+    // ensure that the last sender goes out of scope so that block assembler's exit condition fires
+    std::mem::drop(assembly_tx);
+
+    block_assembler
+        .await
+        .wrap_err("failed assembling sequencer block subsets")
+}
+
+async fn get_rollup(
```
noot, line 308:
```suggestion
async fn get_rollup_data(
```

crates/astria-conductor/src/data_availability.rs
SuperFluffy: Because I want the validation to return `Ok(data)` or `Err(error)` for further processing. Since the return value of the validation step is just `Result<(), E>` this moves the data into the right place. I suppose I can use `?` and `Ok(data)` to make that clearer

crates/astria-conductor/src/executor.rs
```diff
@@ -85,29 +61,23 @@ fn convert_tendermint_to_prost_timestamp(value: Time) -> Result<ProstTimestamp>
...
-    BlockReceivedFromSequencer {
-        block: Box<SequencerBlockData>,
-    },
+    BlockReceivedFromSequencer { block: Box<SequencerBlockData> },
     /// used when a block is received from the reader (Celestia)
-    BlockReceivedFromDataAvailability {
-        block: Box<SequencerBlockSubset>,
-    },
-    Shutdown,
+    FromCelestia(Vec<SequencerBlockSubset>),
```
noot, line 66:
```suggestion
    BlockReceivedFromCelestia(Vec<SequencerBlockSubset>),
```

or update the above enum name for consistency?

crates/astria-conductor/src/executor.rs
```diff
@@ -521,32 +570,27 @@ mod test {
...
+
+        let (_block_tx, block_rx) = mpsc::unbounded_channel();
+        let (_shutdown_tx, shutdown_rx) = oneshot::channel();
+        let mut executor = Executor::new(
+            &cfg.execution_rpc_url,
             chain_id,
-            conf.disable_empty_block_execution,
+            cfg.disable_empty_block_execution,
+            block_rx,
+            shutdown_rx,
```
noot, line 593: the setup for these tests appears to be basically the same in many of them, maybe create a helper func for this?

crates/astria-conductor/src/sequencer/mod.rs
```diff
@@ -0,0 +1,252 @@
...
+            mut shutdown,
+            sync_done,
+        } = self;
+
+        let mut new_blocks: stream::Fuse<NewBlocksStream> = subscribe_new_blocks(pool.clone())
+            .await
+            .wrap_err("failed to start initial new-blocks subscription")?
+            .fuse();
+
+        let mut pending_blocks = FuturesOrdered::new();
```
noot, line 106: it appears only `Ready` futures are ever put in here, which is then used in the `select` statement, what's the purpose of this over just calling `forward_pending_block` directly?

crates/astria-conductor/src/sequencer/mod.rs
```diff
@@ -0,0 +1,252 @@
...
+                // Regular pending blocks will be submitted to the executor in the order they were received.
+                // The condition on `sync` ensures that blocks from the sync process are forwarded first.
+                Some(res) = pending_blocks.next(), if sync.is_terminated() => {
+                    if let Err(e) = forward_pending_block(executor_tx.clone(), res) {
+                        error!("failed forwarding blocks during regular operation; exiting reader");
+                        break 'reader_loop Err(e).wrap_err("failed forwarding blocks regular operation");
+                    }
+                }
+
+                res = &mut resubscribe => {
```
noot, line 188: what is `res` on the first iteration where this is called? since `let mut resubscribe = future::Fuse::terminated();?

crates/astria-conductor/src/sequencer/mod.rs
```diff
@@ -0,0 +1,252 @@
...
+}
+
+fn forward_pending_block(
+    executor_tx: executor::Sender,
+    res: Result<SequencerBlockData, NewBlockStreamError>,
+) -> eyre::Result<()> {
+    let block = match res {
+        Err(e) => {
+            warn!(error.message = %e, error.cause = ?e, "response from sequencer block subscription was bad; dropping it");
+            return Ok(());
```
noot, line 228: shouldn't this case try to resubscribe instead of silently dropping?

crates/astria-conductor/src/sequencer/sync.rs
```diff
@@ -0,0 +1,102 @@
...
+    end: u32,
+    client_pool: Pool<ClientProvider>,
+    executor: crate::executor::Sender,
+) -> eyre::Result<()> {
+    use futures::{
+        FutureExt as _,
+        StreamExt as _,
+    };
+
+    let mut height_stream = futures::stream::iter(start..end);
```
noot, line 44: check if `start > end` and return if so?

crates/astria-sequencer-client/src/extension_trait.rs
```diff
@@ -405,6 +477,32 @@ pub trait SequencerClientExt: Client {
...
+            .map_err(|e| Error::tendermint_rpc("latest_block", e))?;
+        SequencerBlockData::from_tendermint_block(rsp.block)
+            .map_err(|e| Error::deserialization("SequencerBlockData", e))
+    }
+
+    /// Get the sequencer block at the provided height.
+    ///
+    /// This is a convenience method that converts the result [`Client::block`]
+    /// to `SequencerBlockData`.
+    async fn sequencer_block(&self, height: u32) -> Result<SequencerBlockData, Error> {
```
noot, line 497:
```suggestion
    async fn sequencer_block_data(&self, height: u32) -> Result<SequencerBlockData, Error> {
```

crates/astria-conductor/src/client_provider.rs
noot: > You mean having one open websocket connection per component?

yeah, that answers my question :) thanks

noot (approved):
looks good, nice work! approving as to not block. have a few questions/clean up comments only!

crates/astria-sequencer-client/src/extension_trait.rs
SuperFluffy: I want to replace `SequencerBlockData` by just `SequencerBlock` in the short term, which is the nomenclature we originally introduced in our sequencer protos but never implemented.

crates/astria-sequencer-client/src/extension_trait.rs
```diff
@@ -405,6 +477,32 @@ pub trait SequencerClientExt: Client {
         self.get_nonce(address, 0).await
     }
 
+    /// Get the latest sequencer block.
+    ///
+    /// This is a convenience method that converts the result [`Client::latest_block`]
+    /// to `SequencerBlockData`.
+    async fn latest_sequencer_block(&self) -> Result<SequencerBlockData, Error> {
```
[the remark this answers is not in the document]
SuperFluffy: As for `sequencer_block`, I think I want to move to just calling the returned value `SequencerBlock`.

crates/astria-conductor/src/sequencer/sync.rs
SuperFluffy: So this makes use of a bunch of things:
1. assume `start > end`
2. then the iterator impl for `start..end` will always return `None` because it's empty and because it's a `FusedIterator`.
3. this means that `futures::stream::iter` will always resolve to `Ready(None)` immediately
4. this means that the `select!` statement wil poll the stream, immediately get a result, try apply it's result to the `Some(_)` pattern, fail, and hence disable the branch.
5. because the `FuturesOrdered` stream also response with `Ready(None)` if it's empty, its select arm will also be disabled.
6. so because both arms are disabled, we immediately go to `else`

crates/astria-conductor/src/sequencer/mod.rs
SuperFluffy: This case means that a) we got an actual response from the new block stream, but that b) the data we received was bad.

So the stream is still active, but the returned data is bad and we have to drop it.

crates/astria-conductor/src/executor.rs
SuperFluffy: addressed by @sambukowski's commit

crates/astria-conductor/src/client_provider.rs
SuperFluffy: Yeah, you could have an `Arc<ConnectionProvider>` and then `get` connections from it if failed. But that's basically turning into a pool.

crates/astria-conductor/src/executor.rs
SuperFluffy: Moved all of that into a single setup fn.

crates/astria-conductor/src/data_availability.rs
SuperFluffy: I have renamed the function to `fetch_sequencer_blobs_up_to_latest_height, added a comment, and renamed the joinmap to `fetch_sequencer_blobs_at_height`.

crates/astria-conductor/src/data_availability.rs
SuperFluffy: @joroshiba good catch, fixed.

crates/astria-conductor/src/data_availability.rs
SuperFluffy: Renamed it to `verify_sequencer_blobs_and_assemble_rollups` (as well as its joinmap).

crates/astria-conductor/src/data_availability.rs
SuperFluffy: Created an issue

crates/astria-conductor/src/sequencer.rs
SuperFluffy: Created an issue.

crates/astria-conductor/src/data_availability.rs
SuperFluffy: I have renamed the functions. I believe we need not explicitly talk about `da_height` given that we are in the data_availability module (so that the height is explicitly referring to the DA).

crates/astria-conductor/src/data_availability.rs
SuperFluffy: Changed this to `fetch_and_verify_rollup_blobs`

crates/astria-conductor/src/sequencer/mod.rs
SuperFluffy: `pending_blocks` is populated from the new-blocks subscription. However, these are only forwarded after the sync has finished, see the arm in the `select!` macro: `Some(res) = pending_blocks.next(), if sync.is_terminated()`

crates/astria-conductor/src/sequencer/mod.rs
SuperFluffy: `res` would be nothing, since this future will never return anything (if polled, it will yield `Pending` but never `Ready`). I have just added `, if !resubscribe.is_terminated()` to the select arm however to make this clearer.

PR #459, opened 2023-10-04
fix(composer): refetch nonce on submission failure

### LINKED ISSUE #293: composer: refetch nonce on transaction failure due to invalid nonce

### LINKED ISSUE #307: composer: nonce management
Composer currently does not handle transactions that fail due to invalid nonce. The following issues describe the path to solving this. PRs should be implemented in order.

```[tasklist]
### Tasks
- [x] #292 and PR #294
- [ ] #252
- [ ] #293
```

itamarreif:
## Summary
This adds handling for sequencer responses to transaction submission. Failures caused by invalid nonce reset the stored nonce and resubmit the transaction.

## Background
See [#293](https://github.com/astriaorg/astria/issues/293) and [#307](https://github.com/astriaorg/astria/issues/307)

## Changes
- `Executor` methods have been cleaned up and moved around a bit to make error handling more ergonomic
- `SubmitFut` type implements transaction submission, nonce refetching and transaction resubmission as a state machine that is represented using the `SubmitState` enum.
- Transactions that fail with `AbciCode::InvalidNonce` lead to refetching the nonce from the sequencer and resubmit the transaction under the new nonce. 
- Successful transaction submissions and nonce refetching are both logged using spans.
- `AbciCode` method for converting from `tendermint::abci::Code` is added
- The `Executor`'s channel is renamed to `new_bundles`
- We use `sequencer_client::HttpClient` instead of the wrapper type
- Use `tryhard` for exponential backoff instead of `backon`
- Changed the `composer` test object to return a `startup_guard` that needs to be awaited to make sure the startup query guards are only matched in the beginning.

## Testing
- [x] Failed transaction submission due to invalid nonce results in nonce being refetched and the transaction resubmitted.

## Related Issues

closes #293
closes #307


crates/astria-composer/src/searcher/mod.rs
```diff
@@ -130,8 +130,13 @@ impl Searcher {
         let (status, _) = watch::channel(Status::default());
 
         let (executor_tx, executor_rx) = mpsc::channel(256);
-        let executor = Executor::new(&cfg.sequencer_url, &cfg.private_key, executor_rx)
-            .wrap_err("executor construction from config failed")?;
+        let executor = Executor::new(
+            &cfg.sequencer_url,
+            &cfg.private_key,
+            executor_tx.clone(),
```
SuperFluffy, line 136: Why is it holding a channel into itself?

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -120,50 +124,59 @@ impl Executor {
...
+            .sequencer_client
+            .submit_transaction_sync(tx.clone())
             .await
-            .wrap_err("failed submitting signed actions to sequencer")?;
-        Ok(())
+            .map_err(ExeuctionError::SequencerClientFailure)?;
+
+        match AbciCode::from_tendermint(submission_rsp.code) {
+            Some(AbciCode::OK) => Ok(tx),
+            Some(AbciCode::INVALID_NONCE) => Err(ExeuctionError::InvalidNonce {
```
SuperFluffy, line 165: Typo: `ExeuctionError`

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -177,17 +190,49 @@ impl Executor {
...
+                            nonce,
+                            "invalid nonce returned from sequencer; retrieving new nonce and \
+                             resubmitting the transaction"
+                        );
+                        self.nonce = None;
+                        let bundle = {
+                            let (_, _, unsigned) = transaction.into_parts();
+                            unsigned.actions
+                        };
+                        self.handle_bundle_resubmission(bundle).await;
```
SuperFluffy, line 209: This can potentially block indefinitely if I see it right.

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -249,6 +294,19 @@ impl Executor {
         Ok(())
     }
 }
+
+enum ExeuctionError {
+    SequencerClientFailure(eyre::Report),
```
SuperFluffy, line 299: I recommend not creating new error types that wrap an apex-error like `eyre::Report`. It's also not necessary because AFAIK all of sequencer client's errors are fully typed., so it's just this module that inexplicably tosses an `eyre::Report` in the middle of it. :)

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -84,6 +86,7 @@ impl Executor {
     pub(super) fn new(
         sequencer_url: &str,
         private_key: &SecretString,
+        transaction_resubmission_tx: mpsc::Sender<Vec<Action>>,
```
SuperFluffy, line 89: This seems like a code smell. If you need a queue of some sort, it's better to use `FuturesUnordered` (or probably maybe `FuturesOrdered`).

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -120,50 +124,59 @@ impl Executor {
...
-            bail!(
-                "submitting transaction to sequencer returned with error code; code: `{code}`; \
-                 log: `{log}`; hash: `{hash}`",
-                code = rsp.code.value(),
-                log = rsp.log,
-                hash = rsp.hash,
-            );
+            Ok(curr_nonce)
+        } else {
+            let rsp = self.sequencer_client.get_latest_nonce(self.address).await?;
```
SuperFluffy, line 134: See my other comment: it's not necessary to wrap this in an `eyre::Result`. Why not just return the error returned by `get_latest_nonce` itself?

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -120,50 +124,59 @@ impl Executor {
...
     }
 
-    /// Signs and submits the bundle of actions to the sequencer.
-    async fn sign_and_submit(&mut self, actions: Vec<Action>) -> eyre::Result<()> {
+    /// Populates nonce, signs and submits the bundle of actions to the sequencer. Returns the
+    /// transaction if successfully submitted.
+    async fn execute_bundle(
+        &mut self,
+        bundle: Vec<Action>,
+    ) -> Result<SignedTransaction, ExeuctionError> {
```
noot, line 145: don't return the tx here, it appears it's only used for logging anyways, so just log in here

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -120,50 +124,59 @@ impl Executor {
...
             nonce,
-            actions,
+            actions: bundle,
         }
         .into_signed(&self.sequencer_key);
 
-        self.submit_tx(tx)
+        let submission_rsp = self
+            .sequencer_client
+            .submit_transaction_sync(tx.clone())
```
noot, line 159: can remove the `clone` if you don't return the tx

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -120,50 +124,59 @@ impl Executor {
...
+            _ => Err(ExeuctionError::UnknownDeliverTxFailure {
+                code: submission_rsp.code,
+                transaction: tx,
+            }),
+        }
+    }
+
+    async fn handle_bundle_resubmission(&self, bundle: Vec<Action>) {
+        if let Err(e) = self.transaction_resubmission_tx.send(bundle).await {
+            error!(error.msg = %e, error.cause_chain=?e, "failed to resubmit transaction");
```
noot, line 178:
```suggestion
            error!(error.msg = %e, error.cause_chain=?e, "transaction resubmission channel receiver dropped");
```

more accurate i think?

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -50,8 +50,10 @@ use tracing::{
 pub(super) struct Executor {
     // The status of this executor
     status: watch::Sender<Status>,
-    // Channel for receving bundles to pack, sign, and submit
+    // Channel receiver for bundles to pack, sign, and submit
     executor_rx: mpsc::Receiver<Vec<Action>>,
+    // Channel sender for resubmitting transactions that failed submission due to failed nonce
+    transaction_resubmission_tx: mpsc::Sender<Vec<Action>>,
```
noot, line 56: it appears this is the sender counterpart of `executor_rx`?  maybe add a comment for that?

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -177,17 +190,49 @@ impl Executor {
...
-                break;
+            match self.execute_bundle(bundle).await {
+                Err(e) => match e {
+                    ExeuctionError::InvalidNonce {
+                        nonce,
+                        transaction,
+                    } => {
+                        warn!(
+                            nonce,
+                            "invalid nonce returned from sequencer; retrieving new nonce and \
```
noot, line 201:
```suggestion
                            "invalid nonce error returned from sequencer; retrieving new nonce and \
```

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -176,18 +227,75 @@ impl Executor {
...
+                    match ret {
+                        Ok(unsigned_tx) => {
+                            // update self.nonce for next transaction
+                            *Arc::get_mut(&mut self.nonce).expect("should only be called once at a time") =
+                                Some(unsigned_tx.nonce + 1);
+
+                            submission_fut = self.sign_and_submit_fut(unsigned_tx);
+                        }
+                        Err(e) => {
+                            error!(error.msg = %e, "failed to retrieve nonce from sequencer; executor shutting down");
```
SuperFluffy, line 250:
```suggestion
                            let error = &e as &(dyn std::error::Error + "static);
                            error!(error, "failed to retrieve nonce from sequencer; shutting down");
```

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -176,18 +227,75 @@ impl Executor {
...
+                        Err(ExecutionError::InvalidNonce(transaction)) => {
+                            let nonce = transaction.unsigned_transaction().nonce;
+                            warn!(
+                                nonce,
+                                "invalid nonce error returned from sequencer; retrieving new nonce and \
+                                 resubmitting the transaction"
+                            );
+                            nonce_fut = self.bundle_resubmission_from_transaction(transaction);
+                        }
+                        Err(ExecutionError::UnknownDeliverTxFailure {
```
SuperFluffy, line 267: This error and the one below look similar - why are we exiting in one case and not in the other?

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -176,18 +227,75 @@ impl Executor {
...
+                            code,
+                            transaction,
+                        }) => {
+                            warn!(
+                                code=?code,
```
SuperFluffy, line 272: Please don't use debug formatting in events, it's very noisy. AFAIK `AbciCode` implements `Display`:
```suggestion
                                %code,
```

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -176,18 +227,75 @@ impl Executor {
...
+                                transaction = ?transaction,
```
SuperFluffy, line 273: As above: no debug logging. Try and identify the transaction differently. In the worst case scenario, just use something like base64 or hex encoding.

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -176,18 +227,75 @@ impl Executor {
...
+                                "unknown error code returned from sequencer; skipping this \
+                                 transaction"
+                            );
+                        }
+                        Err(ExecutionError::TransactionSubmissionFailed {
+                            error: e,
+                            transaction
+                        }) => {
+                            error!(
+                                error.msg = %e,
```
SuperFluffy, line 283: See my comment above on printing this using a trait object

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -176,18 +227,75 @@ impl Executor {
...
+                                transaction = ?transaction,
```
SuperFluffy, line 284: no debug formatting in events

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -176,18 +227,75 @@ impl Executor {
...
+                                "failed to submit transaction to sequencer; executor shutting down");
+                            break;
+                        }
+                        Ok(tx) => {
+                            let nonce = tx.unsigned_transaction().nonce;
+                            info!(
```
SuperFluffy, line 290: for both fields: no debug logging

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -176,18 +227,75 @@ impl Executor {
...
+                                tx = ?tx,
+                                nonce = ?nonce,
+                                "transaction submitted to sequencer successfully with nonce {}",
```
SuperFluffy, line 293: remove the format `{}`. You have the nonce as a field

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -271,27 +403,30 @@ impl SequencerClient {
...
+        address: Address,
+    ) -> Result<NonceResponse, SequencerClientError> {
         tokio::time::timeout(
             Duration::from_secs(1),
             self.inner.get_latest_nonce(address.0),
         )
         .await
-        .wrap_err("request timed out")?
-        .wrap_err("RPC returned with error")
+        .map_err(|_e| SequencerClientError::Timeout)?
```
SuperFluffy, line 415: Don't ignore the error. I know it doesn't add a lot of info, but it's easier to understand if you don't.

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -271,27 +403,30 @@ impl SequencerClient {
...
-    ) -> eyre::Result<tx_sync::Response> {
+    ) -> Result<tx_sync::Response, SequencerClientError> {
         tokio::time::timeout(
             Duration::from_secs(1),
             self.inner.submit_transaction_sync(signed_tx),
         )
         .await
-        .wrap_err("request timed out")?
-        .wrap_err("RPC returned with error")
+        .map_err(|_e| SequencerClientError::Timeout)?
```
SuperFluffy, line 429: As above: don't drop the error.

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -252,6 +360,30 @@ impl Executor {
         Ok(())
     }
 }
+
+#[derive(Debug, thiserror::Error)]
+enum ExecutionError {
+    #[error("failed to submit sequencer transaction")]
+    TransactionSubmissionFailed {
+        error: SequencerClientError,
```
SuperFluffy, line 368: call this `source` (preferred) or add a `#[source]` attribute. See the thiserror docs

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -252,6 +360,30 @@ impl Executor {
...
+        transaction: SignedTransaction,
```
SuperFluffy, line 369: You don't seem to be doing anything with these transactions but only use them for reporting. If that's the case start a tracing span with a field identifying the tx instead of moving the object around

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -176,18 +227,75 @@ impl Executor {
...
-                    "failed submitting the bundle to the sequencer; bailing",
-                );
-                break;
+        let mut submission_fut = Fuse::terminated();
+        let mut nonce_fut = Fuse::terminated();
+        loop {
+            select! {
+                // receive new bundle for processing
+                Some(bundle) = self.executor_rx.recv(), if nonce_fut.is_terminated() && submission_fut.is_terminated() => {
+                    debug!(bundle = ?bundle, "executor received bundle for processing");
```
SuperFluffy, line 236: Please don't do debug logging, even in `debug!` events. They break the json subscriber.

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -120,50 +137,84 @@ impl Executor {
...
-    /// Gets the next nonce to sign over if it exists and increments the stored nonce counter
-    fn get_and_increment_nonce(&mut self) -> Option<u32> {
-        self.nonce.map(|curr_nonce| {
-            self.nonce = Some(curr_nonce + 1);
-            curr_nonce
-        })
-    }
+    /// Returns a fused future that attaches the nonce to the given bundle. If the stored nonce is
+    /// `None`, the future will retrieve the latest nonce from the sequencer.
+    fn attach_nonce_fut(&mut self, bundle: Vec<Action>) -> NonceFut {
```
SuperFluffy, line 142: Just call this `fn attach_nonce`

crates/astria-composer/tests/blackbox/composer.rs
```diff
@@ -98,6 +174,54 @@ async fn mount_broadcast_tx_sync_mock(
...
+        let Some(sequence_action) = sent_action.as_sequence() else {
+            panic!("mocked sequencer expected a sequence action");
+        };
+        sequence_action.chain_id == expected_chain_id.as_bytes()
+    };
+    let jsonrpc_rsp = response::Wrapper::new_with_id(
+        Id::Num(1),
+        Some(tx_sync::Response {
+            code: AbciCode::INVALID_NONCE.into(),
+            data: vec![].into(),
```
SuperFluffy, line 216: Random idea. We could respond with the next valid nonce from sequencer

crates/astria-composer/tests/blackbox/composer.rs
```diff
@@ -55,9 +71,69 @@ async fn tx_from_two_rollups_are_received_by_sequencer() {
...
+        &test_composer.sequencer,
+        "accounts/nonce",
+        NonceResponse {
+            height: 0,
+            nonce: 1,
+        },
+    )
+    .await;
+
+    // Expect nonce 0 again so that the resubmitted tx is accepted
```
SuperFluffy, line 103: Nonce of 1 right, not 0? That's what the mount before does

crates/astria-composer/tests/blackbox/composer.rs
```diff
@@ -55,9 +71,69 @@ async fn tx_from_two_rollups_are_received_by_sequencer() {
...
+    let valid_nonce_guard =
+        mount_broadcast_tx_sync_mock(&test_composer.sequencer, "test1", 1).await;
+
+    // Push a tx to the rollup node so that it is picked up by the composer and submitted with the
+    // stored nonce of 42, triggering the nonce refetch process
```
SuperFluffy, line 108: The above comment mentions 0, and before that a nonce response with 1 is mounte


Where does 42 come from?

SuperFluffy (commented):
I like the tests, although they are a bit confusingly commented

crates/astria-composer/tests/blackbox/composer.rs
itamarreif: See issue #529: https://github.com/astriaorg/astria/issues/529

crates/astria-composer/tests/blackbox/composer.rs
itamarreif: yeah my bad this was an old comment. fixed.

crates/astria-composer/src/searcher/executor.rs
itamarreif: This is for an unknown AbciCode case that is returned, the other is for transactions that failed to even be sent to the sequencer. If you have suggestions for better naming they are highly welcome :)

crates/astria-composer/src/searcher/executor.rs
itamarreif: this is a `tendermint::abci::Code` though, which does not implement Display. Not sure what the solution here is. Should this be cast into an `AbciCode` at some point?

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -50,15 +65,16 @@ use tracing::{
...
     executor_rx: mpsc::Receiver<Vec<Action>>,
     // The client for submitting wrapped and signed pending eth transactions to the astria
     // sequencer.
     sequencer_client: SequencerClient,
     // Private key used to sign sequencer transactions
     sequencer_key: SigningKey,
-    // Nonce of the sequencer account we sign with
-    nonce: Option<u32>,
+    // Nonce of the sequencer account we sign with. Arc is for static futures. Should only be held
+    // by task at a time.
```
noot, line 76:
```suggestion
    // Nonce of the sequencer account we sign with. Arc is for static futures. Should only be held
    // by one task at a time.
```
typo?

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -176,18 +227,74 @@ impl Executor {
...
+        let mut submission_fut = Fuse::terminated();
+        let mut nonce_fut = Fuse::terminated();
+        loop {
+            select! {
+                // receive new bundle for processing
+                Some(bundle) = self.executor_rx.recv(), if nonce_fut.is_terminated() && submission_fut.is_terminated() => {
+                    debug!(bundle = ?bundle, "executor received bundle for processing");
+                    nonce_fut = self.attach_nonce(bundle);
+                },
+                // attach nonce
```
noot, line 239: wrong place for comment

crates/astria-composer/src/searcher/executor.rs
noot: yeah maybe use the hash (same for errs below)

crates/astria-composer/src/searcher/executor.rs
noot: use `AbciCode::from_tendermint()`?

noot (approved):
approving cause logic looks fine, pls address comments before merging!

crates/astria-composer/src/searcher/executor.rs
itamarreif: not every `tendermint::abci::Code` can be mapped to an `AbciCode` though

crates/astria-composer/src/searcher/executor.rs
itamarreif: Line 237 assigns the future, but line 240 is where it's awaited - this is the correct place for this comment

SuperFluffy:
I pushed a change replacing the ping-pong in the `select!` loop with a typed future that goes between the states `NotStarted -> WaitForSubmit <-> WaitForNonce`. Rather than cloning around tracing spans I realized that it would be easiest to just have everything run under one future.

All of this was also necessary because expressing retries (using `tryhard`) was much easier by tucking everything into one future.

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -120,178 +139,266 @@ impl Executor {
...
+    ///
+    /// If the sequencer returned a non-zero abci code (albeit not `INVALID_NONCE`), this future will return with
+    /// that nonce it used to submit the non-zero abci code request.
+    struct SubmitFut {
+        client: sequencer_client::HttpClient,
+        address: Address,
+        nonce: u32,
+        signing_key: SigningKey,
+        #[pin]
+        state: SubmitState,
```
noot, line 299: is this the only field that needs to be pinned b/c it contains pinned futures?

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -120,178 +139,266 @@ impl Executor {
...
+    .with_config(retry_config)
+    .await
+    .wrap_err("failed submitting a signed transaction after 1024 attempts")
+}
 
-        self.status.send_modify(|status| {
-            status.is_connected = true;
-        });
-        Ok(())
+// TODO: zeroize this
```
noot, line 282: why? cause signing key?

crates/astria-composer/src/searcher/executor.rs
SuperFluffy: Yeah. All other fields are either Unpin or don't need pinning because they are not mutated.

crates/astria-composer/src/searcher/executor.rs
SuperFluffy: Yeah, forgot. Thank you flagging this. I'll derive ZeroizeOnDrop before merge

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -120,178 +139,266 @@ impl Executor {
...
+                        );
+                        return Poll::Ready(
+                            Err(e).wrap_err("failed submitting transaction to sequencer"),
+                        );
+                    }
+                },
+
+                SubmitStateProj::WaitingForNonce {
+                    fut,
+                } => match ready!(fut.poll(cx)) {
```
noot, line 379: if the `waiting_for_nonce` future isn't ready, then the state is unchanged right?

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -120,178 +139,266 @@ impl Executor {
...
-        )
-        .await
-        .wrap_err("request timed out")?
-        .wrap_err("RPC returned with error")
-    }
+impl Future for SubmitFut {
+    type Output = eyre::Result<u32>;
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut std::task::Context<'_>) -> Poll<Self::Output> {
+        loop {
```
noot, line 323: is it ok to loop in a poll function like this? doesn't this block the async runtime? i thought you're supposed to return `Poll::Pending` in the cases where the future isn't ready (for example here, if the `waiting_for_nonce` future is pending)

crates/astria-composer/src/searcher/executor.rs
SuperFluffy: Exactly. The macro short circuits, returning `Poll::Pending` if the underlying future is not ready

crates/astria-composer/src/searcher/executor.rs
noot: i guess this *should* be fine because eventually the tx will submit and we'll get a response which will return `Poll::Ready`, but i'm suspect of the case where we end up stuck in a loop of continuously submitting with invalid nonces.

SuperFluffy: The bit that makes this all work is this" https://docs.rs/futures/latest/futures/macro.ready.html

Have a look at the source: the loop returns early if the currently polled futures aren't ready

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -120,178 +139,266 @@ impl Executor {
...
+                    warn!(
+                        attempt,
+                        wait_duration,
+                        error,
+                        "failed getting latest nonce from sequencer; retrying after backoff",
+                    );
+                }
+                .instrument(span)
+            },
         );
```
noot, line 270: `retry_config` appears to be the same as above - can you make a function for it

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -120,178 +139,266 @@ impl Executor {
...
+                    .map(tracing::field::display);
+                let err = err.clone();
+                let span = warn_span!(parent: span.clone(), "report attempt failure");
+                async move {
+                    let error = &err as &(dyn std::error::Error + 'static);
+                    warn!(
+                        attempt,
+                        wait_duration,
+                        error,
+                        "failed getting latest nonce from sequencer; retrying after backoff",
```
noot, line 265: log is wrong
```suggestion
                        "failed submitting tx to sequencer; retrying after backoff",
```

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -120,178 +139,266 @@ impl Executor {
...
-                error!(
-                    error.message = %e,
-                    error.cause_chain = ?e,
-                    "failed submitting the bundle to the sequencer; bailing",
-                );
-                break;
+                rsp = &mut submission_fut, if !submission_fut.is_terminated() => {
+                    match rsp {
+                        Ok(new_nonce) => self.nonce = new_nonce,
+                        Err(_e) => bail!("eyeyey"),
```
noot, line 159: ???

crates/astria-composer/src/searcher/executor.rs
SuperFluffy: > i guess this *should* be fine because eventually the tx will submit and we'll get a response which will return `Poll::Ready`, but i'm suspect of the case where we end up stuck in a loop of continuously submitting with invalid nonces.

Just saw the edit after you submitted:

Yeah, that would indeed happen. However, that case would be in principle fine in a service. It would not do a lot of useful work but continuously try and resubmit.

The point here is that it shouldn't be too smart: if it gets a nonce from sequencer, then tries to submit the tx, but gets told the nonce is invalid, then something is wrong in other services, but it's ok from the pov of composer

crates/astria-composer/src/searcher/executor.rs
SuperFluffy: :( ty for flagging this. I overlooked it

crates/astria-composer/src/searcher/executor.rs
noot: or does `ready!` wait for the future to be ready? the docs on this macro are kinda bad lol

noot: wait nvm - this returns `Poll::Pending` lol, i get it now

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -120,178 +139,266 @@ impl Executor {
...
+                        client: self.sequencer_client.clone(),
+                        address: self.address,
+                        nonce: self.nonce,
+                        signing_key: self.sequencer_key.clone(),
+                        state: SubmitState::NotStarted,
+                        bundle,
+                    }
+                    .instrument(span)
+                    .fuse();
+                }
```
noot, line 184: doesn't this overwrite `submission_fut` is we get successive bundles from `executor_rx` before the previous `submission_fut` is done?

SuperFluffy: No, because the executor_ex is conditional on the submit fut not being terminated (hence the reason why we use fuses).

noot: missed that haha

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -120,178 +139,266 @@ impl Executor {
...
-        info!(
-            nonce_response.nonce,
-            "retrieved initial nonce from sequencer successfully"
+/// Queries the sequencer for the latest nonce with an exponential backoff
+#[instrument(
+    name = "submit signed transaction",
+    skip_all,
+    fields(
+        nonce = tx.unsigned_transaction().nonce,
+        transaction.hash = hex::encode(tx.to_raw().encode_to_vec()),
```
noot, line 238: this isn't the tx hash

noot (approved):
just a few comments to address before merge!

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -120,178 +139,266 @@ impl Executor {
...
+                    }
+                }
+
+                // receive new bundle for processing
+                Some(bundle) = self.executor_rx.recv(), if submission_fut.is_terminated() => {
+                    // TODO(https://github.com/astriaorg/astria/issues/476): Attach the hash of the
+                    // bundle to the span. Linked GH issue is for agreeing on a hash for `SignedTransaction`,
+                    // but both should be addressed.
+                    let span =  info_span!(
+                        "execute bundle",
```
noot, line 169:
```suggestion
                        "submit bundle",
```

crates/astria-composer/src/searcher/executor.rs
SuperFluffy: In fact, `AbciCode` did not implement `Display`. We are printing with `AbciCode::value` now.

Also, I renamed `AbciCode::from_tendermint` to `AbciCode::from_cometbft`.

crates/astria-composer/src/searcher/executor.rs
SuperFluffy: As you flagged in another comment: the warn message was wrong, so now the config are almost the same, but stil different.

crates/astria-composer/src/searcher/executor.rs
SuperFluffy: fixed

crates/astria-composer/src/searcher/executor.rs
SuperFluffy: I am not properly returning the error. I now also distinguish between "submission" as the whole process of attaching nonce + signing + sending, and just "sending" as the last step of that process. That way it's clearer what failed.

crates/astria-composer/src/searcher/executor.rs
SuperFluffy: Applying sha256 before recording.

crates/astria-composer/src/searcher/executor.rs
SuperFluffy: fixed

crates/astria-composer/src/searcher/executor.rs
SuperFluffy: logging the tx hash now

crates/astria-composer/src/searcher/executor.rs
SuperFluffy: logging the hash of the signed tx now

crates/astria-composer/src/searcher/executor.rs
SuperFluffy: Changed all error reporting in executor to do just that

crates/astria-composer/src/searcher/executor.rs
SuperFluffy: No longer applicable because the state machine in the `SubmitFut` future is selecting on the abci code immediately instead of producing an error.

crates/astria-composer/src/searcher/executor.rs
SuperFluffy: no more formatted event strings

crates/astria-composer/tests/blackbox/composer.rs
SuperFluffy: Addressed by opening the issue

crates/astria-composer/src/searcher/executor.rs
SuperFluffy: removed all debug formatting

SuperFluffy:
Addressed all comments. Waiting for the @itamarreif as the original author to give the ok to merge this.

The PR text should be updated before merge.

SuperFluffy (approved):

PR #460, opened 2023-10-04
composer: log collected txs

itamarreif:
## Summary
This just adds logging for incoming transactions to make debugging easier.


crates/astria-composer/src/searcher/collector.rs
```diff
@@ -123,13 +123,18 @@ impl Collector {
...
+                        inner: tx.clone(),
                     },
                     Duration::from_millis(500),
                 )
                 .await
             {
-                Ok(()) => {}
+                Ok(()) => {
+                    info!(
+                        transaction.hash = %tx.hash,
```
noot, line 134: define `let hash = tx.hash` above the `match searcher_channel` line, so you don't have to `clone()` the `tx`

joroshiba: would it make sense to just have the whole log before the match?

This is only logged if sending to the searcher is succesful, some txs might not be logged.

crates/astria-composer/src/searcher/collector.rs
```diff
@@ -123,13 +123,18 @@ impl Collector {
...
+                        chain_id = %chain_id,
+                        "collected transaction with hash `{}` from rollup with chain_id `{}`", tx.hash, chain_id);
```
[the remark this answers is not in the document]
itamarreif: good point, moved it to before sending to the channel

crates/astria-composer/src/searcher/collector.rs
itamarreif: yeah my bad here, fixed

joroshiba (approved):

crates/astria-composer/src/searcher/collector.rs
```diff
@@ -119,6 +119,7 @@ impl Collector {
             .await
             .wrap_err("failed to subscribe eth client to full pending transactions")?;
         while let Some(tx) = tx_stream.next().await {
+            debug!(transaction.hash = %tx.hash, "collected transaction with from rollup");
```
noot, line 122:
```suggestion
            debug!(transaction.hash = %tx.hash, "collected transaction from rollup");
```

noot: do you want to log which rollup (or is there only one)?

noot (approved):

SuperFluffy (approved):

PR #469, opened 2023-10-06
fix(conductor): fix action tree root inclusion proof verification

noot:
## Summary
fix action tree root inclusion proof verification in conductor. the problem is that the input leaves of the `data_hash` tree are the transaction *hashes*, not the transactions itself, but we were passing `action_tree_root` (the tx itself, not the hash) as the leaf to be verified.

## Changes
- hash `action_tree_root` in conductor before verifying proof of inclusion in `data_hash`

## Testing
unit test which confirms the verification is now correct


crates/astria-conductor/src/block_verifier.rs
```diff
@@ -193,8 +195,9 @@ fn validate_sequencer_namespace_data(
     let Some(data_hash) = header.data_hash else {
         bail!("data hash should not be empty");
     };
+    let action_tree_root_hash = sha2::Sha256::digest(action_tree_root);
     action_tree_root_inclusion_proof
-        .verify(action_tree_root, data_hash)
+        .verify(&action_tree_root_hash, data_hash)
         .wrap_err("failed to verify action tree root inclusion proof")?;
```
noot, line 201: this is where the fix is happening

crates/astria-conductor/src/block_verifier.rs
```diff
@@ -193,8 +195,9 @@ fn validate_sequencer_namespace_data(
     let Some(data_hash) = header.data_hash else {
         bail!("data hash should not be empty");
     };
+    let action_tree_root_hash = sha2::Sha256::digest(action_tree_root);
     action_tree_root_inclusion_proof
-        .verify(action_tree_root, data_hash)
+        .verify(&action_tree_root_hash, data_hash)
```
SuperFluffy, line 200: Somehow this is very surprising to me. As a user, my expectation is that I can pass in the action_tree_root as-is without needing to further manipulate it.

After all, isn't the root just the recursive hash of the tree?

SuperFluffy (approved):
Approving because this fixes the problem, but putting my user-hat on this is very surprising.

sambukowski:
When running tests locally I'm getting some test fails:
```
failures:
    block_verifier::test::validate_rollup_data_ok
    block_verifier::test::validate_sequencer_namespace_data_last_commit_none_ok

test result: FAILED. 11 passed; 2 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
```

```
thread 'block_verifier::test::validate_sequencer_namespace_data_last_commit_none_ok' panicked at 'called `Result::unwrap()` on an `Err` value: failed to verify action tree root inclusion proof

Caused by:
    this root hash doesn't match the proof's root hash w.r.t. the given item

Location:
    crates/astria-conductor/src/block_verifier.rs:201:10', crates/astria-conductor/src/block_verifier.rs:465:10
```


```
thread 'block_verifier::test::validate_sequencer_namespace_data_last_commit_none_ok' panicked at 'called `Result::unwrap()` on an `Err` value: failed to verify action tree root inclusion proof

Caused by:
    this root hash doesn't match the proof's root hash w.r.t. the given item
    
Location:
    crates/astria-conductor/src/block_verifier.rs:201:10', crates/astria-conductor/src/block_verifier.rs:515:10
```

sambukowski:
Ignore my last comment, I pulled an old commit. 
Looks good.

sambukowski (approved):

crates/astria-conductor/src/block_verifier.rs
noot: copy pasting my response in slack:

check this function here: https://github.com/astriaorg/astria/blob/main/crates/astria-sequencer-types/src/sequencer_block_data.rs#L363 this is used in SequencerBlockData::from_tendermint_block

the verification that was broken was the inclusion of the action_tree_root (which is currently the first tx in a block) inside data_hash (which is the root of the tree where the leaves are the transaction hashes)

 since the input to create the tree is the transaction hashes and not the transactions, and action_tree_root is a transaction, we need to hash it before verifying it was included in the data_hash tree

PR #477, opened 2023-10-10
ci: bump rust to 1.72

dereksione:
## Summary
Update Rust from 1.70 to 1.72 in CI and Docker.

## Background

1.70 was giving the error

```bash
cargo +1.70 hack check --all --all-features
info: running `cargo check --all-features` on astria-celestia-jsonrpc-client (1/14)
error: Package `astria-utils v0.1.0 (/Users/astria/dev/astria/astria/crates/astria-utils)` does not have feature `figment`. It has an optional dependency with that name, but that dependency uses the "dep:" syntax in the features table, so it does not have an implicit feature with that name.
error: process didn't exit successfully: `/Users/astria/.rustup/toolchains/1.70-aarch64-apple-darwin/bin/cargo check --all-features --manifest-path crates/astria-celestia-jsonrpc-client/Cargo.toml` (exit status: 101)
```

### DIFF
--- a/containerfiles/Dockerfile
+++ b/containerfiles/Dockerfile
@@ -1,5 +1,5 @@
 # build stage
-FROM --platform=$BUILDPLATFORM lukemathwalker/cargo-chef:0.1.61-rust-1.70.0-bookworm AS chef
+FROM --platform=$BUILDPLATFORM lukemathwalker/cargo-chef:0.1.62-rust-1.72.0-bookworm AS chef
 
 WORKDIR /build/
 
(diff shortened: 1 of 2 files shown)


SuperFluffy (commented):
Approving, but please add the error message you saw while developing #428

containerfiles/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # build stage
-FROM --platform=$BUILDPLATFORM lukemathwalker/cargo-chef:0.1.61-rust-1.70.0-bookworm AS chef
+FROM --platform=$BUILDPLATFORM lukemathwalker/cargo-chef:0.1.61-rust-1.72.0-bookworm AS chef
```
joroshiba, line 2: Will need to update to cargo-chef 0.1.62 I believe

joroshiba (approved):
Added note for fixing failing docker builds.

Approved but with question of: should we be updating to latest release `1.73` instead of `1.72`?

SuperFluffy (approved):

PR #480, opened 2023-10-10
fix(sequencer): move transaction execution to prepare/process proposal

### LINKED ISSUE #478: `sequence::Action`s that fail checks should not be included in block data
currently, transactions (block data) are executed after the block has already been committed by the network. this results in transactions that may fail ending up in the block. for most blockchains, this is fine, as having a failed tx in the block doesn't matter as it didn't affect the state. however, in our case, for `sequence::Action`s, this is bad because we only want data to be sequenced (ie put in a block) if a fee is paid. including `sequence::Action`s in a block which fail the fee payment basically allows data to be sequenced for free.

the solution for this is to perform stateful checks on transactions including `sequence::Action`s before the block is committed by the network, ie. in `prepare_proposal` and `process_proposal`.

noot:
## Summary
We don't want transactions (specifically `sequence::Action`s) which fail fee payment to end up in the block data, as they would essentially get sequenced for free. The easiest way to deal with this is to execute all transactions in a block *before* the block is committed by the tendermint network, and only include txs that successfully execute, which is what this PR does.

## Background
see #478

## Changes
- update `prepare_proposal` and `process_proposal` to execute transactions instead of later on in the usual `deliver_tx` loop.
- the results of tx execution are cached in the `App` and returned when `deliver_tx` is called later.
- as a result, `begin_block` is actually executed *after* all the transactions are executed. this is fine, since the only state change `begin_block` can do at the moment is removal of byzantine validators (plus updating block height + timestamp), none of which affect tx execution.

## Testing
unit tests

## Breaking Changelist
any tx which fails execution will no longer be included in a block's tx data.

## Related Issues

closes #478


crates/astria-sequencer/src/service/consensus.rs
```diff
@@ -200,86 +242,148 @@ impl Consensus {
...
-        txs: res.into_transactions(),
+    /// Generates a commitment to the `sequence::Actions` in the block's transactions.
+    /// This is required so that a rollup can easily verify that the transactions it
+    /// receives are correct (ie. we actually included in a sequencer block, and none
+    /// are missing)
+    /// It puts this special "commitment" as the first transaction in a block.
+    /// When other validators receive the block, they know the first transaction is
+    /// supposed to be the commitment, and verifies that is it correct.
+    #[instrument(skip(self))]
+    async fn handle_prepare_proposal(
```
noot, line 264: I'm going to turn these into methods on `app` actually now that they're stateful. kinda messy to have them here

sambukowski (approved):

crates/astria-sequencer/src/service/consensus.rs
```diff
@@ -150,6 +145,22 @@ impl Consensus {
         })
     }
 
+    #[instrument(skip(self))]
+    async fn handle_prepare_proposal(
+        &mut self,
+        prepare_proposal: request::PrepareProposal,
+    ) -> response::PrepareProposal {
+        self.app.prepare_proposal(prepare_proposal).await
```
SuperFluffy, line 153: My first reaction to this was negative, but looking at the code structure it looks like `App` is actually a part of the `Consensus` service and only used by it. Should we make `app.rs` a submodule of it? Why even have the `App` type at all? It feels like it's not necessary unless I am missing something?

crates/astria-sequencer/src/app.rs
```diff
@@ -113,7 +137,137 @@ impl App {
...
+
+    /// Generates a commitment to the `sequence::Actions` in the block's transactions
+    /// and ensures it matches the commitment created by the proposer, which
+    /// should be the first transaction in the block.
+    #[instrument(name = "App::process_proposal", skip(self))]
+    pub(crate) async fn process_proposal(
+        &mut self,
+        process_proposal: abci::request::ProcessProposal,
+    ) -> anyhow::Result<()> {
+        // clear the cache of transaction execution results
```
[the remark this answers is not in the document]
noot: you mean why is this in both prepare and process proposal? prepare_proposal is called only on the proposer node while process_proposal is called on all other validators, so yeah it has to be in both functions

crates/astria-sequencer/src/app.rs
```diff
@@ -113,7 +137,137 @@ impl App {
         Ok(())
     }
 
-    #[instrument(name = "App:begin_block", skip(self))]
+    /// Generates a commitment to the `sequence::Actions` in the block's transactions.
+    /// This is required so that a rollup can easily verify that the transactions it
+    /// receives are correct (ie. we actually included in a sequencer block, and none
+    /// are missing)
+    /// It puts this special "commitment" as the first transaction in a block.
```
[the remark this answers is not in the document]
noot: this comment was from before, it's regarding the `action_tree_root`, and yeah this is application-specific so cometbft doesn't/can't do this

crates/astria-sequencer/src/service/consensus.rs
noot: the other consensus service functions all call into `app` the same way. and yes it's only used by the consensus service, however I think we should definitely keep `app` separate as it cleanly separates out the application execution logic from all other logic (eg. the consensus service, which is more closely tied to the ABCI server). the `app` also handles the state delta (the successive state changes which occur during execution)

crates/astria-sequencer/src/app.rs
```diff
@@ -113,7 +137,137 @@ impl App {
...
+            .context("no chain IDs commitment in proposal")?
+            .to_vec()
+            .try_into()
+            .map_err(|_| anyhow!("chain IDs commitment must be 32 bytes"))?;
+
+        let expected_txs_len = txs.len();
+
+        let (signed_txs, txs_to_include) = self.execute_block_data(txs.into()).await;
+
+        // all txs in the proposal should be deserializable and executable
```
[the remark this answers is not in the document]
noot: i'll expand on it - if any txs in `txs` fail to deserialize or execute, then `txs_to_include` won't include the same number of txs as `txs`

crates/astria-sequencer/src/service/consensus.rs
SuperFluffy: Yeah, so the reason I wrote the comment was that I thought `App` is a concept/object that operates at a higher hierarchy than `Consensus`.

So it felt wrong that it's passed to `Consensus` which then gets to manipulate it. It felt like an inversion of responsibility (higher level delegating to lower level).

But it turns out that it's the other way round: `App` here is a part of `Consensus`, so the latter delegates to the former.

I'm fine leaving it as its own type, but IMO it should be defined under the consensus module.

SuperFluffy (approved):
Questions were answered, happy to approve.

Consider moving app.rs under consensus.rs. otherwise no blockers

Issue #482, opened 2023-10-11
conductor: simulate bad websocket connections


SuperFluffy:

jbowen93:
This is a good indicator of something where testing in isolation is difficult or impossible and "soak testing" is the best way to validate if a change fixes something. I've run enough different instances of rollups now to say that the conductor crashes and fails to reconnect with high likelihood in <8 hours. 

Would be good to put together a playbook for how to standup a rollup using a WIP change against the remote devnet to perform this kind of soak test.

SuperFluffy:
All websockets were removed fro Conductor.

PR #483, opened 2023-10-11
conductor: reconnect to sequencer websocket with backoff

SuperFluffy:
## Summary
After losing its websocket connection to sequener conductor now tries to reconnect up to 1024 times with exponential backoff.

## Background
Conductor only tried to reestablish its websocket connection to sequencer up to one time, bailing afterwards. This is not enough to overcome network issues, leading to conductor being unnecessarily flaky and bailing too early.

## Changes
- use exponential backoff to reestablish the sequencer websocket connection (up to 1024 attempts)
- merge the startup and reconnect logic as they are essentially the same thing.

## Testing
Testing this is tough and not done here. This is tough to implement and was not done here. Issue to track this: https://github.com/astriaorg/astria/issues/482

### DIFF
--- a/crates/astria-conductor/Cargo.toml
+++ b/crates/astria-conductor/Cargo.toml
@@ -42,6 +42,7 @@ sequencer-client = { package = "astria-sequencer-client", path = "../astria-sequ
 sequencer-validation = { package = "astria-sequencer-validation", path = "../astria-sequencer-validation" }
 telemetry = { package = "astria-telemetry", path = "../astria-telemetry" }
 tryhard = "0.5.1"
+humantime.workspace = true
 
 [dev-dependencies]
 once_cell = { workspace = true }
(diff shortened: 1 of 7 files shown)


crates/astria-conductor/src/client_provider.rs
```diff
@@ -41,47 +41,99 @@ pub(crate) enum Error {
...
+        use tracing::{
+            info,
+            info_span,
+            warn,
+            Instrument as _,
+        };
         let (client_tx, mut client_rx): (ClientTx, ClientRx) = mpsc::unbounded_channel();
+
+        info!(
+            max_attempts = 1024,
```
noot, line 66:
```suggestion
            max_attempts = RECONNECTION_ATTEMPTS,
```

noot (approved):

crates/astria-conductor/Cargo.toml
```diff
@@ -41,6 +41,8 @@ sequencer-client = { package = "astria-sequencer-client", path = "../astria-sequ
 ] }
 sequencer-validation = { package = "astria-sequencer-validation", path = "../astria-sequencer-validation" }
 telemetry = { package = "astria-telemetry", path = "../astria-telemetry" }
+tryhard = "0.5.1"
+humantime.workspace = true
```
joroshiba, line 45: nit: use same format as used elsewhere ie
```
humantime = { workspace = true }
```

I think this may be causing the CI failure even though it shouldn't as well.

joroshiba (approved):

PR #486, opened 2023-10-12
celestia-client: use eiger's version

### LINKED ISSUE #86: make all sequencer-relayer items crate-private except for the main entrypoint, config, telemetry

### LINKED ISSUE #487: Use eiger's celestia client instead of our own
https://github.com/eigerco/celestia-node-rs/ provides jsonrpsee traits for interacting with a celestia node JSONRPC client. We shouldn't maintain our version since they are actively working on theirs.

SuperFluffy:
## Summary
This patch replaces astria's custom celestia node client by https://github.com/eigerco/celestia-node-rs/. It also replaces the data availability in `sequencer-relayer` by an extension trait defined in the new `astria-celestia-client` crate.

## Background
Eiger is actively working on a rust light node and maintains a selection of rust crates to make that possible. Their work is also embraced by celestia, so we should rely on their work and expertise.

## Changes
- Remove `astria-celestia-jsonrpc-client` in favor of github.com/eigerco/celestia-node-rs
- Replace `sequencer_relayer::data_availability` by an extension trait in the new `astria-celestia-client` crate
    - this includes moving the types that are eventually submitted as blobs to that crate
    - verify that verification and signature in sequencer and rollup namespace data are correct and establish them as invariants of these types (distinguish between `UnverifiedSignedNamespaceData<D>` and `SignedNamespaceData`)
    - simplify the logic to filter/convert rollup blobs
- Remove conductor's dependency on sequencer-relayer
- Create an `astria-celestia-mock` to contain the server jsonrpsee traits used for writing blackbox tests
- Remove the `astria-test-utils` crate (used for testing `astria-celestia-jsonrpc-client`)
- Remove all kind/k8s tests from CI
- Install protoc 24.4 in GH workflows using `arduino/setup-protoc`
- Install protoco 24.4 in containerfile using precompiled official binary

## Testing
All tests still pass

## Breaking Changelist
- this should not be a breaking change in our services

## Related Issues

closes #487 
closes #86


SuperFluffy:
Need to update CI to include protoc because the generated rust protobuf isn't committed to their repo.

The tests build and pass locally so this is still ready for review

crates/astria-celestia-client/src/client.rs
```diff
@@ -0,0 +1,306 @@
...
+    }
+
+    /// Returns the rollup data for a given rollup namespace at a given height, if it exists.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if:
+    /// + the verification key could not be constructed from the data stored in `namespace_data`;
+    /// + the RPC to fetch the blobs failed.
+    async fn get_rollup_data_matching_sequencer_data(
```
SuperFluffy, line 113: Flagging this function for review. This used to return just a single roll-up namespace making certain assumptions about how blobs are submitted to celestia.

Now this is more generic, allowing multiple rollup blobs for one roll-up.

crates/astria-conductor/src/data_availability.rs
```diff
@@ -333,37 +349,42 @@ async fn verify_sequencer_blobs_and_assemble_rollups(
...
-        )
-    } else {
-        let subset = SequencerBlockSubset {
-            block_hash: data.data.block_hash,
-            header: data.data.header,
-            rollup_transactions: rollup_data.rollup_txs,
-        };
-        if block_tx.send(subset).await.is_err() {
-            warn!("failed sending validated rollup data to block assembler; receiver dropped");
+    for rollup in rollups {
```
SuperFluffy, line 373: Note this change. This now iterates over the possibly many blocks we receive

crates/astria-conductor/src/data_availability.rs
```diff
@@ -333,37 +349,42 @@ async fn verify_sequencer_blobs_and_assemble_rollups(
...
+                "could not validate rollup data given the action tree root of the sequencer data; \
+                 dropping the block"
+            )
+        } else {
+            let subset = SequencerBlockSubset {
+                block_hash: data.data().block_hash,
+                header: data.data().header.clone(),
+                rollup_transactions: rollup.rollup_txs,
+            };
+            if block_tx.send(subset).await.is_err() {
```
SuperFluffy, line 385: This might be wrong now. I believe we should first aggregate all roll-up blobs into one "subset"?

SuperFluffy (commented):
@joroshiba I flagged some changes I made where I wanted to simplify the retrieval logic.

I think this might pose problems for the executor. Please double check

crates/astria-celestia-client/Cargo.toml
```diff
@@ -0,0 +1,41 @@
...
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+[dependencies]
+async-trait = { workspace = true }
+base64 = { workspace = true }
+base64-serde = "0.7.0"
+ed25519-consensus = { workspace = true }
+serde = { workspace = true, features = ["derive"] }
+serde_json.workspace = true
```
steezeburger, line 14: nit for consistency

crates/astria-celestia-client/Cargo.toml
```diff
@@ -0,0 +1,41 @@
...
+sequencer-validation = { package = "astria-sequencer-validation", path = "../astria-sequencer-validation" }
+sequencer-types = { package = "astria-sequencer-types", path = "../astria-sequencer-types" }
+
+### Celestia specific imports
+#
+# The crates imported here are all specific to eigerco's celestia APIs
+# (including jsonrpsee).
+# As none of them are reexported it is important to keep these in sync
+# when updating.
+jsonrpsee = { version = "0.20", features = ["client-core", "macros"] }
```
noot, line 28: should we try to upstream to have them re-export?

crates/astria-celestia-client/src/blob_space.rs
```diff
@@ -0,0 +1,320 @@
...
+    let result = hasher.finalize();
+    const _: () = assert!(NS_ID_V0_SIZE < 32);
+    let id = <[u8; NS_ID_V0_SIZE]>::try_from(&result[0..NS_ID_V0_SIZE])
+        .expect("must not fail as hash is always 32 bytes and NS_ID_V0_SIZE < 32");
+    Namespace::const_v0(id)
+}
+
+fn hash_json_serialized_bytes<T: Serialize>(val: &T) -> [u8; 32] {
+    use sha2::{
+        Digest,
```
noot, line 45:
```suggestion
        Digest as _,
```

supernit for consistency with above

crates/astria-celestia-client/src/blob_space.rs
```diff
@@ -0,0 +1,320 @@
...
+#[derive(Debug, thiserror::Error)]
+pub enum SignedNamespaceDataVerification {
+    #[error("failed verifying the serialized data with the given verification key and signature")]
+    Verification(#[source] ed25519_consensus::Error),
+}
+
+/// A shadow of `SignedNamespaceData` whose data has not been verified against its
+/// contained pubic key and signature.
+///
+/// This types primary use is for serialization and deserialization.
```
steezeburger, line 164:
```suggestion
/// This type's primary use is for serialization and deserialization.
```

crates/astria-celestia-client/src/blob_space.rs
```diff
@@ -0,0 +1,320 @@
...
+    }
+}
+
+impl<D: Serialize> SignedNamespaceData<D> {
+    /// Constructs a `SignedNamespaceData` by signing the sha256 hashed JSON serialization
+    /// of `data`.
+    ///
+    /// NOTE: Json is unstable under serialization. This must be replaced ASAP.
+    pub fn from_data_and_key(data: D, key: &SigningKey) -> Self {
+        let hash = hash_json_serialized_bytes(&data);
```
noot, line 115: do we want this to be json still or are you going to update this in another PR?

crates/astria-celestia-client/src/blob_space.rs
```diff
@@ -0,0 +1,320 @@
...
+}
+
+/// RollupNamespaceData represents the data written to a rollup namespace.
+#[derive(Clone, Serialize, Deserialize, Debug)]
+pub struct RollupNamespaceData {
+    pub block_hash: Hash,
+    pub chain_id: ChainId,
+    pub rollup_txs: Vec<Vec<u8>>,
+    pub inclusion_proof: InclusionProof,
+}
```
noot, line 225: why are these here? they aren't celestia-specific, imo they should be in the types or proto crate

steezeburger (approved):

crates/astria-celestia-client/src/client.rs
```diff
@@ -0,0 +1,306 @@
...
+        &self,
+        height: u64,
+        namespace: Namespace,
+    ) -> Result<GetSequencerDataResponse, jsonrpsee::core::Error> {
+        let blobs = self.blob_get_all(height, &[namespace]).await?;
+
+        let mut datas = Vec::new();
+        let mut bad_blobs = Vec::new();
+        for blob in blobs {
+            if blob.namespace != namespace {
```
noot, line 83: is there a case where this could happen? seems bad

crates/astria-celestia-client/src/client.rs
```diff
@@ -0,0 +1,306 @@
...
+                bad_blobs.push(BadBlob {
+                    reason: BadBlobReason::WrongNamespace(blob.namespace),
+                    commitment: blob.commitment,
+                });
+            }
+            match serde_json::from_slice(&blob.data) {
+                Ok(data) => datas.push(data),
+                Err(err) => bad_blobs.push(BadBlob {
```
noot, line 91: we can probably ignore this in most cases, since someone else can just push some random blob to the same namespace, but i guess good to keep in case we have deserialization issues

crates/astria-celestia-client/src/client.rs
noot: yeah this seems fine, it should be able to return multiple blobs (since sequencer will be submitting multiple blobs per celestia block anyways)

crates/astria-celestia-client/src/client.rs
```diff
@@ -0,0 +1,306 @@
...
+            Ok(blobs) => blobs,
+            Err(err) if is_blob_not_found(&err) => {
+                return Ok(vec![]);
+            }
+            Err(err) => {
+                return Err(err);
+            }
+        };
+
+        // filter out blobs; we should only be left with either zero or one rollup datas
```
noot, line 138: remove/update comment

noot, line 138: remove/update comment, i assume if multiple rollup datas are returned the caller needs to handle it now?

crates/astria-celestia-client/src/client.rs
```diff
@@ -0,0 +1,306 @@
...
+        let rollup_datas = filter_and_convert_rollup_data_blobs(&blobs, namespace, sequencer_data);
+        Ok(rollup_datas)
+    }
+
+    /// Submits sequencer `blocks` to celestia after converting and signing them.
+    ///
+    /// This calls the `blob.Submit` celestia-node RPC.
```
noot, line 145: add "returns the celestia height at which the data was included"

crates/astria-conductor/src/block_verifier.rs
```diff
@@ -85,8 +85,12 @@ impl BlockVerifier {
...
             .wrap_err("failed to get validator set")?;
 
-        validate_sequencer_namespace_data(&current_validator_set, &parent_validator_set, &data.data)
-            .wrap_err("failed validataing sequencer data inside signed namespace data")
+        validate_sequencer_namespace_data(
+            &current_validator_set,
+            &parent_validator_set,
+            data.data(),
+        )
+        .wrap_err("failed validataing sequencer data inside signed namespace data")
```
noot, line 93:
```suggestion
        .wrap_err("failed validating sequencer data inside signed namespace data")
```

crates/astria-conductor/src/data_availability.rs
noot, line 373: i don't think this is right, if there are multiple blocks for a rollup submitted per sequencer block (which is what the case would be if there are multiple rollup_datas here), then something has gone wrong in the relayer

crates/astria-conductor/src/data_availability.rs
```diff
@@ -333,37 +349,42 @@ async fn verify_sequencer_blobs_and_assemble_rollups(
...
-            rollup_transactions: rollup_data.rollup_txs,
-        };
-        if block_tx.send(subset).await.is_err() {
-            warn!("failed sending validated rollup data to block assembler; receiver dropped");
+    for rollup in rollups {
+        if block_verifier::validate_rollup_data(&rollup, data.data().action_tree_root).is_err() {
+            warn!(
+                "could not validate rollup data given the action tree root of the sequencer data; \
+                 dropping the block"
+            )
```
noot, line 378: i guess this case would catch it - there is only one rollup data that can pass this check

crates/astria-conductor/src/data_availability.rs
noot: i suppose how exactly we handle this depends on what we want to do about a proposer modifying their relayer

crates/astria-conductor/src/data_availability.rs
noot: there should only be 1 blob that passes verification, as the sequencer network will only finalize a block where there is 1 set of data for each rollup inside the action_tree_root. if there is more than 1 blob that gets to this point, then >2/3 of the network has modified their sequencer node

crates/astria-sequencer-relayer/src/relayer.rs
```diff
@@ -86,26 +89,27 @@ impl Relayer {
...
+        )
+        .wrap_err("failed constructing celestia http client")?;
 
         let (state_tx, _) = watch::channel(State::default());
 
         Ok(Self {
             sequencer,
             sequencer_poll_period: Duration::from_millis(cfg.block_time),
             data_availability,
+            fee: 100_000,
```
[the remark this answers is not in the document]
noot: also what unit is this? i'm guessing the smallest TIA denomination?

crates/astria-sequencer-relayer/src/relayer.rs
```diff
@@ -438,17 +451,33 @@ fn convert_block_response_to_sequencer_block_data(
...
-        .await?;
-    Ok(rsp.height)
+    let height = client
+        .submit_sequencer_blocks(
+            SEQUENCER_NAMESPACE,
+            sequencer_block_data,
+            &validator.signing_key,
+            SubmitOptions {
+                fee: Some(fee),
+                gas_limit: Some(gas_limit),
```
noot, line 477: what happens if we don't put a fee or gas limit? does the node/client select them for us?

crates/astria-sequencer-relayer/tests/blackbox/helper.rs
```diff
@@ -240,75 +253,71 @@ impl MockCelestia {
...
-
-        Ok(rsp)
+        Ok(100)
     }
 }
 
-fn create_block_response(validator: &Validator, height: u32) -> endpoint::block::Response {
+fn create_block_response(
+    signing_key: &SigningKey,
+    account: tendermint::account::Id,
```
noot, line 318: derive the account ID from the signing key in here instead of passing it in every time? cause it appears the same thing is passed each time

crates/astria-sequencer-relayer/tests/blackbox/helper.rs
```diff
@@ -460,3 +469,41 @@ pub async fn mount_4_changing_block_responses(
         .await;
     rsps
 }
+
+#[allow(clippy::missing_panics_doc)]
+#[must_use]
+/// Returns a default tendermint block header for test purposes.
+pub fn make_celestia_tendermint_header() -> celestia_client::celestia_tendermint::block::Header {
```
noot, line 476: does it need to be pub?

crates/astria-conductor/src/data_availability.rs
noot: this is also assuming 1 rollup block fits into 1 blob (which it should, as we sort of enforce that in the celestia-client, but we should probably enforce that on the sequencer too)

noot (changes requested):
looks good, mostly minor comments/suggestions, I think we should add an explicit check in `fetch_and_verify_rollup_blobs` that there is only 1 rollup blob/block per sequencer block, since that's enforced in the relayer/celestia-client code currently.

crates/astria-conductor/src/data_availability.rs
SuperFluffy: I have changed it: the rollup blobs are now filtered by whether they pass verification given the sequencer namespace data. If one remains - great, we assemble it and forward to the executor. If more than 1 remain - drop them all.

crates/astria-celestia-client/Cargo.toml
SuperFluffy: good eye, ty

crates/astria-celestia-client/Cargo.toml
SuperFluffy: I opened this:
https://github.com/eigerco/celestia-node-rs/issues/100
Probably makes sense to just submit PRs ourselves, yeah.

crates/astria-celestia-client/src/blob_space.rs
```diff
@@ -0,0 +1,320 @@
...
+
+    pub fn public_key(&self) -> VerificationKey {
+        self.public_key
+    }
+
+    pub fn signature(&self) -> Signature {
+        self.signature
+    }
+
+    pub fn from_unverified_unchecked(unverified: UnverifiedSignedNamespaceData<D>) -> Self {
```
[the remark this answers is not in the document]
SuperFluffy: This is in anticipation of wanting to unwrap the verified data to get some values, and then immediately rewrap into a verified type without doing an additional round of verification. But I agree, we should only add it (and make public) when needed

crates/astria-celestia-client/src/blob_space.rs
SuperFluffy: Yeah, we don't want it to be json. But moving it all to protobuf is a lot of work and outside the scope of this PR. #12 and #443

crates/astria-celestia-client/src/blob_space.rs
```diff
@@ -0,0 +1,320 @@
...
+        let rollup_data_tree = MerkleTree::from_leaves(self.rollup_txs.clone());
+        let rollup_data_root = rollup_data_tree.root();
+        let mut leaf = self.chain_id.as_ref().to_vec();
+        leaf.append(&mut rollup_data_root.to_vec());
+        self.inclusion_proof.verify(&leaf, root_hash)?;
+        Ok(())
+    }
+}
+
+mod _serde {
```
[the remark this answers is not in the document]
SuperFluffy: Yeah, these are used as attributes for the serde derivation of `UnverifiedSignedNamespaceData`

crates/astria-celestia-client/src/client.rs
SuperFluffy: I hope it won't ever happen, but I figured it's worth it to protect against it. :)

crates/astria-celestia-client/src/client.rs
SuperFluffy: My thinking was that we should start thinking about namespaces becoming occupied by 3rd parties. We could construct metrics around this, and it's just interesting data.

crates/astria-celestia-client/src/client.rs
SuperFluffy: good catch! yeah, this should be handled by the caller now.

crates/astria-conductor/src/data_availability.rs
```diff
@@ -148,9 +159,10 @@ impl Reader {
     }
 
     fn get_latest_height(&mut self) {
-        let client = self.celestia_client.inner().clone();
+        use celestia_client::celestia_rpc::HeaderClient;
+        let client = self.celestia_client.clone();
         self.get_latest_height = Some(tokio::spawn(async move {
-            Ok(client.header_network_head().await?.height())
+            Ok(client.header_network_head().await?.header.height.value())
```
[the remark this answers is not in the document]
SuperFluffy: I moved everything from u64 to (celestia-tendermint) Height because the use of `Head::value` everywhere was indeed getting silly.

The extension trait also saw some changes to take height arguments as `T: Into<u64> + Send` instead so that it becomes convenient to use with `Height`.

crates/astria-conductor/src/data_availability.rs
SuperFluffy: Adjusted: for now it will noisily drop all blobs if more than 1 blob survived verification.

crates/astria-conductor/src/data_availability.rs
SuperFluffy: Yeah, so just in case this doesn't happen it will now noisily drop those cases where more than 1 blob remains after verification.

crates/astria-sequencer-relayer/src/relayer.rs
SuperFluffy: I actually don't know. I have submitted this: https://github.com/astriaorg/astria/issues/510

SuperFluffy:
@noot addressed all comments, in particular in regard to dropping/ignoring blobs if conductor receives more than one.

crates/astria-conductor/src/data_availability.rs
```diff
@@ -330,41 +352,61 @@ async fn verify_sequencer_blobs_and_assemble_rollups(
...
+            let subset = SequencerBlockSubset {
+                block_hash: data.data().block_hash,
+                header: data.data().header.clone(),
+                rollup_transactions: rollups.swap_remove(0).rollup_txs,
+            };
+            if block_tx.send(subset).await.is_err() {
+                warn!("failed sending validated rollup data to block assembler; receiver dropped");
+            }
         }
+        num_rollups_verified => info!(
```
noot, line 404: maybe change to warn? this would require investigation if it happened

noot (approved):
looks good, nice work :)

crates/astria-conductor/src/data_availability.rs
SuperFluffy: ty, good catch. That's what I wanted

Issue #489, opened 2023-10-12
Both `composer` and `astria-cli` fail to make calls to sequencer behind https enabled domain


jbowen93:
`astria-cli` call from https://github.com/astriaorg/astria/tree/feature/self-serve-rollup-cli

```
❯ ./target/release/astria-cli sequencer blockheight get --sequencer-url https://rpc.sequencer.dusk-1.devnet.astria.org
failed to get sequencer block

Caused by:
   0: sequencer client method failed
   1: executing tendermint RPC failed
   2: HTTP error

      Caused by:
   2:     error trying to connect: invalid certificate timestamp: UnknownLog

      Location:
   2:     /Users/joshbowen/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flex-error-0.4.4/src/tracer_impl/eyre.rs:10:9
   3: HTTP error
   4: error trying to connect: invalid certificate timestamp: UnknownLog

Location:
    /Users/joshbowen/Code/demo/astria/crates/astria-cli/src/commands/sequencer.rs:108:10
```

Logs from composer
```
❯ kubectl logs josh-geth-fb4fc44b5-cjwvd -c composer --namespace=astria-dev-cluster -f
config:
{"log":"astria_composer=DEBUG","api_listen_addr":"0.0.0.0:0","sequencer_url":"https://rpc.sequencer.dusk-1.devnet.astria.org/","rollups":"0x1234::ws://127.0.0.1:8546/","private_key":"3ab0############################################################"}
service is attached to tty; using human readable formatting
2023-10-12T04:45:01.627504Z  INFO astria_composer: initializing composer config="{\"log\":\"astria_composer=DEBUG\",\"api_listen_addr\":\"0.0.0.0:0\",\"sequencer_url\":\"https://rpc.sequencer.dusk-1.devnet.astria.org/\",\"rollups\":\"0x1234::ws://127.0.0.1:8546/\",\"private_key\":\"3ab0############################################################\"}"
2023-10-12T04:45:01.627519Z DEBUG astria_composer::composer: creating searcher
2023-10-12T04:45:01.629313Z DEBUG astria_composer::composer: creating API server
2023-10-12T04:45:01.629364Z DEBUG astria_composer::composer: API server listening listen_addr=0.0.0.0:46713
[...]

PR #492, opened 2023-10-12
composer: report cause of failed nonce fetch

SuperFluffy:
## Summary
Report the error backtrace when fetching the initial nonce from the sequencer.

## Background
Composer only logged the superficial error reporting what failed, but not why.

## Changes
- log the error source/backtrace when getting the initial nonce from sequencer
- pretty-print the address using display (instead of debug + a format string)
- unrelated: some fixes because clippy started complaining with the bump to 1.72

## Testing
Not applicable.

## Breaking Changelist

## Related Issues
Investigating https://github.com/astriaorg/astria/issues/489 revealed that our events were lacking information to understand what's going on.

### DIFF
--- a/crates/astria-sequencer-types/src/namespace.rs
+++ b/crates/astria-sequencer-types/src/namespace.rs
@@ -47,6 +47,7 @@ impl Namespace {
     #[must_use]
     #[allow(clippy::missing_panics_doc)]
     pub fn from_slice(bytes: &[u8]) -> Namespace {
+        #[allow(clippy::assertions_on_constants)]
         const _: () = assert!(
             NAMESPACE_ID_AVAILABLE_LEN <= 32,
             "this can only be violated if celestia had a breaking change fundamentally altering \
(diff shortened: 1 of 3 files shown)


noot (approved):

crates/astria-sequencer-types/src/namespace.rs
```diff
@@ -41,15 +41,21 @@ impl Namespace {
     /// returns an 10-byte namespace given a byte slice by hashing
     /// the bytes with sha256 and returning the first 10 bytes.
     #[must_use]
+    #[allow(clippy::missing_panic_docs)]
     pub fn from_slice(bytes: &[u8]) -> Namespace {
         let mut hasher = Sha256::new();
         hasher.update(bytes);
         let result = hasher.finalize();
+        const _: () = assert!(
+            NAMESPACE_ID_AVAILABLE_LEN <= 32,
```
joroshiba, line 50: I believe celestia limits us to much less than 32 bytes? https://celestiaorg.github.io/celestia-app/specs/namespace.html

28 bytes of which we can only use 10?

I'm a bit confused of the purpose here, we have an assert on a constant?

joroshiba (approved):

crates/astria-sequencer-types/src/namespace.rs
SuperFluffy: This is compile time stuff and just to ensure that the `expect` right below is correct. I can't get the `32` bytes out of `sha256` so I need to hardcode 32 here.

Just best practices because `NAMESPACE_ID_AVAILABLE_LEN` is imported into this crate and could just be changed without us noticing.

PR #493, opened 2023-10-12
release(composer): Release 0.2.2

SuperFluffy:
## Summary
New minor release of composer.

### DIFF
--- a/crates/astria-composer/Cargo.toml
+++ b/crates/astria-composer/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "astria-composer"
-version = "0.2.1"
+version = "0.2.2"
 edition = "2021"
 rust-version = "1.70.0"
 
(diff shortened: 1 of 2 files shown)


noot (approved):

crates/astria-composer/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "astria-composer"
-version = "0.2.1"
+version = "0.2.2"
```
joroshiba, line 3: I think there should be an update in main cargo.toml as well?

SuperFluffy: cargo.lock yeah - force pushed. But what about cargo.toml?

joroshiba (approved):

crates/astria-composer/Cargo.toml
joroshiba: my bad, cargo.lock what I meant

PR #499, opened 2023-10-13
conductor: never recycle websocket clients

SuperFluffy:
## Summary
Websocket clients to sequencer are never returned to the object pool and instead freshly created

## Background
Websocket clients were returned to the object pool even though their underlying websocket driver was long dead. This meant that a new driver never received any requests because its clients were never used, while actors using the old clients immediately errored.

## Changes
- Implement the `deadpool::managed::Manager::recycle` trait method to always return an error.

## Testing
- Conductor is now able to communicate after a websocket/tcp resets. It was not able before.

### DIFF
--- a/crates/astria-conductor/src/client_provider.rs
+++ b/crates/astria-conductor/src/client_provider.rs
@@ -181,6 +181,8 @@ impl managed::Manager for ClientProvider {
         _obj: &mut Self::Type,
         _: &managed::Metrics,
     ) -> managed::RecycleResult<Self::Error> {
-        Ok(())
+        Err(deadpool::managed::RecycleError::StaticMessage(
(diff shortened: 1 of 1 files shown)


joroshiba (approved):
Should we have a future issue to recycle if the driver is still good?

sambukowski (approved):

SuperFluffy:
@joroshiba 
> Should we have a future issue to recycle if the driver is still good?
There is now way to figure that out from https://docs.rs/tendermint-rpc/latest/tendermint_rpc/client/struct.WebSocketClient.html

But on the other hand cloning these clients is very cheap (since the sender are just `Arc`s), so the advantage of recycling is probably very limited.

PR #501, opened 2023-10-13
conductor: resubscribe with backoff instead of failing

SuperFluffy:
## Summary
Conductor attempts to subscribe to the sequencer with a constant backoff instead of failing after the first attempt.

## Background
Conductor was unnecessarily flaky - instead of trying to resubscribe repeatedly it failed instantly.

## Changes
- Remove the exit condition from the sequencer reader on a failed resubscription to new blocks
- Try to repeatedly resubscribe to new blocks with a constant backoff of 10 seconds.

## Testing
Tested on devnet

### DIFF
--- a/crates/astria-conductor/src/sequencer/mod.rs
+++ b/crates/astria-conductor/src/sequencer/mod.rs
@@ -248,9 +248,12 @@ impl Clone for ResubscriptionError {
 async fn subscribe_new_blocks(
     pool: Pool<ClientProvider>,
 ) -> Result<NewBlocksStream, ResubscriptionError> {
+    use std::time::Duration;
+
     use sequencer_client::SequencerSubscriptionClientExt as _;
-    let retry_config = tryhard::RetryFutureConfig::new(1024)
-        .fixed_backoff(std::time::Duration::from_secs(10))
+    let retry_config = tryhard::RetryFutureConfig::new(10)
+        .exponential_backoff(Duration::from_millis(100))
+        .max_delay(Duration::from_secs(10))
         .on_retry(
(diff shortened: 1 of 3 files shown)


crates/astria-conductor/src/sequencer/mod.rs
```diff
@@ -186,36 +187,99 @@ impl Reader {
...
+    pool: Pool<ClientProvider>,
+) -> Result<NewBlocksStream, ResubscriptionError> {
     use sequencer_client::SequencerSubscriptionClientExt as _;
-    pool.get()
-        .await
-        .wrap_err("failed getting a sequencer client from the pool")?
-        .subscribe_new_block_data()
-        .await
-        .wrap_err("failed subscribing to sequencer to receive new blocks")
+    let retry_config = tryhard::RetryFutureConfig::new(1024)
```
joroshiba, line 252: Should we use a more sophisticated less long pull retry stance here?

Like what about 10 retries on an exponential backoff starting at .5 and that maxes out at 10s? 10s is very slow when blocks are every 2s. You would defacto miss 5 blocks. 

```
tryhard::RetryFutureConfig::new(10)
    .exponential_backoff(std::time::Duration::from_millis(100))
    .max_delay(std::time::Duration::from_secs(10)
```

joroshiba (approved):

crates/astria-conductor/src/sequencer/mod.rs
SuperFluffy: Yeah, that makes sense. Pushed it this way. I don't know about the 10 retries. Effectively what it now does is 11 attempts (10 retries, starting from 100 ms exponential up to 10 s), and then it cycles again.

sambukowski (approved):

PR #505, opened 2023-10-16
fix(conductor): update rollup chain id in conductor example env to match composer

noot:
## Summary
very small fix to have the rollup chain id in conductor example env to match composer's example env rollup chain id (both are now "astriachain")


SuperFluffy (approved):
Good idea.

PR #514, opened 2023-10-17
feat(composer): allow rollup names with dash

SuperFluffy:
## Summary
Allow dashes in rollup names passed to composer through the env var `ASTRIA_COMPOSER_ROLLUPS="astria-dusk-1::ws://127.0.0.1:8546/"`

## Background
Only allowing alphanumeric characters as rollup chain IDs was overly restrictive. It's nice for human readability and for matching subdomain paths to rollup IDs when deploying to allow dashes.

## Changes
- Add `-` to the character classes in the group matching chain IDs names
- Add tests
- Mention alphanumeric ascii and dash in example config
 
## Testing
Tests have been updated to include dashes.


jbowen93 (approved):

noot (approved):
should we allow underscores too?

joroshiba (approved):

joroshiba:
> should we allow underscores too?

We use this as a part of a hostname, so valid characters are alphanumeric and`-`

PR #515, opened 2023-10-17
feat(conductor): add re-sync for missed sequencer blocks

SuperFluffy:
## Summary
Fetch missing sequencer blocks before forwarding to the executor actor.

## Background
If the websocket connection to sequencer drops temporarily (thus not receiving blocks from the `NewBlocks` jsonrpc subscription), blocks for one or more heights might be missed. This is detected by tracking the expected height of the next sequencer block to be forwarded to the executor actor. If the next block taken from the queue of pending blocks has a height greater than the expected height a resync is triggered and the current block rescheduled in front of the queue of regular pending blocks. This patch reuses the sync-mechanism that is run at startup to fetch a contiguous range of missing sequencer blocks. Only after this "re-sync" is completed will regular submission of blocks resume

## Changes
- Track the height of sequencer blocks
- Trigger a resync if there is a mismatch between the height in the next sequencer block and the expected height
- Resume submission of blocks after resync is done

## Testing
Added unit tests for all three forward/resync conditions:
1. forward to executor if block height == expected height
2. drop if block height < expected height
3. sync + reschedule if block height > expected height

No blackbox tests yet. Must be tested on devnet.

### DIFF
--- a/crates/astria-conductor/src/sequencer/mod.rs
+++ b/crates/astria-conductor/src/sequencer/mod.rs
@@ -300,15 +300,16 @@ async fn subscribe_new_blocks(
 /// Forwards a sequencer block to the executor if it contains the expected height, or reschedules
 /// the block for later while fetching blocks for all missing heights.
 ///
-/// The following cases are considered:
+/// The following cases are considered (with `h` the next height expected by the sequencer reader,
+/// and `k` the height recorded in the block):
 ///
-/// 1. if `h == h'` the block is forwarded to the executor. `h+1` is returned as the next expected
+/// 1. if `h == k` the block is forwarded to the executor. `h+1` is returned as the next expected
 ///    height.
-/// 2. if `h < h'` the block is dropped because its height `h'` is below the expected height `h`.
-///    `h` is returned as the next expected height (i.e. no change in the expected height).
-/// 3. if `h > h'` a re-sync is scheduled for the range `h..h'`, the block is pushed to the front of
-///    the queue to be forwarded later. `h'` (the height of the re-scheduled block) is returned as
-///    the next expected height.
+/// 2. if `h < k` the block is dropped. `h` is returned as the next expected height (i.e. there is
+///    no change in the expected height).
+/// 3. if `h > k` a re-sync is scheduled for the range `h..k` (exluding `k`), the block is pushed to
+///    the front of the queue to be forwarded later. `k` (the height of the re-scheduled block) is
+///    returned as the next expected height.
 ///
 /// # Returns
 /// Returns the next expected height, depending on the cases discussed above.
--- a/crates/astria-conductor/src/sequencer/mod.rs
+++ b/crates/astria-conductor/src/sequencer/mod.rs
@@ -24,6 +24,7 @@ use futures::{
 };
 use sequencer_client::{
     extension_trait::NewBlocksStream,
+    tendermint::block::Height,
     NewBlockStreamError,
 };
 use tokio::{
@@ -40,7 +41,6 @@ use tracing::{
 use crate::{
     client_provider::ClientProvider,
     executor,
-    executor::ExecutorCommand,
 };
 
 mod sync;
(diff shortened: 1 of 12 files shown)


crates/astria-conductor/src/sequencer/mod.rs
```diff
@@ -285,22 +289,58 @@ async fn subscribe_new_blocks(
...
-fn forward_pending_block(
+fn forward_pending_block_or_resync(
+    expected_height: u32,
     executor_tx: executor::Sender,
+    pool: Pool<ClientProvider>,
+    pending_blocks: &mut FuturesOrdered<Ready<Result<SequencerBlockData, NewBlockStreamError>>>,
+    sync: &mut future::Fuse<Pin<Box<dyn Future<Output = eyre::Result<()>> + Send>>>,
     res: Result<SequencerBlockData, NewBlockStreamError>,
-) -> eyre::Result<()> {
+) -> eyre::Result<u32> {
```
noot, line 299: can you add a comment explaining the return value?

crates/astria-conductor/src/sequencer/mod.rs
```diff
@@ -285,22 +289,58 @@ async fn subscribe_new_blocks(
...
-        .send(ExecutorCommand::FromSequencer {
-            block: Box::new(block),
-        })
-        .wrap_err("failed sending new block received from sequencer to executor")
+    let block_height = block
+        .header()
+        .height
+        .value()
+        .try_into()
+        .wrap_err("failed converting cometbft to u32, although this should always work")?;
```
noot, line 313:
```suggestion
        .wrap_err("failed converting cometbft block height to u32, although this should always work")?;
```

crates/astria-conductor/src/sequencer/mod.rs
noot: it also doesn't look like the value is used anywhere?

crates/astria-conductor/src/sequencer/mod.rs
```diff
@@ -285,22 +289,58 @@ async fn subscribe_new_blocks(
...
+
+    match expected_height.cmp(&block_height) {
+        // received block is at expected height: send to the executor
```
noot, line 316: what is the meaning of expected height here? is it the target height we should be syncing to?

crates/astria-conductor/src/sequencer/mod.rs
```diff
@@ -285,22 +289,58 @@ async fn subscribe_new_blocks(
     .map_err(|_| ResubscriptionError::BackoffFailed)
 }
 
-fn forward_pending_block(
+fn forward_pending_block_or_resync(
+    expected_height: u32,
```
noot, line 293: can you add a comment clarifying `expected_height`?

crates/astria-conductor/src/sequencer/mod.rs
SuperFluffy: Good catch, that was important to include. I am now using the return value and added a mention in the doc comment on what it returns.

crates/astria-conductor/src/sequencer/mod.rs
SuperFluffy: I have changed all instances of `u32` to use the cometbft `Height` directly, which is cleaner.

crates/astria-conductor/src/sequencer/mod.rs
SuperFluffy: Expanded the doc comment on the function and what it does.

SuperFluffy:
Added unit tests for all three forward/resync conditions:
1. forward to executor if block height == expected height
2. drop if block height < expected height
3. sync + reschedule if block height > expected height

sambukowski (approved):

crates/astria-conductor/src/sequencer/mod.rs
```diff
@@ -160,8 +158,11 @@ impl Reader {
                     } else {
                         info!("sync finished successfully");
                     }
-                    let sync_done = sync_done.take().expect("channel should only be used once and only in this branch; this is a bug");
-                    let _ = sync_done.send(());
+                    // First sync at startup: notify conductor that sync is done.
+                    // Every resync after: don't.
```
noot, line 162: why do we only need to notify the first time?

crates/astria-conductor/src/sequencer/mod.rs
```diff
@@ -285,35 +297,244 @@ async fn subscribe_new_blocks(
     .map_err(|_| ResubscriptionError::BackoffFailed)
 }
 
-fn forward_pending_block(
+/// Forwards a sequencer block to the executor if it contains the expected height, or reschedules
+/// the block for later while fetching blocks for all missing heights.
+///
+/// The following cases are considered:
+///
+/// 1. if `h == h'` the block is forwarded to the executor. `h+1` is returned as the next expected
```
noot, line 305: can we change these to not be `h` and `h` but `h` and something else :p same for rest of the comment

crates/astria-conductor/src/sequencer/mod.rs
SuperFluffy: Because we don't have a mechanism to stop DA execution if the sequencer read goes down. Nor do I think we should (dunno what this should look like).

More technically speaking: we use a one shot channel to signal to conductor that it may spin up the DA reader after the first sequencer sync is done. It's an error to use the one shot channel again, so we have to protect against it.

crates/astria-conductor/src/sequencer/mod.rs
SuperFluffy: It's `h'` read as h-prime. That's what I grew up with 😭

I can see how it's tough to read depending on the font. I'll change to `k`.

noot (approved):

crates/astria-conductor/src/sequencer/mod.rs
noot: ah sorry i didn't see the `'`!!

PR #516, opened 2023-10-17
feat(conductor): replace formatted error backtraces by value impl

SuperFluffy:
## Summary
This replaces explicit instrument formatting using the `?` and `%` sigils by the `tracing::Value` impl for `dyn std::error::Error` giving nicer tracing events.

## Background
So war we were emitting tracing events for errors like `warn!(error.message = %err, error.cause = ?err, "failed to do something");`, which lead to formatted multiline errors that did not match well with normal tracing subscriber formatting. This patch uses the `tracing::Value` implementation for `std::error::Error` trait objects, giving error backtraces in form of a list and. Unfortunately errors have to be explicitly made into trait objects (either through `&err as &dyn Error` or using a binding like `let error: &dyn Error = eyre_report.as_ref()` for eyre reports).

## Changes
- Replace all tracing events using `%` and `?` sigils for error formatting by the `tracing::Value` implementation on their trait objects.

### DIFF
--- a/crates/astria-conductor/src/client_provider.rs
+++ b/crates/astria-conductor/src/client_provider.rs
@@ -79,12 +79,13 @@ impl ClientProvider {
                         .map(humantime::format_duration)
                         .map(tracing::field::display);
                     async move {
+                        let error = &error as &(dyn std::error::Error + 'static);
                         warn!(
                             attempt,
                             wait_duration,
-                            error.message = %error,
-                            error.cause = ?error,
-                            "attempt to connect to sequencer websocket failed; retrying after backoff",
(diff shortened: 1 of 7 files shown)


noot (approved):
what do the error logs look like now?

SuperFluffy:
> what do the error logs look like now?

Instead of something like this:

```
error = Error:
  0: outer most context
  1: inner context
  2. underlying low level error
 ```
You just get a list, essentially:
```
error = ["outer most context", "inner context", "underlying low level error"]
```
This fits much better with events.

steezeburger (approved):

PR #517, opened 2023-10-17
feat(cli): installable cli

joroshiba:
## Summary
Makes it easier to install the `astria-cli`

## Background
In the docs you currently need to download and run from local target, this makes it installed on your path nicely.

## Changes
- Update cargo.toml to support binary
- added just command to make it easy to build with locked versions

## Testing
Ran commands on local machine


WafflesVonMaple (approved):

justfile
```diff
@@ -38,3 +38,6 @@ lint-toml:
 
 lint-md:
   markdownlint-cli2 "**/*.md" "#target" "#.github"
+
+install-cli:
+  cargo install --path ./crates/astria-cli --locked
```
SuperFluffy, line 43: No suggestion in either direction, but you could probably also specify `cargo install --bin astria-cli`

SuperFluffy (approved):
Looks fine, but I would mention that this is most likely installed in `$CARGO_HOME` (if configured) or just `~/.cargo/bin`. It should be immediately available if one that in `$PATH` (which is usually the case if one has set up rust/cargo etc).

PR #520, opened 2023-10-17
feat(cli): grab block height from sequencer + minimize required create inputs

joroshiba:
## Summary
Setting more defaults, making less things required.

## Background
Trying to simplify the number of things which must be set, while still enabling more power when needed.

## Changes
- sequencer block height grab from sequencer
- default network id to `1337`
- Add some docs for commands

## Testing
Running CLI locally + CI on this PR

## Issues
closes https://github.com/astriaorg/astria/issues/513

### DIFF
--- a/crates/astria-cli/src/cli/rollup.rs
+++ b/crates/astria-cli/src/cli/rollup.rs
@@ -11,6 +11,8 @@ const DEFAULT_ROLLUP_CHART_PATH: &str =
     "https://astriaorg.github.io/dev-cluster/astria-evm-rollup-0.4.0.tgz";
 const DEFAULT_SEQUENCER_RPC: &str = "https://rpc.sequencer.dusk-1.devnet.astria.org";
 const DEFAULT_SEQUENCER_WS: &str = "wss://rpc.sequencer.dusk-1.devnet.astria.org/websocket";
+const DEFAULT_LOG_LEVEL: &str = "info";
+const DEFAULT_NETWORK_ID: u64 = 1337;
 
 /// Remove the 0x prefix from a hex string if present
 fn strip_0x_prefix(s: &str) -> &str {
@@ -51,7 +53,7 @@ pub enum ConfigCommand {
 pub struct ConfigCreateArgs {
     #[clap(long, env = "ROLLUP_USE_TTY")]
     pub use_tty: bool,
-    #[clap(long, env = "ROLLUP_LOG_LEVEL")]
+    #[clap(long, env = "ROLLUP_LOG_LEVEL", default_value = DEFAULT_LOG_LEVEL)]
     pub log_level: String,
 
     // rollup config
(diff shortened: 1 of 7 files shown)


crates/astria-cli/src/cli/rollup.rs
```diff
@@ -61,8 +63,10 @@ pub struct ConfigCreateArgs {
     /// Optional. Will be derived from the rollup name if not provided
     #[clap(long = "rollup.chain-id", env = "ROLLUP_CHAIN_ID", required = false)]
     pub chain_id: Option<String>,
-    #[clap(long = "rollup.network-id", env = "ROLLUP_NETWORK_ID")]
+    /// The Network ID for the EVM chain
+    #[clap(long = "rollup.network-id", env = "ROLLUP_NETWORK_ID", default_value_t = DEFAULT_NETWORK_ID)]
     pub network_id: u64,
```
noot, line 68: is this the network id or the chain id for the evm? or both?

crates/astria-cli/src/commands/rollup.rs
```diff
@@ -99,9 +103,24 @@ fn update_yaml_value(
 /// * If the config file cannot be created
 /// * If the arguments cannot be serialized to yaml
 /// * If the yaml cannot be written to the file
-pub(crate) fn create_config(args: &ConfigCreateArgs) -> eyre::Result<()> {
+pub(crate) async fn create_config(args: &ConfigCreateArgs) -> eyre::Result<()> {
     // create rollup from args
-    let rollup = Rollup::try_from(args)?;
+    let mut rollup = Rollup::try_from(args)?;
```
noot, line 108: why not check `if args.sequencer_initial_block_height.is_some()` instead of the check below? or change `initial_sequencer_height` to be an option?

also update here https://github.com/astriaorg/astria/blob/95864ea139ef5532c0bc3e629f89faea49452e50/crates/astria-cli/src/types.rs#L81

noot (approved):

crates/astria-cli/src/cli/rollup.rs
joroshiba: The `network_id` is the evm network ID which you might lookup on chainlist. The `chain_id` is what is used through the shared sequencer. We will derive a `chain_id` from the `name` if not provided, where the name is just something human readable used as an identifier.

sambukowski (approved):

crates/astria-cli/src/cli/rollup.rs
noot: sorry i guess i meant whether this is the `chain_id` or `network_id` for the EVM chain since in eth-land those are two different things (althought often used similarly), network_id is used for the p2p network while chain_id is what's on chainlist. i'm kinda thinking we should change our `chain_id` to be `rollup_id` since it's a bit of a naming conflict

noot: anyways out of scope for this PR

crates/astria-cli/src/commands/rollup.rs
```diff
@@ -99,9 +103,24 @@ fn update_yaml_value(
 /// * If the config file cannot be created
 /// * If the arguments cannot be serialized to yaml
 /// * If the yaml cannot be written to the file
-pub(crate) fn create_config(args: &ConfigCreateArgs) -> eyre::Result<()> {
+pub(crate) async fn create_config(args: &ConfigCreateArgs) -> eyre::Result<()> {
     // create rollup from args
-    let rollup = Rollup::try_from(args)?;
+    let mut conf = (*args).clone();
```
noot, line 108:
```suggestion
    let mut conf = args.clone();
```
don't think you need the deref?

noot (approved):

PR #527, opened 2023-10-19
cli: refactor network_id to be a string on the serialization side

steezeburger:
## Summary
helm likes to turn big integers into scientific notation, so we serialize as a string to ensure proper parsing in helm land

closes https://github.com/astriaorg/astria/issues/526

### DIFF
--- a/crates/astria-cli/src/types.rs
+++ b/crates/astria-cli/src/types.rs
@@ -218,7 +218,7 @@ mod tests {
             log_level: "info".to_string(),
             name: "rollup2".to_string(),
             chain_id: None,
-            network_id: 2211011801,
+            network_id: 2_211_011_801,
             skip_empty_blocks: false,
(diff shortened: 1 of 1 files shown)


crates/astria-cli/src/types.rs
```diff
@@ -114,7 +114,8 @@ impl TryFrom<&ConfigCreateArgs> for RollupDeploymentConfig {
 pub struct RollupConfig {
     name: String,
     chain_id: String,
-    network_id: u64,
+    // NOTE - String here because
```
joroshiba: Can we make this a bit more clear?

joroshiba (approved):

crates/astria-cli/src/types.rs
steezeburger: haha yeah already fixed. Got in a rush :)

sambukowski (approved):

Issue #504, opened 2023-10-16
composer: reconnect to rollup node if websocket connection goes down


noot:
as title says, if the rollup node goes down composer does not auto-reconnect, you have to restart composer when the rollup node is back up which is annoying. we should have some retry mechanism for the rollup ws connection

Issue #283, opened 2023-08-18
composer: fix rollup node readiness check


itamarreif:
# Problem Description
Currently (assuming #278 and related changes are merged), the `EthClient` constructor only calls `Provider<Ws>::connect()`, which [simply establishes a TCP connection](https://docs.rs/ethers-providers/2.0.8/src/ethers_providers/rpc/transports/legacy_ws.rs.html#150) (it uses [`tokio_tungstenite`'s `connect_async` under the hood](https://docs.rs/tokio-tungstenite/latest/src/tokio_tungstenite/connect.rs.html#13-20), which just establishes a TCP connection).

Only when the `Collector` is run, `wait_for_evm` is called. This pings the rollup node's `get_net_version` endpoint to establish readiness. However, since `Provider::connect()` is only run once in the `EthClient` constructor (called in `Collector`'s constructor), the TCP client can fail to be established (e.g. Geth node hasn't finished initializing). This will lead to all the calls to `get_net_version` will be run by a client that failed to connect, and will in turn fail as well.

# Proposed Solution
1. Run `Provider::connect()` in the exponential backoff closure before calling the `get_net_version` endpoint
2. Run the exponential backoff in `EthClient::connect()` so that we know a connection is actually established

PR #532, opened 2023-10-23
composer: resubscribe if rollup subscription stops

SuperFluffy:
## Summary
Added logic to reconnect and resubscribe to a rollup if the subscription ends.

## Background
Composer exited if a websocket/jsonrpc subscription to a remote (geth) rollup failed, requiring a full restart of the service.
With this patch the `searcher` task of composer tries to immediately restart a collector if it exited.

## Changes
- Handle a collector exiting (with or without error) by immediately restarting it.
- Set the retry-count in the ethers `WsClient` (aliased to `Ws`) to 0: the retry logic is opaque and has strange interaction with startup, reconnect on websocket drop, and reconnect on subscription drop. We better explicitly differentiate between connect-failure, subscription-failure, and reconnect-retry.
- Change collector reporting its `is_connected` status only after both the websocket client && the subscription are established.
- Move the`geth` server mock to `astria-test-utils` and put it behind feature flags.

## Testing
- A unit test for the `reconnect_exited_collector` function is provided, which triggers the reconnect-logic by returning a subscription error. Warts:
    - It would be better to test the TCP socket failing, but that would create flaky tests because it would require binding the tcp socket to the same port. An alternative would be to use tokio-turmoil, but ethers does not allow configuring the TcpStream.
    - It would also be better to test the whole `Searcher` instead of the handler (because it feels like we are testing implementation details), but it that would require mocking many more parts of it. A better alternative would be to create a `CollectorManager` that abstracts the reconnection + channel logic away from `Searcher`.
    - 
## Related Issues

closes https://github.com/astriaorg/astria/issues/504
closes https://github.com/astriaorg/astria/issues/283


sambukowski (approved):

itamarreif:
seems to also address this issue: https://github.com/astriaorg/astria/issues/283

SuperFluffy:
> seems to also address this issue: https://github.com/astriaorg/astria/issues/283

Nice, I think it does. I'll update the PR text

joroshiba (approved):

crates/astria-composer/src/searcher/collector.rs
```diff
@@ -102,22 +95,51 @@ impl Collector {
...
+                |attempt, next_delay: Option<Duration>, error: &ProviderError| {
+                    let error = error as &(dyn std::error::Error + 'static);
+                    let wait_duration = next_delay
+                        .map(humantime::format_duration)
+                        .map(tracing::field::display);
+                    warn!(
+                        attempt,
+                        wait_duration,
+                        error,
+                        "attempt to connect to geth node failed; retrying after backoff",
```
noot, line 119:
```suggestion
                        "attempt to connect to eth node failed; retrying after backoff",
```

not sure why this was ever changed to geth lol

crates/astria-composer/src/searcher/mod.rs
```diff
@@ -44,8 +50,10 @@ pub(super) struct Searcher {
     collectors: HashMap<String, Collector>,
     collector_statuses: HashMap<String, watch::Receiver<collector::Status>>,
     // A channel on which the searcher receives transactions from its collectors.
-    new_transactions: Receiver<collector::Transaction>,
-    collector_tasks: tokio_util::task::JoinMap<String, eyre::Result<()>>,
+    new_transactions_rx: Receiver<collector::Transaction>,
+    new_transactions_tx: Sender<collector::Transaction>,
+    rollups: HashMap<String, String>,
```
noot, line 55: can you add a comment explaining what these fields are?

crates/astria-composer/src/searcher/mod.rs
```diff
@@ -44,8 +50,10 @@ pub(super) struct Searcher {
...
+    collector_tasks: JoinMap<String, eyre::Result<()>>,
```
noot, line 56: same here (i guess chain_id to result?)

crates/astria-composer/src/searcher/mod.rs
```diff
@@ -318,3 +324,107 @@ impl Searcher {
...
+        Err(e) => {
+            let error = &e as &(dyn std::error::Error + 'static);
+            warn!(error, "collector task failed; reconnecting");
+        }
+    }
+    let Some(url) = rollups.get(&rollup) else {
+        warn!(
+            "rollup should have had an entry in the rollup->url map but doesn't; not reconnecting \
+             it"
+        );
```
noot, line 351: should never happen right?

noot (approved):
looks good!

crates/astria-composer/src/searcher/collector.rs
SuperFluffy: It does make sense because we are using a geth specific RPC. I believe @itamarreif favored that.

crates/astria-composer/src/searcher/mod.rs
SuperFluffy: Added

crates/astria-composer/src/searcher/mod.rs
SuperFluffy: Added a comment. It's actually just a mechanism to track if a collector task is still running. `collector_tasks.join_next()` should only return `Some(exit_status)` if the collector exited for a reason like the websocket going down.

crates/astria-composer/src/searcher/mod.rs
SuperFluffy: Yeah, it shouldn't. I bumped it to `error!`.

PR #533, opened 2023-10-23
feat(cli): add hostname + namespace

joroshiba:
## Summary
Adds the hostname flag into CLI, and makes the namespace configurable.

## Background
We updated charts, and currently deploying to cloud providers requires manual intervention.

## Changes
- Added the `--hostname` config flag and `ROLLUP_HOSTNAME` env var
- Added the `--namespace` config flag and `ROLLUP_NAMESPACE` env var
- Updated the `astria-cli rollup deployments list` command to list across all namespaces

## Testing
Updated automated tests.

### DIFF
--- a/crates/astria-cli/src/cli/rollup.rs
+++ b/crates/astria-cli/src/cli/rollup.rs
@@ -112,7 +112,7 @@ pub struct ConfigCreateArgs {
     pub hostname: String,
     /// Configures the k8s namespace rollup will be deployed to
     #[clap(long, env = "ROLLUP_NAMESPACE", default_value = DEFAULT_NAMESPACE)]
-    pub(crate) namespace: String,
+    pub namespace: String,
 }
 
 /// `GenesisAccountArg` is a struct that represents a genesis account to be funded.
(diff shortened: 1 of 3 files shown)


crates/astria-cli/src/cli/rollup.rs
```diff
@@ -100,6 +102,17 @@ pub struct ConfigCreateArgs {
...
+    /// configure other hosts
+    #[clap(
+        long = "hostname", 
+        env = "ROLLUP_HOSTNAME", 
+        default_value = DEFAULT_HOSTNAME
+    )]
+    pub hostname: String,
+    /// Configures the k8s namespace rollup will be deployed to
+    #[clap(long, env = "ROLLUP_NAMESPACE", default_value = DEFAULT_NAMESPACE)]
+    pub(crate) namespace: String,
```
noot, line 115: why is this `pub(crate)` but everything else is `pub`?

noot (approved):

crates/astria-cli/src/cli/rollup.rs
joroshiba: I moved it from a different config and missed. Updating.

steezeburger (approved):

PR #537, opened 2023-10-24
config: define service configs in terms of a central crate

SuperFluffy:
## Summary
Provide a central crate for creating astria service configs.

## Background
There was a lot of code duplication for setting up config for astria services. This bundles it in one place.

## Changes
- Define a `astria-config` crate.
- Rewrite all service tests and definitions in terms of this crate.
- Use unique test prefixes

## Testing
- Tests have have been adjusted and still pass.
- `astria-config` comes with doc tests

## Related Issues

closes https://github.com/astriaorg/astria/issues/240
closes https://github.com/astriaorg/astria/issues/398
closes https://github.com/astriaorg/astria/issues/416


crates/astria-config/src/lib.rs
```diff
@@ -0,0 +1,146 @@
...
+/// Returns the same error as `<T as Config>::get`.
+pub fn get<T: Config>() -> Result<T, Error> {
+    T::get()
+}
+
+/// A utility trait for easily creating a config from the environment.
+///
+/// Works for types allowing serde deserialization. Environment variables
+/// are expected to have the form `<Config::PREFIX><fieldname>`. It is
+/// recommended to set the prefix should with a trailing underscore
```
sambukowski, line 92: can remove "should" so that this reads a little better

SuperFluffy: ty, typos

crates/astria-config/src/lib.rs
```diff
@@ -0,0 +1,146 @@
...
+/// let config: MyConfig = config::get()?;
+/// # Ok::<_, Error>(())
+/// ```
+pub trait Config: core::fmt::Debug + DeserializeOwned {
+    const PREFIX: &'static str;
+
+    /// Creates `Self` by reading its fields from the environment.
+    ///
+    /// # Errors
+    /// Returns an error if the a config field could not be read from
```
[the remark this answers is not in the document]
SuperFluffy: thanks for catching this

crates/astria-config/src/tests.rs
```diff
@@ -0,0 +1,115 @@
...
+            );
+            let prefixed_key = format!("{unique_test_prefix}_{key}");
+            jail.set_env(prefixed_key, val);
+        }
+    }
+}
+
+/// Asserts that a config `C` can be created from a string holding env vars.
+///
+/// Such an example environment string could for example be produced by using
```
sambukowski, line 74: this could be changed to:
An environment string could, for example, be produced using

crates/astria-config/src/tests.rs
```diff
@@ -0,0 +1,115 @@
...
+/// `include_str!` on a file holding an example config.
+///
+/// # Panics
+/// As this is intended to be used as a test options and results are unwrapped,
+/// producing panics.
```
sambukowski, line 79: this sections could be:
As this is intended to be used when testing, panics are produced when bad results are unwrapped.

crates/astria-config/src/tests.rs
```diff
@@ -0,0 +1,115 @@
...
+///
+/// Such an example environment string could for example be produced by using
+/// `include_str!` on a file holding an example config.
+///
+/// This effectively tests that a config has `#[serde(deny_unknown_vars)]` set
+/// as other solutions reuqire more work.
+///
+/// # Panics
+/// As this is intended to be used as a test options and results are unwrapped,
+/// producing panics.
```
sambukowski, line 102: see comment above

crates/astria-config/src/tests.rs
```diff
@@ -0,0 +1,115 @@
...
+    Jail::expect_with(|jail| {
+        populate_environment_from_example(jail, unique_test_prefix, example_env);
+        C::get_with_prefix(&full_test_prefix, _internal::Internal).unwrap();
+        Ok(())
+    });
+}
+
+/// Asserts that a config `C` would reject unknown env vars in string holding env vars.
+///
+/// Such an example environment string could for example be produced by using
```
sambukowski, line 94: see comment above

sambukowski: may also be worth noting that the bad env vars are injected in this test automatically so that the user doesn't need to change their working example env vars

sambukowski (approved):

crates/astria-config/src/tests.rs
```diff
@@ -0,0 +1,115 @@
...
+#[track_caller]
+pub fn example_env_config_is_up_to_date<C: Config>(example_env: &str) {
+    let unique_test_prefix = Lazy::force(&TEST_PREFIX);
+    let full_test_prefix = format!("{unique_test_prefix}_{}", C::PREFIX);
+
+    Jail::expect_with(|jail| {
+        populate_environment_from_example(jail, unique_test_prefix, example_env);
+        C::get_with_prefix(&full_test_prefix, _internal::Internal).unwrap();
+        Ok(())
+    });
```
noot, line 89: does this mean we accept env vars in the form of `ASDF_ASTRIA_CONDUCTOR_..` for example?

noot (approved):

crates/astria-config/src/tests.rs
SuperFluffy: This is more a mechanism to ensure that tests are isolated. Instead of checking the file for `ASTRIA_CONFIG_VARIABLE` we test it for `RNG_PREFIX_ASTRIA_CONFIG_VARIABLE`, but we also prepend all variable names that we find in the example conf by `RNG_PREFIX_`.

PR #543, opened 2023-10-25
feat(cli): improve sequencer command set

joroshiba:
## Summary
Adds a couple commands (transfer, nonce), provides default sequencer rpc for all, and streamlines the ways to access balance.

## Background
Had to interact with the sequencer and needed more commands, found incongruence with defaults provided for rollups and found the balance `get` command to be counter intuitive.

Note that specifically did not remove the balance get command for backwards compatibility, docs.

## Changes
- provide default sequencer rpc url
- provide `sequencer account balance` command as alternative to `sequencer balance get`
- adds `sequencer account nonce` command
- adds `sequencer transfer` command

## Testing
Manually tested

### DIFF
--- a/crates/astria-cli/src/cli/sequencer.rs
+++ b/crates/astria-cli/src/cli/sequencer.rs
@@ -32,7 +32,7 @@ pub enum Command {
         command: BlockHeightCommand,
     },
     /// Command for sending balance between accounts
-    Transfer(TransferArgs)
+    Transfer(TransferArgs),
 }
 
 #[derive(Debug, Subcommand)]
--- a/crates/astria-cli/src/cli/sequencer.rs
+++ b/crates/astria-cli/src/cli/sequencer.rs
@@ -10,6 +10,8 @@ use color_eyre::{
     eyre::Context,
 };
 
+const DEFAULT_SEQUENCER_RPC: &str = "https://rpc.sequencer.dusk-1.devnet.astria.org";
+
 /// Interact with a Sequencer node
 #[derive(Debug, Subcommand)]
 pub enum Command {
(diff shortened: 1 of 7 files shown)


sambukowski (approved):

crates/astria-cli/src/cli/sequencer.rs
```diff
@@ -10,6 +10,8 @@ use color_eyre::{
     eyre::Context,
 };
 
+const DEFAULT_SEQUENCER_RPC: &str = "https://rpc.sequencer.dusk-1.devnet.astria.org";
```
noot, line 13: this is already defined as `rollup::DEFAULT_SEQUENCER_RPC`?

crates/astria-cli/src/cli/sequencer.rs
```diff
@@ -29,29 +31,56 @@ pub enum Command {
...
+#[derive(Args, Debug)]
+pub struct TransferArgs {
+    // The address of the Sequencer account to send amount to
+    pub(crate) to_address: SequencerAddressArg,
+    // The amount being sent
+    #[clap(long)]
+    pub(crate) amount: u128,
+    /// The private key of account being sent from
+    #[clap(long, env = "SEQUENCER_PRIVATE_KEY")]
+    pub(crate) private_key: String,
```
noot, line 74: is this the private key itself, or a file which contains the key? also what's the encoding (hex)?

crates/astria-cli/src/commands/sequencer.rs
```diff
@@ -103,6 +138,58 @@ pub(crate) async fn get_block_height(args: &BlockHeightGetArgs) -> eyre::Result<
...
+        })],
+    }
+    .into_signed(&sequencer_key);
+    let res = sequencer_client
+        .submit_transaction_commit(tx)
+        .await
+        .wrap_err("failed to submit transfer transaction")?;
+
+    ensure!(res.tx_result.code.is_ok(), "error with transfer");
+    println!("Transfer completed!");
```
noot, line 188: maybe print tx hash also?

noot (approved):

crates/astria-cli/src/cli/sequencer.rs
SuperFluffy, line 74: If this is a clear text private key then please zero it out at drop.

Use the zeroize crate for that.

SuperFluffy (changes requested):

crates/astria-cli/src/cli/sequencer.rs
joroshiba: The private key itself, hex encoded.

crates/astria-cli/src/commands/sequencer.rs
joroshiba: Did we add a function for the tx hash somewhere?

crates/astria-cli/src/cli/sequencer.rs
joroshiba: We should definitely come up with a standard here, I agree for services but I question if this is correct behavior for a CLI.

I'm handling this the same as we are handling private keys in CLI rollup deployment.

crates/astria-cli/src/cli/sequencer.rs
joroshiba: I can move shared defaults out to it's own module?

I think we ideally avoid sequencer commands being dependent on rollup commands (and vice versa)

PR #548, opened 2023-10-26
fix: implement `chain_ids_commitment` inclusion proof generation and verification

noot:
## Summary
similarly to `action_tree_root`, we need to prove inclusion of `chain_ids_commitment` inside `data_hash` and verify it when receiving `SequencerBlockData`.

## Background
we need this to prove that no chain IDs were omitted when writing to DA by an untrusted relayer.

## Changes
- add `chain_ids_commitment_inclusion_proof` to `SequencerBlockData` and `RawSequencerBlockData`
-  implement `chain_ids_commitment` inclusion proof generation in `SequencerBlockData::from_tendermint_block()`
- implement `chain_ids_commitment` inclusion proof verification inside `SequencerBlockData::try_from_raw`

## Testing
unit tests

## Breaking Changelist
- `SequencerBlockData` written to DA must now contain the `chain_ids_commitment_inclusion_proof`

## Related Issues
n/a

### DIFF
--- a/crates/astria-celestia-client/src/blob_space.rs
+++ b/crates/astria-celestia-client/src/blob_space.rs
@@ -219,6 +219,7 @@ pub struct SequencerNamespaceData {
     pub action_tree_root: [u8; 32],
     pub action_tree_root_inclusion_proof: InclusionProof,
     pub chain_ids_commitment: [u8; 32],
+    pub chain_ids_commitment_inclusion_proof: InclusionProof,
 }
 
 #[derive(Debug, thiserror::Error)]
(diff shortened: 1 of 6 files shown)


joroshiba (approved):

SuperFluffy (approved):
Thank you, this is great to have!

As a followup we should make all fields of `SequencerNamespaceData` private and instead introduce an `UnverifiedSequencerNamespaceData` type (or `RawSequencerNamespaceData`) that is used for serialization. We can then have `SequencerNamespaceData::try_from_unverified` that upholds various invariants (like the chain IDs being valid).

sambukowski (approved):

PR #551, opened 2023-10-26
conductor: empty blocks from da get executed

### LINKED ISSUE #542: feat(conductor): empty blocks / "continuous blocks" when using firm only
Currently when using the `FirmOnly` setting in Conductor, new data from DA/Celestia only shows up in the rollup when that data is written fresh to DA. 
This results in very irregular block times, or no blocks appearing at all, even though the execution of empty blocks should be happening.

The main issue appears to be that even though Celestia is getting polled every 3 seconds from conductor, only blocks with data in them are being seen and sent to the rollup.

The goal for this issue is to pull all blocks from DA (including empty ones) and have ALL of the sequencer blocks in the DA block get sent to the rollup, especially if those blocks are empty.

sambukowski:
## Summary
Previously empty blocks from DA were getting ignored. This was not the correct behavior.
This fix allows the Conductor to now execute empty blocks from DA.

## Changes
- Empty sequencer blobs from DA are now converted to `SequencerBlockSubset` 
- All empty blocks from DA now actually get sent to execution

## Testing
- Manually tested using dev-cluster

closes #542

### DIFF
--- a/crates/astria-conductor/src/data_availability.rs
+++ b/crates/astria-conductor/src/data_availability.rs
@@ -394,10 +394,6 @@ async fn fetch_verify_rollup_blob_and_forward_to_assembly(
             warn!(error, "failed to get rollup data from celestia");
             return;
         }
-        Ok(rollups) if rollups.is_empty() => {
-            info!("could not find rollup data");
-            return;
-        }
         Ok(rollups) => rollups,
     };
 
@@ -410,13 +406,12 @@ async fn fetch_verify_rollup_blob_and_forward_to_assembly(
         block_verifier::validate_rollup_data(rollup, data.data().action_tree_root).is_ok()
     });
     match rollups.len() {
-        0 => info!("all received rollups failed verification; not forwarding to assembler"),
-        1 => {
+        0 | 1 => {
             info!("one rollup successfully verified; forwarding to block assembler");
             let subset = SequencerBlockSubset {
                 block_hash: data.data().block_hash,
                 header: data.data().header.clone(),
-                rollup_transactions: rollups.swap_remove(0).rollup_txs,
+                rollup_transactions: rollups.pop().map_or(vec![], |txs| txs.rollup_txs),
             };
             if block_tx.send(subset).await.is_err() {
                 warn!("failed sending validated rollup data to block assembler; receiver dropped");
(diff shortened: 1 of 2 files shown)


joroshiba (approved):

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -304,56 +304,59 @@ impl Executor {
...
-                self.update_commitments(executed_block)
-                    .await
-                    .wrap_err("executor failed to update both commitments")?;
-                // remove the sequencer block hash from the map, as it's been firmly committed
-                self.sequencer_hash_to_execution_block
-                    .remove(&block.block_hash);
-            }
-        };
+        }
+        for block in blocks {
```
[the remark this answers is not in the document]
sambukowski: Yes. The da blocks are getting sorted here: https://github.com/astriaorg/astria/blob/954ac8b117240f7790ab6a2a3b11cd3e41b1b493/crates/astria-conductor/src/data_availability.rs#L434

crates/astria-conductor/src/data_availability.rs
```diff
@@ -410,13 +406,12 @@ async fn fetch_verify_rollup_blob_and_forward_to_assembly(
         block_verifier::validate_rollup_data(rollup, data.data().action_tree_root).is_ok()
     });
     match rollups.len() {
-        0 => info!("all received rollups failed verification; not forwarding to assembler"),
-        1 => {
+        0 | 1 => {
             info!("one rollup successfully verified; forwarding to block assembler");
```
noot, line 410: update the log here since it's now 0 or 1 rollup datas

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -304,56 +304,59 @@ impl Executor {
...
+                    // - we didn't receive the block from the sequencer stream, or
+                    // - we received it, but the sequencer block didn't contain
+                    // any transactions for this rollup namespace, thus nothing was executed
+                    // on receiving this block.
+
+                    // try executing the block as it hasn't been executed before
+                    // execute_block will check if our namespace has txs; if so, it'll return the
+                    // resulting execution block hash, otherwise None
+                    let Some(executed_block) = self
+                        .execute_block(block.clone())
```
noot, line 338: change `for block in blocks` to `for block in blocks.into_iter()` and i think you can remove the clone here

and add a `let block_hash = block.block_hash;` above this

noot (approved):

Issue #530, opened 2023-10-20
feat(conductor): stateless restarts


joroshiba:
Using the v1alpha2 API we can enable conductor to statelessly restart.

Currently, upon restart it will sync blocks which have already been executed on the rollup node. We can use the `initial_sequencer_block_height` config value + grabbed from rollup node`commitment_state.soft.number` to determine the first block from sequencer which needs to be executed.

There are two versions of this: 
- a simple pass which simply skips blocks until a given height
- more complex which starts sync from the correct sequencer height, this will require some refactor but is the better solution.

Blocked by #411, as requires the v1alpha2 api.

Related to #15 around execution ordering

PR #553, opened 2023-10-27
feat(conductor): added commitment grab to better set sync start height

sambukowski:
## Summary
Conductor now uses the initial commitment state that is pulled when Conductor starts to set the initial sequencer block height more accurately so that sync at startup runs faster.

Basic checks were also put in place to prevent out of order blocks from executing when getting passed to `execute_block`.

## Background
When starting up, Conductor was syncing from it's block 1 even if blocks have already been executed on the rollup. This is inefficient. With the implementation of the execution api v1alpha2, the latest executed block from the rollup can now be pulled and the sync can start where normal execution left off.

## Changes
- Added checks to `execute_block` to prevent out of order execution
- Hight of latest soft block is now pulled from rollup and used to set the block height that sync starts at.

## Testing
Manually tested using dev-cluster

closes https://github.com/astriaorg/astria/issues/530

### DIFF
--- a/crates/astria-conductor/src/conductor.rs
+++ b/crates/astria-conductor/src/conductor.rs
@@ -82,7 +82,7 @@ impl Conductor {
         let signals = spawn_signal_handler();
 
         // Spawn the executor task.
-        let (executor_tx, init_sequencer_height) = {
+        let (executor_tx, sync_start_block_height) = {
             let (block_tx, block_rx) = mpsc::unbounded_channel();
             let (shutdown_tx, shutdown_rx) = oneshot::channel();
             let executor = Executor::new(
@@ -96,12 +96,12 @@ impl Conductor {
             )
             .await
             .wrap_err("failed to construct executor")?;
-            let init_sequencer_height = executor.executable_block_height;
+            let executable_sequencer_block_height = executor.executable_block_height;
 
             tasks.spawn(Self::EXECUTOR, executor.run_until_stopped());
             shutdown_channels.insert(Self::EXECUTOR, shutdown_tx);
 
-            (block_tx, init_sequencer_height)
+            (block_tx, executable_sequencer_block_height)
         };
 
         let sequencer_client_pool = client_provider::start_pool(&cfg.sequencer_url)
@@ -126,7 +126,7 @@ impl Conductor {
             let (sync_done_tx, sync_done_rx) = oneshot::channel();
 
             let sequencer_reader = sequencer::Reader::new(
-                init_sequencer_height,
+                sync_start_block_height,
                 sequencer_client_pool.clone(),
                 shutdown_rx,
                 executor_tx.clone(),
--- a/crates/astria-conductor/src/conductor.rs
+++ b/crates/astria-conductor/src/conductor.rs
@@ -7,10 +7,15 @@ use std::{
 use astria_sequencer_types::ChainId;
 use color_eyre::eyre::{
     self,
+    Result,
     WrapErr as _,
 };
 // use futures::FutureExt;
 use futures::future::Fuse;
+use proto::generated::execution::v1alpha2::{
+    execution_service_client::ExecutionServiceClient,
+    GetCommitmentStateRequest,
+};
 use tokio::{
     select,
     signal::unix::{
@@ -44,6 +49,7 @@ use crate::{
     data_availability,
     executor::Executor,
     sequencer,
+    types::ExecutorCommitmentState,
     Config,
 };
 
@@ -119,11 +125,16 @@ impl Conductor {
         }
 
         if !cfg.execution_commit_level.is_firm_only() {
-            // todo!("set up stuff");
             let (shutdown_tx, shutdown_rx) = oneshot::channel();
             let (sync_done_tx, sync_done_rx) = oneshot::channel();
+
+            // get the commitment state from the execution client
+            let commitment_state = get_commitment_state(cfg.execution_rpc_url.to_owned())
+                .await
+                .wrap_err("failed to get commitment state")?;
+
             let sequencer_reader = sequencer::Reader::new(
-                cfg.initial_sequencer_block_height,
+                cfg.initial_sequencer_block_height + commitment_state.soft.number,
                 sequencer_client_pool.clone(),
                 shutdown_rx,
                 executor_tx.clone(),
@@ -332,4 +343,18 @@ fn spawn_signal_handler() -> SignalReceiver {
         reload_rx,
         stop_rx,
     }
+}
+
+async fn get_commitment_state(rpc_url: String) -> Result<ExecutorCommitmentState> {
+    let mut execution_rpc_client = ExecutionServiceClient::connect(rpc_url)
+        .await
+        .wrap_err("failed to create execution rpc client")?;
+    let commitment_state_resp = execution_rpc_client
+        .get_commitment_state(GetCommitmentStateRequest {})
+        .await
+        .wrap_err("executor failed to get commitment state")?;
+    let commitment_state = ExecutorCommitmentState::from_execution_client_commitment_state(
+        commitment_state_resp.get_ref().clone(),
+    );
+    Ok(commitment_state)
 }
--- a/crates/astria-conductor/src/conductor.rs
+++ b/crates/astria-conductor/src/conductor.rs
@@ -90,7 +90,8 @@ impl Conductor {
                 ChainId::new(cfg.chain_id.as_bytes().to_vec())
                     .wrap_err("failed to create chain ID")?,
                 cfg.disable_empty_block_execution,
-                cfg.initial_sequencer_block_height, // the sequencer block the rollup was start on
+                // the sequencer block that contains the first rollup block
+                cfg.initial_sequencer_block_height,
                 block_rx,
                 shutdown_rx,
             )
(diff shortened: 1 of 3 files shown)


joroshiba (approved):

ygd58:
hi i like and follow your project and i want to collaborate i am a member of core node team we have participated in many projects and we are an experienced team in validators and verifiers and we have projects in collaboration with opside and goracle network cascadia and many of our team members are involved in important projects we also have a core node community. I will send you detailed information about core node github and team members. I think you might want to get to know us thanks. We can manage the Turkey region for your project. We can open official Telegram twitter accounts. We can help node installers and those who have problems. Gökhan, one of our giant teams, is very familiar with your project. We also have an active community of 425 people. If we share, we can help you promote your node or project on Twitter
https://corenode.info/
https://twitter.com/corenodeHQ
https://github.com/core-node-team
https://services.corenode.info/

ygd58:
We have an experienced team, I could write here to communicate, sorry.

crates/astria-conductor/src/conductor.rs
```diff
@@ -83,22 +82,26 @@ impl Conductor {
...
                 ChainId::new(cfg.chain_id.as_bytes().to_vec())
                     .wrap_err("failed to create chain ID")?,
                 cfg.disable_empty_block_execution,
+                cfg.initial_sequencer_block_height,
                 block_rx,
                 shutdown_rx,
             )
             .await
             .wrap_err("failed to construct executor")?;
+            let init_sequencer_height = executor.executable_block_height;
```
noot, line 99: name this `executable_sequencer_block_height` or something as it isn't the `initial_sequencer_block_height`

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -128,12 +132,15 @@ impl Executor {
             .await
             .wrap_err("executor failed to get commitment state")?;
 
+        let executable_block_height = commitment_state.soft.number + init_sequencer_height;
```
noot, line 135: this is only correct if we are executing empty blocks - should we update conductor to always execute empty blocks? otherwise this will not work

SuperFluffy, line 135: Can you add a comment explaining this equation?

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -222,6 +229,29 @@ impl Executor {
...
+                return Ok(None);
+            }
+            std::cmp::Ordering::Greater => {
+                debug!(
+                    sequencer_block_height = block.header.height.value(),
+                    executable_block_height = self.executable_block_height,
+                    "skipping execution of stale block"
+                );
+                return Ok(None);
+            }
```
noot, line 251: either return an error, or update the function comment `/// if there are no relevant transactions in the SequencerBlock, it returns None.`

i think this should be an error, as `execute_block` shouldn't be called with a block that can't be executed

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -222,6 +229,29 @@ impl Executor {
             return Ok(None);
         }
 
+        match self
+            .executable_block_height
+            .cmp(&(block.header.height.value() as u32))
+        {
+            std::cmp::Ordering::Less => {
```
SuperFluffy, line 236: I think it's ok to roll this into one. The distinction between `future` and `stale` will be lost when quickly scanning the logs.

If a `if self.executable_block_height != block.header.height` { ... }` is cleaner.

crates/astria-conductor/src/executor/tests.rs
```diff
@@ -151,6 +151,7 @@ async fn start_mock(disable_empty_block_execution: bool) -> MockEnvironment {
         &server_url,
         chain_id,
         disable_empty_block_execution,
+        1,
```
SuperFluffy, line 154: Can you look into providing a test for skipping this block?

crates/astria-conductor/src/conductor.rs
```diff
@@ -119,11 +122,11 @@ impl Conductor {
         }
 
         if !cfg.execution_commit_level.is_firm_only() {
-            // todo!("set up stuff");
             let (shutdown_tx, shutdown_rx) = oneshot::channel();
             let (sync_done_tx, sync_done_rx) = oneshot::channel();
+
             let sequencer_reader = sequencer::Reader::new(
-                cfg.initial_sequencer_block_height,
+                init_sequencer_height,
```
SuperFluffy, line 129: The relationship between `init_sequencer_height` and how it derives from executor is very obscure at this point. Can you add some comments explaining the relationship?

crates/astria-conductor/src/executor/tests.rs
sambukowski: Added some new tests!

crates/astria-conductor/src/executor/mod.rs
sambukowski: pr to remove disable empty block execution here: https://github.com/astriaorg/astria/pull/556

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -128,12 +133,29 @@ impl Executor {
             .await
             .wrap_err("executor failed to get commitment state")?;
 
+        // The `executable_block_height` is the height of the sequencer block
+        // where the Conductor sync will be started. The `init_sequencer_height`
```
joroshiba, line 137: I think it's strictly the sequencer block height which is executable on top of the current rollup state. That sync starts from this height is not the concern of executor, it only cares that it executes the correct block.

crates/astria-conductor/src/conductor.rs
```diff
@@ -119,11 +122,18 @@ impl Conductor {
...
             let (shutdown_tx, shutdown_rx) = oneshot::channel();
             let (sync_done_tx, sync_done_rx) = oneshot::channel();
+
+            // the `sync_start_block_height` below is the height at which the sync
+            // of Conductor will start. This value is a combination of the
+            // initial sequencer block height, which is set in the config when
+            // the rollup is first created, plus the height of the most recently
+            // executed block on the rollup.
+            // sync_start_block_height =
+            //   initial sequencer block height + most recently executed block height
```
joroshiba, line 134: I think we can simplify this description a bit. The sequencer reader starts sync at the height which is the next executable block. How that's derived is more of an executor concern.

crates/astria-conductor/src/executor/mod.rs
sambukowski: updated the comment

crates/astria-conductor/src/conductor.rs
sambukowski: updated the comment

crates/astria-conductor/src/conductor.rs
```diff
@@ -83,22 +82,26 @@ impl Conductor {
...
-        let executor_tx = {
+        let (executor_tx, sync_start_block_height) = {
             let (block_tx, block_rx) = mpsc::unbounded_channel();
             let (shutdown_tx, shutdown_rx) = oneshot::channel();
             let executor = Executor::new(
                 &cfg.execution_rpc_url,
                 ChainId::new(cfg.chain_id.as_bytes().to_vec())
                     .wrap_err("failed to create chain ID")?,
                 cfg.disable_empty_block_execution,
+                cfg.initial_sequencer_block_height, // the sequencer block the rollup was start on
```
noot, line 93:
```suggestion
                cfg.initial_sequencer_block_height, // the sequencer block that contains the first rollup block
```

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -128,12 +133,29 @@ impl Executor {
             .await
             .wrap_err("executor failed to get commitment state")?;
 
+        // The `executable_block_height` is the height of the fist sequencer block
```
noot, line 136:
```suggestion
        // The `executable_block_height` is the height of the next sequencer block
```

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -128,12 +133,29 @@ impl Executor {
...
+        // that can be executed on top of the rollup state. The `init_sequencer_height`
+        // is set when the rollup is first created and let's us know that this
```
noot, line 138:
```suggestion
        // is set when the rollup is first created and lets us know that this
```

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -128,12 +133,29 @@ impl Executor {
...
+        // rollup's genesis block is in block K of the sequencer chain.
```
noot, line 139:
```suggestion
        // rollup's first block is in block K of the sequencer chain.
```

genesis block is defined in the genesis file, it's the rollup block 1 that's in a sequencer block (as it's the first executable block)

noot (approved):
looks good, only minor comments.

out of scope for this PR, but I find having the `initial_sequencer_block_height` as a config option bad, as it's not something that should be changeable from run to run, it's something that should be a part of the rollup's genesis. imo it's more extensible for rollup developers if this is stored on the execution node (alternatively, the sequencer block hash corresponding to the rollup block hash is stored). this involves updating the API, which i think we should consider.

PR #554, opened 2023-10-30
feat: add an RFC 6962 compliant Merkle tree with flat memory representation

SuperFluffy:
## Summary
Implement a RFC 6962 compliant merkle tree with flat memory representation, remove ct-merkle.

## Background
We want to use a merkle tree that minimizes allocations because constructing merkle trees is a hot operation that happens every time new blocks are generated or verified. Worse, for blocks containing transactions for several distinct rollups a merkle Tree is created per rollup. Finally, Astria wants a minimal API that is tailored exactly to its needs to make audits as simple as possible.

The implementation of this merkle tree library does not allocate space for its leaves but only stores leaf and branch hashes in a flat, one dimensional array. Furthermore, it only supports sha256 hashes. For background reading refer to the following links:

- https://datatracker.ietf.org/doc/html/rfc6962 (reference merkle tree this implementation adheres to)
- https://www.datprotocol.com/deps/0002-hypercore/
- https://datatracker.ietf.org/doc/rfc7574
- https://www.ietf.org/archive/id/draft-ietf-mls-protocol-14.html
- https://mmapped.blog/posts/22-flat-in-order-trees (this is reference this implementation builds on in terms of memory representation)

## Changes
- Add  an `astria-merkle` crate implementing a flat merkle tree.
- Rewrite all services/crates in terms of it.
- Simplify several utility functions by implementing APIs on the merkle tree, avoiding allocations.

## Testing
- Ensure the merkle tree passes all tests with known roots defined by https://github.com/transparency-dev/merkle
- Add unit and doc tests for all high level APIs.
- Ensure that all snapshot tests in astria sequencer still pass without changing the expdcted hashes

## Breaking Changelist
- Sequencer block data and sequencer/rollup blobs are changed, breaking compatibility with prior versions of sequencer, sequencer-relayer, conductor.


crates/astria-merkle/tests/consistency.rs
SuperFluffy, line 1: Remove, we currently don't do consistency proofs. I am not sure we need them either.

crates/astria-celestia-client/src/client.rs
```diff
@@ -199,23 +198,18 @@ pub enum BlobAssemblyError {
         source: serde_json::Error,
         index: usize,
     },
-    #[error("failed to generate inclusion proof for the transaction at index `{index}`")]
-    GenerateInclusionProof {
-        source: IndexOutOfBounds,
-        index: usize,
-    },
+    #[error(
+        "failed to construcgt inclusion proof for the transaction at index `{index}` because its \
```
sambukowski, line 202: spelling: construct

SuperFluffy: ty

crates/astria-celestia-client/src/blob_space.rs
```diff
@@ -84,12 +82,17 @@ impl RollupNamespaceData {
...
-        self.inclusion_proof.verify(&leaf, root_hash)?;
+        let mut leaf = Vec::with_capacity(self.chain_id.as_ref().len() + rollup_data_root.len());
+        leaf.extend_from_slice(self.chain_id.as_ref());
+        leaf.extend_from_slice(&rollup_data_root);
+        let reconstructed_root = self.inclusion_proof.reconstruct_root_with_leaf(&leaf);
+        if root_hash != reconstructed_root {
+            return Err(RollupVerificationFailure {
+                _private: (),
+            });
+        }
```
noot, line 95: i prefer the `inclusion_proof.verify(leaf, root_hash)` API, thoughts?

crates/astria-merkle/src/lib.rs
```diff
@@ -0,0 +1,718 @@
+//! A RFC 6962 compliant merkle tree with a flat representation.
```
noot, line 1: maybe mention "except consistency proofs" since we don't need to implement those?

crates/astria-merkle/src/lib.rs
```diff
@@ -0,0 +1,718 @@
...
+#[must_use]
+pub fn combine(left: &[u8], right: &[u8]) -> [u8; 32] {
+    let mut hasher = Sha256::new();
+    hasher.update([0x01_u8]);
+    hasher.update(left);
+    hasher.update(right);
+    hasher.finalize().into()
+}
+
+/// The proof that a node is included of a Merkle tree.
```
noot, line 108:
```suggestion
/// The proof that a node is included in a Merkle tree.
```

crates/astria-merkle/src/lib.rs
```diff
@@ -0,0 +1,718 @@
...
+///
+/// The proof is the concatenation of all merkle tree hashes from
+/// a node to the root.
```
noot, line 111:
```suggestion
/// The proof is the concatenation of all sibling hashes required to
/// reconstruct the root from a given node.
```

something like this might be clearer?

crates/astria-merkle/src/lib.rs
```diff
@@ -0,0 +1,718 @@
...
+        for sibling in audit_path.chunks(32) {
+            let parent = complete_parent(i, *tree_size);
+            if parent > i {
+                acc = combine(&acc, sibling);
+            } else {
+                acc = combine(sibling, &acc);
+            }
+            i = parent;
+        }
+        acc
```
noot, line 205: could there be an error in the case where the proof is too long/too short? should there be a check that the proof is the expected length?

crates/astria-merkle/src/lib.rs
```diff
@@ -0,0 +1,718 @@
...
+    /// # Examples
+    /// ```
+    /// let mut tree_1 = astria_merkle::Tree::new();
+    /// let mut builder = tree_1.build_leaf();
+    /// builder.write(b"hello");
+    /// builder.write(b"world");
+    /// builder.finish();
+    ///
+    /// let mut tree_2 = astria_merkle::Tree::new();
+    /// tree_2.build_leaf().write(b"hello").write(b"world");
```
noot, line 271:
```suggestion
    /// tree_2.build_leaf().write(b"helloworld");
```

the two trees in the example are built the same way, is this understanding correct?

noot: also `finish` isn't called for `tree_2`

crates/astria-merkle/src/lib.rs
```diff
@@ -0,0 +1,718 @@
...
+            .take()
+            .expect("hasher is set during the leaf builder's lifetime and only taken on drop")
+            .finalize()
+            .into();
+        if tree.nodes.is_empty() {
+            tree.nodes.extend_from_slice(&leaf_hash);
+            return;
+        }
+        // append 2 * 32 = 64 zeros
+        tree.nodes.extend_from_slice(&[0; 64]);
```
noot, line 306: why 64?

noot: isn't it possible that adding the new leaf adds a new level to the tree, thus adding `log2(new_len)` nodes?

crates/astria-merkle/src/lib.rs
SuperFluffy: This is just for demonstration purposes to show the `build_leaf` API. :-) This is used to avoid extra allocations.

Maybe I should choose another example?

crates/astria-merkle/src/lib.rs
SuperFluffy: This comes from memory structure of the tree: For each leaf-node you have to push 1 extra branch node so odd-indexed elements are leaves, even-indexed elements are branches.

crates/astria-merkle/src/lib.rs
SuperFluffy: > also `finish` isn't called for `tree_2`

Just saw that after commenting: `finish` need not be called because the tree is updated on drop. See the comment on this method: 
```
    /// Usually calling this is not necessary because writes to the builder can
    /// be chained, but this is still useful if some other operation has to happen
    /// between writes, requiring the builder to be found.
```

crates/astria-merkle/src/lib.rs
SuperFluffy: > isn't it possible that adding the new leaf adds a new level to the tree, thus adding `log2(new_len)` nodes?

No, that does not happen. You can see that if you draw the evolution of the tree:

```
# Single leaf
x

# Two leaves
  x
 /  \
x  x

# Three leaves
      x
    /   \ 
  x     |
 /  \    |
x  x  x
```
Have a look at the module doc comment at the top of `lib.rs` that shows the evolution and indexing of the tree.

crates/astria-merkle/src/lib.rs
```diff
@@ -0,0 +1,718 @@
...
+    assert!(i < n);
+    let right_child = perfect_right_child(i);
+    if right_child < n {
+        right_child
+    } else {
+        i + 1 + complete_root(n - i - 1)
+    }
+}
+
+/// Returns the parent and sibling of a node at index `i` of a complete binary tree of size `n`.
```
noot, line 692:
```suggestion
/// Returns the parent and sibling infex of a node at index `i` of a complete binary tree of size `n`.
```

crates/astria-merkle/src/lib.rs
```diff
@@ -0,0 +1,718 @@
...
+    where
+        I: IntoIterator<Item = B>,
+        B: AsRef<[u8]>,
+    {
+        let mut tree = Self::new();
+        for item in iter {
+            tree.push(item.as_ref());
+        }
+        tree
+    }
```
noot, line 533: since `push` updates the tree every time it's called, it would be nice to have a `from_leaves` function that doesn't call `push`, but just constructs the tree directly from all the leaves. this way, we can do 1 bigger allocation at once since we know the final tree size, and don't have to hash intermediates unnecessarily. thoughts?

crates/astria-celestia-client/src/blob_space.rs
noot: yes, `fn(&self, leaf: &[u8], root: [u8; 32]) -> bool` seems good!

crates/astria-merkle/src/lib.rs
noot: ah okay, yeah i just meant it's a bit weird that `tree_1` and `tree_2` in the examples are built the same :p

crates/astria-merkle/tests/inclusion.rs
```diff
@@ -0,0 +1,100 @@
...
+    leaf: usize,
+    expected_proof: &'static [u8],
+}
+
+const INCLUSION_TESTS: &[InclusionTest] = &[
+    InclusionTest {
+        num_leaves: 1,
+        leaf: 0,
+        expected_proof: &[],
+    },
```
noot, line 30: how does this work? if there's one leaf, it is set as the root?

crates/astria-merkle/tests/inclusion.rs
```diff
@@ -0,0 +1,100 @@
...
+    &hex!("606162636465666768696a6b6c6d6e6f"),
+];
+
+struct InclusionTest {
+    num_leaves: usize,
+    leaf: usize,
+    expected_proof: &'static [u8],
+}
+
+const INCLUSION_TESTS: &[InclusionTest] = &[
```
noot, line 25: can you add a test for the empty tree (leaves = 0) also?

noot (approved):
a few minor comments, approving since overall looks good and don't want to block. great work!!

crates/astria-merkle/tests/inclusion.rs
SuperFluffy: If there is only one leaf then the expected proof is empty as there are no siblings, i.e. the copath is empty.

crates/astria-merkle/tests/inclusion.rs
SuperFluffy: I added an example to the `Tree::construct_proof` doctest that demonstrates the empty tree leading to `None`.

crates/astria-merkle/src/lib.rs
SuperFluffy: I would love to have that, but to really benefit from this we would need to walk the tree level by level and bottom to top, and only update branches at each level.

But the math comes out a bit annoying and is IMO out of scope for this PR. I would want to spend some time thinking about how to do that.

A thought: according the indexing scheme the trailing set bits indicate the depth of the subtree. So the root for a tree of depth `D` has a binary representation of `011...1` with `D-1` ones. This extends to every subtree. This means we should in principle be able to walk the tree by "roots-of-subtrees" by fixing the number of trailing 1s for each level.

crates/astria-merkle/src/lib.rs
SuperFluffy: Good point, done.

crates/astria-merkle/src/lib.rs
SuperFluffy: I have added a 3rd tree to that example that demonstrates that the two above methods are equivalent to calling `push(b"helloworld");`

crates/astria-celestia-client/src/blob_space.rs
SuperFluffy: I have renamed `Proof::is_leaf_in_tree` -> `Proof::verify`, and added the audit API with `Audit::perform`.

crates/astria-merkle/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "astria-merkle"
+version = "0.1.0"
+edition = "2021"
```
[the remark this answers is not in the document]
SuperFluffy: Added 1.59.0 as determined by cargo-msrv (1.58.1 is incompatible due to destructuring assignments).

crates/astria-merkle/src/lib.rs
```diff
@@ -0,0 +1,683 @@
...
+fn complete_left_child(p: usize) -> usize {
+    perfect_left_child(p)
+}
+
+/// Returns the left child of a branch node `p` of a complete binary tree.
+///
+/// Note: `complete binary tree` here refers to a tree in which all left subtrees
+///       are perfect, which is a stronger assumption than just "complete". Therefore
+///       the left child of `p` is the same as for the perfect binary tree.
+// TODO: check edge cases
```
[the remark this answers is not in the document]
SuperFluffy: removed this comment for now.

itamarreif (approved):
Look good. Only note would be that we should eventually write a `from_leaves()` method that's more efficient with the memory allocations, as @noot mentioned. But this could easily be added later.

Block data construction seems solid and largely matches our spec as described in [`data flow and verification`](https://github.com/astriaorg/astria/blob/main/specs/data-flow-and-verification.md). Only thing that should be updated there is the types: we no longer have `InclusionProof`, instead we have `merkle::Proof`.

Another note is that we should probably also add a document in the specs directory for how the Merkle tree implementation works. This information largely lives as docstrings in the `astria-merkle` crate in this PR which is also fine, though.

SuperFluffy:
I have amended the spec to be in line with this patch, and created https://github.com/astriaorg/astria/issues/587 to track for a more efficient `from_leaves` implementation, and https://github.com/astriaorg/astria/issues/588 to track writing a spec doc.

PR #556, opened 2023-10-30
feat(conductor): remove disable empty block execution config setting

### LINKED ISSUE #555: Remove disable empty block execution config setting
To make sure that the Conductor sync works predictably and that some of the recent optimizations work properly, it has been determined that removing the option to not execute empty blocks should be removed in Conductor.

sambukowski:
## Summary
The 'DISABLE_EMPTY_BLOCK_EXECUTION` config option from Conductor.

## Background
To make sure that the Conductor sync works reliably, it was determined that the option to disable empty block execution should be removed. The makes all block times regular and some of the optimizations to the Conductor sync more efficient.

## Changes
- `DISABLE_EMPTY_BLOCK_EXECUTION` removed
- Updated `astria_conductor::executor::Executor::execute_block` to no longer return an option.
- Updated tests to make sure that executing empty blocks and blocks with transactions work the same way

## Testing
passes cargo code test
manual testing using dev-cluster

## Breaking Changelist
- Requires updates to charts in dev cluster

closes #555


crates/astria-conductor/src/executor/mod.rs
```diff
@@ -214,14 +209,6 @@ impl Executor {
     /// execution block hash.
     /// if there are no relevant transactions in the SequencerBlock, it returns None.
     async fn execute_block(&mut self, block: SequencerBlockSubset) -> Result<Option<Block>> {
```
noot, line 211: change this to just return `Result<Block>` and update the comment above

noot (approved):
please address comment before merging!!

crates/astria-conductor/src/executor/mod.rs
SuperFluffy: ack, this signature can now be simplified

joroshiba (approved):

crates/astria-conductor/src/executor/mod.rs
sambukowski: updated!

PR #558, opened 2023-10-31
fix(sequencer): run only `prepare_proposal` if proposer

noot:
## Summary
turns out both `prepare_proposal` and `process_proposal` are called for a proposer node. since we moved transaction execution to prepare/process_proposal, when the proposer called both these functions, execution would fail during `process_proposal` since the state was already updated for post-execution.

## Background
see summary.

## Changes
- track if we're the proposer during the block lifecycle in the `App` struct, and if we are, don't execute transactions during `process_proposal` 

## Testing
local testing

### DIFF
--- a/crates/astria-conductor/src/executor/mod.rs
+++ b/crates/astria-conductor/src/executor/mod.rs
@@ -213,6 +213,7 @@ impl Executor {
     /// if the block has already been executed, it returns the previously-computed
     /// execution block hash.
     /// if there are no relevant transactions in the SequencerBlock, it returns None.
+    #[instrument(skip(self), fields(sequencer_block_hash = ?block.block_hash, sequencer_block_height = block.header.height.value()))]
     async fn execute_block(&mut self, block: SequencerBlockSubset) -> Result<Option<Block>> {
         if self.disable_empty_block_execution && block.rollup_transactions.is_empty() {
             debug!(
@@ -252,11 +253,8 @@ impl Executor {
 
         // store block hash returned by execution client, as we need it to finalize the block later
         info!(
-            sequencer_block_hash = ?block.block_hash,
-            sequencer_block_height = block.header.height.value(),
             execution_block_hash = hex::encode(&executed_block.hash),
-            tx_count,
-            "executed sequencer block",
+            tx_count, "executed sequencer block",
         );
 
         // store block returned by execution client, as we need it to finalize the block later
--- a/crates/astria-conductor/src/executor/mod.rs
+++ b/crates/astria-conductor/src/executor/mod.rs
@@ -244,6 +244,7 @@ impl Executor {
         let timestamp = convert_tendermint_to_prost_timestamp(block.header.time)
             .wrap_err("failed parsing str as protobuf timestamp")?;
 
+        let tx_count = block.rollup_transactions.len();
         let executed_block = self
             .execution_rpc_client
             .call_execute_block(prev_block_hash, block.rollup_transactions, timestamp)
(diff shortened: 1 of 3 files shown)


crates/astria-conductor/src/executor/mod.rs
```diff
@@ -254,6 +255,7 @@ impl Executor {
             sequencer_block_hash = ?block.block_hash,
             sequencer_block_height = block.header.height.value(),
             execution_block_hash = hex::encode(&executed_block.hash),
+            tx_count,
```
SuperFluffy, line 258: This is getting messy. Can you instrument `execute_block` and move the tx count and the other 2 fields derived from the block into the span?

crates/astria-sequencer/src/app.rs
```diff
@@ -77,6 +77,13 @@ type InterBlockState = Arc<StateDelta<Snapshot>>;
 pub(crate) struct App {
     state: InterBlockState,
 
+    // set to true when `prepare_proposal` is called, indicating we are the proposer for this slot.
+    // set to false when `commit` is called, finalizing the block.
+    // when this is se to true, `process_proposal` is not executed, as we have already executed
```
SuperFluffy, line 82:
```suggestion
    // if true, `process_proposal` is not executed, as we have already executed
```

crates/astria-sequencer/src/app.rs
```diff
@@ -173,6 +182,10 @@ impl App {
         &mut self,
         process_proposal: abci::request::ProcessProposal,
     ) -> anyhow::Result<()> {
+        if self.is_proposer {
+            return Ok(());
```
SuperFluffy, line 186: Add a debug log?

crates/astria-sequencer/src/app.rs
```diff
@@ -448,6 +461,7 @@ impl App {
 
         // Get the latest version of the state, now that we've committed it.
         self.state = Arc::new(StateDelta::new(storage.latest_snapshot()));
+        self.is_proposer = false;
```
SuperFluffy, line 464: Add a comment for why this is necessary to if the reader forgot what `is_propser` is.

SuperFluffy (approved):
Just a few comments.

crates/astria-sequencer/src/app.rs
```diff
@@ -173,6 +182,10 @@ impl App {
         &mut self,
         process_proposal: abci::request::ProcessProposal,
     ) -> anyhow::Result<()> {
+        if self.is_proposer {
```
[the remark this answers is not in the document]
noot: basically everything is here is verification and re-execution, don't think we need to re-do any of it

crates/astria-sequencer/src/app.rs
```diff
@@ -77,6 +77,13 @@ type InterBlockState = Arc<StateDelta<Snapshot>>;
 pub(crate) struct App {
     state: InterBlockState,
 
+    // set to true when `prepare_proposal` is called, indicating we are the proposer for this slot.
+    // set to false when `commit` is called, finalizing the block.
+    // when this is se to true, `process_proposal` is not executed, as we have already executed
+    // the transactions for the block during `prepare_proposal`, and re-executing them would
+    // cause failure.
+    is_proposer: bool,
```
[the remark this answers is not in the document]
noot: yeah, i'll change this to match on the block hash instead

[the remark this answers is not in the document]
noot: yeah i agree, this wouldn't work in that case. i updated this to always set `is_proposer = false` when `process_proposal` is called, so `process_proposal` will skip only in the case where `prepare_proposal` is called directly before it. so if we propose a block, but it's not accepted, and then `process_proposal` is called for the next proposed block, it will execute as normal

sambukowski (approved):

joroshiba (approved):

PR #560, opened 2023-11-01
refactor: verify current block commit in conductor; remove `last_commit` from `SequencerBlockData`

### LINKED ISSUE #401: conductor: remove unnecessary commit verification
as we're now using a trusted sequencer endpoint in conductor, we don't need to validate commits anymore, specifically on data received from DA. we can simply use some `get_header_by_height` endpoint and confirm that the block header (and thus hash) returned (which we know is final) matches the block hash on DA.

as such, we can remove `last_commit` from `SequencerBlockData`.

noot:
## Summary
previously, we were storing a block's `last_commit` in `SequencerBlockData` and verifying it in the conductor. however, we were never doing anything (correctly) with this commit verification - if it failed, we would reject the *current* block, where we should have rejected the previous. also, we don't want to delay block execution 1 block by waiting for the canonical commit in the next block, as we can retrieve a valid commit for the latest block from our sequencer node the moment the block is finalized.

## Background
see summary.

also, note that the tendermint latest block is finalized and thus will always have a commit available for it, so this PR changes the verification code in conductor to fetch and use that commit when verifying a block from DA. we don't care if the commit is canonical, as whichever commit we receive will be valid, since the block would not have finalized without it. (canonical meaning the commit all nodes agreed on, aka `last_commit`. at the moment of finalization, nodes on the network will see different sets of >2/3 votes which consist of their local commit for the block, the "canonical commit" is not chosen until the next block is finalized)

## Changes
-  remove `last_commit` from `SequencerBlockData`
- update conductor block verification code to fetch the commit for *that* block from the sequencer and verify it. `last_commit` is no longer used as it is not relevant.

## Testing
unit tests

## Breaking Changelist
- `SequencerBlockData` no longer contains `last_commit`

## Related Issues

closes #401


sambukowski (approved):

crates/astria-conductor/src/block_verifier.rs
```diff
@@ -22,7 +21,10 @@ use sequencer_client::{
 };
 use tendermint::{
     account,
-    block,
+    block::{
+        self,
+        Commit,
```
SuperFluffy, line 26: Don't import `Commit` directly, just qualify with `block::Commit`.

SuperFluffy (approved):
Looks, I like the simplification.

PR #567, opened 2023-11-07
fix(conductor): remove the bearer token from conductor

joroshiba:
## Summary
Removing the celestia bearer token from conductor

## Background
New versions of celestia node don't require a bearer token to read data, and we are only reading data. This cleans up the flow of using a light client quite a lot. 

## Changes
- remove bearer token from light client. 

## Testing
CI + testing as a part of astriaorg/dev-cluster#124

## Breaking Changelist
- Removes a configuration setting


joroshiba:
I was wrong.... still need token.

PR #568, opened 2023-11-08
feat(sequencer): implement support for arbitrary assets

### LINKED ISSUE #559: sequencer/IBC: support multiple token types and assets for fee payment
as title says - we currently only support the Astria native token on the sequencer app, however for IBC we want to support arbitrary ICS20 token types. as well, we want to be able to support different assets for fee payments, specifically IBC-TIA. 

this will require adding an `Asset` type and updating the transaction format to specify which asset is being used for fee payment or transfer.

noot:
## Summary
This PR implements support for arbitrary assets for fee payments and balance/transfers inside the sequencer. Currently, only one asset exists (the native asset, RIA), but after IBC is implemented we can potentially have many arbitrary assets on the sequencer chain.

## Background
We need to support multiple assets for IBC, and also support fee payments in different token types (probably ibc-TIA)

## Changes
- implement a `Denom` type, which represents the base (smallest) denomination of an arbitrary asset
- implement an `Id` type, which is a hash of the denomination "trace" (see https://github.com/cosmos/ibc-go/blob/main/docs/architecture/adr-001-coin-source-tracing.md for details) in this PR, since there is no IBC, the `Id` is just the hash of the base denomination.
- update `UnsignedTransaction` to have an optional `fee_asset` field; if this is unset, the native asset is used for fee payment
- update `Transfer` to have an optional `asset` field; if this is unset, the native asset is the asset transferred
- update the sequencer app to have a `NATIVE_ASSET` global
- update the genesis state to have a `native_asset_base_denomination` field. the native asset should be set in the genesis file, and stored in (non-consensus) storage for retrieval upon startups. then the `NATIVE_ASSET` global is set upon startup by reading from storage.

## Testing
unit tests, ran it locally

## Breaking Changelist
- `UnsignedTransaction` and `Transfer` now have an extra (optional) field.

## Related Issues

closes #559


crates/astria-proto/src/native/sequencer/v1alpha1.rs
```diff
@@ -209,6 +213,9 @@ impl SignedTransaction {
 pub struct UnsignedTransaction {
     pub nonce: u32,
     pub actions: Vec<Action>,
+    /// optional asset to use for fee payment.
+    /// if None, then the native asset is used.
+    pub fee_asset: Option<asset::Id>,
```
SuperFluffy, line 218: Noted above, I would prefer if we made this behavior explicit rather than special-casing our "native" assets vs. non-native assets.

SuperFluffy (commented):
Leaving some high level comments on the design.

crates/astria-proto/proto/astria/sequencer/v1alpha1/transaction.proto
```diff
@@ -40,6 +41,7 @@ message Action {
 message TransferAction {
     bytes to = 1;
     astria.primitive.v1.Uint128 amount = 2;
+    optional bytes asset = 3;
```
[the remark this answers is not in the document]
noot: i don't think it's possible to do this - we might have arbitrary IBC assets transferred to us, so we don't know the entire potential limit of what we need to support ahead of time. this might help give more background: https://github.com/cosmos/ibc/blob/main/spec/app/ics-020-fungible-token-transfer/README.md i haven't implemented the ICS20 stuff yet, but it'll be the next step after this PR

and yes we can express all amounts this way, as the `asset` is an `ID` that points to a known base denomination of the asset. i'll rename this `asset_id` to make it clearer

crates/astria-proto/src/native/sequencer/asset.rs
```diff
@@ -0,0 +1,85 @@
...
+}
+
+impl Denom {
+    #[must_use]
+    pub fn from_base_denom(base_denom: &str) -> Self {
+        let id = Id::from_denom(base_denom);
+
+        Self {
+            id,
+            base_denom: base_denom.to_owned(),
```
SuperFluffy, line 23: Prefer `to_string` here (does the same, but is more explicit)

crates/astria-proto/src/native/sequencer/mod.rs
```diff
@@ -1 +1,2 @@
+pub mod asset;
```
SuperFluffy, line 1: Do you expect that these need not be versioned?

crates/astria-proto/src/native/sequencer/asset.rs
SuperFluffy, line 1: Noted in the module line of the parent: do you expect that this will not be versioned? I.e., `v1alpha1` etc

crates/astria-proto/src/native/sequencer/v1alpha1.rs
```diff
@@ -548,6 +578,9 @@ impl SequenceAction {
 pub struct TransferAction {
     pub to: Address,
     pub amount: u128,
+    /// optional asset to be transferred.
+    /// if None, then the native asset is used.
```
SuperFluffy, line 582: Noted above, I'd prefer if we didn't special case the native token but used it explicitly

crates/astria-sequencer/src/asset/mod.rs
```diff
@@ -0,0 +1,16 @@
+use once_cell::sync::OnceCell;
```
SuperFluffy, line 1: `OnceCell` is now actually part of std, so you need not import it.

crates/astria-sequencer/src/asset/mod.rs
```diff
@@ -0,0 +1,16 @@
...
+use proto::native::sequencer::v1alpha1::asset::Denom;
+
+pub(crate) static NATIVE_ASSET: OnceCell<Denom> = OnceCell::new();
```
SuperFluffy, line 4: `once_cell::sync::Lazy` might be a slightly nicer API for this.

crates/astria-sequencer/src/asset/mod.rs
```diff
@@ -0,0 +1,16 @@
...
+
+pub(crate) fn initialize_native_asset(native_asset: &str) {
```
SuperFluffy, line 6: Pair this with a `get_native_asset -> &Denom` function so that you don't have to `.expect()` at every call site.

crates/astria-sequencer/src/authority/action.rs
```diff
@@ -23,6 +24,7 @@ impl ActionHandler for tendermint::validator::Update {
         &self,
         state: &S,
         from: Address,
+        _fee_asset_id: &asset::Id,
```
SuperFluffy, line 27: Oh, so all of these APIs are not special casing native asset. I think that's even more of a reason to not special case this in the protof :)

crates/astria-sequencer/src/state_ext.rs
```diff
@@ -61,6 +61,19 @@ pub(crate) trait StateReadExt: StateRead {
         };
         Ok(u64::from_be_bytes(bytes))
     }
+
+    #[instrument(skip(self))]
+    async fn get_native_asset_denom(&self) -> Result<String> {
+        let Some(bytes) = self
+            .nonverifiable_get_raw(b"nativeasset")
```
SuperFluffy, line 68: Can we make this bytestring a const?

crates/astria-sequencer/src/state_ext.rs
```diff
@@ -84,6 +97,11 @@ pub(crate) trait StateWriteExt: StateWrite {
             version.to_be_bytes().to_vec(),
         );
     }
+
+    #[instrument(skip(self))]
+    fn put_native_asset_denom(&mut self, denom: &str) {
+        self.nonverifiable_put_raw(b"nativeasset".to_vec(), denom.as_bytes().to_vec());
```
SuperFluffy, line 103: Also here (i.e. `const NATIVE_ASSET_KEY` or smth like that).

crates/astria-sequencer/src/transaction/mod.rs
```diff
@@ -49,8 +53,12 @@ pub(crate) async fn check_stateful<S: StateReadExt + 'static>(
     state: &S,
 ) -> anyhow::Result<()> {
     let signer_address = Address::from_verification_key(tx.verification_key());
+    let fee_asset_id = tx
+        .unsigned_transaction()
+        .fee_asset_id
+        .unwrap_or(*NATIVE_ASSET.get().expect("native asset must be set").id());
```
SuperFluffy, line 59: Mentioned this above: this could be made nicer using `Lazy` instead of `OnceCell`.

But even if you stick with once_cell, I recommend using something like `asset::get_native_asset()` or `native_asset:get()` and just have the panic in there.

crates/astria-proto/proto/astria/sequencer/v1alpha1/transaction.proto
```diff
@@ -21,6 +21,7 @@ message SignedTransaction {
 message UnsignedTransaction {
     uint32 nonce = 1;
     repeated Action actions = 2;
+    optional bytes fee_asset_id = 3;
```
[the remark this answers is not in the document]
noot: updated to be non-optional - also added comments to the protos and native descriptions

crates/astria-proto/proto/astria/sequencer/v1alpha1/asset.proto
```diff
@@ -0,0 +1,8 @@
+syntax = "proto3";
+
+package astria.sequencer.v1alpha1;
+
+message Denom {
+    bytes id = 1;
```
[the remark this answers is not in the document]
noot: yes, i loosely based this on penumbra's work: https://github.com/penumbra-zone/penumbra/blob/main/crates/core/asset/src/asset/denom_metadata.rs#L29 

having the ID makes sense as the denomination can potentially be very long, as each time an asset is bridged over IBC its prefix gets longer

crates/astria-proto/src/native/sequencer/asset.rs
```diff
@@ -0,0 +1,85 @@
+use std::{
+    error::Error,
+    fmt::Display,
+};
+
+/// Represents a denomination of a sequencer asset.
+#[derive(Debug, Clone)]
+pub struct Denom {
+    id: Id,
```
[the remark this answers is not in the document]
noot: for our purposes - see comment where i linked the penumbra impl

crates/astria-proto/proto/astria/sequencer/v1alpha1/asset.proto
```diff
@@ -0,0 +1,8 @@
...
+    string base_denom = 2;
```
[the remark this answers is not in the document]
noot: this is the `base_denom`, so the `denom` without any prefixes (ie no channel/port info). i'm planning to make an `IbcDenom` type in the IBC PR also which includes both `base_denom` and `prefix`

crates/astria-proto/src/native/sequencer/mod.rs
noot: not necessarily, i just really don't like the 1000+ line v1alpha1.rs file lol. can we turn that into a module and put the types in separate files?

crates/astria-proto/src/native/sequencer/v1alpha1.rs
noot: removed

crates/astria-sequencer/src/asset/mod.rs
noot: don't think we can use `Lazy` as it requires an initializer function but we need to set this at runtime (ie. either from value in storage, or in `init_chain`)

crates/astria-sequencer/src/transaction/mod.rs
noot: updated!

crates/astria-proto/src/native/sequencer/asset.rs
noot: see above

crates/astria-proto/src/native/sequencer/asset.rs
```diff
@@ -0,0 +1,85 @@
...
+        &self.id
+    }
+
+    #[must_use]
+    pub fn base_denom(&self) -> &str {
+        &self.base_denom
+    }
+}
+
+/// Asset ID, which is the hash of the denomination trace.
```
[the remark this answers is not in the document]
noot: no, it's borrowed somewhat from penumbra's impl

crates/astria-proto/src/native/sequencer/mod.rs
SuperFluffy: Haha, yeah. I think it'd be fine to split up `v1alpha` into modules.

crates/astria-proto/src/native/sequencer/asset.rs
SuperFluffy: Responded there - submodules of `v1alpha1` would be great. I don't know if we will end up with lots of duplicate code at some. Probably yes, but I guess we will cross that bridge then.

crates/astria-proto/src/native/sequencer/asset.rs
```diff
@@ -0,0 +1,94 @@
...
+        let id = Id::from_denom(base_denom);
+
+        Self {
+            id,
+            base_denom: base_denom.to_string(),
+        }
+    }
+
+    #[must_use]
+    pub fn id(&self) -> Id {
```
SuperFluffy, line 37: Can you add little doc strings to this that also explain what these things are (i.e. hash of the base denom, etc)

crates/astria-proto/src/native/sequencer/asset.rs
```diff
@@ -0,0 +1,94 @@
...
+
+    #[must_use]
+    pub fn base_denom(&self) -> &str {
+        &self.base_denom
+    }
+}
+
+/// Asset ID, which is the hash of the denomination trace.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct Id([u8; 32]);
```
SuperFluffy, line 49: Is there an advantage of having this be a self-standing type as opposed to leaving this as an implementation detail fo the `Denom` type?

crates/astria-proto/src/native/sequencer/v1alpha1.rs
```diff
@@ -209,6 +213,8 @@ impl SignedTransaction {
 pub struct UnsignedTransaction {
     pub nonce: u32,
     pub actions: Vec<Action>,
+    /// asset to use for fee payment.
+    pub fee_asset_id: asset::Id,
```
SuperFluffy, line 217: See above, not opposed to types, just wondering if leaving this as a bare `[u8; 32]` would be fine.

crates/astria-sequencer/src/accounts/action.rs
```diff
@@ -26,15 +27,36 @@ impl ActionHandler for TransferAction {
...
-            curr_balance >= self.amount + TRANSFER_FEE,
-            "insufficient funds"
-        );
+            .context("failed getting `from` account balance for fee payment")?;
+
+        // if fee asset is same as transfer asset, ensure accounts has enough funds
+        // to cover both the fee and the amount transferred
+        if fee_asset_id == &transfer_asset_id {
+            ensure!(
+                from_fee_balance >= self.amount + TRANSFER_FEE,
```
SuperFluffy, line 43: Can we use `let Some(from_fee_balance) = self.amount.checked_add(TRANSFER_FEE) else {};` here? These things are *usually* fine, but I'd like us to be extra cautious here because the worst case scenario (however unlikely) is bad.

crates/astria-sequencer/src/accounts/action.rs
```diff
@@ -26,15 +27,36 @@ impl ActionHandler for TransferAction {
...
+            );
+        } else {
+            // otherwise, check the fee asset account has enough to cover the fees,
+            // and the transfer asset account has enough to cover the transfer
+            ensure!(
+                from_fee_balance >= TRANSFER_FEE,
+                "insufficient funds for fee payment"
+            );
+
+            let from_transfer_balance = state.get_account_balance(from, &transfer_asset_id).await?;
```
SuperFluffy, line 54:
```suggestion
            let from_transfer_balance = state.get_account_balance(from, &transfer_asset_id)
                .await
                .context("failed getting account balance while checking that enough funds are available")?;
```

crates/astria-sequencer/src/accounts/action.rs
```diff
@@ -46,21 +68,56 @@ impl ActionHandler for TransferAction {
...
-            .context("failed updating `to` account balance")?;
+
+        // if fee payment asset is same asset as transfer asset, deduct fee
+        // from same balance as asset transferred
+        if &transfer_asset_id == fee_asset_id {
+            state
+                .put_account_balance(
+                    from,
+                    &transfer_asset_id,
+                    from_balance - (self.amount + TRANSFER_FEE),
```
SuperFluffy, line 95: As above, can we use checked math for all of this? I.e. `checked_sub` + `checked_add`.

crates/astria-sequencer/src/transaction/action_handler.rs
```diff
@@ -15,10 +18,16 @@ pub(crate) trait ActionHandler {
         &self,
         _state: &S,
         _from: Address,
+        _fee_asset_id: &asset::Id,
```
SuperFluffy, line 21: Isn't `Id: Copy`? Could just `_fee_asset_id: asset::Id`

crates/astria-sequencer/src/accounts/action.rs
```diff
@@ -46,21 +68,56 @@ impl ActionHandler for TransferAction {
...
+                )
+                .context("failed updating `from` account balance")?;
+            state
+                .put_account_balance(self.to, &transfer_asset_id, to_balance + self.amount)
```
SuperFluffy, line 99: checked math?

crates/astria-sequencer/src/accounts/action.rs
```diff
@@ -46,21 +68,56 @@ impl ActionHandler for TransferAction {
...
+                .context("failed updating `to` account balance")?;
+        } else {
+            // otherwise, just transfer the transfer asset and deduct fee from fee asset balance
+            // later
+            state
+                .put_account_balance(from, &transfer_asset_id, from_balance - self.amount)
```
SuperFluffy, line 105: checked math?

crates/astria-sequencer/src/accounts/action.rs
```diff
@@ -46,21 +68,56 @@ impl ActionHandler for TransferAction {
...
+                .put_account_balance(self.to, &transfer_asset_id, to_balance + self.amount)
+                .context("failed updating `to` account balance")?;
+
+            // deduct fee from fee asset balance
+            let from_fee_balance = state
+                .get_account_balance(from, fee_asset_id)
+                .await
+                .context("failed getting `from` account balance for fee payment")?;
+            state
+                .put_account_balance(from, fee_asset_id, from_fee_balance - TRANSFER_FEE)
```
SuperFluffy, line 117: checked math?

crates/astria-sequencer/src/accounts/action.rs
```diff
@@ -46,21 +68,56 @@ impl ActionHandler for TransferAction {
...
+                .put_account_balance(
+                    from,
+                    &transfer_asset_id,
+                    from_balance - (self.amount + TRANSFER_FEE),
+                )
+                .context("failed updating `from` account balance")?;
+            state
+                .put_account_balance(self.to, &transfer_asset_id, to_balance + self.amount)
+                .context("failed updating `to` account balance")?;
+        } else {
```
SuperFluffy, line 101: If any one of the steps in this `else` block fails, don't we need some kind of rollback mechanism to stay consistent?

crates/astria-proto/src/native/sequencer/asset.rs
```diff
@@ -0,0 +1,94 @@
...
+impl Id {
+    #[must_use]
+    pub fn from_denom(denom: &str) -> Self {
+        use sha2::Digest as _;
+        let hash = sha2::Sha256::digest(denom.as_bytes());
+        Self(hash.into())
+    }
+
+    #[must_use]
+    pub fn as_bytes(&self) -> &[u8; 32] {
```
SuperFluffy, line 60: You could add `AsRef` for this.

crates/astria-sequencer/src/accounts/state_ext.rs
```diff
@@ -32,8 +35,12 @@ fn storage_key(address: &str) -> String {
     format!("{ACCOUNTS_PREFIX}/{address}")
 }
 
-pub(crate) fn balance_storage_key(address: Address) -> String {
-    format!("{}/balance", storage_key(&address.encode_hex::<String>()))
+pub(crate) fn balance_storage_key(address: Address, asset: &asset::Id) -> String {
+    format!(
+        "{}/balance/{}",
+        storage_key(&address.encode_hex::<String>()),
+        asset.as_bytes().encode_hex::<String>()
```
SuperFluffy, line 42: See my comment above: `hex::ToHex` is implemented for all `T: AsRef<[u8]>`, so you could just `impl AsRef<[u8]> for Id` to make this a bit more elegant

crates/astria-sequencer/src/asset/mod.rs
```diff
@@ -0,0 +1,21 @@
+use std::sync::OnceLock;
```
SuperFluffy, line 1: Cool. This makes sequencer need msrv 1.70, but it's taken care of in #582

SuperFluffy (approved):
Looks good. The fact that the native asset is encoded now I like.

Other than some requests for error context, please also have a look at this comment (can't link it; it gives you the file, which is confusing, quoting; need grep to find it):
> If any one of the steps in this else block fails, don't we need some kind of rollback mechanism to stay consistent?


Also, I would very much like to go overboard with checked math for the fee calculations as a safety precaution.

crates/astria-sequencer/src/accounts/action.rs
noot: this is already in a `StateDelta` which it per-transaction, so the state changes will revert if this fails

PR #574, opened 2023-11-09
test(composer): add integrity test of eth tx

### LINKED ISSUE #213: composer: check transaction integrity on integration tests
Right now, the integration test is really minimal and only checks the flow of transactions and not their validity (i.e. if the data can be properly decoded and if the decoded data is as expected).

# Definition of done

The integration test should check for the validity of transactions

itamarreif:
## Summary
This adds a test to verify that that payload of the sequencer transaction submitted by the `Searcher` does not change the rollup transaction that it collected.

## Testing
- Added the `single_rollup_tx_payload_integrity` test that verifies that a transaction with a single payload contains a payload that was serialized properly.

* When we support packing multiple rollup transactions into a single sequencer transaction in the `composer` we will need to change this test to encode/decode a `Vec<Transaction>` instead of a single one.

## Related Issues

closes #213


crates/astria-composer/tests/blackbox/composer.rs
```diff
@@ -1,7 +1,14 @@
 use std::time::Duration;
 
-use ethers::types::Transaction;
+use ethers::{
+    types::Transaction,
+    utils::rlp::{
+        Decodable,
```
SuperFluffy, line 6: Even though this is a test, we try to avoid polluting the namespace:
```suggestion
        Decodable as _,
```

crates/astria-composer/tests/blackbox/composer.rs
```diff
@@ -150,6 +157,48 @@ async fn invalid_nonce_failure_causes_tx_resubmission_under_different_nonce() {
     .expect("valid nonce guard failed");
 }
 
+#[tokio::test]
+async fn test_single_tx_integrity() {
+    let test_composer = spawn_composer(&["test1"]).await;
+
+    // blockhash, blocknumber and transactionindex are generated after execution - we can leave them
+    // out
+    let txs = r#"{
```
SuperFluffy, line 166: let's move that out of the test into a const outside of it so that the test is shorter

crates/astria-composer/tests/blackbox/composer.rs
```diff
@@ -215,10 +264,47 @@ async fn mount_broadcast_tx_sync_invalid_nonce_mock(
...
+        let sequencer_tx = signed_tx_from_request(request);
+        let sequence_action = sequencer_tx
+            .actions()
+            .get(0)
+            .unwrap()
+            .as_sequence()
+            .unwrap();
+        let payload = serde_json::from_slice(&sequence_action.data).unwrap();
+
+        debug!(?payload, ?expected_payload, "comparing payloads");
```
SuperFluffy, line 283: remove

crates/astria-composer/tests/blackbox/composer.rs
```diff
@@ -150,6 +157,48 @@ async fn invalid_nonce_failure_causes_tx_resubmission_under_different_nonce() {
...
+    let mock_guard =
+        mount_broadcast_tx_sync_mock_payload_integrity(&test_composer.sequencer, tx.clone()).await;
+
+    test_composer.rollup_nodes["test1"].push_tx(tx).unwrap();
+    tokio::time::timeout(
+        Duration::from_millis(100),
+        mock_guard.wait_until_satisfied(),
+    )
+    .await
+    .expect("mocked sequencer should have received a broadcast message from composer");
```
SuperFluffy, line 199:
```suggestion
    .expect("mock succeeds with transaction intact");
```

crates/astria-composer/tests/blackbox/composer.rs
```diff
@@ -150,6 +157,48 @@ async fn invalid_nonce_failure_causes_tx_resubmission_under_different_nonce() {
     .expect("valid nonce guard failed");
 }
 
+#[tokio::test]
+async fn test_single_tx_integrity() {
```
SuperFluffy, line 161: don't prefix tests with `test_` as it's redundant. Rather give it a name describing what you want to observe. `async fn composer_forwards_transaction_unchanged` or something like that.

crates/astria-composer/tests/blackbox/composer.rs
```diff
@@ -215,10 +264,47 @@ async fn mount_broadcast_tx_sync_invalid_nonce_mock(
         .await
 }
 
-fn chain_id_nonce_from_request(request: &Request) -> (Vec<u8>, u32) {
+/// Deserizalizes the bytes contained in a `tx_sync::Request` to a signed sequencer transaction and
+/// verifies that it contains a sequence action with `expected_payload` as its contents.
+async fn mount_broadcast_tx_sync_mock_payload_integrity(
```
SuperFluffy, line 269: `mount_matcher_verifying_tx_integrity` might be more descriptive.

SuperFluffy (dismissed):
Accepting the PR, but please change the following things before merging:

1. update the PR title. This PR is no longer about adding a test but changing the serialization format of ether txes. The test for integrity is valuable, but is almost of 2nd order
2. address the error logging
3. don't do ad-hoc eyre errors but use `wrap_err`

crates/astria-composer/src/searcher/mod.rs
```diff
@@ -205,7 +207,16 @@ impl Searcher {
                 // submit signed sequencer txs to sequencer
                 Some(join_result) = self.conversion_tasks.join_next(), if !self.conversion_tasks.is_empty() => {
                     match join_result {
-                        Ok(bundle) => self.handle_bundle_execution(bundle).await,
+                        Ok(res) => {
+                            match res {
+                                Ok(bundle) =>self.handle_bundle_execution(bundle).await,
```
noot, line 212:
```suggestion
                                Ok(bundle) => self.handle_bundle_execution(bundle).await,
```

noot (changes requested):
i am pretty sure this will fail when the tx reaches the EVM rollup - the EVM execution layer is expecting the tx to be rlp-encoded, additionally EVM transactions do not contain a `from` field. (from address is derived from signature)

crates/astria-composer/tests/blackbox/composer.rs
noot, line 166: where is this serialized tx from?

itamarreif: Good question - I took this from Derek's PR.

Do you have a better suggestion for what to use?

crates/astria-composer/src/searcher/mod.rs
```diff
@@ -154,14 +154,16 @@ impl Searcher {
...
+                eyre::eyre!("failed to serialize rollup transaction to JSON: {}", e)
+            })?;
             let chain_id = chain_id.into_bytes();
             let seq_action = Action::Sequence(SequenceAction {
                 chain_id,
                 data,
             });
 
-            vec![seq_action]
+            Ok(vec![seq_action])
```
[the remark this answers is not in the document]
itamarreif: The contents of an `UnsignedTransaction` is a `Vec<Action>`. The allocation has to happen either here or when the transaction is submitted. Doing this here makes more sense than at transaction submission, as the `Executor` should not be doing anything with the transaction's payload - just attaching a nonce and submitting it.

itamarreif:
Undid the changes to rollup transaction encoding that were previously part of this PR and cleaned up the tests a bit. The test now compared the RLP encoding of the expected transaction to the one inside the sequencer transaction that the `composer` submits to verify that the transaction was encoded as expected.

noot (approved):
looks good now 🙏

SuperFluffy (approved):
Looks good. See my comment on where that example payload came from. If you don't know where it came from, can you please find another one and link that?

crates/astria-composer/tests/blackbox/composer.rs
```diff
@@ -242,3 +305,21 @@ fn chain_id_nonce_from_request(request: &Request) -> (Vec<u8>, u32) {
         signed_tx.unsigned_transaction().nonce,
     )
 }
+
+const EXAMPLE_ETH_TX_JSON: &str = r#"{
```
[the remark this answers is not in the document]
itamarreif: I think derek just pulled something off of etherscan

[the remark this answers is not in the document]
itamarreif: Took a random uniswap v2 swap from etherscan and used that instead. The transaction can be looked up by its `hash` field on any block explorer so I dont think there's a need to add an etherscan link here but happy to add it if you think I should.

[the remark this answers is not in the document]
SuperFluffy: But it is useful because it saves me a step to Google to figure it out. Adding the link helps others have the same context that you have while you write it.

PR #577, opened 2023-11-12
fix(conductor, relayer): use sequencer chain id for sequencer blobs

joroshiba:
## Summary
Updates `sequencer-relayer` and `conductor` to use the attached sequencer nodes `chain-id` for  blob namespace.

## Background
We were using a constant for the namespace, which would mean when reading from a production network that sequencer blobs might need to be filtered further based on chain-id, accessing and filtering data without any savings for cost to submit (still would be seperate relayers). This makes it such that our various chains through regenesis and all won't have to do this filtering

## Changes
- Sequencer-relayer sequencer blob namespace derived from sequencer block chain ID on header
- Conductor grabs chain id at startup from a block on rpc sequencer.
- Removed the default namespace

## Testing
CI/CD + manual testing in dev-cluster

## Breaking Changelist
- Posted data blobs to celestia will not match from previous releases.

### DIFF
--- a/crates/astria-celestia-client/src/client.rs
+++ b/crates/astria-celestia-client/src/client.rs
@@ -164,13 +164,12 @@ pub trait CelestiaClientExt: BlobClient {
 
         let mut all_blobs = Vec::with_capacity(num_expected_blobs);
         for (i, block) in blocks.into_iter().enumerate() {
-            let mut blobs =
-                assemble_blobs_from_sequencer_block_data(block).map_err(|source| {
-                    SubmitSequencerBlocksError::AssembleBlobs {
-                        source,
-                        index: i,
-                    }
-                })?;
+            let mut blobs = assemble_blobs_from_sequencer_block_data(block).map_err(|source| {
+                SubmitSequencerBlocksError::AssembleBlobs {
+                    source,
+                    index: i,
+                }
+            })?;
             all_blobs.append(&mut blobs);
         }
 
@@ -258,7 +257,8 @@ fn assemble_blobs_from_sequencer_block_data(
         chain_ids.push(chain_id);
     }
 
-    let sequencer_namespace = celestia_namespace_v0_from_hashed_bytes(header.chain_id.clone().as_bytes());
+    let sequencer_namespace =
+        celestia_namespace_v0_from_hashed_bytes(header.chain_id.clone().as_bytes());
     let sequencer_namespace_data = SequencerNamespaceData {
         block_hash,
         header,
@@ -275,7 +275,8 @@ fn assemble_blobs_from_sequencer_block_data(
     );
 
     blobs.push(
-        Blob::new(sequencer_namespace, data).map_err(BlobAssemblyError::ConstructBlobFromSequencerData)?,
+        Blob::new(sequencer_namespace, data)
+            .map_err(BlobAssemblyError::ConstructBlobFromSequencerData)?,
     );
     Ok(blobs)
 }
--- a/crates/astria-celestia-client/src/client.rs
+++ b/crates/astria-celestia-client/src/client.rs
@@ -151,7 +151,6 @@ pub trait CelestiaClientExt: BlobClient {
     /// This calls the `blob.Submit` celestia-node RPC.
     async fn submit_sequencer_blocks(
         &self,
-        namespace: Namespace,
         blocks: Vec<SequencerBlockData>,
         submit_options: SubmitOptions,
     ) -> Result<u64, SubmitSequencerBlocksError> {
@@ -166,7 +165,7 @@ pub trait CelestiaClientExt: BlobClient {
         let mut all_blobs = Vec::with_capacity(num_expected_blobs);
         for (i, block) in blocks.into_iter().enumerate() {
             let mut blobs =
-                assemble_blobs_from_sequencer_block_data(namespace, block).map_err(|source| {
+                assemble_blobs_from_sequencer_block_data(block).map_err(|source| {
                     SubmitSequencerBlocksError::AssembleBlobs {
                         source,
                         index: i,
@@ -206,7 +205,6 @@ pub enum BlobAssemblyError {
 }
 
 fn assemble_blobs_from_sequencer_block_data(
-    namespace: Namespace,
     block_data: SequencerBlockData,
 ) -> Result<Vec<Blob>, BlobAssemblyError> {
     use sequencer_validation::{
@@ -260,6 +258,7 @@ fn assemble_blobs_from_sequencer_block_data(
         chain_ids.push(chain_id);
     }
 
+    let sequencer_namespace = celestia_namespace_v0_from_hashed_bytes(header.chain_id.clone().as_bytes());
     let sequencer_namespace_data = SequencerNamespaceData {
         block_hash,
         header,
@@ -276,7 +275,7 @@ fn assemble_blobs_from_sequencer_block_data(
     );
 
     blobs.push(
-        Blob::new(namespace, data).map_err(BlobAssemblyError::ConstructBlobFromSequencerData)?,
+        Blob::new(sequencer_namespace, data).map_err(BlobAssemblyError::ConstructBlobFromSequencerData)?,
     );
     Ok(blobs)
 }
--- a/crates/astria-celestia-client/src/client.rs
+++ b/crates/astria-celestia-client/src/client.rs
@@ -146,7 +146,8 @@ pub trait CelestiaClientExt: BlobClient {
     }
 
     /// Submits sequencer `blocks` to celestia after converting and signing them, returning the
-    /// height at which they were included.
+    /// height at which they were included. Sequencer data for each block will be posted to a
+    /// namespace derived from the block's chain ID.
     ///
     /// This calls the `blob.Submit` celestia-node RPC.
     async fn submit_sequencer_blocks(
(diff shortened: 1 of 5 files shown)


crates/astria-celestia-client/src/client.rs
```diff
@@ -260,6 +257,8 @@ fn assemble_blobs_from_sequencer_block_data(
         chain_ids.push(chain_id);
     }
 
+    let sequencer_namespace =
+        celestia_namespace_v0_from_hashed_bytes(header.chain_id.clone().as_bytes());
```
noot, line 261: should this be Namespace::const_v0 - i think from_hashed_bytes are for if the input is already hashed?

noot (approved):

crates/astria-conductor/src/conductor.rs
```diff
@@ -152,13 +155,31 @@ impl Conductor {
             let (shutdown_tx, shutdown_rx) = oneshot::channel();
             shutdown_channels.insert(Self::DATA_AVAILABILITY, shutdown_tx);
             let block_verifier = BlockVerifier::new(sequencer_client_pool.clone());
+            let sequencer_namespace = {
+                use sequencer_client::SequencerClientExt as _;
+
+                let client = sequencer_client_pool.get().await?;
```
SuperFluffy, line 161: Please add error context here.

crates/astria-conductor/src/conductor.rs
```diff
@@ -152,13 +155,31 @@ impl Conductor {
...
+                let chain_id = client
+                    .latest_sequencer_block()
+                    .await?
```
SuperFluffy, line 164: Please add error context here.

crates/astria-conductor/src/conductor.rs
```diff
@@ -152,13 +155,31 @@ impl Conductor {
...
+                    .header()
+                    .chain_id
+                    .clone();
```
SuperFluffy, line 167: No need to clone. This field access is moving out of the `Header` struct.

crates/astria-conductor/src/conductor.rs
```diff
@@ -152,13 +155,31 @@ impl Conductor {
             let (shutdown_tx, shutdown_rx) = oneshot::channel();
             shutdown_channels.insert(Self::DATA_AVAILABILITY, shutdown_tx);
             let block_verifier = BlockVerifier::new(sequencer_client_pool.clone());
+            let sequencer_namespace = {
```
SuperFluffy, line 158: I believe this is problematic if `execution_commit_level == FirmOnly` (or whatever the variant is). You might not have a sequencer on the other side, but this code would still run.

The only way to fix this IMO is to not couple conductor-bringup with sequencer and instead set it through conductor's config file.

Then instead when firing up the sequencer reader you can hit the endpoint and verify that the configured sequencer namespace matches the remote one.

But that only works in cases where `! firm_only`

crates/astria-conductor/src/conductor.rs
```diff
@@ -152,13 +155,31 @@ impl Conductor {
...
+                use sequencer_client::SequencerClientExt as _;
+
+                let client = sequencer_client_pool.get().await?;
+                let chain_id = client
+                    .latest_sequencer_block()
```
SuperFluffy, line 163: This should have a retry IMO.

Note that this comment is obsolete if you follow my comment re sequencer being absent in firm-only setups.

crates/astria-conductor/src/data_availability.rs
```diff
@@ -116,22 +117,27 @@ pub(crate) struct Reader {
     shutdown: oneshot::Receiver<()>,
 }
 
+pub(crate) struct CelestiaReaderConfig {
```
SuperFluffy, line 120: I guess this is to work around clippy's "function has too many arguments" lint?

crates/astria-celestia-client/src/client.rs
```diff
@@ -151,7 +151,6 @@ pub trait CelestiaClientExt: BlobClient {
     /// This calls the `blob.Submit` celestia-node RPC.
     async fn submit_sequencer_blocks(
         &self,
-        namespace: Namespace,
```
SuperFluffy, line 154: I am not sure I like that the namespace is becoming an implicit function of the `blocks` sequence.

While I follow the thought process (coupling cometbft chain ID to namespace), it's now difficult to understand where the sequencer blocks will end up.

I prefer to make the namespace explicit and not couple so tightly to cometbft

crates/astria-sequencer-relayer/src/relayer.rs
```diff
@@ -479,9 +478,9 @@ async fn submit_blocks_to_celestia(
         num_blocks = sequencer_block_data.len(),
         "submitting collected sequencer blocks to data availability layer",
     );
+
     let height = client
         .submit_sequencer_blocks(
-            SEQUENCER_NAMESPACE,
```
SuperFluffy, line 484: This made me pause and think about the changes in the client. In principle this enables a relayer to relay blocks from multiple sequencers. This shouldn't happen because a sequencer should not willy-nilly change its chain ID, but I am not sure I am happy with making this a runtime feature.

SuperFluffy (changes requested):
Requesting changes because I see 2 open questions:

1. ~~I think the requirement that sequencer be present has issues if conductor is set to process firm-only blocks~~
2. I believe the sequencer namespace that relayer submits should be explicit and not derived from the cometbft chain IDs.

The latter is a pretty drastic change to the celestia-client abstraction that I'd like to discuss further before doing it.

joroshiba:
To the first point: nope we need to access the sequencer to verify blocks from DA already. 

To the second, this was the discussion on slack. The chain ID is fixed for any given blockchain. If the chain changes it's a different blockchain. So this ensures that dusk-1 has data stored in a different namespace than dusk-2. The chain ID is a part of the genesis that defines the chain so it's programmatic, but should never change for the same sequencing chain.

crates/astria-conductor/src/conductor.rs
SuperFluffy: Resolved: even in `firm-only` conductor needs access to sequencer.

SuperFluffy:
Resolved my requests offline:

1. conductor always needs access to sequencer to verify blocks, so it needs to derive the namespace from sequencer no matter what
2. we do want to strongly couple cometbft chain-IDs to celestia namespaces. In that case the extra argument to `submit_sequencer_blocks` is redundant as the namespace is directly determined from a field in each block.

For 1., please report the chain ID and eventual namespace that conductor will filter on. It seems like celestia likes base64 for this (according to their serialization logic here: https://github.com/eigerco/celestia-node-rs/blob/ee630f2b5303a5d930ea4228c38ecf12b7b5db59/types/src/nmt.rs#L144C6-L144C6). You can use `base64::display` to avoid extra allocations from here (we already have a workspace-level import):
https://docs.rs/base64/latest/base64/display/index.html

For 2 please add a doc comment on `submit_sequencer_blocks` explaining to users where the namespace is derived from.

crates/astria-celestia-client/src/client.rs
joroshiba: Discussed seperately offline. We want this tie, writing it somehwere different than specified by the block would be wrong. Adding logging + comments to address concerns over clarity.

crates/astria-celestia-client/src/client.rs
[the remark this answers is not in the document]
joroshiba: 1. fixed
2. to clarify, pointing out the problem that sha256 solves for us here right? 
3. Yes we do want to couple these

sambukowski (approved):

crates/astria-celestia-client/src/client.rs
```diff
@@ -146,12 +146,12 @@ pub trait CelestiaClientExt: BlobClient {
     }
 
     /// Submits sequencer `blocks` to celestia after converting and signing them, returning the
-    /// height at which they were included.
+    /// height at which they were included. Sequencer data for each block will be posted to a
```
SuperFluffy, line 149: For doc comments try to have a short summary on the first line followed by a space, and then more text like:
```rust
/// Summary of what the function is supposed to do
///
/// More information about the function, interesting background stories,
/// where the function has been, early life, career progression, family related
/// trivia.
```
(the original title was already too long actually)

crates/astria-conductor/src/conductor.rs
```diff
@@ -152,13 +159,42 @@ impl Conductor {
...
+            // Sequencer namespace is defined by the chain id of attached sequencer node
+            // which can be fetched from any block header.
+            let sequencer_namespace = {
+                use sequencer_client::SequencerClientExt as _;
+
+                let client = sequencer_client_pool
+                    .get()
+                    .await
+                    .wrap_err("failed to get a sequencer client from the pool")?;
+                let block = tryhard::retry_fn(|| client.latest_sequencer_block())
```
SuperFluffy, line 172: This has 10 retries but no backoff strategy. Is that intentional? This basically turns this into 10 requests that get fired off immediately one after the other.

crates/astria-conductor/src/data_availability.rs
```diff
@@ -100,6 +99,8 @@ pub(crate) struct Reader {
 
     block_verifier: BlockVerifier,
 
+    /// Sequencer Namespace ID
+    sequencer_namespace: Namespace,
     /// Namespace ID
     namespace: Namespace,
```
SuperFluffy, line 105: We should probably call this `rollup_namespace` now.

SuperFluffy (approved):
Looks good, thank you for addressing this.

crates/astria-conductor/src/conductor.rs
joroshiba: Changed to exponential.

PR #579, opened 2023-11-14
feat(sequencer): support `IbcAction`s and implement ICS20 incoming transfer application logic

noot:
## Summary
Adds IBC support for ICS20 token transfers

## Background
We want IBC support, specifically for token transfers (eg. using TIA as a gas token).

See the cosmos ICS20 spec: https://github.com/cosmos/ibc/blob/fe150abb629de5c1a598e8c7896a7568f2083681/spec/app/ics-020-fungible-token-transfer/README.md

## Changes
- bump penumbra version and related deps to include IBC refactor for generic `AppHandler`s + substore support
- implement an `IbcAsset` type which represents an IBC-transferred-token
- update `Action` proto and native enums to include `IbcAction` variants, `IbcAction` is a type defined in `penumbra_ibc`
- implement `sequencer::accounts::Ics20Transfer` which is the app handler for `IbcAction`s, which is called into during `IbcAction::check_stateless` and  `IbcAction::execute`
- `Ics20Transfer` contains the actual ICS20 transfer logic; namely, handling inbound IBC packets that are transfers. there are two main cases we handle for transfer logic: a token being received on Astria from another chain, or a notification of an outbound transfer failing, in which case we need to refund the escrowed tokens.
- in both cases, we determine if Astria is the source of token we need to issue, and handle the packet accordingly (either minting the synthetic, bridged asset if we are not the source, or sending native Astria tokens from the escrow account to the recipient if we are the source)

## Testing
can't really test the transfer functionality without heavier integration testing at the moment, due to the fact that it requires merkle proofs of state of the counterparty chain. in discussions with the penumbra team about a "mock tendermint" test harness we could use to test this in a unit-test-like environment.

## Breaking Changelist
- none

## Related Issues

closes #79

PR #580, opened 2023-11-14
chore: bump rust version in ci to 1.73.0

noot:
## Summary
bump rust version in ci to 1.73.0

## Background
required for #579; had to bump penumbra deps, which bumped libp2p, which requires rust >=1.73.0

## Changes
- bump rust version in ci to 1.73.0

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,7 +19,7 @@ jobs:
     if: needs.run_checker.outputs.run_tests == 'true'
     steps:
       - uses: actions/checkout@v3
-      - uses: dtolnay/rust-toolchain@1.72.0
+      - uses: dtolnay/rust-toolchain@1.73.0
       - uses: taiki-e/install-action@v2.15.2
         with:
           tool: cargo-hack@0.5.29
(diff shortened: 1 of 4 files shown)


joroshiba:
We should update the docker images as well probably

sambukowski (approved):

SuperFluffy:
This change and the bump to prost/prost-build 0.12 in #579 is currently blocked on protoc-gen-prost still being on prost 0.11.

SuperFluffy (changes requested):
Requesting changes for 2 issues:

I would like to separate wider dependency bumps and feature PRs. This bump to 1.73 is explained by another PR, but ideally the reason for bumping should be tied to that PR.

But either way: I did prepare a PR bumping penumbra (and celestia) to their latest versions, which would allow us to use prost/prost-build 0.12 and tonic 0.10 (as does #579), but that is currently blocked on protoc-gen-prost still being on prost 0.11 (which we are using through buf).

This is an issue.

Other than forking protoc-gen-prost the only way to fix this is to move off `buf` temporarily. I am not sure how feasible it is to fork `protoc-gen-prost` because it would require us to somehow put the fork on buf and make it available to buf.

noot:
@SuperFluffy okay, should i merge this PR into my IBC branch for now then? i figured the rust version bump would be separate enough for its own PR but either way is fine with me!

SuperFluffy:
Closing this in favor of #581 (move away from buf) and #582 (bump all dependencies)

@noot these two should unblock #579

PR #581, opened 2023-11-15
build(proto): replace buf-generate by tonic_build

SuperFluffy:
## Summary
Replaces `buf generate` by `tonic_build` to generate rust code from protobuf.

## Background
`buf generate` relies on the third party plugins `protoc-gen-prost` and `protoc-gen-tonic`, which are not yet updated to prost-build 0.12 and tonic-build 0.10 (and hence prost 0.12, tonic 0.10).

This patch moves code generation away from the `buf generate` executable to the Rust `tonic-build` crate, because we need to update prost/tonic and because these projects do not guarantee that code generated with older versions work with newer versions.

## Changes
- Replace invoking the `buf generate` executable by using `tonic_build` directly
- Use `buf build` to create a file descriptor set that can be fed into `tonic_build` (the latter skips its own invocation of `protoc`)
- Clean the generated code by removing all files that are not prefixed with `astria.` (tonic also creates client and server code for tendermint abci; I am not sure why, probably due to the descriptor set)

## Testing
Not tested directly; this should not affect the actual code generated. Everything should still compile, all tests should pass.

## Related Issues
PR requiring a newer version of penumbra and hence prost: https://github.com/astriaorg/astria/pull/579

PR againast protoc-gen-prost: https://github.com/neoeinstein/protoc-gen-prost/pull/78


crates/astria-proto/src/proto/generated/astria.execution.v1alpha1.tonic.rs
SuperFluffy, line 1: The `*.tonic.rs` files have been absorbed as modules directly into their parent files (these were `include!`ed previously I believe)

noot (approved):
looks good! so are we still using buf for handling dependencies, but tonic for the actual build?

SuperFluffy:
> looks good! so are we still using buf for handling dependencies, but tonic for the actual build?

Yeah, so `buf` fetches the the tendermint protos as defined in `buf.yaml`. and then compiles our protos + the tendermint protos it just pulled, emitting a file descriptor set. We then feed this set into tonic-build.

My understanding is that tonic-build would usually invoke protoc, which would also result in that file descriptor set. But we `.skip_protoc()` because we want `buf` to deal with proto-dependecy resolution.

PR #583, opened 2023-11-16
refactor(conductor): replace executor constructor by type state builder

SuperFluffy:
Replace `Executor::new()` by `Executor::builder()....build()` because the constructor was getting too lengthy.

This uses the type-state pattern to avoid having to check if all required builder fields are set (with the fields all being `Option<T>`). This is more verbose in the executor's type signature and field-setters, but avoids the need for error handling.

### DIFF
--- a/crates/astria-conductor/src/executor/mod.rs
+++ b/crates/astria-conductor/src/executor/mod.rs
@@ -113,9 +113,204 @@ impl From<SequencerBlockData> for ExecutorCommand {
     }
 }
 
+pub(crate) struct NoRollupAddress;
+pub(crate) struct WithRollupAddress(String);
+pub(crate) struct NoChainId;
+pub(crate) struct WithChainId(ChainId);
+pub(crate) struct NoBlockChannel;
+pub(crate) struct WithBlockChannel(Receiver);
+pub(crate) struct NoShutdown;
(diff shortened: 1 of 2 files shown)


crates/astria-conductor/src/executor/mod.rs
```diff
@@ -113,9 +113,204 @@ impl From<SequencerBlockData> for ExecutorCommand {
...
+
+    pub(crate) fn sequencer_height_with_first_rollup_block(
+        mut self,
+        sequencer_height_with_first_rollup_block: u32,
+    ) -> Self {
+        self.sequencer_height_with_first_rollup_block = sequencer_height_with_first_rollup_block;
+        self
+    }
+
+    pub(crate) fn set_optimism_hook(mut self, handler: Option<optimism::Handler>) -> Self {
```
SuperFluffy, line 261: Note that this follows the ecosystem using `set_*` for setters that replace builder pattern fields by a full type.

I would provide `fn optimism_hook(self, handler: Handler)` which would then `self.optimism_hook.replace(handler);`, but since this is not a public interface it would get flagged as never used.

noot (approved):

noot:
nice, thanks!

PR #585, opened 2023-11-17
fix(cli): block height should be string in yaml to avoid scientific notation

steezeburger:
## Summary
I think helm install in hag might be failing due to the yaml having scientific notation for the block height.
This PR serializes the value as a string.

UPDATE: hag wasn't failing due to this, BUT conductor did fail to start because the block height was parsed as a float because of the scientific notation.

Tested locally and the conductor pod started successfully.


crates/astria-cli/src/types.rs
```diff
@@ -73,11 +73,11 @@ impl RollupDeploymentConfig {
     }
 
     #[must_use]
-    pub fn get_initial_sequencer_height(&self) -> u64 {
-        self.sequencer.initial_block_height
+    pub fn get_initial_sequencer_height(&self) -> String {
```
SuperFluffy, line 76: Could to `-> &str`

crates/astria-cli/src/types.rs
```diff
@@ -73,11 +73,11 @@ impl RollupDeploymentConfig {
...
+        self.sequencer.initial_block_height.to_string()
     }
 
-    pub fn set_initial_sequencer_height(&mut self, new_height: u64) {
+    pub fn set_initial_sequencer_height(&mut self, new_height: String) {
```
SuperFluffy, line 80: You could make this `new_height: impl Display` and then `self.sequencer.initial_block_height = new_height.to_string();`

SuperFluffy (approved):

noot (approved):

crates/astria-cli/src/types.rs
steezeburger: This getter/setter pair is never used. Going to remove

PR #586, opened 2023-11-17
fix(relayer): update wait_for_sequencer log to be correct

joroshiba:
## Summary
Fix a log which was incorrect

## Background
We wait for both DA and sequencer, and this log was incorrect.


SuperFluffy (approved):

SuperFluffy (approved):
Approve, but see the `instrument` attribute of `wait_for_sequencer` - its `name` is wrong. Can you also fix before merging?

I think you can just remove the `name = ` thing entirely for both `wait_for_sequencer` and `wait_for_data_availability`. It will take the function names. The context comes from the span within which they run.

noot (approved):

PR #589, opened 2023-11-17
fix(conductor): remove proposer check; use correct validator set in commit verification

noot:
## Summary
remove proposer check from block verifier as it's not needed, also the logic was wrong anyways. the verification of the `Commit` is sufficient to ensure the block was actually committed by the sequencer. 

also update to use correct validator set in commit verification; we need to use the height at `n-1` for a block at height `n`. see code comment also. this makes sense as we have a validator set at genesis, which votes on block 1; then, the set at block 1 votes on block 2, etc.

## Background
see above

## Changes
- remove proposer check from block verifier
- use correct validator set in commit verification

## Testing
these issues were found in devnet testing


sambukowski (approved):

noot:
@SuperFluffy they don't affect the same logic, and this might need to get in first

itamarreif (approved):

joroshiba (approved):

crates/astria-conductor/src/block_verifier.rs
```diff
@@ -70,13 +69,17 @@ impl BlockVerifier {
...
         );
 
-        let current_validator_set = client
-            .validators(height, sequencer_client::tendermint_rpc::Paging::Default)
+        // the validator set which votes on a block at height `n` is the
+        // set at `n-1`. the validator set at height `n` is the set after
+        // the results of block `n`'s execution have occurred.
+        let validator_set = client
+            .validators(
+                height - 1,
```
SuperFluffy, line 77: Please handle the special case `height = 0` by using `checked_sub`.
```rust
let Some(previous_height) = height.checked_sub(1) else {
    return Ok(()); // Ok probably? Not sure; is there another way to validate height 0?
```

crates/astria-conductor/src/block_verifier.rs
```diff
@@ -70,13 +69,17 @@ impl BlockVerifier {
             hex::encode(data.block_hash),
         );
 
-        let current_validator_set = client
-            .validators(height, sequencer_client::tendermint_rpc::Paging::Default)
+        // the validator set which votes on a block at height `n` is the
+        // set at `n-1`. the validator set at height `n` is the set after
+        // the results of block `n`'s execution have occurred.
+        let validator_set = client
```
SuperFluffy, line 75: Could add an info log that we are fetching the validator set for the prev height.

SuperFluffy (approved):
Looks good, only nits and over-cautious requests for wrapping subs.

Issue #509, opened 2023-10-17
provide a mechanism to set celestia fees and gas


SuperFluffy:
We are using a constant celestia fee of `100_000` when submitting blobs. It's not clear what unit this is, nor how this value is obtained.

We should allow making this configurable in a first step, but better provide a mechanism to dynamically set it.

joroshiba:
#590 migrated to using celestia-node fee estimation instead of fixed pricing.

PR #590, opened 2023-11-17
fix(relayer): don't use fixed fee pricing

joroshiba:
## Summary
Updates the fee payments to use default pricing.

## Background
Where were spending .1 TIA for each blob, and it can/should be much cheaper.

## Changes
- removes the gas limit config for relayer
- relies on market pricing from node to write data to celestia

## Breaking Changes
Removed a config entry

## Testing
CI + Deployed against mocha network to test pricing


crates/astria-sequencer-relayer/src/relayer.rs
```diff
@@ -117,8 +117,8 @@ impl Relayer {
             sequencer_poll_period: Duration::from_millis(cfg.block_time),
             data_availability,
             // FIXME (https://github.com/astriaorg/astria/issues/509): allow configuring this
```
noot, line 119: should we address #509 at some point? how does the default pricing get determined (some oracle in the celestia node?)

noot (approved):

crates/astria-sequencer-relayer/src/relayer.rs
```diff
@@ -117,8 +117,8 @@ impl Relayer {
...
-            fee: 100_000,
-            gas_limit: cfg.gas_limit,
```
noot, line 121: remove `cfg.gas_limit` as it's unused now i believe

crates/astria-sequencer-relayer/src/relayer.rs
joroshiba: Celestia node contains code to estimate the gas fee programatically: https://docs.celestia.org/developers/submit-data#estimating-gas-programmatically

At some point it would be good to have an option here to set manually, or use default.

crates/astria-sequencer-relayer/src/relayer.rs
```diff
@@ -35,10 +35,10 @@ pub(crate) struct Relayer {
     data_availability: celestia_client::jsonrpsee::http_client::HttpClient,
 
     // The fee that relayer will pay for submitting sequencer blocks to the DA.
-    fee: u64,
+    fee: Option<u64>,
```
SuperFluffy, line 38: IMO jut remove these fields entirely. They are now dead code.

crates/astria-sequencer-relayer/src/relayer.rs
```diff
@@ -35,10 +35,10 @@ pub(crate) struct Relayer {
...
 
     // The limit that relayer will pay for submitting sequencer blocks to the DA.
-    gas_limit: u64,
+    gas_limit: Option<u64>,
```
SuperFluffy, line 41: Same, remove

crates/astria-sequencer-relayer/src/relayer.rs
```diff
@@ -117,8 +117,8 @@ impl Relayer {
             sequencer_poll_period: Duration::from_millis(cfg.block_time),
             data_availability,
             // FIXME (https://github.com/astriaorg/astria/issues/509): allow configuring this
-            fee: 100_000,
-            gas_limit: cfg.gas_limit,
+            fee: None,
+            gas_limit: None,
```
SuperFluffy, line 121: rm both

crates/astria-sequencer-relayer/src/relayer.rs
```diff
@@ -251,6 +251,10 @@ impl Relayer {
         // Then report update the internal state or report if submission failed
         match submission_result {
             Ok(height) => self.state_tx.send_modify(|state| {
+                debug!(
+                    da_height=%height,
```
SuperFluffy, line 255: Not a fan of terse acronyms like this. I think just `%height` is descriptive enough given the context. Or, if necessary, `data_availability_height` or better `celestia_height` (since its the concrete height of the celestia network, not of some abstract DA network).

crates/astria-sequencer-relayer/src/relayer.rs
```diff
@@ -465,8 +469,8 @@ fn convert_block_response_to_sequencer_block_data(
 #[instrument(skip_all)]
 async fn submit_blocks_to_celestia(
     client: celestia_client::jsonrpsee::http_client::HttpClient,
-    fee: u64,
-    gas_limit: u64,
+    fee: Option<u64>,
+    gas_limit: Option<u64>,
```
SuperFluffy, line 473: dead code. Better rm right now

SuperFluffy (approved):
Looks good. Just have a strong preference for removing dead code and not use the terse `da` acronyms

PR #591, opened 2023-11-18
chore(conductor, relayer, sequencer): release 0.10.0 conductor, 0.8.0 relayer, and 0.6.0 sequencer

joroshiba:
Release cuts, major releases due to fundamental change to the merkle tree shape.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -483,7 +483,7 @@ dependencies = [
 
 [[package]]
 name = "astria-sequencer"
-version = "0.5.0"
+version = "0.6.0"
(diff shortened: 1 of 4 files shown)


SuperFluffy (approved):
Does this warrant a release in composer? I think not, because it doesn't care about any of the APIs or datastructures that have been changed

noot (approved):

joroshiba:
> Does this warrant a release in composer? I think not, because it doesn't care about any of the APIs or datastructures that have been changed

Yeah we shouldn't need a composer release

PR #595, opened 2023-11-21
fix(sequencer): fix instrument logging not to log every tx

noot:
## Summary
logging every single tx before executing each of them is bad for logging.

## Changes
- don't log the entire block's txs

### DIFF
--- a/crates/astria-sequencer/src/app.rs
+++ b/crates/astria-sequencer/src/app.rs
@@ -158,7 +158,7 @@ impl App {
     /// It puts this special "commitment" as the first transaction in a block.
     /// When other validators receive the block, they know the first transaction is
     /// supposed to be the commitment, and verifies that is it correct.
-    #[instrument(name = "App::prepare_proposal", skip(self))]
+    #[instrument(name = "App::prepare_proposal", skip(self, prepare_proposal))]
     pub(crate) async fn prepare_proposal(
         &mut self,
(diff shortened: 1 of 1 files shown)


jbowen93 (approved):
LGTM

crates/astria-sequencer/src/app.rs
```diff
@@ -182,7 +182,7 @@ impl App {
     /// Generates a commitment to the `sequence::Actions` in the block's transactions
     /// and ensures it matches the commitment created by the proposer, which
     /// should be the first transaction in the block.
-    #[instrument(name = "App::process_proposal", skip(self))]
+    #[instrument(name = "App::process_proposal", skip(self, process_proposal))]
```
SuperFluffy, line 185: Do we need to skip all of it or is there some info that's useful?

Also, if you want to skip all args, use `skip_all`

SuperFluffy (approved):
See my comment re useful information. Otherwise this seems very reasonable

SuperFluffy:
Note also that the main issue with this is not the proposal being logged, but the entire thing being attached to every single event that happens under this scope.

jbowen93:
I'm going to merge this as is and cut a new minor release

crates/astria-sequencer/src/app.rs
noot: i just skipped it all for ease - i guess the hash could be good to log

PR #601, opened 2023-11-22
fix(sequencer): cap tx size at 250kb

noot:
## Summary
we saw that on devnet, submitting a tx that exceeds the `max_tx_bytes` per block (~22mb) will result in it hanging around the mempool forever, as it cannot end up in a block, and causes all txs to remain pending. this is bad, and we need to ensure txs that are too large don't end up in the mempool to begin with.

## Background
the size chosen (250kb) was based on @jbowen93 's comment:
> Celestia has 2MB blocks and I believe we batch up to 5 Astria blocks to Celestia. So if we wanted to be conservative like 250KB, but we could go as high as 400KB.

we can change this later if needed, but this is a good starting point.

## Changes
- add a check in `CheckTx` that rejects the tx if it's larger than 250kb

## Testing
devnet (after)

## Breaking Changelist
- txs >250kb will be rejected

### DIFF
--- a/crates/astria-sequencer/src/service/mempool.rs
+++ b/crates/astria-sequencer/src/service/mempool.rs
@@ -86,8 +86,12 @@ async fn handle_check_tx<S: StateReadExt + 'static>(
     } = req;
     if tx.len() > MAX_TX_SIZE {
         return response::CheckTx {
-            code: AbciCode::INVALID_PARAMETER.into(),
-            log: format!("transaction size exceeds maximum allowed size of {MAX_TX_SIZE} bytes"),
+            code: AbciCode::INVALID_SIZE.into(),
+            log: format!(
+                "transaction size too large; received: {}, allowed: {MAX_TX_SIZE}",
+                tx.len()
+            ),
+            info: AbciCode::INVALID_SIZE.to_string(),
             ..response::CheckTx::default()
         };
     }
--- a/crates/astria-sequencer/src/service/mempool.rs
+++ b/crates/astria-sequencer/src/service/mempool.rs
@@ -87,7 +87,7 @@ async fn handle_check_tx<S: StateReadExt + 'static>(
     if tx.len() > MAX_TX_SIZE {
         return response::CheckTx {
             code: AbciCode::INVALID_PARAMETER.into(),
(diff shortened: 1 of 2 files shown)


jbowen93:
LGTM

crates/astria-sequencer/src/service/mempool.rs
```diff
@@ -82,6 +84,14 @@ async fn handle_check_tx<S: StateReadExt + 'static>(
     let request::CheckTx {
         tx, ..
     } = req;
+    if tx.len() > MAX_TX_SIZE {
+        return response::CheckTx {
+            code: AbciCode::INVALID_PARAMETER.into(),
```
SuperFluffy, line 89: Can we add a new AbciCode for this? Will make it easier to understand what went wrong in logs without hving to parse the log

crates/astria-sequencer/src/service/mempool.rs
```diff
@@ -82,6 +84,14 @@ async fn handle_check_tx<S: StateReadExt + 'static>(
     let request::CheckTx {
         tx, ..
     } = req;
+    if tx.len() > MAX_TX_SIZE {
+        return response::CheckTx {
```
SuperFluffy, line 88: This should contain the `info` line, which would be the `Display` impl of `AbciCode::TX_TOO_LARGE` (see comment below adding more abci codes). I think the Display line should be more akin to what's currently in log (see the comment in `log` for more)

crates/astria-sequencer/src/service/mempool.rs
```diff
@@ -82,6 +84,14 @@ async fn handle_check_tx<S: StateReadExt + 'static>(
...
+            code: AbciCode::INVALID_PARAMETER.into(),
+            log: format!("transaction size exceeds maximum allowed size of {MAX_TX_SIZE} bytes"),
```
SuperFluffy, line 90:
```suggestion
            log: format!("transaction size too large; received: {}, allowed: {MAX_TX_SIZE}", tx.len()),
```
(or smth like that)

SuperFluffy (approved):
Looks good. We should extend this to composer and think about moving `MAX_TX_SIZE` to a public place, but we cross that bridge then.

Please have a look at my comments re creating a new `AbciCode` and the `info` and `log` fields.

PR #602, opened 2023-11-22
refactor(proto): split v1alpha1.rs file into multiple files

noot:
## Summary
v1alpha1.rs was >1000 lines before, this PR splits in into multiple files each with their own types. this also ensures `asset` is versioned.

## Changes
- split v1alpha1.rs file into multiple files

## Testing
n/a


crates/astria-proto/src/native/sequencer/v1alpha1.rs
SuperFluffy, line 1: Please `git mv v1alpha.rs v1alpha1/mod.rs`

crates/astria-proto/src/native/sequencer/v1alpha1/mint_action.rs
SuperFluffy, line 1: I think these are too granular.

I think the actions types and enum can all be inside `action.rs`

SuperFluffy (commented):
Quick high level feedback: I think these changes are a bit too granular and we now have one file per type. Agree on the plan to split up `v1alpha` overall, but IMO this is a bit extreme.

I think the *actions should be grouped together, as should the signed/unsigned transactions.

SuperFluffy:
@noot We have a long standing plan to move the proto definitions to the top level of the repository and out of `astria-proto/`. Part of that move would also be to redefine/rename this `astria-proto` as `astria-api` or `astria-messaging` or some better name.

I would like to clean up this file in one go.

noot:
@SuperFluffy ah okay, just saw your comment, I already updated this PR based on the recent updates to main. let me know what you'd like to do

PR #609, opened 2023-11-27
feat(sequencer): implement ICS20 withdrawals

### LINKED ISSUE #597: sequencer + IBC: implement `ICS20Withdrawal` action type
#579 implements receiving ICS20 packets, but we need to handle "outgoing" packets as well, which are ICS20 withdrawal actions a user can submit which attempts to send their tokens from Astria to another chain.

noot:
## Summary
Counterpart to #579 which supports incoming ICS20 transfers, this PR implements outgoing ICS20 transfers.

## Background
required for IBC token transfer support.

See https://github.com/cosmos/ibc/blob/fe150abb629de5c1a598e8c7896a7568f2083681/spec/app/ics-020-fungible-token-transfer/README.md `sendFungibleTokens` for spec.

## Changes
- add `Ics20Withdrawal` proto and native type which is an action (maybe rename to `Ics20WithdrawalAction`?)
- implement check logic (apart from IBC packet check, check the sender has enough balance)
- implement execution logic (escrow tokens if native, or burn if not native)

## Testing
not really - same issue as #579 

## Related Issues

closes #597


crates/astria-sequencer/src/accounts/ics20_withdrawal.rs
```diff
@@ -0,0 +1,132 @@
...
+                .context("failed to update channel balance")?;
+        }
+
+        state.send_packet_execute(checked_packet).await;
+        Ok(())
+    }
+}
+
+fn is_source(source_port: &PortId, source_channel: &ChannelId, asset: &IbcAsset) -> bool {
+    let prefix = format!("{source_port}/{source_channel}/");
```
SuperFluffy, line 130: you could make this check without allocation by making repeated use of `str::strip_prefix` and the fact that `PortId` and `ChannelId` both implement `AsRef<str>`

crates/astria-sequencer/src/accounts/ics20_withdrawal.rs
```diff
@@ -0,0 +1,132 @@
...
+            .get_account_balance(from, self.denom.id())
+            .await
+            .context("failed getting `from` account balance for transfer")?;
+        ensure!(
+            from_transfer_balance > self.amount,
+            "insufficient funds for transfer"
+        );
+
+        state
+            .put_account_balance(from, self.denom.id(), from_transfer_balance - self.amount)
```
SuperFluffy, line 100: Please use checked math here

crates/astria-sequencer/src/accounts/ics20_withdrawal.rs
```diff
@@ -0,0 +1,132 @@
...
+            let channel_balance = state
+                .get_ibc_channel_balance(&self.source_channel, self.denom.id())
+                .await
+                .context("failed to get channel balance")?;
+
+            state
+                .put_ibc_channel_balance(
+                    &self.source_channel,
+                    self.denom.id(),
+                    channel_balance + self.amount,
```
SuperFluffy, line 119: checked math

crates/astria-sequencer/src/accounts/ics20_withdrawal.rs
```diff
@@ -0,0 +1,132 @@
...
+            checked_packet.source_channel(),
+            &self.denom,
+        ) {
+            let channel_balance = state
+                .get_ibc_channel_balance(&self.source_channel, self.denom.id())
+                .await
+                .context("failed to get channel balance")?;
+
+            state
+                .put_ibc_channel_balance(
```
SuperFluffy, line 116: Do we need to worry about making the entirety of these steps a transaction similar to how SQL transactions are atomic (if one fails, all fail or are rolled back)?

crates/astria-proto/src/native/sequencer/ics20_withdrawal.rs
```diff
@@ -0,0 +1,171 @@
...
+use crate::{
+    generated::{
+        primitive::v1::Uint128,
+        sequencer::v1alpha1 as raw,
+    },
+    native::sequencer::v1alpha1::Address,
+};
+
+#[derive(Debug, Clone)]
+pub struct Ics20Withdrawal {
```
SuperFluffy, line 32: I increasingly lean toward abstaining from making all of these private by default.

Can you lock it down?

crates/astria-proto/src/native/sequencer/ics20_withdrawal.rs
```diff
@@ -0,0 +1,171 @@
...
+            return_address,
+            timeout_height,
+            timeout_time: proto.timeout_time,
+            source_channel: ChannelId::from_str(&proto.source_channel)
+                .map_err(Ics20WithdrawalError::InvalidSourceChannel)?,
+        })
+    }
+}
+
+impl From<raw::IbcHeight> for IbcHeight {
```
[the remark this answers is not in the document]
noot: I implemented `From` because the conversion never fails, so I left this for now, but I can change to a `from_raw` or something instead?

crates/astria-sequencer/src/accounts/ics20_withdrawal.rs
```diff
@@ -0,0 +1,132 @@
...
+        }
+
+        state.send_packet_execute(checked_packet).await;
+        Ok(())
+    }
+}
+
+fn is_source(source_port: &PortId, source_channel: &ChannelId, asset: &IbcAsset) -> bool {
+    let prefix = format!("{source_port}/{source_channel}/");
+    !asset.prefix_is(&prefix)
```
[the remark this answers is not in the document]
noot: this is following the ICS20 spec though, so i'd rather leave it like this for now (easier for people reading the spec to understand)

crates/astria-sequencer/src/accounts/ics20_withdrawal.rs
noot: the `"insufficient funds for transfer"` check above covers this

crates/astria-sequencer/src/accounts/ics20_withdrawal.rs
noot: mentioned this before, but check `deliver_tx`, it starts an atomic tx for the db that's only committed if the tx succeeds

crates/astria-proto/src/native/sequencer/ics20_withdrawal.rs
noot: you mean make the fields private? yeah i changed it, made getters instead

crates/astria-proto/src/native/sequencer/ics20_withdrawal.rs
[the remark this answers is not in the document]
SuperFluffy: Yeah, your logic is sound. The reason I don't like `From` here is consistency. The rest of the crate is has `from_raw` and `try_from_raw` so readers will expect to find it.

I intend to expand the `Protobuf` trait to all our proto types to document all `(raw, native)` pairs.

crates/astria-proto/src/native/sequencer/ics20_withdrawal.rs
```diff
@@ -0,0 +1,171 @@
...
+            Self::MissingAmount => f.pad("`amount` field was missing"),
+            Self::InvalidDenom(_) => f.pad("`denom` field was invalid"),
+            Self::InvalidReturnAddress(_) => f.pad("`return_address` field was invalid"),
+            Self::MissingTimeoutHeight => f.pad("`timeout_height` field was missing"),
+            Self::InvalidSourceChannel(_) => f.pad("`source_channel` field was invalid"),
+        }
+    }
+}
+
+impl Error for Ics20WithdrawalError {}
```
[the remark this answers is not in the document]
SuperFluffy: Note on this: in #646 I have reimplemented all errors in terms of thiserror using something like this:
```rust
#[derive(thiserror::Error)]
#[error(transparent)]
struct PublicError(PrivateError);

#[derive(thiserror::Error)]
enum PrivateError{}
```

crates/astria-sequencer/src/accounts/ics20_withdrawal.rs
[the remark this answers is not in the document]
SuperFluffy: Good point, and thanks for pointing that out. Also ugh 😅

crates/astria-proto/src/native/sequencer/ics20_withdrawal.rs
SuperFluffy: Yeah, that's what I meant! Sorry, I meant to write `public` instead of `private`.

crates/astria-sequencer/src/accounts/ics20_withdrawal.rs
```diff
@@ -0,0 +1,140 @@
...
+        ensure!(
+            from_transfer_balance >= self.amount(),
+            "insufficient funds for transfer"
+        );
+
+        state
+            .put_account_balance(
+                from,
+                self.denom().id(),
+                from_transfer_balance - self.amount(),
```
SuperFluffy, line 105: You can do checked math && the check in one go:
```rust
let Some(new_balance) = from_transfer_balance.checked_sub(self.amount()) else {
    bail!("insufficient funds for transfer");
}
```
This addresses my ask for checked math while also defending against accidental refactor.

crates/astria-sequencer/src/accounts/ics20_withdrawal.rs
SuperFluffy: OK, cool. I have to go over how new entries in the DB are created to understand the flow again.

crates/astria-proto/src/native/sequencer/ics20_withdrawal.rs
```diff
@@ -0,0 +1,234 @@
...
+        }
+    }
+
+    #[must_use]
+    pub fn into_raw(self) -> raw::Ics20Withdrawal {
+        raw::Ics20Withdrawal {
+            amount: Some(self.amount.into()),
+            denom: self.denom.to_string(),
+            destination_chain_address: self.destination_chain_address,
+            return_address: self.return_address.0.to_vec(),
```
SuperFluffy, line 114:
```suggestion
            return_address: self.return_address.to_vec(),
```
`Address` has a `to_vec` method

crates/astria-proto/src/native/sequencer/ics20_withdrawal.rs
```diff
@@ -0,0 +1,234 @@
...
+impl From<raw::IbcHeight> for IbcHeight {
+    fn from(h: raw::IbcHeight) -> Self {
+        Self {
+            revision_number: h.revision_number,
+            revision_height: h.revision_height,
+        }
+    }
+}
+
+impl crate::native::Protobuf for IbcHeight {
```
SuperFluffy, line 161: Since all the fields of `raw::IbcHeight` are `Copy` you don't have to impl all the trait methods. The trait definition provides default implementations for `try_from_raw` (in terms of `try_from_raw_ref`) and `into_raw` (in terms of `to_raw)`. So if you impl `to_raw` and `try_from_raw_ref` you get the other two for free.

`try_from_raw` and `into_raw` can be implemented by non-copy types to avoid allocations.

crates/astria-proto/src/native/sequencer/ics20_withdrawal.rs
```diff
@@ -0,0 +1,234 @@
...
+        Self {
+            revision_number: h.revision_number,
+            revision_height: h.revision_height,
+        }
+    }
+}
+
+#[allow(clippy::module_name_repetitions)]
+#[derive(Debug)]
+pub enum Ics20WithdrawalError {
```
SuperFluffy, line 205: See how this is done in PR #646: derive `thiserror::Error` for this and split into a public wrapper type that has `#[error(transparent)]` and a private error type that has all the enums.

crates/astria-sequencer/src/accounts/ics20_withdrawal.rs
```diff
@@ -0,0 +1,140 @@
+use anyhow::{
+    ensure,
+    Context,
```
SuperFluffy, line 3:
```suggestion
    Context as _,
```

proto/astria/sequencer/v1alpha1/transaction.proto
```diff
@@ -75,3 +76,18 @@ message MintAction {
   bytes to = 1;
   astria.primitive.v1.Uint128 amount = 2;
 }
+
+message Ics20Withdrawal {
```
SuperFluffy, line 80: Can you add comments to this type, what IBC spec it matches, maybe with links. and how this is used?

Please also cover the fields and their constraints, i.e. the format of the strings, the number of bytes for things like the return address, units of the time.

proto/astria/sequencer/v1alpha1/transaction.proto
```diff
@@ -75,3 +76,18 @@ message MintAction {
...
+  astria.primitive.v1.Uint128 amount = 1;
+  string denom = 2;
+  string destination_chain_address = 3;
+  bytes return_address = 4;
+  IbcHeight timeout_height = 5;
+  uint64 timeout_time = 6;
```
SuperFluffy, line 86: units? greater 0?

proto/astria/sequencer/v1alpha1/transaction.proto
```diff
@@ -75,3 +76,18 @@ message MintAction {
   bytes to = 1;
   astria.primitive.v1.Uint128 amount = 2;
 }
+
+message Ics20Withdrawal {
+  astria.primitive.v1.Uint128 amount = 1;
+  string denom = 2;
+  string destination_chain_address = 3;
+  bytes return_address = 4;
```
SuperFluffy, line 84: Our address type? 20 bytes?

SuperFluffy (approved):
Looks good know.

I have added a few comments that could be addressed before merge.

Can we rebase this on top of the refactors in #644 #646 in before we merge this?

proto/astria/sequencer/v1alpha1/transaction.proto
noot: yeah it is, in the rust `Ics20Withdrawal` type this is an `Address`

proto/astria/sequencer/v1alpha1/transaction.proto
noot: copied the comments on the rust type here - this is a unix timestamp

proto/astria/sequencer/v1alpha1/transaction.proto
noot: yep, added more description, and added a longer comment + spec link to the rust type

PR #610, opened 2023-11-28
feat(composer, sequencer): add trusted top of block relay

itamarreif:
NOTE: This is very much still a work in progress - doesn't compile. Will fill out details below and provide a design doc as implementation progresses.

## Summary

## Background

## Changes

## Testing
How are these changes tested?

## Breaking Changelist

## Related Issues

closes


crates/astria-proto/src/native/composer/block_relay/v1alpha1/server.rs
```diff
@@ -0,0 +1,81 @@
...
+        sequencer::v1alpha1::Address,
+    },
+};
+
+pub(super) struct TopOfBlockRelayService {
+    // Channel to request the current best bid from the top of block relay actor
+    // TODO: needs a better name
+    best_bid_request_tx: mpsc::Sender<oneshot::Sender<SignedBundle>>,
+    // Bids that have been committed to by proposers
+    pending_commitment: HashMap<Address, SignedBundle>,
```
itamarreif, line 31: A few problems I have with this:
1. Should this state be kept here or should this object communicate via channels to the service object (in the composer crate)?
2. I need to keep the `SignedBundles` that are waiting for a commitment since the commitment will need to be verified against them. If I map it from address, and the `GetBidRequest` has an address field, anyone can call this with a fake address and override the proposer's pending state. Am I overthinking this? Should `GetBidRequest` require the proposer to sign over the block height for which they want the bundle?

crates/astria-composer/src/block_relay/server.rs
```diff
@@ -0,0 +1,65 @@
...
+    generated::block_relay::v1alpha1::top_of_block_relay_server::TopOfBlockRelayServer,
+    native::composer::block_relay::v1alpha1::bid::Bundle,
+};
+use tokio::sync::{
+    mpsc,
+    oneshot,
+};
+use tonic::transport::Server;
+
+pub(super) struct ProposerSubmission {
```
itamarreif, line 17: Should probably just get rid of this and launch the server from `BlockRelay` - still a WIP.

crates/astria-proto/src/native/composer/block_relay/v1alpha1/client.rs
```diff
@@ -0,0 +1,91 @@
...
+    SignedBundle,
+};
+use crate::generated::block_relay::v1alpha1::{
+    top_of_block_relay_client::TopOfBlockRelayClient,
+    GetBidRequest,
+    GetBundleRequest,
+};
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
```
itamarreif, line 20: these need cleaning up

crates/astria-proto/src/native/composer/block_relay/v1alpha1/bid.rs
```diff
@@ -0,0 +1,159 @@
...
+        Ok(Self {
+            block_height: raw.block_height,
+            bid,
+            payload,
+        })
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum OpaqueBidError {
```
itamarreif, line 94: these should probably have something similar to what's done for `SignedTransactionError` and the like in [`sequencer::v1alpha1`](https://github.com/astriaorg/astria/blob/80cbef399b2d6e688671cbda887a1294a55d2ddd/crates/astria-proto/src/native/sequencer/v1alpha1.rs#L23) instead of these silly error enums

crates/astria-composer/src/block_relay/mod.rs
```diff
@@ -0,0 +1,43 @@
...
+        let (committed_bundle_tx, committed_bundle_rx) = mpsc::channel(1);
+        let proposer_submission =
+            ProposerSubmission::new(addr, best_bid_tx, committed_bundle_tx).await?;
+
+        Self {
+            proposer_submission,
+        }
+    }
+
+    pub async fn run_until_stopped(self) -> eyre::Result<()> {
```
itamarreif, line 29: still need to fill in the async stuff here

itamarreif:
closing since this implementation will no longer be used. a new design doc has been posted on notion: https://www.notion.so/astria-org/Trusted-Auction-Relay-PoC-c45ffa2d83bc4f158f8a25b6bea5188e?pvs=4

PR #611, opened 2023-11-28
feat(cli)!: update to use new chart structure, dusk-2

joroshiba:
## Summary
Updates the CLI to work with dusk-2 and the many chart structural updates since then.

## Background
Migrating to dusk-2 which uses new charts which are improved, but have some structural differences

## Changes
- added a `debug` flag and default false on deploy. Now defaults to not printing all k8s
- deduped the sequencer rpc constants
- updated to support namespaces with the new globals
- added naming on the celestia light node to avoid collisions in same namespace
- made namespaces set via the rollup config file

## Testing
Manually deployed a local chart

## Breaking Changelist
- Different APIs hit, supports different charts. The yaml files from previous cli will not be compatible.


SuperFluffy (approved):
Looks fine to me taking into account that you tested it by running

noot (approved):

PR #613, opened 2023-11-29
fix(conductor): validator height should be trailing

joroshiba:
## Summary
Updating to match updated understanding of validator heights. 

## Background
PR https://github.com/astriaorg/astria/pull/589 updated to use the trailing validator, since this is the validator set AFTER the block, but this check was not updated so blocks from DA always rejected.

## Changes
- minor update to da logic

## Testing
CI


noot (approved):
do you want to merge this into main?

PR #617, opened 2023-11-29
feat(conductor): don't attempt to execute finalized blocks

joroshiba:
## Summary
Added a check to check if block is already finalized and bypass attempt at finalization if so. In doing so refactored some logic around how we determine executable block height to avoid duplicating data.

## Background
Accidental double posting of data made a rollup die, and tracking finalized block required slightly different approach due to the way we do block finalization right now (finalized based on latest seen in DA)

## Changes
- Refactor executable block logic
- Add finalizable block logic

## Testing
CI + Manual testing


crates/astria-conductor/src/executor/mod.rs
```diff
@@ -568,19 +568,49 @@ impl Executor {
         Ok(())
     }
 
-    pub(crate) fn get_executable_block_height(&self) -> u32 {
-        self.executable_block_height
+    pub(crate) fn get_executable_block_height(&self) -> Result<u32> {
```
noot, line 571: add a comment saying this returns a sequencer block height

noot (approved):

sambukowski (approved):

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -427,10 +417,11 @@ impl Executor {
     /// returns the previously-computed execution block hash.
     #[instrument(skip(self), fields(sequencer_block_hash = ?block.block_hash, sequencer_block_height = block.header.height.value()))]
     async fn execute_block(&mut self, block: SequencerBlockSubset) -> Result<Block> {
-        if u64::from(self.executable_block_height) != block.header.height.value() {
+        let executable_block_height = self.get_executable_block_height()?;
```
SuperFluffy, line 420: Context

SuperFluffy: https://github.com/astriaorg/astria/pull/620

PR #620, opened 2023-11-30
fix(constructor): add error context, simplify type conversions

SuperFluffy:
## Summary
This patch adds missing error context to fallible operations.

## Background
Errors should not be returned with context.

## Changes
- Add error context with `wrap_err` where missing
- Rename `get_*` to `calculate_*` methods because that's what they do
- Let `calculate_*`methods returned a tendermint `Height` to avoid moving in and out of types.

## Testing
No logic changes, tests still pass.

## Related Issues
Follow-up to https://github.com/astriaorg/astria/pull/617

### DIFF
--- a/crates/astria-conductor/src/executor/mod.rs
+++ b/crates/astria-conductor/src/executor/mod.rs
@@ -3,7 +3,7 @@ use std::collections::HashMap;
 use color_eyre::eyre::{
     self,
     bail,
-    eyre,
+    ensure,
     Result,
     WrapErr as _,
 };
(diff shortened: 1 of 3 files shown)


crates/astria-conductor/src/executor/mod.rs
```diff
@@ -417,15 +417,15 @@ impl Executor {
...
     #[instrument(skip(self), fields(sequencer_block_hash = ?block.block_hash, sequencer_block_height = block.header.height.value()))]
     async fn execute_block(&mut self, block: SequencerBlockSubset) -> Result<Block> {
-        let executable_block_height = self.get_executable_block_height()?;
-        if u64::from(executable_block_height) != block.header.height.value() {
-            error!(
-                sequencer_block_height = block.header.height.value(),
-                executable_block_height = executable_block_height,
-                "block received out of order;"
-            );
-            return Err(eyre!("block received out of order"));
```
SuperFluffy, line 427: this is doubly reported. Actual errors should be handled at the top most level, not deep in the stack.

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -569,7 +571,7 @@ impl Executor {
     }
 
     // Returns the next sequencer block height which can be executed on the rollup
-    pub(crate) fn get_executable_block_height(&self) -> Result<u32> {
+    pub(crate) fn calculate_executable_block_height(&self) -> Result<tendermint::block::Height> {
```
SuperFluffy, line 574: These (this one and the one below) are only ever used as tendermint `Height`s, so it's silly to convert to/from u32 all the time.

PR #621, opened 2023-11-30
fix(conductor): fail hard when executing blocks fails

SuperFluffy:
## Summary
Return an error when executing blocks fails because this means the block chain state is no longer reconcilable.

## Changes
- Return errors rather than just logging them when encountering errors during block execution.
- Refactor shutdown reporting in executor.

## Testing
No testable changes

### DIFF
--- a/crates/astria-conductor/src/executor/mod.rs
+++ b/crates/astria-conductor/src/executor/mod.rs
@@ -344,25 +344,27 @@ impl Executor {
                 biased;
 
                 shutdown = &mut self.shutdown => {
-                    if let Err(e) = shutdown {
-                        let error: &(dyn std::error::Error) = &e;
-                        warn!(error, "shutdown channel return with error; shutting down");
+                    let ret = if let Err(e) = shutdown {
+                        let message = "shutdown channel closed unexpectedly";
+                        error!(error = &e as &dyn std::error::Error, "{message}, shutting down");
+                        Err(e).wrap_err(message)
                     } else {
-                        info!("received shutdown signal; shutting down");
-                    }
-                    break;
+                        info!("received_shutdown_signal, shutting down");
+                        Ok(())
+                    };
+                    break ret;
                 }
 
                 cmd = self.block_channel.recv() => {
                     if let Err(e) = self.handle_executor_command(cmd).await {
-                        let error: &(dyn std::error::Error) = e.as_ref();
-                        error!(error, "failed to handle executor command, breaking from executor loop");
-                        break;
+                        let message = "failed handling executor command";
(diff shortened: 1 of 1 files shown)


SuperFluffy:
Blocked on #620 being merged

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -375,40 +377,36 @@ impl Executor {
...
                 let block_subset =
                     SequencerBlockSubset::from_sequencer_block(*block, self.rollup_id);
 
                 match self.execute_block(block_subset).await {
                     Ok(executed_block) => {
                         if let Err(e) = self.update_soft_commitment(executed_block.clone()).await {
-                            let error: &(dyn std::error::Error) = e.as_ref();
-                            error!(height = height, error, "failed to update soft commitment");
+                            let error: &dyn std::error::Error = e.as_ref();
+                            warn!(%height, error, "failed to update soft commitment");
```
joroshiba, line 395: I think failing to update the soft committment should be treated as an error. Something has gone wrong, the state of the rollup is not correct.

SuperFluffy: The event levels are conceptually:

1. Warn: an error occured, but we can handle it (even if just reporting), and then continue working
2. Error: an error occured and the service cannot handle it. This is usually followed by a shutdown.

So the question here then is: can we go back to a valid state after this error occurred?

From the sound of it we can't - then the logical next step is not to emit an event but stop the service.

joroshiba: Agreed, I don't believe this is reconcilable. Stopping the service sounds reasonable.

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -375,40 +377,36 @@ impl Executor {
...
                         }
                     }
                     Err(e) => {
-                        let error: &(dyn std::error::Error) = e.as_ref();
-                        error!(height = height, error, "failed to execute block");
+                        let error: &dyn std::error::Error = e.as_ref();
+                        warn!(%height, error, "failed to execute block");
```
noot, line 400: same here as above - if we derive an un-executable block, or the execution node fails or whatever reason, that isn't really recoverable

SuperFluffy: failing hard now

SuperFluffy:
Instead of emitting warnings, executor is now returning an error if executing/finalizing blocks failed. This is probably to crass, so I have created https://github.com/astriaorg/astria/issues/624 to at least retry first.

joroshiba (approved):

PR #628, opened 2023-12-01
feat(proto): put many transactions per sequence action

### LINKED ISSUE #623: change `sequence::Action` to a type that takes multiple actions per rollup_id
currently, `sequence::Action` contains `(rollup_id, rollup_tx_bytes)`. we want to change this to `(rollup_id, vec[rollup_tx_bytes])`. this prevents duplication of the 32-byte `rollup_id`s if there are multiple txs for the same rollup within one `SignedTransaction`.

SuperFluffy:
## Summary
Allows putting more than one rollup transaction into a single sequence action. Deprecates the single-transaction field.

## Background
Composer wants to send many transactions belonging to a single rollup to the sequencer. This is wasteful because it has an overhead of `N*32` where `N` is the number of transactions for a rollup and 32 its chain ID. It hence makes sense to group all transactions in a single action.

This is also close to that sequencer does internally at some point, where it groups many transaction bytes by their rollup ID before constructing the merkle hash tree over them.

These changes are not breaking because sequencer will still understand deprecated sequence actions sent to it.

## Changes
- Add `transactions` field to `SequenceAction`.
- Deprecate `data` field on `SequenceAction`.
- Add a conversion method from the deprecated field
- Update sequencer fee calculation to take sum all the sequence action's transaction bytes and the rollup ID
- Update the sequencer block constructor to merge sequence actions by appending rather than per-element pushing
- Make all fields of the idiomatic `SequenceAction` private
- Update all code points that construct sequence actions to use its constructor and setter
- Update all tests to understand the new sequence actions
- Move test utils to construct cometbft headers out of sequencer-types to astria-proto where they belong (as these are now closer to the `SequencerBock::try_from_cometbft` constructor)
- Deduplicate the cometbft header test factory (sequencer-relayer had a copied version)

## Testing
All tests are updated and still pass.
Tests have been added to check that sequence action construction is correct, and to reject sequence actions that have both the new and the old deprecated field set.

## Related Issues

closes #623


jbowen93:
I’m confused by what functionality we gain at the cost of making the API less general with this change. 

Previously with a sequence action containing a single bytes value I could 
1. serialize multiple rollup txs into those bytes
2. serialize a whole rollup block including a header into those bytes

Now with the single bytes being replaced with a repeated transactions field, each of which is still just bytes, I would have a different structure for (1) or (2). Further, (2) would seem semantically incorrect as the naming implies a correct way to structure a sequence action.

crates/astria-proto/proto/astria/sequencer/v1alpha1/transaction.proto
```diff
@@ -54,8 +54,17 @@ message TransferAction {
 // It contains the rollup ID of the destination chain, and the
 // opaque transaction data.
 message SequenceAction {
+    // The rollup ID identifying the rollup these transactions belong to.
+    // Always 32 byyes.
```
joroshiba, line 58: nit: bytes

SuperFluffy:
We decided to shelve this for now. While this API would be more efficient from composer's point of view, it does not add new capabilities that we did not have before, yet would be a breaking change with respect to a rollup needing a new condutor that's able to ingest the changed sequence action format.

PR #629, opened 2023-12-01
chore(ci): move protobuf specs to repository top level

SuperFluffy:
## Summary
Move the Astria protobuf specs into a top-level `proto/` folder and perform Protobuf to Rust code generation in a self-standing compilation tool.

## Background
The protobuf specifications are the single point of truth for Astria's API. As such they should stand by themselves in the repository hierarchy and not be a detail of the `astria-proto` Rust crate as a language-specific implementation.

Furthermore, the current state using integration tests to compile protobuf to rust code via integration tests made for bad developer experience: a error compiler higher level abstractions on top of the generated lower level interfaces lead to the test - and hence code generation - never running. For example, this happened frequently due to the programmer anticipating changes made to a protobuf message but before re-generating code.

Protobuf-to-Rust compilation now happens in the `astria-protobuf-compiler` binary crate, and the check for the Protobuf spec and generate Rust code being in sync is moved to a github action.

## Changes
- Move `crates/astria-proto/proto` under the repo root to `proto/astria`
- Create an `astria-protobuf-compiler` tool that invokes `buf` and calls `tonic-build`
- Run the protobuf compiler in CI and test if it leaves the git tree unchanged, fail otherwise (as opposed to doing this check in Rust before)

## Testing
All Rust code using the protobuf definitions still compiles. Deleting the generated code prior to this change and then re-generating with the new tool leaves the git tree unchanged. Deleting a single file makes the github action fail (meaning the job checking for proto spec and Rust being synced works).


.github/workflows/test.yml
```diff
@@ -12,7 +12,37 @@ on:
 jobs:
   run_checker:
     uses: ./.github/workflows/reusable-run-checker.yml
-              
+
+  protos-compiled:
```
SuperFluffy, line 16: @joroshiba I am not sure if the `tests` workflow is the right place for the CI job. On the one hand we compile rust code there, on the other this feels more fitting for the `release` workflow.

crates/astria-proto/README.md
```diff
@@ -1,33 +1,28 @@
...
+the Rust sources with Astria's protobuf compiler tool from the root of the monorepo:
 
----- build stdout ----
-thread 'build' panicked at 'the generated files have changed; please commit the
-changes', crates/astria-proto/tests/proto_build.rs:126:5
-$ git commit -am "<message about protobuf changes>"
-$ cargo test -p astria-proto build
-test build ... ok
+```sh
+$ cargo run --manifest-path tools/protobuf-compiler
```
noot, line 24: this gives me an error:
```
$ cargo run --manifest-path tools/protobuf-compiler
error: the manifest-path must be a path to a Cargo.toml file
```

noot, line 24:
```suggestion
$ cargo run --manifest-path tools/protobuf-compiler/Cargo.toml
```

noot (approved):
looks good, i like the protobuf-compiler tool :D

SuperFluffy:
Ran pedantic clippy on the protobuf-compiler tool, added the pedantic clippy check in CI.

proto/buf.work.yaml
```diff
@@ -1,3 +1,3 @@
 version: v1
 directories:
-  - proto
+  - astria
```
joroshiba, line 3: should we name the top level folder `src` so its `src/astria` will that break the module naming stuff?

.github/workflows/test.yml
joroshiba: Release in this instance is the push to buf. This seems fine here.

crates/astria-proto/README.md
joroshiba, line 1: nit/bikesheddy: should this be `astria-types` instead of `astria-proto`?

joroshiba (approved):
Added a couple discussion questions, but overall looks fine.

crates/astria-proto/README.md
SuperFluffy: Yes! That's absolutely the plan. I am not a fan of `astria-types` because I find it too non-descript. We will bikeshed over this in the next PR. :-)

proto/buf.work.yaml
SuperFluffy: The issue that this repetition addresses is potentially vendoring other protobufs. Here for example is what penumbra is doing: https://github.com/penumbra-zone/penumbra/tree/main/proto

They have `penumbra` for their own protobuf, but also `rust-vendored` and another.

Buf in their example has the `paymentapis` and `petapis`: https://buf.build/docs/reference/workspaces#configuration

I think the right move here is to get rid of the workspace until we actually make use of it. I will change this.

joroshiba (approved):

PR #631, opened 2023-12-01
feat(sequencer): add IBC gRPC server to sequencer app

### LINKED ISSUE #604: sequencer/IBC: integrate penumbra-ibc's RPC module
we need to implement an RPC server on the sequencer application side, as the IBC relayer needs to make queries into it, generate proofs, etc. we can do this by using the `IbcQuery` type inside penumbra-ibc, and use it with gRPC server (see https://github.com/penumbra-zone/penumbra/blob/main/crates/bin/pd/src/main.rs#L350)

noot:
## Summary
as title says! required for the IBC relayer.

## Changes
- start a gRPC server at startup which serves the relevant IBC endpoints needed. it uses the penumbra IBC RPC implementation, but with our storage as the backend.
- add a config option for `grpc_addr`

## Testing
testing w/ hermes relayer

## Related Issues

closes #604


SuperFluffy:
Damn I feel like this is cursed. It turns the logic of the sequencer app inside out.

Is there a other way of doing this like having a proxy that translates gRPC requests into abci requests over the running cometbft server?

noot:
@SuperFluffy discussed with @joroshiba and we are going to have a gRPC on the app, however it is read-only for now!

noot:
drafting this for now - it turns out that IBC relayers require ABCI queries for state proofs, and the gRPC interface doesn't provide these, so i need to implement ABCI queries for these methods anyways.

crates/astria-sequencer/src/sequencer.rs
```diff
@@ -98,3 +104,47 @@ impl Sequencer {
         Ok(())
     }
 }
+
+fn start_grpc_server(
```
joroshiba, line 108: Should we have create, start, and stop functions on the server? Something akin to what we do in go execution API server? https://github.com/astriaorg/go-ethereum/pull/36/files#diff-cf421ae83f75df74a828a831be8b14967d79fb76fd0703606d804724e32d832f

noot: just wondering why? we don't have start/stop methods on the existing ABCI server

joroshiba: Specifically with grpc because grpc supports streaming, even if these apis don't and ideally want to close those cleanly.

In general we want to move towards having a graceful shutdown though, and closing the server is a part of that.

noot:
@joroshiba the gRPC server already uses `with_graceful_shutdown`, i don't think we need to implement graceful shutdown on top of this.

crates/astria-sequencer/Cargo.toml
```diff
@@ -26,7 +26,8 @@ ed25519-consensus = { workspace = true }
 futures = { workspace = true }
 hex = { workspace = true, features = ["serde"] }
 ibc-types = { workspace = true, features = ["with_serde"] }
-penumbra-ibc = { workspace = true, features = ["component"] }
+# TODO: do we want to feature gate "rpc" as well?
```
SuperFluffy, line 29: IMO no, at least not for now.

crates/astria-sequencer/src/sequencer.rs
```diff
@@ -98,3 +104,44 @@ impl Sequencer {
...
+    use tower_http::cors::CorsLayer;
+
+    let ibc = penumbra_ibc::component::rpc::IbcQuery::new(storage.clone());
+    let cors_layer = CorsLayer::permissive();
+
+    // TODO: setup HTTPS?
+    let grpc_server = Server::builder()
+        .trace_fn(|req| {
+            if let Some(remote_addr) = remote_addr(req) {
+                tracing::error_span!("grpc", ?remote_addr)
```
SuperFluffy, line 125: no debug fields

crates/astria-sequencer/src/sequencer.rs
```diff
@@ -98,3 +104,44 @@ impl Sequencer {
...
+        // allows local applications such as web browsers to talk to pd.
+        .accept_http1(true)
+        // (from Penumbra) Add permissive CORS headers, so pd's gRPC services are accessible
+        // from arbitrary web contexts, including from localhost.
+        .layer(cors_layer)
+        .add_service(ClientQueryServer::new(ibc.clone()))
+        .add_service(ChannelQueryServer::new(ibc.clone()))
+        .add_service(ConnectionQueryServer::new(ibc.clone()));
+
+    tokio::task::Builder::new()
```
SuperFluffy, line 142: This creates a handle to a future that itself is a tokio handle, which is a bit much. Just make this method an `async fn start(...) -> Result<(), Error>` and assign this to a variable in the `run_until_stopped` so that you can join your futures/tasks at the same time.

Also, it's not ideal (because we cannot attach our own span to the spawned task), but it's better to use this

However, in this case it's fine to use https://docs.rs/tonic/latest/tonic/transport/server/struct.Router.html#method.serve_with_shutdown

crates/astria-sequencer/src/sequencer.rs
```diff
@@ -98,3 +104,44 @@ impl Sequencer {
...
 }
+
+fn start_grpc_server(storage: &cnidarium::Storage, grpc_addr: std::net::SocketAddr) -> Result<()> {
+    use ibc_proto::ibc::core::{
+        channel::v1::query_server::QueryServer as ChannelQueryServer,
+        client::v1::query_server::QueryServer as ClientQueryServer,
+        connection::v1::query_server::QueryServer as ConnectionQueryServer,
+    };
+    use penumbra_tower_trace::remote_addr;
+    use tonic::transport::Server;
```
SuperFluffy, line 115: It's confusing that this shadows the `Server` import at the module namespace. IMO just `tonic::transport::Server::builder` further down is fine

crates/astria-sequencer/src/sequencer.rs
```diff
@@ -98,3 +104,44 @@ impl Sequencer {
         Ok(())
     }
 }
+
+fn start_grpc_server(storage: &cnidarium::Storage, grpc_addr: std::net::SocketAddr) -> Result<()> {
```
SuperFluffy, line 108: IDK if we want to multiplex non-ibc things over this grpc server. Either way, I would prefer to call name it for what it does (serve ibc query requests) rather than for how it does it (being a grpc server):
```suggestion
fn start_ibc_query_server(storage: &cnidarium::Storage, grpc_addr: std::net::SocketAddr) -> Result<()> {
```

crates/astria-sequencer/src/sequencer.rs
```diff
@@ -90,6 +90,12 @@ impl Sequencer {
             .finish()
             .ok_or_else(|| anyhow!("server builder didn't return server; are all fields set?"))?;
 
+        let grpc_addr = config
+            .grpc_addr
+            .parse()
+            .context("failed to parse grpc_addr address")?;
+        start_grpc_server(&storage, grpc_addr)?;
```
SuperFluffy, line 97: context

SuperFluffy (changes requested):
Agree with @joroshiba. Please add graceful using `https://docs.rs/tonic/latest/tonic/transport/server/struct.Router.html#method.serve_with_shutdown

noot:
@SuperFluffy if you go to `grpc_serve.serve()` it calls `serve_with_shutdown` internally, which calls `.serve().with_graceful_shutdown()` already

crates/astria-sequencer/src/sequencer.rs
noot: renamed to `start_ibc_grpc_server` for now, in the future i think we are going to add a lot more stuff to this

joroshiba (approved):

crates/astria-sequencer/src/sequencer.rs
```diff
@@ -141,7 +179,41 @@ fn start_grpc_server(storage: &cnidarium::Storage, grpc_addr: std::net::SocketAd
 
     tokio::task::Builder::new()
         .name("grpc_server")
-        .spawn(grpc_server.serve(grpc_addr))
+        .spawn(
+            grpc_server.serve_with_shutdown(grpc_addr, shutdown_rx.map_ok_or_else(|_| (), |()| ())),
```
SuperFluffy:
```suggest
            grpc_server.serve_with_shutdown(grpc_addr, shutdown_rx.unwrap_or_else(|| ()));
```

crates/astria-sequencer/src/sequencer.rs
```diff
@@ -90,22 +104,46 @@ impl Sequencer {
...
+            }
+
+            res = server.listen_tcp(&config.listen_addr) => {
+                match res {
+                    Ok(()) => {
+                        // this shouldn't happen, as there isn't a way for the ABCI server to exit
+                        info!("server exited successfully");
+                    }
+                    Err(e) => {
+                        error!(?e, "server exited with error");
```
SuperFluffy, line 129: mention that the abci server exited. We have 2 servers now ;)

Also, no debug fields: `e as &dyn std::error::Error` or `as_ref()` depending on what this thing is

crates/astria-sequencer/src/sequencer.rs
```diff
@@ -90,22 +104,46 @@ impl Sequencer {
...
+        select! {
+            _ = signals.stop_rx.changed() => {
+                info!("shutting down sequencer");
+            }
+
+            res = server.listen_tcp(&config.listen_addr) => {
+                match res {
+                    Ok(()) => {
+                        // this shouldn't happen, as there isn't a way for the ABCI server to exit
+                        info!("server exited successfully");
```
SuperFluffy, line 126: `abci server`

crates/astria-sequencer/src/sequencer.rs
```diff
@@ -90,22 +104,46 @@ impl Sequencer {
...
             .ok_or_else(|| anyhow!("server builder didn't return server; are all fields set?"))?;
 
+        let (shutdown_tx, shutdown_rx) = tokio::sync::oneshot::channel();
+
         let grpc_addr = config
             .grpc_addr
             .parse()
             .context("failed to parse grpc_addr address")?;
-        start_grpc_server(&storage, grpc_addr)?;
+        start_grpc_server(&storage, grpc_addr, shutdown_rx)?;
```
SuperFluffy: error context

SuperFluffy (approved):
Please address the few comments before merging.

Also please create a follow-up issue to take down the tower-abci server one receiving a sigint/sigterm signal

PR #637, opened 2023-12-06
ci(proto): add proto formatting, cleanup justfile

joroshiba:
## Summary
Cleans up some justfile, updates to check protos against buf, updates readmes,  and adds default formatting of proto files. 

## Background
We moved protos, this caused an issue with breaking changes. Additionally we had non consistent proto formatting.

## Changes
- remove no longer needed just commands
- restructure fmt & lint to take args, add an `all` which is default
- add buf format, run against the repo
- use the buf posted files for breaking changes checks.
- update readmes

## Testing
CI/CD tests, manual testing commands

### DIFF
--- a/justfile
+++ b/justfile
@@ -1,46 +1,67 @@
 default:
   @just --list
 
-create-cluster:
-  kind create cluster --config ./crates/astria-celestia-jsonrpc-client/k8s/cluster-config.yml
-
-deploy-ingress-controller:
-  kubectl apply -f https://raw.githubusercontent.com/kubernetes/ingress-nginx/main/deploy/static/provider/kind/deploy.yaml
-
-wait-for-ingress-controller:
-  kubectl wait --namespace ingress-nginx --for=condition=ready pod --selector=app.kubernetes.io/component=controller --timeout=600s
-
-start-celestia-jsonrpc-test-deployment:
-  kubectl apply -k crates/astria-celestia-jsonrpc-client/k8s/
-
-wait-for-celestia-jsonrpc-test-deployment:
-  kubectl wait --namespace astria-celestia-jsonrpc-client-test deployment --for=condition=Available --selector=app.kubernetes.io/name=astria-celestia-jsonrpc-client-test --timeout=600s
-
-delete-cluster:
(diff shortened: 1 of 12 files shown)


justfile
```diff
@@ -1,46 +1,67 @@
...
+  @just _fmt-rust 
+  @just _fmt-toml
+  @just _fmt-proto
 
-fmt-rust:
+@_lint-all: 
+  -just _lint-rust 
+  -just _lint-toml 
+  -just _lint-proto
+  -just _lint-md
```
joroshiba, line 37: Error codes are swallowed on lint so all outputs will be listed

justfile
```diff
@@ -1,46 +1,67 @@
...
+# Can lint 'rust', 'toml', 'proto', 'md' or 'all'. Defaults to all.
+lint lang=default_lang:
+  @just _lint-{{lang}}
 
-docker-build crate tag=default_docker_tag:
-  docker buildx build --load --build-arg TARGETBINARY={{crate}} -f containerfiles/Dockerfile -t {{crate}}:{{tag}} .
+_fmt-all: 
+  @just _fmt-rust 
+  @just _fmt-toml
+  @just _fmt-proto
```
joroshiba, line 31: this command is run silently, since it's just a pass through

SuperFluffy: also very cool

justfile
SuperFluffy: I am not sure what this means.

SuperFluffy (approved):
These are great. I didn't know about all these just features.

joroshiba:
The output from `just` or `just --list` now:

```bash
Available recipes:
    compile-protos         # Compiles the generated rust code from protos which are used in crates.
    default
    docker-build crate tag=default_docker_tag # Builds docker image for the crate. Defaults to 'local' tag.
    fmt lang=default_lang  # Can format 'rust', 'toml', 'proto', or 'all'. Defaults to all
    install-cli
    lint lang=default_lang # Can lint 'rust', 'toml', 'proto', 'md' or 'all'. Defaults to all.
```

noot (approved):

PR #644, opened 2023-12-11
refactor(proto): rename astria-proto to astria-core

SuperFluffy:
## Summary
Renames `astria-proto` to `astria-core`.

## Background
Originally `astria-proto` contained the astria protobuf specs and the Rust types to interact with them. Since then the protobuf specs have moved to their own root-level directory and the crate contains more idiomatic types and methods.

This PR contains no logic changes.

## Changes
- Rename `astria-proto` to `astria-core`
- Point the protobuf compiler tool to `astria-core/src/generated`
- Remove the module at `astria-core/src/native` (formerly `astria-proto/src/native`), moving all its definitions under `astria-core/src`

## Testing
All tests are still running after adapting the imports


crates/astria-api/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [package]
-name = "astria-proto"
+name = "astria-api"
```
SuperFluffy, line 2: @joroshiba @noot Happy to bikeshed naming. See this as a placeholder.

SuperFluffy: Renamed to `astria-core` after discussion.

crates/astria-core/src/lib.rs
```diff
@@ -1,3 +1,13 @@
+#[cfg(not(target_pointer_width = "64"))]
+compile_error!(
+    "library is only guaranteed to run on 64 bit machines due to casts from/to u64 and usize"
+);
+
+// #[allow(warnings)]
+// #[allow(unreachable_pub)]
```
noot, line 7: remove

noot (approved):
nice!

PR #645, opened 2023-12-11
ci(cli): fix archive upload

joroshiba:
## Summary
Last version always ran as dry run, this fixes while still allowing workflow dispatches on custom builds to ensure binaries build

## Background
Testing github actions is cursed, and didn't realize wouldn't upload previously.

## Changes
- makes tag checking generic
- moves cli to it's own build with check for if it should run.
- dry run checked based on tag instead of event
- adds a name

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,14 +1,19 @@
 on:
+  # TODO: allow workflow dispatch to specify which binaries to build
   workflow_dispatch:
   push:
     tags:
-      - "cli-v[0-9]+.[0-9]+.[0-9]+"
-      - "cli-v[0-9]+.[0-9]+.[0-9]+-alpha.[0-9]+"
-      - "cli-v[0-9]+.[0-9]+.[0-9]+-beta.[0-9]+"
-      - "cli-v[0-9]+.[0-9]+.[0-9]+-rc[0-9]+"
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -1,14 +1,19 @@
...
-      - "cli-v[0-9]+.[0-9]+.[0-9]+"
-      - "cli-v[0-9]+.[0-9]+.[0-9]+-alpha.[0-9]+"
-      - "cli-v[0-9]+.[0-9]+.[0-9]+-beta.[0-9]+"
-      - "cli-v[0-9]+.[0-9]+.[0-9]+-rc[0-9]+"
+      - "**-v[0-9]+.[0-9]+.[0-9]+"
+      - "**-v[0-9]+.[0-9]+.[0-9]+-alpha.[0-9]+"
+      - "**-v[0-9]+.[0-9]+.[0-9]+-beta.[0-9]+"
+      - "**-v[0-9]+.[0-9]+.[0-9]+-rc[0-9]+"
+
+name: Build Binaries
```
jbowen93, line 11: I think name should be the first line. See: https://github.com/astriaorg/astria/blob/main/.github/workflows/release.yml#L1

jbowen93 (approved):

PR #651, opened 2023-12-14
feat(composer): bundle multiple rollup transactions into a single sequencer transaction

### LINKED ISSUE #598: Send bundles with more than 1 transaction to sequencer
At the moment composer sends bundles to sequencer that contain only one transaction.

It should collect many transactions into a bundle before sending it.

### LINKED ISSUE #599: Send bundles up to a max size
Composer should collect transactions up to a maximum size of 256KB before sending them to the sequencer. Quoting @jbowen93:

> Celestia has 2MB blocks and I believe we batch up to 5 Astria blocks to Celestia. So if we wanted to be conservative like 250KB, but we could go as high as 400KB.

itamarreif:
## Summary
This adds logic for bundling multiple `RollupTransaction`s into a single sequencer `SignedTransaction`. It also introduces a `max_bundle_size` cap to the amount of bytes that the `Executor` will submit.

The `Executor` now adds incoming `SequenceAction`s into the `BundleFactory`, which manages a buffer of `SizedBundle`s waiting to be submitted to the `Sequencer`. A `SizedBundle` is a `Vec<Action>` with a limit on how many bytes can be packed into it.
The `Executor` submits a finished bundle as soon as an in-flight submission returns a response.
If there are no finished bundles, the `Executor` will preempt the current bundle being built within 1 sequencer block time of the last submission response by forcing the `BundleFactory` to flush.

## Background
Product requirements of the `composer` service to be able to submit bundles with more than a single `RollupTransaction`. 

## Changes
- Add `SizedBundle`, `BundleFactory`, and `NextFinishedBundle` as well as the logic driving it to the `Executor`
  - Added `max_bundle_sz` and `block_time` config vars
  - Added block timer logic to the `Executor`'s `select!` loop
- Removed `bundle_pending_tx` in the `Searcher` in favor of serializing and forwarding a `SequenceAction` directly from the `Collector` to the `Executor`
  - Removed `collector:RollupTransaction` since it is no longer needed
  - Renamed `bundle_tx/rx` channel handles to `seq_actions_tx/rx` for clarity
  - `Searcher` will now only manage dead `Collector`s and `Executor` states.
- Cleaned up some of the logging and spans in the `Executor`

## Testing
Added unit tests to the executor for the following:
- [x] Check that bundles are flushed when they go over capacity
- [x] Check that bundles are flushed when the block timer ticks
- [x] Check that two `SequenceAction`s are bundled into a single `SignedTransaction` submission
Added unit tests to the bundling logic & associated structs:
- [x] `SizedBundle`
- [x] `BundleFactory` and `NextFinishedBundle`

## Related Issues

closes #599
closes #598


crates/astria-composer/src/config.rs
```diff
@@ -30,6 +30,12 @@ pub struct Config {
     /// Private key for the sequencer account used for signing transactions
     #[serde(serialize_with = "serialize_private_key")]
     pub private_key: SecretString,
+
+    /// Block time in milliseconds
```
noot, line 34: sequencer block time?

crates/astria-composer/local.env.example
```diff
@@ -14,3 +14,10 @@ ASTRIA_COMPOSER_ROLLUPS="astriachain::ws://127.0.0.1:8545"
 # Private key for the sequencer account used for signing transactions
 # Must be a hex-encoded 32-byte array (64-character hex string)
 ASTRIA_COMPOSER_PRIVATE_KEY="2bd806c97f0e00af1a1fc3328fa763a9269723c8db8fac4f93af71db186d6e90"
+
+# Block time in milliseconds
+# Should match the sequencer node configuration for 'timeout_commit', as specified in https://docs.tendermint.com/v0.34/tendermint-core/configuration.html
+ASTRIA_COMPOSER_BLOCK_TIME=2000
+
+# Max bytes to send in a single sequencer `SignedTransaction` (sum of all `SequenceAction`s)
```
noot, line 22:
```suggestion
# Max bytes to send in a single sequencer `SignedTransaction`
```

crates/astria-composer/src/config.rs
```diff
@@ -30,6 +30,12 @@ pub struct Config {
...
+    pub block_time: u64,
+
+    /// Max bytes to send in a single sequencer `SignedTransaction` (sum of all `SequenceAction`s)
+    pub max_bundle_sz: usize,
```
noot, line 38:
```suggestion
    pub max_bundle_size: usize,
```
nit but prefer this :p

crates/astria-composer/src/config.rs
```diff
@@ -30,6 +30,12 @@ pub struct Config {
     /// Private key for the sequencer account used for signing transactions
     #[serde(serialize_with = "serialize_private_key")]
     pub private_key: SecretString,
+
+    /// Block time in milliseconds
+    pub block_time: u64,
+
+    /// Max bytes to send in a single sequencer `SignedTransaction` (sum of all `SequenceAction`s)
```
noot, line 37:
```suggestion
    /// Max bytes to send in a single sequencer `SignedTransaction`
```
since it's actually the encoded size

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -73,14 +135,18 @@ pub(super) struct Executor {
     // The status of this executor
     status: watch::Sender<Status>,
     // Channel for receiving action bundles for submission to the sequencer.
```
noot, line 137: update comment

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -168,23 +243,68 @@ impl Executor {
...
+                    debug!(
+                        new_bundle_sz = ?curr_bundle.curr_sz,
+                        seq_action_sz = ?seq_action_sz,
+                        "bundled new sequence action"
+                    );
+                }
+
+                // submit bundle on block timer tick
+                _ = block_timer.tick(), if !curr_bundle.is_empty() && submission_fut.is_terminated() => {
+                    // receive oneshot from executor
```
noot, line 287:
```suggestion
```

crates/astria-composer/src/config.rs
noot: ok, i see now that in the `SizedBundle` it submits once the sum of the sequence::Action lengths is > this size. however this doesn't take into account signature + pubkey size, nonce, encoding size, etc. that's put into the `SignedTransaction`, so this needs to be less than 250kb by default

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -168,23 +243,68 @@ impl Executor {
...
+                    // flush bundle to the oneshot
```
noot, line 288:
```suggestion
```

crates/astria-composer/src/searcher/mod.rs
```diff
@@ -204,10 +215,26 @@ impl Searcher {
...
                 Some(join_result) = self.conversion_tasks.join_next(), if !self.conversion_tasks.is_empty() => {
                     match join_result {
-                        Ok(bundle) => self.handle_bundle_execution(bundle).await,
+                        Ok(seq_action) => {
+                            let seq_action_sz = seq_action.data.len() + ROLLUP_ID_LEN;
+                            match self.seq_actions_tx.try_send(seq_action) {
+                                Ok(()) => {
+                                    // TODO: use span from the rollup transaction to log here instead
+                                    debug!(
+                                        bytes = seq_action_sz,
```
noot, line 227:
```suggestion
                                        bytes_size = seq_action_sz,
```

crates/astria-composer/src/searcher/mod.rs
```diff
@@ -204,10 +215,26 @@ impl Searcher {
...
+                                        "bundled rollup transaction",
+                                    );
+                                }
+                                Err(e) => warn!(
+                                    error.message = %e,
+                                    error.cause_chain = ?e,
+                                    "failed to forward sequence action to the executor due to backpressure",
```
noot, line 234: could also be due to executor exiting

crates/astria-composer/src/searcher/mod.rs
```diff
@@ -204,10 +215,26 @@ impl Searcher {
...
+                                ),
```
noot, line 235: i think we're changing these to:
```rust
let error: &dyn std::error::Error = e.as_ref();
error!(error, "message here");
```

noot (commented):
code looks basically fine, but i'm wondering what the motivation for this change is? the downside would be that if the signed tx fails to execute (due to running out of fees for example), then all the rollup txs in that tx will fail to be included, whereas previously the txs up until fees ran out would have been included. is this to simplify nonce logic, or some other logic?

crates/astria-composer/src/config.rs
itamarreif: fixed comments

crates/astria-composer/src/searcher/mod.rs
itamarreif: shouldnt that trigger the branch that handles the dead executor tho?

itamarreif:
> code looks basically fine, but i'm wondering what the motivation for this change is? the downside would be that if the signed tx fails to execute (due to running out of fees for example), then all the rollup txs in that tx will fail to be included, whereas previously the txs up until fees ran out would have been included. is this to simplify nonce logic, or some other logic? or is it more so that this makes composer more "generic"?

The main thing is that this demonstrates bundling transactions from multiple rollups into a single atomic bundle.

It also limits the size of transactions that will be submitted to the sequencer, addressing #599, #598.

crates/astria-composer/src/searcher/bundler.rs
```diff
@@ -0,0 +1,135 @@
...
+            buffer: vec![],
+            curr_size: 0,
+            max_size,
+        }
+    }
+
+    /// Buffer `seq_action` into the bundle. If the bundle won't fit `seq_action`, flush `buffer`,
+    /// returning it, and start building up a new buffer using `seq_action`.
+    fn push(&mut self, seq_action: SequenceAction) -> Result<(), SizedBundleError> {
+        let seq_action_size = seq_action.data.len() + ROLLUP_ID_LEN;
```
noot, line 51: two points:
1. i'm seeing this many places, should we make a function for it?
2. this doesn't take into account encoding size, should we have some amount of buffer for that?

crates/astria-composer/src/searcher/bundler.rs
```diff
@@ -0,0 +1,135 @@
...
+
+/// Manages the bundling of sequence actions into `SizedBundle`s. A `SizedBundle` is flushed and
+/// added to the `finished` queue when an incoming `SequenceAction` won't fit in the current bundle.
+/// The `finished` queue operates in FIFO order, where `SizedBundle`s are added to the back and
+/// taken off from the front.
+pub(super) struct Bundler {
+    /// The current bundle being built.
+    pub(super) curr_bundle: SizedBundle,
+    /// The queue of bundles that have been built but not yet sent to the sequencer.
+    pub(super) finished: VecDeque<SizedBundle>,
```
noot, line 87: I think `VecQueue` is fine, it's safer than `Option` as there is potentially a case where we can have multiple bundles waiting for submission, ie the case where a lot of sequence txs/large sequence txs are being submitted, and the `bundler.get_next()` case isn't selected enough (seems very unlikely/niche but still)

crates/astria-composer/src/searcher/bundler.rs
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    /// Get the next bundle from the `finished` queue. If the queue is empty, flush `curr_bundle`
+    pub(super) fn get_next(&mut self) -> Ready<Option<SizedBundle>> {
+        future::ready(self.finished.pop_front())
+    }
+
+    /// Flush the current bundle into the `finished` queue. A bundle can be preempted if there are
+    /// no pending `finished` and the `curr_bundle` is not empty.
+    pub(super) fn preempt_curr_bundle(&mut self) {
```
noot, line 126:
```suggestion
    pub(super) fn flush_curr_bundle(&mut self) {
```
or
`move_curr_bundle_to_finished_queue`? `preempt` is a bit confusing imo

crates/astria-composer/src/searcher/bundler.rs
```diff
@@ -0,0 +1,135 @@
...
+        if !self.curr_bundle.is_empty() && self.finished.is_empty() {
+            debug!(
+                bundle_size=?self.curr_bundle.curr_size,
```
noot, line 129:
```suggestion
                bundle_size=self.curr_bundle.curr_size,
```

noot: don't think you need debug for size stuff? same in other places

crates/astria-composer/local.env.example
```diff
@@ -14,3 +14,11 @@ ASTRIA_COMPOSER_ROLLUPS="astriachain::ws://127.0.0.1:8545"
...
 # Must be a hex-encoded 32-byte array (64-character hex string)
 ASTRIA_COMPOSER_PRIVATE_KEY="2bd806c97f0e00af1a1fc3328fa763a9269723c8db8fac4f93af71db186d6e90"
+
+# Block time in milliseconds
+# Should match the sequencer node configuration for 'timeout_commit', as specified in https://docs.tendermint.com/v0.34/tendermint-core/configuration.html
+ASTRIA_COMPOSER_BLOCK_TIME=2000
+
+# Max bytes to encode into a single sequencer `SignedTransaction`, not including signature,
+# public key, nonce. This is the sum of the sizes of all the `SequenceAction`s
+ASTRIA_COMPOSER_MAX_BUNDLE_SIZE=250000
```
noot, line 24: we should probably take into account signature, pubkey, nonce, encoding size, even if it's just an estimate/upper bound on it, otherwise we can end up submitting bundles the sequencer will always reject.

crates/astria-composer/src/searcher/bundler.rs
```diff
@@ -0,0 +1,135 @@
...
+    /// Buffer `seq_action` into the current bundle. If the bundle won't fit `seq_action`, flush
+    /// `curr_bundle` into the `finished` queue and start a new bundle
+    pub(super) fn push(&mut self, seq_action: SequenceAction) {
+        let seq_action_size = seq_action.data.len() + ROLLUP_ID_LEN;
+        if let Err(SizedBundleError::NotEnoughSpace(seq_action)) = self.curr_bundle.push(seq_action)
+        {
+            // if the bundle is full, flush it and start a new one
+            self.finished.push_back(self.curr_bundle.flush());
+            // TODO: this will error if the seq_action size is larger than the max bundle size -
+            // should panic if its reached this point
```
noot, line 107: is this a TODO?

crates/astria-composer/src/searcher/bundler.rs
```diff
@@ -0,0 +1,135 @@
...
+            new_bundle_size = ?self.curr_bundle.curr_size,
+            seq_action_size = ?seq_action_size,
+            "bundled new sequence action"
+        );
+    }
+
+    /// Get the next bundle from the `finished` queue. If the queue is empty, flush `curr_bundle`
+    pub(super) fn get_next(&mut self) -> Ready<Option<SizedBundle>> {
+        future::ready(self.finished.pop_front())
+    }
```
noot, line 122: the fact that this always returns `Ready` is sus to me. is this going to break the select statement @SuperFluffy ?

you should probably implement `Future` for `Bundler` or some other type inside `Bundler`, eg
```
impl Future for Bundler {
    type Output = SizedBundle;

   fn poll(self: Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<Self::Output> {
        match self.unpin().finished.pop_front() {
             Some(bundle) => Poll::Ready(bundle),
             None => Poll::Pending,
        }
}
```

maybe there's a type for this already in the futures lib somewhere? will look

crates/astria-composer/src/searcher/bundler.rs
itamarreif: thanks - my bad

crates/astria-composer/src/searcher/bundler.rs
itamarreif: forgot to remove - my bad

crates/astria-composer/src/searcher/bundler.rs
itamarreif: This was @SuperFluffy 's suggestions so let's see what he says after reviewing the whole PR

crates/astria-composer/local.env.example
SuperFluffy: Agree. I think this comment can mention that it's an estimate without going into the specifics. Also, don't mention `SignedTransaction` as that is implementation specific naming.

IMO a better name (given that you won't see the comment in a k8s configmap):
```suggestion
ASTRIA_COMPOSER_MAX_BUNDLE_BYTES=250000
```

crates/astria-composer/local.env.example
```diff
@@ -14,3 +14,11 @@ ASTRIA_COMPOSER_ROLLUPS="astriachain::ws://127.0.0.1:8545"
 # Private key for the sequencer account used for signing transactions
 # Must be a hex-encoded 32-byte array (64-character hex string)
 ASTRIA_COMPOSER_PRIVATE_KEY="2bd806c97f0e00af1a1fc3328fa763a9269723c8db8fac4f93af71db186d6e90"
+
+# Block time in milliseconds
+# Should match the sequencer node configuration for 'timeout_commit', as specified in https://docs.tendermint.com/v0.34/tendermint-core/configuration.html
+ASTRIA_COMPOSER_BLOCK_TIME=2000
```
SuperFluffy, line 20:
```suggestion
ASTRIA_COMPOSER_BLOCK_TIME_MS
```
Better to make it clear that this is milliseconds because you won't see the comment in a k8s configmap

crates/astria-composer/src/searcher/bundler.rs
```diff
@@ -0,0 +1,133 @@
...
+#[derive(Debug, thiserror::Error)]
+enum SizedBundleError {
+    #[error("bundle does not have enough space left for the given sequence action")]
+    NotEnoughSpace(SequenceAction),
+}
+
+/// A bundle sequence actions to be submitted to the sequencer. Maintains the total size of the
+/// bytes pushed to it and enforces a max size in bytes passed in the constructor. If an incoming
+/// `seq_action` won't fit in the buffer it is flushed and a new bundle is started.
+pub(super) struct SizedBundle {
```
SuperFluffy, line 28: This `SizedBundle` should probably be an implementation detail of this module and not be leaked outside of it.

crates/astria-composer/src/searcher/bundler.rs
SuperFluffy: I am not a fan of the naming here.

`get_*` taking a mutable borrow is very surprising. It should also not return a ready-future.

IMO this would be a better signature:
```rust
fn pop(&mut self) -> Option<Vec<Action>>
```

Also note the return value: IMO `SizedBundle` should remain an implementation detail. It's fine if downstream users can rely on receiving a bundle of the correct size.

crates/astria-composer/src/searcher/bundler.rs
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    /// Get the next bundle from the `finished` queue. If the queue is empty, flush `curr_bundle`
+    pub(super) fn get_next(&mut self) -> Ready<Option<SizedBundle>> {
+        future::ready(self.finished.pop_front())
+    }
+
+    /// Flush the current bundle into the `finished` queue. A bundle can be preempted if there are
+    /// no pending `finished` and the `curr_bundle` is not empty.
+    pub(super) fn flush_curr_bundle(&mut self) {
```
SuperFluffy, line 124: This smells like we are floating implementation details to the top.

I think this bundling mechanism should expose two methods (can come up with lots of names; just `pop` or `force_pop`, etc):
```rust
fn pop_completed(&mut self) -> Option<Vec<Action>>
fn pop_now(&mut self) -> Vec<Action>
```

`pop_completed` gives you a completed bundle if one is ready, while `force_pop` gives you the bundle that's currently under construction. It might return just an empty vec.

IMO these two are clear and would cover all uses cases without you mechanically telling the collection where to move which bytes. That should remain an implementation detail

crates/astria-composer/src/searcher/collector.rs
```diff
@@ -24,7 +24,7 @@ use tracing::{
 /// A wrapper around an [`ethers::types::Transaction`] that includes the chain ID.
 ///
 /// Used to send new transactions to the searcher.
-pub(super) struct Transaction {
+pub(super) struct RollupTransaction {
```
SuperFluffy, line 27: Not against renaming this, but I don't see the need to rename this in this PR

crates/astria-composer/src/searcher/mod.rs
```diff
@@ -148,34 +160,34 @@ impl Searcher {
...
-    async fn handle_bundle_execution(&self, bundle: Vec<Action>) {
-        // send bundle to executor
-        if let Err(e) = self.bundle_tx.send(bundle).await {
-            error!(
-                error.message = %e,
-                error.cause_chain = ?e,
-                "failed to send bundle to executor",
-            );
-        }
+            if seq_action_sz > max_bundle_sz {
```
SuperFluffy, line 181: This logic is now effectively duplicated here and in your bundle factory. Better have the bundle factory reject the bundle when you try to push it in

crates/astria-composer/src/searcher/mod.rs
```diff
@@ -204,10 +216,28 @@ impl Searcher {
...
                 Some(join_result) = self.conversion_tasks.join_next(), if !self.conversion_tasks.is_empty() => {
                     match join_result {
-                        Ok(bundle) => self.handle_bundle_execution(bundle).await,
+                        Ok(seq_action) => {
+                            let seq_action_size = seq_action.data.len() + ROLLUP_ID_LEN;
+                            match self.seq_actions_tx.try_send(seq_action) {
+                                Ok(()) => {
+                                    // TODO: use span from the rollup transaction to log here instead
+                                    debug!(
+                                        bytes_size = seq_action_size,
```
SuperFluffy, line 228: No need to log this here

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -151,10 +165,15 @@ impl Executor {
             .await
             .wrap_err("failed getting initial nonce from sequencer")?;
         self.status.send_modify(|status| status.is_connected = true);
+
+        let mut block_timer = time::interval(Duration::from_millis(self.block_time));
```
SuperFluffy, line 169: This should not be an interval timer. If you have submitted a tx right before the period completes you will attempt to send a (potentially empty) tx right after

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -164,28 +183,46 @@ impl Executor {
...
+                    );
+
                     // TODO(https://github.com/astriaorg/astria/issues/476): Attach the hash of the
                     // bundle to the span. Linked GH issue is for agreeing on a hash for `SignedTransaction`,
                     // but both should be addressed.
                     let span =  info_span!(
                         "submit bundle",
                         nonce.initial = nonce,
                         bundle.len = bundle.len(),
+                        bundle.size = bundle.curr_size,
```
SuperFluffy, line 204: Don't insert this into the span

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -164,28 +183,46 @@ impl Executor {
...
                     .fuse();
+                    }
+
+                // receive new seq_action and bundle it
+                Some(seq_action) = self.seq_actions_rx.recv(), if submission_fut.is_terminated() => {
+                    bundler.push(seq_action);
+                }
+
+                // try to preempt current bundle if the timer has ticked without submitting the next bundle
+                _ = block_timer.tick(), if submission_fut.is_terminated() => {
```
SuperFluffy, line 224: The way it's written right now this can fire in a hot loop if the interval accumulated too long.

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -164,28 +183,46 @@ impl Executor {
...
+                    bundler.flush_curr_bundle();
```
SuperFluffy, line 225: Just submit a new signed transaction here. Don't drive the data structure hoping that the other select statement will be hit

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -164,28 +183,46 @@ impl Executor {
...
-                        bundle,
+                        bundle: bundle.into_actions(),
                     }
                     .instrument(span)
                     .fuse();
+                    }
+
+                // receive new seq_action and bundle it
+                Some(seq_action) = self.seq_actions_rx.recv(), if submission_fut.is_terminated() => {
+                    bundler.push(seq_action);
```
SuperFluffy, line 220: Could make a `debug!` log here about receiving a new tx

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -164,28 +183,46 @@ impl Executor {
                             break Err(e).wrap_err("failed submitting bundle to sequencer");
                         }
                     }
+                    // reset timer for preempt bundle if it doesn't fill up within the sequencer block time
+                    block_timer.reset_immediately();
```
SuperFluffy, line 187: The interval timer is IMO the wrong thing to use here. I think a countdown timer makes more sense.

SuperFluffy (changes requested):

crates/astria-composer/src/searcher/executor.rs
itamarreif: i do this inside the `Bundler`/`BundleFactory`

crates/astria-composer/src/searcher/bundler.rs
```diff
@@ -0,0 +1,163 @@
...
+    fn into_actions(self) -> Vec<Action> {
+        self.buffer
+    }
+
+    fn is_empty(&self) -> bool {
+        self.buffer.is_empty()
+    }
+}
+
+/// Manages the bundling of sequence actions into `SizedBundle`s. A `SizedBundle` is flushed and
```
SuperFluffy, line 80: `SizedBundle` is an implementation detail and does not show up outside of this module so probably shouldn't be mentioned here.

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -164,28 +216,43 @@ impl Executor {
...
-                        address: self.address,
-                        nonce,
-                        signing_key: self.sequencer_key.clone(),
-                        state: SubmitState::NotStarted,
-                        bundle,
+                    submission_fut = self.make_submission_fut(nonce, bundle);
+                }
+
+                // receive new seq_action and bundle it
+                Some(seq_action) = self.seq_actions_rx.recv(), if submission_fut.is_terminated() => {
```
SuperFluffy, line 232: Should this be disabled if the submission future is in flight? Doesn't that defeat the purpose of having this as an extra branch?

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -164,28 +216,43 @@ impl Executor {
...
-                // receive new bundle for processing
-                Some(bundle) = self.new_bundles.recv(), if submission_fut.is_terminated() => {
-                    // TODO(https://github.com/astriaorg/astria/issues/476): Attach the hash of the
-                    // bundle to the span. Linked GH issue is for agreeing on a hash for `SignedTransaction`,
-                    // but both should be addressed.
-                    let span =  info_span!(
-                        "submit bundle",
-                        nonce.initial = nonce,
-                        bundle.len = bundle.len(),
+                Some(bundle) = bundle_factory.pop_complete(), if submission_fut.is_terminated() => {
```
SuperFluffy, line 223: IMO remove this arm and merge this into  the "submission completed" branch - if the arm returns and there is a new bundle ready to be sent, schedule it immediately without taking an extra iteration through the loop first.

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -164,28 +216,43 @@ impl Executor {
...
+                        }
+                    }
+                }
+
+                // try to preempt current bundle if the timer has ticked without submitting the next bundle
+                () = &mut block_timer, if submission_fut.is_terminated() => {
+                    if let Some(bundle) = bundle_factory.pop_now() {
+                        debug!(
+                            bundle_len=bundle.len(),
+                            "forcing bundle submission to sequencer"
```
SuperFluffy, line 250: This should explain why (because the timer was up)

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -164,28 +216,43 @@ impl Executor {
...
+                        );
+                        submission_fut = self.make_submission_fut(nonce, bundle);
                     }
-                    .instrument(span)
-                    .fuse();
+
+                    block_timer.as_mut().reset(Instant::now() + Duration::from_millis(self.block_time_ms));
```
SuperFluffy, line 255: I don't think this reset here is useful. You have two cases here:

1. the submission terminates in less than the block time => the arm matching on submission completion will reset this
2. the submission exceeds the block time => this arm is disabled until it does, but the timer will be reset either way

SuperFluffy (changes requested):
I like the direction this PR is moving in. It's already simplified a lot compared to its initial state.

I am however requesting changes because I am very sure there is currently a bug (unwanted behaviour) due to the use of `Option::or` rather than `Option::or_else` in the bundle factory.

This might have been caught with unit tests, so I am requesting to add these.

crates/astria-composer/src/searcher/mod.rs
```diff
@@ -201,23 +194,37 @@ impl Searcher {
         loop {
             select!(
                 // serialize and sign sequencer tx for incoming pending rollup txs
-                Some(rollup_tx) = self.new_transactions_rx.recv() => self.bundle_pending_tx(rollup_tx),
+                Some(rollup_tx) = self.new_transactions_rx.recv() => self.serialize_pending_tx(rollup_tx),
```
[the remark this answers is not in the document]
itamarreif: I did this initially but then decided against it because then serialization blocks on reaping other transactions from the geth mempool. Happy to add this in tho.

crates/astria-composer/src/searcher/bundler.rs
```diff
@@ -0,0 +1,163 @@
...
+            Err(SizedBundleError::NotEnoughSpace(seq_action)) => {
+                // if the bundle is full, flush it and start a new one
+                self.finished.push_back(self.curr_bundle.flush());
+                self.curr_bundle
+                    .push(seq_action)
+                    .expect("seq_action should not be larger than max bundle size");
+                Ok(())
+            }
+            Ok(()) => {
+                debug!(
```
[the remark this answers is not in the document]
itamarreif: changed to trace level

crates/astria-composer/src/searcher/executor.rs
itamarreif: good point

crates/astria-composer/src/searcher/bundler.rs
itamarreif: my bad - forgot to update docs after making that change.

crates/astria-composer/src/searcher/executor.rs
itamarreif: I tried this but it results in the first bundle not being fired until the block timer hits, even if it is filled and a new one starts. for that matter, not matter how many `SequenceAction`s are pushed into the `BundleFactory` - they won't be fired until after the first timer tick.

This happens if the submission drivers are only submission responses and timer ticks. Since there won't be a submission response before anything is submitted, it will have to wait for the block timer to tick.

This behavior will also happen at any point where the `BundleFactory` is emptied out entirely:
1. The response for the last bundle submission will be returned
2. The `BundleFactory` will be checked for additional bundles (`pop_completed`), which will return `None`
3. This branch won't fire off another `submission_fut`
4. Loop will continue until next timer tick

Possible solutions:
1. Reducing the timer to less than half of a block time, to make sure it'll hit. This is an easy fix but feel finnicky?
2. Not caring about this and deciding it is acceptable behavior. Will add documentation to reflect this if we want to do this.
3. Current state of the code, and to my understanding this requires the interface on `BundleFactory.pop_completed()` to return a `futures::Ready` so that the `pop_completed` call can be put in the select loop.

This felt like the best solution. WDYT?

crates/astria-composer/src/searcher/executor.rs
itamarreif: good catch

crates/astria-composer/src/searcher/bundler.rs
itamarreif: noted in comments

crates/astria-composer/src/searcher/bundler.rs
```diff
@@ -0,0 +1,163 @@
...
+            })
+            .filter(|bundle| !bundle.is_empty())
+            .map(SizedBundle::into_actions)
+    }
+
+    /// Flush the current bundle into the `finished` queue. A bundle can be preempted if there are
+    /// no pending `finished` and the `curr_bundle` is not empty.
+    fn flush_curr_bundle(&mut self) {
+        if !self.curr_bundle.is_empty() && self.finished.is_empty() {
+            debug!(
```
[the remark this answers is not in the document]
itamarreif: changed to trace

crates/astria-composer/src/searcher/bundler.rs
```diff
@@ -0,0 +1,163 @@
...
+    pub(super) fn pop_complete(&mut self) -> future::Ready<Option<Vec<Action>>> {
+        future::ready(self.finished.pop_front().map(SizedBundle::into_actions))
+    }
+
+    /// Get the next bundle from the `finished` queue. If the queue is empty, flush `curr_bundle`.
+    /// If `finished` and `curr_bundle` are both empty, return `None`.
+    pub(super) fn pop_now(&mut self) -> Option<Vec<Action>> {
+        self.finished
+            .pop_front()
+            .or({
```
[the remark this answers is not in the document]
itamarreif: good catch - fixed

crates/astria-composer/src/searcher/bundler.rs
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    /// Get the next bundle from the `finished` queue.
+    pub(super) fn pop_complete(&mut self) -> future::Ready<Option<Vec<Action>>> {
+        future::ready(self.finished.pop_front().map(SizedBundle::into_actions))
+    }
+
+    /// Get the next bundle from the `finished` queue. If the queue is empty, flush `curr_bundle`.
+    /// If `finished` and `curr_bundle` are both empty, return `None`.
+    pub(super) fn pop_now(&mut self) -> Option<Vec<Action>> {
```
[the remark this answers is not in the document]
itamarreif: Its clearer to match on an option (whether there was a bundle to flush or not) than to have to check whether the bundle you popped is valid or not (checking if the vec returned is empty)

crates/astria-composer/src/searcher/bundler.rs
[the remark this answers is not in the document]
itamarreif: good call. will add

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -164,28 +216,43 @@ impl Executor {
...
-                Some(bundle) = self.new_bundles.recv(), if submission_fut.is_terminated() => {
-                    // TODO(https://github.com/astriaorg/astria/issues/476): Attach the hash of the
-                    // bundle to the span. Linked GH issue is for agreeing on a hash for `SignedTransaction`,
-                    // but both should be addressed.
-                    let span =  info_span!(
-                        "submit bundle",
-                        nonce.initial = nonce,
-                        bundle.len = bundle.len(),
+                Some(bundle) = bundle_factory.pop_complete(), if submission_fut.is_terminated() => {
+                    debug!(
```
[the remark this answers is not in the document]
itamarreif: I agree - this should be replaced with properly managing spans: create one on tx collection, then merge them into a bundle span or something like that. Keeping this in here until that's all added seems fine to me tho? I'll change it to trace level so that it's not as noisy.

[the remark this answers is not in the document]
SuperFluffy: I commented on this in `fn make_submission_fut` (which should probably be named "submit_bundle"): if you properly instrument it you a) don't need to create an `info_span!` inside it, and b) you can put this `debug!` event there so that it will be emitted under the correct span.

crates/astria-composer/src/searcher/bundle_factory.rs
```diff
@@ -0,0 +1,479 @@
...
+                    size: seq_action_size,
+                    max_size: self.curr_bundle.max_size,
+                })
+            }
+            Err(SizedBundleError::NotEnoughSpace(seq_action)) => {
+                // if the bundle is full, flush it and start a new one
+                self.finished.push_back(self.curr_bundle.flush());
+                self.curr_bundle
+                    .push(seq_action)
+                    .expect("seq_action should not be larger than max bundle size");
```
SuperFluffy, line 121: Why is panicking here ok? Because of `SizedBundle::push`? Can you change the `expect` message to that effect?

crates/astria-composer/src/searcher/bundle_factory.rs
```diff
@@ -0,0 +1,479 @@
...
+                    new_bundle_size = self.curr_bundle.curr_size,
+                    seq_action_size = seq_action_size,
+                    "bundled new sequence action"
+                );
+                Ok(())
+            }
+        }
+    }
+
+    /// Get the next bundle from the `finished` queue.
```
SuperFluffy, line 135:
```suggestion
    /// Returns a handle to the next finished bundle if it exists.
    ///
    /// The bundle is only removed from the factory on calling [`NextFinishedBundle::pop`].
    /// This method primarily exists to work around async cancellation.
```

crates/astria-composer/src/searcher/bundle_factory.rs
```diff
@@ -0,0 +1,479 @@
...
+    pub(super) fn next_finished(&mut self) -> Option<NextFinishedBundle> {
```
noot, line 136: why does this return `NextFinishedBundle` instead of just doing the `pop_front` directly?

noot: or rather, when/why is this used instead of `pop_now`?

crates/astria-composer/src/searcher/bundle_factory.rs
```diff
@@ -0,0 +1,479 @@
...
+        if self.finished.is_empty() {
+            None
+        } else {
+            Some(NextFinishedBundle {
+                bundle_factory: self,
+            })
+        }
+    }
+
+    /// Get the next bundle from the `finished` queue. If the queue is empty, flush `curr_bundle`.
```
SuperFluffy, line 146:
```suggestion
    /// Immediately the currently aggregating bundle.
    ///
    /// Returns an empty bundle if there are no bundled transactions.
```

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -70,15 +84,19 @@ use tracing::{
 pub(super) struct Executor {
     // The status of this executor
     status: watch::Sender<Status>,
-    // Channel for receiving action bundles for submission to the sequencer.
-    new_bundles: mpsc::Receiver<Vec<Action>>,
+    // Channel for receiving `SequenceAction`s to be bundled.
+    serialized_rollup_transactions_rx: mpsc::Receiver<SequenceAction>,
```
noot, line 88:
```suggestion
    sequence_actions_rx: mpsc::Receiver<SequenceAction>,
```
perhaps?

crates/astria-composer/src/searcher/bundle_factory.rs
```diff
@@ -0,0 +1,479 @@
...
+    }
+}
+
+/// The size of the `seq_action` in bytes, including the rollup id.
+fn estimate_size_of_sequence_action(seq_action: &SequenceAction) -> usize {
+    seq_action.data.len() + ROLLUP_ID_LEN
+}
+
+#[cfg(test)]
+mod sized_bundle_tests {
```
SuperFluffy, line 176: Consider moving the tests into a separate `bundle_factory/tests.rs` module.

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -70,15 +84,19 @@ use tracing::{
...
     // The client for submitting wrapped and signed pending eth transactions to the astria
     // sequencer.
     sequencer_client: sequencer_client::HttpClient,
     // Private key used to sign sequencer transactions
     sequencer_key: SigningKey,
     // The sequencer address associated with the private key
     address: Address,
+    // Milliseconds for bundle timer to make sure bundles are submitted at least once per block.
+    block_time_ms: u64,
```
SuperFluffy, line 97: Maket this a `std::time::Duration` in the constructor

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -139,22 +161,41 @@ impl Executor {
...
+    fn submit_bundle(&self, nonce: u32, bundle: Vec<Action>) -> Fuse<SubmitFut> {
+        SubmitFut {
+            client: self.sequencer_client.clone(),
+            address: self.address,
+            nonce,
+            signing_key: self.sequencer_key.clone(),
+            state: SubmitState::NotStarted,
+            bundle,
+        }
+        .fuse()
```
SuperFluffy, line 175: This future is not instrumented. Use `tracing::Instrument::in_current_span`

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -164,28 +205,43 @@ impl Executor {
                             break Err(e).wrap_err("failed submitting bundle to sequencer");
                         }
                     }
+
+                    block_timer.as_mut().reset(Instant::now() + Duration::from_millis(self.block_time_ms));
+                }
+
+                Some(next_bundle) = future::ready(bundle_factory.next_finished()), if submission_fut.is_terminated() => {
+                    let bundle = next_bundle.pop();
+                    if !bundle.is_empty() {
```
SuperFluffy, line 214: A debug level event to the effect "timer is but no transactions to send" would be  great

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -164,28 +205,43 @@ impl Executor {
...
+                        submission_fut = self.submit_bundle(nonce, bundle);
+                    }
+                }
+
+                // receive new seq_action and bundle it
+                Some(seq_action) = self.serialized_rollup_transactions_rx.recv() => {
+                    let rollup_id = seq_action.rollup_id;
+                    match bundle_factory.try_push(seq_action) {
+                        Ok(()) => {}
```
SuperFluffy, line 223: Replace this `match` with:
```rust
if let Err(e) = bundle_factory.try_push(seq_action) {}
```
you don't need that `Ok` branch

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -164,28 +205,43 @@ impl Executor {
...
+                        Err(BundleFactoryError::SequenceActionTooLarge{size, max_size}) => {
+                            warn!(
+                                rollup_id = %rollup_id,
+                                seq_action_size = size,
+                                max_size = max_size,
+                                "failed to bundle sequence action: too large. sequence action is dropped."
```
SuperFluffy, line 229: The `seq_action_size` and `max_size` fields should come out of the display-formatted error. I.e. derive `thiserror::Error` with the same information and don't set the fields here explicitly. Just match the error with `Err(e)` and then `error = %e as &StdError`

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -164,28 +205,43 @@ impl Executor {
...
-                        client: self.sequencer_client.clone(),
-                        address: self.address,
-                        nonce,
-                        signing_key: self.sequencer_key.clone(),
-                        state: SubmitState::NotStarted,
-                        bundle,
+                // try to preempt current bundle if the timer has ticked without submitting the next bundle
+                () = &mut block_timer, if submission_fut.is_terminated() => {
+                    let bundle = bundle_factory.pop_now();
+                    if !bundle.is_empty() {
```
SuperFluffy, line 238: Just like above, I think  it might be valuable info to notify that the timer was up, but the bundle was empty

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -237,12 +293,15 @@ async fn get_latest_nonce(
     fields(
         nonce = tx.unsigned_transaction().nonce,
         transaction.hash = hex::encode(sha256(&tx.to_raw().encode_to_vec())),
+        bundle.len = %tx.actions().len()
```
SuperFluffy, line 296: I don't think this should be in the span

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -407,11 +466,420 @@ impl Future for SubmitFut {
...
 
 fn sha256(data: &[u8]) -> [u8; 32] {
-    use sha2::{
-        Digest as _,
-        Sha256,
+    use sha2::Sha256;
+    Sha256::digest(data)
+}
+
+#[cfg(test)]
```
SuperFluffy, line 473: Thank you for going through the effort of adding these tests. I am wondering if we are overspecifying these

crates/astria-composer/src/searcher/executor.rs
```diff
@@ -407,11 +466,420 @@ impl Future for SubmitFut {
...
+mod tests {
```
SuperFluffy, line 474: Please move `executor.rs -> executor/mod.rs` and this module into its own `executor/tests.rs`

SuperFluffy (approved):
Some comments here and there (like the wrong span for the submission future), but looks good all in all. Please address the comments before you merge.

Thank you for going through the effort of adding all the tests. I do wonder if we are overspecifying (and hence calcifying) our tests, but I trust that this patch is doing what it's supposed to be doing.

crates/astria-composer/src/searcher/bundle_factory.rs
SuperFluffy: This is required to work around async cancellation. In short, assume you encounter something like this:

```rust
loop {
    select!(
        Some(bundle) = std::future::ready(the_factory.pop()) => {}
        Some(tx) = a_channel.recv() => {}
    }
}
```
Here, the futures are constructed upon entering the `select!` statement, but they are not polled in this but in a random order. So the ready-future is constructed by popping a bundle, moving it out of the factory and into the future. But if the channel is polled and immediately ready, then the ready-future is dropped (cancelled), and you lose the bundle.

Making the removal of the bundle lazy works around this async cancellation.

PR #655, opened 2023-12-14
chore(sequencer)!: bump penumbra deps, remove `AppHash`

noot:
## Summary
bump penumbra deps to 0.64.1

## Changes
- bump penumbra deps to 0.64.1 (renames penumbra-storage to cnidarium)
- also remove `AppHash` from sequencer as it will break IBC proofs, penumbra removed their analogous type a while back.

## Breaking Changelist
- `AppHash` type is removed from sequencer, this will change the `app_hash` and `block_hash` of cometbft blocks.


crates/astria-sequencer/Cargo.toml
```diff
@@ -24,10 +24,10 @@ ed25519-consensus = { workspace = true }
 futures = { workspace = true }
 hex = { workspace = true, features = ["serde"] }
 ibc-types = { workspace = true, features = ["with_serde"] }
-penumbra-component = { workspace = true }
+cnidarium = { workspace = true }
```
SuperFluffy, line 27: wonder if this should no longer be a workspace dependency but a sequencer-only dependency

Cargo.toml
```diff
@@ -44,11 +44,11 @@ once_cell = "1.17.1"
 sha2 = "0.10"
 serde = "1"
 serde_json = "1"
-penumbra-component = { git = "https://github.com/penumbra-zone/penumbra.git", rev = "ab1f82970af76091b962509831ec0f53cfeb85f4" }
-penumbra-ibc = { git = "https://github.com/penumbra-zone/penumbra.git", rev = "ab1f82970af76091b962509831ec0f53cfeb85f4", default-features = false }
-penumbra-proto = { git = "https://github.com/penumbra-zone/penumbra.git", rev = "ab1f82970af76091b962509831ec0f53cfeb85f4" }
-penumbra-storage = { git = "https://github.com/penumbra-zone/penumbra.git", rev = "ab1f82970af76091b962509831ec0f53cfeb85f4" }
-penumbra-tower-trace = { git = "https://github.com/penumbra-zone/penumbra.git", rev = "ab1f82970af76091b962509831ec0f53cfeb85f4" }
+cnidarium = { git = "https://github.com/penumbra-zone/penumbra.git", tag = "v0.64.1" }
```
SuperFluffy, line 47: please mention the rename in the PR description.

Also, should we have a standing issue to migrate to crates.io? Looks like they are prepping a crate?

crates/astria-sequencer/src/service/info/mod.rs
```diff
@@ -72,13 +72,12 @@ impl Info {
...
-                        .root_hash()
-                        .await
-                        .context("failed to get app hash")?,
-                );
+                let app_hash = self
+                    .storage
+                    .latest_snapshot()
+                    .root_hash()
+                    .await
+                    .context("failed to get app hash")?;
```
SuperFluffy, line 80: is calling this an app hash still accurate?

SuperFluffy (approved):
Changes look straightforward to me. Only some comments re cnidarium and imprecise naming of app hash vs root hash.

Also, is this truly an externally observable breaking change?

crates/astria-sequencer/Cargo.toml
noot: yeah, looks like only sequencer uses this now. will update

crates/astria-sequencer/src/service/info/mod.rs
noot: yeah, it's returned as the `app_hash` inside the cometbft block

noot:
@SuperFluffy yeah, i updated the description, but this is breaking as it changes the block's `app_hash` and thus `block_hash`.

PR #656, opened 2023-12-15
feat(telemetry)!: use opentelemetry

SuperFluffy:
## Summary
Export trace data as opentelemetry, replacing the tracing json and pretty-print formatting subscribers.

## Background
We want to export our trace data in a formalized way, targeting observability platforms that understand the opentelemetry protocol (OTLP).

If the opentelemetry-otlp has hard-coded env vars that are always read at runtime to configure it. These are (see https://github.com/open-telemetry/opentelemetry-rust/blob/4e80e3a2ebf96a166917b169f7d27e44e83cd3da/opentelemetry-otlp/src/span.rs#L30-L42):

+ `OTEL_EXPORTER_OTLP_ENDPOINT`: to set the OTLP endpoint
+ `OTEL_EXPORTER_OTLP_TRACE_ENDPOINT`: to set the trace OTLP endpoint (taking precedence over the general one)
+ `OTEL_EXPORTER_OTLP_TRACE_TIMEOUT`: to set the timeout for the backend to process span batches
+ `OTEL_EXPORTER_OTLP_TRACE_COMPRESSION`: the compression algorithm to use for compressing spans
+ `OTEL_EXPORTER_OTLP_HEADERS`: key-value pairs to set as grpc/http headers
+ `OTEL_EXPORTER_OTLP_TRACE_HEADERS`: key-value pairs to set as grpc/http headers taking precedence over the more general one.

If none of the above are set then their default values are used according to the `opentelemetry-otlp` crate.

Because these are always read and take precedence over custom per-service settings the service config do not contain extra settings.

## Changes
- Replace the tracing_subscriber fmt layer by tracing-opentelemetry
- Send tracing data to an OTLP endpoint (this is understood to replace the json-formatted output to stdout)
- Allow disabling OTLP with `ASTRIA_<SERVICE>_NO_OTLP=true`
- Optionally emit otel traces to stdout if connected to a tty (this replaces the pretty printer)
- Allow forcing writing to stdout with `ASTRIA_<SERVICE>_FORCE_STDOUT=true`

## Testing
In tests: trace data is written to stdout
In a cluster: **not yet tested**

## Breaking Changelist
- The various services got new settings with env vars `ASTRIA_<SERVICE>_NO_OTEL` and `ASTRIA_<SERVICE>_FORCE_STDOUT`, which must be set
- if `ASTRIA_<SERVICE>_NO_OTEL=false` then a GRPC-ready opentelemetry endpoint must be available at the default location `http://localhost:4317` or at a location defined via `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` or `OTEL_EXPORTER_OTLP_ENDPOINT`

## Related Issues
closes https://github.com/astriaorg/astria/issues/652

### DIFF
--- a/crates/astria-composer/src/main.rs
+++ b/crates/astria-composer/src/main.rs
@@ -1,24 +1,37 @@
+use std::process::ExitCode;
+
 use astria_composer::{
     telemetry,
     Composer,
     Config,
 };
+use color_eyre::eyre::WrapErr as _;
 use tracing::info;
 
 #[tokio::main]
-async fn main() {
+async fn main() -> ExitCode {
     let cfg: Config = match config::get() {
         Ok(cfg) => cfg,
         Err(e) => {
             eprintln!("failed to read configuration: {e}");
-            std::process::exit(2);
+            return ExitCode::FAILURE;
         }
     };
+
+    if let Err(e) = telemetry::configure()
+        .set_no_otel(cfg.no_otel)
+        .set_force_stdout(cfg.force_stdout)
+        .filter_directives(&cfg.log)
+        .try_init()
+        .wrap_err("failed to setup telemetry")
+    {
+        eprintln!("initializing composer failed:\n{e:?}");
+        return ExitCode::FAILURE;
+    }
+
     let cfg_ser = serde_json::to_string(&cfg)
         .expect("the json serializer should never fail when serializing to a string");
     eprintln!("config:\n{cfg_ser}");
-
-    telemetry::init(std::io::stdout, &cfg.log).expect("failed to initialize tracing");
 
     info!(config = cfg_ser, "initializing composer",);
 
@@ -26,4 +39,5 @@ async fn main() {
         .expect("failed creating composer")
         .run_until_stopped()
         .await;
+    ExitCode::SUCCESS
 }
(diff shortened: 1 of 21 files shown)


crates/astria-composer/src/main.rs
```diff
@@ -1,29 +1,42 @@
...
+
+    if let Err(e) = telemetry::configure()
+        .otel_endpoint("http://otel-collector.monitoring:4317")
+        .filter_directives(&cfg.log)
+        .try_init()
+        .wrap_err("failed to setup telemetry")
+    {
+        eprintln!("initializing sequencer failed:\n{e:?}");
+        return ExitCode::FAILURE;
+    }
```
noot, line 29: is the otel-collector something run locally or remotely? will starting up the service fail is the otel-collector is not available?

SuperFluffy: After your comment I questioned my decision to call this a non-breaking change. I added config options optionally disable OTLP so that this can be run locally without an otel collector.

noot (approved):

crates/astria-sequencer-relayer/local.env.example
```diff
@@ -1,3 +1,13 @@
+# A list of filter directives of the form target[span{field=value}]=level.
+ASTRIA_SEQUENCER_RELAYER_LOG=astria_sequencer_relayer=info
```
sambukowski, line 2: Should this have quotes? `ASTRIA_SEQUENCER_RELAYER_LOG="astria_sequencer_relayer=info"`

SuperFluffy: No, that's not necessary:

```sh
$ export FOO=bar
$ echo $FOO
bar
```

PR #660, opened 2023-12-20
fix(sequencer): clear processed tx count in begin_block

noot:
same as #659 but on top of sequencer tag 0.7.0


noot:
gonna close for now as i assume @joroshiba has already cherry picked necessary changes from main

PR #662, opened 2023-12-21
feat(sequencer-utils): add `--chain-id` to sequencer genesis utils

noot:
## Summary
utils to set the chain ID in the cometbft genesis.json file.

## Background
need it for ibc testing (we need to set the chain id when starting hermes) this makes my life a lot easier, cause otherwise it picks a random chain ID.

## Changes
- add `--chain-id` to sequencer genesis utils which sets the `chain_id` field in the cometbft genesis file to the given string

## Testing
unit tests + i ran it (`just run-cometbft` and `cat ~/.cometbft/config/genesis.json`)

### DIFF
--- a/crates/astria-sequencer-utils/src/genesis_parser.rs
+++ b/crates/astria-sequencer-utils/src/genesis_parser.rs
@@ -48,7 +48,11 @@ impl GenesisParser {
                 .wrap_err("failed deserializing cometbft genesis state from file")?;
 
         // insert sequencer app genesis data into cometbft genesis data
-        insert_app_state(&mut destination_genesis_data, &source_genesis_data);
+        insert_app_state_and_chain_id(
+            &mut destination_genesis_data,
+            &source_genesis_data,
+            data.chain_id,
+        );
 
         // write new state
         let dest_file = File::create(Path::new(data.destination_genesis_file.as_str()))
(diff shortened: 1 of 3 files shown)


crates/astria-sequencer-utils/src/genesis_parser.rs
```diff
@@ -60,11 +64,12 @@ impl GenesisParser {
     }
 }
 
-fn insert_app_state(dst: &mut Value, app_state: &Value) {
+fn insert_app_state_and_chain_id(dst: &mut Value, app_state: &Value, chain_id: String) {
```
SuperFluffy, line 67: This doesn't really scale IMO. Better to replace this by
```rust
dst["app_state"] = json!(source_genesis_data);
dst["chain_id"] = json!(chain_id);
```

sambukowski (approved):

crates/astria-sequencer-utils/src/genesis_parser.rs
noot: like replace the whole function? that is harder for unit tests :/

SuperFluffy: Yeah, this should be rewritten so that not everything is stuffed into a single function.

SuperFluffy (approved):
unhappy with how this tool looks right now, but makes sense to not block it on a refactor.

PR #663, opened 2023-12-21
fix(sequencer): handle abci events properly + improve tx logging

### LINKED ISSUE #476: Transaction Hashes
There's currently no agreed upon way to hash our `SignedTransaction`s (i.e. no method or trait implemented). This is useful for logging as transactions need to be tracked across services.

- [ ] look at how `CometBFT` implements transaction hashing
- [ ] add the functionality to the `SignedTransaction` type

noot:
## Summary
we were ignoring ABCI events (bad). needed to fix this for hermes testing

## Changes
- return ABCI events emitted in `deliver_tx`
- also minor logging improvements, the full tx in `deliver_tx` was getting logged twice but the tx hash wasn't logged anywhere, this was extremely annoying for testing, so i implemented the `SignedTransaction::hash()` and logged that

## Testing
i ran it with hermes, also cometbft gets the same tx hashes.

## Related Issues

closes #476

### DIFF
--- a/crates/astria-sequencer/src/app.rs
+++ b/crates/astria-sequencer/src/app.rs
@@ -361,7 +361,7 @@ impl App {
     /// successfully.
     ///
     /// Note that `begin_block` is now called *after* transaction execution.
-    #[instrument(name = "App::deliver_tx", skip(self))]
+    #[instrument(name = "App::deliver_tx", skip_all, fields(tx_hash = hex::encode(signed_tx.hash())))]
     pub(crate) async fn deliver_tx(
         &mut self,
         signed_tx: astria_core::sequencer::v1alpha1::SignedTransaction,
@@ -392,7 +392,7 @@ impl App {
         transaction::execute(&signed_tx, &mut state_tx)
             .await
             .context("failed executing transaction")?;
-        state_tx.apply();
+        let (_, events) = state_tx.apply();
 
         // note: deliver_tx is now called (internally) before begin_block,
         // so increment the logged height by 1.
@@ -400,12 +400,15 @@ impl App {
             "block height must be set, as `begin_block` is always called before `deliver_tx`",
         );
         info!(
+            tx_hash = hex::encode(signed_tx.hash()),
             ?signed_tx,
             height = height + 1,
             sender = %Address::from_verification_key(signed_tx.verification_key()),
+            events = ?events,
             "executed transaction"
         );
-        Ok(vec![])
+
+        Ok(events)
     }
 
     #[instrument(name = "App::end_block", skip(self))]
--- a/crates/astria-core/src/sequencer/v1alpha1/transaction/mod.rs
+++ b/crates/astria-core/src/sequencer/v1alpha1/transaction/mod.rs
@@ -4,6 +4,7 @@ use ed25519_consensus::{
     VerificationKey,
 };
 use prost::Message as _;
+use sha2::Sha256;
 
 use super::raw;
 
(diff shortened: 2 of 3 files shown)


crates/astria-sequencer/src/app.rs
```diff
@@ -361,7 +361,7 @@ impl App {
     /// successfully.
     ///
     /// Note that `begin_block` is now called *after* transaction execution.
-    #[instrument(name = "App::deliver_tx", skip(self))]
+    #[instrument(name = "App::deliver_tx", skip_all, fields(tx_hash = hex::encode(signed_tx.hash())))]
```
SuperFluffy, line 364: Several notes on this:

1. I want to get away from `hex::encode` because it has an unnecessary allocation. Can you write a newtype wrapper like`DisplayHex<'a, T>(&'a T)` with `impl<'a, T> std::fmt::Display for DisplayHex<'a, T> where T: AsRef<[u8>]` that walks the bytes and writes the individual bytes to the formatter like here? https://github.com/astriaorg/astria/blob/04cfccb5109df7aa953d2767b7c26a5605c0931e/crates/astria-core/src/sequencer/v1alpha1/mod.rs#L97-L104
2. I think it would be nice to have the hash of both the signed and the unsigned transactions here, and the public key. Not sure if this is overkill, but IMO very useful. The reason to have the hash of the unsigned transactions is that on the composer side we only get the hash of the signed transaction immediately before sending. The signed tx hash also changes if the nonce is wrong, while the unsigned tx remains stable.
3. Also please use more verbose field names. I am thinking something like `hash.signed_transaction`, `hash.unsigned_transaction` and `sender`

crates/astria-sequencer/src/app.rs
```diff
@@ -392,20 +392,23 @@ impl App {
...
-        state_tx.apply();
+        let (_, events) = state_tx.apply();
 
         // note: deliver_tx is now called (internally) before begin_block,
         // so increment the logged height by 1.
         let height = self.state.get_block_height().await.expect(
             "block height must be set, as `begin_block` is always called before `deliver_tx`",
         );
         info!(
+            tx_hash = hex::encode(signed_tx.hash()),
```
SuperFluffy, line 403: Remove this. It's already in the span

crates/astria-sequencer/src/app.rs
```diff
@@ -392,20 +392,23 @@ impl App {
...
             ?signed_tx,
             height = height + 1,
             sender = %Address::from_verification_key(signed_tx.verification_key()),
```
SuperFluffy, line 406: Ah, this is in regard to my comment above logging the pub key - let's move this out of here

crates/astria-sequencer/src/app.rs
```diff
@@ -392,20 +392,23 @@ impl App {
...
+        let (_, events) = state_tx.apply();
 
         // note: deliver_tx is now called (internally) before begin_block,
         // so increment the logged height by 1.
         let height = self.state.get_block_height().await.expect(
             "block height must be set, as `begin_block` is always called before `deliver_tx`",
         );
         info!(
+            tx_hash = hex::encode(signed_tx.hash()),
             ?signed_tx,
```
SuperFluffy, line 404: If we touch this event let's fix it to remove this field entirely. It's too verbose.

crates/astria-sequencer/src/app.rs
```diff
@@ -392,20 +392,23 @@ impl App {
...
             height = height + 1,
             sender = %Address::from_verification_key(signed_tx.verification_key()),
+            events = ?events,
```
SuperFluffy, line 407: no debug formatted logs in events

SuperFluffy (changes requested):

noot:
closing cause slightly stale, will re-open another PR

PR #664, opened 2023-12-21
feat(ci): lint debug fields in tracing events

SuperFluffy:
## Summary
Add a custom lint using the [`dylint`](https://github.com/trailofbits/dylint) tool to detect and warn about debug-formatted fields in tracing events.

## Background
Debug-formatted fields in tracing events are bad practice. They are hard to read and parse, and they surface language-specific implementation details that have no place in observability.

The present lint only acts on debug-formatted fields in `tracing::event!` macros (which includes macros like `info!` and `warn!` that are implemented in terms of it). Debug-formatted fields in `#[instrument]` proc macros will be detected in a separate lint.

## Changes
- Implement a custom lint using dylint that flags that debug-formatted fields in `tracing::event!` macros
- Adds tests for the lint
- Add a github job to run dylints (the lint failing does currently not fail CI)
- Add a `default-members` field to the virtual workspace so that the lints are exempt from most CI jobs (the lints require a nightly rust)

## Testing
This patch provides `compiletest-rs` macros that correctly detects debug-formatted fields in event macros and not in instrument-proc-macros.


lint/tracing_debug_field/ui/main.rs
```diff
@@ -0,0 +1,11 @@
...
+
+#[derive(Clone, Copy, Debug)]
+struct Wrapped(&'static str);
+
+fn main() {
+    let field = Wrapped("wrapped");
+    info!(field = tracing::field::debug(field), "using the function");
+    info!(field = ?field, "using the sigil");
+    info!(?field, "using shorthand with sigil");
+}
```
noot, line 11: what is this for? just testing?

noot (approved):
looks good! the `if_chain` is cool :D

lint/tracing_debug_field/ui/main.rs
SuperFluffy: Yeah, this `ui/main.rs` produces some output when compiled. https://github.com/Manishearth/compiletest-rs then tests the output against `ui/main.stderr`.

This is to make sure that all the cases that I want the lint to catch are caught.

joroshiba (approved):

PR #666, opened 2023-12-21
fix(sequencer): fix ABCI event handling

noot:
## Summary
we were ignoring ABCI events (bad). needed to fix this for hermes testing

## Changes
- return ABCI events emitted in `deliver_tx`

## Testing
i ran it with hermes, this fixes errors due to no events

### DIFF
--- a/crates/astria-sequencer/src/app.rs
+++ b/crates/astria-sequencer/src/app.rs
@@ -392,7 +392,7 @@ impl App {
         transaction::execute(&signed_tx, &mut state_tx)
             .await
             .context("failed executing transaction")?;
-        state_tx.apply();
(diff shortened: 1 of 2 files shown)


SuperFluffy (approved):
Approving because you say that this fixes issues in Hermes.

However, I don't understand what these events are needed for and why Hermes doesn't like it. Can you provide links to the relevant ibc/cometbft/abci docs and mention why Hermes does not like it for learning and understanding context?

noot:
@SuperFluffy mentioned on the call, but basically this is a bug, hermes expected ABCI events to be returned in response to doing certain IBC actions, so if they don't happen then hermes just thinks the tx failed and errors.

PR #670, opened 2024-01-01
refactor(sequencer): call abort on ABCI server on signal

### LINKED ISSUE #667: sequencer: shut down ABCI server on shutdown signal
can spawn the ABCI server in a task and do `task.abort()` when the signal is received. workaround since the server has no way to shutdown otherwise.

mattdf:
## Summary
Spawns the server task with tokio and aborts the server properly when a signal is received.

## Changes
Added a oneshot channel and created a tokio task inside `run_until_stopped` with a `server_handle.abort()` at the before return.

## Testing
Ran it and the server shuts down properly.

## Related Issues

closes #667

### DIFF
--- a/crates/astria-sequencer/src/sequencer.rs
+++ b/crates/astria-sequencer/src/sequencer.rs
@@ -106,6 +106,7 @@ impl Sequencer {
             .ok_or_else(|| anyhow!("server builder didn't return server; are all fields set?"))?;
 
         let (shutdown_tx, shutdown_rx) = tokio::sync::oneshot::channel();
+        let (server_exit_tx, server_exit_rx) = tokio::sync::oneshot::channel();
 
         let grpc_addr = config
             .grpc_addr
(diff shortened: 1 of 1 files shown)


crates/astria-sequencer/src/sequencer.rs
```diff
@@ -114,21 +115,26 @@ impl Sequencer {
...
-                    Ok(()) => {
-                        // this shouldn't happen, as there isn't a way for the ABCI server to exit
-                        info!("ABCI server exited successfully");
-                    }
-                    Err(e) => {
-                        error!(err = e.as_ref(), "ABCI server exited with error");
-                    }
-                }
+            _ = server_exit_rx => {
+                info!("ABCI server task exited")
```
noot, line 137:
```suggestion
               error!("ABCI server task exited");
```

noot (approved):

SuperFluffy:
Small note on this: killing the task outright is the right thing to do given the API we have. What would be better is to have a way to kill the [tower-abci](https://github.com/penumbra-zone/tower-abci) server given a signal. For example how tonic is doing it: https://docs.rs/tonic/0.10.2/tonic/transport/server/struct.Router.html

Upstreaming a PR there would be awesome.

PR #671, opened 2024-01-02
refactor: use display formatting instead of debug formatting in tracing events

SuperFluffy:
## Summary
Replace all tracing event fields flagged by the `tracing-debug-field` lint with display formatting

## Background
Debug formatting is noisy and impossible to parse in observability. It also leaks language-specific implementation details.

## Changes
- Fix all fields in events flagged by the tracing-debug-field:
- errors are emitted using their `tracing::field::Value` implementation on the `dyn std::error::Error` trait object
- bytes slices (like sha256 hashes) are formatted using a new `astria_telemetry::display::hex` utility
- emitting of big types (like `SignedTransaction`) is removed in favor of emitting its sha256 hash as hex
- make the `custom-lints` job a hard requirement to pass tests


noot (approved):

SuperFluffy:
Using my admin privileges to merge this change into main.

Justification: the PR has approval for the relevant changes from codeowner @noot. The code change that requires approval from @astriaorg/infra makes the `custom-lints` job a hard requirement where before it was a soft requirement.

PR #675, opened 2024-01-04
fix(sequencer): allow blocksync to complete successfully

joroshiba:
## Summary
Allows `deliver_tx` to execute txs when the full block has not previously been executed during a proposal

## Background
During blocksync proposals are not received and txs will not have already been executed. We need to allow for a node to have either missed the proposal, and still execute the block or in basic operation sync.

## Changes
- tracks the hash executed during proposal phase.
- allows txs to execute if begin block doesn't match tracked hash, uses cache otherwise


crates/astria-sequencer/src/app.rs
```diff
@@ -360,19 +374,50 @@ impl App {
...
             return Some(Ok(vec![]));
         }
 
-        self.execution_result.remove(tx_hash)
+        if self.proposal_executed {
+            let tx_hash: [u8; 32] = sha2::Sha256::digest(&tx.tx).into();
+            return self.execution_result.remove(&tx_hash);
+        }
+
+        let Some(signed_tx) = raw::SignedTransaction::decode(&*tx.tx)
```
SuperFluffy, line 396: As you have a return value of `Option` you don't need to use the `let ... else` construct.

Here:
```suggestion
        let signed_tx = raw::SignedTransaction::decode(&*tx.tx)
```
(see comment below for rest)

crates/astria-sequencer/src/app.rs
```diff
@@ -360,19 +374,50 @@ impl App {
...
+                SignedTransaction::try_from_raw(raw_tx)
+                    .map_err(|e| {
+                        debug!(
+                            error = &e as &dyn std::error::Error,
+                            "failed to convert raw signed transaction to native signed transaction"
+                        );
+                        e
+                    })
+                    .ok()
+            })
```
SuperFluffy, line 415:
```suggestion
            })?;
```

crates/astria-sequencer/src/app.rs
```diff
@@ -360,19 +374,50 @@ impl App {
...
+        else {
+            return None;
+        };
```
SuperFluffy, line 418: remove these

crates/astria-sequencer/src/app.rs
```diff
@@ -360,19 +374,50 @@ impl App {
...
+            .and_then(|raw_tx| {
+                SignedTransaction::try_from_raw(raw_tx)
+                    .map_err(|e| {
+                        debug!(
+                            error = &e as &dyn std::error::Error,
+                            "failed to convert raw signed transaction to native signed transaction"
+                        );
+                        e
+                    })
+                    .ok()
```
SuperFluffy, line 414: this entire block looks a bit crazy. Could make a helper function to make this look cleaner:
```rust
fn signed_transaction_from_bytes(bytes: &[u8]) -> anyhow::Result<SignedTransaction> {
    let raw = raw::SignedTransaction::decode(bytes).context("failed to decode protobuf to signed transaction")?;
    let tx = SignedTransaction::try_from_raw(raw).context("failed to transform raw signed transaction to verified type")?;
    Ok(tx)
}
```
Then in the body:
```rust
let signed_tx = match signed_transaction_from_bytes(&tx.tx) {
    Err(e) => {
        debug!(
            error = AsRef::<dyn std::error::Error>::as_ref(&e),
            "failed to decode deliver tx payload to signed transaction; ignoring it",
        );
        return None;.
    }
    Ok(tx) => tx,
};
```

SuperFluffy (approved):
Some suggestions for code cleanup

crates/astria-sequencer/src/app.rs
```diff
@@ -360,19 +374,50 @@ impl App {
     /// Since transaction execution now happens in the proposal phase, results
     /// are cached in the app and returned here during the usual ABCI block execution process.
     ///
+    /// If the proposal was not executed, the transaction will be executed.
```
noot, line 377:
```suggestion
    /// If the tx was not executed during the proposal phase, the transaction will be executed here.
```

crates/astria-sequencer/src/app.rs
```diff
@@ -163,6 +174,7 @@ impl App {
         // clear the cache of transaction execution results
         self.execution_result.clear();
         self.processed_txs = 0;
+        self.proposal_executed = false;
```
noot, line 177: put this in `commit` instead - this function is only called by `prepare_proposal`/`process_proposal`, so if the node was a validator for the previous round but not the current round, then `proposal_executed` will be stuck being set to true

crates/astria-sequencer/src/app.rs
noot: this is copy-pasted from `execute_block_data` so we can just make a function used by both

crates/astria-sequencer/src/app.rs
noot: nvm, you did put this in commit. this is fine then!

noot (approved):
approving since there are only minor comments!

crates/astria-sequencer/src/app.rs
```diff
@@ -89,6 +92,16 @@ pub(crate) struct App {
     // `prepare_proposal`, and re-executing them would cause failure.
     is_proposer: bool,
 
+    // This is set to the executed hash of the proposal during `process_proposal`
+    //
+    // If it does not match the hash given during begin_block, then we clear and
+    // reset the execution results cache + state delta. Transactions are reexecuted.
+    // If it does match we utilize cached results to reduce computation.
+    //
+    // Resets to null hash at the beggining of prepare_proposal, and process_proposal if
```
noot, line 101:
```suggestion
    // Resets to the default hash at the beginning of prepare_proposal, and process_proposal if
```

noot: probably want to specify what the default hash is

crates/astria-sequencer/src/app.rs
```diff
@@ -330,10 +334,17 @@ impl App {
     pub(crate) async fn begin_block(
         &mut self,
         begin_block: &abci::request::BeginBlock,
+        storage: Storage,
     ) -> anyhow::Result<Vec<abci::Event>> {
         // clear the processed_txs count when beginning block execution
         self.processed_txs = 0;
 
+        // If we previously executed a different proposal than is being processed reset cached
```
noot, line 342:
```suggestion
        // If we previously executed txs in a different proposal than is being processed, reset cached
```

crates/astria-sequencer/src/app.rs
```diff
@@ -362,21 +373,40 @@ impl App {
...
+        let signed_tx = match signed_transaction_from_bytes(&tx.tx) {
+            Err(e) => {
+                debug!(
+                    error = AsRef::<dyn std::error::Error>::as_ref(&e),
+                    "failed to decode deliver tx payload to signed transaction; ignoring it",
+                );
+                return None;
+            }
+            Ok(tx) => tx,
+        };
```
noot, line 407:
```suggestion
        let signed_tx = signed_transaction_from_bytes(&tx.tx).map_err(|e|
                debug!(
                    error = AsRef::<dyn std::error::Error>::as_ref(&e),
                    "failed to decode deliver tx payload to signed transaction; ignoring it",
                );
                None
            )?;
```

debating whether this should error or not - we should never get an un-deserializable tx

crates/astria-sequencer/src/app.rs
```diff
@@ -672,7 +712,7 @@ mod test {
 
     #[tokio::test]
     async fn app_genesis_and_init_chain() {
-        let app = initialize_app(None, vec![]).await;
+        let (app, _storage) = initialize_app(None, vec![]).await;
```
noot, line 715: make an `initialize_app` and `initiialize_app_with_storage` where `initiialize_app_with_storage` returns the storage but `initialize_app` doesn't (it can just call `initiialize_app_with_storage` and ignore the storage) 

then you can use just `initialize_app` where the storages are ignored

crates/astria-sequencer/src/app.rs
noot: i guess returning `None` hits the `expect` in the caller, so that's fine

noot (approved):
nice, i think the logic looks good, only smaller comments!

crates/astria-sequencer/src/app.rs
```diff
@@ -115,6 +128,7 @@ impl App {
         Self {
             state,
             is_proposer: false,
+            executed_proposal_hash: Hash::default(),
```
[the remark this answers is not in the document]
joroshiba: It is None, and it is also `empty`

crates/astria-sequencer/src/app.rs
joroshiba: I don't think it really matters what the default hash is? It's default, anything more is client library implementation detail?

crates/astria-sequencer/src/app.rs
joroshiba: I don't think this behavior should be changed in this PR, maintaining the previous behavior

joroshiba: Yeah this maintains parity with what happens in the return above if there is an issue with execution.

Issue #524, opened 2023-10-19
sequencer: determine max block size


noot:
we need to determine the maximum block size for sequencer blocks, specifically for `sequence::Action`s (as only those are written to DA). factors to take into account:
- celestia block size 
- celestia blob size
- compression
- computation cost for sequencer nodes and relayer nodes

joroshiba:
676 sets this at 256kb, aligned with the current max transaction size. Easily changed via a constant in the future.

PR #676, opened 2024-01-05
feat(sequencer)!: max sequencer bytes per block limit

joroshiba:
## Summary
Adds a limit to the amount of sequenced data which is allowed into a single block, important to avoid overflowing celestia + ensuring that we can propogate full blocks successfully.

## Background
We had a limit on transaction sizes but not on aggregate of transactions, this adds a basic limit of 256kb for total sequenced data. This size was chosen based on Celestia block size with max of ~5-6 blocks per celestia block completely filled blocks would use up ~60% of 2mb Celestia block size.

## Changes
- Adds a limitation to execute block transactions to limit the number of bytes that make it into full block.

## Breaking Changes
- This a network upgrade, all future proposals will be rejected if they have too many txs. However, old blocks will still be successfully executed through blocksync.

## Issues
closes https://github.com/astriaorg/astria/issues/524

### DIFF
--- a/crates/astria-sequencer/src/app.rs
+++ b/crates/astria-sequencer/src/app.rs
@@ -68,6 +68,9 @@ use crate::{
 /// The inter-block state being written to by the application.
 type InterBlockState = Arc<StateDelta<Snapshot>>;
 
+/// The maximum number of bytes allowed in sequencer block txs.
+const MAX_BLOCK_SIZE: usize = 256_000;
+
 /// The Sequencer application, written as a bundle of [`Component`]s.
 ///
 /// Note: this is called `App` because this is a Tendermint ABCI application,
@@ -267,7 +270,20 @@ impl App {
         let mut signed_txs = Vec::with_capacity(txs.len());
         let mut validated_txs = Vec::with_capacity(txs.len());
 
+        let mut total_tx_size: usize = 0;
+
         for tx in txs {
+            let tx_size = tx.len();
+            // Don't include tx if it would make the block too large.
+
+            if total_tx_size + tx_size > MAX_BLOCK_SIZE {
+                debug!(
+                    total_tx_size,
+                    tx_size, "block size limit reached, not including transaction in block"
+                );
+                continue;
+            }
+
             let Some(signed_tx) = raw::SignedTransaction::decode(&*tx)
                 .map_err(|e| {
                     debug!(
@@ -300,6 +316,7 @@ impl App {
                     self.execution_result.insert(tx_hash.into(), Ok(events));
                     signed_txs.push(signed_tx);
                     validated_txs.push(tx);
+                    total_tx_size += tx_size;
                 }
                 Err(e) => {
                     debug!(
--- a/crates/astria-sequencer/src/app.rs
+++ b/crates/astria-sequencer/src/app.rs
@@ -14,6 +14,7 @@ use anyhow::{
 use astria_core::{
     generated::sequencer::v1alpha1 as raw,
     sequencer::v1alpha1::{
+        transaction::Action,
         Address,
         SignedTransaction,
     },
@@ -69,7 +70,7 @@ use crate::{
 type InterBlockState = Arc<StateDelta<Snapshot>>;
 
 /// The maximum number of bytes allowed in sequencer action data.
-const MAX_BLOCK_SEQUENCE_DATA_SIZE: usize = 256_000;
+const MAX_SEQUENCE_DATA_BYTES_PER_BLOCK: usize = 256_000;
 
 /// The Sequencer application, written as a bundle of [`Component`]s.
 ///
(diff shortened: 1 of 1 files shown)


crates/astria-sequencer/src/app.rs
```diff
@@ -68,6 +68,9 @@ use crate::{
 /// The inter-block state being written to by the application.
 type InterBlockState = Arc<StateDelta<Snapshot>>;
 
+/// The maximum number of bytes allowed in sequencer action data.
+const MAX_BLOCK_SEQUENCE_DATA_SIZE: usize = 256_000;
```
SuperFluffy, line 72: To make it clear what we are talking about.
```suggestion
const MAX_BLOCK_SEQUENCE_BYTES: usize = 256_000;
```
Alternatively `_LEN` because you construct this from the length of byte buffers (just going off of `Vec::len`).

Bikeshed: I wonder if we should move away from calling this `sequence` and just start calling it `rollup_data` or something like that because that's much clearer.

crates/astria-sequencer/src/app.rs
```diff
@@ -266,6 +269,7 @@ impl App {
     ) -> (Vec<SignedTransaction>, Vec<bytes::Bytes>) {
         let mut signed_txs = Vec::with_capacity(txs.len());
         let mut validated_txs = Vec::with_capacity(txs.len());
+        let mut total_sequenced_data_size: usize = 0;
```
SuperFluffy, line 272: Nit: I think I would prefer calling this either `total_sequenced_num_bytes` or `total_sequenced_data_len` just because you construct this from lengths.

crates/astria-sequencer/src/app.rs
```diff
@@ -293,13 +297,35 @@ impl App {
...
+            // Don't include tx if it would make the sequenced block data too large.
+            if total_sequenced_data_size + sequenced_data_size > MAX_BLOCK_SEQUENCE_DATA_SIZE {
+                debug!(
+                    transaction_hash = %telemetry::display::hex(&tx_hash),
+                    included_data_bytes = total_sequenced_data_size,
+                    tx_data_bytes = sequenced_data_size,
+                    max_data_bytes = MAX_BLOCK_SEQUENCE_DATA_SIZE,
+                    "excluding transaction: max block sequenced data limit reached"
+                );
+                continue;
```
SuperFluffy, line 319: I first wanted to suggest `break;`ing here, but in fact this just drops signed txes that don't fit the rest, but attempts to fill up the buffer as much as possible, correct?

crates/astria-sequencer/src/app.rs
```diff
@@ -293,13 +297,35 @@ impl App {
                 continue;
             };
 
-            // store transaction execution result, indexed by tx hash
             let tx_hash = Sha256::digest(&tx);
+
+            // Sum up the total bytes in sequence actions
+            let mut sequenced_data_size = 0;
+            for action in signed_tx.clone().into_unsigned().actions {
```
SuperFluffy, line 304: No need to clone here:
```suggestion
            for action in &signed_tx.unsigned_transaction().actions {
```
If you want to be fancy:

```
let total_len = signed_tx
    .unsigned_transaction()
    .iter()
    .filter_map(UnsignedTransaction::as_sequence)
    .fold(0usize, |acc, seq| acc + seq.data.len());
```

crates/astria-sequencer/src/app.rs
```diff
@@ -293,13 +297,35 @@ impl App {
...
+                if let Some(sequence_action) = action.as_sequence() {
+                    sequenced_data_size += sequence_action.data.len();
+                }
+            }
+
+            // Don't include tx if it would make the sequenced block data too large.
+            if total_sequenced_data_size + sequenced_data_size > MAX_BLOCK_SEQUENCE_DATA_SIZE {
+                debug!(
```
SuperFluffy, line 312: I wonder if this should be flagged more prominently. Maybe an `info!` level that informs you of how many txes were dropped?

SuperFluffy (approved):
Looks good, just some minor nits.

crates/astria-sequencer/src/app.rs
joroshiba: Correct, if there is a follow on smaller tx data that can fit we should grab it.

We are explicitly not going into the whole napsack problem, but a single too big tx shouldn't stop us fitting more.

crates/astria-sequencer/src/app.rs
joroshiba: This is probably a place we want metrics, I can add a counter of some sort and log for now.

This was done as a debug to give as much information as should be needed about why tx not included.

noot (approved):
approving but can you change the title and PR description before merging, as this isn't a block size limit, but a limit on the amount of sequenced data, which is different. can you also add a quick explainer to the description of how this number was chosen, w.r.t celestia?

also, do we care about having an actual block size limit (ie. cap the entire data size), or just what's posted to celestia?

crates/astria-sequencer/src/app.rs
joroshiba: discussed offline and updated name.

Will save the bike shedding on sequencer vs rollup data bikeshed for later date. This is correct choice given current naming.

joroshiba:
> approving but can you change the title and PR description before merging, as this isn't a block size limit, but a limit on the amount of sequenced data, which is different. can you also add a quick explainer to the description of how this number was chosen, w.r.t celestia?
> 
> also, do we care about having an actual block size limit (ie. cap the entire data size), or just what's posted to celestia?

Done.

We probably still need some sort of block size limit (ie we can't have 20k transfer actions in a block). I'm not sure how we want to implement that though, it could be a gas limit as opposed to actual block size limit potentially?

PR #677, opened 2024-01-05
chore(sequencer): update instrumentation for all consensus & app functions

joroshiba:
## Summary
Updates all consensus API calls to have specifically chosen data in the spans, and omits anything but context of function from children (ie handle has context, the funciton it calls doesn't need it)

## Background
Debugging sequencer is hard, spans from handle request on process proposal were unreadable. Similar changes made in sequencer for debugging fix. 

## Changes
- Update all functions to skip_all with specific fields logged

### DIFF
--- a/crates/astria-sequencer/src/app.rs
+++ b/crates/astria-sequencer/src/app.rs
@@ -120,7 +120,7 @@ impl App {
         }
     }
 
-    #[instrument(name = "App:init_chain", skip(self))]
+    #[instrument(name = "App:init_chain", skip_all)]
     pub(crate) async fn init_chain(
         &mut self,
         genesis_state: GenesisState,
@@ -161,7 +161,7 @@ impl App {
     /// It puts this special "commitment" as the first transaction in a block.
     /// When other validators receive the block, they know the first transaction is
     /// supposed to be the commitment, and verifies that is it correct.
-    #[instrument(name = "App::prepare_proposal", skip(self, prepare_proposal))]
+    #[instrument(name = "App::prepare_proposal", skip_all)]
     pub(crate) async fn prepare_proposal(
         &mut self,
         prepare_proposal: abci::request::PrepareProposal,
@@ -185,7 +185,7 @@ impl App {
     /// Generates a commitment to the `sequence::Actions` in the block's transactions
     /// and ensures it matches the commitment created by the proposer, which
     /// should be the first transaction in the block.
-    #[instrument(name = "App::process_proposal", skip(self, process_proposal))]
+    #[instrument(name = "App::process_proposal", skip_all)]
     pub(crate) async fn process_proposal(
         &mut self,
         process_proposal: abci::request::ProcessProposal,
@@ -259,7 +259,9 @@ impl App {
     ///
     /// Returns the transactions which were successfully decoded and executed
     /// in both their [`SignedTransaction`] and raw bytes form.
-    #[instrument(name = "App::execute_block_data", skip(self, txs))]
+    #[instrument(name = "App::execute_block_data", skip_all, fields(
+        tx_count = txs.len()
+    ))]
     async fn execute_block_data(
         &mut self,
         txs: Vec<bytes::Bytes>,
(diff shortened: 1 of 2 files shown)


crates/astria-sequencer/src/app.rs
```diff
@@ -259,7 +259,9 @@ impl App {
     ///
     /// Returns the transactions which were successfully decoded and executed
     /// in both their [`SignedTransaction`] and raw bytes form.
-    #[instrument(name = "App::execute_block_data", skip(self, txs))]
+    #[instrument(name = "App::execute_block_data", skip_all, fields(
+        tx_count = txs.len()
```
SuperFluffy, line 263: Nit, I prefer to not abbreviate things.
```suggestion
        transaction_count = txs.len()
```

crates/astria-sequencer/src/app.rs
```diff
@@ -353,7 +355,9 @@ impl App {
     ///
     /// Note that the first two "transactions" in the block, which are the proposer-generated
     /// commitments, are ignored.
-    #[instrument(name = "App::deliver_tx_after_execution", skip(self))]
+    #[instrument(name = "App::deliver_tx_after_execution", skip_all, fields(
+        tx_hash =  %telemetry::display::hex(tx_hash),
```
SuperFluffy, line 359: Nit: same here, I'd prefer to not abbreviate.

crates/astria-sequencer/src/service/consensus.rs
```diff
@@ -151,23 +159,39 @@ impl Consensus {
...
+    ))]
     async fn handle_prepare_proposal(
         &mut self,
         prepare_proposal: request::PrepareProposal,
     ) -> response::PrepareProposal {
         self.app.prepare_proposal(prepare_proposal).await
     }
 
-    #[instrument(skip(self))]
+    #[instrument(skip_all, fields(
```
SuperFluffy, line 174: This might a be a few too many fields in the span. On the other hand these all seem useful. Not sure

crates/astria-sequencer/src/service/consensus.rs
```diff
@@ -151,23 +159,39 @@ impl Consensus {
...
+        height = %process_proposal.height,
+        time = %process_proposal.time,
+        tx_count = process_proposal.txs.len(),
+        proposer = %process_proposal.proposer_address,
+        hash = %telemetry::display::hex(&process_proposal.hash),
+        next_validators_hash = %process_proposal.next_validators_hash,
```
SuperFluffy, line 180: We should probably decide whether we want to use `tendermint::Hash: Display` or our `telemetry::display::hex` (for both `next_validators_hash` and `hash`).

Note that their `Display` implementation:
1. is upper hex, and
2. always allocates because they rely on the `hex` crate.

I personally try to avoid unnecessary allocations. so `telemetry::display::hex` as much as possible (and consistently lower-cased) is my preference.

crates/astria-sequencer/src/service/consensus.rs
```diff
@@ -182,7 +206,9 @@ impl Consensus {
         })
     }
 
-    #[instrument(skip(self))]
+    #[instrument(skip_all, fields(
+        tx_hash = %telemetry::display::hex(&Sha256::digest(&deliver_tx.tx))
```
SuperFluffy, line 210: Nit: see my other comment re abbreviation. It seems like `tx` is widely used in cometbft, so maybe it makes sense to use it here.

SuperFluffy (approved):
Thank you for doing the work and going through these!

I only have a few nits that you can mostly ignore (maybe except the note on using `telemetry::display::hex` vs the `Display` impl for `tendermint::Hash`).

As a side note (not introduced by this PR so need not be done here): we probably want to think about how best to name our spans. If the `instrument` macro is provided sans a `name` field, the name of its function is taken.

 I find the names of the explicitly named functions too rusty (like `App::deliver_tx_after_execution`).

noot (approved):
great, this looks a lot better. thanks!!

crates/astria-sequencer/src/service/consensus.rs
joroshiba: It's large but I think it's all useful if something goes wrong here, and it should all be relatively small amounts of data. A bunch of hashes.

crates/astria-sequencer/src/app.rs
joroshiba: I would argue this isn't an abbreviation at least not by us, it's the name of it: Blocks have data which is an array of txs, this is executing those. https://docs.cometbft.com/main/spec/core/data_structures#data

PR #678, opened 2024-01-08
refactor(conductor): use idiomatic types, streamline some executor code

SuperFluffy:
## Summary
Defined idiomatic execution types in `astria-core`, refactored and streamlined executor code.

## Background
The conversion method for commitment states would panic, it returns an error now and the conversion happens in `astria-core`, similar to the sequencer types.

There was also a lot of unused code in the execution service extension trait, and unnecessary redirection when updating commitment states (single-use functions that make reading the code opaque). All of that has been removed. 

## Changes
- Defined `Block` and `CommitmentState` in `astria-core` with conversion methods from the raw generated prost types.
- Changed the exection service extension crate to a newtype wrapper on `ExecutionServiceClient<Channel>` (it used to be that, but the newtype wrapper fits better in this instance)
- Removed a few methods by inlining their logic at the call site.

## Testing
Executor tests still run, only refactoring.

### DIFF
--- a/crates/astria-conductor/src/executor/client.rs
+++ b/crates/astria-conductor/src/executor/client.rs
@@ -1,71 +1,32 @@
-use astria_core::generated::execution::v1alpha2::{
-    execution_service_client::ExecutionServiceClient,
-    BatchGetBlocksRequest,
-    BatchGetBlocksResponse,
-    Block,
-    BlockIdentifier,
-    CommitmentState,
-    ExecuteBlockRequest,
-    GetBlockRequest,
-    GetCommitmentStateRequest,
-    UpdateCommitmentStateRequest,
+use astria_core::{
+    execution::v1alpha2::{
+        Block,
+        CommitmentState,
+    },
+    generated::execution::{
+        v1alpha2 as raw,
+        v1alpha2::execution_service_client::ExecutionServiceClient,
+    },
+    Protobuf as _,
 };
 use color_eyre::eyre::{
-    Result,
-    WrapErr,
+    self,
+    WrapErr as _,
 };
 use prost_types::Timestamp;
 use tonic::transport::Channel;
 
-use super::ExecutorCommitmentState;
-
-/// Extension trait for the ExecutionServiceClient that makes it easier to interact with the
-/// execution client.
-#[async_trait::async_trait]
-pub(super) trait ExecutionClientExt {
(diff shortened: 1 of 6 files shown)


crates/astria-conductor/src/executor/client.rs
```diff
@@ -1,154 +1,98 @@
...
 };
 use prost_types::Timestamp;
 use tonic::transport::Channel;
 
-use super::ExecutorCommitmentState;
-
-/// Extension trait for the ExecutionServiceClient that makes it easier to interact with the
-/// execution client.
-#[async_trait::async_trait]
-pub(super) trait ExecutionClientExt {
```
noot, line 25: what's the reasoning for changing this to a wrapper type instead of an extension trait?

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -391,9 +360,9 @@ impl Executor {
                     .execute_block(block_subset)
                     .await
                     .wrap_err("failed to execute block")?;
-                self.update_soft_commitment(executed_block)
+                self.update_commitment_state(self.commitment_state.firm().clone(), executed_block)
```
noot, line 363: i preferred the previous `update_soft_commitment`/`update_firm_commitment` for readability

crates/astria-conductor/src/executor/client.rs
SuperFluffy: I started with the extension trait, but you cannot name the trait methods the same as the inherent methods (`execute_block‘ etc... You actually could but it's hecking annoying to use).

So the only way around this is a newtype wrapper.

noot (approved):
going to approve, only have smaller comments

crates/astria-conductor/src/executor/mod.rs
SuperFluffy: I think you might be right.

SuperFluffy: I changed the signature to
```rust
fn update_commitment_state(&mut self, update: Update) {}

enum Update {
    OnlySoft(Block),
    OnlyFirm(Block),
    ToSame(Block),
}
```

This means that calls will now look like:
```rust
self.update_commitment_state(Update::OnlyFirm(executed_block));
```
This should make it even clearer what's intended.

PR #683, opened 2024-01-15
feat(sequencer, cli)!: update balance queries to return every asset owned by account

### LINKED ISSUE #682: sequencer: update balance query to return balance for every denom
as title says - only supports the native asset right now, but the balance query should return every asset owned by that account and their respective denoms/IDs.

noot:
## Summary
update balance queries to return every asset owned by account.

## Background
since we support multiple assets on the sequencer via IBC, we need to be able to check what they are.

## Changes
- implement `get_account_balances` which returns every (asset, balance) owned by the account.
- update ABCI balance query to support this
- update proto `BalanceResponse` to handle vec of (asset, balance)
- consolidate `Denom` and `IbcAsset` into one `Denom` type that handles IBC and non IBC assets. this is needed for multiple reasons: makes handling between native/non-native easier, and when an asset is received via IBC, it is not yet prefixed, so we couldn't previously have converted it into a `IbcAsset`.

## Testing

cli testing (with IBC), eg:
```
./target/debug/astria-cli sequencer balance get 1c0c490f1b5528d8173c5de46d131160e4b2c0c3 --sequencer-url=http://localhost:26657
```

## Breaking Changelist
- ABCI balance query format is changed.

## Related Issues

closes #682


sambukowski (approved):

crates/astria-sequencer/src/accounts/state_ext.rs
```diff
@@ -57,6 +59,56 @@ fn channel_balance_storage_key(channel: &ChannelId, asset: asset::Id) -> String
...
 pub(crate) trait StateReadExt: StateRead {
+    #[instrument(skip(self))]
+    async fn get_account_balances(&self, address: Address) -> Result<Vec<AssetBalance>> {
+        use crate::asset::state_ext::StateReadExt as _;
+
+        let prefix = format!("{}/balance/", storage_key(&address.encode_hex::<String>()));
+        let mut balances: Vec<AssetBalance> = Vec::new();
+
+        let mut stream = std::pin::pin!(self.prefix_keys(&prefix));
+        while let Some(Ok(key)) = stream.next().await {
```
SuperFluffy, line 70: Q about the entire body of the loop: should this entire block fail if some account balances could not be retrieved? Or is it OK to log these and carry on?

SuperFluffy:
Looks fine to me. Only have a question about the `get_account_balances` loop if hard failing is necessary or if it's fine to just skip the failed fields.

I have marked the PR as breaking (the `!` in `feat(...)!`)

SuperFluffy (approved):

crates/astria-sequencer/src/accounts/state_ext.rs
noot: I think it's fine as this is just used in the CLI right now

PR #685, opened 2024-01-16
feat(conductor, proto)!: essential rollup info from genesis

joroshiba:
## Summary
Adds a `GenesisInfo` rpc to the execution API to pass definitional rollup information through rollup genesis instead of conductor config, as well as implementing conductor to implement this.

## Background
Certain information about the rollup should be considered a core part of the rollup definition, but was done through environmental variable configuration making it harder to spin up nodes that follow the same rules.

## Changes
- Create `GenesisInfo` protos
- Added types to astria-core
- Replaced information from config. 

## Testing
Added basic CI tests. Manual testing TBD

## Breaking Changelist
- Changes conductor config and requires `GenesisInfo` be implemented on rollup.

### DIFF
--- a/crates/astria-conductor/local.env.example
+++ b/crates/astria-conductor/local.env.example
@@ -10,10 +10,6 @@ ASTRIA_CONDUCTOR_CELESTIA_BEARER_TOKEN="<JWT Bearer token>"
 # This url is used to read astria blocks from the Data Availability layer
 ASTRIA_CONDUCTOR_CELESTIA_NODE_URL="http://127.0.0.1:26659"
 
-# The chain id of the chain that is being read from the astria-sequencer or the
-# Data Availability layer
-ASTRIA_CONDUCTOR_CHAIN_ID="astriachain"
-
 # Execution RPC URL
 ASTRIA_CONDUCTOR_EXECUTION_RPC_URL="http://127.0.0.1:50051"
 
@@ -29,9 +25,6 @@ ASTRIA_CONDUCTOR_EXECUTION_COMMIT_LEVEL="SoftAndFirm"
 # retrieving validators.
 # 127.0.0.1:26657 is the default socket address in comebft's `rpc.laddr` setting.
 ASTRIA_CONDUCTOR_SEQUENCER_URL="ws://127.0.0.1:26657/websocket"
-
-# the sequencer block height that the rollup's first block was in
-ASTRIA_CONDUCTOR_INITIAL_SEQUENCER_BLOCK_HEIGHT=1
 
 # set to true to enable op-stack deposit derivations
 ASTRIA_CONDUCTOR_ENABLE_OPTIMISM=false
--- a/crates/astria-conductor/src/conductor.rs
+++ b/crates/astria-conductor/src/conductor.rs
@@ -112,7 +112,7 @@ impl Conductor {
                 .build()
                 .await
                 .wrap_err("failed to construct executor")?;
-            let genesis = executor.get_genesis();
+            let genesis = executor.genesis();
             let executable_sequencer_block_height = executor
                 .calculate_executable_block_height()
                 .wrap_err("failed calculating the next executable block height")?;
--- a/crates/astria-conductor/src/conductor.rs
+++ b/crates/astria-conductor/src/conductor.rs
@@ -5,7 +5,6 @@ use std::{
     time::Duration,
 };
 
-use astria_core::sequencer::v1alpha1::RollupId;
 use celestia_client::celestia_types::nmt::Namespace;
 use color_eyre::eyre::{
     self,
@@ -96,10 +95,8 @@ impl Conductor {
 
         let signals = spawn_signal_handler();
 
-        let rollup_id = RollupId::from_unhashed_bytes(&cfg.chain_id);
-
         // Spawn the executor task.
-        let (executor_tx, sync_start_block_height) = {
+        let (executor_tx, genesis, sync_start_block_height) = {
             let (block_tx, block_rx) = mpsc::unbounded_channel();
             let (shutdown_tx, shutdown_rx) = oneshot::channel();
 
@@ -109,14 +106,13 @@ impl Conductor {
 
             let executor = Executor::builder()
                 .rollup_address(&cfg.execution_rpc_url)
-                .rollup_id(rollup_id)
-                .sequencer_height_with_first_rollup_block(cfg.initial_sequencer_block_height)
                 .block_channel(block_rx)
                 .shutdown(shutdown_rx)
                 .set_optimism_hook(hook)
                 .build()
                 .await
                 .wrap_err("failed to construct executor")?;
+            let genesis = executor.get_genesis();
             let executable_sequencer_block_height = executor
                 .calculate_executable_block_height()
                 .wrap_err("failed calculating the next executable block height")?;
@@ -124,7 +120,7 @@ impl Conductor {
             tasks.spawn(Self::EXECUTOR, executor.run_until_stopped());
             shutdown_channels.insert(Self::EXECUTOR, shutdown_tx);
 
-            (block_tx, executable_sequencer_block_height)
+            (block_tx, genesis, executable_sequencer_block_height)
         };
 
         let sequencer_client_pool = client_provider::start_pool(&cfg.sequencer_url)
@@ -200,7 +196,7 @@ impl Conductor {
                 executor_tx.clone(),
                 sequencer_client_pool.clone(),
                 sequencer_namespace,
-                celestia_client::celestia_namespace_v0_from_rollup_id(rollup_id),
+                celestia_client::celestia_namespace_v0_from_rollup_id(genesis.rollup_id()),
                 shutdown_rx,
             )
             .await
--- a/crates/astria-conductor/src/executor/mod.rs
+++ b/crates/astria-conductor/src/executor/mod.rs
@@ -446,13 +446,13 @@ impl Executor {
         &self,
     ) -> eyre::Result<tendermint::block::Height> {
         let Some(executable_block_height) = calculate_sequencer_block_height(
-            self.genesis.sequencer_genesis_block_number(),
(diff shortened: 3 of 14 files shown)


crates/astria-conductor/src/executor/mod.rs
```diff
@@ -501,32 +446,36 @@ impl Executor {
...
                 self.commitment_state.soft().number(),
             );
         };
 
         Ok(executable_block_height.into())
     }
 
+    pub(crate) fn get_genesis(&self) -> GenesisInfo {
+        self.genesis.clone()
+    }
```
noot, line 465: it appears only the rollup_id is used from this - maybe just make a `get_rollup_id` instead?

proto/astria/execution/v1alpha2/execution.proto
```diff
@@ -4,6 +4,17 @@ package astria.execution.v1alpha2;
...
 
+// GenesisInfo contains the information needed to start a rollup chain.
+//
+// This information is used to determine which sequencer & celestia data to
+// use from the Astria & Celestia networks.
+message GenesisInfo {
+  bytes rollup_id = 1;
+  uint32 sequencer_genesis_block_number = 2;
+  uint32 celestia_base_block_number = 3;
+  uint32 celestia_block_variance = 4;
```
SuperFluffy, line 15: I think I'd leave this out for now and go with only the naive height.

proto/astria/execution/v1alpha2/execution.proto
```diff
@@ -4,6 +4,17 @@ package astria.execution.v1alpha2;
...
 import "google/protobuf/timestamp.proto";
 
+// GenesisInfo contains the information needed to start a rollup chain.
+//
+// This information is used to determine which sequencer & celestia data to
+// use from the Astria & Celestia networks.
+message GenesisInfo {
+  bytes rollup_id = 1;
+  uint32 sequencer_genesis_block_number = 2;
+  uint32 celestia_base_block_number = 3;
```
SuperFluffy, line 14: In cometbft these are called the heights, not numbers. I'd prefer their convention.

crates/astria-core/src/execution/v1alpha2/mod.rs
```diff
@@ -2,9 +2,107 @@ use prost_types::Timestamp;
...
 use crate::{
     generated::execution::v1alpha2 as raw,
+    sequencer::v1alpha1::{
+        IncorrectRollupIdLength,
+        RollupId,
+    },
     Protobuf,
 };
 
+// An error when transforming a [`raw::Block`] into a [`Block`].
```
SuperFluffy, line 12:
```suggestion
/// An error when transforming a [`raw::Block`] into a [`Block`].
```
to make this into a doccomment

crates/astria-core/src/execution/v1alpha2/mod.rs
```diff
@@ -2,9 +2,107 @@ use prost_types::Timestamp;
...
+/// its parent block's hash, and timestamp.
+///
+/// Usually constructed its [`Protobuf`] implementation from a
+/// [`raw::Block`].
+#[derive(Clone, Debug)]
+pub struct GenesisInfo {
+    /// The block number
+    rollup_id: RollupId,
+    /// The hash of the block
+    sequencer_genesis_block_number: u32,
```
SuperFluffy, line 41: This should ideally be a `tendermint::block::Height`.

crates/astria-core/src/execution/v1alpha2/mod.rs
```diff
@@ -2,9 +2,107 @@ use prost_types::Timestamp;
...
+    celestia_base_block_number: u32,
```
SuperFluffy, line 42: And this as a `celestia_tendermint::block::Height`. Check our `astria-celestia-client` for the exact module.

crates/astria-core/src/execution/v1alpha2/mod.rs
```diff
@@ -2,9 +2,107 @@ use prost_types::Timestamp;
...
+}
+
+/// An Astria execution block on a rollup.
+///
+/// Contains information about the block number, its hash,
+/// its parent block's hash, and timestamp.
+///
+/// Usually constructed its [`Protobuf`] implementation from a
+/// [`raw::Block`].
+#[derive(Clone, Debug)]
```
SuperFluffy, line 36: This can be `Copy`

crates/astria-conductor/src/executor/mod.rs
SuperFluffy, line 465:
```suggestion
    pub(crate) fn genesis(&self) -> &GenesisInfo {
        self.genesis
    }
```
accessors are conventionally named without `get_*` unless there is also a setter to avoid confusion.

Clone's should also be done explicitly by the caller of the method, not the method itself.

But either way - `GenesisInfo` can be `Copy`, so the cloning is not necessary

crates/astria-core/src/execution/v1alpha2/mod.rs
```diff
@@ -2,9 +2,107 @@ use prost_types::Timestamp;
 
 use crate::{
     generated::execution::v1alpha2 as raw,
+    sequencer::v1alpha1::{
+        IncorrectRollupIdLength,
+        RollupId,
```
SuperFluffy, line 7: side remark: we should probably move these out of sequencer into a `common` module

crates/astria-core/src/execution/v1alpha2/mod.rs
```diff
@@ -2,9 +2,107 @@ use prost_types::Timestamp;
...
+            sequencer_genesis_block_number,
+            celestia_base_block_number,
+            celestia_block_variance,
+        } = raw;
+        let rollup_id =
+            RollupId::try_from_slice(rollup_id).map_err(Self::Error::incorrect_rollup_id_length)?;
+
+        Ok(Self {
+            rollup_id,
+            sequencer_genesis_block_number: *sequencer_genesis_block_number,
```
SuperFluffy, line 84: I wonder if these should guard against being `0` because the rollup being included in either sequencer or celestia genesis would be very strange.

crates/astria-conductor/local.env.example
SuperFluffy, line 1: getting rid of these is awesome

crates/astria-conductor/src/conductor.rs
```diff
@@ -109,22 +106,21 @@ impl Conductor {
...
+            let genesis = executor.get_genesis();
             let executable_sequencer_block_height = executor
                 .calculate_executable_block_height()
                 .wrap_err("failed calculating the next executable block height")?;
 
             tasks.spawn(Self::EXECUTOR, executor.run_until_stopped());
             shutdown_channels.insert(Self::EXECUTOR, shutdown_tx);
 
-            (block_tx, executable_sequencer_block_height)
+            (block_tx, genesis, executable_sequencer_block_height)
```
SuperFluffy, line 123: side remark: i am switching around the logic in my refactor so that this block just returns the executor and the other tasks can be spun up by using e.g. `celestia_namespace_v0_from_rollup_id(executor.genesis_info().rollup_id())` because the executor task is launched later.

crates/astria-core/src/execution/v1alpha2/mod.rs
```diff
@@ -2,9 +2,107 @@ use prost_types::Timestamp;
...
+/// An Astria execution block on a rollup.
+///
+/// Contains information about the block number, its hash,
+/// its parent block's hash, and timestamp.
+///
+/// Usually constructed its [`Protobuf`] implementation from a
+/// [`raw::Block`].
+#[derive(Clone, Debug)]
+pub struct GenesisInfo {
+    /// The block number
```
noot, line 38: remove/update comment

crates/astria-core/src/execution/v1alpha2/mod.rs
```diff
@@ -2,9 +2,107 @@ use prost_types::Timestamp;
...
+    rollup_id: RollupId,
+    /// The hash of the block
```
noot, line 40: remove

crates/astria-core/src/execution/v1alpha2/mod.rs
```diff
@@ -2,9 +2,107 @@ use prost_types::Timestamp;
...
+    IncorrectRollupIdLength(IncorrectRollupIdLength),
+}
+
+/// An Astria execution block on a rollup.
+///
+/// Contains information about the block number, its hash,
+/// its parent block's hash, and timestamp.
+///
+/// Usually constructed its [`Protobuf`] implementation from a
+/// [`raw::Block`].
```
noot, line 35: update comment

crates/astria-core/src/execution/v1alpha2/mod.rs
noot, line 12: update comment

proto/astria/execution/v1alpha2/execution.proto
noot, line 14: add comment somewhere for this? i'm guessing it's the first celestia block the rollup has data in?

proto/astria/execution/v1alpha2/execution.proto
noot, line 15: same for this, add comment

noot (approved):
mostly minor comments (clean up, comments, etc), overall i think this is a great change!

crates/astria-core/src/execution/v1alpha2/mod.rs
joroshiba: I understand from our perspective, but that would require rollups import tendermint types, which we don't want to require.

crates/astria-core/src/execution/v1alpha2/mod.rs
joroshiba: Same as above.

sambukowski (approved):

joroshiba:
Closing since it has been rolled into another PR with major conductor changes.

PR #687, opened 2024-01-17
feat(core): add `SignedTransaction::sha256_of_proto_encoding()` method

noot:
## Summary

add transaction hashes to match what cometbft/hermes are logging/use internally.

## Changes
- add `SignedTransaction::sha256_of_proto_encoding()` method

## Testing

unit test + I ran it and cometbft/hermes get the same tx hashes.

## Related Issues

closes https://github.com/astriaorg/astria/issues/476

### DIFF
--- a/crates/astria-core/src/sequencer/v1alpha1/transaction/mod.rs
+++ b/crates/astria-core/src/sequencer/v1alpha1/transaction/mod.rs
@@ -273,3 +273,59 @@ enum UnsignedTransactionErrorKind {
     #[error("`fee_asset_id` field is invalid")]
     FeeAsset(#[source] super::IncorrectAssetIdLength),
 }
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use crate::sequencer::v1alpha1::{
+        asset::{
+            default_native_asset_id,
+            Id,
+        },
+        transaction::action::{
+            Action::Transfer,
+            TransferAction,
+        },
+        Address,
+    };
+
+    #[test]
+    fn signed_transaction_hash() {
+        let verification_key = VerificationKey::try_from([
+            213, 191, 74, 63, 204, 231, 23, 176, 56, 139, 204, 39, 73, 235, 193, 72, 173, 153, 105,
+            178, 63, 69, 238, 27, 96, 95, 213, 135, 120, 87, 106, 196,
+        ])
+        .unwrap();
+        let signature = Signature::try_from([
+            227, 85, 139, 137, 185, 81, 103, 226, 85, 208, 68, 190, 196, 105, 191, 191, 37, 227,
+            167, 21, 69, 165, 229, 163, 187, 104, 165, 40, 92, 8, 113, 67, 166, 194, 232, 156, 232,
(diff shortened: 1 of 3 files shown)


crates/astria-core/src/sequencer/v1alpha1/transaction/mod.rs
```diff
@@ -63,6 +64,17 @@ pub struct SignedTransaction {
 }
 
 impl SignedTransaction {
+    /// Returns the transaction hash.
+    ///
+    /// The transaction hash is calculated by protobuf-encoding the transaction
+    /// and hashing the resulting bytes with sha256.
+    #[must_use]
+    pub fn hash(&self) -> [u8; 32] {
```
SuperFluffy, line 72: We can do this, but please change the name to something more verbose like `sha256_of_proto_encoding`.

The reason is that for an inherent method I expect a hash to a) be in terms of the rust memory representation (like if you `#[derive(Hash)]`), and b) be relatively cheap (just the cost of hashing, not an additional serialization + allocation).

SuperFluffy:
I can see the merit for wanting a shorter method for reporting, but I find the name both to innocent and surprising for what it does.

crates/astria-core/src/sequencer/v1alpha1/transaction/mod.rs
noot: done!

crates/astria-core/src/sequencer/v1alpha1/transaction/mod.rs
```diff
@@ -261,3 +273,59 @@ enum UnsignedTransactionErrorKind {
...
+
+        let transfer = TransferAction {
+            to: Address::from([0; 20]),
+            amount: 0,
+            asset_id: default_native_asset_id(),
+        };
+
+        let unsigned = UnsignedTransaction {
+            nonce: 0,
+            actions: vec![Transfer(transfer)],
```
SuperFluffy, line 319: note that I think `impl From<TransferAction> for Action` so that
```suggestion
            actions: vec![transfer.into()],
```

crates/astria-core/src/sequencer/v1alpha1/transaction/mod.rs
```diff
@@ -261,3 +273,59 @@ enum UnsignedTransactionErrorKind {
...
+#[cfg(test)]
+mod test {
+    use super::*;
+    use crate::sequencer::v1alpha1::{
+        asset::{
+            default_native_asset_id,
+            Id,
+        },
+        transaction::action::{
+            Action::Transfer,
```
SuperFluffy, line 286: think you don't need to import this, but no big deal

crates/astria-core/src/sequencer/v1alpha1/transaction/mod.rs
```diff
@@ -4,6 +4,7 @@ use ed25519_consensus::{
     VerificationKey,
 };
 use prost::Message as _;
+use sha2::Sha256;
```
SuperFluffy, line 7: nit: Would move this into the `use` statements of the method because it's only used there.

SuperFluffy (approved):
looks good, nits only.

PR #688, opened 2024-01-17
fix(ci): change foundry commit in ci

noot:
## Summary
previous foundry commit is no longer available, changed it to one that is.


SuperFluffy (approved):
That is awful that these are being pulled.

jbowen93 (approved):

PR #689, opened 2024-01-18
chore(cli): bump for dusk-3

joroshiba:
## Summary
release new cli version which points to dusk-3

## Background
We are launching a new devnet, and cli should point to the correct rpc by default


jbowen93 (approved):
We also need to bump strings in the docs

noot (approved):

PR #690, opened 2024-01-19
feat(sequencer): transfer fees to block proposer instead of burning

### LINKED ISSUE #649: sequencer: transfer fees to block proposer instead of burning
as title says

noot:
## Summary
accumulate block fees during tx execution and transfer them to the block proposer in `end_block`.

## Background
we don't want to burn tx fees especially not if they are IBC assets.

## Changes
- implement `get_block_fees`, `get_and_increase_block_fees`, and `clear_block_fees` which get, increase, and clear the block fees respectively
- during action execution where there is fee payment (transfer and sequencer), update the current block fees state.
- in `begin_block` store the proposer address (needed b/c `end_block` doesn't receive it, this can be removed when we move to comet 0.38)
- in `end_block` transfer the fees to the proposer and clear the cached address

## Testing
unit test

## Breaking Changelist
changes the state by not burning tokens.

## Related Issues

closes  #649


joroshiba (approved):
LGTM, but I believe this should be considered a breaking change. 

If run over existing blocks a blocksync call would achieve incorrect state, would require a forkchoice rule added in order to added to an existing network.

PR #691, opened 2024-01-19
feat(conductor, core, proto)!: add firm block syncing

SuperFluffy:
## Summary

Allows conductor to sync using firm commitments
read from Celestia. Adds genesis info RPCs to the
execution API. Reworks and refactors syncing in
conductor to be part of normal execution.

## Background

Soft-syncing (meaning bringing up a conductor's
associated rollup to the latest state using blocks
read from the Sequencer) was patched into
conductor as an after-thought. With the
requirement that a rollup also be synced using
firm blocks (read from Celestia) the design needed
to be refactored.

## Changes

The main changes are:

- soft- and firm-syncing are reworked to not be
  separate processes but part of normal operation.
- the Sequencer/Celestia readers forward sequencer
  blocks to the executor in strictly sequential
  order starting from a given Sequencer height.
- the starting Sequencer height is obtained by
  calling the genesis info and commitment state
  RPCs on the Rollup.
- the genesis info RPC is added to the execution
  API.
- the executor now considers the spread between
  firm and soft commitments, ignoring soft blocks
  until firm blocks have caught (currently
  hard-coded to a difference of 16 heights).
- older soft-blocks are simply dropped.
- the Celestia Reader now considers a newly
  introduced concept of "celestia height variance"
  when looking for sequencer blocks: sequencer
  blocks at heights S and S+1 must be found within
  Celestia heights [C, C+V], where V is the
  variance and also obtained from the genesis info
  RPC.
- the commitment state is now broadcast from
  executor to the Sequencer/Celestia readers.
- Sequencer reader uses this information to not
  fetch/forward outdated blocks, but simply drops
  them.
- Celestia reader uses that information to advance
  its permitted Celestia heights (to be within the
  variance mentioned above).

Minor changes:

- rework instrumentation, events
- the executor, Sequencer reader, and Celestia
  reader now all have non-async builders so
  creating them won't block the others from
  advancing.
- signal handling was reworked (only SIGTERM is
  used).
- communication between Sequencer reader and
  executor and Celestia reader and executor is
  now over separate channels to allow for
  different backpressure.
- remove all mention of "data availability" in
  favor of "celestia".
- Celestia reader no longer polls for new blocks
  at fixed intervals but uses a JSONRPC websocket
  subscription to Celestia headers to determine
  if Celestia produces new blocks.
- Celestia reader attempts to resubscribe to new
  Celestia headers if the subscription fails
- Celestia reader attempts to reconnect to the
  HTTP endpoint repeatedly before failing
[...]

PR #693, opened 2024-01-19
refactor(conductor): remove sync-completed signal

SuperFluffy:
## Summary
Removes the sync-completed signal from the sequencer reader.

## Background
In the new implementation of syncing there is no need for reporting the sync status back to conductor because the celestia and sequencer readers can run concurrently.

## Changes
- Remove the sync-completed signal.

## Related Issues

### DIFF
--- a/crates/astria-conductor/src/conductor.rs
+++ b/crates/astria-conductor/src/conductor.rs
@@ -16,7 +16,6 @@ use ethers::prelude::{
     Provider,
     Ws,
 };
-use futures::future::Fuse;
 use tokio::{
     select,
     signal::unix::{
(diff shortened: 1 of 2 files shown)


sambukowski (approved):

noot (approved):
i'm guessing we now don't distinguish between syncing and not in the DA reader?

SuperFluffy:
> i'm guessing we now don't distinguish between syncing and not in the DA reader?

Yes, the distinction is effectively removed. Syncin'ing and normal operation will the the same thing, but implemented in https://github.com/astriaorg/astria/pull/695 and ultimately https://github.com/astriaorg/astria/pull/699

PR #694, opened 2024-01-19
refactor(conductor): use separate channels for celestia, sequencer blocks

SuperFluffy:
## Summary
Splits the legacy "executor command" channel into separate channels for the celestia and sequencer reader sending their blocks to the executor.

## Background
This allows for a cleaner implementation with the celestia/sequencer readers being completely agnostic to each other where before they had to be wrapped in a common enum.

Furthermore it allows separate back pressure on both channels so that frequent sequencer blocks don't flood out celestia blocks (although this is not currently implemented).

## Changes
- Removed the common "executor command" mpsc channel, replace by two separate mpsc channels.

## Testing
No functional changes other than using two different channels.

## Related Issues

### DIFF
--- a/crates/astria-conductor/src/executor/mod.rs
+++ b/crates/astria-conductor/src/executor/mod.rs
@@ -16,17 +16,10 @@ use color_eyre::eyre::{
 };
 use prost_types::Timestamp as ProstTimestamp;
 use sequencer_client::SequencerBlock;
-use tendermint::{
-    Hash,
-    Time,
-};
 use tokio::{
     select,
     sync::{
-        mpsc::{
-            UnboundedReceiver,
-            UnboundedSender,
-        },
+        mpsc,
         oneshot,
     },
 };
@@ -45,13 +38,8 @@ mod client;
 #[cfg(test)]
 mod tests;
 
-/// The channel for sending commands to the executor task.
-pub(crate) type Sender = UnboundedSender<ExecutorCommand>;
-/// The channel the executor task uses to listen for commands.
-pub(crate) type Receiver = UnboundedReceiver<ExecutorCommand>;
-
 // Given `Time`, convert to protobuf timestamp
-fn convert_tendermint_to_prost_timestamp(value: Time) -> ProstTimestamp {
+fn convert_tendermint_to_prost_timestamp(value: tendermint::Time) -> ProstTimestamp {
     use tendermint_proto::google::protobuf::Timestamp as TendermintTimestamp;
     let TendermintTimestamp {
         seconds,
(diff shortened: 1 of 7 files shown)


sambukowski (approved):

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -265,8 +207,10 @@ impl<TBlockChannel, TRollupAddress, TRollupId, TShutdown>
 }
 
 pub(crate) struct Executor {
-    /// Channel on which executor commands are received.
-    block_channel: Receiver,
+    celestia_rx: mpsc::UnboundedReceiver<Vec<SequencerBlockSubset>>,
+    celestia_tx: mpsc::WeakUnboundedSender<Vec<SequencerBlockSubset>>,
```
noot, line 211: why `WeakUnboundedSender` vs `UnboundedSender`?

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -305,6 +249,18 @@ impl Executor {
         ExecutorBuilder::new()
     }
 
+    pub(super) fn celestia_channel(&self) -> mpsc::UnboundedSender<Vec<SequencerBlockSubset>> {
+        self.celestia_tx.upgrade().expect(
+            "should work because the channel is held by self, is open, and other senders exist",
+        )
+    }
```
noot, line 256: can this be called multiple times to return multiple senders?

crates/astria-conductor/src/executor/mod.rs
noot: is it because `Executor` will never send on these tx channels? if so can you add a comment

noot (approved):
approving but have some questions!

crates/astria-conductor/src/executor/mod.rs
SuperFluffy: After rereading https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.WeakUnboundedSender.html I realized that I misunderstood `WeakSender`. I thought upgrading fails if there are no other senders (either weak or full/upgraded), but that actually does not make sense. Upgrading fails if there are no other full senders, so what I implemented here will never work.

Thank you for flagging, I will fix this!

crates/astria-conductor/src/executor/mod.rs
SuperFluffy: Yes. Although I am going to change the `celestia_tx -> UnboundedSender` and the `upgrade -> clone`

crates/astria-conductor/src/executor/mod.rs
SuperFluffy: I have fixed this in PR https://github.com/astriaorg/astria/pull/711

crates/astria-conductor/src/executor/mod.rs
SuperFluffy: Does no longer apply because we don't use weak channels (was buggy; see PR #711)

PR #695, opened 2024-01-19
feat(conductor)!: forward firm blocks read from celestia in sequential order

SuperFluffy:
## Summary
The celestia reader task only forwards blocks to the executor if their heights are in sequential order.

## Background
As part of the changed contract the sequencer and celestia readers forward their blocks to the executor only if they are in sequential order of the sequencer heights recorded inside them.

## Changes
- Introduce a block cache that yields a new block with height H only if it has previously yielded a block with height H-1
- Change the http-based celestia client to websocket
- Remove polling celestia for new blocks at fixed interval times
- Subscribe to new celestia headers, extracting the newest celestia height
- Add a `ReconstructedBlockStream` that acts as a buffered queue fetching celestia blobs at the the height obtained from the header subscription

## Testing
No explicit tests. This must be tested end-to-end. Mock tests should be added in a follow-up.

## Breaking Changelist
- must connect to celestia using websocket instead of http

## Related Issues
Part of https://github.com/astriaorg/astria/pull/691

PR #697, opened 2024-01-19
refactor(conductor): forward blocks from sequencer in sequential order

SuperFluffy:
## Summary
The sequencer reader task only forwards blocks to the executor if their heights are in sequential order.

## Background
This follows the patch for the celestia reader in https://github.com/astriaorg/astria/pull/695 using the `BlockCache` introduced there to only yield new blocks if blocks at the previous height have been seen.

## Changes
- Refactor `sequencer::Reader` to use the `BlockCache` to forward sequencer blocks to the executor in sequential order
- Add a `BlocksFromHeightStream` to fetch sequencer blocks at heights that have been missed
- Remove the explicit sync job as this is now provided by the cache and the stream.

## Testing
This should be tested end-to-end. Mock tests don't exist and should be added in a follow-up.

## Related Issues
Part of #691


crates/astria-conductor/src/sequencer.rs
```diff
@@ -0,0 +1,365 @@
...
+    stream::FuturesUnordered,
+    FutureExt as _,
+    Stream,
+    StreamExt as _,
+};
+use pin_project_lite::pin_project;
+use sequencer_client::{
+    extension_trait::NewBlocksStream,
+    tendermint::block::Height,
+    // NewBlockStreamError,
```
sambukowski, line 33: remove?

sambukowski (approved):

crates/astria-conductor/src/block_cache.rs
```diff
@@ -0,0 +1,220 @@
+//! A cache of sequencer blocks that are only yielded in sequential order.
```
noot, line 1: is this something that can be re-used for celestia blocks?

crates/astria-conductor/src/block_cache.rs
```diff
@@ -0,0 +1,220 @@
...
+
+use pin_project_lite::pin_project;
+use sequencer_client::{
+    tendermint::block::Height,
+    SequencerBlock,
+};
+
+pub(crate) trait GetSequencerHeight {
+    fn get_height(&self) -> Height;
+}
```
noot, line 15: is this only so it can be re-used with multiple block types? but it's only being used with one block type here, right?

crates/astria-conductor/src/sequencer.rs
```diff
@@ -0,0 +1,365 @@
...
+                new_block = subscription.next(), if !subscription.is_done() => {
+                    match new_block {
+                        Some(Ok(block)) => {
+                            let height = block.height();
+                            if let Err(e) = sequential_blocks.insert(block) {
+                                // XXX: we could temporarily kill the subscription if we put an upper limit on the cache size
+                                warn!(error = &e as &dyn std::error::Error, "failed pushing block into cache, dropping it");
+                            } else {
+                                blocks_from_height.advance_height(height);
+                            }
```
noot, line 172: why not have the `BlocksFromHeightStream` advance the height internally?

noot (approved):

crates/astria-conductor/src/block_cache.rs
SuperFluffy: Yes, exactly. The "reconstructed block" type that conductor defines and `SequencerBlock` coming out of astria-core both have heights.

crates/astria-conductor/src/sequencer.rs
SuperFluffy: Because that information is not available within that stream. Fundamentally, `BlocksFromHeightStream` is modelled around a `FuturesUnordered` into which new requests are pushed. The assumption is that if a height is requested, then it is expected to exist. The stream will then eagerly request as many heights as possible.

On the other hand, if one requests a cometbft height that does not exist (yet), cometbft will respond with a 500. So the stream could just autoadvance its heights (or use an infinite loop), but then you'd get a ton of 500s which you need to spin on until you get an actual response.

I thought it's a cleaner solution to just set the heights externally (kind of like pushing new heights into the stream).

Note that this is changed in https://github.com/astriaorg/astria/pull/725

crates/astria-conductor/src/block_cache.rs
SuperFluffy: yeah

PR #700, opened 2024-01-19
feat(sequencer): use `IbcComponent` and penumbra `HostInterface` 

### LINKED ISSUE #650: sequencer: bind IBC ports on startup
for now, we just need to bind the `transfer` port to the `accounts` component.

noot:
## Summary
need to set specific state keys on `InitChain` and `BeginBlock` for the IBC impl to work.

also uses the `HostInterface` implemented here: https://github.com/penumbra-zone/penumbra/pull/3598 needed so that the IBC impl can access astria-specific state keys.

## Background
as above

## Changes
- use the penumbra `IBCComponent` within our app, which deals with the required IBC state
- don't need to explicitly bind `transfer` ports as this is checked in the penumbra IBC implementation (`PortId::transfer()`)
- implement `HostInterface` for `AstriaHost` 

## Testing
with hermes

## Breaking Changelist
this updates the values in storage (adds `chain_id`)

## Related Issues

closes #650 (not exactly related but closing anyways)


crates/astria-sequencer/src/app.rs
```diff
@@ -152,6 +154,8 @@ impl App {
         crate::asset::initialize_native_asset(&genesis_state.native_asset_base_denomination);
         state_tx.put_native_asset_denom(&genesis_state.native_asset_base_denomination);
 
+        // TODO: pass from config
```
noot, line 157: need to put this in the genesis file, i'll make another PR for this

crates/astria-sequencer/justfile
```diff
@@ -14,7 +14,7 @@ run:
 run-cometbft:
   cometbft init
   ../../target/debug/astria-sequencer-utils --genesis-app-state-file=test-genesis-app-state.json --destination-genesis-file=$HOME/.cometbft/config/genesis.json --chain-id=astria
-  sed -i'.bak' 's/timeout_commit = "1s"/timeout_commit = "15s"/g' ~/.cometbft/config/config.toml
+  sed -i'.bak' 's/timeout_commit = "1s"/timeout_commit = "3s"/g' ~/.cometbft/config/config.toml
```
joroshiba, line 17: We use 2s for deployed networks.

crates/astria-sequencer/src/app.rs
joroshiba: A chain ID is already set in the cometbft genesis?

noot: yeah woops i'll update to pass it in from init_chain

crates/astria-sequencer/justfile
noot: ok, should i change it here then?

SuperFluffy (approved):
Seems fine in that it just wires up the definitions used by penumbra, but truth be told I am lacking context to evaluate this.

sambukowski (approved):

PR #701, opened 2024-01-19
feat(proto): add sequencer service proto

joroshiba:
## Summary
Defines an API service to be implemented in sequencer which can be used to decouple conductor & sequencer relayer from core transaction types.

## Background
Currently conductor & relayer utilize CometBFT apis to grab sequencer data, and parse the transactions. This tightly couples them with transaction formats, and increases computation cost in parsing each new block. When implemented, this will enable the types to be decoupled in conductor so we can change transaction formats without breaking rollups, and streamline the code for relayer. 

## Changes
- Adds a new service definitions


proto/astria/sequencer/v1alpha1/block.proto
```diff
@@ -41,3 +41,15 @@ message SequencerBlock {
   // the rollup transactions.
   astria.sequencer.v1alpha1.Proof rollup_ids_proof = 4;
 }
+
+// `RollupSpecificSequencerBlock` is a subset of `SequencerBlock` that only contains
+// transaction data for a specific rollup.
+message RollupSpecificSequencerBlock {
```
SuperFluffy, line 47: Should this be flattened or contain the `RollupTransactions` object?

noot: this needs the proofs + commitments also i believe? the conductor will be calling `GetRollupSpecificSequencerBlock` if i understand correctly, so needs to verify it

noot: also not a huge fan of the name, maybe `RollupDataSubset` or something is better?

joroshiba: If you get the data directly from a trusted sequencer node what would be verified?

I guess if we added a light client to conductor you could verify from a sequencer node against header?

noot:
```
    rollup_transactions_proof: merkle::Proof,
    rollup_ids_proof: merkle::Proof,
```

see `SequencerBlock` in astria-core, you would just need to add these. and yep I'm thinking in the case that the sequencer node is not trusted and you're just getting the data for light client verification

joroshiba: Should we just eliminate this structure altogether? We could use the same SequencerBlock type, but filtered down. This would allow us to expand the API in the future and say request the transactions for multiple rollups.

joroshiba:
I have removed the streaming services as they aren't day 1 requirements. I also simplified to use the same object for subsets and full sequencer blocks. These were very close to the same shape no strong upside, provides more options in the future (ie fetching multiple rollups data)

proto/sequencerapis/astria/sequencer/v1alpha1/service.proto
```diff
@@ -0,0 +1,25 @@
+syntax = "proto3";
+
+package astria.sequencer.v1alpha1;
+
+import "astria/sequencer/v1alpha1/block.proto";
+
+message GetSequencerBlockRequest {
```
SuperFluffy, line 7: I wonder if a nice API would be:

```proto
message GetSequencerBlockRequest {
    int64 height = 1; // or uint32
    optional bytes rollup_id = 2;
}
```

This means you can either get the full block if you leave the field unset. Or you can get the block with the rollup specific transactions filtered out.

Bonus: `optional` and `repeated` are compatible and not a breaking change if we decide to `optional -> repeated` later

proto/sequencerapis/astria/sequencer/v1alpha1/service.proto
```diff
@@ -0,0 +1,25 @@
...
+  // The height of the block to retrieve.
+  int64 height = 1;
+}
+
+message GetFilteredSequencerBlockRequest {
```
SuperFluffy, line 12: See my suggestion above to add an `optional bytes rollup_id` field to `GetSequencerBlockRequet`.

Other than that I think I liked your original naming suggestion of calling this `GetRollupSpecificSequencerBlockRequest`. Although wordy it's much more concrete than `Filtered` which sounds very general.

proto/sequencerapis/astria/sequencer/v1alpha1/service.proto
joroshiba: My general view is to align with standard resource based RPC type architectures there should be a standard `Get` accessor, such that we could map with an http proxy to an http `GET` request with no body or params only an id. 

The other method I would consider "custom". This loosely aligns with practices suggested here: https://google.aip.dev/131, https://google.aip.dev/136

proto/sequencerapis/astria/sequencer/v1alpha1/service.proto
```diff
@@ -0,0 +1,25 @@
+syntax = "proto3";
+
+package astria.sequencer.v1alpha1;
+
+import "astria/sequencer/v1alpha1/block.proto";
+
+message GetSequencerBlockRequest {
+  // The height of the block to retrieve.
+  int64 height = 1;
```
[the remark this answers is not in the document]
joroshiba: I updated the uint64, would rather be forward looking here.

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -544,10 +544,10 @@ impl ExecutableBlock {
...
         let height = block.height();
         let timestamp = convert_tendermint_to_prost_timestamp(block.header().time);
-        let transactions = block
-            .into_rollup_transactions()
-            .swap_remove(&id)
-            .unwrap_or_default();
+        let transactions = match block.into_rollup_transactions().swap_remove(&id) {
+            Some(rollup_transactions) => rollup_transactions.transactions().to_vec(),
+            None => vec![],
+        };
```
noot, line 550:
```suggestion
        let transactions = match block.into_rollup_transactions().swap_remove(&id).map(|txs| txs.transactions().to_vec()).unwrap_or_default();`
```

crates/astria-core/src/sequencer/v1alpha1/block.rs
```diff
@@ -28,21 +28,35 @@ impl RollupTransactionsError {
...
 
 /// The opaque transactions belonging to a rollup identified by its rollup ID.
-#[derive(Clone)]
+#[derive(Clone, Debug, PartialEq)]
 pub struct RollupTransactions {
     /// The 32 bytes identifying a rollup. Usually the sha256 hash of a plain rollup name.
     id: RollupId,
     /// The serialized opaque bytes of the rollup transactions.
     transactions: Vec<Vec<u8>>,
+    /// Proof that this set of transactions belongs in the merkle tree
```
noot, line 58:
```suggestion
    /// Proof that this set of transactions belongs in the `sequence::Action` tree
```

crates/astria-core/src/sequencer/v1alpha1/block.rs
```diff
@@ -366,6 +393,25 @@ impl SequencerBlock {
         }
     }
 
+    #[must_use]
+    pub fn filtered_block(self, rollup_ids: Vec<RollupId>) -> Self {
```
noot, line 397: this should take `&self`? unless you intended for this to take the rollup txs out of `self`?

crates/astria-core/src/sequencer/v1alpha1/block.rs
```diff
@@ -366,6 +393,25 @@ impl SequencerBlock {
...
+        let mut rollup_transactions = IndexMap::with_capacity(rollup_ids.len());
+        for id in rollup_ids {
+            let Some(rollup_transaction) = self.rollup_transactions.get(&id) else {
```
noot, line 400: naming here is a bit weird since this is still multiple txs right? maybe name `rollup_data` or something?

crates/astria-core/src/sequencer/v1alpha1/block.rs
```diff
@@ -477,9 +538,9 @@ impl SequencerBlock {
     pub fn try_from_raw(raw: raw::SequencerBlock) -> Result<Self, SequencerBlockError> {
         fn rollup_txs_to_tuple(
             raw: raw::RollupTransactions,
-        ) -> Result<(RollupId, Vec<Vec<u8>>), IncorrectRollupIdLength> {
-            let id = RollupId::try_from_slice(&raw.id)?;
-            Ok((id, raw.transactions))
+        ) -> Result<(RollupId, RollupTransactions), RollupTransactionsError> {
+            let rollup_transactions = RollupTransactions::try_from_raw(raw)?;
+            Ok((rollup_transactions.id, rollup_transactions))
```
noot, line 543: `RollupTransactions` contains the id, can just return `RollupTransactions`

crates/astria-core/src/sequencer/v1alpha1/celestia.rs
```diff
@@ -44,15 +47,12 @@ impl CelestiaBlobBundle {
...
-                .construct_proof(i)
-                .expect("the proof must exist because the tree was derived with the same leaf");
+        for (rollup_id, rollup_txs) in rollup_transactions {
             tail.push(CelestiaRollupBlob {
                 sequencer_block_hash: block_hash,
                 rollup_id,
-                transactions,
-                proof,
+                transactions: rollup_txs.transactions().to_vec(),
+                proof: rollup_txs.proof().clone(),
```
noot, line 55: would rather not clone these, can you take them out of `rollup_txs`?

crates/astria-sequencer/src/grpc/sequencer.rs
```diff
@@ -0,0 +1,120 @@
...
+        if curr_block_height < request.height {
+            return Err(Status::invalid_argument(
+                "requested height is greater than current block height",
+            ));
+        }
+
+        let height: u32 = request
+            .height
+            .try_into()
+            .expect("height should be a valid u32");
```
noot, line 64: no expect here - rpc methods should never panic

crates/astria-sequencer/src/grpc/sequencer.rs
```diff
@@ -0,0 +1,120 @@
...
+                return Err(Status::invalid_argument("Rollup ID must be 32 bytes"));
+            };
+            rollup_ids.push(rollup_id);
+        }
+
+        if curr_block_height < request.height {
+            return Err(Status::invalid_argument(
+                "requested height is greater than current block height",
+            ));
+        }
```
noot, line 103: put this check above the prev block

crates/astria-sequencer/src/grpc/sequencer.rs
```diff
@@ -0,0 +1,120 @@
...
+    ) -> Result<Response<RawSequencerBlock>, Status> {
+        let snapshot = self.storage.latest_snapshot();
+        let Ok(curr_block_height) = snapshot.get_block_height().await else {
+            return Err(Status::internal("failed to get block height from storage"));
+        };
+
+        let request = request.into_inner();
+
+        let mut rollup_ids: Vec<RollupId> = vec![];
+        for id in request.rollup_ids {
```
SuperFluffy, line 92: This conversion (here and below) should live in `astria-core` to have symmetry with the other types.

crates/astria-sequencer/src/grpc/sequencer.rs
```diff
@@ -0,0 +1,120 @@
...
+            ));
+        }
+
+        let height: u32 = request
+            .height
+            .try_into()
+            .expect("height should be a valid u32");
+        let block = match self.client.sequencer_block(height).await {
+            Ok(block) => block.filtered_block(rollup_ids),
+            Err(_) => {
```
SuperFluffy, line 111: Why ignore the error? Let's report it, might be useful.

crates/astria-sequencer/src/grpc/sequencer.rs
SuperFluffy, line 1: The endpoints here could be tested with some simple mocks.

SuperFluffy (commented):
First set of comments. Will do more tomorrow.

.github/labeler.yml
SuperFluffy, line 1: move out of this PR?

.gitattributes
SuperFluffy, line 1: move out of this PR?

crates/astria-core/src/sequencer/v1alpha1/block.rs
```diff
@@ -28,21 +28,35 @@ impl RollupTransactionsError {
...
+    }
 }
 
 #[derive(Debug, thiserror::Error)]
 pub enum RollupTransactionsErrorKind {
     #[error("`id` field is invalid")]
     RollupId(#[source] IncorrectRollupIdLength),
+    #[error("the expected field in the raw source type was not set: `{0}`")]
+    FieldNotSet(&'static str),
+    #[error("failed constructing a proof from the raw protobuf transaction proof")]
```
SuperFluffy, line 47: Mention the field name that is problematic.

crates/astria-core/src/sequencer/v1alpha1/block.rs
```diff
@@ -58,17 +72,25 @@ impl RollupTransactions {
         &self.transactions
     }
 
+    /// Returns the opaque transactions bytes.
```
SuperFluffy, line 75: Wrong comment?

proto/sequencerapis/astria/sequencer/v1alpha1/service.proto
```diff
@@ -0,0 +1,35 @@
...
+
+service SequencerService {
+  // Given a block height, returns the sequencer block at that height.
+  rpc GetSequencerBlock(GetSequencerBlockRequest) returns (SequencerBlock) {
+    option (google.api.http) = {get: "/v1alpha1/sequencer/{height}"};
+  }
+
+  // Given a block height and set of rollup ids, returns a SequencerBlock which
+  // is filtered to contain only the transactions that are relevant to the given rollup.
+  rpc FilteredSequencerBlock(FilteredSequencerBlockRequest) returns (SequencerBlock) {
```
SuperFluffy, line 29: To be symmetric with the other functions, this should IMO called `GetFilteredSequencerBlock` (although I still prefer a more explicit naming like `GetSequencerBlockFilteredByRollupIds` or something like that).

Also, this should return a different value - name it `FilteredSequencerBlock` or `SequencerBlockSubset` or something like that. These objects have different semantics.

crates/astria-core/src/sequencer/v1alpha1/block.rs
SuperFluffy, line 397: The naming of this method does not feel idiomatic:
1. manipulations of the same type should have a verb in the name, e.g. `filter_blocks_by_rollup_id` or something like that.
2. transformations to a different type should be `to_*` or `into_*`
3. This returning `Self` feels wrong. I'd prefer to call it `FilteredSequencerBlock` or `SequencerBlockSubset` or something like that
4. I'd expect that a function named `filter` takes a function `F: FnMut(&RollupId) -> bool` to match the ecosystem. But even if it doesn't a better signature would be `f<T: IntoIterator<Item = RollupId>>(self, rollup_ids: T)` or at least `f(self, rollup_ids: &[RollupId])`

crates/astria-core/src/sequencer/v1alpha1/celestia.rs
```diff
@@ -32,7 +32,10 @@ impl CelestiaBlobBundle {
             rollup_ids_proof,
         } = block.into_unchecked();
 
-        let tree = super::derive_merkle_tree_from_rollup_txs(&rollup_transactions);
+        let rollup_base_transactions = rollup_transactions
```
SuperFluffy, line 35: Is all of this still necessary if the individual`RollupTransaction` types also contain proofs?

crates/astria-core/src/sequencer/v1alpha1/celestia.rs
```diff
@@ -32,7 +32,10 @@ impl CelestiaBlobBundle {
...
+            .iter()
+            .map(|(rollup_id, txs)| (rollup_id, txs.transactions()));
+        let tree = super::derive_merkle_tree_from_rollup_txs(rollup_base_transactions);
```
SuperFluffy, line 38: What happens if this tree is constructed from filtered block? That is, the sequencer block yielded by the transformation `filtered_block(SequencerBlock, [RollupId]) -> SequencerBlock`? Isn't this tree wrong?

crates/astria-core/src/sequencer/v1alpha1/mod.rs
```diff
@@ -308,13 +310,15 @@ impl Protobuf for merkle::Proof {
 /// It is the responsbility if the caller to ensure that the iterable is
 /// deterministic. Prefer types like `Vec`, `BTreeMap` or `IndexMap` over
 /// `HashMap`.
-pub fn derive_merkle_tree_from_rollup_txs<'a, T: 'a>(rollup_ids_to_txs: T) -> merkle::Tree
+pub fn derive_merkle_tree_from_rollup_txs<'a, T: 'a, U: 'a>(rollup_ids_to_txs: T) -> merkle::Tree
 where
-    T: IntoIterator<Item = (&'a RollupId, &'a Vec<Vec<u8>>)>,
+    T: IntoIterator<Item = (&'a RollupId, &'a U)>,
+    U: AsRef<[Vec<u8>]> + 'a + ?Sized,
```
SuperFluffy, line 316: Would this also work?
```rust
    U: AsRef<[u8]> + 'a + ?Sized,
```

SuperFluffy (changes requested):
Requesting changes because `SequencerBlock::filtered_block -> SequencerBlock` followed by the construction of the `(CelestiaSequencerBlob, [CelestiaRollupBlob])` can lead to very surprising results (namely that the full Merkle Tree across all rollup transactions cannot be reconstructed, leading to different proofs).

I also believe that doing this construction is no longer strictly necessary with the `RollupTransaction` type having been enriched by proofs?


Also a bunch of other comments regarding idiomatic naming.

crates/astria-core/src/sequencer/v1alpha1/celestia.rs
joroshiba: It's a different proof we are constructing here, and is needed for Celestia metadata blob posting I believe.

joroshiba:
> Requesting changes because `SequencerBlock::filtered_block -> SequencerBlock` followed by the construction of the `(CelestiaSequencerBlob, [CelestiaRollupBlob])` can lead to very surprising results (namely that the full Merkle Tree across all rollup transactions cannot be reconstructed, leading to different proofs).
> 
> I also believe that doing this construction is no longer strictly necessary with the `RollupTransaction` type having been enriched by proofs?
> 
> Also a bunch of other comments regarding idiomatic naming.

This makes sesnse to me. Protobuf should probably have two different types such that it's clear that you can't construct the whole tree from the filtered object. Even if all the fields are the same. We could not have the rollup specific proofs on the full block, but that isn't really cleaner when stored in DB, and makes more work to generate all the proofs when relaying.

crates/astria-core/src/sequencer/v1alpha1/block.rs
```diff
@@ -417,38 +463,53 @@ impl SequencerBlock {
...
                 SequencerBlockError::rollup_transactions_root_does_not_match_reconstructed(),
             );
         }
 
-        // ensure the rollup IDs commitment matches the one calculated from the rollup data
-        if rollup_ids_root != merkle::Tree::from_leaves(rollup_transactions.keys()).root() {
-            return Err(SequencerBlockError::rollup_ids_root_does_not_match_reconstructed());
+        let mut rollup_transactions = IndexMap::new();
+        for (i, (id, transactions)) in rollup_base_transactions.into_iter().enumerate() {
+            let proof = tree
```
SuperFluffy, line 500: I believe the construction of these proofs are wrong. Reason:

1. `tree` is constructed from *all* the elements of `data`
2. `data[0]` and `data[1]` are `rollup_transactions_root` and `rollup_ids_root`, respectively.
3. this loop asks to construct proofs for leaves `0..len(rollup_txs)`, respectively, but that does not match the leaves of three.
4. Proof: try proving `rollup_base_transactions[0]` against the proof constructed for `i = 0` - you will not get the data hash.

crates/astria-core/src/sequencer/v1alpha1/celestia.rs
SuperFluffy: I believe it should be the same proof but currently isn't. I have re-read and added a comment to the loop in `try_from_cometbft` that constructs the individual proofs. I believe as it stands the construction is not correct.

crates/astria-sequencer/src/grpc/sequencer.rs
noot: what conversion? turning the vec of bytes into a `Vec<RollupId>`?

crates/astria-core/src/sequencer/v1alpha1/mod.rs
noot: no, it complains:
```
the trait bound `u8: AsRef<[u8]>` is not satisfied
required for `&u8` to implement `AsRef<[u8]>`rustc[Click for full compiler diagnostic](rust-analyzer-diagnostics-view:/diagnostic%20message%20%5B1%5D?1#file%3A%2F%2F%2Fhome%2Fe%2Fastria%2Fcrates%2Fastria-core%2Fsrc%2Fsequencer%2Fv1alpha1%2Fmod.rs)
lib.rs(489, 12): required by a bound in `Tree::from_leaves`
astria_merkle::Tree
pub fn from_leaves<I, B>(iter: I) -> Self
where
    I: IntoIterator<Item = B>,
    B: AsRef<[u8]>,
```

crates/astria-core/src/sequencer/v1alpha1/block.rs
noot: updated to use `rollup_transaction_tree` (the right tree) for constructing the proofs, will add tests

proto/sequencerapis/astria/sequencer/v1alpha1/service.proto
noot: updated

crates/astria-core/src/sequencer/v1alpha1/block.rs
noot: updated to be `into_filtered_block`, thoughts?

crates/astria-core/src/sequencer/v1alpha1/celestia.rs
noot: fixed this, should be correct now, adding unit tests also

noot:
@SuperFluffy addressed your comments, added unit tests to ensure the conversions from cometbft block and to/from raw filtered block are working. also created the `FilteredSequencerBlock` type for clarity. ready for re-review

noot (approved):
approving let's get this in!!!

SuperFluffy (approved):

PR #703, opened 2024-01-22
feat(conductor): sync executor commitment state with sequencer reader

SuperFluffy:
## Summary
Sync the executor commitment state and sequencer reader so that it only fetches and forwards blocks at the current level.

## Background
Soft blocks should only be read from the sequencer and forwarded to the executor if the executor expects them. This patch ensures that obsolete heights are not fetched from sequencer (heights that have already been firmly confirmed).

This also lays the groundwork for allowing sequencer syncing and block forwarding to start at a later height than the firm block sync (although not implemented in this patch).

## Changes
- wrap the executor commitment state in a watch channel
- monitor the executor state in the sequencer reader
- drop obsolete heights from both the block cache and the height reader

## Testing
This interaction should be tested end to end.

## Related Issues

### DIFF
--- a/crates/astria-conductor/src/block_cache.rs
+++ b/crates/astria-conductor/src/block_cache.rs
@@ -51,6 +51,15 @@ impl<T> BlockCache<T> {
         Some(block)
     }
 
+    pub(crate) fn drop_obsolete(&mut self, latest_height: Height) {
+        self.next_height = std::cmp::max(self.next_height, latest_height);
+        // Splitting the btree always involves an allocation, so only do it if necessary
+        if self.inner.first_key_value().map(|(&height, _)| height) < Some(latest_height) {
+            let only_non_obsolete = self.inner.split_off(&latest_height);
+            self.inner = only_non_obsolete;
+        }
+    }
+
     /// Return a handle to the next block in the cache.
     ///
     /// This method exists to make fetching the next block async cancellation safe.
--- a/crates/astria-conductor/src/sequencer.rs
+++ b/crates/astria-conductor/src/sequencer.rs
@@ -1,6 +1,7 @@
 //! [`Reader`] reads reads blocks from sequencer and forwards them to [`crate::executor::Executor`].
 
 use std::{
+    cmp,
     collections::VecDeque,
     error::Error as StdError,
     pin::Pin,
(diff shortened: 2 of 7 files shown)


sambukowski (approved):

crates/astria-conductor/src/block_cache.rs
```diff
@@ -51,6 +51,15 @@ impl<T> BlockCache<T> {
...
     }
 
+    pub(crate) fn drop_obsolete(&mut self, latest_height: Height) {
+        self.next_height = std::cmp::max(self.next_height, latest_height);
+        // Splitting the btree always involves an allocation, so only do it if necessary
+        if self.inner.first_key_value().map(|(&height, _)| height) < Some(latest_height) {
+            let only_non_obsolete = self.inner.split_off(&latest_height);
+            self.inner = only_non_obsolete;
+        }
+    }
```
noot, line 61: is splitting more efficient than clearing everything before `latest_height`?

crates/astria-conductor/src/sequencer.rs
```diff
@@ -52,34 +55,42 @@ use crate::{
...
     /// The channel used to send messages to the executor task.
     executor_channel: mpsc::UnboundedSender<SequencerBlock>,
 
+    /// The state of the executor to receive information about the next
+    /// expected sequencer height.
+    executor_state: watch::Receiver<executor::State>,
+
+    /// The minimum height the executor must reach before this reader will forward
+    /// blocks to it.
+    not_before_height: Height,
```
noot, line 71: which height? `soft`?

noot: also this looks like it's only read from and never updated (unless i missed something)

crates/astria-conductor/src/block_cache.rs
SuperFluffy: I tried understanding both `split_off` and `remove_entry` and honestly didn't get far with the latter. `BTreeMap` might likely not be the right datastructure for this but I am going to leave `split_off` because it's easier to understand.

crates/astria-conductor/src/sequencer.rs
SuperFluffy: Yeah, I removed this in a subsequent PR. Sorry about that.

PR #709, opened 2024-01-24
fix(core, conductor): relax size requirements of hash buffers

SuperFluffy:
## Summary
Makes `hash` and `parent_block_hash` fields `Bytes` instead of `[u8; 32]` in the idiomatic Rust types for the execution v1alpha2 API.

## Background
The `hash` and `parent_block_hash` were over-constrained in the assumption that they behaved the same as the hashes emitted by cometbft/sequencer. However, the hashes are actually defined by the rollup and can thus be of any length.

## Changes
- Adjust `tools/protobuf-compiler` to emit `Bytes` for all types in the `astria.execution.v1alpha2` protobuf package (instead of `Vec<u8>`)
- Update `astria-core::execution::v1alpha2` to expect arbitrarily sized buffers instead of fixed-size array.
- Relax version constraints for the bytes crate


noot (approved):
just curious, why `Bytes` over `Vec<u8>`?

sambukowski (approved):

SuperFluffy:
> just curious, why `Bytes` over `Vec<u8>`?

`Bytes` are cheap to clone, `Vec`s are expensive. I want to start migrating all of byte buffers in `core` to `Bytes`. Thought this is a good starting point.

joroshiba (approved):

PR #710, opened 2024-01-25
feat(conductor, proto): add genesis info RPC, initialize in terms of it

SuperFluffy:
## Summary
Adds a genesis info RPC to the executor v1alpha2 API and initializes conductor using it.

## Background
Certain information about the rollup should be considered a core part of the rollup definition, but was done through environmental variable configuration making it harder to spin up nodes that follow the same rules.

## Changes
- Create `GenesisInfo` protos
- Added idiomatic types to `astria-core`
- Remove fields from config that are now derived from genesis info
- Initialize the executor actor in terms of the genesis info and propagate it to the celestia and sequencer readers

## Testing
Updated executor tests to still pass. Rest has to be tested end-to-end.

## Breaking Changelist
- Requires an updated rollup node to implement the genesis info RPCs

## Related Issues

### DIFF
--- a/crates/astria-conductor/src/executor/mod.rs
+++ b/crates/astria-conductor/src/executor/mod.rs
@@ -42,6 +42,7 @@ mod state;
 #[cfg(test)]
 mod tests;
 
+pub(super) use client::Client;
 pub(crate) use state::State;
 
 pub(crate) struct Executor {
@@ -52,14 +53,10 @@ pub(crate) struct Executor {
 
     shutdown: oneshot::Receiver<()>,
 
-    /// The execution rpc client that we use to send messages to the execution service
-    client: client::Client,
-
-    /// Chain ID
-    rollup_id: RollupId,
+    rollup_address: tonic::transport::Uri,
 
     /// Tracks SOFT and FIRM on the execution chain
-    state: watch::Sender<state::State>,
+    state: watch::Sender<State>,
 
     /// Tracks executed blocks as soft commitments.
     ///
@@ -95,6 +92,14 @@ impl Executor {
 
     #[instrument(skip_all)]
     pub(crate) async fn run_until_stopped(mut self) -> eyre::Result<()> {
+        let client = Client::connect(self.rollup_address.clone())
+            .await
+            .wrap_err("failed connecting to rollup node")?;
+
+        self.set_initial_node_state(client.clone())
+            .await
+            .wrap_err("failed setting initial rollup node state")?;
+
         loop {
             select!(
                 biased;
@@ -117,7 +122,7 @@ impl Executor {
                         block.hash = %telemetry::display::hex(&block.block_hash),
                         "received block from celestia reader",
                     );
-                    if let Err(e) = self.execute_firm(block).await {
+                    if let Err(e) = self.execute_firm(client.clone(), block).await {
                         let reason = "failed executing firm block";
                         error!(
                             error = AsRef::<dyn std::error::Error>::as_ref(&e),
@@ -134,7 +139,7 @@ impl Executor {
                         block.hash = %telemetry::display::hex(&block.block_hash()),
                         "received block from sequencer reader",
                     );
-                    if let Err(e) = self.execute_soft(block).await {
+                    if let Err(e) = self.execute_soft(client.clone(), block).await {
                         let reason = "failed executing soft block";
                         error!(
                             error = AsRef::<dyn std::error::Error>::as_ref(&e),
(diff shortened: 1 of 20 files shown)


crates/astria-conductor/src/executor/state.rs
```diff
@@ -1,60 +1,95 @@
...
+forward_impls!(
+    [firm -> &Block],
+    [soft -> &Block],
+    [firm_parent_hash -> [u8; 32]],
+    [soft_parent_hash -> [u8; 32]],
+    [celestia_base_block_height -> CelestiaHeight],
+    [rollup_id -> RollupId],
+    [next_firm_sequencer_height -> Height],
+    [next_soft_sequencer_height -> Height],
+);
```
noot, line 75: what is this for / where is it used?

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -266,7 +274,38 @@ impl Executor {
...
+    #[instrument(skip_all)]
+    async fn set_initial_node_state(&self, client: Client) -> eyre::Result<()> {
+        let genesis_info = {
+            let mut client = client.clone();
+            async move {
+                client
+                    .get_genesis_info()
+                    .await
+                    .wrap_err("failed getting genesis info")
+            }
```
noot, line 286: why the `async move` block?

crates/astria-conductor/src/executor/state.rs
```diff
@@ -1,60 +1,95 @@
...
+    pub(super) fn new() -> Self {
+        Self {
+            inner: None,
+        }
+    }
+
+    pub(super) fn init(&mut self, genesis_info: GenesisInfo, commitment_state: CommitmentState) {
+        self.inner
+            .replace(StateImpl::new(genesis_info, commitment_state));
+    }
```
noot, line 36: should we make sure this is called only once?

crates/astria-conductor/src/executor/state.rs
```diff
@@ -100,18 +143,42 @@ impl State {
     }
 }
 
+// Maps a rollup height to a sequencer heights.
+/// # Panics
```
noot, line 147:
```suggestion
/// Maps a rollup height to a sequencer height.
///
/// # Panics
```

crates/astria-core/src/execution/v1alpha2/mod.rs
```diff
@@ -2,9 +2,119 @@ use prost_types::Timestamp;
...
+///
+/// Contains information about the rollup id, and base heights for both sequencer & celestia.
+///
+/// Usually constructed its [`Protobuf`] implementation from a
+/// [`raw::GenesisInfo`].
+#[derive(Clone, Copy, Debug)]
+pub struct GenesisInfo {
+    /// The rollup id which is used to identify the rollup txs.
+    rollup_id: RollupId,
+    /// The first height of sequencer which is used to identify the first block of the rollup.
```
noot, line 39:
```suggestion
    /// The sequencer block height which contains the first block of the rollup.
```

crates/astria-core/src/execution/v1alpha2/mod.rs
```diff
@@ -2,9 +2,119 @@ use prost_types::Timestamp;
...
+    sequencer_genesis_block_height: tendermint::block::Height,
+    /// The first height of celestia which to look for sequencer blocks.
+    celestia_base_block_height: celestia_tendermint::block::Height,
```
noot, line 42: is this the only place `celestia_tendermint` is used? is their height type internally a different type than normal tendermint?

crates/astria-conductor/src/executor/state.rs
```diff
@@ -100,18 +143,42 @@ impl State {
...
+) -> Height {
+    let first_sequencer_height: u32 = first_sequencer_height
+        .value()
+        .try_into()
+        .expect("should not fail; cometbft heights are internally represented as int64");
+    first_sequencer_height
+        .checked_add(current_rollup_height)
+        .expect(
+            "should not overflow; if this overflows either the first sequencer height or current \
+             rollup height have been incorrectly set incorrectly set, are in the future, or the \
```
sambukowski, line 163: duplicated "incorrectly set"

sambukowski (approved):

crates/astria-conductor/src/executor/state.rs
SuperFluffy: This is to avoid having to write `self.inner.as_ref().expect("").<method>()` for every single method on on the inner state.

crates/astria-conductor/src/executor/state.rs
SuperFluffy: Replace this with `*self.inner.as_mut().expect("msg") = ...`.

crates/astria-core/src/execution/v1alpha2/mod.rs
SuperFluffy: The type is internally exactly the same, but unfortunately from a type-system perspective it's different.

PR #711, opened 2024-01-29
refactor(conductor): interact with executor via handle

SuperFluffy:
## Summary
Added an `executor::Handle` type to send blocks, read state.

## Background
The previous implementation required too much manual plumbing to set up connections to the executor. A handle to a long running task is a more elegant and idiomatic solution. It also allows to ensure at the type-level that a `Handle::wait_for_init -> Handle<StateIsInit>` is called prior to reading its state and sending data to it because all methods to do so are only implemented on `Handle<StateIsInit>`.

## Changes
- Introduce an `executor::Handle<T>` type to interact with the `Executor`.
- Move all send channels out of the executor and into the Handle
- Introduce `StateIsInit` and `StateNotInit` marker types to ensure the executor state is initialized (which is the case when it read the genesis info and commitment state off its rollup node)

## Testing
Needs to be done end-to-end

## Related 
Part of #691.

Also removes the `WeakUnboundedSender` in favor of `UnboundedSender` because that was buggy (noticed in https://github.com/astriaorg/astria/pull/694)

### DIFF
--- a/crates/astria-conductor/src/executor/mod.rs
+++ b/crates/astria-conductor/src/executor/mod.rs
@@ -45,14 +45,41 @@ mod tests;
 pub(super) use client::Client;
 pub(crate) use state::State;
 
+#[derive(Clone, Debug)]
+pub(crate) struct StateNotInit;
+#[derive(Clone, Debug)]
+pub(crate) struct StateIsInit;
+
 #[derive(Debug, Clone)]
-pub(crate) struct Handle {
+pub(crate) struct Handle<TStateInit = StateNotInit> {
     firm_blocks: mpsc::UnboundedSender<ReconstructedBlock>,
     soft_blocks: mpsc::UnboundedSender<SequencerBlock>,
     state: watch::Receiver<State>,
+    _state_init: TStateInit,
 }
 
-impl Handle {
+impl Handle<StateNotInit> {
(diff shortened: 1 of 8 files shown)


crates/astria-conductor/src/executor/mod.rs
```diff
@@ -45,11 +52,83 @@ mod tests;
...
+/// To be be useful, [`Handle<StateNotInit>::wait_for_init`] must be called in
+/// order to obtain a [`Handle<StateInit>`]. This is to ensure that the executor
+/// state was primed before using its other methods. See [`State`] for more
+/// information.
+#[derive(Debug, Clone)]
+pub(crate) struct Handle<TStateInit = StateNotInit> {
+    firm_blocks: mpsc::UnboundedSender<ReconstructedBlock>,
+    soft_blocks: mpsc::UnboundedSender<SequencerBlock>,
+    state: watch::Receiver<State>,
+    _state_init: TStateInit,
```
noot, line 71: is this instead of a `PhantomData`?

noot (approved):
nice, this looks good!

sambukowski (approved):

crates/astria-conductor/src/executor/mod.rs
SuperFluffy: Yes. But since we are only using zero sized unit types there is no advantage to using `PhantomData`, so this is simpler.

PR #721, opened 2024-02-01
feat(sequencer)!: add `ibc_sudo_address` to genesis, only allow `IbcRelay` actions from this address

### LINKED ISSUE #720: sequencer: permissioning for opening IBC channels
currently, anyone can open IBC clients/connections/channels. on our real network we want to have this be permissioned, so we can control which tokens (and denoms) will be on the network. for now we can have this be permissioned via a sudo key.

noot:
## Summary
as title says, make modifying IBC clients/connections/channels permissioned. this allows only a specific key (should be the key used by hermes) to send `IbcRelay` actions.

## Background
we probably don't want IBC channels being arbitrarily opened at this time.

## Changes
- add `ibc_sudo_address` to genesis
- store this in state
- only allow `IbcRelay` actions from this address

## Testing
unit tests

## Breaking Changelist
- updates state by adding `ibc_sudo_address`

## Related Issues

closes #720

### DIFF
--- a/crates/astria-sequencer/src/accounts/component.rs
+++ b/crates/astria-sequencer/src/accounts/component.rs
@@ -32,6 +32,9 @@ impl Component for AccountsComponent {
                 .put_account_balance(account.address, native_asset.id(), account.balance)
                 .context("failed writing account balance to state")?;
         }
+        state
+            .put_ibc_sudo_address(app_state.ibc_sudo_address)
+            .context("failed to set IBC sudo key")?;
         Ok(())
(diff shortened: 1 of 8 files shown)


joroshiba (approved):
I'm not seeing the change in state from sudo_key to sudo_address? Is that implemented somehwere, otherwise LGTM

noot:
@joroshiba the state key doesn't need to be changed, it's unrelated to the name of the field in genesis, so this is fine!

PR #725, opened 2024-02-05
feat(conductor): restrict firm<>soft commitment spread and add celestia search windows

SuperFluffy:
## Summary
This patch restricts the maximum permitted spread between soft and firm commitments to 16 heights and adds a search window to celestia heights within which the next sequencer block must be found.

## Background
As part of the refactor it was found that the loop reading off of sequencer was too hot, reading more sequencer blocks than could be executed against the rollup node leading to conductor crashing with out of memory. To fix this a, the soft-block channel from the sequencer reader to the executor needed to be changed from an unlimited to a limited channel.

With this rework it made sense to go all the way and also backpressure on the celestia reader.

## Changes
- Executor tests if soft commitments are too far ahead of firm commitments, and stops pulling soft blocks from sequencer-reader until firm commitments have caught up (currently hard-coded at a difference of 16).
- The above leads to backpressure on sequencer reader, which stops reading new sequencer heights until the channel is free
- Because the tendermint-rs JSONRPC subscription only allows for reading full blocks the code is changed to not deserialize raw tx bytes and only keep sequencer block heights. This is wasteful in network costs but doesn't put pressure on conductor memory and will hopefully be addressed with future tendermint-rs (or cometbft-rs) releases.
- The backpressure on the celestia-reader is indirect: it actively monitors the commitment state reported by executor and only advances the maximum celestia height that it will read based on the firm commitment number. 

## Testing
Conductor was redeployed several times with different configurations of `FirmOnly`, `SoftOnly`, and `SoftAndFirm` commitment levels. It was observed that it always reached the same commitment hash at a non-trivial height (in this case 1961).

## Related Issues


crates/astria-conductor/src/executor/builder.rs
```diff
@@ -24,6 +24,7 @@ pub(crate) struct NoShutdown;
 pub(crate) struct WithShutdown(oneshot::Receiver<()>);
 
 pub(crate) struct ExecutorBuilder<TRollupAddress = NoRollupAddress, TShutdown = NoShutdown> {
+    consider_commitment_spread: bool,
```
joroshiba, line 27: What's the purpose of providing the option in builder to not consider the commitment spread?

SuperFluffy: You can only consider the spread if you are both syncing soft && firm - although you only have to really disable it for soft syncing as here the firm block num will always stay at 0.

Executor does not a priori know that it will never receive firm blocks in the soft-only case and so you have to disable the spread.

sambukowski (approved):

PR #728, opened 2024-02-06
chore: build all binaries, fix pr title ci

joroshiba:
## Summary
Builds binaries for all of our releases, and makes it such pr title check runs on pr edit.

## Background
We added binary builds for the cli, but we don't have prebuilt binaries for our other services.

## Changes
- adds binary builds for conductor, composer, sequencer, and sequencer-relayer
- moves the conventional commit checker to admin so it will run on pr edit

## Testing
Manual workflow triggers on this PR.


sambukowski (approved):

crates/astria-composer/Cargo.toml
```diff
@@ -4,6 +4,11 @@ version = "0.3.1"
 edition = "2021"
 rust-version = "1.70.0"
 
+[[bin]]
+name = "astria-composer"
+test = false
```
noot, line 9: why is this false? doesn't this turn tests off for this crate when running `cargo test`? same for others
https://doc.rust-lang.org/cargo/reference/cargo-targets.html#the-test-field

.github/workflows/reusable-build.yml
```diff
@@ -0,0 +1,45 @@
...
+          - target: aarch64-apple-darwin
+            os: macos-latest
+            build-tool: cargo
+          - target: x86_64-apple-darwin
+            os: macos-latest
+            build-tool: cargo
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@1.73.0
```
SuperFluffy, line 31: We should look into how we can make this toolchain available for all our actions so that we don't have to update it everywhere

SuperFluffy (approved):

WafflesVonMaple (approved):

aajimal (approved):

PR #732, opened 2024-02-07
chore: split protos into multiple buf repos

joroshiba:
## Summary
Breaks out the execution api protos into seperate BUF image/package. This is done without any changes to local types, and maintains the same buf repo for core astria sequencer types on buf.

## Background
Importing modules via BUF for execution API currently pulls in everything and the kitchen sync. Notably this includes penumbra & cometbft types which can create type collisions. Execution API doesn't need these, is fully independent so breaking out into it's own module is clean.

## Changes
- Split execution API out
- Vendor penumbra & minimal set of cosmos types

## Testing
CI tests, locally running `buf lint` and `buf breaking` to confirm workspace updates work.

## Breaking Changelist
- Dependent code on execution api using buf build will have to update their module imports for any new updates.


tools/protobuf-compiler/src/main.rs
```diff
@@ -37,8 +37,7 @@ fn main() {
     cmd.arg("build")
         .arg("--output")
         .arg(buf_img.path())
-        .arg("--as-file-descriptor-set")
-        .arg(&src_dir);
```
SuperFluffy, line 41: Indeed not necessary? Or does this now require that the command is always run from the repo root? How does this interact with `find_protos` further down (line 60/61)?

buf.work.yaml
SuperFluffy, line 1: nice that we now make use of the workspace feature

SuperFluffy (commented):
quick comment on our proto compiler tool and the src argument.

tools/protobuf-compiler/src/main.rs
joroshiba: Seems to not be, workspace seems to solve. I've tested by adding changes and generating to make sure they show up. Find protos still needs the src dir, the workspace features maps things correctly

noot (approved):
looks good, could you also mention the tendermint/penumbra protos were vendored in the PR description and also mention the commit/version of the tendermint/penumbra protos included?

SuperFluffy (approved):
I like the changes. Before you merge, can you please add blurbs on all dependencies used in `extern_path` (probably in the workspace-wide `Cargo.toml` but also in the vendored protos) which commits/SHAs they are taken from? And also add that the some text that the Rust crates and vendored deps should be updated in lock-step (that might just be verifying that the underlying protos didn't change and potentially updating a rev)?

You probably also did the vendoring by just copying the files over?

I wonder if we can write a small utility to do that. For example, the (now removed) `proto.buf.yaml` contained explicit revs of penumbra and cosmos-sdk. For the tool we could write a config that a) checks buf protos at a specific version, and b) copies those protobuf packages that we actually desire.

joroshiba:
> I like the changes. Before you merge, can you please add blurbs on all dependencies used in `extern_path` (probably in the workspace-wide `Cargo.toml` but also in the vendored protos) which commits/SHAs they are taken from? And also add that the some text that the Rust crates and vendored deps should be updated in lock-step (that might just be verifying that the underlying protos didn't change and potentially updating a rev)?
> 
> You probably also did the vendoring by just copying the files over?
> 
> I wonder if we can write a small utility to do that. For example, the (now removed) `proto.buf.yaml` contained explicit revs of penumbra and cosmos-sdk. For the tool we could write a config that a) checks buf protos at a specific version, and b) copies those protobuf packages that we actually desire.

I've added notes of the specific versions vendored. We could potentially build a utility. The types we have actually imported outside of the penumbra one which will need to be updated to v1 when we upgrade to release 0.65 or newer are relatively unlikely to change (the tendermint types copied here have not been changed outside of formatting in 4 years). I think building the tool at this point more effort than reward.

We may change the header type ourselves in a fork at some point in the future although that would be complicated.

PR #737, opened 2024-02-09
feat(sequencer): add `ibc_relayer_addresses` list and allow modifications via `ibc_sudo_address`

### LINKED ISSUE #734: sequencer: change `ibc_sudo_address` to be a list and allow modifications
update `ibc_sudo_address` to be a list instead of one specific address, as we will have multiple relayers. also need to allow this list to be updateable (by `authority_sudo_address` or some other address?)

noot:
## Summary
we want to have multiple authorized addresses for IBC relayers. this adds a list of `ibc_relayer_addresses` which can submit `IbcRelay` actions. `ibc_sudo_address` is allowed to modify this list via `IbcRelayerChangeAction`.

## Changes
- add list of `ibc_relayer_addresses` to genesis and store in state
- implement `IbcRelayerChangeAction` which represents adding/deleting an address from the `ibc_relayer_addresses` list, only `ibc_sudo_address` can execute this.

## Testing
unit tests

## Breaking Changelist
- changes app state by adding ibc relayer addresses

## Related Issues

closes #734


proto/astria/sequencer/v1alpha1/transaction.proto
```diff
@@ -102,3 +103,10 @@ message IbcHeight {
   uint64 revision_number = 1;
   uint64 revision_height = 2;
 }
+
+message IbcRelayerChangeAction {
+  // the address to update
+  bytes address = 1;
+  // set to true to add the address to the relayer list, false to remove it
+  bool addition = 2;
```
joroshiba, line 111: Should we use an enum for this instead?

noot: up to you, might be a bit clearer/more self-documenting, i just went with a bool for simplicity for now

crates/astria-core/src/sequencer/v1alpha1/transaction/action.rs
```diff
@@ -755,3 +767,61 @@ enum Ics20WithdrawalErrorKind {
...
+impl IbcRelayerChangeActionError {
+    #[must_use]
+    fn invalid_address(err: IncorrectAddressLength) -> Self {
+        Self(IbcRelayerChangeActionErrorKind::InvalidAddress(err))
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+enum IbcRelayerChangeActionErrorKind {
+    #[error("an address was invalid")]
```
SuperFluffy, line 825: Name the field here.

crates/astria-sequencer/src/accounts/ibc_relayer_change.rs
```diff
@@ -0,0 +1,42 @@
...
+        StateReadExt,
+        StateWriteExt,
+    },
+    transaction::action_handler::ActionHandler,
+};
+
+#[async_trait]
+impl ActionHandler for IbcRelayerChangeAction {
+    async fn check_stateful<S: StateRead + 'static>(&self, state: &S, from: Address) -> Result<()> {
+        let ibc_sudo_address = state.get_ibc_sudo_address().await?;
```
SuperFluffy, line 26: context

SuperFluffy (commented):
The design of the proto message can lead to confusion IMO

proto/astria/sequencer/v1alpha1/transaction.proto
noot: okay, i'll change it then!

noot: @SuperFluffy @joroshiba updated to an enum

crates/astria-sequencer/src/app.rs
```diff
@@ -1167,6 +1170,58 @@ mod test {
         assert_eq!(validator_updates.get(&pub_key.into()).unwrap(), &update);
     }
 
+    #[tokio::test]
+    async fn app_deliver_tx_ibc_relayer_change_addition() {
```
joroshiba, line 1174: Should we add a test to validate it doesn't change when using the wrong sudo key?

joroshiba (approved):
approved with note of potential testing coverage gap

PR #739, opened 2024-02-09
feat!: metrics setup, and first metrics

joroshiba:
## Summary
Adds the setup for metrics into core telemetry init, adds config for it to all our standard services.

## Background
Metrics are currently only available through cometbft or rollup nodes, we want to have metrics in our configuration

## Changes
- Updates all service configs
- Updates telemetry to enable initializing prometheus metrics
- Adds a couple of metrics for sequencer-relayer, and celestia-client

## Breaking Changelist
- new configuration fields


joroshiba:
This is a replacement PR of https://github.com/astriaorg/astria/pull/716. After otel merges there were significant changes that made it easier to start over.

erwanor:
Heads up that we discovered a couple of warts with the 22 release of `metrics`. It accidentally broke our reporting because we upgraded main crate without upgrading the `metrics-util@0.15` and `metrics-prometheus-exporter@0.13` crates in lockstep.

Unfortunately, for now, this isn't easily remediated because `metrics-util` _pins a dependency_ and this create an intractable conflict with other hashbrown imports (in your case, one from the `jmt` crate, and another from the `astria-core` crate). I have reached out to the upstream to help with releasing a version that removes the pinned dep, and maybe discuss how it affects the "downstream experience".

Sorry about the churn, but although I'm hopeful that this will be sorted out early next week, we might decide to revert back to `metrics@0.19` if this stalls for too long (we need metrics).

joroshiba:
> Heads up that we discovered a couple of warts with the 22 release of `metrics`. It accidentally broke our reporting because we upgraded main crate without upgrading the `metrics-util@0.15` and `metrics-prometheus-exporter@0.13` crates in lockstep.
> 
> Unfortunately, for now, this isn't easily remediated because `metrics-util` _pins a dependency_ and this create an intractable conflict with other hashbrown imports (in your case, one from the `jmt` crate, and another from the `astria-core` crate). I have reached out to the upstream to help with releasing a version that removes the pinned dep, and maybe discuss how it affects the "downstream experience".
> 
> Sorry about the churn, but although I'm hopeful that this will be sorted out early next week, we might decide to revert back to `metrics@0.19` if this stalls for too long (we need metrics).

Thanks for the heads up. Ran into the dependency issue you mentioned myself while trying to update. Will revert to 0.19 for now and upgrade if resolved (as you noted.... we need metrics).

crates/astria-telemetry/src/lib.rs
```diff
@@ -219,6 +280,21 @@ impl Config {
             .try_init()
             .map_err(Error::init_subscriber)?;
 
+        if let Some(metrics_addr) = metrics_addr {
+            let addr: SocketAddr = metrics_addr.parse().map_err(Error::metrics_addr)?;
+            let mut metrics_builder = PrometheusBuilder::new()
+                .with_http_listener(addr)
+                .add_global_label("service", service_name);
```
SuperFluffy, line 287: So if one were to leave `service_name` unset then there would be a global label `"service"` with value `""`? Does that make sense or would that be confusing?

crates/astria-composer/src/main.rs
```diff
@@ -18,10 +18,18 @@ async fn main() -> ExitCode {
...
+    let mut telemetry_conf = telemetry::configure()
         .set_no_otel(cfg.no_otel)
         .set_force_stdout(cfg.force_stdout)
-        .filter_directives(&cfg.log)
+        .filter_directives(&cfg.log);
+
+    if cfg.metrics_enabled {
+        telemetry_conf = telemetry_conf
+            .metrics_addr(&cfg.prometheus_http_listener_addr)
+            .service_name(env!("CARGO_CRATE_NAME"));
```
SuperFluffy, line 29: Should this be configurable so that operators can inject their own global service name to identify metrics?

crates/astria-sequencer-relayer/src/main.rs
```diff
@@ -12,10 +13,19 @@ use tracing::info;
...
+    let mut telemetry_conf = telemetry::configure()
         .set_no_otel(cfg.no_otel)
         .set_force_stdout(cfg.force_stdout)
-        .filter_directives(&cfg.log)
+        .filter_directives(&cfg.log);
+
+    if cfg.metrics_enabled {
+        telemetry_conf = telemetry_conf
+            .metrics_addr(&cfg.prometheus_http_listener_addr)
+            .service_name(env!("CARGO_CRATE_NAME"));
```
SuperFluffy, line 24: Same as for the other services - should this be configurable in case operators want to give their services custom names?

Cargo.toml
```diff
@@ -66,6 +66,7 @@ once_cell = "1.17.1"
 sha2 = "0.10"
 serde = "1"
 serde_json = "1"
+metrics = "0.19.0"
```
SuperFluffy, line 69: Could add a github issue to bump this once the ahash forced requirement is figured out // once penumbra updates.

SuperFluffy (approved):
Looks great. My only question is around potentially configuring the global service names.

For example, in dev cluster we have several sequencer deployments. Does it make sense to give them their own global service name? Or should they carry the the same service name given that they are running the same service?

I don't know how prometheus is set up to discriminate such services, so I guess this is fundamentally a devops question.

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -467,7 +467,7 @@ impl ExecutableBlock {
         let timestamp = convert_tendermint_to_prost_timestamp(block.header().time);
         let transactions = block
             .into_rollup_transactions()
-            .remove(&id)
+            .swap_remove(&id)
```
joroshiba, line 470: Note: dependency bump required this update `remove` was deprecated and is equivalent to `swap_remove`

crates/astria-sequencer-relayer/src/main.rs
joroshiba: Could be a nice future feature, I changed to package name for now.

crates/astria-telemetry/src/lib.rs
joroshiba: Updated to only set if service_name is not empty string.

PR #743, opened 2024-02-12
feat(conductor): set permitted commitment spread from rollup genesis

SuperFluffy:
## Summary
Sets the permitted commitment spread between soft and firm commitment heights to a number derived from the rollup genesis info.

## Background
The spread implemented in the refactoring https://github.com/astriaorg/astria/pull/691 was set to a fixed number of 16 which is very small and not dynamic. This patch takes the spread as the `6 * celestia_block_variance`, with the variance set in the rollup's genesis info. 6 is a heuristic taken as the upper limit of expected sequencer heights per celestia height.

The channel to send soft blocks from sequencer reader to executor was rewritten in terms of a semaphore that acts as a bound to an unbounded mpsc channel. That is done because the bounded tokio mpsc channel (and any other mpsc channel currently found in the ecosystem) does not allow to have its capacity changed dynamically.

## Changes
- refactor the soft block channel in the executor to an unbounded channel bounded by a semaphore
- set the channel's capacity from the celestia block variance times 6
- add permits to the channel if firm blocks have caught up with sequencer blocks
- simplified rescheduling of rejected blocks due to backpressure by leaning on async send methods (instead of putting them back in the cache)
- hid the implementation details of sending blocks to the executor by removing access to the channels in its handler

## Testing
Added unit tests to test the general flow of the channel (initializing, adding permits, exhausting it).


crates/astria-conductor/src/block_cache.rs
```diff
@@ -95,31 +91,6 @@ impl<T: GetSequencerHeight> BlockCache<T> {
...
     }
-
-    /// Reschedules a block at the previous height.
-    ///
-    /// # Errors
-    ///
-    /// Returns an error if:
-    /// + the height of the block is not equal to the previous height recorded in the struct.
-    /// + the height of the block is at level 0.
-    pub(crate) fn reschedule_block(&mut self, block: T) -> Result<(), Error> {
```
SuperFluffy, line 106: Removed because this was replaced by a simpler rescheduling mechanism - schedule the blocks to be sent once there is space in the channel.

crates/astria-conductor/src/celestia/mod.rs
```diff
@@ -214,17 +214,15 @@ impl Reader {
                     break;
                 }
 
-                res = &mut executor_full, if !executor_full.is_terminated() => {
-                    // we just check the error here and drop the permit without using it.
-                    // because the future is now fused the branch fetching the next block will trigger.
+                res = &mut scheduled_block, if !scheduled_block.is_terminated() => {
```
SuperFluffy, line 217: Scheduling the blocks directly instead of putting them back into the cache.

crates/astria-conductor/src/executor/builder.rs
```diff
@@ -52,14 +52,14 @@ impl ExecutorBuilder<WithRollupAddress, WithShutdown> {
         let WithRollupAddress(rollup_address) = rollup_address;
         let WithShutdown(shutdown) = shutdown;
 
-        let (firm_blocks_tx, firm_blocks_rx) = mpsc::channel(16);
-        let (soft_blocks_tx, soft_blocks_rx) = mpsc::channel(16);
+        let (firm_block_tx, firm_block_rx) = mpsc::channel(16);
```
SuperFluffy, line 55: Leaving this at 16 for now. This is just a simple backpressure mechanism that ideally should not even be triggered because executor pulls firm blocks with a higher priority than soft blocks.

crates/astria-conductor/src/executor/channel.rs
```diff
@@ -0,0 +1,194 @@
+//! An mpsc channel bounded by an externally driven semaphore.
+//!
+//! While the main purpose of this channel is to send [`sequencer_client::SequencerBlock`]s
+//! form a sequencer reader to the executor, the channel is generic over the values that are
```
sambukowski, line 4: typo: form should be from

sambukowski (dismissed):

crates/astria-conductor/src/executor/channel.rs
```diff
@@ -0,0 +1,194 @@
...
+        let (tx, _rx) = soft_block_channel::<()>();
+        assert_eq!(
+            tx.try_send(()).unwrap_err(),
+            TrySendError::NoPermits(()),
+            "a fresh channel starts without permits"
+        );
+    }
+
+    #[test]
+    fn permits_are_filled_to_capacity() {
```
joroshiba, line 163: Should this test first expect 1 to be filled, and fail only when at capacity?

crates/astria-conductor/src/executor/channel.rs
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    /// Receives a block over the channel.
+    pub(super) async fn recv(&mut self) -> Option<T> {
+        self.chan.recv().await
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
joroshiba, line 146: This test sweet has all tests which expect `TrySendError::NoPermits(())`. I'd like to have a test for closed as well as a success path.

If I were to replace `try_send` with just return an `TrySendError::NoPermits(())` wouldn't these all pass?

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -226,13 +273,45 @@ impl Executor {
...
-        let next_soft = self.state.borrow().next_soft_sequencer_height().value();
-        let is_too_far_ahead = next_soft.saturating_sub(next_firm) >= 16;
+        let (next_firm, next_soft) = {
+            let state = self.state.borrow();
+            let next_firm = state.next_firm_sequencer_height().value();
+            let next_soft = state.next_soft_sequencer_height().value();
+            (next_firm, next_soft)
+        };
+
+        let is_too_far_ahead = usize::try_from(next_soft.saturating_sub(next_firm))
```
joroshiba, line 311: In what case would this `try_from` not be successful? It seems like this would be a larger issue?

joroshiba (changes requested):

crates/astria-conductor/src/executor/mod.rs
SuperFluffy: If you ran this on a 16bit architecture (or just sub-32bit architecture) this would fail. I think it's fair to assume that we are running on at least 32 bit (and even that is unrealistic nowadays) so that there is an `expect` right after.

crates/astria-conductor/src/executor/channel.rs
SuperFluffy: This was a good call. Adding tests to check for the channel being closed revealed that the `Closed` variant was never hit because the channel senders held an `Arc`. I have downgraded the senders to hold a `Weak` instead, and am also explicitly closing the semaphore when dropping the receiver. I have also added tests for the immediate `try_send`, the async `send`, and the `recv` methods.

crates/astria-conductor/src/executor/channel.rs
SuperFluffy: I am not sure I understand. Can you give me an example?

joroshiba: Nevermind, misread the test.

joroshiba (approved):

PR #749, opened 2024-02-13
fix(sequencer-relayer): re-queue blocks that fail to submit

### LINKED ISSUE #745: sequencer-relayer: re-queue and submit blobs that fail to submit
as title says - we currently don't re-queue and re-submit blobs that fail to submit to DA, resulting in data is that is never made available, which is bad.

noot:
## Summary
 re-queue blocks that fail to submit to celestia.

not sure if i like this impl as returning the blocks in the error field of `Result` seems jank but maybe it's fine for now? going to give it some thought.

## Background
it's really bad if we drop blocks completely, as they will never get written to celestial, which is what was happening.

## Changes
-  re-queue blocks that fail to submit to celestia

## Testing
not yet - can try running in dev cluster?

## Related Issues

closes #745


joroshiba:
Generally in support of this, but we probably shouldn't do this without also adding a max number of bytes per celestia blob pulled from queue. Otherwise a failure could result in giant transactions if it goes for too long.

noot:
@joroshiba i made https://github.com/astriaorg/astria/issues/751 for that - should i address them in the same PR? or i can do a PR that merges into this one?

crates/astria-sequencer-relayer/src/relayer.rs
```diff
@@ -359,11 +373,11 @@ impl Relayer {
             // layer if no submission is in flight.
             if !self.queued_blocks.is_empty() && self.submission_task.is_none() {
                 let client = self.data_availability.clone();
+                let queued_blocks = std::mem::replace(&mut self.queued_blocks, vec![]);
                 self.submission_task = Some(task::spawn(submit_blocks_to_celestia(
                     client,
-                    self.queued_blocks.clone(),
+                    queued_blocks,
```
SuperFluffy, line 377: If there are issues with endpoint this will effectively be a hot loop that:

1. spawns the task
2. immediately fails
3. respawns

Depending on the setup of the other futures in the select! statement it might be that this drowns out all other branches.

SuperFluffy (changes requested):
Requesting changes because this is effectively spinning on the resubmission in the worst case which can lead to unexpected runtime behaviour.

It's best to reimplement this in terms of a backoff, for example using `tryhard`

joroshiba:
> @SuperFluffy agreed, I was originally thinking of waiting for the next celestia block but a backoff also seems fine, thoughts? I'll implement the backoff for now

I think if we implement a backoff retry instead of adding back onto queue that will be fine for first PR. The other was just going to immediately follow into "might make all future requests fail". Still possible if queue grows too large during retries.

We should emit an error and probably a metric counter upon failure as well here. Maybe count submission attempts for a given bundle via bundle via metric? Reset to 0 on first send. Do this probably as a gauge.

SuperFluffy:
@noot 
> @SuperFluffy agreed, I was originally thinking of waiting for the next celestia block but a backoff also seems fine, thoughts? I'll implement the backoff for now

I think a backoff provides an immediate fix to the issue. However, with a potentially long-running backoff we immediately run into the issue of backpressure: right now we'd keep accumulating sequencer blocks until we are OOM.

I think this warrants a refactoring of sequencer-relayer to bring it closer to conductor (and composer) in functionality.

noot:
closing for now as @SuperFluffy will take this on

PR #752, opened 2024-02-14
feat(sequencer): implement ability to update fee assets using sudo key

### LINKED ISSUE #731: sequencer: configure allowed fee assets with a sudo key
follow up to https://github.com/astriaorg/astria/pull/730

noot:
## Summary
as title says, implement ability to update fee assets using sudo key.

## Background
need to be able to update the assets allowed for fee payment while the chain is running.

## Changes
- add `FeeAssetChangeAction` which represents the addition or deletion of an authorized fee asset
- ensure that at least one asset has to be allowed for fee payment (should we make it impossible to delete the default asset? probably?)

## Testing
unit tests

## Related Issues

closes #731


joroshiba (approved):

joroshiba:
to fix proto format `just fmt proto` should work.

PR #753, opened 2024-02-14
feat: print build info in all services

SuperFluffy:
## Summary
Print build info for all services.

## Background
Astria services emit no build information which means we are potentially flying blind.

This patch injects various bits of build information using our services' `build.rs` build scripts and reads them out at compile time. The injected build information is then printed on stderr immediately after startup following the convention that informational and debug messages are not sent to stdout.

The emitted build information is json formatted and looks like so (for sequencer compiled locally and ran through `jq`. It's not actually pretty-printed):
```json
{
  "build_timestamp": "2024-02-14T14:13:18.583116000Z",
  "cargo_opt_level": "0",
  "cargo_pkg_name": "astria-sequencer",
  "cargo_target_triple": "aarch64-apple-darwin",
  "git_branch": "superfluffy/version-and-sha-into-services",
  "git_commit_date": "2024-02-14",
  "git_describe": "sequencer-v0.8.0-25-g268bbba",
  "git_sha": "268bbba35155fcf2f218a52afda20557dad8f294",
  "rustc_channel": "stable",
  "rustc_commit_hash": "07dca489ac2d933c78d3c5158e3f43beefeb02ce",
  "rustc_host_triple": "aarch64-apple-darwin"
}
```

## Changes
- Add a new crate `astria-build-info` to emit and inject build information at compile time
- Update all services to print build info.
- Set the minimum version requirement of `ibc-types` to `0.11.1` because of conflicts in the version resolution of the `time` dependency (a mutual dependency of `vergen` and `ibc-types`).

## Testing
Nothing to test as this is fundamentally a thin wrapper around the functionality provided by the `vergen` crate. All services, composer, conductor, sequencer, and sequencer-relayer, were executed locally and observed to emit the expected build info.


crates/astria-build-info/README.md
```diff
@@ -0,0 +1,51 @@
...
+    astria_build_info::emit("<release-tag-of-service>")?;
+    Ok(())
+}
+```
+
+And pick up the emitted variables like so:
+
+```rust,no_run
+use astria_build_info::BuildInfo;
+const BUILD_INFO: BuildInfo = astria_build_info::get!();
```
SuperFluffy, line 50: This doctest is failing because the required env vars are not injected into the doctest's build. I am going to manually inject the env vars to fix this test.

crates/astria-build-info/src/lib.rs
```diff
@@ -0,0 +1,86 @@
...
+///
+/// Note that if two services share a common prefix like `sequencer` and
+/// `sequencer-relayer`, `<prefix>` should be supplied as `"sequencer-v"` and
+/// `"sequencer-relayer-v"` for tags like `sequencer-v1.2.3` or
+/// `sequencer-relayer-v1.2.3`. This is to avoid matching on the wrong prefix
+///
+/// # Errors
+/// Emits the same errors as [`vergen::EmitBuilder::emit`].
+pub fn emit(prefix: &str) -> Result<(), Box<dyn std::error::Error>> {
+    let git_describe_prefix = Box::leak(format!("{prefix}*").into_boxed_str());
```
joroshiba, line 71: Should we just add the -v here instead of requiring each package add it?

crates/astria-conductor/build.rs
```diff
@@ -0,0 +1,4 @@
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    astria_build_info::emit("conductor-v")?;
```
joroshiba, line 2: Should we just use `env!("CARGO_PKG_NAME")`?

SuperFluffy: We are using `astria-conductor` for pkg name but `conductor-vx.x.x` for the tag so that doesn't work unfortunately.

crates/astria-build-info/src/lib.rs
SuperFluffy: I considered this initially but since the output of the build scripts is hidden from maintainers this is marginally clearer (although build scripts are not in a direct line of sight either, so this benefit is questionable).

I don't have strong opinions either way. Do you prefer adding it in the method?

joroshiba: Long term it's probably more correct to have to specify the `-v`. Fundamentally the tag chosen is a part of our github action workflow. I don't love that configuration which comes out of our github action build process get's injected here. Let's keep as is for now.

Not required before merge, but long term: My thinking is that we shouldn't have to specify anything here. It defaults to the name of the pkg/binary being built, but is overridable with an environment variable. We would set this in our Docker & release builds. Say `RUST_BUILD_INFO_TAG_PREFIX`? Then in reusable-build/reusable-docker we can set these environment variables so the code that picks the tags also sets the prefix here?

joroshiba (approved):

PR #754, opened 2024-02-14
chore: add some information to crates update msrv

joroshiba:
## Summary
Updates all crate cargo tomls to have some more metadata + correct rust-versions

## Background
We didn't have msrv set on some packages, we didn't have information about package ownership and sometimes correct licenses.

## Changes
- checked + updated all cargo rust versions
- added links to readmes where appropriate
- consistent licenses, repo, website

## Testing
ran cargo-msrv on all crates.


SuperFluffy (approved):
Thank you for going through all of these. The versions seem oddly recent, but apparently cargo has some issues withh the `dep:foo` and `foo/feat` syntax. We have a few dependencies that make use of this syntax and we also use it ourselves. Apparently it is resolved in newer versions of Rust (or cargo?).

PR #755, opened 2024-02-14
chore: rename folders to not all start with astria

joroshiba:
## Summary
Removes `astria-` from the beggining of each crate folder in the repo. Note this will affect open PRs and branches, but the code is otherwise identical and generated packages are identical.

## Background
Every crate produced by us starts with `astria-` this is useful for packaging for annoying for browsing the code, IMO.

## Changes
- Change folder references and folders to remove `astria-`
- Update docker and general build scripts to make work with new naming

## Testing
CI/CD tests


joroshiba:
After further discussion, closing and killing.

The goal had been to align folders with alias names, preference to migrate to not aliasing over the folder renaming.

PR #761, opened 2024-02-16
fix(conductor): try adding path, scheme to Celestia RPC URL

SuperFluffy:
## Summary
If missing, adds a dummy scheme and/or empty path, to the provided celestia RPC URI if missing.

## Background
Conductor fails when trying to construct Celestia endpoints from URIs like `astria.org` (set via the env var `ASTRIA_CONDUCTOR_CELESTIA_NODE_URL`) that contain only an authority but no scheme nor path with errors like this:
```
2024-02-16T16:21:08.212985Z ERROR astria_conductor: failed initializing conductor error=failed constructing data availability reader error.sources=[failed constructing websocket endpoint from provided celestia endpoint URL, path missing]
```

On the other hand, the following strings would work: `http://astria.org`, `http://astria.org/` or `wss://astria.org/` (notice that the first input lacks both scheme and path, and the latter are trailed by a `/` "path").

## Changes
- Set `/` as the URI path if missing.
- Set URI scheme if missing
- Provide tests to check various input scenarios

## Testing
Provided new tests. Ran conductor in dev cluster with a value for `ASTRIA_CONDUCTOR_CELESTIA_NODE_URL` that was previously broken and now works.

### DIFF
--- a/crates/astria-conductor/src/celestia/builder.rs
+++ b/crates/astria-conductor/src/celestia/builder.rs
@@ -335,9 +335,9 @@ impl FromStr for Endpoints {
             .wrap_err("failed to parse as URI")?;
         let is_tls = matches!(uri.scheme().map(Scheme::as_str), Some("https" | "wss"));
 
-        let http = make_uri(uri.clone(), EndpointKind::Http, is_tls)
+        let http = make_uri(&uri, EndpointKind::Http, is_tls)
             .wrap_err("failed constructing http endpoint from parsed URI")?;
-        let websocket = make_uri(uri.clone(), EndpointKind::Ws, is_tls)
+        let websocket = make_uri(&uri, EndpointKind::Ws, is_tls)
             .wrap_err("failed constructing websocket endpoint parsed URI")?;
 
         Ok(Self {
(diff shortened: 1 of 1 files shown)


crates/astria-conductor/src/celestia/builder.rs
```diff
@@ -319,14 +308,226 @@ fn wss_scheme() -> Scheme {
...
+    type Err = eyre::Report;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        // The docs promise that inputs like `<authority>/<scheme>` work,
+        // even though that is not a valid URI because it's lacking a scheme.
+        // If parsing `s` as a URI fails the code below prepends the input with
+        // `dummy://{s}` and tries parsing that instead. If that also fails, the
+        // error of the first parse is returned.
+        //
+        // XXX: This is a hack to not have a breaking change in the example config.
```
SuperFluffy, line 327: Flagging this. We should probably just not do this and instead just have 2 env vars, one for HTTP, one for WS.

noot: yeah that would be better

crates/astria-conductor/src/celestia/builder.rs
```diff
@@ -319,14 +308,226 @@ fn wss_scheme() -> Scheme {
...
+        //      However, this is pretty janky and potentially confusing to the operator.
+        //      Best to have two variables for the Celestia HTTP and WS RPC endpoints.
+        let uri = s
+            .parse::<Uri>()
+            // Insert a meaningless scheme here because it will be replaced later.
+            // If parsing this fails, the error of the initial parse is returned.
+            .or_else(|e| format!("dummy://{s}").parse::<Uri>().or(Err(e)))
+            .wrap_err("failed to parse as URI")?;
+        let is_tls = matches!(uri.scheme().map(Scheme::as_str), Some("https" | "wss"));
```
noot, line 336: does celestia require https/wss?

noot (approved):

SuperFluffy:
Closing this in favor of https://github.com/astriaorg/astria/pull/780. This PR is unncessarily complex.

PR #765, opened 2024-02-18
feat(telemetry): simplify emitting error fields with cause chains

SuperFluffy:
## Summary
Installs a global eyre error hook to format all eyre reports with cause chains.

## Background
By default display-formatting of errors only prints the top-most error, while a debug formatting also prints the sources (either using Rust debug formatting or in the case of eyre or anyhow as human-readable strings). This however does not mix well with structured logging and observability because there events are expected to be machine-readable.

Tracing indeed implements `tracing::field::Value for dyn std::error::Error`, but that requires the author to remember to cast errors to this trait object, either with `&e as &dyn std::error::Error` for typed errors or `AsRef::<dyn std::error::Error>::as_ref(&err)` for eyre or anyhow reports/errors. This is both annoying and can be easily forgotten.

This patch introduces the `astria-eyre` crate that provieds an `install()` hook, which ensures that all `eyre::Report` types are printed as JSON-formatted strings with their complete cause-chain.

Unfortunately this does not automatically give full cause-chains for typed errors (unless they are wrapped in an eyre Report first), and it also does not help with sequencer because it uses anyhow, for which no global formatting hook exist. But it still makes life easier for the other Astria services for which we control error handling.

## Changes
- add a `astria-eyre` crate exposing an `install()` function that installs a global formatting hook for eyre reports
- re-write composer, conductor, sequencer-relayer in terms of it
- change `err_report.as_ref()` where appropriate, leave `&err as &dyn StdError` untouched

## Testing
Not applicable for services (need to observe the changes while running the service).


crates/astria-eyre/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "astria-eyre"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+eyre = { workspace = true }
```
noot, line 9: do any other crates depend on `eyre` now? should this be non-workspace?

SuperFluffy: Updated. `eyre` is no longer a workspace dependency.

joroshiba (approved):

noot (approved):
nice!

PR #769, opened 2024-02-27
feat(relayer): retry sequencer block fetch, celestia blob submission

SuperFluffy:
## Summary
Adds retries for fetching blocks from sequencer and submitting blobs to Celestia. Also refactors state-tracking and simplifies tests.

## Background
Relayer immediately failed when fetching blocks from sequencer or submitting blobs to Celestia failed, leading to critical loss of data. This patch adds retry logic to both up to `u32::MAX` times, which is effectively infinity.

Because very long retries means that the internal block/blob cache can grow infinitely this patch also adds a backpressure mechanism: if the task writing blobs to Celestia is congested reading blocks from Sequencer stops until there is more capacity.

The change also affected how readiness and health are tracked, requiring a refactor (and simplification) of tests.

## Missing

This patch does not include, leaving to followup PRs:
1. a limiter on the maximum blob size (both individual blobs and across all blobs submitted in one `blob.Submit` RPC);
2. state-tracking to ensure that already submitted blobs will not be resubmitted.

## Changes
- Add retry logic to sequencer block fetch
- Add retry logic to celestia blob submission
- Add a backpressure mechanism to halt block fetching if blob submission is congested
- Refactor status reporting taking the above changes into account
- Rework and simplify tests

## Testing
Reworked and simplified tests to ensure that blobs still arrive.

## Related Issues

closes https://github.com/astriaorg/astria/issues/745


crates/astria-celestia-client/src/client.rs
```diff
@@ -171,7 +178,7 @@ pub trait CelestiaClientExt: BlobClient {
         Ok(rollup_datas)
     }
 
-    /// Submits sequencer `blocks` to celestia
+    /// Submits a sequencer `block` to celestia
     ///
     /// `Blocks` after converted into celestia blobs and then posted. Rollup
```
noot, line 183: should this say `the block is converted...`?

crates/astria-sequencer-relayer/src/relayer/mod.rs
```diff
@@ -0,0 +1,304 @@
...
+            make_latest_height_stream(self.sequencer.clone(), self.sequencer_poll_period);
+        pin!(latest_height_stream);
+
+        let (submitter_task, submitter) =
+            spawn_submitter(self.celestia.clone(), self.state.clone());
+
+        let mut block_stream = read::BlockStream::new(self.sequencer.clone(), self.state.clone());
+        block_stream.set_block_time(self.sequencer_poll_period);
+
+        let mut forward_once_free: Fuse<
```
noot, line 142: what is this for exactly? could you add a comment?

crates/astria-sequencer-relayer/src/relayer/mod.rs
```diff
@@ -0,0 +1,304 @@
...
+        > = Fuse::terminated();
+
+        self.state.set_ready();
+
+        let reason = loop {
+            select!(
+                biased;
+
+                res = &mut forward_once_free, if !forward_once_free.is_terminated() => {
+                    block_stream.resume();
```
noot, line 153: should this resume after the `if` block?

crates/astria-sequencer-relayer/src/relayer/mod.rs
```diff
@@ -0,0 +1,304 @@
...
+                    // XXX: exiting because submitter only returns an error after u32::MAX
+                    // retries, which is practically infinity.
+                    if res.is_err() {
+                        let err = res.wrap_err(
+                            "submitter exited unexpectly while trying to forward block"
+                        );
+                        break err;
+                    }
```
noot, line 161: the errors returned here can be the blob submit channel closed, or it's at capacity, right? are there any other cases?

crates/astria-sequencer-relayer/src/relayer/read.rs
```diff
@@ -0,0 +1,218 @@
...
+    pub(super) fn set_block_time(&mut self, block_time: Duration) {
+        self.block_time = block_time;
+    }
+
+    pub(super) fn set_latest_sequencer_height(&mut self, height: Height) {
+        self.heights.last_observed.replace(height);
+    }
+
+    pub(super) fn pause(&mut self) {
+        debug!("stream paused");
```
noot, line 93:
```suggestion
        debug!("block stream paused");
```

same for resume

crates/astria-sequencer-relayer/src/relayer/read.rs
```diff
@@ -0,0 +1,218 @@
...
+            client,
+            heights: Heights::default(),
+            future: None,
+            block_time: Duration::from_millis(1_000),
+            paused: false,
+            state,
+        }
+    }
+
+    pub(super) fn set_block_time(&mut self, block_time: Duration) {
```
noot, line 84: why have this over setting it in `new`?

crates/astria-sequencer-relayer/src/relayer/write.rs
```diff
@@ -0,0 +1,366 @@
...
+    ) -> Result<(), SendError<SequencerBlock>> {
+        self.tx.send(block).await
+    }
+}
+
+pub(super) struct BlobSubmitter {
+    // The client to submit blobs to Celestia.
+    client: HttpClient,
+
+    // The channel over which sequencer blocks to receive.
```
noot, line 135:
```suggestion
    // The channel over which sequencer blocks are received.
```

crates/astria-sequencer-relayer/src/relayer/write.rs
```diff
@@ -0,0 +1,366 @@
...
+    blocks: Receiver<SequencerBlock>,
+
+    // The maximum number of conversions that can be active at the same time.
+    // Submitter will refuse to accept more blocks until there are fewer active
+    // conversions.
```
noot, line 140: what happens to blocks that are refused? are they resent to the submitter later?

crates/astria-sequencer-relayer/tests/blackbox/main.rs
```diff
@@ -61,116 +43,95 @@ async fn one_block_is_relayed_to_celestia() {
...
-    if let Some(blobs_seen_by_celestia) = sequencer_relayer
-        .celestia
-        .state_rpc_confirmed_rx
-        .recv()
-        .await
-    {
-        assert_eq!(6, blobs_seen_by_celestia.len());
-    }
+    let blobs_seen = timeout(
+        // timeout after (3 + 1) block times to ensure that 3 blocks are definitley picked up
```
noot, line 126:
```suggestion
        // timeout after (3 + 1) block times to ensure that 3 blocks are definitely picked up
```

crates/astria-sequencer-relayer/src/relayer/write.rs
```diff
@@ -0,0 +1,366 @@
...
+            client
+                .blob_submit(
+                    &blobs,
+                    SubmitOptions {
+                        fee: None,
+                        gas_limit: None,
+                    },
+                )
+                .await
+                .wrap_err("failed submitting sequencer blocks to celestia")
```
noot, line 315: for a future PR, maybe we should check what kind of error is returned, eg a nonce error or other error, and handle based on that

crates/astria-sequencer-relayer/src/relayer/write.rs
```diff
@@ -0,0 +1,366 @@
...
+                }
+
+                Some((sequencer_height, conversion_result)) = self.conversions.join_next() => {
+                     match crate::utils::flatten(conversion_result) {
+                        // XXX: Emitting at ERROR level because failing to convert contitutes
+                        // a fundamental problem for the relayer, even though it can happily
+                        // continue chugging along.
+                        // XXX: Should there instead be a mechanism to bubble up the error and
+                        // have sequencer-relayer return with an error code (so that k8s can halt
+                        // the chain)? This should probably be part of the protocal/sequencer proper.
```
noot, line 198:
```suggestion
                        // the chain)? This should probably be part of the protocol/sequencer proper.
```

yeah, failing to convert would be very bad - in what cases would that ever happen?

crates/astria-sequencer-relayer/src/relayer/mod.rs
```diff
@@ -0,0 +1,304 @@
...
+                    if let Err(err) = self.forward_block_for_submission(
+                        height,
+                        block,
+                        &mut block_stream,
+                        submitter.clone(),
+                        &mut forward_once_free,
+                    ).wrap_err("submitter exited unexpectly while trying to forward block") {
+                        // XXX: exiting because submitter only returns an error after u32::MAX
+                        // retries, which is practically infinity.
+                        break Err(err);
```
noot, line 201: i don't see the retry in `forward_block_for_submission`?

noot (commented):
mostly looks good, just have a question about `forward_block_for_submission`/ what happens if the submitter is at capacity

crates/astria-sequencer-relayer/src/relayer/mod.rs
SuperFluffy: I shifted it after the if-block. Agree that it might be confusing to readers. Note though that it practically makes no difference (since it only affects the block stream in the next iteration of the loop).

crates/astria-sequencer-relayer/src/relayer/read.rs
SuperFluffy: Added and also moved the debug logs out of the methods to the places that actually call the methods.

crates/astria-sequencer-relayer/src/relayer/write.rs
```diff
@@ -0,0 +1,366 @@
...
+            blobs: Vec::new(),
+        }
+    }
+
+    fn push(&mut self, mut blobs: Vec<Blob>, height: SequencerHeight) {
+        self.blobs.append(&mut blobs);
+        self.heights.push(height);
+    }
+
+    fn take(&mut self) -> TakeBlobs<'_> {
```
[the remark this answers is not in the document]
SuperFluffy: Yes, I added some text to `QueuedBlobs::take`. We have this construction elsewhere in the code. The reason for having a future here (instead of something like `futures::ready(queued_blobs.take_immediately())`, where `take_immediately` is eagerly, not lazily, moving the blobs out of the queue) is async cancellation.

crates/astria-sequencer-relayer/src/relayer/write.rs
SuperFluffy: Yes, have a look at the usage of `try_send` vs `send` in the relayer select-loop. It always `try_send`s a block first. If that returns an error it immediately schedules a future with `send(the_same_block)` and pauses the block stream.

crates/astria-sequencer-relayer/src/relayer/write.rs
SuperFluffy: We should. Unfortunately Celestia didn't spec that out anywhere which makes this difficult. :(

crates/astria-sequencer-relayer/src/relayer/mod.rs
SuperFluffy: Yeah, I can see how that comment is confusing. I have changed the comment to talk about how there is no logic to restart the task to which we are forwarding the blocks.

noot (approved):
looks great :)

PR #771, opened 2024-02-28
fix(relayer): add metrics

SuperFluffy:
## Summary
Adds metrics to relayer.

## Background
Some metrics were not ported when refactoring relayer in https://github.com/astriaorg/astria/pull/769, some other metrics did not make sense where they were.

## Changes
- Add new metrics to relayer
- 
## Metrics

- `astria_sequencer_relayer_blobs_per_celestia_tx`
- `astria_sequencer_relayer_celestia_submission_latency`
- `astria_sequencer_relayer_sequencer_block_fetch_failure_count`
- `astria_sequencer_relayer_sequencer_height_fetch_failure_count`

### DIFF
--- a/crates/astria-sequencer-relayer/src/metrics_init.rs
+++ b/crates/astria-sequencer-relayer/src/metrics_init.rs
@@ -14,27 +14,46 @@ pub fn register() {
     celestia_client::metrics_init::register();
 
     describe_counter!(
-        CELESTIA_SUBMISSION_HEIGHT,
-        Unit::Count,
-        "The height of the last blob submitted to Celestia"
-    );
-
-    describe_counter!(
         CELESTIA_SUBMISSION_COUNT,
         Unit::Count,
         "The number of calls made to submit to celestia"
     );
 
     describe_counter!(
-        CELESTIA_SUBMISSION_COUNT,
+        CELESTIA_SUBMISSION_HEIGHT,
+        Unit::Count,
+        "The height of the last blob submitted to Celestia"
+    );
+
+    describe_counter!(
+        CELESTIA_SUBMISSION_FAILURE_COUNT,
         Unit::Count,
         "The number of calls made to submit to celestia which have failed"
+    );
+
+    describe_counter!(
+        SEQUENCER_BLOCK_FETCH_FAILURE_COUNT,
+        Unit::Count,
+        "The number of calls made to fetch a block from sequencer which have failed"
+    );
+
+    describe_counter!(
+        SEQUENCER_HEIGHT_FETCH_FAILURE_COUNT,
+        Unit::Count,
+        "The number of calls made to fetch the current height from sequencer which have failed"
     );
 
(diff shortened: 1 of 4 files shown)


crates/astria-sequencer-relayer/src/metrics_init.rs
```diff
@@ -13,28 +13,47 @@ use metrics::{
...
+
     describe_counter!(
         CELESTIA_SUBMISSION_HEIGHT,
         Unit::Count,
         "The height of the last blob submitted to Celestia"
     );
 
     describe_counter!(
-        CELESTIA_SUBMISSION_COUNT,
+        CELESTIA_SUBMISSION_FAILURE_COUNT,
```
SuperFluffy, line 29: The diff looks weird, but the point here is that `CELESTIA_SUBMISSION_COUNT` was described twice (instead of `CELESTIA_SUBMISSION_COUNT` and `CELESTIA_SUBMISSION_FAILURE_COUNT`)

crates/astria-sequencer-relayer/src/relayer/read.rs
```diff
@@ -185,6 +185,9 @@ async fn fetch_block(
         .max_delay(block_time)
         .on_retry(
             |attempt: u32, next_delay: Option<Duration>, error: &eyre::Report| {
+                metrics::counter!(crate::metrics_init::SEQUENCER_BLOCK_FETCH_FAILURE_COUNT)
```
SuperFluffy, line 188: This counter only makes sense in the context of the retry future since the fetch fails hard (after `u32::MAX` retries).

crates/astria-sequencer-relayer/src/relayer/write.rs
```diff
@@ -277,16 +301,16 @@ async fn submit_with_retry(
         .max_delay(Duration::from_secs(12))
         .on_retry(
             |attempt: u32, next_delay: Option<Duration>, error: &eyre::Report| {
+                metrics::counter!(crate::metrics_init::CELESTIA_SUBMISSION_FAILURE_COUNT)
```
SuperFluffy, line 304: Similar to counting the number of failed celestia submits, would this make sense as a gauge?

crates/astria-sequencer-relayer/src/relayer/read.rs
joroshiba: Agreed an additional guage would make sense here.

crates/astria-sequencer-relayer/src/relayer/write.rs
joroshiba: As above I think having a counter & a guage makes sense.

joroshiba (approved):
Approved as is, I think adding guages for the retry loops would be good as well so we can track total failures & failures per retry loop.

SuperFluffy:
I will add gauges in a follow-up PR.

PR #772, opened 2024-03-01
feat(conductor): assert host fulfills execution API contract

SuperFluffy:
## Summary
Check that the host response to `astria.execution.v1alpha2.ExecuteBlock` fulfills the contract (namely, that the number in the response is +1 the current number tracked by conductor).

## Background
Conductor implicitly checked the contract by enforcing that the blocks received from sequencer and/or celestia have the expected rollup block number. There was no test however that the execution API server (i.e. the rollup node) responds with the expected block number. This lead to confusing error logs.

## Changes
- Assert that the host response to the `astria.execution.v1alpha2.ExecuteBlock` RPC is as expected.

## Testing
Added unit tests to check that the assertion is correct.

### DIFF
--- a/crates/astria-conductor/src/executor/mod.rs
+++ b/crates/astria-conductor/src/executor/mod.rs
@@ -344,6 +344,13 @@ impl Executor {
             .await
             .wrap_err("failed to execute block")?;
 
+        does_block_response_fulfill_contract(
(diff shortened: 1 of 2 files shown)


joroshiba (approved):
LGTM

PR #775, opened 2024-03-02
chore(conductor)!: remove all optimism functionality

SuperFluffy:
## Summary
Removes all functionality to interact with optimism rollups from the monorepo.

## Background
There is no astria deployment that uses optimism, yet having this functionality in conductor carries a maintenance burden. Prior to the upcoming it makes sense to remove this functionality. It will be brought back at a later point.

## Changes
- remove the astria-optimism crate
- remove all optimism related functionality from conductor
- adjust CI to not run steps required to set up binaries required to run optimism tests

## Testing
N/A. Functionality is removed.

## Breaking Changelist
- This is technically a breaking change because functionality is removed. But in practice there are no services/rollups that make use of this.


aajimal (approved):

joroshiba (approved):
Can we get this into a feature branch as well?

SuperFluffy:
Created a draft PR to revert these changes here: https://github.com/astriaorg/astria/pull/781

PR #778, opened 2024-03-04
Superfluffy/stateful relayer

SuperFluffy:
## Summary

## Background

## Changes

## Testing
How are these changes tested?

## Metrics

## Breaking Changelist

## Related Issues

closes


crates/astria-sequencer-relayer/tests/blackbox/main.rs
```diff
@@ -42,9 +47,57 @@ async fn one_block_is_relayed_to_celestia() {
...
+        .await
+    else {
+        panic!("celestia must have seen blobs")
+    };
+    // We can reconstruct the individual blobs here, but let's just assert that it's
+    // two blobs for now: one transaction in the original block + sequencer namespace
+    // data.
+    assert_eq!(blobs_seen_by_celestia.len(), 2);
+
+    // XXX: Waiting in async tests is generally. Fix this by providing different means of
```
SuperFluffy, line 84: I will create an issue for this.

crates/astria-sequencer-relayer/local.env.example
joroshiba, line 1: Could we make it such that a non existent file would be treated the same as an empty file? Allowing it to be created by the relayer?

Makes startup more awkward to say "go create two files".

Documentation on the data held within?

crates/astria-sequencer-relayer/src/relayer/mod.rs
```diff
@@ -273,13 +296,63 @@ impl Relayer {
...
+        match (presubmit_state, postsubmit_state) {
+            (None, None) => Ok(None),
+            (Some(pre), Some(post)) if write::are_states_consistent(pre, post) => {
+                let determined_height = post.sequencer_height();
+                let height = determined_height.try_into().wrap_err_with(|| {
+                    format!(
+                        "last submitted sequencer height was determined to be \
+                         `{determined_height}`, but couldn't be turned into cometbft/sequencer \
+                         height"
+                    )
```
joroshiba, line 323: Is this comment correct? Wouldn't another primary these get out of sync be that crashed while sending the data to celestia? IE submission started but response not received, so no clear idea of if got submitted or not?

crates/astria-sequencer-relayer/local.env.example
```diff
@@ -45,6 +45,18 @@ ASTRIA_SEQUENCER_RELAYER_RELAY_ONLY_VALIDATOR_KEY_BLOCKS=false
...
+# A file must exist at this path and the file must be readable and writable.
+# The file must either be empty or contain a JSON object that can be read by
+# by Relayer (this is the case if it was written to by sequencer relayer).
+ASTRIA_SEQUENCER_RELAYER_PRESUBMIT_PATH=/path/to/presubmit.json
+
+# The path to which relayer will write its state after submitting to Celestia.
+# A file must exist at this path and the file must be readable and writable.
+# The file must either be empty or contain a JSON object that can be read by
+# by Relayer (this is the case if it was written to by sequencer relayer).
+ASTRIA_SEQUENCER_RELAYER_POSTSUBMIT_PATH=/path/to/postsubmit.json
```
noot, line 58: what's the reasoning for having two different files for the states? why not one file of the current state?

crates/astria-sequencer-relayer/src/relayer/mod.rs
```diff
@@ -273,13 +296,63 @@ impl Relayer {
...
+                })?;
+                Ok(Some(height))
+            }
+            (..) => bail!(
+                "manual intervention required: pre- and post-submit states must either both be \
+                 unset (their files present but empty),
+                    or both be set and consistent (files present and deserializable, equal \
+                 sequencer heights, celestia heights pre-submission less
+                    then post-submission)"
```
noot, line 332: the files have to exist before relayer is started for the first time? or does relayer init the files?

crates/astria-sequencer-relayer/src/relayer/read.rs
```diff
@@ -142,21 +135,24 @@ impl Stream for BlockStream {
...
-            .try_into()
-            .expect(
-                "height differences should convert to usize on all reasonable architectures; 64 \
-                 bit: heights are represented as u64; 32 bit: cometbft heights are non-negative \
-                 i64",
-            );
+        let next_height = self.heights.next.value();
+        let mut lower_limit = latest_height.saturating_sub(next_height);
+
+        // A new future will be spawned as long as next_heigth <= latest_height. So
```
noot, line 141:
```suggestion
        // A new future will be spawned as long as next_height <= latest_height. So
```

crates/astria-sequencer-relayer/src/relayer/mod.rs
SuperFluffy: This line is just trying to convert a u64 to a i64. The if-guard of the match arm handles consistency - `write::are_states_consistent`

crates/astria-sequencer-relayer/local.env.example
SuperFluffy: This is a common pattern - if you use one file and you get a corruption (service crashes/gets killed while writing, etc) you lose the entire file. By separating this you will retain the old state (if writing `presubmit` failed) or retain information about what you tried to submit.

Thinking about it, `presubmit` should actually also store the last submitted state, because losing that is arguably worse.

crates/astria-sequencer-relayer/local.env.example
SuperFluffy: So this is actually a feature to force manual intervention. Otherwise, if the files at the `presubmit_path` and `postsubmit_path` did not exist, then relayer would automatically try and start relaying everything beginning from sequencer block 1 upon restart.

Assume you are an operator and set up everything correctly but you have a typo in the filepath (or you set up the volumes incorrectly), then that might be a very costly mistake.

SuperFluffy:
Expanded `local.env.example` to explain the possible values of the pre/post submit files.

Required the files to not only be present, but always initialized with valid values.

Reworked the submission state tracking to be more explicit in its types.

Added a ton of tests to test various consistency failures when reading state from disk.

crates/astria-sequencer-relayer/src/relayer/read.rs
```diff
@@ -25,35 +25,37 @@ use sequencer_client::{
...
     Instrument as _,
     Span,
 };
 
-#[derive(Default, Debug)]
+#[derive(Debug)]
 struct Heights {
     last_observed: Option<Height>,
-    last_requested: Option<Height>,
+    next: Height,
```
noot, line 38: could you add a comment for this?

crates/astria-sequencer-relayer/src/relayer/write.rs
```diff
@@ -378,6 +472,62 @@ impl<'a> std::fmt::Display for ReportSequencerHeights<'a> {
     }
 }
 
+/// Currently running conversions of Sequencer blocks to Celestia blobs.
+///
+/// The conversion result will be returned in the order they are pushed
+/// into this queue.
+///
+/// Note on the implementation: the conversions are done in a block tokio
```
noot, line 480:
```suggestion
/// Note on the implementation: the conversions are done in a blocking tokio
```

crates/astria-sequencer-relayer/src/relayer/write.rs
```diff
@@ -189,18 +202,24 @@ impl BlobSubmitter {
...
 
+        // Fetch the latest Celestia height so there is a starting point when tracking
+        // submissions.
+        fetch_latest_celestia_height(self.client.clone(), self.state.clone())
+            .await
+            .wrap_err("failed fetching latest celestia height after many retries")?;
+
         loop {
             select!(
                 Some(block) = self.blocks.recv(), if self.has_capacity() => {
```
noot, line 213: just double checking - `if self.has_capacity()` is called/checked before `self.blocks.recv()` right? so if there is no capacity, `block` will not be taken out of the channel and dropped, but only taken out when there is capacity?

crates/astria-sequencer-relayer/src/relayer/write.rs
```diff
@@ -378,6 +472,62 @@ impl<'a> std::fmt::Display for ReportSequencerHeights<'a> {
...
+    }
+
+    fn has_capacity(&self) -> bool {
+        self.active.len() < self.max_conversions
+    }
+
+    fn push(&mut self, block: SequencerBlock) {
+        let height = block.height();
+        let fut = async move {
+            let res = tokio::task::spawn_blocking(move || convert(block)).await;
```
noot, line 509: does this still start the task immediately even if it's inside an `async move` scope?

crates/astria-sequencer-relayer/tests/blackbox/main.rs
noot:
```suggestion
    // XXX: Waiting in async tests is generally bad. Fix this by providing different means of
```

crates/astria-sequencer-relayer/src/relayer/submission.rs
```diff
@@ -0,0 +1,458 @@
...
+
+impl SubmissionState {
+    pub(super) fn last_fetched_height(&self) -> Option<u64> {
+        match self.post {
+            PostSubmission::Fresh => None,
+            PostSubmission::Submitted {
+                sequencer_height, ..
+            } => Some(sequencer_height),
+        }
+    }
```
noot, line 134: last fetched height or last submitted height?

crates/astria-sequencer-relayer/src/relayer/submission.rs
```diff
@@ -0,0 +1,458 @@
...
+        sequencer_height: height_in_current,
+        ..
+    } = current_submission
+    else {
+        return Ok(());
+    };
+    ensure!(
+        height_in_last <= height_in_current,
+        "sequencer height in `last_submission` in the pre-submit file cannot exceed the sequencer \
+         height recorded as submitted in the post-submit file"
```
noot, line 283:
```suggestion
        "sequencer height in `last_submission` in the pre-submit file cannot exceed or be equal to the sequencer \
         height recorded as submitted in the post-submit file"
```

crates/astria-sequencer-relayer/src/relayer/submission.rs
```diff
@@ -0,0 +1,458 @@
...
+        sequencer_height: height_in_current,
+        ..
+    } = current_submission
+    else {
+        return Ok(());
+    };
+    ensure!(
+        next_height >= height_in_current,
+        "sequencer height as next submitted height in the pre-submit file cannot be less than the \
+         sequencer height recorded as submitted in the post-submit file"
```
noot, line 302: are they allowed to be equal? that seems wrong and conflicts with the `ensure_height_in_last_does_not_exceed_height_in_current` check

crates/astria-sequencer-relayer/src/relayer/write.rs
SuperFluffy: Yeah, the `select!` macro first evaluates the guard, if present. It then polls the futures to the left of the guard if and only if the guard evaluates to `true`.

So in this case the channel will only be polled if the submitter thinks it has capacity.

Have a read here: https://docs.rs/tokio/latest/tokio/macro.select.html

crates/astria-sequencer-relayer/src/relayer/write.rs
SuperFluffy: That's a pretty good question actually!

I think it's indeed only started after the future is polled for the first time (so once the select macro is entered again).

I'm going to move the task spawn outside the block to start the task immediately. This could lead to funny gotchas otherwise.

Thank you!

crates/astria-sequencer-relayer/src/relayer/submission.rs
SuperFluffy: Yes, thank you! I started getting confused with the naming here. I renamed this to `last_submitted_height` and also replaced the `u64` by `SequencerHeight` so that this already fails when parsing the file (not so important for the celestia height counterpart because `blob.Submit -> u64` and it's just for reporting).

crates/astria-sequencer-relayer/src/relayer/submission.rs
SuperFluffy: These are confusing, but I think this check is wrong for other reasons: a submission state is consistent and complete if `presubmit.sequencer_height == postsubmit.sequencer_height`. I have updated this (and the other consistency tests) to reflect that.

crates/astria-sequencer-relayer/src/relayer/submission.rs
SuperFluffy: I have tightened these tests (similar reason to the one below): `ensure_height_in_last_is_less_than_height_in_current` now enforces that the newest successful submission is larger than the previous successful submission.

crates/astria-sequencer-relayer/local.env.example
SuperFluffy: I made these requirements even stricter for the reason outlined above: we want to avoid at all costs that relayer accidentally starts re-relaying everything from height 1 onward. Operators are now expected to explicitly create a pre-submit file containing `{"state": "ignore"}` and a post-submit file with `{"state": "fresh"}`. Only then will relayer start relaying everything form 1 onward.

I have also documented this in the `astria-sequencer-relayer/local.env.example`

crates/astria-sequencer-relayer/src/relayer/mod.rs
SuperFluffy: No longer relevant: the states now contain `tendermint::block::Height`s so this is ensured at deserialization time (which is what it should have been).

joroshiba (approved):
LGTM, should we make sure on startup that we have read write permissions on the lock files? 

It's possible we can read successfully, write to presubmit and post submit fails due to file permissions? Might be a good startup check? I don't think this is necessarily blocking but potential follow on task.

SuperFluffy:
> LGTM, should we make sure on startup that we have read write permissions on the lock files?
> 
> It's possible we can read successfully, write to presubmit and post submit fails due to file permissions? Might be a good startup check? I don't think this is necessarily blocking but potential follow on task.

I originally punted this because it was annoying with the original empty-file idea. But now I don't have excuses not to do it.

I will add it to the `SubmissionState::from_paths` constructor.

PR #783, opened 2024-03-05
chore(proto)!: delete unused proto file

noot:
## Summary
delete unused proto file, it was also confusing since we have other types that are actually used with the same/similar naming

## Changes
- delete unused proto file and recompile protos

## Breaking change

This is technically a breaking change and hence marked as such, but in practice this was not used by any of our services.

## Testing
n/a


SuperFluffy (approved):
Cleanups are always nice! Thank you.

I have changed the PR title to `chore!` and added `Breaking changes` section mentioning that this technically is a breaking change.

Approved despite failing CI: buf fails because a proto file was removed, which is the point of this PR.

PR #784, opened 2024-03-05
feat(composer): Add a gRPC collector

bharath-123:
## Summary
We add a gRPC collector to the Searcher in the Composer module. The gRPC collector exposes an API which allows end-users to send bytes of txs along with their corresponding rollup ids. 

## Background
Currently, the Composer has a Searcher component which spawns up various rollup specific Collectors. These Collectors stream txs sent to the rollup node on a websocket connection. Essentially we subscribe the `getPendingTransactions` websocket stream of the rollup node. Here, we assume that the rollup node is a Geth node.

We want to add an API to which users can submit a list of sequence actions which will be bundled and sent to the sequencer. This will make testing a lot easier and allow non-geth based rollups to integrate with Astria composer easily. 

We also want to keep the existing Composer flow of streaming txs from the rollup nodes alongside with this API.

## Changes
- Added a gRPC server in the `Composer`.
- Add a new configuration `grpc_addr`. This is the address to which the gRPC collector is binded to. We bind to this address in `Composer::from_config`
- We add the protobufs for this gRPC server in the `proto/composerapis` directory. We specify the protos for the `SubmitRollupTransactionsRequest` which is the request sent to the gRPC server and the `SubmitRollupTransactions` rpc which handles validating the payload received in `SubmitSequenceActionsRequest` and its passing to the `Executor`
- Added the tonic_health healthcheck service to the gRPC server.

## Testing
Added integration tests where we submit txs to the gRPC collector and verify if it was included in the Sequencer block with a mock sequencer.

## Notes
1. There is a field called `rollups` in cfg which is basically a map from rollup to the websocket url to be used by the GethCollector. We could better call it `rollup_by_id` but thats a breaking change which I think we can reserve for future PR if required since it could break any existing integration testing systems.
2. We do not track the status of the gRPC collector in the composer_status. This is because the gRPC collector essentially just starts up a gRPC server and if there is a failure during the startup/running of the gRPC server, it is a pretty critical issue.

## Issue
https://github.com/astriaorg/astria/issues/795

closes


crates/astria-composer/src/composer.rs
```diff
@@ -59,22 +114,42 @@ impl Composer {
...
+                .serve_with_incoming(tokio_stream::wrappers::TcpListenerStream::new(
+                    grpc_collector_listener,
+                ))
+                .await
+        });
+
         tokio::select! {
             o = api_task => report_exit("api server", o),
             o = searcher_task => report_exit("searcher", o),
+            o = grpc_server_handler => println!("grpc server ended: {:?}", o),
```
noot, line 152: use `report_exit` here?

crates/astria-composer/src/composer.rs
```diff
@@ -59,22 +114,42 @@ impl Composer {
...
 
+        // run the grpc server
+        let composer_service = ComposerServiceServer::new(executor_handle);
+        let grpc_server = tonic::transport::Server::builder().add_service(composer_service);
+        let grpc_server_handler = tokio::spawn(async move {
+            grpc_server
+                .serve_with_incoming(tokio_stream::wrappers::TcpListenerStream::new(
+                    grpc_collector_listener,
+                ))
+                .await
```
noot, line 146: why use the tcp stream instead of just `grpc_server.serve(listen_addr)`?

also, ideally this would use `serve_with_shutdown` (see here https://github.com/astriaorg/astria/blob/main/crates/astria-sequencer/src/sequencer.rs#L199) but it seems composer doesn't have clean shutdown logic yet :(

crates/astria-composer/src/composer.rs
```diff
@@ -22,12 +51,22 @@ use crate::{
...
+    /// This is at the Composer level to allow its sharing to various different collectors.
+    executor_handle: ExecutorHandle,
+    /// `GrpcCollectorListener` is the tcp connection on which the gRPC collector is running
+    grpc_collector_listener: TcpListener,
+}
+
+#[derive(Clone)]
+struct ExecutorHandle {
+    send_bundles: Sender<SequenceAction>,
 }
```
noot, line 70: move this to above  the `impl ComposerService for ExecutorHandle` block

crates/astria-composer/src/composer.rs
```diff
@@ -22,12 +51,22 @@ use crate::{
...
+    /// `ExecutorHandle` to communicate SequenceActions to the Executor
+    /// This is at the Composer level to allow its sharing to various different collectors.
+    executor_handle: ExecutorHandle,
+    /// `GrpcCollectorListener` is the tcp connection on which the gRPC collector is running
+    grpc_collector_listener: TcpListener,
+}
+
+#[derive(Clone)]
+struct ExecutorHandle {
+    send_bundles: Sender<SequenceAction>,
```
noot, line 69:
```suggestion
    sequence_action_tx: Sender<SequenceAction>,
```

since this sends individual actions not bundles

crates/astria-composer/src/composer.rs
```diff
@@ -90,3 +165,45 @@ fn report_exit(task_name: &str, outcome: Result<eyre::Result<()>, JoinError>) {
...
+                    return Err(tonic::Status::deadline_exceeded(
+                        "timeout while sending txs to searcher",
+                    ));
+                }
+                Err(SendTimeoutError::Closed(_seq_action)) => {
+                    return Err(tonic::Status::unavailable("searcher is not available"));
+                }
+            }
+        }
+
```
noot, line 206: question: is the intention here to send all the actions within the rpc request as a bundle to be executed atomically? because right now it's sending individually, which should probably be documented clearer in the grpc proto definition if that is the intended usage. however it seems there should be a way to send a bunch of actions as a bundle?

proto/composerapis/astria/composer/v1alpha1/generic_collector.proto
```diff
@@ -0,0 +1,19 @@
+syntax = 'proto3';
+
+package astria.composer.v1alpha1;
+
+import "google/protobuf/empty.proto";
+
+message SequenceAction {
+  string rollup_id = 1;
```
noot, line 8: `rollup_name` (ie plaintext string) or the 32 byte `rollup_id`? should probably be `rollup_id` to match other APIs, but in that case, you can just use the `SequenceAction` defined in sequencerapis

proto/composerapis/astria/composer/v1alpha1/generic_collector.proto
```diff
@@ -0,0 +1,19 @@
...
+  bytes tx_bytes = 2;
+}
+
+message SubmitSequenceActionsRequest {
+  // Sequence Actions the user wants to send to the sequencer app.
+  repeated SequenceAction sequence_actions = 1;
+}
+
+service ComposerService {
+  rpc SubmitSequenceActions(SubmitSequenceActionsRequest) returns (google.protobuf.Empty) {}
```
noot, line 18: it would be cool if this returned the tx hashes of the txs that were submitted to sequencer containing the submitted sequence actions

crates/astria-composer/src/composer.rs
```diff
@@ -90,3 +165,45 @@ fn report_exit(task_name: &str, outcome: Result<eyre::Result<()>, JoinError>) {
         }
     }
 }
+
+#[async_trait::async_trait]
+impl ComposerService for ExecutorHandle {
```
noot, line 170: would prefer for all of `ExecutorHandle` to go in its own file/module, and also maybe name it `ComposerServer` or something like that?

crates/astria-composer/src/composer.rs
bharath-123: yes!! apologies for this!

crates/astria-composer/src/composer.rs
bharath-123: yeah :( that is a future task as of now.

crates/astria-composer/src/composer.rs
bharath-123: will do

crates/astria-composer/src/composer.rs
bharath-123: also the tcp stream is to be able to support tests which use port 0 to allow the OS to randomly assign a port. We can't use port 0 with just `serve`. TcpStream is mainly being used to enhance testability of the service.

crates/astria-composer/src/composer.rs
bharath-123: thats a good question. I believe this can be taken as a future enhancement for now. will discuss with Itamar.

crates/astria-composer/src/composer.rs
bharath-123: i will move the code to its own file/module.

proto/composerapis/astria/composer/v1alpha1/generic_collector.proto
bharath-123: we were initially thinking of such a design. But on further discussions, we chose to stuck with the design of sending the sequence actions over to the executor over a channel. 
I do believe composer will evolve a bit from here.

crates/astria-composer/src/composer.rs
bharath-123: Looks like buf complains when I call it `ComposerServer` vs `ComposerService`. Plus we seem to be using the `Service` suffix for all our gRPC servers(https://github.com/astriaorg/astria/blob/72645d41c63515766528fba5fe616480553dcace/proto/executionapis/astria/execution/v1alpha2/execution.proto#L103). Should we just stuck to that?

crates/astria-composer/src/composer.rs
bharath-123: this is done

proto/composerapis/astria/composer/v1alpha1/generic_collector.proto
bharath-123: i was actually think we don't want them to specify any native_fee_asset because composer is technically a gas station in this case. we want to pay the fee for sending the sequence actions.

crates/astria-composer/src/composer.rs
noot: sorry, I meant to rename `ExecutorHandle` not `ComposerService`, but it's fine to leave as is

proto/composerapis/astria/composer/v1alpha1/generic_collector.proto
noot: that makes sense!

itamarreif (commented):
Looks good. Summarizing my comments/thoughts:

- A few nits around imports and such.
- The way the `Composer`'s `select!` loop currently is will cause the thing to crash if the `GrpcCollector` dies for some reason. Seems reasonable since you'll be adding graceful shutdowns in a next PR.
- See my comment on the protobuf file for the gRPC service. Taking in a bundle and then unbundling it and bundling it back, potentially in a different way, seems kind of weird to me.
  - In most cases it doesn't seem like that would be a problem but I put together an example in the comments where it would happen.
  - One alternative is to just take in a single `SequenceAction` at a time. Not really sure what the consequences of that are, besides it being slightly more annoying client side.
  - Another option is to change the `serialized_rollup_transactions` channel to be for `Vec<SequenceAction>` and rename it back to `new_bundles` or something. We had this before but removed it since the `GethCollector` was only reaping one transaction at a time and so we were always sending a `vec![tx]` which felt redundant.
  - It doesn't seem to me like we currently need the functionality to submit bundles to the gRPC collector, considering our mental model until now was reaping one rollup transaction at a time from Geth via WS.
  - I agree that this would be nicer to have tho, from the perspective of someone using the `Composer` for their searcher bot's bundle submission or something like that. If you do end up doing this, I would just make sure to clarify in the PR description.

crates/astria-composer/src/searcher/executor/mod.rs
```diff
@@ -98,7 +98,7 @@ pub(super) struct Executor {
     // The sequencer address associated with the private key
     address: Address,
     // Milliseconds for bundle timer to make sure bundles are submitted at least once per block.
-    block_time: tokio::time::Duration,
+    block_time: Duration,
```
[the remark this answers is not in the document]
bharath-123: hmmm that was part of my editor. i can revert this.

crates/astria-composer/src/searcher/executor/tests.rs
```diff
@@ -243,7 +244,7 @@ async fn full_bundle() {
 
     // verify only one signed transaction was received by the mock sequencer
     // i.e. only the full bundle was sent and not the second one due to the block timer
-    let expected_seq_actions = vec![seq0];
+    let expected_seq_actions = [seq0];
```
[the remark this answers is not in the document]
bharath-123: yeah, let me revert these lint changes anyways.

crates/astria-composer/src/composer.rs
```diff
@@ -37,11 +50,25 @@ impl Composer {
...
+        let searcher = Searcher::from_config(
+            cfg,
+            serialized_rollup_transactions_tx.clone(),
+            serialized_rollup_transactions_rx,
+        )
+        .wrap_err("failed to initialize searcher")?;
 
         let searcher_status = searcher.subscribe_to_state();
 
+        let grpc_collector_listener = TcpListener::bind(cfg.grpc_collector_addr).await?;
```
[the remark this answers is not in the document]
bharath-123: I would ideally like to save the TcpListener address because that will give the final correct address considering the port to be bound.

proto/composerapis/astria/composer/v1alpha1/grpc_collector.proto
```diff
@@ -0,0 +1,19 @@
+syntax = 'proto3';
+
+package astria.composer.v1alpha1;
+
+import "google/protobuf/empty.proto";
+
+message SequenceAction {
```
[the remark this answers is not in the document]
bharath-123: we could but it includes the `default_fee_asset_id` which we do not the user to specify

crates/astria-composer/src/composer.rs
```diff
@@ -59,22 +88,53 @@ impl Composer {
...
+                ))
+                .await
+        });
+
         tokio::select! {
-            o = api_task => report_exit("api server", o),
-            o = searcher_task => report_exit("searcher", o),
+                o = api_task => report_exit("api server", o),
+                o = searcher_task => report_exit("searcher", o),
+                o = grpc_server_handler => {
```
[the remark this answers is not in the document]
bharath-123: i ll add that in the graceful shutdown PR

proto/composerapis/astria/composer/v1alpha1/grpc_collector.proto
```diff
@@ -0,0 +1,19 @@
...
+  string rollup_id = 1;
+  bytes tx_bytes = 2;
+}
+
+message SubmitSequenceActionsRequest {
+  // Sequence Actions the user wants to send to the sequencer app.
+  repeated SequenceAction sequence_actions = 1;
```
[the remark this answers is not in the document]
bharath-123: I see what you are trying to say and it makes sense. I think we should submit a Vec of bundles to solve for this. And I also propose that we do this in a follow up once we get to merge this gRPC collector stuff.

bharath-123:
I don't think we need to add gRPC collector status to `ComposerStatus`. This is because, gRPC collectors just spins up a gRPC server and if it fails then we exit the composer itself(since gRPC server exiting is a fatal event)

crates/astria-composer/src/composer.rs
```diff
@@ -15,42 +28,113 @@ use crate::{
...
+    /// Constructs a new Composer service from config.
     ///
     /// # Errors
     ///
-    /// An error is returned if the searcher fails to be initialized.
-    /// See `[Searcher::from_config]` for its error scenarios.
-    pub fn from_config(cfg: &Config) -> eyre::Result<Self> {
-        let searcher = Searcher::from_config(cfg).wrap_err("failed to initialize searcher")?;
+    /// An error is returned if the composer fails to be initialized.
+    /// See `[Composer::from_config]` for its error scenarios.
```
noot, line 78: remove/update this line? it's referencing itself

crates/astria-composer/src/composer.rs
```diff
@@ -15,42 +28,113 @@ use crate::{
...
+    /// The sequencer transactions are received from various collectors.
+    executor: Executor,
+    /// `GethCollectors` is the collection of geth collectors and their rollup names.
+    geth_collectors: HashMap<String, GethCollector>,
+    /// `GethCollectorStatuses` The collection of the geth collector statuses.
+    geth_collector_statuses: HashMap<String, watch::Receiver<geth_collector::Status>>,
+    /// `GethCollectorTasks` is the set of tasks tracking if the geth collectors are still running.
+    geth_collector_tasks: JoinMap<String, eyre::Result<()>>,
+    /// `Rollups` The map of chain ID to the URLs to which geth collectors should connect.
+    rollups: HashMap<String, String>,
```
noot, line 67:
```suggestion
    rollup_id_to_url: HashMap<String, String>,
```

more descriptive is good here

crates/astria-composer/src/composer.rs
```diff
@@ -59,26 +143,166 @@ impl Composer {
...
+            o = &mut executor_task => {
+                    report_exit("executor unexpectedly ended", o);
+                    return Ok(());
+            },
+            o = &mut grpc_server_handle => {
+                    report_exit("grpc server unexpectedly ended", o);
+                    return Ok(());
+            },
+            Some((rollup, collector_exit)) = geth_collector_tasks.join_next() => {
+                reconnect_exited_collector(
```
noot, line 214: if the geth node goes down for instance, this will continuously loop right? we should probably have a retry limit/backoff here

crates/astria-composer/src/grpc_collector/mod.rs
```diff
@@ -0,0 +1,127 @@
...
+
+        match grpc_server
+            .serve_with_incoming(tokio_stream::wrappers::TcpListenerStream::new(
+                self.grpc_collector_listener,
+            ))
+            .await
+        {
+            Ok(()) => Ok(()),
+            Err(err) => Err(eyre!(err)),
+        }
```
noot, line 83:
```suggestion
    grpc_server
            .serve_with_incoming(tokio_stream::wrappers::TcpListenerStream::new(
                self.grpc_collector_listener,
            ))
            .await
            .wrap_err("failed to run grpc server")
```

crates/astria-composer/src/grpc_collector/mod.rs
```diff
@@ -0,0 +1,127 @@
...
+    ) -> Result<Response<()>, tonic::Status> {
+        let submit_rollup_txs_request = request.into_inner();
+        if submit_rollup_txs_request.rollup_txs.is_empty() {
+            return Err(tonic::Status::invalid_argument(
+                "No sequence actions provided",
+            ));
+        }
+
+        // package the sequence actions into a SequenceAction and send it to the searcher
+        for sequence_action in submit_rollup_txs_request.rollup_txs {
```
noot, line 101: i still find it somewhat unexpected that the txs passed in via grpc are unbundled and sent separately - can you add a comment on this behaviour on the protobuf? do you plan to change this later?

noot (commented):
nice, mostly smaller comments.

also, why was the functionality for running multiple collectors moved from `Searcher` to `Composer`? I thought the point of `Searcher` was to search over some set of rollups, and if a composer wanted to have different searchers for different sets of rollups, then that could be done easily. this sort of removes that functionality, which we weren't using anyways, but curious as to the rationale

crates/astria-composer/src/composer.rs
bharath-123: makes sense. we can add it in a seperate PR

crates/astria-composer/src/grpc_collector/mod.rs
bharath-123: Yes, we will make these channel take in a `Vec<SequenceAction>`, rather than just sequence action in a future PR!

bharath-123:
> nice, mostly smaller comments.
> 
> also, why was the functionality for running multiple collectors moved from `Searcher` to `Composer`? I thought the point of `Searcher` was to search over some set of rollups, and if a composer wanted to have different searchers for different sets of rollups, then that could be done easily. this sort of removes that functionality, which we weren't using anyways, but curious as to the rationale

It was mainly for testability and maintainability. It became harder to write tests for collectors like the gRPC collector if the collectors were 2 levels below the composer. We weren't able to get the port of the gRPC collector. So, to make things more testable, we moved it to the top level.

crates/astria-composer/src/collectors/grpc.rs
```diff
@@ -0,0 +1,139 @@
...
+    Request,
+    Response,
+};
+
+use crate::{
+    executor,
+    executor::Handle,
+};
+
+/// `GrpcCollector` listens for incoming gRPC requests and sends the Rollup transactions to the
```
SuperFluffy, line 45:
```suggestions
/// Listens for incoming gRPC requests and sends the Rollup transactions to the
```

crates/astria-composer/src/collectors/grpc.rs
```diff
@@ -0,0 +1,139 @@
...
+    /// Returns an error if the listener is not bound
+    pub(crate) fn local_addr(&self) -> io::Result<SocketAddr> {
+        self.listener.local_addr()
+    }
+
+    pub(crate) async fn run_until_stopped(self) -> eyre::Result<()> {
+        let (mut health_reporter, health_service) = tonic_health::server::health_reporter();
+
+        let composer_service = GrpcCollectorServiceServer::new(self.executor_handle);
+        let grpc_server = tonic::transport::Server::builder()
```
SuperFluffy, line 82: With the health service added (and this changing the nature of composer so much), we should have a root-level `grpc` module which is the actual gRPC server. `collectors::grpc` can then just contain the implementation of `GrpcCollectorServiceServer`.

So basically I think this logic should be moved there.

crates/astria-composer/src/collectors/grpc.rs
```diff
@@ -0,0 +1,139 @@
...
+            .serve_with_incoming(tokio_stream::wrappers::TcpListenerStream::new(
+                self.listener,
+            ))
+            .await
+            .wrap_err("failed to run grpc server")
+    }
+}
+
+#[async_trait::async_trait]
+impl GrpcCollectorService for executor::Handle {
```
SuperFluffy, line 101: Please don't implement this trait on `executor::Handle`. It's very surprising. Can you instead do something like this:

```rust
struct GrpcCollectorServiceImpl {
    executor: executor::Handle,
}
```

crates/astria-composer/src/collectors/grpc.rs
```diff
@@ -0,0 +1,139 @@
...
+    async fn submit_rollup_txs(
+        &self,
+        request: Request<SubmitRollupTxsRequest>,
+    ) -> Result<Response<()>, tonic::Status> {
+        let submit_rollup_txs_request = request.into_inner();
+        if submit_rollup_txs_request.rollup_txs.is_empty() {
+            return Err(tonic::Status::invalid_argument(
+                "No sequence actions provided",
```
SuperFluffy, line 109: small case messages (I actually don't know the gRPC convention here, but let's be consistent with our/rust error message conventions).

proto/composerapis/astria/composer/v1alpha1/grpc_collector.proto
```diff
@@ -0,0 +1,18 @@
+syntax = 'proto3';
+
+package astria.composer.v1alpha1;
+
+import "google/protobuf/empty.proto";
+
+message RollupTx {
```
SuperFluffy, line 7: We mostly don't have abbreviations in the codebase.
```suggestion
message RollupTransaction {
```

proto/composerapis/astria/composer/v1alpha1/grpc_collector.proto
```diff
@@ -0,0 +1,18 @@
...
+  string rollup_id = 1;
```
SuperFluffy, line 8: Should this be a generic `string` or should we also require this be `bytes` like all other protobuf specs?

The confusing thing is that this is called `rollup_id` but is actually an arbitrary rollup name that has to be hashed before getting the final `RollupId` (in this PR using`RollupId::from_unhashed_bytes`). On the other hand, the rest of the codebase refers to `RollupId` as a sequence of 32 bytes.

proto/composerapis/astria/composer/v1alpha1/grpc_collector.proto
```diff
@@ -0,0 +1,18 @@
...
+  bytes tx_bytes = 2;
+}
+
+message SubmitRollupTxsRequest {
+  repeated RollupTx rollup_txs = 1;
+}
+
+service GrpcCollectorService {
+  rpc SubmitRollupTxs(SubmitRollupTxsRequest) returns (google.protobuf.Empty) {}
```
SuperFluffy, line 17: `google.protobuf.Empty` or `SubmitRollupTransactionsResponse` to add more fields in the future?

justfile
```diff
@@ -68,3 +68,6 @@ _lint-proto:
   buf format -d --exit-code
   buf breaking proto/executionapis --against 'buf.build/astria/execution-apis'
   buf breaking proto/sequencerapis --against 'buf.build/astria/astria'
+  buf breaking proto/composerapis --against 'buf.build/astria/astria'
+
+
```
SuperFluffy, line 73: can you remove the new empty lines here?

crates/astria-composer/src/lib.rs
```diff
@@ -43,6 +43,7 @@ mod collectors;
 mod composer;
 pub mod config;
 mod executor;
+
```
SuperFluffy, line 46: no newline

crates/astria-composer/src/lib.rs
```diff
@@ -30,7 +30,7 @@
 //!     .expect("failed to setup telemetry");
 //! info!(config = cfg_ser, "initializing composer",);
 //!
-//! let _composer = Composer::from_config(&cfg)
+//! let _composer = Composer::from_config(&cfg).await
```
SuperFluffy, line 33: Consistent doc formatting:
```suggestion
//! let _composer = Composer::from_config(&cfg)
//!     .await

File: crates/astria-composer/src/collectors/grpc.rs [hunk 30]
```diff
@@ -0,0 +1,139 @@
...
+            ));
+        }
+
+        // package the rollup txs into a SequenceAction and send it to the searcher
+        for rollup_txs in submit_rollup_txs_request.rollup_txs {
+            let sequence_action = SequenceAction {
+                rollup_id: RollupId::from_unhashed_bytes(rollup_txs.rollup_id),
```
SuperFluffy, line 116: Commented in the protobuf spec: I think I'd like to be stricter how data is submitted to the server. The `SubmitRollupTransactionsRequest` should IMO have `bytes rollup_id`, and the server expect that to be `[u8; 32]`

SuperFluffy (commented):
Quick set of comments to get started

crates/astria-composer/src/collectors/grpc.rs
```diff
@@ -0,0 +1,139 @@
+// ! `GrpcCollector` allows users to directly send Rollup transactions to the Composer
```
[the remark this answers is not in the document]
bharath-123: thanks! will take better care of this next time!

proto/composerapis/astria/composer/v1alpha1/grpc_collector.proto
bharath-123: I think we shouldn't use `from_unhashed_bytes` then. Let me see how I can make it just a sequence of 32 bytes and avoid using `from_unhashed_bytes`

bharath-123: used bytes instead of string. that should be better.

proto/composerapis/astria/composer/v1alpha1/grpc_collector.proto
bharath-123: @joroshiba could you let us know your thoughts on this?

Cargo.toml
```diff
@@ -95,3 +95,4 @@ tracing = "0.1"
 tryhard = "0.5.1"
 which = "4.4.0"
 wiremock = "0.5"
+tokio-stream = { version = "0.1.14", features = ["net"] }
```
SuperFluffy, line 98: Don't activate features in the workspace crate - activate them in the crates that make use of the workspace dependency.

SuperFluffy (commented):
Thank you very much for this PR. I think this is almost ready to merge minus a few nits + request to provide documentation on the new protobuf spec.

Before I approve I would like to get @joroshiba's input on the design of the RPC return message.

proto/composerapis/astria/composer/v1alpha1/grpc_collector.proto
```diff
@@ -0,0 +1,18 @@
...
+  bytes rollup_id = 1;
+  bytes tx_bytes = 2;
+}
+
+message SubmitRollupTransactionsRequest {
+  repeated RollupTransaction rollup_txs = 1;
+}
+
+service GrpcCollectorService {
+  rpc SubmitRollupTransactions(SubmitRollupTransactionsRequest) returns (google.protobuf.Empty) {}
```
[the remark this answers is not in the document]
joroshiba: I believe it would be a breaking change, at least based on how we have defined as strictly as possible (based on file). We should have an empty return object instead. 

Similar to how we use empty request objects for `GetGenesisInfoRequest` in sequencer API.

justfile
```diff
@@ -68,3 +68,4 @@ _lint-proto:
   buf format -d --exit-code
   buf breaking proto/executionapis --against 'buf.build/astria/execution-apis'
   buf breaking proto/sequencerapis --against 'buf.build/astria/astria'
+  buf breaking proto/composerapis --against 'buf.build/astria/astria'
```
joroshiba, line 71: The new one will have to be against `astria/composer-apis` and you will need to update `.github/workflows/release.yml` to push these to buf.

itamarreif (commented):
Looks good. Agree with @SuperFluffy about needing comments on the proto files and the return codes/messages in `collectors::grpc.rs`.

itamarreif (approved):

crates/astria-composer/src/collectors/grpc.rs
```diff
@@ -0,0 +1,83 @@
...
+
+            match self
+                .executor
+                .send_timeout(sequence_action, Duration::from_millis(500))
+                .await
+            {
+                Ok(()) => {}
+                Err(SendTimeoutError::Timeout(_seq_action)) => {
+                    return Err(tonic::Status::deadline_exceeded(
+                        "timeout while sending txs to searcher",
```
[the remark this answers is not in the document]
bharath-123: I think we can just mention composer. Because we anyways describe this module as composer.

crates/astria-composer/src/collectors/grpc.rs
```diff
@@ -0,0 +1,83 @@
...
+                    ));
+                }
+                Err(SendTimeoutError::Closed(_seq_action)) => {
+                    return Err(tonic::Status::unavailable("searcher is not available"));
```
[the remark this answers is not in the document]
bharath-123: I think closed makes more sense? Because this can only happen when the executor is shutting down and it closes the receiving end of the mpsc channel?

crates/astria-composer/src/composer.rs
```diff
@@ -52,8 +55,11 @@ pub struct Composer {
     geth_collector_statuses: HashMap<String, watch::Receiver<collectors::geth::Status>>,
     /// `GethCollectorTasks` is the set of tasks tracking if the geth collectors are still running.
     geth_collector_tasks: JoinMap<String, eyre::Result<()>>,
-    /// `Rollups` The map of chain ID to the URLs to which geth collectors should connect.
-    rollups: HashMap<String, String>,
+    /// `RollupIdToUrl` The map of chain ID to the URLs to which geth collectors should connect.
+    rollup_id_to_url: HashMap<String, String>,
```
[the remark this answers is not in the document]
bharath-123: we can do it later.

crates/astria-composer/Cargo.toml
```diff
@@ -50,6 +50,9 @@ tokio = { workspace = true, features = [
 tokio-util = { workspace = true, features = ["rt"] }
 tracing = { workspace = true, features = ["attributes"] }
 tryhard = { workspace = true }
+tonic = { workspace = true }
+tokio-stream = { workspace = true, features = ["net"] }
+tonic-health = "0.10.2"
```
bharath-123, line 55: Might be worth moving it to the workspace as we standardize health service across grpc servers.

proto/composerapis/astria/composer/v1alpha1/grpc_collector.proto
```diff
@@ -0,0 +1,30 @@
...
+  bytes rollup_id = 1;
+  // the raw data bytes of the rollup transaction
+  bytes data = 2;
+}
+
+// SubmitRollupTransactionsRequest contains a batch of rollup transactions to be submitted to the Shared Sequencer
+// Network via the Composer
+message SubmitRollupTransactionsRequest {
+  // an array of rollup transactions to be submitted
+  repeated RollupTransaction rollup_transactions = 1;
```
itamarreif, line 18: Should not be a `repeated` - keeping a single tx per request will make tracing easier. We end up breaking these apart into bundles inside `composer` anyways, so keeping it at one tx per request would allow rollup nodes to attach an id to the request that would identify the tx.

bharath-123: makes sense. I think we would have to understand how to define the id. But I think keeping it at 1 tx should be pretty good.

joroshiba (approved):
I have not had time to review the rust code, approving CI/CD changes and API design.

charts/evm-rollup/templates/configmap.yaml
```diff
@@ -72,6 +72,7 @@ data:
   {{- end }}
   {{- if not .Values.global.dev }}
   {{- else }}
+  ASTRIA_COMPOSER_GRPC_ADDR: "0.0.0.0:0"
```
SuperFluffy, line 75: Yeah, I guess this is fine albeit odd - we don't have anything talking to this gRPC service in the testnet, so it can just bind whatever it wants.

crates/astria-composer/Cargo.toml
```diff
@@ -13,7 +13,7 @@ name = "astria-composer"
 
 [dependencies]
 astria-build-info = { path = "../astria-build-info", features = ["runtime"] }
-astria-core = { path = "../astria-core", features = ["serde"] }
+astria-core = { path = "../astria-core", features = ["server", "serde"] }
```
SuperFluffy, line 16: very very nit: can you order features alphabetically?

crates/astria-composer/src/composer.rs
```diff
@@ -95,7 +101,17 @@ impl Composer {
         )
         .wrap_err("executor construction from config failed")?;
 
+        let grpc_server = GrpcServer::new(cfg.grpc_addr, executor_handle.clone())
+            .await
+            .wrap_err("grpc collector construction from config failed")?;
```
SuperFluffy, line 106:
```suggestion
            .wrap_err("failed to create grpc server")?;
```

crates/astria-composer/src/composer.rs
```diff
@@ -140,6 +157,13 @@ impl Composer {
         self.api_server.local_addr()
     }
 
+    /// Returns the socker address the grpc collector is served over
+    /// # Errors
+    /// Returns an error if the listener is not bound
+    pub fn grpc_collector_local_addr(&self) -> io::Result<SocketAddr> {
```
SuperFluffy, line 163:
```suggestion
    pub fn grpc_local_addr(&self) -> io::Result<SocketAddr> {
```
This is after all not the address of the collector, but the address of the entire server that is also serving the collector service.

crates/astria-composer/tests/blackbox/composer.rs
```diff
@@ -54,6 +60,46 @@ async fn tx_from_one_rollup_is_received_by_sequencer() {
     .expect("mocked sequencer should have received a broadcast message from composer");
 }
 
+#[tokio::test]
+async fn tx_from_one_rollup_is_received_by_sequencer_from_grpc_collector() {
+    let test_composer = spawn_composer(&[]).await;
+    tokio::time::timeout(
+        Duration::from_millis(100),
+        test_composer.setup_guard.wait_until_satisfied(),
```
SuperFluffy, line 68: not for this PR: since this setup guard is likely the same for all tests, you can just move this into `spawn_composer` so that readers of the test don't have to worry about this (since it's kinda orthogonal to the aim of this PR).

crates/astria-composer/tests/blackbox/composer.rs
```diff
@@ -54,6 +60,46 @@ async fn tx_from_one_rollup_is_received_by_sequencer() {
...
+    ))
+    .await
+    .unwrap();
+    composer_client
+        .submit_rollup_transaction(SubmitRollupTransactionRequest {
+            rollup_id: rollup_id.as_ref().to_vec(),
+            data: tx.rlp().to_vec(),
+        })
+        .await
+        .expect("error submitting sequence actions to generic collector");
```
SuperFluffy, line 92: The `expect` message should tell you that something failed, but instead what you expected to happen:
```suggestion
        .expect("can submit a transaction to composer's GrpcCollector service");
```
have a read here: https://doc.rust-lang.org/std/error/index.html#common-message-styles

crates/astria-composer/tests/blackbox/composer.rs
```diff
@@ -54,6 +60,46 @@ async fn tx_from_one_rollup_is_received_by_sequencer() {
...
+
+    // wait for 1 sequencer block time to make sure the bundle is preempted
+    tokio::time::timeout(
+        Duration::from_millis(test_composer.cfg.block_time_ms),
+        mock_guard.wait_until_satisfied(),
+    )
+    .await
+    .expect("mocked sequencer should have received a broadcast message from composer");
```
SuperFluffy, line 100: Yeah, this expect message makes sense

crates/astria-composer/tests/blackbox/composer.rs
```diff
@@ -161,7 +207,83 @@ async fn invalid_nonce_failure_causes_tx_resubmission_under_different_nonce() {
...
+    // Spawn a composer with a mock sequencer and a mock rollup node
+    // Initial nonce is 0
+    let rollup_id = RollupId::from_unhashed_bytes("test1");
+    let test_composer = spawn_composer(&[rollup_id.to_string().as_str()]).await;
+    tokio::time::timeout(
+        Duration::from_millis(100),
+        test_composer.setup_guard.wait_until_satisfied(),
+    )
+    .await
+    .expect("setup guard failed");
```
SuperFluffy, line 222: Maybe you copied it from another test - but this `expect` message (like above) kinda states the obvious.

crates/astria-composer/tests/blackbox/composer.rs
```diff
@@ -161,7 +207,83 @@ async fn invalid_nonce_failure_causes_tx_resubmission_under_different_nonce() {
 }
 
 #[tokio::test]
-async fn single_rollup_tx_payload_integrity() {
+async fn invalid_nonce_failure_causes_tx_resubmission_under_different_nonce_grpc_collector() {
```
SuperFluffy, line 210:
```suggestion
async fn invalid_nonce_causes_resubmission_under_different_nonce() {
```

crates/astria-composer/tests/blackbox/composer.rs
```diff
@@ -161,7 +207,83 @@ async fn invalid_nonce_failure_causes_tx_resubmission_under_different_nonce() {
...
+    )
+    .await
+    .expect("nonce refetch guard failed");
+
+    tokio::time::timeout(
+        Duration::from_millis(100),
+        valid_nonce_guard.wait_until_satisfied(),
+    )
+    .await
+    .expect("valid nonce guard failed");
```
SuperFluffy, line 282: all these `expect` messages don't state why you expected these tests to work

crates/astria-composer/tests/blackbox/composer.rs
SuperFluffy, line 1: Are the tests in here basically copies of the non-grpc tests that verify if the payload ended up in the sequencer?

If yes, instead of adding a `_grpc_collector` suffix, can you just move them to their own test module?

I am not very happy with all the duplicated code in the tests because we will have to maintain that too, but let's have some separation at least.

bharath-123: sure will move the tests to their own module

crates/astria-composer/tests/blackbox/composer.rs
bharath-123: i ll update all the expect messages

crates/astria-composer/tests/blackbox/composer.rs
bharath-123: I ll refactor the tests in a future PR

crates/astria-composer/tests/blackbox/composer.rs
bharath-123: updating them to make more sense

itamarreif (approved):
looks good! seems like everything we discussed was addressed.

PR #633, opened 2023-12-04
chore(conductor, composer)!: update chain_id strings to rollup_name

joroshiba:
## Summary
changed references to `chain_id` that were strings to `rollup_name` or `rollup_id` from the sequencer viewpoint.

## Background
Chain ID is hugely overloaded, we previously updated our protos, but we haven't updated some configuration to remove the overloading.

## Changes
- updated conductor config
- updated composer + comments in composer

## Breaking Changes
- changes to conductor configuration

PR #791, opened 2024-03-06
chore(composer): disambiguate chain-id

joroshiba:
## Summary
Disambiguate uses of chain-id in core services, making sure it's clear when we mean `rollup-id`, `rollup-name`, or `chain-id`

## Background
We previously used to have a lot of chain id overlap. Now chain id (outside of the cli crate), should refer to only the cometbft chain id of either the sequencer or another IBC compatible chain id. Instances referring to rollups are now referenced wholly by `rollup-name` when it is a string or `rollup-id` when it is a 32 bit byte array.

There is a previous stale PR (https://github.com/astriaorg/astria/pull/633) that accomplished this, but many change since then opted to go from scratch.

## Changes
- update references mostly in composer to `chain-id`


crates/astria-composer/src/searcher/executor/tests.rs
```diff
@@ -144,7 +144,7 @@ fn signed_tx_from_request(request: &Request) -> SignedTransaction {
 }
 
 /// Deserizalizes the bytes contained in a `tx_sync::Request` to a signed sequencer transaction
-/// and verifies that the contained sequence action is in the given `expected_chain_ids` and
+/// and verifies that the contained sequence action is in the given `expected_rolup_ids` and
```
itamarreif, line 147: `expected_rollup_ids`

crates/astria-sequencer-relayer/tests/blackbox/helper.rs
```diff
@@ -329,7 +329,7 @@ fn create_block_response(
         block,
         Hash,
     };
-    let rollup_id = RollupId::from_unhashed_bytes(b"test_chain_id_1");
+    let rollup_id = RollupId::from_unhashed_bytes(b"test_chain_name_1");
```
itamarreif, line 332: `b"test_rollup_id_1"`?

itamarreif (approved):

noot (approved):

PR #792, opened 2024-03-07
chore(collector): Rename Collector to GethCollector

bharath-123:
## Summary

This PR renames Collector to GethCollector. It also renames the `Status` structs to make it easier to distinguish.

## Background

As we are working towards adding different types of collectors to Composer, it is better to re-name the existing Collector implementation to `GethCollector` since it streams txs from the rollup node's geth mempool.

It also includes refactoring the `Status` structs to `CollectorStatus`, `SearcherStatus` and `ExecutorStatus` to make it more easier to distinguish b/w these structs.

## Changes
- Rename `Collector` -> `GethCollector`
- Renames collectors field in `Searcher` struct to `geth_collectors` 
- Renames `searcher::Status` -> `SearcherStatus`
- Renames `collector::Status` -> `CollectorStatus`
- Renames `executor::Status` -> `ExecutorStatus`

Note: I chose not to name other fields of `Searcher` struct like `collector_statuses` and `collector_tasks` since they can be generic on different types of collectors and not just `GethCollectors`

## Testing
Compiled the codebase and ran tests successfully.

closes


itamarreif:
I would prefer if we stuck to `searcher:Status` over `searcher:SearcherStatus`

noot (commented):
is the current collector geth-specific or eth-specific? it should work with any ethereum execution node as they all implement the same json-rpc api, right?

bharath-123:
> is the current collector geth-specific or eth-specific? it should work with any ethereum execution node as they all implement the same json-rpc api, right?

Thats a good point. It makes sense to me. @itamarreif any thoughts from your end on this?

crates/astria-composer/src/searcher/mod.rs
```diff
@@ -25,25 +25,24 @@ use tracing::{
...
 /// sequencer, all of them are going to be executed in the same Astria block.
 pub(super) struct Searcher {
     // Channel to report the internal status of the searcher to other parts of the system.
     status: watch::Sender<Status>,
     // The collection of collectors and their rollup names.
-    collectors: HashMap<String, Collector>,
+    geth_collectors: HashMap<String, GethCollector>,
     // The collection of the collector statuses.
-    collector_statuses: HashMap<String, watch::Receiver<collector::Status>>,
+    collector_statuses: HashMap<String, watch::Receiver<geth_collector::Status>>,
```
itamarreif, line 45: `geth_collector_statuses` maybe?

crates/astria-composer/src/searcher/mod.rs
```diff
@@ -121,7 +120,7 @@ impl Searcher {
 
         Ok(Searcher {
             status,
-            collectors,
+            geth_collectors,
             collector_statuses,
             collector_tasks: JoinMap::new(),
```
itamarreif, line 125: `geth_collector_tasks` maybe?

crates/astria-composer/src/searcher/mod.rs
```diff
@@ -266,7 +265,7 @@ impl Searcher {
 }
 
 fn reconnect_exited_collector(
```
itamarreif, line 267: `reconnect_exited_geth_collector` maybe?

itamarreif (approved):
just a couple spots in the `Searcher`'s fields that may be worth changing to disambiguate between `GethCollector`s and `GrpcCollector`s in the future, but I'll leave it up to @bharath-123 to decide if he wants to change those as well.

itamarreif:
> is the current collector geth-specific or eth-specific? it should work with any ethereum execution node as they all implement the same json-rpc api, right?

This is my bad - for some reason I thought the websocket endpoints of JSON-RPC were Geth specific but from looking into it, it seems to be a general Eth JSON-RPC thing:
https://github.com/NethermindEth/docs/blob/master/ethereum-client/json-rpc/subscribe.md

bharath-123:
> > is the current collector geth-specific or eth-specific? it should work with any ethereum execution node as they all implement the same json-rpc api, right?
> 
> This is my bad - for some reason I thought the websocket endpoints of JSON-RPC were Geth specific but from looking into it, it seems to be a general Eth JSON-RPC thing: https://github.com/NethermindEth/docs/blob/master/ethereum-client/json-rpc/subscribe.md
> 
> So maybe let's go with `EthCollector` then?

I ll refactor this in a PR after this gRPC collector + refactor goes in. It ll just be easier given the way i have worked on it.

crates/astria-composer/src/searcher/mod.rs
bharath-123: I will rename it

crates/astria-composer/src/searcher/mod.rs
bharath-123: yeah that would be better. I initially thought we would have multiple collectors.

crates/astria-composer/src/searcher/mod.rs
bharath-123: sure that is doable.

crates/astria-composer/src/searcher/collector.rs
SuperFluffy, line 1: Consider renaming this module `collector -> geth_collector`. The main point of this module is provide a single collector (`GethCollector` after this PR) at least for now.

SuperFluffy:
@noot 
> is the current collector geth-specific or eth-specific? it should work with any ethereum execution node as they all implement the same json-rpc api, right?

I believe that's not the case, unfortunately, because we use geth-specific extensions of the JSON-RPC - namely subscribing to new events. Last time I checked this was not a feature generally supported outside of it.

SuperFluffy (commented):
It makes sense to disambiguate the purpose of the collector by its name.

Can you please add a module documentation for the purpose of the `collector.rs` (or `geth_collector.rs` module? Specifically, can you go into how it's geth-specific?

Have a look at https://geth.ethereum.org/docs/interacting-with-geth/rpc/pubsub and the "geth mock" that we use in the tests. (When I wrote the mock I worked myself backward from ethers is sending to the node to subscribe to new transactions)

crates/astria-composer/src/searcher/collector.rs
bharath-123: cool will do

crates/astria-composer/src/searcher/geth_collector.rs
```diff
@@ -1,3 +1,14 @@
+//! `GethCollector` is responsible for fetching pending transactions submitted to rollup
```
SuperFluffy, line 1: Make the first sentence in the module shorter. It's used as a summary when listing the module (we are not generating docs because this module is private, but it's still good practice).

```suggestion
//! Fetch transactions from a Geth rollup node.
```

crates/astria-composer/src/searcher/geth_collector.rs
```diff
@@ -1,3 +1,14 @@
...
+//! nodes and then passing them downstream for the searcher to process.
+//!
+//! It is responsible for fetching pending transactions submitted to the rollup Geth nodes and then
```
SuperFluffy, line 4:
```suggestion
//! [`GethCollector`] subscribes to pending transactions from a [go-ethereum](https://geth.ethereum.org) rollup node, and forwards them for more processing.
```
The rest of the information is orthogonal to this collector (the collector does not know about the searcher at this point, or if it can have multiple such collectors).

crates/astria-composer/src/searcher/geth_collector.rs
```diff
@@ -1,3 +1,14 @@
...
+//! passing them downstream for the executor to process. Thus, a searcher can have multiple
+//! collectors running at the same time funneling data from multiple rollup nodes.
+//!
+//! `GethCollector` uses the <https://geth.ethereum.org/docs/interacting-with-geth/rpc/pubsub#newpendingtransactions>
```
SuperFluffy, line 8:
```suggestion
//! ## Note
//! This collector is likely specific to go-ethereum and only checked to work wit. It makes use of the [`eth_subscribe`](https://geth.ethereum.org/docs/interacting-with-geth/rpc/pubsub#newpendingtransactions) JSON-RPC with arguments shown below. It appears as if go-ethereum is the only ethereum node that documents this.
//! ```json
//! { "id": 1, "jsonrpc": "2.0", "method": "eth_subscribe", "params": ["newPendingTransactions"] }
//! ```
```

SuperFluffy (approved):
Looks good to me.

Please change/add the suggested bits before merge (really want to become more explicit documenting APIs, behvavior, etc).

crates/astria-composer/src/searcher/geth_collector.rs
bharath-123: thanks! this gave me a better sense on writing these docs

Issue #795, opened 2024-03-08
Composer Enhancements


bharath-123:
This issue tracks the progess of Enhancements of Composer to be mainnet ready.

```
### Tasks
- [x] Re-name Collectors to GethCollector to distinguish b/w the various collector. (Done in https://github.com/astriaorg/astria/pull/792)
- [x] Move Executor and Collector tasks to the Composer (Done in https://github.com/astriaorg/astria/pull/796)
- [x] Add the gRPC collector (done in https://github.com/astriaorg/astria/pull/784)
- [ ] Add signal handling for graceful shutdown of the gRPC Collector, GethCollector and Executor (Being worked on in https://github.com/astriaorg/astria/pull/854)
- [ ] Remove the API Server in favour of using the gRPC server for healthchecks.
- [ ] Add logging and metrics
- [ ] Add tracing logic for rollup transaction to sequencer transaction tracking.
```

PR #796, opened 2024-03-08
refactor(composer): flatten module structure

bharath-123:
## Summary
This is a wholesale refactor of Composer where we move the Executor and GethCollector tasks to the Composer top level.

## Background
Previously, Collectors and Executors were part of a higher level Searcher task i.e they were leaf'ed of the Composer task. This makes the testing of Collectors and Executors much harder as they were very deep in the task tree. 
In the future as we add the gRPC collectors and other types of collectors, this refactor aims to improve the readability and testability of Composer.

## Changes
- Spawn the Executor and Collector tasks off Composer. 
- Move Collectors, Executor related data fields to the Composer
- Remove the Searcher struct.
- Restructure the Composer module by removing the Searcher module and flattening the GethCollector and Executor module.
- Add Composer specific documentation
- Rename SearcherStatus specific structs to ComposerStatus
- Added error handling in main.rs which is orthogonal to the PR but clippy started complaining.
 
## Testing
Since this refactor is adding no functional code but is just a refactor. We can validate it by making sure the codebase compiles and all the tests successfully runs.

## Issues
#795


crates/astria-composer/src/composer.rs
```diff
@@ -28,6 +32,13 @@ pub struct Composer {
     /// pending transactions from them and wraps them as sequencer transactions
     /// for submission.
     searcher: Searcher,
+    /// The handle to communicate SequenceActions to the Executor
+    /// This is at the Composer level to allow its sharing to various different collectors.
+    executor_handle: ExecutorHandle,
+}
+
+struct ExecutorHandle {
+    send_bundles: Sender<SequenceAction>,
```
noot, line 41: is this meant to send bundles (ie `Vec<SequenceAction>`) like the name, or it is intended to send individual `SequenceAction`s?

crates/astria-composer/src/searcher/collector.rs
```diff
@@ -36,7 +36,7 @@ type StdError = dyn std::error::Error;
 /// passing them downstream for the searcher to process. Thus, a searcher can have multiple
 /// collectors running at the same time funneling data from multiple rollup nodes.
 #[derive(Debug)]
-pub(super) struct Collector {
+pub(super) struct GethCollector {
```
noot, line 39: can you remove these changes that were from your other PR?

bharath-123: yes i will do that. I thought i would work these changes in order i.e in this order https://github.com/astriaorg/astria/issues/795 . Merge each PR to main. 

What do you think?

crates/astria-composer/src/composer.rs
bharath-123: Right now, it sends individual bundles. converting it to an array would require batching of sorts in `GethCollector`(or `EthCollector`) which can complicate things for now. We can do it if required.

crates/astria-composer/src/composer.rs
```diff
@@ -66,11 +90,13 @@ impl Composer {
         let Self {
             api_server,
             searcher,
+            executor_handle,
         } = self;
 
         let api_task =
             tokio::spawn(async move { api_server.await.wrap_err("api server ended unexpectedly") });
         let searcher_task = tokio::spawn(searcher.run());
+        let _ = executor_handle.send_bundles;
```
itamarreif, line 99: if you added this to avoid the errors from an unused variable, you can name it `_executor_handle` to avoid getting that error

crates/astria-composer/src/composer.rs
itamarreif: i'd keep the name `serialized_rollup_transactions_tx/rx` as it appears in `searcher`

crates/astria-composer/src/searcher/mod.rs
```diff
@@ -77,7 +76,11 @@ impl Searcher {
     /// Errors are returned in the following scenarios:
     /// + failed to connect to the eth RPC server;
     /// + failed to construct a sequencer clinet
-    pub(super) fn from_config(cfg: &Config) -> eyre::Result<Self> {
+    pub(super) fn from_config(
+        cfg: &Config,
+        serialized_rollup_transactions_tx: Sender<SequenceAction>,
```
itamarreif, line 81: nit: i would qualify these as `mpsc::Sender/Receiver` since we have `watch::Sender/Receiver`s here as well.

itamarreif (approved):
looks good aside from a couple nits regarding naming

crates/astria-composer/src/composer.rs
bharath-123: yes sounds good. will do that

crates/astria-composer/src/searcher/mod.rs
bharath-123: sure sure

crates/astria-composer/src/searcher/mod.rs
```diff
@@ -76,7 +75,11 @@ impl Searcher {
     /// Errors are returned in the following scenarios:
     /// + failed to connect to the eth RPC server;
     /// + failed to construct a sequencer clinet
-    pub(super) fn from_config(cfg: &Config) -> eyre::Result<Self> {
+    pub(super) fn from_config(
```
SuperFluffy, line 78: This constructor is now a misnomer. Please don't add fields to this.

As a side note (not focus of this PR): we should have explicit constructors for all the types instead of treating them like "grab bags" from the config.

crates/astria-composer/src/searcher/mod.rs
```diff
@@ -76,7 +75,11 @@ impl Searcher {
...
+        cfg: &Config,
+        serialized_rollup_transactions_tx: mpsc::Sender<SequenceAction>,
+        serialized_rollup_transactions_rx: mpsc::Receiver<SequenceAction>,
+    ) -> eyre::Result<Self> {
```
SuperFluffy, line 82: In conductor we have a similar construction, but where the constructor returns a tuple `(Executor, ExecutorHandle)`. Can you do the same thing here?

crates/astria-composer/src/composer.rs
```diff
@@ -28,6 +32,13 @@ pub struct Composer {
     /// pending transactions from them and wraps them as sequencer transactions
     /// for submission.
     searcher: Searcher,
+    /// The handle to communicate `SequenceActions` to the Executor
+    /// This is at the Composer level to allow its sharing to various different collectors.
+    executor_handle: ExecutorHandle,
+}
+
+struct ExecutorHandle {
```
SuperFluffy, line 40: This type should be defined in the executor module

crates/astria-composer/src/composer.rs
```diff
@@ -66,11 +90,13 @@ impl Composer {
         let Self {
             api_server,
             searcher,
+            executor_handle,
         } = self;
 
         let api_task =
             tokio::spawn(async move { api_server.await.wrap_err("api server ended unexpectedly") });
         let searcher_task = tokio::spawn(searcher.run());
+        let _ = executor_handle.serialized_rollup_transactions_tx;
```
SuperFluffy, line 99: What's this doing?

SuperFluffy (changes requested):
I think this PR needs fundamental changes. `executor` should be singularly responsible for the creation of a handle to it, instead of receiving its handle "from the top".

crates/astria-composer/tests/blackbox/helper/mod.rs
```diff
@@ -50,11 +51,6 @@ pub struct TestComposer {
 pub async fn spawn_composer(rollup_ids: &[&str]) -> TestComposer {
     Lazy::force(&TELEMETRY);
 
-    assert!(
-        !rollup_ids.is_empty(),
-        "must provide at least one rollup ID for tests"
-    );
-
```
bharath-123, line 57: Removed this because we can have deployments where there are no geth collectors but only one gRPC collector.

crates/astria-composer/src/geth_collector/mod.rs
SuperFluffy, line 1: Please just `git mv` this to `geth_collector.rs`. Looks lke `geth_collector` is a single-file module, so no need to the extra directory structure.

I think I wouldn't mind having a collectors module with this structure (module structure need not be the same as the task hierchary under which the collectors run).
```
collectors/
    geth.rs
    grpc.rs
```

crates/astria-composer/src/composer_status.rs
```diff
@@ -0,0 +1,20 @@
+/// Announces the current status of the Composer for other modules in the crate to use
+#[derive(Debug, Default)]
+pub(super) struct ComposerStatus {
```
SuperFluffy, line 3: Please move this type into `composer` and refer to it as `composer::Status`.

crates/astria-composer/src/composer_status.rs
SuperFluffy, line 1: Doesn't warrant an extra module IMO

crates/astria-composer/src/api.rs
```diff
@@ -20,29 +20,29 @@ use serde::Serialize;
 use tokio::sync::watch;
 use tracing::debug;
 
-use crate::searcher;
+use crate::composer_status::ComposerStatus;
 
 pub(super) type ApiServer = axum::Server<AddrIncoming, IntoMakeService<Router>>;
 
-type SearcherStatus = watch::Receiver<searcher::Status>;
+type ComposerStatusWatch = watch::Receiver<ComposerStatus>;
```
SuperFluffy, line 27: Please no Hungarian notation (or anything that goes into that direction).
```suggestion
type ComposerStatus = watch::Receiver<ComposerStatus>;
```

crates/astria-composer/src/composer.rs
```diff
@@ -61,24 +136,144 @@ impl Composer {
...
+        }
+    }
+
+    Ok(())
+}
+
+fn reconnect_exited_collector(
+    collector_statuses: &mut HashMap<String, watch::Receiver<geth_collector::Status>>,
+    collector_tasks: &mut JoinMap<String, eyre::Result<()>>,
+    serialized_rolup_transactions_tx: Sender<SequenceAction>,
```
SuperFluffy, line 254:
```suggestion
    serialized_rollup_transactions_tx: Sender<SequenceAction>,
```

crates/astria-composer/src/composer.rs
```diff
@@ -90,3 +285,91 @@ fn report_exit(task_name: &str, outcome: Result<eyre::Result<()>, JoinError>) {
         }
     }
 }
+
+#[cfg(test)]
```
SuperFluffy, line 289: Please move this module into a new file `composer/tests.rs`

crates/astria-composer/src/composer.rs
```diff
@@ -61,24 +136,144 @@ impl Composer {
...
+            tokio::select!(
+            o = &mut api_task => {
+                    report_exit("api server unexpectedly ended", o);
+                    return Ok(());
+            },
+            o = &mut executor_task => {
+                    report_exit("executor unexpectedly ended", o);
+                    return Ok(());
+            },
+            Some((rollup, collector_exit)) = geth_collector_tasks.join_next() => {
```
SuperFluffy, line 184: IMO this makes sense the way this crate was structured so far. I think it will make sense to move the entire geth collectors handling into its dedicated type so that `composer` is only concerned with watching the "geth collectors supervisor" (terrible name) task. Future PR

tools/protobuf-compiler/src/main.rs
```diff
@@ -105,8 +105,8 @@ fn clean_non_astria_code(generated: &mut ContentMap) {
         .filter(|name| !name.starts_with("astria."))
         .cloned()
         .collect();
-    // also mask mod.rs because we need are defining it
-    foreign_file_names.remove("mod.rs");
+    // also mask geth_collector because we need are defining it
+    foreign_file_names.remove("geth_collector");
```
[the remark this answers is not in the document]
bharath-123: this is was a mistake during renaming which was removed.

buf.work.yaml
```diff
@@ -1,4 +1,4 @@
 version: v1
 directories:
   - proto/executionapis
-  - proto/sequencerapis
+  - proto/sequencerapis
```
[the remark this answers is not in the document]
bharath-123: got it! will set to that.

crates/astria-core/src/generated/mod.rs
```diff
@@ -25,4 +25,4 @@ pub mod primitive {
 pub mod sequencer {
     #[path = "astria.sequencer.v1.rs"]
     pub mod v1;
-}
+}
```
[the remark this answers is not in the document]
bharath-123: yikes its a newline thing :(

justfile
```diff
@@ -68,3 +68,4 @@ _lint-proto:
   buf format -d --exit-code
   buf breaking proto/executionapis --against 'buf.build/astria/execution-apis'
   buf breaking proto/sequencerapis --against 'buf.build/astria/astria'
+
```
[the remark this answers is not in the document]
bharath-123: newline thing. will see what i can do about it

tools/protobuf-compiler/src/main.rs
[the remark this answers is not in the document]
bharath-123: my bad. this was an issue during renaming

crates/astria-composer/tests/blackbox/helper/mod.rs
```diff
@@ -50,11 +51,6 @@ pub struct TestComposer {
 pub async fn spawn_composer(rollup_ids: &[&str]) -> TestComposer {
     Lazy::force(&TELEMETRY);
 
-    assert!(
```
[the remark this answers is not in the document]
bharath-123: this can happen if we are not running geth collectors but we are running a gRPC collector. so there could be a case where rollup_ids is empty. I can revert the code but the tests will look a bit clunky.

crates/astria-composer/src/searcher/mod.rs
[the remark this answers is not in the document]
bharath-123: yes.

crates/astria-composer/src/composer_status.rs
bharath-123: sure will do

crates/astria-composer/src/composer_status.rs
bharath-123: cool cool

crates/astria-composer/src/geth_collector/mod.rs
bharath-123: sounds good will do!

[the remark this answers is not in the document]
bharath-123: will add a collectors module!

crates/astria-composer/tests/blackbox/helper/mod.rs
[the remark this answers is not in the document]
bharath-123: reverted

crates/astria-composer/src/main.rs
```diff
@@ -51,9 +51,17 @@ async fn main() -> ExitCode {
 
     info!(config = cfg_ser, "initializing composer",);
 
-    Composer::from_config(&cfg)
+    match Composer::from_config(&cfg)
```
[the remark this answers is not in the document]
bharath-123: will rever.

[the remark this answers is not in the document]
bharath-123: don't think i can revert this since it will cause clippy errors.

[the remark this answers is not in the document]
SuperFluffy: Yeah, that's valid. Sometimes `clippy` starts complaining about lines that were ok before - although in this case it's interesting that it had a false negative (meaning, it should have caught this before but didn't).

Please change this to to use tracing everywhere though while you are at it:

```rust
let composer = match Composer::new(&cfg) {
    Err(error) => {
        error!(%error, "failed initializing Composer");
        return ExitCode::Failure;
    }
    Ok(composer) => composer,
}

match composer.run_until_stopped().await {
    Ok(()) => {
        info!("composer stopped");
        ExitCode::SUCCESS
    }
    Err(error) => {
        error!(%error, "Composer exited with errro");
        ExitCode::FAILURE;
}
```

crates/astria-composer/src/collectors/geth.rs
```diff
@@ -46,16 +46,16 @@ type StdError = dyn std::error::Error;
 /// them downstream for further processing.
 ///
 /// It is responsible for fetching pending transactions submitted to the rollup Geth nodes and then
-/// passing them downstream for the executor to process. Thus, a searcher can have multiple
+/// passing them downstream for the executor to process. Thus, a composer can have multiple
 /// collectors running at the same time funneling data from multiple rollup nodes.
 #[derive(Debug)]
-pub(super) struct GethCollector {
+pub(crate) struct GethCollector {
```
SuperFluffy, line 52: Please name this just `Geth`

crates/astria-composer/src/collectors/mod.rs
```diff
@@ -0,0 +1 @@
+pub(crate) mod geth;
```
SuperFluffy, line 1: Rexport `geth::Geth` here, so that we can refer to it as `collectors::Geth` elsewhere

crates/astria-composer/src/composer.rs
```diff
@@ -15,42 +26,126 @@ use crate::{
         self,
         ApiServer,
     },
-    searcher::Searcher,
+    collectors,
+    collectors::geth::GethCollector,
```
SuperFluffy, line 30: Just import the `collectors` module and refer to the geth collectors as `collectors::Geth`

crates/astria-composer/src/executor/mod.rs
```diff
@@ -103,6 +104,11 @@ pub(super) struct Executor {
     max_bytes_per_bundle: usize,
 }
 
+#[derive(Clone)]
+pub(super) struct ExecutorHandle {
```
SuperFluffy, line 108: Remove from this PR, this is not about flattening

crates/astria-composer/src/executor/mod.rs
```diff
@@ -130,13 +136,12 @@ impl Executor {
     pub(super) fn new(
         sequencer_url: &str,
         private_key: &SecretString,
-        serialized_rollup_transactions_rx: mpsc::Receiver<SequenceAction>,
```
SuperFluffy, line 133: I agree with the change, but not in this PR. Keep passing in the channel here for now, we will construct the `Handle` in a followup

SuperFluffy (approved):
Looks good, thanks for iterating on this. Please address the last few comments and then merge.

PR #797, opened 2024-03-08
fix(conductor): simplify mapping Celestia heights to Sequencer heights

SuperFluffy:
## Summary
Simplifies mapping Celestia heights to Sequencer heights by adding the source Celestia height to final type that contains the reconstructed sequencer block information. Fixes an assertion (and panic) being triggered by making it unecessary.

## Background
Conductor kept a map of `sequencer_height -> celestia_height` to track which blobs contained which sequencer height. This implementation was unnecessary and error prone, and ultimately out-of-sync with the rest of the Celestia reader. It's select-loop contained an `assert!` macro intended to tie assumptions about height-progress together and ultimately revealed a bug.

Instead of keeping an extra map, the Celestia heights can be flattened directly into the objects that are reconstructed from the blobs read off of Celestia. This makes the assertion unnecessary and simplifies the code.

## Changes
- Add a field `ReconstructedSequencerBlock::celestia_height` to track the Celestia source heights.
- Remove the `SequencerHeightToCelestiaHeight` map
- Streamline the height tracking in the block stream.

## Testing
Should be tested end-to-end

### DIFF
--- a/crates/astria-conductor/src/celestia/mod.rs
+++ b/crates/astria-conductor/src/celestia/mod.rs
@@ -314,8 +314,10 @@ impl TrackHeights {
         }
     }
 
-    fn increment_next(&mut self) {
-        self.next_height += 1;
+    fn increment_next_height_to_fetch(&mut self) {
+        self.next_height
+            .checked_add(1)
+            .expect("this value should never reach u64::MAX");
     }
 
     fn update_reference_height_if_greater(&mut self, height: u64) -> bool {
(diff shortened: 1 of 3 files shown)


crates/astria-conductor/src/celestia/mod.rs
```diff
@@ -302,13 +296,50 @@ impl Reader {
...
+        let max_permissible = self.reference_height.saturating_add(self.variance.into());
+        if self.next_height < max_permissible && self.next_height <= self.last_observed {
+            Some(self.next_height)
+        } else {
+            None
+        }
+    }
+
+    fn increment_next(&mut self) {
+        self.next_height += 1;
```
joroshiba, line 318: Should this be a checked add?

SuperFluffy: It doesn't hurt.

PR #802, opened 2024-03-08
feat(sequencer)!: include deposit data as part of rollup data

### LINKED ISSUE #774: sequencer/relayer: write deposit info to rollup blobs
follow up to #768 - `Deposit`s need to be written to DA for usage by the conductor, this can be implemented by including them in the `SequencerBlock` structure returned by the sequencer API.

noot:
## Summary
include deposit data generated from `BridgeLockAction`s as part of rollup data. this allows the relayer to write the deposit data to DA as part of rollup data, and for rollups to receive deposits from the conductor (once relayer/conductor are updated to use the sequencer API).

## Background
required for the bridge accounts implementation

## Changes
- implement `RollupData` enum type which consists of either `SequencedData` or `Deposit`. this type is used to wrap either a `SequenceAction`'s byte data or a `Deposit`.
- this type is now what's written to each rollup data field + the rollup data merkle tree in the form of a proto-encoded `RollupData` . this allows the rollup execution node to decode each piece of data it receives from conductor into a `SequencedData` or `Deposit`. a rollup can still choose to ignore deposits.
- update building/storing of `SequencerBlock` in the app to include deposits generated by the block's execution.

## Testing
unit tests + i ran it with cometbft

## Breaking Changelist
- this changes the rollup data commitment by including deposits, as well as formatting them as encoded `RollupData`. thus after this PR, commitments made before this PR will not be valid.

## Related Issues

closes  #774


crates/astria-core/src/sequencer/v1alpha1/block/mod.rs
```diff
@@ -209,7 +211,7 @@ impl SequencerBlockError {
         Self(SequencerBlockErrorKind::RollupIdsNotInSequencerBlock)
     }
 
-    fn signed_transaction_prootbof_decode(source: prost::DecodeError) -> Self {
+    fn signed_transaction_protobuf_decode(source: prost::DecodeError) -> Self {
```
joroshiba, line 214: lol, good catch

proto/sequencerapis/astria/sequencer/v1alpha1/block.proto
```diff
@@ -14,7 +14,8 @@ import "astria_vendored/tendermint/types/types.proto";
 message RollupTransactions {
   // The 32 bytes identifying a rollup. Usually the sha256 hash of a plain rollup name.
   bytes id = 1;
-  // The serialized opaque bytes of the rollup transactions.
+  // The serialized bytes of the rollup data.
+  // Each entry is a protobuf-encoded `RollupData` message.
   repeated bytes transactions = 2;
```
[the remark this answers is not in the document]
noot: yeah i tried that and the code got messy fast with constantly converting between `RollupData` and bytes eg in the merkle tree and verification steps. `RollupData` doesn't need to be represented at any point really except at the entry/exit points so it seemed like a lot of excessive allocations/conversions for nothing

joroshiba (approved):

proto/sequencerapis/astria/sequencer/v1alpha1/block.proto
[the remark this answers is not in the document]
noot: another option would be to have the `RollupTransactions` proto contain `RollupData`, but have the `RollupTransactions` native type contain the encoded bytes, but that felt a bit not-great also

proto/sequencerapis/astria/sequencer/v1alpha1/block.proto
```diff
@@ -116,3 +117,16 @@ message FilteredSequencerBlock {
...
+
+// A piece of data that is sent to a rollup execution node.
+//
+// The data can be either sequenced data (originating from a `SequenceAction`
+// submitted by a user) or a `Deposit` originating from a `BridgeLockAction`.
+//
+// The rollup node receives this type from conductor and must decode them accordingly.
+message RollupData {
+  oneof value {
+    bytes sequenced_data = 1;
```
SuperFluffy, line 129: Are these just opaque bytes?

proto/sequencerapis/astria/sequencer/v1alpha1/block.proto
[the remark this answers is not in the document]
SuperFluffy: I can see how moving in and out of these types could get weird.

I had a similar idea for other parts of the protobufs that we only "materialize"  the bytes once we actually need them.

One thing that is worrisome though is that now there are no (type-level) guarantees what's inside `transactions`. It's entirely application (= sequencer) defined.

crates/astria-sequencer/src/proposal/commitment.rs
```diff
@@ -22,35 +32,49 @@ impl GeneratedCommitments {
...
+    deposits: HashMap<RollupId, Vec<Deposit>>,
 ) -> GeneratedCommitments {
-    let rollup_ids_to_txs = astria_core::sequencer::v1alpha1::group_sequence_actions_in_signed_transaction_transactions_by_rollup_id(signed_txs);
+    use prost::Message as _;
+
+    let mut rollup_ids_to_txs =
+        group_sequence_actions_in_signed_transaction_transactions_by_rollup_id(signed_txs);
     let rollup_ids_root = merkle::Tree::from_leaves(rollup_ids_to_txs.keys()).root();
 
+    for (rollup_id, deposit) in deposits {
```
SuperFluffy, line 62: `rollup_ids_to_txs` *must* be srotede after this operation.

SuperFluffy (changes requested):
`rollup_ids_to_txs` must be sorted after populating it with more information.

crates/astria-sequencer/src/proposal/commitment.rs
noot: fixed

proto/sequencerapis/astria/sequencer/v1alpha1/block.proto
noot: yes, they're what's in sequence actions

crates/astria-core/src/sequencer/v1alpha1/block/mod.rs
```diff
@@ -1275,3 +1291,71 @@ enum DepositErrorKind {
     #[error("the asset ID length is not 32 bytes")]
     IncorrectAssetIdLength(#[source] asset::IncorrectAssetIdLength),
 }
+
+/// A piece of data that is sent to a rollup execution node.
+///
+/// The data can be either sequenced data (originating from a [`SequenceAction`]
+/// submitted by a user) or a [`Deposit`] originating from a [`BridgeLockAction`].
+///
+/// The rollup node receives this type, protobuf-encoded, from conductor and
```
SuperFluffy, line 1300:
```suggestion
/// The rollup node receives this type as opaque, protobuf-encoded bytes from conductor, and

File: crates/astria-core/src/sequencer/v1alpha1/block/mod.rs [hunk 6]
```diff
@@ -1275,3 +1291,71 @@ enum DepositErrorKind {
...
+    fn deposit(source: DepositError) -> Self {
+        Self(RollupDataErrorKind::Deposit(source))
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+enum RollupDataErrorKind {
+    #[error("the expected field in the raw source type was not set: `{0}`")]
+    FieldNotSet(&'static str),
+    #[error(transparent)]
```
SuperFluffy, line 1359: Please don't make this `transparent`, but give it a proper description, in this case probably:
```suggestion
    #[error("failed to validate `deposit` field)]
```

crates/astria-core/src/sequencer/v1alpha1/block/mod.rs
```diff
@@ -1275,3 +1291,71 @@ enum DepositErrorKind {
...
+    Deposit(#[from] DepositError),
```
SuperFluffy, line 1360: Please make this `#[source]`. Not a big deal, but to be in line with the rest of the crate.

crates/astria-sequencer/src/bridge/state_ext.rs
```diff
@@ -198,6 +227,15 @@ pub(crate) trait StateWriteExt: StateWrite {
             self.nonverifiable_delete(key);
         }
     }
+
+    #[instrument(skip(self))]
+    async fn clear_block_deposits(&mut self) -> Result<()> {
+        let deposit_rollup_ids = self.get_deposit_rollup_ids().await?;
```
SuperFluffy, line 233: context

proto/sequencerapis/astria/sequencer/v1alpha1/block.proto
SuperFluffy: What if we called this `opaque_transaction_bytes` to really bring home what this is supposed to be?

crates/astria-core/src/sequencer/v1alpha1/block/mod.rs
```diff
@@ -667,35 +675,43 @@ impl SequencerBlock {
...
         }
-        rollup_base_transactions.sort_unstable_keys();
+        for (id, deposits) in deposits {
+            rollup_datas.entry(id).or_default().extend(
+                deposits
+                    .into_iter()
+                    .map(|deposit| RollupData::Deposit(deposit).into_raw().encode_to_vec()),
+            );
+        }
+        rollup_datas.sort_unstable_keys();
```
SuperFluffy, line 691: Can you add a comment above this line to bring home why sorting is necessary? I should have done that in the original impl. It's too easy to overlook this. E.g:

```rust
// XXX: The rollup data must be sorted by its keys before constructing the merkle tree.
// Since its constructed from non-deterministically ordered sources there is otherwise no
// guarantee that the same data will give the root.
```

SuperFluffy (approved):
Approved with comments

PR #803, opened 2024-03-08
fix(conductor): update reference celestia height correctly

joroshiba:
## Summary
Updates the celestia reference height even when the execution channel is full

## Background
During large syncs the channel gets full of blocks to execute from celestia and then with smaller variance heights especially would exit the window since the window wasn't being updated on scheduled block updates.

## Changes
- Add logs about sequencer blocks received from Celestia
- Update the reference base for celestia variance window correctly.

## Testing
Run manually via dev-cluster

### DIFF
--- a/crates/astria-conductor/src/celestia/mod.rs
+++ b/crates/astria-conductor/src/celestia/mod.rs
@@ -55,7 +55,10 @@ use sequencer_client::tendermint::{
 use tokio::{
     select,
     sync::{
-        mpsc::error::{SendError, TrySendError},
+        mpsc::error::{
+            SendError,
+            TrySendError,
+        },
         oneshot,
     },
 };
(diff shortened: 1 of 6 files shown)


crates/astria-conductor/src/celestia/mod.rs
```diff
@@ -225,7 +230,13 @@ impl Reader {
                         }
                         Err(TrySendError::Full(block)) => {
                             trace!("executor channel is full; rescheduling block fetch until the channel opens up");
-                            scheduled_block = executor.clone().send_firm_block(block).boxed().fuse();
+                            let executor_clone = executor.clone();
+                            // Using async block returning celestia height so reference height be updated
```
SuperFluffy, line 234: Small nit: just mention the need for returning the height. No need to mention the async block (since that's obvious from the next line)

SuperFluffy (approved):
Good find

PR #807, opened 2024-03-11
feat(relayer): report information on sequencer block submitted to Celestia

SuperFluffy:
## Summary
For each sequencer block and height that was submitted to Celestia, reports the rollup IDs, Celestia namespaces (derived from rollup IDs), and the number of transactions contained therein. 

## Background
We need to have more visibility into what relayer submits to Celestia.

## Changes
- Emit a JSON formatted field mapping `sequencer height -> (rollup ID, namespace, num of transactions)` at the start of celestia submission

## Testing
Punted, needs to checked in observability.


crates/astria-sequencer-relayer/src/relayer/write.rs
```diff
@@ -264,21 +308,17 @@ impl BlobSubmitter {
 /// Submits new blobs Celestia.
 ///
 /// # Panics
-/// Panics if `blobs` or `sequencer_heights` are empty. This function
+/// Panics if `blobs`. This function
```
noot, line 311: > Panics if `blobs`
panics if what?

crates/astria-telemetry/src/display.rs
```diff
@@ -26,13 +37,75 @@ pub fn hex<T: AsRef<[u8]> + ?Sized>(bytes: &T) -> Hex<'_> {
...
+        struct WriterFormatter<'a, 'b: 'a> {
+            inner: &'a mut Formatter<'b>,
+        }
+
+        impl<'a, 'b> io::Write for WriterFormatter<'a, 'b> {
+            fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+                // NOTE: This implementation is lifted straight from serde_json.
+                // Safety: the serializer below only emits valid utf8 when using
+                // the default formatter.
+                let s = unsafe { str::from_utf8_unchecked(buf) };
```
noot, line 90: is this guaranteed safe? what does `non-utf8` fields mean, since this is serializing something? if this function already returns a result why not just return an error instead of panicking?

joroshiba:
Test failures likely related? Tests that are running long in relayer

crates/astria-telemetry/src/display.rs
SuperFluffy: I have added links to serde_json's `impl Display for Value`, as that's lifted from there. But yeah - it's guaranteed safe since the serde_json serializer will return an error if it encounters non-utf8 strings.

SuperFluffy:
> Test failures likely related? Tests that are running long in relayer

I haven't observed this?

crates/astria-sequencer-relayer/src/relayer/write/mod.rs
```diff
@@ -264,21 +290,20 @@ impl BlobSubmitter {
...
-/// should only be called if there is something to submit.
+/// Panics if `blocks` is empty. This function should only be called if there is something to
+/// submit.
 #[instrument(
     skip_all,
-    fields(
-        num_blobs = blobs.len(),
-        sequencer_heights = %ReportSequencerHeights(&sequencer_heights),
-))]
+    fields(blocks = %telemetry::display::json(&blocks.infos))
```
SuperFluffy, line 297: All this information gets injected into the span. My reasoning is that I wanted to have it in every single event in here. Maybe that's excessive and should just be logged in the first event?

joroshiba: Could we tie a uuid or something like that to the blocks? Then we can trace with a UUID and have blocks.info in the first debug log.

SuperFluffy: This should technically be handled by our observability platform. Inject UUIDs manually duplicates that work.

I have removed the field from the span, emitting all this data in an event at the start of the submission tasks instead.

noot (approved):

PR #808, opened 2024-03-11
chore(composer): logging of rollup ids on startup, and submission

joroshiba:
## Summary
Fixed the bundle instrumentation, and added logging of rollup ids on startup.

## Background
It was unclear from logs what rollup ids were being used, additionally upon submission logs had gibberish of full bundle bytes, this simplifies while providing more useful information.

## Changes
- Added log when rollup names parsed into Rollup ID
- Don't instrument whole bundle, instead a list of rollup ids and counts within bundle.

## Testing
Used while manually testing dev-cluster.


crates/astria-composer/src/searcher/executor/mod.rs
```diff
@@ -466,3 +467,30 @@ fn sha256(data: &[u8]) -> [u8; 32] {
...
+                *counts.entry(action.rollup_id).or_insert(0) += 1;
+            }
+        }
+
+        f.write_char('[')?;
+        let mut counts_iter = counts.iter();
+        if let Some((rollup_id, count)) = counts_iter.next() {
+            write!(f, "{rollup_id}: {count}")?;
+        };
+        for (rollup_id, count) in counts {
```
noot, line 489: in `counts` or `counts_iter`? does this double print the first/one of the rollups?

crates/astria-composer/src/searcher/collector.rs
```diff
@@ -75,8 +76,14 @@ impl Collector {
         new_bundles: Sender<SequenceAction>,
     ) -> Self {
         let (status, _) = watch::channel(Status::new());
+        let rollup_id = RollupId::from_unhashed_bytes(&chain_name);
+        info!(
+            rollup_name = %chain_name,
+            rollup_id = %rollup_id,
+            "Creating new collector for rollup",
```
SuperFluffy, line 83: messages should started small-cased

crates/astria-composer/src/searcher/collector.rs
```diff
@@ -75,8 +76,14 @@ impl Collector {
         new_bundles: Sender<SequenceAction>,
     ) -> Self {
         let (status, _) = watch::channel(Status::new());
+        let rollup_id = RollupId::from_unhashed_bytes(&chain_name);
+        info!(
```
SuperFluffy, line 80: Commenting here because github doesn't permit commenting far away from diffs:

I think `rollup_id` should be injected into `Collector::run_until_stopped`, and a message generated there, like "running collector` or something like that.

crates/astria-composer/src/searcher/executor/mod.rs
```diff
@@ -165,7 +166,7 @@ impl Executor {
     }
 
     /// Create a future to submit a bundle to the sequencer.
-    #[instrument(skip(self), fields(nonce.initial = %nonce))]
+    #[instrument(skip_all, fields(nonce.initial = %nonce, rollup_counts = %ReportBundleRollupIdCounts(&bundle)))]
```
SuperFluffy, line 169: I think we should pay the price for this in `BundleFactory` while constructing these bundles, similar to what I have done here: https://github.com/astriaorg/astria/pull/807

crates/astria-composer/src/searcher/executor/mod.rs
SuperFluffy: Agree with @noot. This should consume `counts_iter`

crates/astria-composer/src/executor/bundle_factory/mod.rs
```diff
@@ -27,14 +31,16 @@ enum SizedBundleError {
...
+pub(super) struct SizedBundle {
     /// The buffer of actions
     buffer: Vec<Action>,
     /// The current size of the bundle in bytes. This is equal to the sum of the size of the
     /// `seq_action`s + `ROLLUP_ID_LEN` for each.
     curr_size: usize,
     /// The max bundle size in bytes to enforce.
     max_size: usize,
+    /// Mapping of rollup id to the number of sequence actions for that rollup id in the bundle.
+    rollup_counts: HashMap<RollupId, u32>,
```
noot, line 43:
```suggestion
    rollup_transaction_counts: HashMap<RollupId, u32>,
```

crates/astria-composer/src/executor/bundle_factory/mod.rs
```diff
@@ -71,9 +82,31 @@ impl SizedBundle {
...
+    }
+}
+
+impl std::fmt::Display for SizedBundle {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "SizedBundle {{ size: {}, actions: [", self.curr_size)?;
+
+        let mut counts_iter = self.rollup_counts.iter();
+        if let Some((rollup_id, count)) = counts_iter.next() {
+            write!(f, "{rollup_id}: {count}, ")?;
```
noot, line 101:
```suggestion
            write!(f, "{rollup_id}: {count}")?;
```

noot, line 101:
```suggestion
            write!(f, "{rollup_id}: {count}")?;
```

crates/astria-composer/src/executor/bundle_factory/mod.rs
```diff
@@ -71,9 +82,31 @@ impl SizedBundle {
...
+        };
+        for (rollup_id, count) in counts_iter {
+            f.write_str(", ")?;
+            write!(f, "{rollup_id}: {count}, ")?;
```
noot, line 105:
```suggestion
            write!(f, "{rollup_id}: {count}")?;
```

crates/astria-composer/src/executor/bundle_factory/mod.rs
```diff
@@ -71,9 +82,31 @@ impl SizedBundle {
...
+
+    /// Returns true if the bundle is empty.
+    pub(super) fn is_empty(&self) -> bool {
+        self.buffer.is_empty()
+    }
+}
+
+impl std::fmt::Display for SizedBundle {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "SizedBundle {{ size: {}, actions: [", self.curr_size)?;
```
noot, line 97:
```suggestion
        write!(f, "SizedBundle {{ size: {} bytes, actions: [", self.curr_size)?;
```

crates/astria-composer/src/executor/mod.rs
```diff
@@ -192,15 +193,16 @@ impl Executor {
     }
 
     /// Create a future to submit a bundle to the sequencer.
-    #[instrument(skip(self), fields(nonce.initial = %nonce))]
-    fn submit_bundle(&self, nonce: u32, bundle: Vec<Action>) -> Fuse<Instrumented<SubmitFut>> {
+    // TODO: fix instrumentation
```
noot, line 196: what about it?

crates/astria-composer/src/executor/bundle_factory/mod.rs
```diff
@@ -153,12 +186,11 @@ impl BundleFactory {
...
     ///
     /// Returns an empty bundle if there are no bundled transactions.
-    pub(super) fn pop_now(&mut self) -> Vec<Action> {
+    pub(super) fn pop_now(&mut self) -> SizedBundle {
         self.finished
             .pop_front()
             .or_else(|| Some(self.curr_bundle.flush()))
-            .map(SizedBundle::into_actions)
-            .unwrap_or_default()
+            .unwrap_or(SizedBundle::new(self.curr_bundle.max_size))
```
noot, line 193: why is this changed? it looks like `into_actions()` is always called on the returned value?

crates/astria-composer/src/executor/bundle_factory/mod.rs
```diff
@@ -167,11 +199,10 @@ pub(super) struct NextFinishedBundle<'a> {
 }
 
 impl<'a> NextFinishedBundle<'a> {
-    pub(super) fn pop(self) -> Vec<Action> {
+    pub(super) fn pop(self) -> SizedBundle {
         self.bundle_factory
             .finished
             .pop_front()
-            .map(SizedBundle::into_actions)
```
noot, line 174: same here

crates/astria-composer/src/executor/mod.rs
itamarreif: forgot to remove this oops

crates/astria-composer/src/executor/bundle_factory/mod.rs
itamarreif: Popping returns a `SizedBundle` now, which contains info that we use for tracing (instruments `Executor.submit_bundle()`).
`into_actions` is called in all the tests since we dont care about the byte count and rollup counts there.

Cargo.toml
```diff
@@ -95,3 +95,4 @@ tracing = "0.1"
 tryhard = "0.5.1"
 which = "4.4.0"
 wiremock = "0.5"
+insta = { version = "1.36.1", features = ["json"] }
```
SuperFluffy:
```suggestion
insta = "1.36.1"
```

crates/astria-composer/Cargo.toml
```diff
@@ -66,5 +68,7 @@ tokio-test = { workspace = true }
 tendermint-rpc = { workspace = true }
 wiremock = { workspace = true }
```
SuperFluffy, line 70: nit: please order these alphabetically (astria-configs first, then workspace)

crates/astria-composer/src/collectors/geth.rs
```diff
@@ -82,8 +83,14 @@ impl Geth {
     /// Initializes a new collector instance
     pub(crate) fn new(chain_name: String, url: String, executor_handle: executor::Handle) -> Self {
         let (status, _) = watch::channel(Status::new());
+        let rollup_id = RollupId::from_unhashed_bytes(&chain_name);
+        info!(
```
SuperFluffy, line 87: nit: I prefer after-the-fact event messages like "I have done the thing" over "I am going to do the thing" especially for short, non-async constructors.

crates/astria-composer/src/executor/mod.rs
```diff
@@ -192,15 +193,19 @@ impl Executor {
     }
 
     /// Create a future to submit a bundle to the sequencer.
-    #[instrument(skip(self), fields(nonce.initial = %nonce))]
-    fn submit_bundle(&self, nonce: u32, bundle: Vec<Action>) -> Fuse<Instrumented<SubmitFut>> {
+    #[instrument(skip_all, fields(
+        nonce.initial = %nonce,
+        bundle_bytes = %bundle.curr_size,
+        rollup_actions = %telemetry::display::json(&bundle.rollup_counts)
```
SuperFluffy, line 199: hm, let's leave this for now but this might end up being very noisy. I haven't thought about the all the events emitted by this future, but just a thought.

crates/astria-composer/src/executor/bundle_factory/mod.rs
```diff
@@ -24,17 +33,34 @@ enum SizedBundleError {
     SequenceActionTooLarge(SequenceAction),
 }
 
+pub(super) struct RollupCountsReport(HashMap<RollupId, u32>);
+
+impl Serialize for RollupCountsReport {
```
SuperFluffy, line 38: Does this need a manual `Serialize` impl? We need it in conductor because we construct information ad-hoc, but I think in this case you might get away with something like this (since `RollupId` should impl `Serialize`; you might need to adjust the features of `astria-core`):
```rust
#[derive(serde::Serialize)]
#[serde(transparent)]
pub(super) struct RollupCountsReport(HashMap<RollupId, u32>);
```
Less work

crates/astria-composer/src/executor/bundle_factory/mod.rs
```diff
@@ -24,17 +33,34 @@ enum SizedBundleError {
     SequenceActionTooLarge(SequenceAction),
 }
 
+pub(super) struct RollupCountsReport(HashMap<RollupId, u32>);
```
SuperFluffy, line 36: Can this remain an implementation deatil of this module? Can you make it non-`pub(super)`?

crates/astria-composer/src/executor/bundle_factory/mod.rs
```diff
@@ -24,17 +33,34 @@ enum SizedBundleError {
...
+        }
+        map.end()
+    }
+}
+
 /// A bundle sequence actions to be submitted to the sequencer. Maintains the total size of the
 /// bytes pushed to it and enforces a max size in bytes passed in the constructor. If an incoming
 /// `seq_action` won't fit in the buffer it is flushed and a new bundle is started.
-struct SizedBundle {
+pub(super) struct SizedBundle {
```
SuperFluffy, line 54: Same with all of these fields - why are they no longer module-private? Due to instrumentation? Can you just provide accessors?

crates/astria-composer/src/executor/bundle_factory/snapshots/astria_composer__executor__bundle_factory__tests__sized_bundle_tests__snapshots.snap
```diff
@@ -0,0 +1,8 @@
+---
+source: crates/astria-composer/src/executor/bundle_factory/tests.rs
+expression: bundle.rollup_counts
+---
+{
+  "0000000000000000000000000000000000000000000000000000000000000000": 1,
+  "0101010101010101010101010101010101010101010101010101010101010101": 1
```
SuperFluffy, line 7: could you make these slightly more complex? Add a count of 2 for example, make the rollup IDs non-zero

SuperFluffy (approved):
I like the additions.

Could you please have a second look over how the submission future is instrumented? I can see this potentially becoming noisy, but I also see the merit in attaching it to every error that the submission fut generates.

We will have better insights into whether this is too noisy or ok once we have obsersvability running.

crates/astria-composer/src/executor/bundle_factory/mod.rs
itamarreif: Changed this `Report` thing to resemble what you did in `conductor` more closely, should get rid of the weird visibility stuff

crates/astria-composer/src/executor/bundle_factory/mod.rs
```diff
@@ -16,7 +16,10 @@ use astria_core::sequencer::v1::{
     RollupId,
     ROLLUP_ID_LEN,
 };
-use serde::Serialize;
+use serde::{
+    self,
```
SuperFluffy: Why are you importing `serde::self`?

crates/astria-composer/src/executor/bundle_factory/mod.rs
```diff
@@ -30,6 +33,7 @@ enum SizedBundleError {
 }
 
 #[derive(Serialize, Clone)]
```
SuperFluffy: Very nit: I think just `#[derive(serde::Serialize)]` would be fine.

crates/astria-composer/src/executor/mod.rs
```diff
@@ -421,9 +424,7 @@ impl Future for SubmitFut {
                     .into_signed(this.signing_key);
                     debug!(
```
SuperFluffy, line 425: Given the submitting things to sequencer is the main task of composer, I think submitting this as `info!` is justified.

crates/astria-composer/src/executor/mod.rs
```diff
@@ -481,9 +482,7 @@ impl Future for SubmitFut {
                         .into_signed(this.signing_key);
                         debug!(
```
SuperFluffy, line 483: Same here - I think making this `info!` is justified.

PR #810, opened 2024-03-12
reafactor(relayer)!: update relayer to use sequencer API to pull blocks

### LINKED ISSUE #790: relayer: use `GetSequencerBlock` in place of tendermint API
We added the new Sequencer API, as a grpc service which can be used to get the blocks directly from the sequencer app instead of needing to transpose SequencerBlocks from CometBFT blocks.

This is required for us to add new bridging txs into SequencerBlocks.

noot:
## Summary
update relayer to use the sequencer gRPC API to pull blocks instead of from cometbft and converting them.

## Background
needed for bridge txs + using the sequencer API in general + less conversions

## Changes
- update the relayer reader to use the sequencer gRPC API `get_sequencer_block` instead of cometbft
- note that the relayer still uses cometbft for the new heights stream via `abci_info`, we probably want to remove this and have an `info`/`latest_height` on the sequencer API

## Testing
blackbox tests (todo: fix)

## Breaking Changelist
- relayer now requires access to the sequencer app gRPC server

## Related Issues

closes #790


crates/astria-sequencer-relayer/src/main.rs
```diff
@@ -3,7 +3,6 @@ use std::process::ExitCode;
 use astria_eyre::eyre::WrapErr as _;
 use astria_sequencer_relayer::{
     metrics_init,
-    telemetry,
```
SuperFluffy, line 6: Good catch. Is there a reexport in `lib.rs` that should also go?

crates/astria-sequencer-relayer/tests/blackbox/helper.rs
```diff
@@ -20,17 +35,23 @@ use serde_json::json;
...
-    endpoint,
     response::Wrapper,
     Id,
 };
 use tokio::{
+    net::TcpListener,
     sync::{
         mpsc,
         oneshot,
+        Mutex,
```
SuperFluffy, line 46: Side comment: tokio recommends to prefer `std::sync::Mutex` over `tokio::sync::Mutex`. `tokio::sync::Mutex` is only needed if you plan to hold locks across await-points (which is tricky in general). In all other cases a `std::sync::Mutex` is better as one doesn't pay for the async-specific bits.

crates/astria-sequencer-relayer/tests/blackbox/main.rs
```diff
@@ -51,14 +51,11 @@ async fn report_degraded_if_block_fetch_fails() {
         actual: readyz.json::<serde_json::Value>().await.unwrap(),
     );
 
+    // don't mount another block, so the relayer will fail to fetch the block
     let abci_guard = sequencer_relayer.mount_abci_response(1).await;
-    let block_guard = sequencer_relayer.mount_bad_block_response(1).await;
```
SuperFluffy, line 55: This made this test more shaky (flaky?) then before, because it relies on relayer attempting to fetch a new bock within `2 * block_time` failing, and reporting that.

I think it's fine for this PR, but could you add a big `NOTE` or even `XXX` to make note of that?

crates/astria-sequencer-relayer/tests/blackbox/main.rs
```diff
@@ -89,17 +88,15 @@ async fn one_block_is_relayed_to_celestia() {
...
     timeout(
         Duration::from_millis(100),
-        futures::future::join(
-            abci_guard.wait_until_satisfied(),
-            block_guard.wait_until_satisfied(),
-        ),
+        abci_guard.wait_until_satisfied(),
     )
     .await
     .expect("requesting abci info and block must have occured");
```
SuperFluffy, line 96: this expect msg is no longer correct

crates/astria-sequencer-relayer/tests/blackbox/main.rs
```diff
@@ -51,14 +51,11 @@ async fn report_degraded_if_block_fetch_fails() {
...
+    // don't mount another block, so the relayer will fail to fetch the block
     let abci_guard = sequencer_relayer.mount_abci_response(1).await;
-    let block_guard = sequencer_relayer.mount_bad_block_response(1).await;
     timeout(
         Duration::from_millis(2 * sequencer_relayer.config.block_time),
-        futures::future::join(
-            abci_guard.wait_until_satisfied(),
-            block_guard.wait_until_satisfied(),
-        ),
+        abci_guard.wait_until_satisfied(),
```
SuperFluffy, line 58: Thought about this some more: this actually is much more flaky than before.

This test now relies on the healthz request triggering after the block-request going through. That's because the timeout will already have returned after `abci_guard` is satisfied.

But there is actually no guarantee that the healthz is peformed after the block-request is done.

SuperFluffy (commented):
Can you provide a handrolled `Guard` for `mount_block_response`? I think this could be relatively easy (didn't check, but this should work):

Do a:
```rust
let (tx, rx) = tokio::sync::mpsc::oneshot::channel();
```
Pass the `tx` part into the mockserver, i.e. something like `blocks.push_back((tx, block))`, return `rx` as a:
```rust
struct BlockGuard {
    inner: oneshot::Receiver<()>,
}
impl BlockGuard {
   async fn wait_until_satisfied(self) -> Result<(), RecvError> {
        self.inner.await
    }
}
```

Then, in the `MockServer` you can pop off the `(tx, block)` pair and in the request handler perform `tx.send(())`.

crates/astria-sequencer-relayer/tests/blackbox/helper.rs
```diff
@@ -78,6 +99,31 @@ const PRIVATE_VALIDATOR_KEY: &str = r#"
 }
 "#;
 
+pub struct MockSequencerServer {
+    blocks: Arc<Mutex<VecDeque<SequencerBlock>>>,
```
SuperFluffy, line 103: Could this just be a `tokio::sync::mpsc::UnboundedReceiver` instead of this relatively complex `Arc<Mutex<VecDeque<>>>`?

crates/astria-sequencer-relayer/local.env.example
```diff
@@ -19,10 +19,14 @@ ASTRIA_SEQUENCER_RELAYER_PRETTY_PRINT=false
...
-# Address of sequencer/cometbft/tendermint to request new blocks.
+# Address of cometbft/tendermint to request new block heights.
 # 127.0.0.1:26657 is the default socket address at which cometbft
 # serves RPCs.
-ASTRIA_SEQUENCER_RELAYER_SEQUENCER_ENDPOINT="http://127.0.0.1:26657"
+ASTRIA_SEQUENCER_RELAYER_COMETBFT_ENDPOINT="http://127.0.0.1:26657"
+
+# Address of sequencer applcation gRPC service for requesting new blocks.
+# 127.0.0.1:8080 is the default socket address for its gRPC server.
+ASTRIA_SEQUENCER_RELAYER_SEQUENCER_ENDPOINT="http://127.0.0.1:8080"
```
joroshiba, line 29: Can we call this like `ASTRIA_SEQUENCER_RELAYER_SEQUENCER_GRPC_ENDPOINT`? Avoid name collision with previous config when different expected value

crates/astria-sequencer-relayer/src/main.rs
noot: yes, removed!

crates/astria-sequencer-relayer/tests/blackbox/helper.rs
noot: tried that initially but it requires `&mut self` in the grpc methods, which doesn't work as the trait requires `&self`

crates/astria-sequencer-relayer/tests/blackbox/main.rs
noot: @SuperFluffy i updated this to mount a bad block again, should be good now

crates/astria-sequencer-relayer/tests/blackbox/main.rs
noot: re-added block guards

crates/astria-sequencer-relayer/local.env.example
noot: done

crates/astria-sequencer-relayer/tests/blackbox/helper.rs
```diff
@@ -232,7 +331,28 @@ impl TestSequencerRelayerConfig {
...
+            grpc_server
+                .serve_with_incoming(tokio_stream::wrappers::TcpListenerStream::new(
+                    grpc_listener,
+                ))
+                .await
+                .unwrap();
+        });
+
+        // wait for server to start; TODO don't use sleep?
+        tokio::time::sleep(tokio::time::Duration::from_millis(100)).await;
```
SuperFluffy, line 355: Is this actually necessary? The server is not immediately ready to use? What if in addition you constructed everything outside of the task and moved it in?
```rust
let sequencer = {
    let serve = grpc_server.serve_with_incoming(TcpListenerStream::new(grpc_listener));
    tokio::task::spawn(async move { serve.await.unwrap() });
};
```

crates/astria-sequencer-relayer/local.env.example
```diff
@@ -19,10 +19,14 @@ ASTRIA_SEQUENCER_RELAYER_PRETTY_PRINT=false
 # `ASTRIA_SEQUENCER_RELAYER_PRETTY_PRINT` is set to `true`.
 NO_COLOR=
 
-# Address of sequencer/cometbft/tendermint to request new blocks.
+# Address of cometbft/tendermint to request new block heights.
 # 127.0.0.1:26657 is the default socket address at which cometbft
 # serves RPCs.
-ASTRIA_SEQUENCER_RELAYER_SEQUENCER_ENDPOINT="http://127.0.0.1:26657"
+ASTRIA_SEQUENCER_RELAYER_COMETBFT_ENDPOINT="http://127.0.0.1:26657"
```
SuperFluffy, line 25: If we do this for sequencer, should we also explicitly call this `ASTRIA_SEQUENCER_RELAYER_COMETBFT_HTTP_ENDPOINT`?

crates/astria-sequencer-relayer/tests/blackbox/main.rs
```diff
@@ -61,7 +64,9 @@ async fn report_degraded_if_block_fetch_fails() {
         ),
     )
     .await
-    .expect("requesting abci info and block must have occured");
+    .expect("requesting abci info and block must have occured")
+    .1
```
SuperFluffy, line 68: Yeah, that's weird but OK right now. We should fix this (I added a comment on `wait_until_satisifed()`)

crates/astria-sequencer-relayer/tests/blackbox/helper.rs
```diff
@@ -78,6 +101,44 @@ const PRIVATE_VALIDATOR_KEY: &str = r#"
 }
 "#;
 
+pub struct BlockGuard {
+    inner: oneshot::Receiver<()>,
+}
+
+impl BlockGuard {
+    pub async fn wait_until_satisfied(self) -> Result<(), tokio::sync::oneshot::error::RecvError> {
```
SuperFluffy, line 109: Can you add a note that we should harmonize this and the abci mock guard so that we don't have to do the `.1.unwrap()` stuff when using this in a test?

crates/astria-sequencer-relayer/tests/blackbox/helper.rs
```diff
@@ -106,6 +173,12 @@ pub struct TestSequencerRelayer {
 }
 
 impl TestSequencerRelayer {
+    pub fn shutdown(&self) {
```
SuperFluffy, line 176: I think the aborts on the tasks are implied when they are droppoed.

Did you impl this because the server misbehaved otherwise?

Because this is a pretty harsh "shutdown" (more like "plug power") anyway, I wonder if this could just be moved into a drop impl so you don't have to explicitly `TestSequencerRelayer::shutdown` at the end of a test:

```rust
impl Drop for TestSequencerRelayer {
    fn drop(&mut self) {
        self.sequencer_relayer.abort();
        self.sequencer.abort();
        self.celestia.server_handle.stop().unwrap();
    }
}
```

crates/astria-sequencer-relayer/tests/blackbox/main.rs
```diff
@@ -77,6 +82,8 @@ async fn report_degraded_if_block_fetch_fails() {
         expected: json!({"status": "degraded"}),
         actual: readyz.json::<serde_json::Value>().await.unwrap(),
     );
+
+    sequencer_relayer.shutdown();
```
SuperFluffy, line 86: See my comment on replacing `shutdown` with  a `Drop` impl on the mock server so you can get rid of the calls here.

crates/astria-sequencer-relayer/tests/blackbox/main.rs
```diff
@@ -246,12 +260,13 @@ async fn block_from_other_proposer_is_skipped() {
                 break;
             }
         }
+        sequencer_relayer.shutdown();
```
SuperFluffy, line 263: Why is the shutdown happening here? Because the object is moved into the block?

Yeah, that's ugly. It's not the best fix, but can you just hand back the object instead and call `shutdown` outside? That's marginally better. I mean this:

```rust
let observe_blobs = async move {
    // todo!("do things with sequencer_relayer")
   (sequencer_relayer, observe_blobs)
};

Then you can get the object back into the outer scope.

SuperFluffy (approved):
Just a few comments

joroshiba (approved):
clippy failing but otherwise LGTM

crates/astria-sequencer-relayer/tests/blackbox/main.rs
noot: removed in favour of `Drop`

PR #811, opened 2024-03-12
feat(conductor): emit more information about blocks received from Sequencer, Celestia

SuperFluffy:
## Summary
Emits more information on the sequencer blocks received from Sequencer, and blocks reconstructed from Celestia blobs.

## Background
We have very little insight into how many blocks we receive from Sequencer and reconstructed from Celestia.

## Changes
- Add various newtype-wrappers implementing `serde::Serialize` for blocks reconstructed from Celestia blobs, and for blocks received from Sequencer
- Use these reporting-wrappers to emit more information about the received data, like number of transactions, heights, rollup IDs, et 

## Testing
Should be checked in observability.


crates/astria-conductor/src/celestia/reporting.rs
```diff
@@ -0,0 +1,87 @@
...
+    {
+        const FIELDS: [&str; 4] = [
+            "celestia_height",
+            "sequencer_namespace",
+            "rollup_namespace",
+            "reconstructed_blocks",
+        ];
+        let mut state = serializer.serialize_struct("ReconstructedBlocksInfo", FIELDS.len())?;
+        state.serialize_field(FIELDS[0], &self.0.celestia_height)?;
+        state.serialize_field(FIELDS[1], &hex(&self.0.sequencer_namespace.as_bytes()))?;
```
SuperFluffy, line 47: namespaces are hex-formatted bytes here (as opposed to base64). This seems to be in-line with block explorers like celenium.

crates/astria-telemetry/src/display.rs
```diff
@@ -37,6 +38,7 @@ pub fn hex<T: AsRef<[u8]> + ?Sized>(bytes: &T) -> Hex<'_> {
 /// A newtype wrapper of a byte slice that implements [`std::fmt::Display`].
 ///
 /// To be used in tracing contexts. See the [`self::hex`] utility.
+#[derive(SerializeDisplay)]
```
SuperFluffy, line 41: Introducing `serde_with` into the dependency tree because this avoids extra allocations when serializing these.

crates/astria-conductor/src/celestia/mod.rs
```diff
@@ -300,14 +316,8 @@ impl Reader {
                     }
                 }
 
-                Some((celestia_height, blocks)) = block_stream.next() => {
-                    debug!(
```
SuperFluffy, line 304: moved the event construction out of this select-loop into the function that actually performs the fetch.

It's better practice to emit events in their correct scopes (instead of glue, like here)

crates/astria-conductor/src/celestia/mod.rs
```diff
@@ -549,7 +563,17 @@ async fn fetch_blocks_at_celestia_height(
...
         .collect()
         .await;
-    Ok(blocks)
+    let reconstructed = ReconstructedBlocks {
+        celestia_height,
+        sequencer_namespace,
+        rollup_namespace,
+        blocks,
+    };
+    info!(
```
SuperFluffy, line 572: Emitting this at info level because this strikes me as the main functionality of conductor (receiving new blocks).

Or should the level be stepped down?

crates/astria-conductor/src/sequencer/reporting.rs
```diff
@@ -0,0 +1,41 @@
...
+        S: serde::Serializer,
+    {
+        let mut state = serializer.serialize_struct("SequencerBlockInfo", 2)?;
+        state.serialize_field("sequencer_height", &self.0.height().value())?;
+        state.serialize_field("rollups", &ReportRollups(self.0.rollup_transactions()))?;
+        state.end()
+    }
+}
+
+struct ReportRollups<'a>(&'a IndexMap<RollupId, RollupTransactions>);
```
SuperFluffy, line 28: This reports the each rollup ID and the number of transactions per rollup. With the move to the sequencer API this will be reduced.

It could also report hashes of the actual opaque bytes. Maybe at a lower debug event level?

crates/astria-conductor/src/celestia/reporting.rs
```diff
@@ -0,0 +1,87 @@
...
+        const FIELDS: [&str; 4] = [
+            "celestia_height",
+            "block_hash",
+            "number_of_transactions",
+            "from_celestia_height",
+        ];
+        let mut state = serializer.serialize_struct("ReconstructedBlockInfo", FIELDS.len())?;
+        state.serialize_field(FIELDS[0], &self.0.celestia_height)?;
+        state.serialize_field(FIELDS[1], &hex(&self.0.block_hash))?;
+        state.serialize_field(FIELDS[2], &self.0.transactions.len())?;
```
SuperFluffy, line 83: We could also report the hashes of all the transactions. If we are just reporting for an individual transactions this might not be so bad.

crates/astria-conductor/src/celestia/mod.rs
```diff
@@ -582,6 +607,11 @@ async fn process_sequencer_blob(
         )
         .await
         .wrap_err("failed fetching rollup blobs from celestia")?;
+    info!(
```
joroshiba, line 610: Maybe debug level here with the info log being the full reconstructed?

joroshiba (approved):

joroshiba (approved):

PR #813, opened 2024-03-13
chore(composer): flatten searcher to top level composer

itamarreif:
## Summary
This removes the `searcher` module from `composer`, moving all the logic that was previously in the `Searcher` actor into the `Composer` top-level actor.

## Background
In order to implement the gRPC collector (#795) we needed to move around some of the parts of `Searcher` into `Composer` (#796). The plan was to do this piecemeal, and afterwards move the rest of `Searcher` into `Composer`, but after looking into it, seems like removing the `Searcher` right away was simple enough.

## Changes
- Moved the fields and associated constructor logic from `Searcher` up into `Composer`
- Moved the logic for spawning, waiting for, and handling tasks for the `Executor` and `Collector`s up into the `Composer`'s `run_until_stopped()`.
- Moved some of the helper methods into the `composer` module and converted them from methods to static functions.
- Fixed all the places where the word Searcher was used to replace it with Composer
- Moved the `executor`, `collector`, `rollup` modules up to the top-level.

## Testing
No new functionality, all existing tests pass.

## Related Issues
Part of #795

closes #796


itamarreif:
closing this in favor of @bharath-123's work in #795

Issue #789, opened 2024-03-06
conductor: use `GetFilteredSequencerBlock` instead of tendermint API


joroshiba:
We added the new Sequencer API, as a grpc service which can be used to get the blocks directly from a trusted sequencer app instead of needing to transpose `SequencerBlocks` from CometBFT blocks.

This is required for us to add new bridging txs into `SequencerBlocks`

PR #815, opened 2024-03-13
feat(conductor)!: use Sequencer gRPC API to fetch soft bocks.

SuperFluffy:
## Summary
Uses the new gRPC based Sequencer API to fetch soft blocks from Sequencer.

## Background
https://github.com/astriaorg/astria/pull/701 introduced a gRPC based API to fetch block information from sequencer for just the rollup IDs that are desired (compared to fetching the full block + a filter afterwards to get the desired data).

This patch removes the websocket based fetch in favor of this simpler API.

## Changes
- replace the websocket-based sequencer block fetch by the gRPC based filtered block fetch
- remove the websocket client pool (it was very complex)
- use the cometbft http based RPC to fetch the most recent sequencer heights (similar to what sequencer-relayer is doing) 

## Testing
All tests were updated to work with the newest changes. This patch must be tested end-to-end.

## Breaking Changelist
New environment variables were introduced:
- `ASTRIA_CONDUCTOR_SEQUENCER_COMETBFT_URL`: a String containing the URL of the cometbft HTTP RPC endpoint
- `ASTRIA_CONDUCTOR_SEQUENCER_GRPC_URL`: a String containing URL of the sequencer gRPC endpoint
- `ASTRIA_CONDUCTOR_SEQUENCER_BLOCK_TIME_MS`: an integer containing the duration between subsequent cometbft ABCI requests to figure out the latest sequencer height

An environment variable was removed:
- `ASTRIA_CONDUCTOR_SEQUENCER_URL`

## Related Issues
Closes https://github.com/astriaorg/astria/issues/789


crates/astria-conductor/src/sequencer/block_stream.rs
SuperFluffy, line 1: This module was moved from `sequencer/mod.rs` but also simplified.

crates/astria-sequencer-client/src/lib.rs
SuperFluffy, line 1: Leaving this untouched in this PR, but I believe most of the functionality in here can now go.

crates/astria-conductor/local.env.example
```diff
@@ -46,11 +46,19 @@ ASTRIA_CONDUCTOR_PRETTY_PRINT=false
...
+# The URL of a fully trusted CometBFT/Sequencer node to get the latest height of
+# the Sequencer network, and to get the validators for a given Sequencer height
+# to verify Sequencer data read off Celestia.
 # 127.0.0.1:26657 is the default socket address in comebft's `rpc.laddr` setting.
-ASTRIA_CONDUCTOR_SEQUENCER_URL="ws://127.0.0.1:26657/websocket"
+ASTRIA_CONDUCTOR_SEQUENCER_COMETBFT_URL="http://127.0.0.1:26657"
+
+# The duration in milliseconds that sequencer-relayer sleeps before
+# requesting the latest block height from sequencer.
+ASTRIA_CONDUCTOR_SEQUENCER_BLOCK_TIME_MS=1000
```
joroshiba, line 61: This is about half of the best case scendario block time is that a good ratio or should this be 2000?

crates/astria-conductor/local.env.example
```diff
@@ -46,11 +46,19 @@ ASTRIA_CONDUCTOR_PRETTY_PRINT=false
...
+ASTRIA_CONDUCTOR_SEQUENCER_GRPC_URL="http://127.0.0.1:8080"
+
+# The URL of a fully trusted CometBFT/Sequencer node to get the latest height of
+# the Sequencer network, and to get the validators for a given Sequencer height
+# to verify Sequencer data read off Celestia.
 # 127.0.0.1:26657 is the default socket address in comebft's `rpc.laddr` setting.
-ASTRIA_CONDUCTOR_SEQUENCER_URL="ws://127.0.0.1:26657/websocket"
+ASTRIA_CONDUCTOR_SEQUENCER_COMETBFT_URL="http://127.0.0.1:26657"
+
+# The duration in milliseconds that sequencer-relayer sleeps before
```
joroshiba, line 59:
```suggestion
# The duration in milliseconds that conductor sleeps before
```

joroshiba (commented):
In general looks good, couple nits and questions.

Noting we were previously verifying the data we generated against the data hash in cometbft header and are no longer doing that. Sequencer is fully trusted, verifying against it's own data doesn't make a ton of sense so I think this is fine.

crates/astria-conductor/local.env.example
SuperFluffy: 2000 is our default right? I have changed it to that number and have noted that that's the default.

crates/astria-conductor/src/executor/tests.rs
```diff
@@ -302,7 +306,9 @@ async fn soft_blocks_at_expected_heights_are_executed() {
 
     let mut block = make_cometbft_block();
     block.header.height = SequencerHeight::from(100u32);
-    let block = SequencerBlock::try_from_cometbft(block).unwrap();
+    let block = SequencerBlock::try_from_cometbft(block)
```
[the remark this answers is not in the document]
SuperFluffy: Fair enough, but we need to create a more convenient way to construct valid sequencer blocks for testing. We should do that in a follow up.

crates/astria-conductor/src/sequencer/mod.rs
```diff
@@ -109,10 +104,14 @@ impl Reader {
...
         let mut sequential_blocks = BlockCache::with_next_height(next_expected_height)
             .wrap_err("failed constructing sequential block cache")?;
-        let mut blocks_from_heights =
-            BlocksFromHeightStream::new(next_expected_height, latest_height, pool.clone(), 20);
+        let mut blocks_from_heights = block_stream::BlocksFromHeightStream::new(
+            executor.rollup_id(),
+            next_expected_height,
+            latest_height,
+            sequencer_grpc_client.clone(),
+            20,
```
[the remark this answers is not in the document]
SuperFluffy: Moved the magic number away from here and into the constructor. Added a note there.

joroshiba (approved):

PR #816, opened 2024-03-13
feat(conductor): sequencer api supports rollupdata type

joroshiba:
## Summary
Changes required to make the Sequencer API utilize the new `RollupData` type instead of raw bytes.

## Background
In order to update all rollups after recent change to serialization format, they would need to import and decode tx bytes into RollupData shape. This moves that into the execution API.

## Changes
- Update protobuf for execution api
- Update buf build process to utilize sequencerapis as dependency in executionapis
- Modify conductor to parse bytes into `RollupData` before calling execution api.

## Testing
Updated automated tests.


crates/astria-conductor/src/executor/client.rs
```diff
@@ -68,7 +74,18 @@ impl Client {
         transactions: Vec<Vec<u8>>,
         timestamp: Timestamp,
     ) -> eyre::Result<Block> {
-        let transactions = transactions.into_iter().map(Bytes::from).collect();
+        use prost::Message as _;
+        let transactions = transactions
+            .into_iter()
+            .filter_map(|tx| match RollupData::decode(tx.as_slice()) {
```
joroshiba, line 80: Should this actually throw an error instead of skipping?

noot: yeah, failing to decode here means there is an issue creating the `SequencerBlock` type in the sequencer

crates/astria-conductor/src/executor/client.rs
```diff
@@ -68,7 +71,14 @@ impl Client {
         transactions: Vec<Vec<u8>>,
         timestamp: Timestamp,
     ) -> eyre::Result<Block> {
-        let transactions = transactions.into_iter().map(Bytes::from).collect();
+        use prost::Message;
+
+        let transactions = transactions
+            .into_iter()
+            .map(|tx| RollupData::decode(tx.as_slice()))
```
SuperFluffy, line 78: Now that I see this bit I am wondering if making `RollupData` an opaque field inside `RollupTransactions` was the right call.

There is now a lot of wrapping/unwrapping going on and I am not quite sure if that's good.

crates/astria-conductor/src/executor/client.rs
```diff
@@ -68,7 +71,14 @@ impl Client {
...
+            .collect::<Result<_, _>>()
+            .wrap_err("failed to decode tx bytes as RollupData")?;
```
SuperFluffy, line 80: just calling out that this will eventually halt conductor

crates/astria-conductor/src/executor/tests.rs
```diff
@@ -261,16 +268,21 @@ async fn start_mock() -> MockEnvironment {
     }
 }
 
+fn test_transaction(data: &str) -> RawRollupData {
```
SuperFluffy, line 271: Can you prefix this with a verb? Just `make_transaction()` or `make_rollup_data()`

SuperFluffy (approved):
The change looks simple enough. Just a few comments and request to change a name.

crates/astria-conductor/src/executor/client.rs
noot: yeah, my intention was that the rollup node decodes the bytes inside `RollupTransactions`, and we consider conductor part of a rollup node, so this is intended. the decoding will happen at one place or another, if we don't put it here, it will be when conductor pulls the sequencer block from sequencer/DA

noot (approved):

crates/astria-conductor/src/executor/client.rs
joroshiba: Yep this is expected, if we have data that can't be deserialized it's an issue that we shouldnt skip over and just execute.

PR #817, opened 2024-03-13
chore: migrate v1alpha1 sequencer apis to v1

joroshiba:
## Summary
Now that we have entered audit over things that use these core types, these should be considered v1 and breaking changes cannot be allowed.


noot (approved):
approved, but want to note I'll likely be changing the protos in the upgrade to cometbft v0.38 if we decide to not have `SequencerBlock` contain the entire cometbft block header.

joroshiba:
> approved, but want to note I'll likely be changing the protos in the upgrade to cometbft v0.38 if we decide to not have `SequencerBlock` contain the entire cometbft block header.

We can do this with a v2 change instead of making the breaking change as well

PR #819, opened 2024-03-14
fix(sequencer): filtered blocks success when no data expected

joroshiba:
## Summary
Updates `GetFilteredSequencerBlock` to only return an error if the rollup should have data in storage and doesn't as opposed to error each time data in storage.

## Background
Conductor calls `GetFilteredSeqeuncerBlock` for every height and does not know if rollup contains data, should get a response even if the rollup block has no data not an error message.

## Changes
- Filter out ids to those expected to have data before fetching rollup data.

## Testing
Manually tested with dev-cluster


crates/astria-sequencer/src/grpc/sequencer.rs
```diff
@@ -125,9 +125,13 @@ impl SequencerService for SequencerServer {
             .into_iter()
             .map(RollupId::to_vec)
             .collect::<Vec<_>>();
+        let rollup_ids_with_data: Vec<RollupId> = rollup_ids
+            .into_iter()
+            .filter(|id| all_rollup_ids.contains(&id.to_vec()))
```
SuperFluffy, line 130: This is pretty expensive because with `O(n)` behaviour for every single check. Also, unnecessary heap allocations everywhere. I would suggest doing the following instead:
```rust
let all_rollup_ids = snapshot
            .get_rollup_ids_by_block_hash(&block_hash)
            .await
            .map_err(|e| Status::internal(format!("failed to get rollup ids from storage: {e}")))?
            .into_iter()
            .collect::<HashSet<_>>();
for rollup_id in rollup_ids {
    if all_rollup_ids.contains(&rollup_id) {
        // perform the lookup, push to `rollup_transactions`
    }
}
let all_rollup_transactions = all_rollup_transactions.into_iter().map(RollupId::to_vec).collect();
```

crates/astria-sequencer/src/grpc/sequencer.rs
```diff
@@ -118,14 +118,21 @@ impl SequencerService for SequencerServer {
...
             .into_iter()
             .map(RollupId::to_vec)
             .collect::<Vec<_>>();
+        all_rollup_ids.sort_unstable();
 
+        // Filter out the Rollup Ids requested which have no data before grabbing
+        // so as to not error because the block had no data for the requested rollup
+        let rollup_ids: Vec<RollupId> = rollup_ids
+            .into_iter()
+            .filter(|id| all_rollup_ids.binary_search(&id.to_vec()).is_ok())
```
SuperFluffy, line 134: I prefer to not turn all rollup IDs into vecs prematurely, but even so this allocation could still be avoided by using [`slice::binary_search_by`](https://doc.rust-lang.org/std/primitive.slice.html#method.binary_search_by): `.filter(|id| all_rollup_ids.binay_search_by(|probe| probe.as_slice().cmp(id.as_ref()))`

crates/astria-sequencer/src/grpc/sequencer.rs
```diff
@@ -118,14 +118,21 @@ impl SequencerService for SequencerServer {
                 ))
             })?;
 
-        let all_rollup_ids = snapshot
+        let mut all_rollup_ids = snapshot
             .get_rollup_ids_by_block_hash(&block_hash)
             .await
             .map_err(|e| Status::internal(format!("failed to get rollup ids from storage: {e}")))?
             .into_iter()
             .map(RollupId::to_vec)
```
SuperFluffy, line 126: IMO this should be done at the end of the method, when the final return value is constructed. You can avoid the extra allocation when looking up the IDs, and it's neater to read.

SuperFluffy (approved):
Approving to not bikeshed on this unduly, but the transformation of `all_rollup_ids` should be moved to the very end.

crates/astria-sequencer/src/grpc/sequencer.rs
```diff
@@ -84,15 +84,16 @@ impl SequencerService for SequencerServer {
...
-        let mut rollup_ids: Vec<RollupId> = vec![];
-        for id in request.rollup_ids {
-            let Ok(rollup_id) = RollupId::try_from_vec(id) else {
-                return Err(Status::invalid_argument(
-                    "invalid rollup ID; must be 32 bytes",
-                ));
-            };
-            rollup_ids.push(rollup_id);
-        }
+        let Ok(rollup_ids) = request
```
joroshiba, line 87: note this is secondary, but updated to match the style of the rest of the code here closer.

PR #821, opened 2024-03-14
fix(conductor): serialize rollup IDs as strings so telemetry doesn't crash

SuperFluffy:
## Summary
Ensures that rollup IDs are serialized as strings.

## Background
The serde derivation for `RollupId` was missing a `serde(transparent)` attribute. This means that, for JSON, rollup IDs were not serialized has `"<hex-of-32-bytes>"` but as `{ "inner": "<hex-of-32-bytes>" }`.

This in turn caused the JSON-serialization of some structs used for reporting received blocks from sequencer to fail, causing a panic in the tracing stack (as it assumes that `Display` formatting won't fail).

## Changes
- Add `serde(transparent)` attribute on `RollupId`
- Add snapshot tests to reporting types

## Testing
Added snapshot tests for types used in conductor tracing (that's where we saw the failed display formatting and subsequent panic).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -601,6 +601,7 @@ dependencies = [
  "hex",
  "humantime",
  "indexmap 2.2.3",
+ "insta",
  "itoa",
  "jsonrpsee",
  "pin-project-lite",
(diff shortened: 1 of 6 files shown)


joroshiba (approved):
love the snapshot tests

PR #822, opened 2024-03-15
refactor(relayer): replace grab-bag constructor with config-like builder

SuperFluffy:
## Summary
Explicitly populates a builder and constructs the relayer actor from it instead of passing in the config.

## Background
Using a config as a grab-bag to initialize types is bad practice because it's not immediately clear which option are relevant to which part of the stack.

This patch adds a `relayer::Builder` type with crate-public fields which are explicitly populated and then used to construct the final `Relayer` object. This fulfills a similar need as the type-state builder pattern used in conductor, but needs significantly less boilerplate. And because there are no extra invariants that are supposed to be held up by the builder-type itself this is all around nicer.

## Changes
- Add a `relayer::Builder` type
- Replace `Relayer::new` with `relayer::Builder::build`
- Remove unnecessary sources of `async` so that `relayer::Builder::build` need not be `await`ed on:
    - instead of `celestia_rpc::Client` use `jsonrpsee::HttpClient`
    - use `tonic::Endpoint::connect_lazy` to delay connecting until the first attempt

## Testing
+ Most changes were shuffling code around and tests still pass.
+ Block fetching over gRPC still works, which means lazy client initialization still works.
+ Added an Authorization middleware to the mocked celestia server to ensure the changes to Celestia construction still work.


crates/astria-sequencer-relayer/src/relayer/builder.rs
```diff
@@ -0,0 +1,96 @@
...
+    };
+    let mut headers = HeaderMap::new();
+    let auth_value = format!("Bearer {bearer_token}").parse().wrap_err(
+        "failed to construct Authorization header value from provided Celestia bearer token",
+    )?;
+    headers.insert(http::header::AUTHORIZATION, auth_value);
+    let client = HttpClientBuilder::default()
+        .set_headers(headers)
+        .build(endpoint)
+        .wrap_err("failed constructing Celestia JSONRPC HTTP Client")?;
```
joroshiba, line 93: Why construct this way vs previous method of using `celestia_client::celestia_rpc::Client::new`?

crates/astria-sequencer-client/src/lib.rs
```diff
@@ -54,6 +54,7 @@ pub trait StreamLatestHeight {
 #[cfg(feature = "http")]
 impl StreamLatestHeight for HttpClient {
     fn stream_latest_height(&self, poll_period: Duration) -> LatestHeightStream {
+        println!("client: {self:?}");
```
joroshiba, line 57:
```suggestion
```

crates/astria-sequencer-relayer/src/relayer/builder.rs
```diff
@@ -0,0 +1,96 @@
...
+    println!("{client:?}");
```
joroshiba, line 94:
```suggestion
```

joroshiba (approved):
approved assuming removal of printlns done. Only one open question

crates/astria-sequencer-relayer/src/relayer/builder.rs
SuperFluffy: The `celestia_rpc` impl forces you into `async` because they combine both websocket and http into one (where the http client builder is not async).

I didn't like this constructor being async (hence the lazy grpc connection), and I wanted to get rid of the ugly uwrapping we had before.

I wish they had just re-exported jsonrpsee since they are exporting a lot if its items anyhow.

PR #824, opened 2024-03-16
chore: releases for new devnet

joroshiba:
## Summary
Cut new core binaries as needed for new devnet release. 

## Background
We have updated core protobufs, migrated to using sequencer api for block information. This cuts new breaking changes for all the core binaries.


crates/astria-composer/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "astria-composer"
-version = "0.4.0"
+version = "0.5.0"
```
SuperFluffy, line 3: Did we have a breaking change here?

Answer: yeah, we had breaking changes in the protobufs, composer uses the protobufs in its public interfaces, hence this needs to bump the protobufs.

SuperFluffy (approved):

PR #827, opened 2024-03-18
feat(composer): Implement graceful shutdown

bharath-123:
## Summary
Implement graceful shutdown for composer tasks.

## Background
Currently, composer tasks do not support any sort of graceful shutdown on a SIGTERM or so. Also, when certain tasks error out, there is no graceful way of exiting all the composer tasks.

## Changes
- Add a field to `composer` called `composer_shutdown_channels` which consists of the shutdown channels of the ApiServer, gRPC collector and Executor.
- Add a field to `composer` called `geth_collector_shutdown_channels` which consists of the shutdown channels of each individual collector.
- Implement a `shutdown` function which is called after a task exits due to an error or when a SIGTERM is called.

## Testing
Builds and tests run successfully. Need to test it by deploying it. 

## Related Issues
https://github.com/astriaorg/astria/issues/795

closes

### DIFF
--- a/crates/astria-composer/Cargo.toml
+++ b/crates/astria-composer/Cargo.toml
@@ -44,6 +44,7 @@ tokio = { workspace = true, features = [
   "rt-multi-thread",
   "sync",
   "time",
+  "signal",
 ] }
 tokio-util = { workspace = true, features = ["rt"] }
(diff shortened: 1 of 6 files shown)


bharath-123:
I will re-work shutdowns in a new PR.

PR #829, opened 2024-03-18
refactor(conductor): simplify builder types by config-like

SuperFluffy:
## Summary
Replaces type-state builders by simpler config-like builders.

## Background
The type-state builder patterns we use in conductor contains a lot of boiler plate for arguably little gain because the objects they construct are only created once at app initialization. Worse, they are painful to extend and confusing to future maintainers.

Using config-like builders with crate-public fields gives us similar upsides (all fields are required to be set, it's obvious where fields from `Config` end up, the `build` method checks that the correct final type is constructed) without the downsides (so much boiler plate).

## Changes
- Replace executor and conductor type-state builders by builders with crate-private fields
- Replace multi-argument sequencer-reader constructor by the same config-like builder

## Testing
Minimal impact on the final program. End-to-end test should be sufficient.


noot (approved):
nice!

Issue #679, opened 2024-01-09
sequencer: update to cometbft 0.38, utilize FinalizeBlock, cache execution.


joroshiba:
`FinalizeBlock` will give us much more control over tx execution flows. 

In order to remove edge cases around failed proposal rounds we removed the caching of executed txs in this PR: https://github.com/astriaorg/astria/pull/675

We could potentially work around these edge cases other ways, but it's cleaner to simply double execute with cometbft 0.37, because in `BeginBlock` we don't have information about the txs which will be executed. We only get them piece by piece through each DeliverTX calls. Migrating to 0.38 allows us to get them all at once and have more control over the execution of the txs. We can verify that all txs were executed already, validate by validating a tx root, or clear state then execute.

PR #830, opened 2024-03-18
refactor(proto)!: update `SequencerBlockHeader` and related proto types to not use cometbft header

noot:
## Summary
update `SequencerBlockHeader`, `SequencerBlock`, `FilteredSequencerBlock`, and `CelestiaSequencerBlob` to not contain an entire cometbft header, but only to contain relevant fields required. 

## Background
this is required for the update to ABCI v0.38, as `finalize_block` no longer passes in the entire block header (like `begin_block` did). we keep needed fields like height, chain_id, timestamp, proposer_address, data_hash in the `SequencerBlockHeader`.

## Changes
- update `SequencerBlockHeader` to not contain an entire cometbft header but have height, chain_id, timestamp, proposer_address, data_hash - all fields are now explicitly needed by other components of the stack
- update `SequencerBlock`, `FilteredSequencerBlock`, and `CelestiaSequencerBlob` to have `block_hash` as this can no longer be calculated from the cometbft header as it doesn't exist anymore
- deprecate `rollup_ids_root` in `SequencerBlockHeader` as it can be calculated by the list of rollup IDs in a block

## Testing
unit tests

## Breaking Changelist
- the `SequencerBlockHeader`, `SequencerBlock`, `FilteredSequencerBlock`, and `CelestiaSequencerBlob` are changed.

## Related Issues

 related to  #679


crates/astria-core/src/sequencer/v1/block/mod.rs
```diff
@@ -327,25 +337,41 @@ enum SequencerBlockErrorKind {
...
 }
 
 #[derive(Debug, Clone, PartialEq)]
 pub struct SequencerBlockHeader {
     // the cometbft header for this sequencer block
-    cometbft_header: tendermint::block::header::Header,
+    // cometbft_header: tendermint::block::header::Header,
+    chain_id: tendermint::chain::Id,
+    height: tendermint::block::Height,
+    time: Time,
```
SuperFluffy, line 354: If we redo these, can't we get rid of tendermint here for good? I am looking at the changed `SequencerBlockHeader` in the redefined protos, and these don't show up at the tendermint level anymore? Would much rather use native protos (or well known types).

SuperFluffy (commented):
I think making changes on top of the `v1` sequencer API might make maintaining this a bit of a nightmare if we need extra changes on `main`.

Given that we consider changes to the sequencer proto specs breaking (as in, breaking at the network level even though spec evolution is not breaking for protobus), I think this warrants calling this a `v2alpha1`. It will also make it much easier to merge this back into main at a later point.

crates/astria-sequencer-relayer/postsubmit.json
```diff
@@ -0,0 +1 @@
+{"state":"fresh"}
```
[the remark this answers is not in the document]
noot: woops

crates/astria-conductor/src/conductor.rs
```diff
@@ -217,7 +217,9 @@ async fn get_sequencer_namespace(client: HttpClient) -> eyre::Result<Namespace>
         .await
         .wrap_err("failed to get block from sequencer after 10 attempts")?;
 
-    Ok(celestia_client::celestia_namespace_v0_from_cometbft_header(
-        block.header().cometbft_header(),
-    ))
+    Ok(
+        celestia_client::celestia_namespace_v0_from_cometbft_chain_id(
+            block.header().chain_id().as_str(),
```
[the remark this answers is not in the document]
noot: it only needs the chain id field though, so that seems worse for code understandability to me

crates/astria-core/src/sequencer/v1/block/mod.rs
noot: yeah we can - i left them as tendermint as they're originating from a tendermint chain, but can change to our own types also

noot:
@SuperFluffy moved changed protos (block, celestia, and service) to a v2 - ready for rereview!

crates/astria-core/src/sequencer/mod.rs
```diff
@@ -1 +1,39 @@
 pub mod v1;
+
+/// A trait to convert from raw decoded protobuf types to idiomatic astria types.
+///
+/// The primary use of this trait is to convert to/from foreign types.
+pub trait Protobuf: Sized {
```
noot, line 6: I had to move the Protobuf trait into the crates that were implementing it - the problem is that it was implemented on the external `merkle::Proof` type, however we then had two implementations of it (one in `sequencer` and one in `sequencerblock`) which conflicted as the `Raw` types weren't the same.

crates/astria-conductor/src/lib.rs
```diff
@@ -15,7 +15,7 @@ pub mod conductor;
 pub mod config;
 pub(crate) mod executor;
 pub(crate) mod sequencer;
-pub(crate) mod utils;
+mod utils;
```
joroshiba, line 18: why did this publicity here change?

crates/astria-core/src/generated/astria.sequencer.v2.rs
joroshiba, line 1: I think the v2 generated bits here should be removed?

crates/astria-core/src/lib.rs
```diff
@@ -9,44 +9,83 @@ pub mod generated;
...
-
-    /// Convert to the raw protobuf type, dropping `self`.
-    ///
-    /// This method provides a default implementation in terms of
-    /// [`Self::to_raw`].
-    fn into_raw(self) -> Self::Raw {
-        Self::to_raw(&self)
-    }
-}
+// /// A trait to convert from raw decoded protobuf types to idiomatic astria types.
```
joroshiba, line 17: remove?

crates/astria-core/src/sequencer/v1/block/mod.rs
```diff
@@ -14,8 +14,8 @@ use super::{
     IncorrectRollupIdLength,
     RollupId,
 };
-use crate::{
-    sequencer::v1::{
+use crate::sequencer::{
+    v1::{
```
joroshiba, line 18: nit: unnecessary change, kinda surprised format doesn't dislike it

crates/astria-core/src/sequencer/v1/transaction/action.rs
```diff
@@ -9,8 +9,8 @@ use penumbra_ibc::IbcRelay;
 use penumbra_proto::penumbra::core::component::ibc::v1::FungibleTokenPacketData;
 
 use super::raw;
-use crate::{
-    sequencer::v1::{
+use crate::sequencer::{
+    v1::{
```
joroshiba, line 13: nit: here as well

crates/astria-conductor/src/lib.rs
noot: just didn't need `pub(crate)`

crates/astria-core/src/sequencer/v1/block/mod.rs
noot: what do you mean? this has to be changed because of the protobuf trait change

crates/astria-core/src/sequencerblock/v1alpha1/block.rs
```diff
@@ -0,0 +1,1529 @@
...
+    pub chain_id: tendermint::chain::Id,
+    pub height: tendermint::block::Height,
+    pub time: Time,
+    pub rollup_transactions_root: [u8; 32],
+    pub data_hash: [u8; 32],
+    pub proposer_address: account::Id,
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct SequencerBlockHeader {
```
noot, line 349: this type is updated

crates/astria-core/src/sequencerblock/v1alpha1/block.rs
```diff
@@ -0,0 +1,1529 @@
...
+    pub rollup_transactions: IndexMap<RollupId, RollupTransactions>,
+    pub rollup_transactions_proof: merkle::Proof,
+    pub rollup_ids_proof: merkle::Proof,
+}
+
+/// `SequencerBlock` is constructed from a tendermint/cometbft block by
+/// converting its opaque `data` bytes into sequencer specific types.
+#[derive(Clone, Debug, PartialEq)]
+#[allow(clippy::module_name_repetitions)]
+pub struct SequencerBlock {
```
noot, line 552: this type is updated

crates/astria-core/src/sequencerblock/v1alpha1/block.rs
```diff
@@ -0,0 +1,1529 @@
...
+    pub rollup_transactions_proof: merkle::Proof,
+    // all rollup ids in the sequencer block
+    pub all_rollup_ids: Vec<RollupId>,
+    // proof that `rollup_ids` is included in `data_hash`
+    pub rollup_ids_proof: merkle::Proof,
+}
+
+#[derive(Debug, Clone, PartialEq)]
+#[allow(clippy::module_name_repetitions)]
+pub struct FilteredSequencerBlock {
```
noot, line 1011: this type is updated

crates/astria-core/src/sequencerblock/v1alpha1/celestia.rs
```diff
@@ -0,0 +1,616 @@
...
+            rollup_ids,
+            rollup_transactions_proof,
+            rollup_ids_proof,
+        })
+    }
+}
+
+#[derive(Clone, Debug)]
+#[allow(clippy::module_name_repetitions)]
+pub struct CelestiaSequencerBlob {
```
noot, line 481: this type is updated

crates/astria-core/src/sequencerblock/v1alpha1/mod.rs
```diff
@@ -0,0 +1,124 @@
...
+use crate::{
+    generated::sequencerblock::v1alpha1 as raw,
+    sequencer::v1::{
+        derive_merkle_tree_from_rollup_txs,
+        RollupId,
+    },
+    sequencerblock::Protobuf,
+};
+
+pub(crate) fn are_rollup_ids_included<'a, TRollupIds: 'a>(
```
noot, line 27: this file is copy pasted from v1

crates/astria-celestia-client/src/lib.rs
```diff
@@ -55,10 +55,10 @@ pub const fn celestia_namespace_v0_from_rollup_id(
 }
 
 #[must_use = "a celestia namespace must be used in order to be useful"]
-pub fn celestia_namespace_v0_from_cometbft_header(header: &tendermint::block::Header) -> Namespace {
+pub fn celestia_namespace_v0_from_cometbft_chain_id(chain_id: &str) -> Namespace {
```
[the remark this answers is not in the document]
SuperFluffy: Raising this again: please actually provide an `chain::Id` argument or rename this ti `celestia_namespace_v0_from_str`. Although I think I would prefer just calling this `celestia_namespace_v0_from_bytes` (as done in #946 )

crates/astria-core/src/sequencerblock/mod.rs
```diff
@@ -0,0 +1,39 @@
+pub mod v1alpha1;
+
+/// A trait to convert from raw decoded protobuf types to idiomatic astria types.
+///
+/// The primary use of this trait is to convert to/from foreign types.
+pub trait Protobuf: Sized {
```
SuperFluffy, line 6: Please move this to the top level of `astria-core`

crates/astria-core/src/sequencer/mod.rs
SuperFluffy: Ok, that's not good. We have to find a different solution because that's pretty confusing then. I really don't want to duplicate traits like this.

crates/astria-conductor/src/conductor.rs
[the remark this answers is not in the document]
SuperFluffy: It's actually a pretty common (and preferred) pattern - law of demeter. You need not know that `block` has a `header` that in turn provides a `chain_id`. It's enough that `block` gives you a chain ID - where it comes from is an implementation detail.

crates/astria-core/src/sequencerblock/v1alpha1/mod.rs
SuperFluffy: Please move this to `lib.rs` or into a utilty module (same for the others) so that we avoid unnecessary code duplication.

crates/astria-core/src/sequencerblock/v1alpha1/mod.rs
```diff
@@ -0,0 +1,124 @@
...
+        .audit()
+        .with_root(root)
+        .with_leaf_builder()
+        .write(id.as_ref())
+        .write(&merkle::Tree::from_leaves(rollup_transactions.transactions()).root())
+        .finish_leaf()
+        .perform()
+}
+
+impl Protobuf for merkle::Proof {
```
SuperFluffy, line 69: This is really not ideal. We should either modify the `Protobuf` trait (which I don't like), define newtype wrappers, or have a common `raw::Proof` type.

SuperFluffy (changes requested):
This introduces a lot of (IMO unnecessary) duplicated code. Especially the fact that we have duplicate definintions for `merkle::Proof` and duplicate definitions of the `Protobuf` trait points to the API design not being sound.

SuperFluffy (approved):
Approving to unblock the move to abci 0.38 and with a nod to #958 that will address my concerns

PR #831, opened 2024-03-18
feat(sequencer)!: update to ABCI v0.38

### LINKED ISSUE #679: sequencer: update to cometbft 0.38, utilize FinalizeBlock, cache execution.

noot:
## Summary

updates the sequencer app to use ABCI v0.38, which replaces `begin_block`/`deliver_tx`/`end_block` with one call, `finalize_block`.

relevant penumbra PRs:

- update penumbra-tower-trace to re-add instrumentation to the ABCI methods (done in https://github.com/penumbra-zone/penumbra/pull/4160)
- update cnidarium to be able to get the root hash of the state without calling `commit` (done in https://github.com/penumbra-zone/penumbra/pull/4122)

## Background
this was inevitable, also it cleans up the block execution flow inside the sequencer application. (removed `processed_txs` and `current_sequencer_block_builder` from the app)

## Changes
- implement `finalize_block`; for the most part, I left the `begin_block`/`deliver_tx`/`end_block` as they are but called them inside `finalize_block`
- however the block execution flow is cleaned up as we no longer need to track how many txs have been delivered (for the commitments) and we can construct the `SequencerBlock` at the end of `finalize_block` without needing to track things between calls

## Testing
unit tests 

ran it with cometbft release [v0.38.6](https://github.com/cometbft/cometbft/releases/tag/v0.38.6)

## Breaking Changelist
- the sequencer app will now only work with a cometbft version that supports ABCI v0.38

## Related Issues

closes #679


crates/astria-sequencer/src/app.rs
```diff
@@ -482,44 +679,24 @@ impl App {
...
-                    error = AsRef::<dyn std::error::Error>::as_ref(&e),
-                    "failed to decode deliver tx payload to signed transaction; ignoring it",
-                );
-                return None;
-            }
-            Ok(tx) => tx,
-        };
-
-        Some(self.deliver_tx(signed_tx).await)
+        let signed_tx = signed_transaction_from_bytes(tx)
```
Fraser999, line 697: Couldn't this just be `signed_transaction_from_bytes(tx)?;`?  We're providing context to the error inside `signed_transaction_from_bytes`, and `deliver_tx_after_proposal` is already returning an `anyhow::Result`.

crates/astria-sequencer/src/app.rs
```diff
@@ -715,6 +840,11 @@ impl App {
 
         // Get the latest version of the state, now that we've committed it.
         self.state = Arc::new(StateDelta::new(storage.latest_snapshot()));
+        self.app_hash = app_hash
+            .0
+            .to_vec()
+            .try_into()
+            .expect("root hash to app hash must succeed; both are 32 bytes");
 
         app_hash
```
Fraser999, line 849: Seems like no callsite really needs the app_hash - could we change this method to not return it?

Fraser999 (approved):
No blockers, but a few questions/suggestions/nitpicks.

crates/astria-sequencer/src/app.rs
```diff
@@ -206,7 +205,13 @@ impl App {
             .context("failed to call init_chain on IbcComponent")?;
 
         state_tx.apply();
-        Ok(())
+
+        let app_hash = self
+            .prepare_commit(storage)
+            .await
+            .context("failed to prepare commit")?;
+        debug!(app_hash = %telemetry::display::base64(&app_hash), "init_chain completed");
```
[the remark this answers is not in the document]
noot: i actually do call `commit` right after `init_chain` in the `init_chain` message handler in the consensus service, so the changes are committed. i'm not sure if that's necessary though

crates/astria-sequencer/src/app.rs
```diff
@@ -426,22 +430,215 @@ impl App {
...
+
+        // cometbft expects a result for every tx in the block, so we need to return a
+        // tx result for the commitments, even though they're not actually user txs.
+        let mut tx_results: Vec<ExecTxResult> = Vec::with_capacity(finalize_block.txs.len());
+        tx_results.push(ExecTxResult {
+            ..Default::default()
+        });
+        tx_results.push(ExecTxResult {
+            ..Default::default()
+        });
```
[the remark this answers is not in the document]
noot: nice!

crates/astria-sequencer/src/app.rs
noot: you mean returning the error as opposed to panicking? this needs to be an `expect` as it's an unrecoverable protocol error, it would only happen if the chain's consensus failed to commit valid transactions. if we return an error here instead of panicking, the chain will halt anyways as cometbft will error due to the transaction results not matching, or some other error

crates/astria-sequencer/src/app.rs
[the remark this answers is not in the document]
Fraser999: Ah - sorry - I wasn't meaning you weren't calling `commit` right after `init_chain` in this PR.

I was more thinking of down the line if there end up being more places calling `init_chain` then it would be good to just call `commit` inside `init_chain` so it doesn't get forgotten.

crates/astria-sequencer/src/service/consensus.rs
```diff
@@ -193,78 +185,35 @@ impl Consensus {
...
-        self.app.end_block(&end_block).await
+            .context("failed to call App::finalize_block")?;
+        Ok(finalize_block)
     }
 
     #[instrument(skip_all)]
     async fn commit(&mut self) -> anyhow::Result<response::Commit> {
-        let app_hash = self.app.commit(self.storage.clone()).await;
+        self.app.commit(self.storage.clone()).await;
         Ok(response::Commit {
```
Fraser999, line 214: Could be just `Ok(response::Commit::default())` I think.

crates/astria-sequencer/src/app.rs
Fraser999: I think that's ok though?  If the higher logic knows to exit the process, or it causes the process to exit because cometbft dies then we have the correct outcome.

By panicking here we lose the source chain of the error - we'll only see the contents of the eyre `.context` value.  If we do want to panic, it might be better to do it inside `signed_transaction_from_bytes` (assuming there's never a case where that fails but we can continue to run) and we can replace `.context` with `.expect` to see the more relevant details.

Cargo.toml
```diff
@@ -76,9 +76,10 @@ serde_json = "1"
 metrics = "0.22.1"
 pbjson-types = "0.6"
 # Note that when updating the penumbra versions, vendored types in `proto/sequencerapis/astria_vendored` may need to be updated as well.
-penumbra-ibc = { git = "https://github.com/penumbra-zone/penumbra.git", tag = "v0.69.1", default-features = false }
-penumbra-proto = { git = "https://github.com/penumbra-zone/penumbra.git", tag = "v0.69.1" }
-penumbra-tower-trace = { git = "https://github.com/penumbra-zone/penumbra.git", tag = "v0.69.1" }
+# update once https://github.com/penumbra-zone/penumbra/commit/8b06546af43bf073fd99f3f9d82b8afb51872489 makes it into a release
```
SuperFluffy, line 79: Let's have a tracking issue for this?

crates/astria-sequencer/Cargo.toml
```diff
@@ -28,10 +28,10 @@ anyhow = "1"
 borsh = { version = "1", features = ["derive"] }
 matchit = "0.7.2"
 tower = "0.4"
-tower-abci = "0.11.0"
+tower-abci = "0.12.0"
 tower-actor = "0.1.0"
-cnidarium = { git = "https://github.com/penumbra-zone/penumbra.git", tag = "v0.69.1" }
-cnidarium-component = { git = "https://github.com/penumbra-zone/penumbra.git", tag = "v0.69.1" }
+cnidarium = { git = "https://github.com/penumbra-zone/penumbra.git", rev = "8b06546af43bf073fd99f3f9d82b8afb51872489" }
```
SuperFluffy, line 33: Same as for the penumbra deps, right? Let's add this to a tracking issue

crates/astria-sequencer/src/app.rs
```diff
@@ -161,19 +158,20 @@ impl App {
             validator_address: None,
             executed_proposal_hash: Hash::default(),
             execution_result: HashMap::new(),
-            processed_txs: 0,
             current_proposer: None,
-            current_sequencer_block_builder: None,
+            write_batch: None,
+            app_hash: AppHash::default(),
```
SuperFluffy, line 163: This is just a `Vec<u8>` <- let's toss the sha in here? It's already available at the crate root:

```rust
            app_hash: AppHash::try_from(crate::BUILD_INFO.git_sha.to_vec()).expect("TryFrom<Vec<u8>> for AppHash cannot fail"),
```

crates/astria-sequencer/src/app.rs
```diff
@@ -426,22 +429,206 @@ impl App {
...
+            byzantine_validators: finalize_block.misbehavior.clone(),
+            header: Header {
+                app_hash: self.app_hash.clone(),
+                chain_id: chain_id.clone(),
+                consensus_hash: Hash::default(), // unused
+                data_hash: Some(Hash::try_from(data_hash.to_vec()).unwrap()),
+                evidence_hash: Some(Hash::default()), // unused
+                height: finalize_block.height,
+                last_block_id: None,                      // unused
+                last_commit_hash: Some(Hash::default()),  // unused
```
SuperFluffy, line 473: Turns out, this is just: `Some(Hash::default()) -> Some(Hash::None)`

SuperFluffy (commented):
Just comments re issues for tracking the next penumbra release && populating `AppHash` with already available data.

crates/astria-sequencer/src/app.rs
noot: why? this makes it non-deterministic which is bad, also it'll immediately get overwritten when `commit` is called after `init_chain`

Cargo.toml
noot: #964

crates/astria-sequencer/src/app.rs
noot: yes :)

crates/astria-sequencer/src/app.rs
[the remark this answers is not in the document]
noot: ah okay, the cometbft behaviour is that it only ever calls `init_chain` once, so this should be fine

crates/astria-sequencer/src/app.rs
SuperFluffy: Think I misunderstood the purpose of app hash. :)

Fraser999 (approved):

joroshiba (approved):
infra approval for charts

PR #832, opened 2024-03-18
fix(sequencer): fix bug in `get_sequencer_block_by_hash`

noot:
## Summary
found a bug :/

## Changes
- `get_sequencer_block_by_hash` to push to vec instead of indexing at index that doesn't exist

## Testing
unit tests


joroshiba (approved):

SuperFluffy:
We should add unit tests for bugs we find to get started on having a full test harness.

PR #834, opened 2024-03-19
refactor(composer): interact with executor through handle

bharath-123:
## Summary
This is a minor refactor PR which moves the creation of the channel through which executor receives SequenceActions to submit to the Shared Sequence N/w to the executor.

## Background
Previously, the channels were created by the Composer which creates the Executor. It is better practice to allow the Executor to create the channel which it owns and just return the sending end of it back to the Composer.

## Changes
- Create a struct `Handle` in Executor which is referenced as `executor::Handle`.
- Handle creation of the `executor::Handle` in Executor and allow Executor to pass it back.

## Testing
Since this is not a functional PR change, making sure that the code compiles and the tests run is enough to ensure that this works.

## Note
The tests were initially using just `send` on the channel but the executor uses `send_with_timeout`. We could add just a `send` method to the `executor::Handle`. But to avoid clippy errors w.r.t unused method `send` and avoiding a `#[cfg(test)]` in the method, I updated the tests to use `send_with_timeout` instead of `send`.

closes


crates/astria-composer/src/executor/tests.rs
```diff
@@ -229,14 +226,8 @@ async fn full_bundle() {
...
     // push both sequence actions to the executor in order to force the full bundle to be sent
-    serialized_rollup_transactions_tx
-        .send(seq0.clone())
-        .await
-        .unwrap();
-    serialized_rollup_transactions_tx
-        .send(seq1.clone())
-        .await
-        .unwrap();
+    executor_handle.get().send(seq0.clone()).await.unwrap();
```
bharath-123, line 229: might be a bad function name.

noot (approved):

crates/astria-composer/src/executor/mod.rs
```diff
@@ -103,6 +104,29 @@ pub(super) struct Executor {
     max_bytes_per_bundle: usize,
 }
 
+#[derive(Clone)]
+pub(super) struct Handle {
+    serialized_rollup_transactions_tx: mpsc::Sender<SequenceAction>,
+}
+
+impl Handle {
+    pub(super) fn new(serialized_rollup_transactions_tx: mpsc::Sender<SequenceAction>) -> Self {
```
SuperFluffy, line 113: This constructor should not be `pub(super)` - it should only be callable from inside this module

(Very minor nit: I don't think this needs an explicit constructor - just create the handle in `Executor::new` (or however it's constructor is called).)

crates/astria-composer/src/executor/mod.rs
```diff
@@ -103,6 +104,29 @@ pub(super) struct Executor {
...
+        Self {
+            serialized_rollup_transactions_tx,
+        }
+    }
+
+    pub(super) async fn send_with_timeout(
```
SuperFluffy, line 119: i like the thin wrapper - but IMO just call this `send_timeout`.

crates/astria-composer/src/executor/mod.rs
```diff
@@ -89,7 +90,7 @@ pub(super) struct Executor {
     // The status of this executor
     status: watch::Sender<Status>,
     // Channel for receiving `SequenceAction`s to be bundled.
-    serialized_rollup_transactions_rx: mpsc::Receiver<SequenceAction>,
+    sequence_action_rx: mpsc::Receiver<SequenceAction>,
```
SuperFluffy, line 93: Personally, I'd just call this `sequence_actions` or new_sequence_actions`. The `_rx` doesn't add anything of substance.

SuperFluffy (approved):
Looks good and is ready to merge after addressing the privacy issue around `executor::Handle` - it doesn't make sense allowing the construction of a handle to a thing that does not exist.

crates/astria-composer/src/executor/mod.rs
bharath-123: i think thats a good point. i ll try just creating it in the Executor::new

crates/astria-composer/src/tests.rs
```diff
@@ -27,7 +28,12 @@ async fn collector_is_reconnected_after_exit() {
     let (tx, mut rx) = tokio::sync::mpsc::channel(16);
 
     let mut collector_tasks = JoinMap::new();
-    let collector = Geth::new(rollup_name.clone(), rollup_url.clone(), tx.clone());
+    let executor_handle = executor::Handle::new(tx.clone());
```
bharath-123, line 31: here

crates/astria-test-utils/src/mock/geth.rs
```diff
@@ -238,11 +238,11 @@ impl Geth {
     /// # Errors
     ///
     /// Returns the same error as tokio's [`Sender::send`].
-    pub fn abort(&self) -> Result<usize, SendError<SubscriptionCommand>> {
+    pub fn cancel_subscriptions(&self) -> Result<usize, SendError<SubscriptionCommand>> {
```
SuperFluffy, line 241: Yeah, this is more descriptive. I was confused by `abort`.

crates/astria-composer/tests/blackbox/composer.rs
```diff
@@ -54,6 +54,43 @@ async fn tx_from_one_rollup_is_received_by_sequencer() {
...
+
+    // the collector will be restarted now, we should be able to send a tx normally
+    let expected_rollup_ids = vec![RollupId::from_unhashed_bytes("test1")];
+    let mock_guard =
+        mount_broadcast_tx_sync_mock(&test_composer.sequencer, expected_rollup_ids, vec![0]).await;
+    test_composer.rollup_nodes["test1"]
+        .push_tx(Transaction::default())
+        .unwrap();
+
+    // wait for 1 sequencer block time to make sure the bundle is preempted
```
SuperFluffy, line 82: Add a comment that you add 1000ms to make sure the websocket subscription to the collector comes back up

PR #844, opened 2024-03-20
fix(sequencer): fix `is_source` prefix check

noot:
## Summary
caught by the zellic audit, the prefix in the `Denom` type does not contain the final slash, so `prefix_is` would always be false.

## Changes
- fix `is_source` by removing final slash in prefix

## Testing
unit tests


crates/astria-sequencer/src/ibc/ics20_transfer.rs
```diff
@@ -207,7 +207,7 @@ fn is_source(
     asset: &Denom,
     is_refund: bool,
```
SuperFluffy, line 208: This isn't strictly speaking part of this PR, but this extra argument is very confusing/surprising and only really used in one spot (the function is called in 2 places, but in the argument is always `true`).

I'd remove the argument in favor of just handling this case there.

crates/astria-sequencer/src/ibc/ics20_transfer.rs
```diff
@@ -412,3 +412,21 @@ async fn execute_ics20_transfer<S: StateWriteExt>(
...
+    fn is_source_test() {
+        let source_port = "source_port".to_string().parse().unwrap();
+        let source_channel = "source_channel".to_string().parse().unwrap();
+        let asset = Denom::from("source_port/source_channel/asset".to_string());
+        // in the case of a transfer in that is not a refund,
+        // we are the source if the packets are prefixed by the sending chain
+        assert!(is_source(&source_port, &source_channel, &asset, false));
+        // in the case of a refund, we are the source if the packets are not
+        // prefixed by the sending chain
+        assert!(!is_source(&source_port, &source_channel, &asset, true));
```
SuperFluffy, line 430: These are effectively testing the same thing, right?

crates/astria-sequencer/src/ibc/ics20_transfer.rs
noot: the argument isn't always true, it's false when called from recv_packet_execute -> execute_ics20_transfer

crates/astria-sequencer/src/ibc/ics20_transfer.rs
noot: one is for the refund case and one is for non refund, but yeah they both test that the prefix checking is ok

crates/astria-sequencer/src/ibc/ics20_transfer.rs
SuperFluffy: It's used in two places:
1. Here it's hardcoded with `true`: https://github.com/astriaorg/astria/blob/561572ce1600092b1af0a036527616fb98088641/crates/astria-sequencer/src/ibc/ics20_transfer.rs#L183
2. Here it's actually a variable: https://github.com/astriaorg/astria/blob/561572ce1600092b1af0a036527616fb98088641/crates/astria-sequencer/src/ibc/ics20_transfer.rs#L345

SuperFluffy (approved):
Please also add a test for the negative case - i.e. where `is_source_prefix` returns `false`.

Also, I really think the last argument which is effectively just a post-script `not` operation should be removed and just inlined where this is used.

The last test shows why: you are running the same test twice, just one with `false` and one with `!true`.

noot:
@SuperFluffy updated is_source to just check is_prefix and then the refund/non refund operation is done in the caller!

PR #845, opened 2024-03-20
refactor(conductor) use cancellation tokens for shutdown

SuperFluffy:
## Summary
Switches from using oneshot channels to cancellation tokens.

## Background
This is much neater than using channels to signal shutdown. One token can be passed to all tasks, instead of creating (and storing) one channel per task.

Note that conductor will shut down immediately and not try to drain its queue of pending blocks.

Also note that this PR does not address conductor hanging if it receives SIGTERM during startup, where various async initializers might hang (leaving to a followup PR).

## Changes
- Replace all `oneshot` channels used for signalling shutdown with `CancellationToken`
- Make all `select!` loops biased, favoring cancellation over all other branches
- reorder the other `select!` arms, favoring sending existing blocks to executor over pulling new blocks, and heights
- Add comments to various unclear code parts
- Unify shutdown messages for all long lived tasks.

## Testing
This needs to be tested end-to-end.


crates/astria-conductor/src/sequencer/mod.rs
```diff
@@ -97,56 +93,66 @@ impl Reader {
...
+                    {
+                        Err(error) => break Err(error),
+                        Ok(block) => block,
+                    };
+                    if let Err(error) = sequential_blocks.insert(block).wrap_err("failed adding block to sequential cache") {
+                        warn!(%error, "failed pushing block into cache, dropping it");
+                    }
+                }
+
+                // Record the latst height of the Sequencer network, allowing `blocks_from_heights` to progress.
```
noot, line 155:
```suggestion
                // Record the latest height of the Sequencer network, allowing `blocks_from_heights` to progress.
```

noot (approved):

PR #852, opened 2024-03-20
refactor(conductor)!: use HTTP to periodically poll Celestia heights 

SuperFluffy:
## Summary
Replaces the websocket subscription to get the latest Celestia heights by a HTTP fetch running periodically.

## Background
Websocket adds complexity for questionable gain. Tests in particular are much more complex to set up, but also for normal operation disconnects and resubscriptions need to be handled explicitly because websocket does not have a reconnect mechanism.

This PR essentially comes full circle, removing every last trace of websocket again.

## Changes
- Replace the websocket subscription by an interval stream + HTTP fetch (not calling this GET because this is still JSONRPC dong a POST)
- Removes the `ASTRIA_CONDUCTOR_CELESTIA_NODE_WEBSOCKET_URL` config
- Adds the `ASTRIA_CONDUCTOR_CELESTIA_BLOCK_TIME_MS` config which takes an integer representing the duration in milliseconds between two successive Celestia block heights.
- Update charts to reflect the env var changes

## Testing
Will be tested using blackbox tests and smoke tests (out of scope of this PR)

## Breaking Changelist
- Removed `ASTRIA_CONDUCTOR_CELESTIA_NODE_WEBSOCKET_URL`
- Added `ASTRIA_CONDUCTOR_CELESTIA_BLOCK_TIME_MS`

### DIFF
--- a/crates/astria-conductor/src/celestia/latest_height_stream.rs
+++ b/crates/astria-conductor/src/celestia/latest_height_stream.rs
@@ -1,0 +1,63 @@
+use std::{
+    pin::Pin,
+    time::Duration,
+};
+
+use astria_eyre::eyre::{
+    Result,
+    WrapErr as _,
+};
+use celestia_client::{
+    celestia_rpc::HeaderClient as _,
+    jsonrpsee::http_client::HttpClient,
+};
+use futures::{
+    Future,
+    FutureExt as _,
+    Stream,
+    StreamExt as _,
+};
+use tokio_stream::wrappers::IntervalStream;
+
+pub(super) fn stream_latest_heights(
+    client: HttpClient,
+    poll_period: Duration,
+) -> LatestHeightStream {
+    let f = Box::new(move |_: tokio::time::Instant| {
+        let client = client.clone();
+        async move {
+            client
+                .header_network_head()
+                .await
+                .wrap_err("failed to fetch network head")
+                .map(|header| header.height().value())
+        }
+        .boxed()
+    });
+    let mut interval = tokio::time::interval(poll_period);
+    interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
+    LatestHeightStream {
+        inner: IntervalStream::new(interval).then(f),
+    }
+}
+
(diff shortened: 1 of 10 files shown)


aajimal (approved):

crates/astria-conductor/src/celestia/latest_height_stream.rs
```diff
@@ -0,0 +1,63 @@
...
+                .await
+                .wrap_err("failed to fetch network head")
+                .map(|header| header.height().value())
+        }
+        .boxed()
+    });
+    let mut interval = tokio::time::interval(poll_period);
+    interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
+    LatestHeightStream {
+        inner: IntervalStream::new(interval).then(f),
```
noot, line 40: did not know about this, very cool

crates/astria-conductor/src/celestia/latest_height_stream.rs
```diff
@@ -0,0 +1,63 @@
...
+            client
+                .header_network_head()
+                .await
+                .wrap_err("failed to fetch network head")
+                .map(|header| header.height().value())
+        }
+        .boxed()
+    });
+    let mut interval = tokio::time::interval(poll_period);
+    interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
```
noot, line 38: skipping is ok b/c conductor will fetch all blocks if a few heights are missed in the stream right?

noot (approved):
looks good!

crates/astria-conductor/src/celestia/latest_height_stream.rs
SuperFluffy: Yeah, so say the last block fetched is 10 and the latest celestia height we observed is 10 - in this case no new blocks will be fetched.

If the stream fails for 9 consecutive heights but then 10th one succeeds, then the latest observed height jumps from 10 to 20. In that moment the stream schedules 11 to 20 one after the other.

PR #853, opened 2024-03-21
refactor: generate Arc<Self> target for grpc servers

SuperFluffy:
## Summary
Change code generation to emit `Arc<Self>` as the target for tonic gRPC service traits.

## Background
Using `self: Arc<Self>` instead of `&self` is more convenient because it allows access to a service's inner attributes. The motivation for doing this is writing tests for mocks right now, but also streaming gRPC services later.

## Changes
- Add `use_arc_self(true)` to the tonic builder inside the protobuf generation tool.
- Update all types that implement generated gRPC services (mainly in sequencer, but also in sequencer-relayer and conductor for mocks).

## Testing
Tests still run, code still compiles. This change should be entirely opaque.

### DIFF
--- a/crates/astria-conductor/src/executor/tests.rs
+++ b/crates/astria-conductor/src/executor/tests.rs
@@ -1,7 +1,10 @@
 use std::{
     collections::HashMap,
     net::SocketAddr,
-    sync::Mutex,
+    sync::{
+        Arc,
+        Mutex,
+    },
 };
 
 use astria_core::{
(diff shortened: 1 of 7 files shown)


noot (approved):
does this allow for mutating self (eg maybe for subscription APIs later?)

SuperFluffy:
> does this allow for mutating self (eg maybe for subscription APIs later?)

Yes it does - albeit through mutexes, rwlocks etc. I wanted this for writing blackbox tests against conductor, so that I can hold an `Arc<Self>` of the grpc server in a mock-object and mount mocks on it.

The original reason to introduce this in tonic was situations where you'd `tokio::spawn` something from a handler (like a long-lived stream). But when you do that you can't access `&self` anymore (because it's not `'static'`). So the fix to that is wrapping whatever fields you need in `Arc<YourType>`.

PR #854, opened 2024-03-21
feat(composer): add graceful shutdown

bharath-123:
## Summary
Implements graceful shutdown for Composer components. 

## Background
Composer has 3 main components:
1. Geth Collectors
2. Grpc Server/Collector
3. Executor

Currently, we do not have any sort of graceful shutdown for these Components. This PR implements it.

When a Geth Collector shuts down, we do the following:
1. We need to unsubscribe to the websocket stream
2. We need to set the status of the geth collector to false(not connected)

For the Grpc Server, we just pass in a future which resolves when a shutdown signal is passed to the server by running it as `serve_with_incoming_shutdown`. 

When an Executor shuts down, we do the following:
1. Close the `serialized_rollup_transactions` channel to avoid receiving transactions. 
2. Set Executor status to false(not connected).
3. Drain any remaining bundles

We ensure that the API server shuts down at the end by giving it its own CancellationToken. This is because we need the API server to be up in order for k8s to report liveness.

## Changes
- Add a shutdown token which is a tokio CancellationToken
- Pass it to the GethCollectors, GrpcServer and Executor
- Convert the while loop which listens to txs on the Geth Wss stream to a select stmt in a loop to listen for the shutdown signal from the shutdown CancellationToken.
- Add shutdown logic to Executor
- Pass in the shutdown CancellationToken to the gRPC server with the `serve_with_incoming_shutdown` method.
- Pass a new CancellationToken to the API server and ensure it shuts down after every other task is shutdown.

**Note:** 
I had to refactor the run_until_stopped method in Composer as Clippy was complaining a lot about the below:
https://rust-lang.github.io/rust-clippy/master/index.html#/too_many_lines

## Testing
Tested by manually deploying a local shared sequencer cluster and running a Composer instance against it. Sent a SIGTERM signal to Composer and tested the shutdown.

### DIFF
--- a/crates/astria-composer/src/collectors/geth.rs
+++ b/crates/astria-composer/src/collectors/geth.rs
@@ -47,12 +47,13 @@ use tracing::{
     warn,
 };
 
-use crate::executor;
+use crate::{
+    collectors::EXECUTOR_SEND_TIMEOUT,
+    executor,
+};
 type StdError = dyn std::error::Error;
 
 const WSS_UNSUBSCRIBE_TIMEOUT: Duration = Duration::from_secs(2);
-
-pub(super) const EXECUTOR_SEND_TIMEOUT: Duration = Duration::from_millis(500);
 
 /// `GethCollector` Collects transactions submitted to a Geth rollup node and passes
 /// them downstream for further processing.
--- a/crates/astria-composer/src/collectors/geth.rs
+++ b/crates/astria-composer/src/collectors/geth.rs
@@ -13,6 +13,8 @@
 //! { "id": 1, "jsonrpc": "2.0", "method": "eth_subscribe", "params": ["newPendingTransactions"] }
 //! ```
 
+use std::time::Duration;
+
 use astria_core::sequencer::v1::{
     asset::default_native_asset_id,
     transaction::action::SequenceAction,
@@ -20,6 +22,8 @@ use astria_core::sequencer::v1::{
 };
 use astria_eyre::eyre::{
     self,
+    eyre,
+    Report,
     WrapErr as _,
 };
 use ethers::providers::{
@@ -27,12 +31,17 @@ use ethers::providers::{
     ProviderError,
     Ws,
 };
-use tokio::sync::{
-    mpsc::error::SendTimeoutError,
-    watch,
+use tokio::{
+    select,
+    sync::{
+        mpsc::error::SendTimeoutError,
+        watch,
+    },
 };
+use tokio_util::sync::CancellationToken;
 use tracing::{
     debug,
+    error,
     info,
     instrument,
     warn,
@@ -40,6 +49,10 @@ use tracing::{
 
 use crate::executor;
 type StdError = dyn std::error::Error;
+
+const WSS_UNSUBSCRIBE_TIMEOUT: Duration = Duration::from_secs(2);
+
+pub(super) const EXECUTOR_SEND_TIMEOUT: Duration = Duration::from_millis(500);
 
 /// `GethCollector` Collects transactions submitted to a Geth rollup node and passes
 /// them downstream for further processing.
@@ -57,8 +70,10 @@ pub(crate) struct Geth {
     executor_handle: executor::Handle,
     // The status of this collector instance.
     status: watch::Sender<Status>,
-    /// Rollup URL
+    // Rollup URL
     url: String,
+    // Token to signal the geth collector to stop upon shutdown.
+    shutdown_token: CancellationToken,
 }
 
 #[derive(Debug)]
@@ -80,7 +95,12 @@ impl Status {
 
 impl Geth {
     /// Initializes a new collector instance
-    pub(crate) fn new(chain_name: String, url: String, executor_handle: executor::Handle) -> Self {
+    pub(crate) fn new(
+        chain_name: String,
+        url: String,
+        executor_handle: executor::Handle,
+        shutdown_token: CancellationToken,
+    ) -> Self {
         let (status, _) = watch::channel(Status::new());
         let rollup_id = RollupId::from_unhashed_bytes(&chain_name);
         info!(
@@ -94,6 +114,7 @@ impl Geth {
             executor_handle,
             status,
             url,
+            shutdown_token,
         }
     }
 
@@ -116,6 +137,7 @@ impl Geth {
             executor_handle,
             status,
             url,
+            shutdown_token,
             ..
         } = self;
 
@@ -155,37 +177,79 @@ impl Geth {
 
         status.send_modify(|status| status.is_connected = true);
 
-        while let Some(tx) = tx_stream.next().await {
-            let tx_hash = tx.hash;
-            debug!(transaction.hash = %tx_hash, "collected transaction from rollup");
-            let data = tx.rlp().to_vec();
-            let seq_action = SequenceAction {
-                rollup_id,
-                data,
-                fee_asset_id: default_native_asset_id(),
-            };
+        let reason = loop {
+            select! {
+                biased;
+                () = shutdown_token.cancelled() => {
+                    break Ok("shutdown signal received");
+                },
+                tx_res = tx_stream.next() => {
+                    if let Some(tx) = tx_res {
+                        let tx_hash = tx.hash;
+                        debug!(transaction.hash = %tx_hash, "collected transaction from rollup");
+                        let data = tx.rlp().to_vec();
+                        let seq_action = SequenceAction {
+                            rollup_id,
+                            data,
+                            fee_asset_id: default_native_asset_id(),
+                        };
 
-            match executor_handle
-                .send_timeout(seq_action, Duration::from_millis(500))
-                .await
-            {
-                Ok(()) => {}
-                Err(SendTimeoutError::Timeout(_seq_action)) => {
-                    warn!(
-                        transaction.hash = %tx_hash,
-                        "timed out sending new transaction to executor after 500ms; dropping tx"
-                    );
-                }
-                Err(SendTimeoutError::Closed(_seq_action)) => {
-                    warn!(
-                        transaction.hash = %tx_hash,
-                        "executor channel closed while sending transaction; dropping transaction \
-                         and exiting event loop"
-                    );
-                    break;
+                        match executor_handle
+                            .send_timeout(seq_action, EXECUTOR_SEND_TIMEOUT)
+                            .await
+                        {
+                            Ok(()) => {}
+                            Err(SendTimeoutError::Timeout(_seq_action)) => {
+                                warn!(
+                                    transaction.hash = %tx_hash,
+                                    timeout_ms = EXECUTOR_SEND_TIMEOUT.as_millis(),
+                                    "timed out sending new transaction to executor; dropping tx",
+                                );
+                            }
+                            Err(SendTimeoutError::Closed(_seq_action)) => {
+                                warn!(
+                                    transaction.hash = %tx_hash,
+                                    "executor channel closed while sending transaction; dropping transaction \
+                                     and exiting event loop"
+                                );
+                                break Err(eyre!("executor channel closed while sending transaction"));
+                            }
+                        }
+                    } else {
+                        break Err(eyre!("geth tx stream ended"));
+                    }
                 }
             }
+        };
+
+        match &reason {
+            Ok(reason) => {
+                info!(reason, "shutting down");
+            }
+            Err(reason) => {
+                error!(%reason, "shutting down");
+            }
+        };
+
+        status.send_modify(|status| status.is_connected = false);
+
+        // if the loop exits with an error, we can still proceed with unsubscribing the WSS
+        // stream as we could have exited due to an error in sending messages via the executor
+        // channel.
+
+        // give 2s for the websocket connection to be unsubscribed as we want to avoid having
+        // this hang for too long
+        match tokio::time::timeout(WSS_UNSUBSCRIBE_TIMEOUT, tx_stream.unsubscribe()).await {
+            Ok(Ok(true)) => info!("unsubscribed from geth tx stream"),
+            Ok(Ok(false)) => warn!("failed to unsubscribe from geth tx stream"),
+            Ok(Err(err)) => {
+                error!(error = %Report::new(err), "failed unsubscribing from the geth tx stream");
+            }
+            Err(err) => {
+                error!(error = %Report::new(err), "timed out while unsubscribing from the geth tx stream");
+            }
         }
-        Ok(())
+
+        reason.map(|_| ())
     }
 }
--- a/crates/astria-composer/src/collectors/grpc.rs
+++ b/crates/astria-composer/src/collectors/grpc.rs
@@ -1,9 +1,6 @@
 //! `GrpcCollector` implements the `GrpcCollectorService` rpc service.
 
-use std::{
-    sync::Arc,
-    time::Duration,
-};
+use std::sync::Arc;
 
 use astria_core::{
     generated::composer::v1alpha1::{
@@ -24,7 +21,10 @@ use tonic::{
     Status,
 };
 
-use crate::executor;
+use crate::{
+    collectors::geth::EXECUTOR_SEND_TIMEOUT,
+    executor,
+};
 
 /// Implements the `GrpcCollectorService` which listens for incoming gRPC requests and
 /// sends the Rollup transactions to the Executor. The Executor then sends the transactions
@@ -61,7 +61,7 @@ impl GrpcCollectorService for Grpc {
 
         match self
             .executor
-            .send_timeout(sequence_action, Duration::from_millis(500))
+            .send_timeout(sequence_action, EXECUTOR_SEND_TIMEOUT)
             .await
         {
             Ok(()) => {}
--- a/crates/astria-composer/src/collectors/grpc.rs
+++ b/crates/astria-composer/src/collectors/grpc.rs
@@ -22,7 +22,7 @@ use tonic::{
 };
 
 use crate::{
-    collectors::geth::EXECUTOR_SEND_TIMEOUT,
+    collectors::EXECUTOR_SEND_TIMEOUT,
     executor,
 };
 
--- a/crates/astria-composer/src/collectors/mod.rs
+++ b/crates/astria-composer/src/collectors/mod.rs
@@ -1,5 +1,9 @@
 pub(crate) mod geth;
 pub(crate) mod grpc;
 
+const EXECUTOR_SEND_TIMEOUT: Duration = Duration::from_millis(500);
+
+use std::time::Duration;
+
 pub(crate) use geth::Geth;
 pub(crate) use grpc::Grpc;
--- a/crates/astria-composer/src/composer.rs
+++ b/crates/astria-composer/src/composer.rs
@@ -49,6 +49,7 @@ use crate::{
 
 const API_SERVER_SHUTDOWN_DURATION: Duration = Duration::from_secs(2);
 const GRPC_SERVER_SHUTDOWN_DURATION: Duration = Duration::from_secs(5);
+const EXECUTOR_SHUTDOWN_DURATION: Duration = Duration::from_secs(17);
 const GETH_COLLECTOR_SHUTDOWN_DURATION: Duration = Duration::from_secs(5);
 
 /// `Composer` is a service responsible for spinning up `GethCollectors` which are responsible
@@ -349,10 +350,13 @@ impl ShutdownInfo {
         // We give executor 17 seconds to shut down. The logic to timeout is in the
         // executor itself. We wait 17s for all the bundles to be drained.
         if let Some(executor_task_handle) = executor_task_handle {
-            match executor_task_handle.await {
-                Ok(Ok(())) => info!("executor shut down successfully"),
+            match tokio::time::timeout(EXECUTOR_SHUTDOWN_DURATION, executor_task_handle)
+                .await
+                .map(flatten_result)
+            {
+                Ok(Ok(())) => info!("executor task shut down"),
                 Ok(Err(error)) => error!(%error, "executor task shut down with error"),
-                Err(error) => error!(%error, "executor task didn't complete successfully"),
+                Err(error) => error!(%error, "executor task failed to shut down in time"),
             }
         } else {
             info!("executor task was already dead");
--- a/crates/astria-composer/src/composer.rs
+++ b/crates/astria-composer/src/composer.rs
@@ -1,21 +1,35 @@
 use std::{
     collections::HashMap,
     net::SocketAddr,
+    time::Duration,
 };
 
 use astria_eyre::eyre::{
     self,
     WrapErr as _,
 };
+use itertools::Itertools as _;
 use tokio::{
     io,
+    signal::unix::{
+        signal,
+        SignalKind,
+    },
     sync::watch,
-    task::JoinError,
+    task::{
+        JoinError,
+        JoinHandle,
+    },
+    time::timeout,
 };
-use tokio_util::task::JoinMap;
+use tokio_util::{
+    sync::CancellationToken,
+    task::JoinMap,
+};
 use tracing::{
     error,
     info,
+    warn,
 };
 
 use crate::{
@@ -32,6 +46,10 @@ use crate::{
     rollup::Rollup,
     Config,
 };
+
+const API_SERVER_SHUTDOWN_DURATION: Duration = Duration::from_secs(2);
+const GRPC_SERVER_SHUTDOWN_DURATION: Duration = Duration::from_secs(5);
+const GETH_COLLECTOR_SHUTDOWN_DURATION: Duration = Duration::from_secs(5);
 
 /// `Composer` is a service responsible for spinning up `GethCollectors` which are responsible
 /// for fetching pending transactions submitted to the rollup Geth nodes and then passing them
@@ -61,6 +79,8 @@ pub struct Composer {
     /// The gRPC server that listens for incoming requests from the collectors via the
     /// GrpcCollector service. It also exposes a health service.
     grpc_server: GrpcServer,
+    /// Used to signal the Composer to shut down.
+    shutdown_token: CancellationToken,
 }
 
 /// Announces the current status of the Composer for other modules in the crate to use
@@ -93,17 +113,24 @@ impl Composer {
     /// See `[from_config]` for its error scenarios.
     pub async fn from_config(cfg: &Config) -> eyre::Result<Self> {
         let (composer_status_sender, _) = watch::channel(Status::default());
+        let shutdown_token = CancellationToken::new();
+
         let (executor, executor_handle) = Executor::new(
             &cfg.sequencer_url,
             &cfg.private_key,
             cfg.block_time_ms,
             cfg.max_bytes_per_bundle,
+            shutdown_token.clone(),
         )
         .wrap_err("executor construction from config failed")?;
 
-        let grpc_server = GrpcServer::new(cfg.grpc_addr, executor_handle.clone())
-            .await
-            .wrap_err("failed to create grpc server")?;
+        let grpc_server = GrpcServer::new(
+            cfg.grpc_addr,
+            executor_handle.clone(),
+            shutdown_token.clone(),
+        )
+        .await
+        .wrap_err("grpc collector construction from config failed")?;
 
         info!(
             listen_addr = %grpc_server.local_addr().wrap_err("grpc server listener not bound")?,
@@ -128,8 +155,12 @@ impl Composer {
         let geth_collectors = rollups
             .iter()
             .map(|(rollup_name, url)| {
-                let collector =
-                    Geth::new(rollup_name.clone(), url.clone(), executor_handle.clone());
+                let collector = Geth::new(
+                    rollup_name.clone(),
+                    url.clone(),
+                    executor_handle.clone(),
+                    shutdown_token.clone(),
+                );
                 (rollup_name.clone(), collector)
             })
             .collect::<HashMap<_, _>>();
@@ -149,6 +180,7 @@ impl Composer {
             geth_collector_statuses,
             geth_collector_tasks: JoinMap::new(),
             grpc_server,
+            shutdown_token,
         })
     }
 
@@ -168,10 +200,13 @@ impl Composer {
     ///
     /// # Errors
     /// It errors out if the API Server, Executor or any of the Geth Collectors fail to start.
+    ///
+    /// # Panics
+    /// It panics if the Composer cannot set the SIGTERM listener.
     pub async fn run_until_stopped(self) -> eyre::Result<()> {
         let Self {
             api_server,
-            composer_status_sender,
+            mut composer_status_sender,
             executor,
             executor_handle,
             mut geth_collector_tasks,
@@ -179,28 +214,35 @@ impl Composer {
             rollups,
             mut geth_collector_statuses,
             grpc_server,
+            shutdown_token,
         } = self;
 
+        // we need the API server to shutdown at the end, since it is used by k8s
+        // to report the liveness of the service
+        let api_server_shutdown_token = CancellationToken::new();
+
+        let api_server_task_shutdown_token = api_server_shutdown_token.clone();
         // run the api server
-        let mut api_task =
-            tokio::spawn(async move { api_server.await.wrap_err("api server ended unexpectedly") });
+        let mut api_task = tokio::spawn(async move {
+            api_server
+                .with_graceful_shutdown(api_server_task_shutdown_token.cancelled())
+                .await
+                .wrap_err("api server ended unexpectedly")
+        });
 
         // run the collectors and executor
-        for (chain_id, collector) in geth_collectors.drain() {
-            geth_collector_tasks.spawn(chain_id, collector.run_until_stopped());
-        }
+        spawn_geth_collectors(&mut geth_collectors, &mut geth_collector_tasks);
+
         let executor_status = executor.subscribe().clone();
         let mut executor_task = tokio::spawn(executor.run_until_stopped());
 
         // wait for collectors and executor to come online
-        wait_for_collectors(&geth_collector_statuses).await?;
-        composer_status_sender.send_modify(|status| {
-            status.set_all_collectors_connected(true);
-        });
-        wait_for_executor(executor_status).await?;
-        composer_status_sender.send_modify(|status| {
-            status.set_executor_connected(true);
-        });
+        wait_for_collectors(&geth_collector_statuses, &mut composer_status_sender)
+            .await
+            .wrap_err("geth collectors failed to become ready")?;
+        wait_for_executor(executor_status, &mut composer_status_sender)
+            .await
+            .wrap_err("executor failed to become ready")?;
 
         // run the grpc server
         let mut grpc_server_handle = tokio::spawn(async move {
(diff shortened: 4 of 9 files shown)


crates/astria-composer/src/executor/mod.rs
```diff
@@ -226,10 +231,19 @@ impl Executor {
...
 
-        loop {
+        let reason = loop {
             select! {
                 biased;
 
+                () = self.shutdown_token.cancelled() => {
+                    info!("executor shutting down.");
+                    // do not accept any new txs during shutdown.
+                    self.serialized_rollup_transactions.close();
```
Fraser999, line 241: Is this required?  Looks like the `bundle_factory` won't receive any more txs if we've exited this loop.

Fraser999 (changes requested):
Mostly lgtm.  A few nitpicks/questions, but I'll block for now due to the potential panic in `fn shutdown`.

crates/astria-composer/src/executor/mod.rs
bharath-123: yeah if we are shutting the executor down, we do not want to receive any more txs from the collectors.

crates/astria-composer/src/collectors/geth.rs
```diff
@@ -155,37 +168,53 @@ impl Geth {
...
+                                "timed out sending new transaction to executor after 500ms; dropping tx"
+                            );
+                            }
+                            Err(SendTimeoutError::Closed(_seq_action)) => {
+                                warn!(
+                                transaction.hash = %tx_hash,
+                                "executor channel closed while sending transaction; dropping transaction \
+                                 and exiting event loop"
+                            );
+                                break;
```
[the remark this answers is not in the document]
bharath-123: good catch. I ll do that.

crates/astria-composer/src/collectors/geth.rs
```diff
@@ -155,37 +168,53 @@ impl Geth {
...
-                        "timed out sending new transaction to executor after 500ms; dropping tx"
-                    );
-                }
-                Err(SendTimeoutError::Closed(_seq_action)) => {
-                    warn!(
-                        transaction.hash = %tx_hash,
-                        "executor channel closed while sending transaction; dropping transaction \
-                         and exiting event loop"
-                    );
+        loop {
```
SuperFluffy, line 171: Can you streamline the exit process of this loop? Something like this:

```rust
let reason = loop { select!( select_arm = some_future { break Ok("received shutdown signal"; }  ) };

Then you can a) report why the select-loop exited, and b) have all the shutdown logic for your particular task at the bottm.

File: crates/astria-composer/src/collectors/geth.rs [hunk 4]
```diff
@@ -80,7 +86,12 @@ impl Status {
 
 impl Geth {
     /// Initializes a new collector instance
-    pub(crate) fn new(chain_name: String, url: String, executor_handle: executor::Handle) -> Self {
+    pub(crate) fn new(
```
SuperFluffy, line 89: I noted that for the `grpc` server, but would like to see this here as well: can you replace this with a `GethBuilder` that has `pub(crate)` fields for each of the arguments of this constructor? These methods have a tendency to grow indefinitely and become very hard to understand.

I am fine if you do it in a followup.

crates/astria-composer/src/composer.rs
```diff
@@ -93,17 +109,24 @@ impl Composer {
...
         )
         .wrap_err("executor construction from config failed")?;
 
-        let grpc_server = GrpcServer::new(cfg.grpc_addr, executor_handle.clone())
-            .await
-            .wrap_err("failed to create grpc server")?;
+        let grpc_server = GrpcServer::new(
+            cfg.grpc_addr,
+            executor_handle.clone(),
+            shutdown_token.clone(),
```
SuperFluffy, line 126: I think this shutdown token on the gRPC server is too simple. It fulfills a similar role as the API server, so it should do two things:

1. stop receiving requests on its composer service,
2. but keep serving requests on its health service.

crates/astria-composer/src/grpc.rs
```diff
@@ -68,9 +72,10 @@ impl GrpcServer {
             .await;
 
         grpc_server
-            .serve_with_incoming(tokio_stream::wrappers::TcpListenerStream::new(
-                self.listener,
-            ))
+            .serve_with_incoming_shutdown(
+                tokio_stream::wrappers::TcpListenerStream::new(self.listener),
+                self.shutdown_token.cancelled(),
```
SuperFluffy, line 77: Noted above: I think this doesn't work.

We need instead some kind of `shutting_down: AtomicBool` that is checked in the server's `ComposerService`, refusing new transactions while still serving health requests.

crates/astria-composer/src/executor/mod.rs
SuperFluffy: I understand the logic behind using `Receiver::close`, but I believe the channel is not currently drained after the loop.

crates/astria-composer/src/executor/mod.rs
```diff
@@ -276,6 +290,37 @@ impl Executor {
                     }
                 }
             }
+        };
+
+        match reason {
```
SuperFluffy, line 295: The logs here are confusing - shouldn't they mention 1) that the service is starting its shutdown procedure, 2) providing the `reason`, 3) reporting what other steps are taken prior to shutdown, and finally 4) reporting that it is now shutting down?

crates/astria-composer/src/executor/mod.rs
```diff
@@ -276,6 +290,37 @@ impl Executor {
...
+                    }
+                    bundles_to_drain.push(bundle);
+                }
+                let num_bundles_to_drain = bundles_to_drain.len();
+
+                // we need to submit the bundles sequentially since that is the guarantee
+                // the bundle factory gives to the end users.
+                for bundle in bundles_to_drain {
+                    let res = self.submit_bundle(nonce, bundle).await?;
+                    nonce = res;
```
SuperFluffy, line 315: This entire process should be guarded by a timeout. If your 25 (or 30) seconds are reached, you don't want to exit here uncontrolled. It's better to try and submit and - if it failed - report the txes that could not be submitted.

Also, please no `?` without context.

SuperFluffy (changes requested):
Left a few comments that I think need rework.

crates/astria-composer/src/collectors/geth.rs
```diff
@@ -155,37 +168,53 @@ impl Geth {
...
+
+                        match executor_handle
+                            .send_timeout(seq_action, Duration::from_millis(500))
+                            .await
+                        {
+                            Ok(()) => {}
+                            Err(SendTimeoutError::Timeout(_seq_action)) => {
+                                warn!(
+                                transaction.hash = %tx_hash,
+                                "timed out sending new transaction to executor after 500ms; dropping tx"
```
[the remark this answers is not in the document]
bharath-123: problem is that rustfmt is not formatting select! stmts :(

crates/astria-composer/src/executor/mod.rs
bharath-123: will remove the `?`. I ll add a timeout here. I initially thought we could just have the timeout in the Composer, but i think it makes sense for reporting the bundles which we lost.

crates/astria-composer/src/collectors/geth.rs
bharath-123: I ll do these in a future PR.

crates/astria-composer/src/executor/mod.rs
bharath-123: Makes sense. I ll structure the logs accordingly.

crates/astria-composer/src/composer.rs
bharath-123: hmm i had thought of this initially but I thought it was fine just letting the gRPC server shutdown gracefully as it ll take these services down. let me try doing what you suggested.

crates/astria-composer/src/executor/mod.rs
bharath-123: its a good point. the receve future might not have been polled yet and we would loose txs sent. I did not think of that. I will drain the channel!

crates/astria-composer/src/collectors/geth.rs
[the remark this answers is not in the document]
Fraser999: Yeah rustfmt doesn't like macros!

crates/astria-composer/src/executor/mod.rs
```diff
@@ -276,7 +285,93 @@ impl Executor {
...
+                        return Err(e);
+                    }
+                }
+            }
+            info!("drained {} bundles successfully", num_bundles_to_drain);
+
+            Ok(())
+        };
+
+        match tokio::time::timeout(Duration::from_secs(13), shutdown_logic).await {
```
Fraser999, line 361: I think we should avoid using a timeout here _and_ for the `executor_task_handle` future in the composer's `ShutdownInfo::run`.  Seems like they're both defending against the same thing - i.e. this executor task not returning in time, so we can probably remove the timeout from the composer (maybe with a comment saying it already has protection against running for too long?)  If the composer's timeout happens to execute before we finish up shutting down in the executor, then we lose the benefit of changes here, i.e. we'd potentially not send all the bundles, or report unsent bundles.

If we do want to keep both timeouts, then I'd suggest providing a `pub(super)` const for the timeout duration in the executor module, and use that (+ some wiggle room) in the composer.  But that seems suboptimal to me.

bharath-123: I think we can just keep it in executor, it could make more sense. my thinking was that we could have the lower level timeout task(i.e executor) be much bigger and the 13s would be attributed to only that specific shutdown logic of re-submitting the bundles but other logic needn't be in a timeout. But on 2nd thought, thats overkill for now. I ll just use a timeout in the executor and await the future in the composer.

crates/astria-composer/src/executor/mod.rs
```diff
@@ -276,7 +285,93 @@ impl Executor {
...
+            Ok(res) => {
+                if let Err(e) = res {
+                    error!(%e, "executor shutdown tasks aborted");
+                } else {
+                    info!("executor shutdown tasks completed successfully");
+                }
+            }
+            Err(err) => {
+                error!(%err, "executor shutdown tasks failed to complete in time");
```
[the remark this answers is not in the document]
bharath-123: Yeah my initial thought is that i don't see the point of displaying the bundles? plus a bundle is a bunch of txs which wouldn't make sense displayed on the screen. I do not see the value of logging or tracking that, since we cannot as of now retry any submission offchain with that info.
I ll spend some more time thinking of it, but i don't see any strong reason as of now to display the missed bundles rather than just mentioning the number of bundles missed which could be good info, since that would inform us of the backpressure in executor.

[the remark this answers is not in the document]
bharath-123: i was able to get it to report the bundles which weren't drained for now. if in prod, we see that it makes the logs unreadable or so, i can remove it!

crates/astria-composer/src/executor/mod.rs
bharath-123: i have done this! moved the timeout to executor and I am just awaiting the task in composer

Fraser999 (approved):
LGTM

crates/astria-composer/src/composer.rs
```diff
@@ -232,26 +312,166 @@ impl Composer {
...
+            while let Some((name, res)) = geth_collector_tasks.join_next().await {
+                let message = "task shut down";
+                match flatten(res) {
+                    Ok(()) => info!(name, message),
+                    Err(error) => error!(name, %error, message),
+                }
+            }
+        };
+
+        if geth_collectors_tasks_no > 0 {
```
SuperFluffy, line 398: IMO this is unnecessary information. Just timeout the shutdown loop without checking the number of collector tasks.

crates/astria-composer/src/composer.rs
```diff
@@ -232,26 +312,166 @@ impl Composer {
...
+                warn!(
+                    tasks = format_args!("[{tasks}]"),
+                    "aborting all geth collector tasks that have not yet shut down",
+                );
+                geth_collector_tasks.abort_all();
+            } else {
+                info!("all geth collector tasks shut down regularly");
+            }
+        } else {
+            info!("no geth collector tasks were running to shut down");
```
SuperFluffy, line 419: As above, no need for this IMO

crates/astria-composer/src/executor/mod.rs
```diff
@@ -276,7 +287,101 @@ impl Executor {
...
+        );
+
+        // log all the bundles that have not been drained
+        loop {
+            let bundle = bundle_factory.pop_now();
+            if bundle.is_empty() {
+                break;
+            }
+
+            warn!(
```
SuperFluffy, line 378: Please don't fire events in a hot loop. `serde::Serialize` is implemented for `[T], T: Serialize` and also for `Vec<T>: T: Serialize`, so you could just do this:
```rust
// feel free to come up with a better name ^.^
let mut undrained_bundles = Vec::new(); // potentially ::with_capacity since BundleFactory might have that info
loop {
    let bundle = bundle_factory.pop_now();
    if bundle.is_empty() {
        break;
    }
    undrained_bundles.push(bundle);
}
warn!(
    undrained_bundles = %telemetry::display::json(&undrained_bundles);
    "unable to drain all bundles within the allocated time",
);
```

crates/astria-composer/src/executor/mod.rs
```diff
@@ -276,7 +287,101 @@ impl Executor {
...
+                        debug!(
+                            bundle = %telemetry::display::json(&SizedBundleReport(&bundle)),
+                            "drained bundle successfully; new nonce: {}", new_nonce
+                        );
+                        nonce = new_nonce;
+                        total_bundles_drained_successfully += 1;
+                    }
+                    Err(e) => {
+                        error!(bundle =  %telemetry::display::json(&SizedBundleReport(&bundle)),
+                            %e, "failed submitting bundle to sequencer during shutdown; \
```
SuperFluffy, line 339: Please call all errors fields `error`

crates/astria-composer/src/executor/mod.rs
```diff
@@ -226,10 +233,14 @@ impl Executor {
 
         let reset_time = || Instant::now() + self.block_time;
 
-        loop {
+        let reason = loop {
             select! {
                 biased;
 
+                () = self.shutdown_token.cancelled() => {
+                    info!("shutdown signal received");
```
SuperFluffy, line 241: That info is redundant. You are going to print the same info twice immediately after exiting this loop

crates/astria-composer/src/executor/mod.rs
```diff
@@ -276,7 +287,101 @@ impl Executor {
...
+
+        let mut total_bundles_drained_successfully = 0;
+        let shutdown_logic = async {
+            info!("setting executor connection status to disconnected");
+            self.status
+                .send_modify(|status| status.is_connected = false);
+
+            info!("draining sequence actions from the executor receiver channel");
+            // drain the receiver channel
+            while let Ok(seq_action) = self.serialized_rollup_transactions.try_recv() {
```
SuperFluffy, line 309: Shouldn't this channel be closed before trying to drain it?

crates/astria-composer/src/executor/mod.rs
```diff
@@ -276,7 +287,101 @@ impl Executor {
...
+                if let Err(e) = bundle_factory.try_push(seq_action) {
+                    warn!(
+                        rollup_id = %rollup_id,
+                        error = &e as &StdError,
+                        "failed to bundle sequence action: too large. sequence action is dropped."
+                    );
+                }
+            }
+
+            info!("draining remaining bundles from bundle factory");
```
SuperFluffy, line 320: If you pop all the bundles from the factory before you start submitting them, you can report the total number of not-yet-submitted bundles.

crates/astria-composer/src/executor/mod.rs
```diff
@@ -276,7 +287,101 @@ impl Executor {
...
+                }
+            }
+
+            Ok(())
+        };
+
+        match tokio::time::timeout(Duration::from_secs(EXECUTOR_SHUTDOWN_TIME), shutdown_logic)
+            .await
+        {
+            Ok(res) => {
```
SuperFluffy, line 354: You can nest the matches to make this neater:
```rust
match foo() {
    Ok(Ok(_)) => todo!(),
    Ok(Err(error)) => todo!(),
    Err(error) => todo("timeout error"),
}
```

crates/astria-composer/src/executor/mod.rs
```diff
@@ -276,7 +287,101 @@ impl Executor {
...
+                if let Err(e) = res {
+                    error!(%e, "executor shutdown tasks aborted");
```
SuperFluffy, line 356: Name the field `error`

crates/astria-composer/src/executor/mod.rs
```diff
@@ -276,7 +287,101 @@ impl Executor {
...
+                } else {
+                    info!("executor shutdown tasks completed successfully");
+                }
+            }
+            Err(err) => {
+                error!(%err, "executor shutdown tasks failed to complete in time");
```
SuperFluffy, line 362: Please name this field `error` as well.

crates/astria-composer/src/executor/mod.rs
```diff
@@ -276,7 +287,101 @@ impl Executor {
...
+            loop {
+                let bundle = bundle_factory.pop_now();
+                if bundle.is_empty() {
+                    break;
+                }
+                match self.submit_bundle(nonce, bundle.clone()).await {
+                    Ok(new_nonce) => {
+                        debug!(
+                            bundle = %telemetry::display::json(&SizedBundleReport(&bundle)),
+                            "drained bundle successfully; new nonce: {}", new_nonce
```
SuperFluffy, line 332: put `new_nonce` into a field instead of the format string.

crates/astria-composer/src/executor/mod.rs
```diff
@@ -276,7 +287,101 @@ impl Executor {
...
+                }
+            }
+            Err(err) => {
+                error!(%err, "executor shutdown tasks failed to complete in time");
+            }
+        }
+
+        info!(
+            "{} bundles drained successfully during shutdown",
+            total_bundles_drained_successfully
```
SuperFluffy, line 368: This is a pretty confusing message if there were no bundles to be submitted ever:

Maybe:
```rust
info!(total, "bundles submitted during shutdown");
```

crates/astria-composer/src/executor/mod.rs
```diff
@@ -276,7 +287,101 @@ impl Executor {
...
+                        nonce = new_nonce;
+                        total_bundles_drained_successfully += 1;
+                    }
+                    Err(e) => {
+                        error!(bundle =  %telemetry::display::json(&SizedBundleReport(&bundle)),
+                            %e, "failed submitting bundle to sequencer during shutdown; \
+                                aborting shutdown");
+                        // if we can't submit a bundle after multiple retries, we can abort
+                        // the shutdown process
+                        return Err(e);
```
SuperFluffy, line 343: Comment: You will probably never get here here if `submit_bundle` hangs because it usually tries indefinitely.

crates/astria-composer/src/executor/mod.rs
```diff
@@ -276,7 +287,101 @@ impl Executor {
...
+
+            info!("draining sequence actions from the executor receiver channel");
+            // drain the receiver channel
+            while let Ok(seq_action) = self.serialized_rollup_transactions.try_recv() {
+                let rollup_id = seq_action.rollup_id;
+                if let Err(e) = bundle_factory.try_push(seq_action) {
+                    warn!(
+                        rollup_id = %rollup_id,
+                        error = &e as &StdError,
+                        "failed to bundle sequence action: too large. sequence action is dropped."
```
SuperFluffy, line 315:
```suggestion
                        "failed to bundle sequence action, dropping it"
```
You don't know if the size is the reason - the reason is report by the specific error that `try_push` return, so better to rely on that.

SuperFluffy (changes requested):
I believe the shutdown logic in the executor does not yet handle a submission being in flight.

What happens if you receive a shutdown signal while a bundle is in flight? You will likely end up with conflicts between the `submission_fut` and the shutdown loop.

crates/astria-composer/src/executor/mod.rs
bharath-123: you are right, i missed that

crates/astria-composer/src/executor/mod.rs
bharath-123: that doesn't seem right no? but how likely is that to happen?

crates/astria-composer/src/executor/mod.rs
SuperFluffy: Apologies, this should have read something like this:
```rust
let report = undrained_bundles.iter().map(SizedBundleReport);
```
and then report that ^

bharath-123: that works too

crates/astria-composer/src/composer.rs
```diff
@@ -232,26 +312,166 @@ impl Composer {
...
+
+        // if the composer is shutting down because of an unexpected shutdown from any one of the
+        // components(and not because of a SIGTERM), we need to send the cancel signal to all
+        // the other components.
+        composer_shutdown_token.cancel();
+        // k8s issues SIGKILL in 30s, so we need to make sure that the shutdown happens before 30s.
+
+        // We give executor 17 seconds to shut down. The logic to timeout is in the
+        // executor itself. We wait 17s for all the bundles to be drained.
+        if let Some(executor_task_handle) = executor_task_handle {
```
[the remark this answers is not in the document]
bharath-123: its provided in executor. its not a good idea to nest timeouts since they can be polled in different orders.

crates/astria-composer/src/collectors/geth.rs
```diff
@@ -153,39 +168,82 @@ impl Geth {
...
+            Err(reason) => {
+                error!(%reason, "shutting down");
+            }
+        };
+
+        info!("sending geth collector connection status to false");
+        status.send_modify(|status| status.is_connected = false);
+
+        info!("unsubscribing from geth tx stream");
+        match tx_stream.unsubscribe().await {
```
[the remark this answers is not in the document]
SuperFluffy: Coming back to this, please add the timeout explicitly. If the task gets killed because the server does not respond then we will not know why.

crates/astria-composer/src/composer.rs
```diff
@@ -1,21 +1,36 @@
 use std::{
+    borrow::BorrowMut,
```
SuperFluffy, line 2: This is very unusual and likely used wrong.

crates/astria-composer/src/composer.rs
```diff
@@ -1,21 +1,36 @@
...
     collections::HashMap,
     net::SocketAddr,
+    time::Duration,
 };
 
 use astria_eyre::eyre::{
     self,
     WrapErr as _,
 };
+use itertools::Itertools;
```
SuperFluffy, line 12:
```suggestion
use itertools::Itertools as _;
```

crates/astria-composer/src/composer.rs
```diff
@@ -168,39 +201,49 @@ impl Composer {
...
-        });
-        wait_for_executor(executor_status).await?;
-        composer_status_sender.send_modify(|status| {
-            status.set_executor_connected(true);
-        });
+        wait_for_collectors(
+            &geth_collector_statuses,
+            composer_status_sender.borrow_mut(),
+        )
+        .await?;
```
SuperFluffy, line 245: Please add a `.wrap_err before `?`

crates/astria-composer/src/composer.rs
```diff
@@ -168,39 +201,49 @@ impl Composer {
...
+        wait_for_executor(executor_status, composer_status_sender.borrow_mut()).await?;
```
SuperFluffy, line 246: `borrow_mut` is unnecessary. Also, please add error context

crates/astria-composer/src/composer.rs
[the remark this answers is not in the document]
SuperFluffy: Add an explicit timeout here.

crates/astria-composer/src/executor/mod.rs
```diff
@@ -276,11 +286,129 @@ impl Executor {
...
+                "unable to drain all bundles within the allocated time"
+            );
+        }
+
+        reason.map(|_| ())
     }
 }
 
-/// Queries the sequencer for the latest nonce with an exponential backoff
+/// Queries the seque ncer for the latest nonce with an exponential backoff
```
SuperFluffy, line 411: typo

crates/astria-composer/src/executor/mod.rs
```diff
@@ -276,11 +286,129 @@ impl Executor {
...
+            Ok(reason) => {
+                info!(reason, "shutting down");
+            }
+            Err(reason) => {
+                error!(%reason, "shutting down");
+            }
+        };
+
+        let mut total_bundles_drained_successfully = 0;
+        let shutdown_logic = async {
```
SuperFluffy, line 301: Please report how many bundles need to be drained before you enter this shutdown logic

crates/astria-composer/src/executor/mod.rs
SuperFluffy: It is not good enough. Please report the number of bundles that need to be submitted

crates/astria-composer/src/grpc.rs
SuperFluffy: for a follow-up

crates/astria-composer/src/composer.rs
```diff
@@ -232,26 +311,144 @@ impl Composer {
...
+            api_server_task_handle,
+            executor_task_handle,
+            grpc_server_task_handle,
+            mut geth_collector_tasks,
+        } = self;
+
+        // if the composer is shutting down because of an unexpected shutdown from any one of the
+        // components(and not because of a SIGTERM), we need to send the cancel signal to all
+        // the other components.
+        composer_shutdown_token.cancel();
```
itamarreif, line 346: this seems like the only spot where the cancel signal is sent.
this is fine for the collectors, but for executor it will only return to the main composer loop and trigger this shutdown after the executor's shutdown logic finished (seq actions and bundles all drained etc).
not sure if this is intended but figured i'd flag it here

itamarreif (commented):
looks good, only thing that stood out to me is that the cancellation token's signal is only sent out from the composer's shutdown. this means that if executor dies, the rest of the services will only shutdown after it finishes with its shutdown logic.
not sure what (if any) issues this causes, just seemed like weird behavior.

crates/astria-composer/src/composer.rs
bharath-123: its a good catch. I don't think its a critical issue because the executor would exit mainly due to failures while submitting to sequencer. The shutdown logic would anyways fail while bundle submission. so it wouldn't cause any problems but yeah its not desirable and we would have unecessary logs.

crates/astria-composer/src/executor/mod.rs
```diff
@@ -276,7 +287,125 @@ impl Executor {
...
+                    "failed to bundle sequence action, dropping it."
+                );
+            }
         }
+
+        // when shutting down, drain all the remaining bundles and submit to the sequencer
+        // to avoid any bundle loss.
+        loop {
+            let bundle = bundle_factory.pop_now();
+            if bundle.is_empty() {
```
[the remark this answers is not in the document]
bharath-123: don't think so :( cause pop_now() is not an iterable right?

crates/astria-composer/src/composer.rs
```diff
@@ -232,26 +311,144 @@ impl Composer {
...
+            }
+        } else {
+            info!("api server task was already dead");
+        };
+
+        Ok(())
+    }
+}
+
+fn spawn_geth_collectors(
```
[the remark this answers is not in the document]
bharath-123: this is mainly cause clippy was complaining that the composer `run_until_stopped` was too big. so i had to do this to slim it out.

crates/astria-composer/src/collectors/geth.rs
```diff
@@ -153,39 +173,79 @@ impl Geth {
             .await
             .wrap_err("failed to subscribe eth client to full pending transactions")?;
 
+        let send_timeout_duration = Duration::from_millis(500);
```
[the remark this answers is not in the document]
bharath-123: yeah sure will do that

crates/astria-composer/src/composer.rs
```diff
@@ -327,3 +529,11 @@ fn report_exit(task_name: &str, outcome: Result<eyre::Result<()>, JoinError>) {
         }
     }
 }
+
+pub(crate) fn flatten<T>(res: Result<eyre::Result<T>, JoinError>) -> eyre::Result<T> {
```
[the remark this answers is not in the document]
bharath-123: thats a good rename. flatten is also an array thing.

crates/astria-composer/src/executor/mod.rs
```diff
@@ -276,7 +287,125 @@ impl Executor {
...
+
+        match &reason {
+            Ok(reason) => {
+                info!(reason, "shutting down");
+            }
+            Err(reason) => {
+                error!(%reason, "shutting down");
+                // we error out mainly because of a failure to submit a bundle to the sequencer
+                // we do not want to proceed with the shutdown process in this case
+                return Err(eyre!(reason.to_string()));
```
bharath-123, line 300: @SuperFluffy what do you think of the way i m returning the error reason?

crates/astria-composer/src/collectors/mod.rs
```diff
@@ -1,5 +1,9 @@
 pub(crate) mod geth;
 pub(crate) mod grpc;
 
+use std::time::Duration;
+
 pub(crate) use geth::Geth;
 pub(crate) use grpc::Grpc;
+
+pub(super) const EXECUTOR_SEND_TIMEOUT: Duration = Duration::from_millis(500);
```
bharath-123, line 9: also is keeping constants here fine?

crates/astria-composer/src/collectors/grpc.rs
```diff
@@ -24,7 +21,10 @@ use tonic::{
     Status,
 };
 
-use crate::executor;
+use crate::{
+    collectors::geth::EXECUTOR_SEND_TIMEOUT,
```
SuperFluffy, line 25: Remove this. The grpc collector schould be entirely independent from what's going on in geth

crates/astria-composer/src/executor/mod.rs
bharath-123: Also note for reviewer: I am choosing to exit at this point rather than continuing, because its very unlikely we even reach this point because we retry almost infinitely. 
In a practical scenario, we would observe the retries and manually shutdown composer.

SuperFluffy: The return at this point (including skipping the drain) is fine - you know if it makes sense to attempt to drain or not.

But I don't see a point formatting the error at this point?. Just `return Err(reason)`

crates/astria-composer/src/executor/mod.rs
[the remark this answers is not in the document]
SuperFluffy: The reason is that `pop_now()` doesn't return an `Option` but just an empty vec, so you can't `while let Some(bundle) = bundle_factory.pop_now()`. I suppose we could change this, but other PR.

crates/astria-composer/src/executor/mod.rs
```diff
@@ -276,7 +287,122 @@ impl Executor {
...
+                    }
+                    Err(error) => {
+                        error!(%error, "failed submitting bundle to sequencer during shutdown; \
+                                aborting shutdown");
+                        return Err(error);
+                    }
+                }
+            }
+
+            info!(
```
SuperFluffy, line 356: Move this to the beginning of your `shutdown_logic` outside of it. If `submission_fut` never returns and the task is aborted you will never emit this event.

crates/astria-composer/src/executor/mod.rs
```diff
@@ -276,7 +287,122 @@ impl Executor {
...
+        self.status
+            .send_modify(|status| status.is_connected = false);
+
+        // close the channel to avoid receiving any other txs before we drain the remaining
+        // sequence actions
+        self.serialized_rollup_transactions.close();
+
+        match &reason {
+            Ok(reason) => {
+                info!(reason, "shutting down");
```
SuperFluffy, line 301:
```suggestion
                info!(reason, "starting shutdown process");
```

crates/astria-composer/src/executor/mod.rs
```diff
@@ -276,7 +287,122 @@ impl Executor {
...
+                // we error out because of a failure to submit a bundle to the sequencer
+                // we do not want to proceed with the shutdown process in this case
+                return Err(eyre!(reason.to_string()));
+            }
+        };
+
+        let mut bundles_to_drain: VecDeque<SizedBundle> = VecDeque::new();
+        let mut bundles_drained = 0;
+
+        info!("draining sequence actions from the executor receiver channel");
```
SuperFluffy, line 314: All of these words are implementation details that you don't need in logs.
```suggestion
        info!("draining already received transactions");
```

crates/astria-composer/src/executor/mod.rs
```diff
@@ -276,7 +287,122 @@ impl Executor {
...
+
+        // drain the receiver channel
+        while let Ok(seq_action) = self.serialized_rollup_transactions.try_recv() {
+            let rollup_id = seq_action.rollup_id;
+            if let Err(e) = bundle_factory.try_push(seq_action) {
+                warn!(
+                    rollup_id = %rollup_id,
+                    error = &e as &StdError,
+                    "failed to bundle sequence action, dropping it."
```
SuperFluffy, line 323:
```suggestion
                    "failed to bundle transaction, dropping it."
```

crates/astria-composer/src/executor/mod.rs
```diff
@@ -276,7 +287,122 @@ impl Executor {
...
+                break;
+            }
+
+            bundles_to_drain.push_back(bundle);
+        }
+
+        let shutdown_logic = async {
+            // wait for the last bundle to be submitted
+            if !submission_fut.is_terminated() {
+                info!("waiting for the last bundle to be submitted to the sequencer");
```
SuperFluffy, line 342:
```suggestion
                info!("waiting for the last bundle of transactions to be submitted to the sequencer");
```

crates/astria-composer/src/executor/mod.rs
```diff
@@ -276,7 +287,122 @@ impl Executor {
...
+                match submission_fut.await {
+                    Ok(new_nonce) => {
+                        debug!(new_nonce = new_nonce, "drained bundle successfully");
```
SuperFluffy, line 345:
```suggestion
                        debug!(new_nonce = new_nonce, "successfully submitted bundle of transactions");
```

crates/astria-composer/src/executor/mod.rs
```diff
@@ -276,7 +287,122 @@ impl Executor {
...
+                        error!(%error, "failed submitting bundle to sequencer during shutdown; \
+                                aborting shutdown");
+                        return Err(error);
+                    }
+                }
+            }
+
+            info!(
+                no_of_bundles_to_drain = bundles_to_drain.len(),
+                "draining remaining bundles from bundle factory"
```
SuperFluffy, line 358: don't mention `bundle factory`. It's an implementation detail that operators need not know.
```suggestion
                "submitting remaining transaction bundles to sequencer"
```

crates/astria-composer/src/executor/mod.rs
```diff
@@ -276,7 +287,122 @@ impl Executor {
...
+            );
+            while let Some(bundle) = bundles_to_drain.pop_front() {
+                match self.submit_bundle(nonce, bundle.clone()).await {
+                    Ok(new_nonce) => {
+                        debug!(
+                            bundle = %telemetry::display::json(&SizedBundleReport(&bundle)),
+                            new_nonce = new_nonce,
+                            "drained bundle successfully"
```
SuperFluffy, line 366:
```suggestion
                            "successfully submitted transction bundle"
```

crates/astria-composer/src/executor/mod.rs
```diff
@@ -276,7 +287,122 @@ impl Executor {
...
+                        error!(bundle = %telemetry::display::json(&SizedBundleReport(&bundle)),
+                                %error, "failed submitting bundle to sequencer during shutdown; \
+                                    aborting shutdown");
+                        // if we can't submit a bundle after multiple retries, we can abort
+                        // the shutdown process
+                        return Err(error);
+                    }
+                }
+            }
+
```
SuperFluffy, line 381:
```suggestion
        info!("shutdown target reached; shutting down");
```

crates/astria-composer/src/executor/mod.rs
```diff
@@ -276,7 +287,122 @@ impl Executor {
...
+            Ok(())
+        };
+
+        match tokio::time::timeout(BUNDLE_DRAINING_DURATION, shutdown_logic).await {
+            Ok(Ok(())) => info!("executor shutdown tasks completed successfully"),
+            Ok(Err(error)) => error!(%error, "executor shutdown tasks failed"),
+            Err(error) => error!(%error, "executor shutdown tasks failed to complete in time"),
+        }
+
+        info!(bundles_drained, "bundles drained during shutdown");
```
SuperFluffy, line 391: This even and the one below should probably be merged:

```rust
if !bundles_to_drain_is_empty() {
    let report = ...;
    warn!(
        number_of_bundles_submitted = bundles_drained,
        number_of_missing_bundles = report.len(),
        missing_bundles = %json(&report),
        "failed submitting all transaction to sequencer during shutdown",
    )
} else {
    info!(number_of_submitted_bundles = bundles_drained, "submitted all outstanding bundles to sequencer during shutdown");
}
```

crates/astria-composer/src/executor/mod.rs
```diff
@@ -79,6 +81,8 @@ mod bundle_factory;
 #[cfg(test)]
 mod tests;
 
+const BUNDLE_DRAINING_DURATION: Duration = Duration::from_secs(16);
```
SuperFluffy, line 84: Add a comment above here why 16 (-> because the higher level task has 17s).

SuperFluffy (approved):
Excellent, thank you very much for pushing this over the line.

I left a few comments/nits that would be great to be addressed prior to merge, but I don't see blockers.

Please have a pass over your emitted events: are you communicating implementation details? Or what's the high level information that's relevant to an operator. For example, an operator need not know what a "bundle factory" is (it's just a composer-internal concept). The same applies to e.g. the "executor task". On the other hand "bundle" is too imprecise - prefer "transaction bundle" or "bundle of rollup transactions".

PR #857, opened 2024-03-22
feat(core): add serialization to execution v1alpha2 compliant with protobuf json mapping

SuperFluffy:
## Summary
Uses pbjson to generate serde `Serialize` and `Deserialize` implementations compliant with the protobuf JSON mapping for all types defined or used by the execution v1alpha2 API.

## Background
It is convenient to have a standard for displaying our protobuf spec as JSON. This is useful for enforcing serialization standards across the codebase and will make maintenance easier. It is also useful for writing tests (for example, to use tools like `assert_json_diff`).

NOTE: the protobuf JSON mapping prescribes base64 as the byte encoding. Therefore this patch deviates from this crate's convention to use hex. Because this is an entirely cosmetic change this is not considered breaking (we communicate using protobuf bytes, not string containing hex-encoded bytes).

## Changes
- Update the compiler tool to emit code implementing `serde::Serialize` and `serde::Deserialize` (for now only for `astria.execution.v1alpha2` and imports that it uses, namely `astria.sequencer.v1.RollupData`, `astria.sequencer.v1.Deposit`, and `astria.primitive.v1.Uint128`)
- Replace `prost-types` by `pbjson-types`.
- Update the native types to delegate to the protobuf types when serializing to json.
- Change display formatting of `RollupId` to base64

## Testing
TODO: add insta snapshot tests?


crates/astria-conductor/src/sequencer/snapshots/astria_conductor__sequencer__reporting__tests__snapshots.snap
```diff
@@ -3,6 +3,6 @@ source: crates/astria-conductor/src/sequencer/reporting.rs
 expression: ReportRollups(block.rollup_transactions())
 ---
 {
-  "2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a": 2,
-  "4545454545454545454545454545454545454545454545454545454545454545": 1
+  "KioqKioqKioqKioqKioqKioqKioqKioqKioqKioqKio=": 2,
```
SuperFluffy, line 6: This snapshot is updated because the display-formatting of `RollupId` changed from hex to base64 to be compliant with protobuf <> JSON mapping (and because that's just what `pbjson` gives us)

noot (approved):

PR #858, opened 2024-03-25
test(sequencer): add unit tests for src/state_ext file

Lilyjjo:
## Summary
First set of state-ext unit tests. In total there are seven files which need tests:
- `astria-sequencer/src/api_state_ext.rs`
- `astria-sequencer/src/state_ext.rs`
- `astria-sequencer/src/accounts/state_ext.rs`
- `astria-sequencer/src/asset/state_ext.rs`
- `astria-sequencer/src/authority/state_ext.rs`
- `astria-sequencer/src/bridge/state_ext.rs`
- `astria-sequencer/src/ibc/state_ext.rs`

This PR just tests the `src/state_ext.rs` file. I will add more commits / PRs for the others after getting feedback on this one.

## Background
It is good to have unit tests to ensure that the database works as intended. 

## Changes
Unit tests for the functionality in the file `src/state_ext.rs` were added.

## Testing
:)

### DIFF
--- a/crates/astria-sequencer/src/state_ext.rs
+++ b/crates/astria-sequencer/src/state_ext.rs
@@ -322,12 +322,12 @@ mod test {
         assert!(state.get_block_timestamp().await.is_err());
 
         // can write new
-        let block_timestamp_orig = Time::from_unix_timestamp(1577836800, 0).unwrap();
+        let block_timestamp_orig = Time::from_unix_timestamp(1_577_836_800, 0).unwrap();
         state.put_block_timestamp(block_timestamp_orig);
         assert!(state.get_block_timestamp().await.unwrap() == block_timestamp_orig);
 
         // can rewrite with new value
-        let block_timestamp_update = Time::from_unix_timestamp(1577836801, 0).unwrap();
+        let block_timestamp_update = Time::from_unix_timestamp(1_577_836_801, 0).unwrap();
         state.put_block_timestamp(block_timestamp_update);
         assert!(state.get_block_timestamp().await.unwrap() == block_timestamp_update);
     }
@@ -417,18 +417,15 @@ mod test {
 
         // doesn't exist at first
         let fee_balances_orig = state.get_block_fees().await.unwrap();
-        assert!(fee_balances_orig.len() == 0);
+        assert!(fee_balances_orig.is_empty());
 
         // can write
         let asset = astria_core::sequencer::v1::asset::Id::from_denom("asset_0");
         let amount = 100u128;
-        assert!(
-            state
-                .get_and_increase_block_fees(asset, amount)
-                .await
-                .unwrap()
-                == ()
-        );
+        state
+            .get_and_increase_block_fees(asset, amount)
+            .await
+            .unwrap();
 
         // holds expected
         let fee_balances_updated = state.get_block_fees().await.unwrap();
@@ -448,21 +445,15 @@ mod test {
         let asset_second = astria_core::sequencer::v1::asset::Id::from_denom("asset_1");
         let amount_first = 100u128;
         let amount_second = 200u128;
-        assert!(
-            state
-                .get_and_increase_block_fees(asset_first, amount_first)
-                .await
-                .unwrap()
-                == ()
-        );
-        assert!(
-            state
-                .get_and_increase_block_fees(asset_second, amount_second)
-                .await
-                .unwrap()
-                == ()
-        );
 
+        state
+            .get_and_increase_block_fees(asset_first, amount_first)
+            .await
+            .unwrap();
+        state
+            .get_and_increase_block_fees(asset_second, amount_second)
+            .await
+            .unwrap();
         // holds expected
         let fee_balances = state.get_block_fees().await.unwrap();
         for val in fee_balances {
@@ -473,7 +464,7 @@ mod test {
         state.clear_block_fees().await;
 
         let fee_balances_updated = state.get_block_fees().await.unwrap();
-        assert!(fee_balances_updated.len() == 0);
+        assert!(fee_balances_updated.is_empty());
     }
 
     #[tokio::test]
@@ -515,7 +506,7 @@ mod test {
 
         // see is deleted
         let assets = state.get_allowed_fee_assets().await.unwrap();
-        assert!(assets.len() == 0);
+        assert!(assets.is_empty());
     }
 
     #[tokio::test]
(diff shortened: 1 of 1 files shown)


crates/astria-sequencer/src/state_ext.rs
```diff
@@ -243,3 +243,298 @@ pub(crate) trait StateWriteExt: StateWrite {
...
+    use super::{
+        StateReadExt as _,
+        StateWriteExt as _,
+    };
+
+    #[tokio::test]
+    async fn chain_id() {
+        let storage = cnidarium::TempStorage::new()
+            .await
+            .expect("failed to create temp storage backing chain state");
```
SuperFluffy, line 261: High level observation on this usage of `expect` (applies to all instances below here as well):

1. the `expect` message should contain what you *expected* to happen, not the opposite, see this: https://doc.rust-lang.org/std/error/index.html#common-message-styles
2. I think in this case your intent is completely clear - you just want there to be a temp storage, so `unwrap()`ing is ok

crates/astria-sequencer/src/state_ext.rs
```diff
@@ -243,3 +243,298 @@ pub(crate) trait StateWriteExt: StateWrite {
...
+        let snapshot = storage.latest_snapshot();
+        let mut state = StateDelta::new(snapshot);
+
+        // doesn't exist at first
+        assert!(state.get_chain_id().await.is_err());
```
SuperFluffy, line 266: I just using `expect_err("no chain ID should exist at first")` over `assert!` together with `Result::is_ok` or `Result::is_err`. In case the test is actually wrong the `is_*` methods just paper over what was actually inside (and that might be valuable).

crates/astria-sequencer/src/state_ext.rs
```diff
@@ -243,3 +243,298 @@ pub(crate) trait StateWriteExt: StateWrite {
...
+
+        // can write new
+        let chain_id_orig = "test-chain-orig";
+        state.put_chain_id(chain_id_orig.to_string());
+        assert!(state.get_chain_id().await.unwrap() == chain_id_orig);
```
SuperFluffy, line 271:
```suggestion
        assert_eq!(state.get_chain_id().await.unwrap(), chain_id_orig);
```
This will give you better diagnostics in case smth went wrong.

Also, this might be a good use `expect("a chain ID was written and must exist in the database")`

crates/astria-sequencer/src/state_ext.rs
```diff
@@ -243,3 +243,298 @@ pub(crate) trait StateWriteExt: StateWrite {
...
+
+        // can rewrite with new value
+        let chain_id_update = "test-chain-update";
+        state.put_chain_id(chain_id_update.to_string());
+        assert!(state.get_chain_id().await.unwrap() == chain_id_update);
```
SuperFluffy, line 276: same thing re `expect` and `assert_eq!`

crates/astria-sequencer/src/state_ext.rs
```diff
@@ -243,3 +243,298 @@ pub(crate) trait StateWriteExt: StateWrite {
...
+    #[tokio::test]
+    async fn revision_number() {
+        let storage = cnidarium::TempStorage::new()
+            .await
+            .expect("failed to create temp storage backing chain state");
+        let snapshot = storage.latest_snapshot();
+        let state = StateDelta::new(snapshot);
+
+        // current impl just returns 'ok'
+        assert!(state.get_revision_number().await.unwrap() == 0u64);
```
SuperFluffy, line 288: `assert_eq!`

SuperFluffy (approved):
The test in general look great, but please make use of `assert_eq!` and more expressive `expect("msg")` lines.

Also, remember that `assert!` and `assert_eq` also allow with a format-string, so you could do:
```rust
assert!(assets.is_empty(), "all assets should have been deleted from the database");
```

crates/astria-sequencer/src/state_ext.rs
noot: this is my bad as i put this in the other unit tests originally - @Lilyjjo feel free to replace all the `expect`s on these lines with `unwrap`

crates/astria-sequencer/src/state_ext.rs
```diff
@@ -243,3 +243,298 @@ pub(crate) trait StateWriteExt: StateWrite {
...
+            .unwrap();
+        state
+            .get_and_increase_block_fees(asset_second, amount_second)
+            .await
+            .unwrap();
+        // holds expected
+        let fee_balances = state.get_block_fees().await.unwrap();
+        for val in fee_balances {
+            assert!(val == (asset_first, amount_first) || val == (asset_second, amount_second));
+        }
```
noot, line 461: i think you can just construct the vec and compare directly: `vec![(asset_first, amount_first), (asset_second, amount_second)]`

crates/astria-sequencer/src/state_ext.rs
```diff
@@ -243,3 +243,298 @@ pub(crate) trait StateWriteExt: StateWrite {
...
+        assert!(state.is_allowed_fee_asset(asset_third).await.unwrap());
+
+        // can delete
+        state.delete_allowed_fee_asset(asset_second);
+
+        // see is deleted
+        let assets = state.get_allowed_fee_assets().await.unwrap();
+        for asset in assets {
+            assert!(asset == asset_first || asset == asset_third);
+        }
```
noot, line 538: same here, just compare against `vec![asset_first, asset_third]`

crates/astria-sequencer/src/state_ext.rs
```diff
@@ -243,3 +243,298 @@ pub(crate) trait StateWriteExt: StateWrite {
...
+        let asset = astria_core::sequencer::v1::asset::Id::from_denom("asset_0");
+        state.put_allowed_fee_asset(asset);
+        assert!(state.is_allowed_fee_asset(asset).await.unwrap());
+
+        // see can get fee asset
+        let assets = state.get_allowed_fee_assets().await.unwrap();
+        assert!(assets[0] == asset);
+
+        // can delete
+        state.delete_allowed_fee_asset(asset);
```
noot, line 505: check what happens if you delete a fee asset that wasn't stored - should error

Lilyjjo: Right now `delete_allowed_fee_asset()` doesn't ensure that the fee asset exists before deleting. Do you want me to change the behavior to err in this case? I can make a follow up PR for it

Lilyjjo:
Refactored code to reflect all comments by @SuperFluffy and all but one of @noot's.

crates/astria-sequencer/src/state_ext.rs
noot: you can leave as-is, thinking about it it doesn't really make a difference if this errors or is a no-op

noot (approved):
looks great!!

PR #861, opened 2024-03-25
ci: chart release setup

joroshiba:
## Summary
Updates chart release flow to promote any chart changes to the astriaorg/charts repo, which then manages chart release.

## Background
Charts were migrated into the monorepo to ease our development flow, but this doesn't solve our release flow. By copying changes out externally and then promoting chart releases there, we can keep chart releases external, provide a clear view for people only interested in our charts, but maintain the easy development flow and smoke test from from within monorepo.

## Changes
- Replaces in repo chart release to pushing code to external repo.

## Testing
Tested an initial release flow by commenting out if statement.


aajimal (approved):
👍

noot (approved):

PR #863, opened 2024-03-26
fix(sequencer): fix rollup ids commitment for deposits

noot:
## Summary
fix generation of rollup ids commitment for deposits

## Background
this was wrong, figured it out while testing. it wasn't including the deposit rollup id in the commitment in prepare_proposal, but it was in end_block, causing a block failure.

## Changes
- fix generation of rollup ids commitment for deposits

## Testing
end to end bridge testing


crates/astria-sequencer/src/proposal/commitment.rs
```diff
@@ -67,6 +66,7 @@ pub(crate) fn generate_rollup_datas_commitment(
         );
     }
 
+    let rollup_ids_root = merkle::Tree::from_leaves(rollup_ids_to_txs.keys()).root();
     rollup_ids_to_txs.sort_unstable_keys();
```
SuperFluffy, line 70: Shouldn't the sorting go above the merkle tree generation her? You are potentially inserting new rollup IDs in the previous loop `(rollup_ids_to_txs.entry(rollup_id).or_default())`, so the order of this object is not guaranteed to be stable.

We should provide a unit test to ensure this.

noot: yeah good point, coding at midnight got me missing. will fix!!

SuperFluffy (approved):
Approving to push this along, but I have created https://github.com/astriaorg/astria/issues/870 to get this unit tested.

PR #864, opened 2024-03-26
feat(sequencer): add display for deposits in `end_block`

noot:
## Summary
add display for deposits in `end_block` and minor logging fixes

## Background
it's nice for testing and debugging :)

## Changes
- add display for deposits in `end_block` and minor logging fixes


crates/astria-sequencer/src/app.rs
```diff
@@ -733,6 +734,28 @@ impl App {
     }
 }
 
+/// Display wrapper for `HashMap<RollupId, Vec<Deposit>>`.
+struct DisplayDeposits<'a>(&'a HashMap<RollupId, Vec<Deposit>>);
+
+impl<'a> std::fmt::Display for DisplayDeposits<'a> {
```
SuperFluffy, line 740: When writing a `Display` impl for these complex types, please ensure that they are done using a json formatting. That makes analyzing them easier (once grafana etc is working).

I suggest deriving `serde::Serialize` with your preferred way of formatting the map. Alternatively, impl your own `serde::Serialize` for your reporting newtype wrapper, and then use `astria_telemetry::display::json` to emit the JSON in tracing.

crates/astria-conductor/src/sequencer/mod.rs
```diff
@@ -129,7 +128,7 @@ impl Reader {
                             }
 
                             executor::channel::TrySendError::NoPermits(block) => {
-                                trace!("executor channel is full; scheduling block and stopping block fetch until a slot opens up");
+                                debug!("executor channel is full; scheduling block and stopping block fetch until a slot opens up");
```
SuperFluffy, line 131: This was intentionally kept at trace-level because it's pretty low level because right now this can trigger extremely frequently == create lots of noise.

crates/astria-conductor/src/sequencer/block_stream.rs
```diff
@@ -115,7 +115,7 @@ impl BlocksFromHeightStream {
     )]
     pub(super) fn set_latest_observed_height_if_greater(&mut self, height: Height) {
         if !self.heights.set_latest_observed_if_greater(height) {
-            info!("observed latest sequencer height older than previous; ignoring it");
+            info!("observed latest sequencer height older or the same as previous; ignoring it");
```
SuperFluffy, line 118: That change makes sense, but is orthogonal to the aim of the PR.

SuperFluffy (changes requested):
The reporting in sequencer makes sense, but please try to make it valid JSON. Also provide a snapshot test for it.

Please move the changes to conductor to a separate PR.

crates/astria-conductor/src/sequencer/mod.rs
noot: removed

noot:
@SuperFluffy updated to format as json!

crates/astria-sequencer/src/app.rs
```diff
@@ -639,6 +639,10 @@ impl App {
             .get_block_deposits()
             .await
             .context("failed to get block deposits in end_block")?;
+        debug!(
+            deposits = %telemetry::display::json(&DisplayDeposits(&deposits)),
+            "end_block: got block deposits"
```
SuperFluffy, line 644: The `end_block:` is not necessary. You already running under a scope called `App::end_block`
```suggestion
            "read block deposits from state"
```

crates/astria-sequencer/src/app.rs
```diff
@@ -733,6 +737,10 @@ impl App {
     }
 }
 
+/// Display wrapper for `HashMap<RollupId, Vec<Deposit>>`.
+#[derive(Debug, serde::Serialize)]
+struct DisplayDeposits<'a>(&'a HashMap<RollupId, Vec<Deposit>>);
```
SuperFluffy, line 742: This is extra wrapper is not necessary since serde implements `Serialize for HashMap`.

However, this newtype wrapper might be useful in case you also add snapshots tests using inta (see for example composer and conductor which employ them).

crates/astria-sequencer/src/app.rs
```diff
@@ -639,6 +639,10 @@ impl App {
             .get_block_deposits()
             .await
             .context("failed to get block deposits in end_block")?;
+        debug!(
+            deposits = %telemetry::display::json(&DisplayDeposits(&deposits)),
```
SuperFluffy, line 643: mentioned below, just do:
```suggestion
            deposits = %telemetry::display::json(&deposits),
```

crates/astria-core/src/sequencer/v1/block/mod.rs
```diff
@@ -1,6 +1,8 @@
 use std::collections::HashMap;
 
 use indexmap::IndexMap;
+#[cfg(feature = "serde")]
+use serde::Serialize;
```
SuperFluffy, line 5: remove this import. You are using the full path in the `cfg_attr` below

crates/astria-core/src/sequencer/v1/asset.rs
```diff
@@ -119,6 +122,7 @@ impl From<String> for Denom {
 
 /// Asset ID, which is the hash of the denomination trace.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize))]
```
SuperFluffy, line 125: This is very likely not going to do what you want. Have a look at `RollupId`. `astria_core::serde` provides a utility for deriving a serde impl that serialises bytes as base64

crates/astria-core/src/sequencer/v1/block/mod.rs
```diff
@@ -1239,6 +1241,7 @@ impl FilteredSequencerBlockError {
 /// A [`Deposit`] is constructed whenever a [`BridgeLockAction`] is executed
 /// and stored as part of the block's events.
 #[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize))]
```
SuperFluffy, line 1244: This should serialize exactly the same way as the protobuf. Use:
```rust
#[cfg_attr(feature = "serde", derive(serde::Serialize))]
#[cfg_attr(feature = "serde", serde(into = "::core::generated::sequencer::v1::Deposit"))]
```
Also update the compiler tool to generate code for the `Deposity` type and add a `impl From<Deposit> for ::core::generated::sequencer::v1::Deposit`

SuperFluffy (changes requested):

crates/astria-core/src/sequencer/v1/mod.rs
```diff
@@ -43,6 +43,7 @@ pub const ADDRESS_LEN: usize = 20;
 pub const ROLLUP_ID_LEN: usize = 32;
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize))]
 pub struct Address([u8; ADDRESS_LEN]);
```
Fraser999, line 47: Do we not want `#[cfg_attr(feature = "serde", serde(serialize_with = "crate::serde::base64"))]` here too?

SuperFluffy (approved):
Approving because this looks most mostly good, but before you merge, please do the following:

1. use `base64` to derive the serialization format for `Address` (as flagged by @Fraser999)
2. please add snapshot tests for `Address` using insta. See for example in astria-conductor how that's done: https://github.com/astriaorg/astria/blob/6665948892b39a583e0cc623d1355954328bebff/crates/astria-conductor/src/sequencer/reporting.rs#L83

PR #865, opened 2024-03-26
fix(composer): Make snapshot testing in composer deterministic

bharath-123:
## Summary
Make the Composer bundle factory snapshot testing more deterministic by sorting the `rollup_counts` map before testing it against the generated json.

## Background
From https://github.com/astriaorg/astria/pull/808, we started tracking the number of sequence actions per rollup id in a bundle via a hashmap. We are testing this feature using insta snapshot testing where we store snapshot outputs in a `.snap` file and ensure that the output matches the expected output in the `.snap` file. 

The `snapshots` test in `bundle_factory` tends to be flaky due to the non-determinism of Hashmaps.

To fix it, we enforce sorting of maps as part of insta snapshot settings.

## Changes
- Before running the test, we update the snapshot test settings and `set_sort_maps` to true and run the snapshot assertion in the context of this test.
- We update the `.snap` file to contain the json sorted lexicographically by keys which is the default way a HashMap with a String key is sorted.

## Testing
Ran `cargo test` in `astria-composer` crate 10 times to ensure there is no flakiness.


SuperFluffy (approved):
NIce, thank you for fixing this. I will have to pay better attention to this.

PR #868, opened 2024-03-26
fix(conductor): clarify conductor log

noot:
## Summary
fix wording in a log in conductor.

## Changes
- clarify conductor log

### DIFF
--- a/crates/astria-conductor/src/sequencer/block_stream.rs
+++ b/crates/astria-conductor/src/sequencer/block_stream.rs
@@ -115,7 +115,7 @@ impl BlocksFromHeightStream {
     )]
     pub(super) fn set_latest_observed_height_if_greater(&mut self, height: Height) {
         if !self.heights.set_latest_observed_if_greater(height) {
-            info!("observed latest sequencer height older than previous; ignoring it");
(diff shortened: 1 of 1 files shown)


crates/astria-conductor/src/sequencer/block_stream.rs
```diff
@@ -115,7 +115,7 @@ impl BlocksFromHeightStream {
     )]
     pub(super) fn set_latest_observed_height_if_greater(&mut self, height: Height) {
         if !self.heights.set_latest_observed_if_greater(height) {
-            info!("observed latest sequencer height older than previous; ignoring it");
+            info!("observed latest sequencer height older or the same as previous; ignoring it");
```
SuperFluffy, line 118: I am wondering if this should be `debug!` instead of `info!`. On the other hand, it is arguably one of the core purposes of conductor to observe (and hence) the current height of sequencer.

SuperFluffy (approved):
Thank you for fixing this. This log was indeed confusing.

crates/astria-conductor/src/sequencer/block_stream.rs
noot: i think info is fine, if we see it polluting we can change it but i don't see it too often

PR #869, opened 2024-03-26
feat(sequencer): check for sufficient balance in `check_tx`

### LINKED ISSUE #855: sequencer: add balance check in stateful check for `check_tx`
currently, we don't check sufficient balance for fee payment in the stateful `check_tx` check. however, we want to check this to avoid having excessive invalid txs in the mempool.

noot:
## Summary
check for sufficient balance for all actions/assets in `check_tx`

## Background
don't want the mempool to be filled with invalid txs

## Changes
- check for sufficient balance for all actions/assets in `check_tx`. sums the amount that will be used by each action in the tx by their asset ID and checks for sufficient balance for all asset IDs

## Testing
unit tests

## Related Issues

closes  #855


crates/astria-core/src/sequencer/v1/abci.rs
```diff
@@ -17,6 +17,7 @@ impl AbciErrorCode {
     pub const INTERNAL_ERROR: Self = Self(3);
     pub const INVALID_NONCE: Self = Self(4);
     pub const TRANSACTION_TOO_LARGE: Self = Self(5);
+    pub const INSUFFICIENT_FUNDS: Self = Self(6);
```
joroshiba, line 20: Should there be a matching ABCI error code added here? https://github.com/astriaorg/astria/blob/main/proto/sequencerapis/astria/sequencer/v1/abci.proto#L5

noot: turns out these are unused, so i removed them

noot: ugh this is going to break the protos in ci even though these were unused :/ let me know if that's ok

crates/astria-sequencer/src/transaction/mod.rs
```diff
@@ -309,3 +388,121 @@ impl ActionHandler for UnsignedTransaction {
...
+        };
+
+        let signed_tx = tx.into_signed(&alice_signing_key);
+        check_balance_mempool(&signed_tx, &state_tx)
+            .await
+            .expect("sufficient balance for all actions");
+    }
+
+    #[tokio::test]
+    async fn check_balance_mempool_insuffient_other_asset_balance() {
```
Fraser999, line 463:
```suggestion
    async fn check_balance_mempool_insufficient_other_asset_balance() {
```

crates/astria-sequencer/src/transaction/mod.rs
```diff
@@ -309,3 +388,121 @@ impl ActionHandler for UnsignedTransaction {
...
+
+        let tx = UnsignedTransaction {
+            nonce: 0,
+            actions,
+        };
+
+        let signed_tx = tx.into_signed(&alice_signing_key);
+        check_balance_mempool(&signed_tx, &state_tx)
+            .await
+            .expect_err("insuffient funds for `other` asset");
```
Fraser999, line 506:
```suggestion
            .expect_err("insufficient funds for `other` asset");
```
Should we also be checking that the error contains the expected string to ensure it was really `other`'s ID mentioned as being underfunded.

Fraser999 (approved):
Just a few non-blocking nitpicks, otherwise LGTM.

PR #872, opened 2024-03-27
feat: add gRPC mocking library

SuperFluffy:
## Summary
Adds a library for mocking gRPC servers heavily inspired by wiremock-rs.

## Background
conductor, sequencer-relayer, and soon composer are interacting with gRPC servers. In blackbox tests these services use ad-hoc implementations of gRPC servers with manually crafted internal state to return desired responses. This is very tedious and error prone work, and moving forward not sustainable.

Specifically conductor is lacking blackbox tests entirely. Implementing blackbox tests for it is the motivation for `astria-grpc-mock`.

This library is heavily inspired by https://github.com/LukeMathWalker/wiremock-rs. Deviating from it, `astria-grpc-mock::MockServer` does not stand in as a full server in its own right, but is run next to a `tonic::Server`. Implementors of a tonic-generated service then delegate individual trait methods to `MockServer::handle_request`. See the tests in `astria-grpc-mock-test` how this is done.

The other big difference is that we are heavily leaning on `std::any::Any` to erase the types of concrete gRPC requests and responses, i.e. for functions of the form `rpc: T -> U` both `T` and `U` are erased. Because a mock author needs to provide an RPC respsonse `V`, `astria-grpc-mock` panics if `U` and `V` don't match, failing the test. This check happens if the `MockServer` goes out of scope, if a `MockGuard` goes out of scope, or in the `MockGuard::wait_until_satisfied` future (different from `wiremock-rs`, where `MockGuard::wait_until_satisfied` cannot panic).

Documentation and ergonomics of this library will be improved in follow-up PRs (an idea is to use a proc macro to automatically generate a desired mock implementation for a server, but the details around how to this are complex).

**Note to the reviewer:** The best way to review this library is to see how it is used in `astria-grpc-mock-test` and move backward from there.

## Changes
Added the following libraries:
- `astria-grpc-mock` to write grpc server mocks
- `astria-grpc-mock-test` to test the public interface of `astria-grpc-mock`
- `astria-grpc-mock-codegen` to generate Rust bindings with pbjson `Serialize` impls for the google `grpc.health.v1` package (this taken as is from `tonic-health`)

## Testing
The `astria-grpc-mock-test` crate tests the `astria-grpc-mock` crate. No other tests are necessary as this crate does not show up in release builds of any service.


crates/astria-grpc-mock-test/proto/health.proto
SuperFluffy, line 1: this does not need review. It was copied verbatim from `tonic-health`

crates/astria-grpc-mock-test/src/generated/grpc.health.v1.rs
SuperFluffy, line 1: does not need review; generated source code

crates/astria-grpc-mock-test/src/generated/grpc.health.v1.serde.rs
SuperFluffy, line 1: does not need review; generated source code

crates/astria-grpc-mock/src/mock.rs
```diff
@@ -0,0 +1,147 @@
...
+            matchers,
+            response: Box::new(rsp),
+            name: None,
+            expectation_range: Times(TimesEnum::Unbounded(RangeFull)),
+        }
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct Times(TimesEnum);
```
SuperFluffy, line 85: This struct is copied verbatim from wiremock.

crates/astria-grpc-mock/src/lib.rs
```diff
@@ -0,0 +1,119 @@
...
+    tonic::Response::from_parts(metadata, boxed, extensions)
+}
+
+fn clone_request<T: Clone>(req: &tonic::Request<T>) -> tonic::Request<T> {
+    let mut req = tonic::Request::new(req.get_ref().clone());
+    *req.metadata_mut() = req.metadata().clone();
+    req
+}
+
+fn clone_response<T: Clone>(req: &tonic::Response<T>) -> tonic::Response<T> {
```
Fraser999, line 115: `req` should probably be `resp` here?

crates/astria-grpc-mock/src/mounted_mock.rs
```diff
@@ -0,0 +1,209 @@
...
+                        .push((clone_request(request), Ok(mock_response)));
+                    Ok(Ok(rsp))
+                } else {
+                    let actual = mock_response.type_name;
+                    self.bad_responses
+                        .push((clone_request(request), mock_response).into());
+                    let expected = std::any::type_name::<U>();
+                    #[rustfmt::skip]
+                        let msg = format!(
+                            "failed downcasting mock response to conrete type:\n\
```
Fraser999, line 153:
```suggestion
                            "failed downcasting mock response to concrete type:\n\
```

crates/astria-grpc-mock-test/tests/health/main.rs
Fraser999, line 1: We could maybe have a slightly more complex request and response with more than a single field each, but not a big deal (I think you already called this out @SuperFluffy).

We should maybe also use `Mock::with_name` in all the tests in case they're used as a reference point for others using the mock crate (I think the usage of that is recommended in wiremock).

Also, it might be nice to have a test which exercises the `MockBuilder::and` method, but again, not a big deal.

Fraser999 (commented):
Generally looks good to me - nice one!

The only drawback I see so far is that metadata and extensions on requests and responses are effectively ignored, but that's maybe not a real problem?

crates/astria-grpc-mock/src/mounted_mock.rs
```diff
@@ -0,0 +1,209 @@
...
+///
+/// Only used to implement `Clone` because [`Request`] doesn't.
+pub(crate) struct BadResponse {
+    pub(crate) request: Request<AnyMessage>,
+    pub(crate) mock_response: MockResponse,
+}
+
+impl BadResponse {
+    pub(crate) fn print(&self, mut buffer: impl std::fmt::Write, indent: &str) -> std::fmt::Result {
+        // TODO: Print the metadata map
```
noot, line 40: remove or implement?

SuperFluffy: Leave for now. Gonna implement later because more work. I am going to make an issue.

crates/astria-grpc-mock/src/mock.rs
```diff
@@ -0,0 +1,147 @@
+use std::ops::{
```
noot, line 1: would be nice to have a small doc comment on usage (although looks similar to wiremock?)

crates/astria-grpc-mock/src/lib.rs
```diff
@@ -0,0 +1,119 @@
...
+    T: Clone + erased_serde::Serialize + prost::Name + Send + Sync + 'static,
+{
+    let (metadata, message, extensions) = rsp.into_parts();
+    let boxed = Box::new(message) as AnyMessage;
+    tonic::Response::from_parts(metadata, boxed, extensions)
+}
+
+fn clone_request<T: Clone>(req: &tonic::Request<T>) -> tonic::Request<T> {
+    let mut req = tonic::Request::new(req.get_ref().clone());
+    *req.metadata_mut() = req.metadata().clone();
```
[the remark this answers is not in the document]
SuperFluffy: Nice catch, my propensity for shadowing variables caught me there!

crates/astria-grpc-mock/src/lib.rs
```diff
@@ -0,0 +1,119 @@
...
+    req
```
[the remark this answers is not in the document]
SuperFluffy: I contemplated this as well but I don't think this will be an issue here: extensions are used to enrich requests and responses in middlewares to gain access to server-state specific objects, for example injecting a database handle into a requests to use it downstream.

So these are very much specific to a server implementation, which for these mocks we are agnostic to.

crates/astria-grpc-mock/src/verification.rs
```diff
@@ -0,0 +1,78 @@
...
+                self.position_in_set, self.rpc, self.expectation_range, self.n_successful_requests, self.bad_responses.len(),
+            )
+        };
+
+        if !self.bad_responses.is_empty() {
+            let _ = writeln!(msg, "Bad responses:");
+            msg = self.bad_responses.iter().enumerate().fold(
+                msg,
+                |mut msg, (index, response)| {
+                    _ = writeln!(msg, "\t - Bad response #{}", index+1,);
```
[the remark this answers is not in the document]
SuperFluffy: I guess I followed wiremock's example to enumerate the requests by one ("first request") for readers. But I agree - it's surprising that these have different indices. I changed it to use zero-indices it here in `verification.rs`, but also in the mock server itself to report the actually received requests.

crates/astria-grpc-mock/src/mock_server.rs
```diff
@@ -0,0 +1,255 @@
...
+            mock_id,
+            server_state: self.state.clone(),
+        }
+    }
+
+    pub async fn verify(&self) {
+        debug!("verifying mock expectations");
+        if let VerificationOutcome::Failure(failed_verifications) = self.state.read().await.verify()
+        {
+            let received_requests_message =
```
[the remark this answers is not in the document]
SuperFluffy: I have moved the construction of the received requests into a separate function. The rest is bespoke.

crates/astria-grpc-mock/src/mock_server.rs
```diff
@@ -0,0 +1,255 @@
...
+        notification.await;
+    }
+}
+
+struct MockRequest {
+    inner: AnyRequest,
+}
+
+impl MockRequest {
+    fn print(&self, mut buffer: impl std::fmt::Write) -> std::fmt::Result {
```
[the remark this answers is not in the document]
SuperFluffy: wiremock has a limiter for the number of chars each report should print. I have removed that config for this PR but I am inclined to bring it back later (because they have probably arrived there for a good reason). It doesn't match very well with `Display`.

crates/astria-grpc-mock/src/verification.rs
```diff
@@ -0,0 +1,78 @@
...
+                    _ = response.print(&mut msg, "\t");
```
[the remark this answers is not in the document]
SuperFluffy: that is very nice, thank you! I was looking for something like that but then gave up.

crates/astria-grpc-mock/src/mock_set.rs
```diff
@@ -0,0 +1,143 @@
...
+                continue;
+            }
+            match mock.match_and_respond::<U>(rpc, &erased) {
+                MockResult::NoMatch => continue,
+                MockResult::BadResponse(status) => {
+                    mock_response.replace(Err(status));
+                }
+                MockResult::Success(response) => {
+                    mock_response.replace(response);
+                }
```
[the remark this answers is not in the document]
SuperFluffy: I think this was my oversight after a refactor. Eagerly breaking from the loop now.

crates/astria-grpc-mock/src/lib.rs
```diff
@@ -0,0 +1,119 @@
...
+mod mock;
+mod mock_server;
+mod mock_set;
+mod mounted_mock;
+pub mod response;
+mod verification;
+
+pub use mock::Mock;
+pub use mock_server::MockServer;
+
```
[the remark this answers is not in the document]
SuperFluffy: Good hunch, I actually needed all of these when writing conductor tests, so I now they are exported.

crates/astria-grpc-mock/src/mounted_mock.rs
SuperFluffy: Added https://github.com/astriaorg/astria/issues/900 to the TODO comment

crates/astria-grpc-mock/src/mock.rs
SuperFluffy: Will do, but leave it to a follow up. https://github.com/astriaorg/astria/issues/901

crates/astria-grpc-mock-test/tests/health/main.rs
SuperFluffy: I have created https://github.com/astriaorg/astria/issues/902 and https://github.com/astriaorg/astria/issues/901 as follow-ups.

crates/astria-grpc-mock-test/proto/health.proto
```diff
@@ -0,0 +1,63 @@
...
+
+syntax = "proto3";
+
+package grpc.health.v1;
+
+option csharp_namespace = "Grpc.Health.V1";
+option go_package = "google.golang.org/grpc/health/grpc_health_v1";
+option java_multiple_files = true;
+option java_outer_classname = "HealthProto";
+option java_package = "io.grpc.health.v1";
```
noot, line 26: java o.o

crates/astria-grpc-mock-test/tests/health/main.rs
```diff
@@ -0,0 +1,270 @@
...
+        mocked: mock_server,
+    }
+}
+
+struct HealthService {
+    mock_server: astria_grpc_mock::MockServer,
+}
+
+#[tonic::async_trait]
+impl Health for HealthService {
```
noot, line 72: what's the purpose of the health service? is it just for testing the mock server?

crates/astria-grpc-mock/src/mock_set.rs
```diff
@@ -0,0 +1,145 @@
...
+
+    pub(crate) fn handle_request<
+        T: erased_serde::Serialize + prost::Name + Clone + Send + Sync + 'static,
+        U: Send + Sync + 'static,
+    >(
+        &mut self,
+        rpc: &'static str,
+        req: tonic::Request<T>,
+    ) -> tonic::Result<tonic::Response<U>> {
+        debug!("handling request.");
```
noot, line 59: remove or add more info?

crates/astria-grpc-mock/src/mounted_mock.rs
```diff
@@ -0,0 +1,202 @@
...
+                        let msg = format!(
+                            "failed downcasting mock response to concrete type:\n\
+                             required type of gRPC response: `{expected}`\n\
+                             type of mock response: `{actual}`\n\
+                             JSON serialization:\n\
+                             {}",
+                            serde_json::to_string_pretty(erased_message.as_serialize())
+                                    .expect("can map registered protobuf response to json")
+                        );
+                    Err(tonic::Status::internal(msg))
```
noot, line 154: when would this error case happen? if the server is implemented to return type X but the mock response implements type Y?

noot (approved):
looks good, just a few questions :)

Fraser999 (approved):
Nice job - LGTM.

crates/astria-grpc-mock-test/proto/health.proto
SuperFluffy: IDK, I literally just copied it over. :)

crates/astria-grpc-mock-test/tests/health/main.rs
SuperFluffy: yeah

crates/astria-grpc-mock/src/mounted_mock.rs
SuperFluffy: yes, exactly. On line 134 we try and downcast the response to to `U`. If we are unable to do that the mock server returns an error (because we need to return something).

crates/astria-grpc-mock/src/mock_set.rs
SuperFluffy: yeah, fair point. It's also reporting which rpc is being served. wiremock didn't have that because their requests are all dynamic, whereas ours aren't.

PR #883, opened 2024-03-29
fix(sequencer-relayer): shutdown not propagating on API server fail

Fraser999:
## Summary
A fix for a minor issue in the shutdown flow, and some typo corrections.

## Background
If the `api_task` in `SequencerRelayer::run` joined before the `relayer_task` (and no `SIGTERM` was received), then `shutdown_token.cancel();` was never called meaning the `relayer_task` was always aborted under these conditions rather than being gracefully shut down.

## Changes
`shutdown_token.cancel();` is now always called irrespective of the reason for shutting down.  It is effectively a no-op if the reason for shutting down was due to the `relayer_task` joining, since the `api_task` has a different notifier, but it seems to be a little more future-proof to call this in every scenario, in case more tasks depending upon it are added in the future.

## Testing
This functionality isn't able to be easily tested currently, so no explicit tests for this have been added.


SuperFluffy (approved):
Thank you for catching this! You are right that is not testable at the moment. Maybe we should a) move the signal handler outside the `SequencerRelayer` object into `main.rs`, and b) instead have a `SequencerRelayerHandler` that exposes a single `SequencerRelayerHandler::shutdown` method. That way proper shutdown can be checked on every integration test.

PR #887, opened 2024-04-01
ci: add cargo audit

Fraser999:
## Summary
This adds `cargo audit` to the CI pipeline.

## Background
This is useful to catch issues published to [RUSTSEC](https://rustsec.org/).

## Changes
Added a job to test.yml to run it.  I used the instructions from [`audit-check`](https://github.com/rustsec/audit-check) (as recommended by `cargo-audit` [here](https://github.com/rustsec/rustsec/blob/main/cargo-audit/README.md#using-cargo-audit-on-github-action)), but chose to not restrict the check to run only if `Cargo.[toml|lock]` are modified, since `cargo audit` can start warning on a given dependency at any time.  We could probably change to follow their suggestion fully, i.e. run audit nightly and otherwise only run on changes to those two files, but I'll leave that up to SRE to decide :)

Note that we currently have three audit warnings, but they're all deemed as "allowed" by default (i.e. they're all informational only).  It seems like we should maybe look to address [the borsch one](https://rustsec.org/advisories/RUSTSEC-2023-0033) since, while it's not deemed a security issue, it's still best to avoid unsound code!

## Testing
N/A

### DIFF
--- a/.github/workflows/scheduled.yml
+++ b/.github/workflows/scheduled.yml
@@ -2,8 +2,12 @@ name: Scheduled
 on:
   schedule:
     - cron: '0 0 * * *'
+
 jobs:
   audit:
+    permissions:
+      issues: write
+      checks: write
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
--- a/.github/workflows/scheduled.yml
+++ b/.github/workflows/scheduled.yml
@@ -1,0 +1,12 @@
+name: Scheduled
+on:
+  schedule:
+    - cron: '0 0 * * *'
+jobs:
+  audit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: rustsec/audit-check@v1.4.1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,7 +2,7 @@ name: Test
 env:
   CI: true
   RUSTFLAGS: "-D warnings -D unreachable-pub --cfg tokio_unstable"
-on: 
+on:
   pull_request:
   merge_group:
   push:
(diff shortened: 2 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -79,7 +79,17 @@ jobs:
           cache-provider: "github"
       - name: Check that the lockfile is updated
         run: cargo fetch --locked
-    
+
+  audit:
+    runs-on: buildjet-2vcpu-ubuntu-2204
```
joroshiba, line 84: we can probably run this on a standard ubuntu runner. We use the buildjet for things that require more compute power, but can use standard github runners in many places.

joroshiba (commented):
Should we run this seperately from normal CI as a cron?

SuperFluffy:
I think we should gate running this on changes to `Cargo.lock` for new PRs, but also run this periodically (every 12 hours?) using a `schedule` workflow trigger: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule

.github/workflows/scheduled.yml
```diff
@@ -0,0 +1,12 @@
...
+  schedule:
+    - cron: '0 0 * * *'
+jobs:
+  audit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: rustsec/audit-check@v1.4.1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
```
joroshiba: nit: new line

.github/workflows/scheduled.yml
```diff
@@ -0,0 +1,12 @@
+name: Scheduled
+on:
+  schedule:
+    - cron: '0 0 * * *'
+jobs:
```
joroshiba, line 5: I think we need to add above this:

```yaml
permissions:
   issues: write
```

docs: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

.github/workflows/test.yml
```diff
@@ -79,7 +79,17 @@ jobs:
...
+    runs-on: ubuntu-22.04
+    needs: run_checker
+    if: needs.run_checker.outputs.run_audit == 'true'
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@1.76.0
+      - uses: rustsec/audit-check@v1.4.1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
   rust:
```
joroshiba, line 93: nit: new line between jobs

joroshiba (approved):
Approved with a couple notes

PR #889, opened 2024-04-01
feat(sequencer-relayer): provide a shutdown controller

### LINKED ISSUE #882: sequencer-relayer: improve shutdown control
Shutting down the `SequencerRelayer` should be able to be invoked from outwith the object itself.  Currently that can be done via sending a `SIGTERM`, but we also want the ability to invoke the shutdown sequence from `main` and from tests.

Fraser999:
## Summary
Adds a shutdown handle for the sequencer-relayer.

## Background
We want the ability to invoke the shutdown sequence from `main` and from tests.

## Changes
A new RAII object `ShutdownHandle` has been added.  This cancels the wrapped `CancellationToken` when the `ShutdownHandle` is dropped or when its `shutdown` method is called.

`SIGTERM` handling has been moved to main.rs; on receiving a `SIGTERM`, `shutdown` is called on the shutdown handle associated with the relayer.

The relayer's main  `select!` loop was simplified to only contain the two tasks.

## Testing
The `TestSequencerRelayer` was updated to make use of the new shutdown controller.  Also manually tested by sending a `SIGTERM` to a running sequencer-relayer.

## Related Issues
Closes #882


crates/astria-sequencer-relayer/Cargo.toml
```diff
@@ -69,7 +69,7 @@ jsonrpsee = { workspace = true, features = ["server"] }
 once_cell = { workspace = true }
 tempfile = { workspace = true }
 tokio = { workspace = true, features = ["test-util"] }
-tokio-stream = { workspace = true }
+tokio-stream = { workspace = true, features = ["net"] }
```
Fraser999, line 72: This was required if I ran the tests from just this crate.

crates/astria-sequencer-relayer/src/sequencer_relayer.rs
```diff
@@ -200,14 +253,18 @@ impl ShutDown {
...
+            match timeout(limit, &mut api_task)
                 .await
                 .map(crate::utils::flatten)
             {
-                Ok(Ok(())) => info!("api server exited gracefully"),
-                Ok(Err(error)) => error!(%error, "api server exited with an error"),
+                Ok(Ok(())) => info!("API server exited gracefully"),
+                Ok(Err(error)) => error!(%error, "API server exited with an error"),
                 Err(_) => {
-                    error!("api server did not shut down after 25 seconds; killing it");
```
Fraser999, line 210: There was a copy-paste error here (`25` should have been `4`), so I extracted the value out to a variable.

I prefer having variable data separated from the main static message in `tracing` macros, so I did that here and also in the equivalent block above for the `relayer_task`.  Let me know if this is not a preference here please.

crates/astria-sequencer-relayer/tests/blackbox/helper.rs
```diff
@@ -180,7 +190,18 @@ pub struct TestSequencerRelayer {
...
-        self.sequencer_relayer.abort();
+        // We drop the shutdown controller here to cause the sequencer relayer to shut down.
+        let _ = self.relayer_shutdown_controller.take();
+
+        let sequencer_relayer = mem::replace(&mut self.sequencer_relayer, tokio::spawn(async {}));
+        let _ = futures::executor::block_on(async move {
+            timeout(Duration::from_secs(30), sequencer_relayer)
+                .await
+                .unwrap_or_else(|_| {
+                    panic!("timed out waiting for sequencer relayer to shut down");
```
Fraser999, line 201: I generally avoid putting panicking code in drop impls, but I think it's probably ok here since this is a test-only object, and having a test process abort seems like a better option than just logging here since the log message could easily be missed.

crates/astria-sequencer-relayer/tests/blackbox/helper.rs
```diff
@@ -305,6 +326,12 @@ pub struct TestSequencerRelayerConfig {
 
 impl TestSequencerRelayerConfig {
     pub async fn spawn_relayer(self) -> TestSequencerRelayer {
+        assert_ne!(
```
Fraser999, line 329: We need the tokio runtime to not be of the "current thread" flavour so as to allow the `TestSequencerRelayer::drop` to not hang at the `futures::executor::block_on` call.

crates/astria-sequencer-relayer/src/sequencer_relayer.rs
```diff
@@ -145,6 +143,58 @@ impl SequencerRelayer {
...
+    /// # Panics
+    ///
+    /// Panics if a signal listener could not be constructed (usually if this is not run on Unix).
+    #[must_use]
+    pub fn new() -> (Self, mpsc::Receiver<()>) {
+        let (sender, receiver) = mpsc::channel(1);
+        let cloned_sender = sender.clone();
+        let mut sigterm = signal(SignalKind::terminate())
+            .expect("setting a SIGTERM listener should always work on Unix");
+        tokio::spawn(async move {
```
SuperFluffy, line 169: Leaving task `JoinHandle` dangling always make me a bit uneasy. This one is simple enough, but my preference is to alway `.join` or .await` them.

SuperFluffy (commented):
I like explicitly handling the shutdowns very much so that we can test this on every test.

I think this is almost ready to merge, but I would like to explore whether we can both simplify the `ShutdownHandle` (by moving the signal trap into `main.rs` and relying on a `oneshot` channel or a `CancellationToken`/`DropGuard`) and making it more explicit (by providing an `ShutdownHandle::shutdown` method).

crates/astria-sequencer-relayer/Cargo.toml
Fraser999: Thanks!

crates/astria-sequencer-relayer/tests/blackbox/helper.rs
Fraser999: Well, it's not a problem with code currently in the `main` branch.  It's the addition in this PR of the call to `futures::executor::block_on` in the `TestSequencerRelayer::drop`.

The async function passed to `block_on` can't be cancelled or paused, and I believe that's stopping the relayer making progress inside the main `select!` macro as the single tokio thread is blocked.

crates/astria-sequencer-relayer/src/main.rs
```diff
@@ -44,10 +51,30 @@ async fn main() -> ExitCode {
...
 
-    SequencerRelayer::new(cfg)
-        .expect("could not initialize sequencer relayer")
-        .run()
-        .await;
+    let mut sigterm = signal(SignalKind::terminate())
+        .expect("setting a SIGTERM listener should always work on Unix");
+    let (sequencer_relayer, shutdown_handle) =
+        SequencerRelayer::new(cfg).expect("could not initialize sequencer relayer");
+    let sigterm_handle = tokio::spawn(async move {
```
SuperFluffy, line 58: Does it make sense to invert this, i.e. `spawn(sequencer_relayer.run()` instead of the signal handlers?

crates/astria-sequencer-relayer/src/main.rs
```diff
@@ -44,10 +51,30 @@ async fn main() -> ExitCode {
...
+        let shutdown_token = shutdown_handle.token();
+        tokio::select!(
+            _ = sigterm.recv() => {
+                // We don't care about the result (i.e. whether there could be more SIGTERM signals
+                // incoming); we just want to shut down as soon as we receive the first `SIGTERM`.
+                info!("received SIGTERM, issuing shutdown to all services");
+                shutdown_handle.shutdown();
+            }
+            () = shutdown_token.cancelled() => {
+                // Nothing to log here - we're just finishing up waiting for a SIGTERM having
```
SuperFluffy, line 68: Can you report this at a `warn!` level? This arm should never match - if it does match, then something strange is going on inside sequencer relayer (i.e. it shouldn't just shut itself down).

crates/astria-sequencer-relayer/src/sequencer_relayer.rs
```diff
@@ -83,22 +79,19 @@ impl SequencerRelayer {
...
         self.api_server.local_addr()
     }
 
-    /// Run Sequencer Relayer.
-    ///
-    /// # Panics
-    /// Panics if a signal listener could not be constructed (usually if this binary is not run on a
-    /// Unix).
+    /// Runs Sequencer Relayer.
     pub async fn run(self) {
```
SuperFluffy, line 95: I wonder if we should just go with what hyper and others are doing and just call this `spawn`, and internally `tokio::spawn` the process, returning a `Handle` to the object (which itself contains the `JoinHandle` and the token).

SuperFluffy (approved):
This is neat, thank you!

crates/astria-sequencer-relayer/src/main.rs
Fraser999: Yup - sure.

crates/astria-sequencer-relayer/src/sequencer_relayer.rs
Fraser999: As discussed elsewhere, I think it's probably worth keeping the creation of the token to the `new` function so we have the option to have shutdown handling already in place before we actually start running the relayer.

PR #890, opened 2024-04-01
test(sequencer): add unit tests for src/api_state_ext

Lilyjjo:
## Summary
Seventh set of state-ext unit tests. In total there are seven files which
need tests:
- ~`astria-sequencer/src/api_state_ext.rs`~ (This PR)
- ~`astria-sequencer/src/state_ext.rs`~
- ~`astria-sequencer/src/accounts/state_ext.rs`~
- ~`astria-sequencer/src/asset/state_ext.rs`~
- ~`astria-sequencer/src/authority/state_ext.rs`~ 
- ~`astria-sequencer/src/bridge/state_ext.rs`~
- ~`astria-sequencer/src/ibc/state_ext.rs`~

This PR just tests the `src/api_state_ext/state_ext.rs` file.

## Background
It is good to have unit tests to ensure that the database works as intended.

## Changes
Unit tests for the functionality in the file `src/api_state_ext/state_ext.rs` were added.

## Testing
:)

### DIFF
--- a/crates/astria-sequencer/src/api_state_ext.rs
+++ b/crates/astria-sequencer/src/api_state_ext.rs
@@ -441,7 +441,7 @@ mod test {
         // create inner rollup id/tx data
         let mut deposits = HashMap::new();
         for _ in 0..2 {
-            let rollup_id = RollupId::new([rng.gen(); 32]);
+            let rollup_id = RollupId::new(rng.gen());
             let bridge_address = Address::try_from_slice(&[rng.gen(); 20]).unwrap();
             let amount = rng.gen::<u128>();
             let asset_id = Id::from_denom(&rng.gen::<u8>().to_string());
--- a/crates/astria-sequencer/src/api_state_ext.rs
+++ b/crates/astria-sequencer/src/api_state_ext.rs
@@ -375,3 +375,413 @@ pub(crate) trait StateWriteExt: StateWrite {
 }
 
 impl<T: StateWrite> StateWriteExt for T {}
+
+#[cfg(test)]
+mod test {
+
+    use std::collections::HashMap;
+
+    use astria_core::{
+        sequencer::v1::{
+            asset::Id,
+            block::{
+                Deposit,
+                SequencerBlock,
+            },
+            Address,
+            RollupId,
+        },
+        Protobuf,
+    };
+    use cnidarium::StateDelta;
+    use sha2::{
+        Digest as _,
+        Sha256,
+    };
+    use tendermint::{
+        account,
+        block::{
+            header::Version,
+            Header,
+        },
+        AppHash,
+        Hash,
+        Time,
+    };
+    use tendermint_proto::v0_37::types::{
+        BlockId,
+        PartSetHeader,
+    };
+
+    use super::*;
+    use crate::proposal::commitment::generate_rollup_datas_commitment;
+
+    struct ValueGenerator {
+        value: u8,
+    }
+
+    // used to ensure that unique data is being written for all sequencer block components
+    impl ValueGenerator {
+        fn new(start: u8) -> Self {
+            ValueGenerator {
+                value: start,
+            }
+        }
+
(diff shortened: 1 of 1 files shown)


crates/astria-sequencer/src/api_state_ext.rs
```diff
@@ -375,3 +375,413 @@ pub(crate) trait StateWriteExt: StateWrite {
...
+            last_results_hash: Some(Hash::try_from([gen.generate(); 32].to_vec()).unwrap()),
+            next_validators_hash: Hash::try_from([gen.generate(); 32].to_vec()).unwrap(),
+            proposer_address: account::Id::try_from([gen.generate(); 20].to_vec()).unwrap(),
+            time: Time::now(),
+            validators_hash: Hash::try_from([gen.generate(); 32].to_vec()).unwrap(),
+            version: Version {
+                app: u64::from(gen.generate()),
+                block: u64::from(gen.generate()),
+            },
+        };
```
noot, line 477: if the generator is used for creating essentially random numbers, maybe use rand instead? https://docs.rs/rand/latest/rand/ then you can do:
```
app_hash: rng.gen(),
...
```

for example

crates/astria-sequencer/src/api_state_ext.rs
```diff
@@ -375,3 +375,413 @@ pub(crate) trait StateWriteExt: StateWrite {
...
+
+        // create inner rollup id/tx data
+        let mut deposits = HashMap::new();
+        for _ in 0..2 {
```
noot, line 481: is this just adding 2 arbitrary deposit types?

crates/astria-sequencer/src/api_state_ext.rs
```diff
@@ -375,3 +375,413 @@ pub(crate) trait StateWriteExt: StateWrite {
...
+                rollup_id,
+                amount,
+                asset_id,
+                destination_chain_address,
+            );
+            deposits.insert(rollup_id, vec![deposit]);
+        }
+
+        // create block's other data
+        let committments = generate_rollup_datas_commitment(&[], deposits.clone());
```
noot, line 498:
```suggestion
        let commitments = generate_rollup_datas_commitment(&[], deposits.clone());
```

noot (commented):
looks good, main comment would be to use `rand` instead of the manual number generator for generating the test data - you can probably use all zeros/constant values for the test data except for the fields you need to vary (height, deposit info, etc)

crates/astria-sequencer/src/api_state_ext.rs
Lilyjjo: Oh nice, I'll replace with rng(). I was trying to avoid doing all zeros/default values because I wanted the tests to ensure that unique data was being written and retrieved from the correct places.

crates/astria-sequencer/src/api_state_ext.rs
Lilyjjo: Yeah, I wanted to put rollup data in that could be read out. I can replace with normal non-deposit generated rollup transactions if you'd prefer.

noot: this is fine!

crates/astria-sequencer/src/api_state_ext.rs
```diff
@@ -375,3 +375,350 @@ pub(crate) trait StateWriteExt: StateWrite {
...
+            version: Version {
+                app: 0,
+                block: 0,
+            },
+        };
+
+        // create inner rollup id/tx data
+        let mut deposits = HashMap::new();
+        for _ in 0..2 {
+            let rollup_id = RollupId::new([rng.gen(); 32]);
```
noot, line 444: just curious, can you do:
```suggestion
            let rollup_id = RollupId::new(rng.gen());
```

Lilyjjo: yes you can! changed

Issue #850, opened 2024-03-20
sequencer: respect max_tx_bytes when preparing proposal


noot:
from zellic:
>  https://github.com/astriaorg/astria/blob/105474286ea1803244dcf2851bb5a84f5e16daa8/crates/astria-sequencer/src/app.rs#L232-L255 - The PrepareProposal request has a max_tx_bytes field which the proposer must respect, but currently this isn't checked. Depending on how CometBFT is configured, the supplied transactions could already be larger than this limit (refs https://docs.cometbft.com/v0.37/spec/abci/abci++_methods#parameters-and-types), or if the block is almost full then prepending the commitment transactions could push it over the limit.

PR #911, opened 2024-04-03
fix(sequencer): respect max_tx_bytes when preparing proposals

Lilyjjo:
## Summary
Ensures that `cometBFT` blocks will not become too large during `prepare_proposal()` from the addition of commitments to the front of the block. 

## Background
`cometBFT` sends in a `max_tx_bytes` parameter to the ABCI `prepare_proposal()` that should be respected when modifying the block. We currently add commitments to the front of the block for the included rollup ids and included rollup transactions. It is possible that adding these bytes will cause the `cometBFT` block to be larger than the `max_tx_bytes` parameter.

## Changes
- Refactored `execute_block()` into `execute_block()` and `execute_and_filter_block()` and pulled out the shared logic into `execute_transaction()`.
- Prevents `cometBFT` blocks from becoming too large from adding the commitments to the front of the block. 

## Testing
- Added new unit test ensuring that overflow is okay.
- Ran local setup to ensure that transactions still flow through the system in general. 

## Issues
closes https://github.com/astriaorg/astria/issues/850


crates/astria-sequencer/src/proposal/commitment.rs
```diff
@@ -30,6 +30,25 @@ impl GeneratedCommitments {
...
+impl Example for GeneratedCommitments {
+    fn example() -> Self {
+        GeneratedCommitments {
+            rollup_datas_root: [0; 32],
+            rollup_ids_root: [0; 32],
+        }
+    }
+
+    // used for accounting for commitments' data size in cometbft block
+    fn commitments_size(self) -> usize {
```
Lilyjjo, line 47: I'm not sure what the idiomatic way to get the size of a struct's inner parts before having a concrete version of the struct is. I didn't want to hardcode the size in case changes were made later to the `GeneratedCommitments` logic/struct.

noot: left a comment, you can use `std::mem::size_of`

crates/astria-sequencer/src/app.rs
```diff
@@ -344,6 +355,7 @@ impl App {
     async fn execute_block_data(
         &mut self,
         txs: Vec<bytes::Bytes>,
+        mut remaining_cometbft_bytes: Option<i64>,
```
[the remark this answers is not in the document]
noot: yeah this seems fine - i'd change the i64 to a usize tho so you don't need to do conversions

crates/astria-sequencer/src/app.rs
```diff
@@ -387,9 +399,28 @@ impl App {
                 continue;
             }
 
+            // Don't include tx if it would make the cometbft block too large
+            if let Some(remaining) = remaining_cometbft_bytes {
+                if remaining < i64::try_from(tx.len()).expect("failed to turn tx length into i64") {
```
noot, line 404: can just use usize - same for other places

crates/astria-sequencer/src/app.rs
```diff
@@ -387,9 +399,28 @@ impl App {
...
+                    debug!(
+                        transaction_hash = %telemetry::display::hex(&tx_hash),
+                        remaining_cometbft_bytes = remaining,
+                        tx_len = tx.len(),
+                        "excluding transaction: max cometbft data limit reached"
+                    );
+                    excluded_tx_count += 1;
+                    continue;
```
noot, line 412: at this point we know we can't include any more txs, so just update `excluded_tx_count += remaining_tx_count` and return - you can update `for tx in txs` to `for (i, tx) in txs.enumerate()` to get the tx index you're at

noot (changes requested):
nice, logic looks good, just a few things to address!

Lilyjjo:
addressed all the comments!

crates/astria-sequencer/src/app.rs
```diff
@@ -344,13 +351,15 @@ impl App {
...
+        mut remaining_cometbft_bytes: Option<usize>,
     ) -> (Vec<SignedTransaction>, Vec<bytes::Bytes>) {
         let mut signed_txs = Vec::with_capacity(txs.len());
         let mut validated_txs = Vec::with_capacity(txs.len());
         let mut block_sequence_data_bytes: usize = 0;
         let mut excluded_tx_count: usize = 0;
+        let txs_cached_length = txs.len();
 
-        for tx in txs {
+        for (i, tx) in txs.into_iter().enumerate() {
```
SuperFluffy, line 362: The contents of this block look incredibly complex. Is there a way to move the loop body into a freestanding function and simplify? The explicit flow control using the `continue` and `break` statements could then be simplified using the `?` operator.

crates/astria-sequencer/src/app.rs
```diff
@@ -387,9 +396,27 @@ impl App {
...
+            // Don't include tx if it would make the cometbft block too large
+            if let Some(remaining) = remaining_cometbft_bytes {
+                if remaining < tx.len() {
+                    debug!(
+                        transaction_hash = %telemetry::display::hex(&tx_hash),
+                        remaining_cometbft_bytes = remaining,
+                        tx_len = tx.len(),
+                        "excluding transaction: max cometbft data limit reached"
+                    );
+                    excluded_tx_count += txs_cached_length - i;
```
SuperFluffy, line 408: This operation looks very complex and prone to error.

crates/astria-sequencer/src/app.rs
```diff
@@ -344,13 +351,15 @@ impl App {
     async fn execute_block_data(
         &mut self,
         txs: Vec<bytes::Bytes>,
+        mut remaining_cometbft_bytes: Option<usize>,
```
SuperFluffy, line 354: I am not a fan of arguments of this type which fundamentally change the behavior of the contained code. When reading the code it's not obvious what the `None` argument is supposed to do.

IMO it's cleaner to have 2 separate functions `execute_block_data_with_limit` and `execute_block_data_without_limit` (bad names; feel free to bikeshed) that might internally delegate to `Some(capacity)` and `None` so that the purpose is clearer to the reader

SuperFluffy (changes requested):
`execute_block_data` is approaching 100 lines in length. I think it's time to factor the loop into a freestanding function.

Also please consider using saturating arithmetic.

crates/astria-sequencer/src/app.rs
Lilyjjo: Cool, yeah it's def not great in this state. I'll refactor into freestanding functions

crates/astria-sequencer/src/app.rs
```diff
@@ -418,6 +428,67 @@ impl App {
...
+                self.execution_result.insert(tx_hash.into(), Ok(events));
+                return Ok(signed_tx);
+            }
+            Err(e) => {
+                debug!(
+                    transaction_hash = %telemetry::display::hex(&Sha256::digest(tx_hash)),
+                    error = AsRef::<dyn std::error::Error>::as_ref(&e),
+                    "failed to execute transaction"
+                );
+                self.execution_result.insert(tx_hash.into(), Err(e));
```
Lilyjjo, line 486: I couldn't figure out how to copy the error so I couldn't bubble it up. I don't like returning this error message but I wasn't sure what a better thing to do was.

Lilyjjo, line 486: I couldn't figure out how to copy the error so I couldn't bubble it up. I don't like returning this error message but I wasn't sure what a better thing to do was.

Lilyjjo:
I'm not 100% on what I did regarding returning/managing errors. I don't have a good sense yet of the idiomatic way to do error handling is for the codebase 🫣

crates/astria-sequencer/src/app.rs
```diff
@@ -237,7 +238,14 @@ impl App {
...
 
-        let (signed_txs, txs_to_include) = self.execute_block_data(prepare_proposal.txs).await;
+        let remaining_tx_bytes: usize = usize::try_from(prepare_proposal.max_tx_bytes)
+            .expect("should be able to convert max_tx_bytes to usize")
+            .checked_sub(std::mem::size_of::<GeneratedCommitments>())
+            .expect("commitment size should not be larger than the prepare proposal's max bytes");
+
+        let (signed_txs, txs_to_include) = self
+            .execute_and_filter_block_data(VecDeque::from(prepare_proposal.txs), remaining_tx_bytes)
+            .await;
```
noot, line 248: both this function and `execute_block_data` execute the txs, but we don't want to double-execute everything (this will also cause everything to fail on second execution) - tbh i preferred the previous solution as it also doesn't loop over all the txs twice. i don't think there's a way to split apart the length-check and execution unfortunately as we only include successfully-executing txs in a block. i'd suggest moving it back into one `filter_and_execute_block_data` function but breaking that function into multiple functions (ie inside the for loop, it calls a fn that handles just that tx).

Lilyjjo: I think I'm missing something, I don't see where we're doing double execution

crates/astria-sequencer/src/app.rs
```diff
@@ -387,23 +391,29 @@ impl App {
...
             }
 
+            // Don't include tx if it would make the cometbft block too large
+            if remaining_cometbft_bytes < tx.len() {
+                debug!(
+                    remaining_cometbft_bytes = remaining_cometbft_bytes,
+                    number_transactions_excluded = tx.len() + 1,
+                    "excluding rest of transactions: max cometbft data limit reached"
+                );
+                excluded_tx_count += txs.len() + 1;
```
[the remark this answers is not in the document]
Lilyjjo: Changed back to executing all of the transactions

crates/astria-sequencer/src/app.rs
```diff
@@ -387,23 +391,29 @@ impl App {
                 continue;
             }
 
+            // Don't include tx if it would make the cometbft block too large
+            if remaining_cometbft_bytes < tx.len() {
```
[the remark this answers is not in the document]
Lilyjjo: Changed back to executing all of the transactions

crates/astria-sequencer/src/app.rs
```diff
@@ -372,37 +380,50 @@ impl App {
...
+                debug!(
+                    transaction_hash = %telemetry::display::base64(&tx_hash),
+                    max_cometbft_bytes = block_size_constraints.max_cometbft,
+                    included_data_bytes = block_size_constraints.current_cometbft,
+                    tx_data_bytes = tx.len(),
+                    "excluding transactions: max cometBFT data limit reached"
+                );
+                excluded_tx_count += 1;
+                continue;
+            }
```
noot, line 407: i would move this to directly under the `for` statement (also move `let tx_hash = Sha256::digest(&tx);` right above this block), since we know if the cometbft max size has been reached, the rest of the txs will be skipped also, so we don't need to spend resources decoding them

crates/astria-sequencer/src/app.rs
```diff
@@ -372,37 +380,50 @@ impl App {
...
                     signed_txs.push(signed_tx);
                     validated_txs.push(tx);
-                    block_sequence_data_bytes += tx_sequence_data_bytes;
                 }
-                Err(e) => {
+                Err(_e) => {
                     debug!(
                         transaction_hash = %telemetry::display::base64(&tx_hash),
-                        error = AsRef::<dyn std::error::Error>::as_ref(&e),
                         "failed to execute transaction, not including in block"
```
noot, line 424: we should leave this error log

crates/astria-sequencer/src/app.rs
```diff
@@ -415,7 +436,67 @@ impl App {
...
+    ))]
+    async fn execute_block_data(
+        &mut self,
+        txs: VecDeque<bytes::Bytes>,
+    ) -> anyhow::Result<Vec<SignedTransaction>, anyhow::Error> {
+        let mut signed_txs = Vec::new();
+        for tx in txs {
+            // all txs in the proposal should be deserializable and executable,
+            // if any txs were not deserializeable or executable, they would not have been
+            // returned by `execute_block_data`
```
noot, line 460:
```suggestion
            // returned by `execute_and_filter_block_data`
```

crates/astria-sequencer/src/app.rs
Lilyjjo: I keep on going back and forth on if to short circuit the loop or not. If one transaction doesn't fit, others may still be able to. If we're trying to optimize for largest block (and max profit) it'd be better if we keep on checking. This isn't super relevant for this PR because the commitment size is only like 64 bytes and at most will kick out a single transaction. 

This will probably have to be refactored anyways once we start adding ToB bundles from the relay, where we'll be more likely to hit the max size.

crates/astria-sequencer/src/app.rs
noot: see my previous comment - can change this fn to not return anything then check `self.execution_result` in callers

crates/astria-sequencer/src/app.rs
```diff
@@ -2186,4 +2267,90 @@ mod test {
...
+        assert_eq!(
+            std::mem::size_of::<GeneratedCommitments>(),
+            64,
+            "expecting the size to be equal to the length of the commitments being appended to \
+             the cometBFT block"
+        );
+    }
+
+    #[tokio::test]
+    async fn app_prepare_proposal_cometbft_max_bytes_overflow_ok() {
```
noot, line 2284: nice, could you add a test for testing `MAX_SEQUENCE_DATA_BYTES_PER_BLOCK` exceeded as well?

noot (commented):
cool, this looks good now! just have a few smaller comments/suggestions

crates/astria-sequencer/src/app.rs
```diff
@@ -415,7 +436,67 @@ impl App {
...
+            Err(e) => {
+                debug!(
+                    transaction_hash = %telemetry::display::base64(&Sha256::digest(tx_hash)),
+                    error = AsRef::<dyn std::error::Error>::as_ref(&e),
+                    "failed to execute transaction"
+                );
+                self.execution_result.insert(tx_hash.into(), Err(e));
+                Err(anyhow!("failed to execute transaction"))
+            }
+        }
```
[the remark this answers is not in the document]
Lilyjjo: nice, added! this made it cleaner

crates/astria-sequencer/src/app.rs
Lilyjjo: Put it up at the top!

crates/astria-sequencer/src/app.rs
```diff
@@ -415,7 +436,68 @@ impl App {
...
+                 prepare proposal",
+            )?;
+
+            // execute transaction and ensure that it succeeded
+            let tx_hash = self.execute_and_store_signed_tx(&signed_tx).await;
+            if let Some(Err(_e)) = self.execution_result.get(&tx_hash) {
+                bail!(
+                    "transaction failed during execution, should've been filtered out in prepare \
+                     proposal"
+                )
```
noot, line 472: should also remove the tx from the execution_results map as this will cause a protocol error, but it's already a protocol error if a tx in a finalized block failed :/ so this is fine, just gonna note this for when i merge with my branch later

noot (approved):
looks good now :D

crates/astria-sequencer/src/app.rs
```diff
@@ -237,7 +238,20 @@ impl App {
         self.validator_address = Some(prepare_proposal.proposer_address);
         self.update_state_for_new_round(&storage);
 
-        let (signed_txs, txs_to_include) = self.execute_block_data(prepare_proposal.txs).await;
+        let mut block_size_constraints = BlockSizeConstraints::new(
+            usize::try_from(prepare_proposal.max_tx_bytes)
+                .context("failed to convert max_tx_bytes to usize")?,
+        );
+        block_size_constraints
+            .cometbft_checked_add(std::mem::size_of::<GeneratedCommitments>())
```
SuperFluffy, line 246: Seeing this again I still hate it. :D

I have commented on it there.

crates/astria-sequencer/src/proposal/block_size_constraints.rs
```diff
@@ -0,0 +1,96 @@
...
+    pub(crate) current_cometbft: usize,
+}
+
+impl BlockSizeConstraints {
+    pub(crate) fn new(cometbft_max_size: usize) -> Self {
+        BlockSizeConstraints {
+            max_sequencer: MAX_SEQUENCE_DATA_BYTES_PER_BLOCK,
+            max_cometbft: cometbft_max_size,
+            current_sequencer: 0,
+            current_cometbft: 0,
```
SuperFluffy, line 20: Set this to `current_cometbft: GeneratedCommitments::TOTAL_SIZE` and also provide an associated const on it:
```rust
impl GeneratedCommitments {
    const TOTAL_SIZE: usize = 64; // or whatever it was
}
```

crates/astria-sequencer/src/proposal/block_size_constraints.rs
```diff
@@ -0,0 +1,96 @@
...
+        }
+    }
+
+    pub(crate) fn sequencer_has_space(&self, size: usize) -> bool {
```
SuperFluffy, line 24: Please use checked or saturating arithmetic for all functions in this file.

crates/astria-sequencer/src/proposal/block_size_constraints.rs
```diff
@@ -0,0 +1,96 @@
...
+        self.current_sequencer + size <= self.max_sequencer
```
SuperFluffy, line 25: For example here:
```suggestion
        self.current_sequencer.saturating_add(size) <= self.max_sequencer
```

crates/astria-sequencer/src/proposal/block_size_constraints.rs
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    pub(crate) fn cometbft_has_space(&self, size: usize) -> bool {
+        self.current_cometbft + size <= self.max_cometbft
+    }
+
+    pub(crate) fn sequencer_checked_add(&mut self, size: usize) -> anyhow::Result<()> {
+        if self.current_sequencer + size > self.max_sequencer {
```
SuperFluffy, line 33: Please use `self.current_sequencer.saturating_add(size) > self.max_sequencer`

PR #916, opened 2024-04-04
Lilyjjo/sequencer state duplicate rollup ids

### LINKED ISSUE #894: sequencer: `get_deposit_rollup_ids` can return duplicate rollup IDs
`get_deposit_rollup_ids` can return duplicate rollup IDs, which is fine but not intended, each id should be unique.

Lilyjjo:
## Summary
Fix for #894. Now calls to `get_deposit_rollup_ids()` will not return duplicate `rollup_ids`.

## Testing
Added unit test.

## Related Issues
closes #894


crates/astria-sequencer/src/bridge/state_ext.rs
```diff
@@ -139,9 +142,9 @@ pub(crate) trait StateReadExt: StateRead {
                 hex::decode(key_parts[1]).context("invalid rollup ID hex string")?;
             let rollup_id =
                 RollupId::try_from_slice(&rollup_id_bytes).context("invalid rollup ID bytes")?;
-            rollup_ids.push(rollup_id);
+            rollup_ids.insert(rollup_id);
```
SuperFluffy, line 145: Q: Should dupes be reported? Maybe at the debug level? Maybe not necessary

crates/astria-sequencer/src/bridge/state_ext.rs
```diff
@@ -126,7 +129,7 @@ pub(crate) trait StateReadExt: StateRead {
     #[instrument(skip(self))]
     async fn get_deposit_rollup_ids(&self) -> Result<Vec<RollupId>> {
```
SuperFluffy, line 130: Why not just just return a `HashSet` here?

crates/astria-sequencer/src/bridge/state_ext.rs
```diff
@@ -615,6 +618,81 @@ mod test {
         );
     }
 
+    #[tokio::test]
+    async fn get_deposit_rollup_ids() {
```
SuperFluffy, line 622: I think the extra check in the middle of this test is redundant. IMO just put 3 deposits, 1 of them a duplicate. In the end, ensure that a) final result has `len == 2`, and b) that rollup IDs are in the set.

SuperFluffy (approved):
The logic looks sound. I think it would be fine to have `get_deposit_rollup_ids -> HashSet` and let the caller decide whether to make it into a vec or not.

PR #917, opened 2024-04-04
feat(conductor): add blackbox tests for conductor running in soft-only mode

SuperFluffy:
## Summary
Add blackbox tests for conductor running in soft-only mode.

## Background
Conductor has virtually no test coverage and is completely lacking blackbox tests. This patch starts getting some rudimentary testing going.

## Changes
- Add blackbox tests for conductor running in soft-only mode and test the following scenarios:
    1. Genesis state and initial commitment state are read from the rollup, sequencer is queried for the correct rollup ID and height, and the block is executed against the rollup.
    2. As in 1., but 2 subsequent blocks are fetched and executed
    3. As in 1. but the rollup's commitment state is already at a later soft-number and conductor only fetches and executes a block at that later height
    4. As in 1., but the sequencer height stored in the genesis is later, and conductor correctly fetches the first soft rollup number starting at that later sequencer height.
- testing revealed that startup is blocked on conductor trying to determine a sequencer namespace, even though that's only required when running in firm-only or firm-and-soft modes; this has been moved into the celestia reader.

## Testing
These are tests.


crates/astria-conductor/tests/blackbox/soft_only.rs
```diff
@@ -0,0 +1,326 @@
...
+    mount_get_filtered_sequencer_block,
+    mount_get_genesis_info,
+    mount_update_commitment_state,
+};
+
+#[tokio::test]
+async fn simple() {
+    let test_conductor = spawn_conductor(CommitLevel::SoftOnly).await;
+
+    mount_get_genesis_info!(
```
SuperFluffy, line 24: Some of the macros can probably be combined. For example, mounting the genesis info and (initial) commitment state could be made neater by using one macro. Mounting block execution + updating the commitment state also make sense as one macro.

crates/astria-conductor/src/celestia/builder.rs
```diff
@@ -39,24 +34,21 @@ impl Builder {
...
-
         let celestia_client = create_celestia_client(celestia_http_endpoint, &celestia_token)
             .wrap_err("failed initializing client for Celestia HTTP RPC")?;
 
         let latest_celestia_heights =
             stream_latest_heights(celestia_client.clone(), celestia_block_time);
 
         Ok(Reader {
-            block_verifier,
+            // block_verifier,
```
noot, line 47: remove

crates/astria-conductor/tests/blackbox/helpers/macros.rs
```diff
@@ -0,0 +1,237 @@
...
+            $crate::genesis_info!(
+                sequencer_genesis_block_height: $sequencer_height,
+                celestia_base_block_height: $celestia_height,
+                celestia_block_variance: $variance,
+            )
+        ).await;
+    };
+}
+
+// fn latest_commit() -> response::Wrapper<tendermint_rpc::endpoint::commit::Response> {
```
noot, line 189: remove?

crates/astria-conductor/src/celestia/mod.rs
```diff
@@ -590,6 +599,40 @@ async fn process_sequencer_blob(
...
+                    attempt,
+                    wait_duration,
+                    error = error as &dyn std::error::Error,
+                    "attempt to grab sequencer block failed; retrying after backoff",
+                );
+                futures::future::ready(())
+            },
+        );
+
+    let response = tryhard::retry_fn(|| client.latest_commit())
```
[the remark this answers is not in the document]
SuperFluffy: Getting the latest sequencer block was wasteful (we just grab the header after all and throw everything else away).

Kinda figured this out because it felt weird having to mount a mocked block just for this call.

Actutally, I think this one might be even better: https://docs.rs/tendermint-rpc/0.34.1/tendermint_rpc/endpoint/genesis/index.html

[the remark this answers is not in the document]
SuperFluffy: I am going to move this diff out of this PR. I initially set out to also mock firm-only mode, which is what this change was for.

crates/astria-conductor/tests/blackbox/helpers/telemetry.rs
SuperFluffy, line 1: will remove

Fraser999 (approved):

crates/astria-conductor/tests/blackbox/helpers/macros.rs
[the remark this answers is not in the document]
SuperFluffy: Fair point regarding debugging and macros being opaque. I agree and thought hard about this. I tried other ways of increasing test clarity (for example types), but none of them quite worked. I finally decided on macros because at least a reader/reviewer will understand what conductor is supposed to fetch and submit.

PR #925, opened 2024-04-05
fix(conductor): enable tls for grpc connections

joroshiba:
## Summary
Added a couple of tonic features to conductor so it can use tls and http2 correctly.

## Background
When connecting over an https http2 (h2) connection conductor would get errors stating `http2 error: connection error detected: frame with invalid size, connection error detected: frame with invalid size`. Tonic wasn't using tls because it wasn't turned on. Adding these two features allows "native" certs for client to be used by default and uses tls and http2 correctly when making client requests.

## Changes
- enabled `tls` and `tls-roots` features on tonic in conductor 

## Testing
Ran conductor locally against  dusk-4 network. Additionally ran local smoke test using the newly built image which confirmed still works over non https urls.


jbowen93 (approved):
Rubber stamp

jbowen93:
lol I am no longer a code owner, fair

sambukowski (approved):

SuperFluffy (approved):
I suppose we don't need a more specific TLS config at this time (identity, CA Certs, client auth)? https://docs.rs/tonic/latest/tonic/transport/server/struct.ServerTlsConfig.html

joroshiba:
> I suppose we don't need a more specific TLS config at this time (identity, CA Certs, client auth)? https://docs.rs/tonic/latest/tonic/transport/server/struct.ServerTlsConfig.html

I believe that since this is client side we should be ok with the native tls roots. 

Server side we are letting cloudflare handle ssl termination.

PR #930, opened 2024-04-06
feat(cli): support dusk-4 and latest charts

joroshiba:
## Summary
Updates the cli to use dusk4 network by default, and updates the rollup deployment to support latest chart formats required. 

## Background
Dusk-4 network is live updating CLI to give us feature parity with it.


crates/astria-cli/src/cli/rollup.rs
```diff
@@ -108,6 +130,16 @@ pub struct ConfigCreateArgs {
...
     pub namespace: String,
+    /// Choose to enable the Celestia feature
+    #[clap(long = "celestia-node.enabled", env = "ROLLUP_ENABLE_CELESTIA_NODE")]
+    pub enable_celestia_node: bool,
+}
+
+fn validate_initial_block_height(val: &str) -> Result<u64, String> {
+    match val.parse::<u64>() {
+        Ok(height) if height >= 2 => Ok(height),
+        _ => Err(String::from("The block height must be at least 2.")),
```
SuperFluffy, line 141: Why throw away the error? Also if a static string is desired this could just be `Err("the block height must be at least 2")` and the return value `Result<u64, &'static str>`.

I guess the minimum must be `2` because it needs `height - 1` to validate `height`?

SuperFluffy (approved):

Fraser999 (approved):

crates/astria-cli/src/cli/rollup.rs
sambukowski: Correct. Currently conductor can't pull block 0 or the genesis specifically, so needs to be 2 so that it can grab block 1 for validation.
I've also updated the validation function to pass on any parse errors. I added the `parsing to u64:` to the error formatting because testing showed that the parse errors didn't specify what type the value was actually trying to be parsed to.

Fraser999 (approved):

PR #932, opened 2024-04-08
feat(composer): add initial set of metrics

bharath-123:
## Summary
This PR setups and instruments metrics with Composer.

## Background
Composer currently doesn't track any metrics. We require metrics in production for observability and alerting.

## Changes
- Metrics setup for Composer crate

More changes detailed in the `Metrics` section

## Testing
Tested these changes by launching composer against messenger rollup and bombarding messenger rollup with txs. The metrics were accurate w.r.t to the load composer received.

## Metrics
- `transactions_dropped`: A counter of the  transactions dropped by the Collectors labelled per rollup id and collector type. This can be due to the Executor channel being closed or the channel has timed out.
- `transactions_received`: A counter of transactions received by the collectors labelled by rollup id and collector type
- `transactions_dropped_too_large`: A counter of transactions dropped because they were too large to fit in a bundle
- `nonce_fetch_count`: A counter of the number of times we have attempted to fetch the nonce
- `nonce_fetch_failure_count`: A counter of the number of times we have failed to fetch the nonce
- `nonce_fetch_latency`: A histogram of the latency of the nonce fetch call
- `current_nonce`: A gauge of the current nonce
- `sequencer_submission_latency`: A histogram of the latency of submitting a transaction to the sequencer
- `sequencer_submission_failure_count`: A counter of the number of failed transaction submission to the sequencer
- `transactions_per_submission`: A histogram of the number of rollup transactions successfully sent to the sequencer per submission
- `bytes_per_submission`: A histogram of the number of bytes successfully sent to the sequencer per submission

### DIFF
--- a/crates/astria-composer/src/collectors/geth.rs
+++ b/crates/astria-composer/src/collectors/geth.rs
@@ -51,8 +51,8 @@ use tracing::{
 
 use crate::{
     collectors::{
-        CollectorType,
         EXECUTOR_SEND_TIMEOUT,
+        GETH,
     },
     executor,
     metrics_init::{
@@ -218,7 +218,7 @@ impl Geth {
                             crate::metrics_init::TRANSACTIONS_RECEIVED,
                             &[
                                 (ROLLUP_ID_LABEL, chain_name.clone()),
-                                (COLLECTOR_TYPE_LABEL, CollectorType::Geth.to_string())
+                                (COLLECTOR_TYPE_LABEL, GETH.to_string())
                             ]).increment(1);
 
                         match executor_handle
@@ -236,7 +236,7 @@ impl Geth {
                                     crate::metrics_init::TRANSACTIONS_DROPPED,
                                     &[
                                         (ROLLUP_ID_LABEL, chain_name.clone()),
-                                        (COLLECTOR_TYPE_LABEL, CollectorType::Geth.to_string())
+                                        (COLLECTOR_TYPE_LABEL, GETH.to_string())
                                     ]
                                 ).increment(1);
                             }
@@ -250,7 +250,7 @@ impl Geth {
                                     crate::metrics_init::TRANSACTIONS_DROPPED,
                                     &[
                                         (ROLLUP_ID_LABEL, chain_name.clone()),
-                                        (COLLECTOR_TYPE_LABEL, CollectorType::Geth.to_string())
+                                        (COLLECTOR_TYPE_LABEL, GETH.to_string())
                                     ]
                                 ).increment(1);
                                 break Err(eyre!("executor channel closed while sending transaction"));
--- a/crates/astria-composer/src/collectors/geth.rs
+++ b/crates/astria-composer/src/collectors/geth.rs
@@ -215,7 +215,7 @@ impl Geth {
                         };
 
                         metrics::counter!(
-                            crate::metrics_init::TRANSACTIONS_COLLECTED,
+                            crate::metrics_init::TRANSACTIONS_RECEIVED,
                             &[
                                 (ROLLUP_ID_LABEL, chain_name.clone()),
                                 (COLLECTOR_TYPE_LABEL, CollectorType::Geth.to_string())
@@ -225,15 +225,7 @@ impl Geth {
                             .send_timeout(seq_action, EXECUTOR_SEND_TIMEOUT)
                             .await
                         {
-                            Ok(()) => {
-                                metrics::counter!(
-                                    crate::metrics_init::TRANSACTIONS_FORWARDED,
-                                    &[
-                                        (ROLLUP_ID_LABEL, chain_name.clone()),
-                                        (COLLECTOR_TYPE_LABEL, CollectorType::Geth.to_string())
-                                    ]
-                                ).increment(1);
-                            },
+                            Ok(()) => {},
                             Err(SendTimeoutError::Timeout(_seq_action)) => {
                                 warn!(
                                     transaction.hash = %tx_hash,
--- a/crates/astria-composer/src/collectors/geth.rs
+++ b/crates/astria-composer/src/collectors/geth.rs
@@ -52,7 +52,9 @@ use tracing::{
 use crate::{
     collectors::EXECUTOR_SEND_TIMEOUT,
     executor,
+    metrics_init::ROLLUP_ID_LABEL,
 };
+
 type StdError = dyn std::error::Error;
 
 const WSS_UNSUBSCRIBE_TIMEOUT: Duration = Duration::from_secs(2);
@@ -150,7 +152,7 @@ impl Geth {
             status,
             url,
             shutdown_token,
-            ..
+            chain_name,
         } = self;
 
         let retry_config = tryhard::RetryFutureConfig::new(1024)
@@ -171,6 +173,7 @@ impl Geth {
                 },
             );
 
+        let start = std::time::Instant::now();
         let client = tryhard::retry_fn(|| {
             let url = url.clone();
             async move {
@@ -186,6 +189,10 @@ impl Geth {
             .subscribe_full_pending_txs()
             .await
             .wrap_err("failed to subscribe eth client to full pending transactions")?;
+
+        metrics::histogram!(crate::metrics_init::GETH_COLLECTOR_CONNECTION_LATENCY,
+            ROLLUP_ID_LABEL => chain_name.clone())
+        .record(start.elapsed());
 
         status.send_modify(|status| status.is_connected = true);
 
@@ -206,17 +213,25 @@ impl Geth {
                             fee_asset_id: default_native_asset_id(),
                         };
 
+                        metrics::counter!(crate::metrics_init::GETH_COLLECTOR_TRANSACTIONS_COLLECTED,
+                            ROLLUP_ID_LABEL => chain_name.clone()).increment(1);
+
                         match executor_handle
                             .send_timeout(seq_action, EXECUTOR_SEND_TIMEOUT)
                             .await
                         {
-                            Ok(()) => {}
+                            Ok(()) => {
+                                metrics::counter!(crate::metrics_init::GETH_COLLECTOR_TRANSACTIONS_FORWARDED,
+                                    ROLLUP_ID_LABEL => chain_name.clone()).increment(1);
+                            },
                             Err(SendTimeoutError::Timeout(_seq_action)) => {
                                 warn!(
                                     transaction.hash = %tx_hash,
                                     timeout_ms = EXECUTOR_SEND_TIMEOUT.as_millis(),
                                     "timed out sending new transaction to executor; dropping tx",
                                 );
+                                metrics::counter!(crate::metrics_init::GETH_COLLECTOR_TRANSACTIONS_DROPPED,
+                                    ROLLUP_ID_LABEL => chain_name.clone()).increment(1);
                             }
                             Err(SendTimeoutError::Closed(_seq_action)) => {
                                 warn!(
@@ -224,6 +239,8 @@ impl Geth {
                                     "executor channel closed while sending transaction; dropping transaction \
                                      and exiting event loop"
                                 );
+                                metrics::counter!(crate::metrics_init::GETH_COLLECTOR_TRANSACTIONS_DROPPED,
+                                    ROLLUP_ID_LABEL => chain_name.clone()).increment(1);
                                 break Err(eyre!("executor channel closed while sending transaction"));
                             }
                         }
--- a/crates/astria-composer/src/collectors/geth.rs
+++ b/crates/astria-composer/src/collectors/geth.rs
@@ -50,9 +50,15 @@ use tracing::{
 };
 
 use crate::{
-    collectors::EXECUTOR_SEND_TIMEOUT,
+    collectors::{
+        CollectorType,
+        EXECUTOR_SEND_TIMEOUT,
+    },
     executor,
-    metrics_init::ROLLUP_ID_LABEL,
+    metrics_init::{
+        COLLECTOR_TYPE_LABEL,
+        ROLLUP_ID_LABEL,
+    },
 };
 
 type StdError = dyn std::error::Error;
@@ -173,7 +179,6 @@ impl Geth {
                 },
             );
 
-        let start = std::time::Instant::now();
         let client = tryhard::retry_fn(|| {
             let url = url.clone();
             async move {
@@ -189,10 +194,6 @@ impl Geth {
             .subscribe_full_pending_txs()
             .await
             .wrap_err("failed to subscribe eth client to full pending transactions")?;
-
-        metrics::histogram!(crate::metrics_init::GETH_COLLECTOR_CONNECTION_LATENCY,
-            ROLLUP_ID_LABEL => chain_name.clone())
-        .record(start.elapsed());
 
         status.send_modify(|status| status.is_connected = true);
 
@@ -213,16 +214,25 @@ impl Geth {
                             fee_asset_id: default_native_asset_id(),
                         };
 
-                        metrics::counter!(crate::metrics_init::GETH_COLLECTOR_TRANSACTIONS_COLLECTED,
-                            ROLLUP_ID_LABEL => chain_name.clone()).increment(1);
+                        metrics::counter!(
+                            crate::metrics_init::TRANSACTIONS_COLLECTED,
+                            &[
+                                (ROLLUP_ID_LABEL, chain_name.clone()),
+                                (COLLECTOR_TYPE_LABEL, CollectorType::Geth.to_string())
+                            ]).increment(1);
 
                         match executor_handle
                             .send_timeout(seq_action, EXECUTOR_SEND_TIMEOUT)
                             .await
                         {
                             Ok(()) => {
-                                metrics::counter!(crate::metrics_init::GETH_COLLECTOR_TRANSACTIONS_FORWARDED,
-                                    ROLLUP_ID_LABEL => chain_name.clone()).increment(1);
+                                metrics::counter!(
+                                    crate::metrics_init::TRANSACTIONS_FORWARDED,
+                                    &[
+                                        (ROLLUP_ID_LABEL, chain_name.clone()),
+                                        (COLLECTOR_TYPE_LABEL, CollectorType::Geth.to_string())
+                                    ]
+                                ).increment(1);
                             },
                             Err(SendTimeoutError::Timeout(_seq_action)) => {
                                 warn!(
@@ -230,8 +240,13 @@ impl Geth {
                                     timeout_ms = EXECUTOR_SEND_TIMEOUT.as_millis(),
                                     "timed out sending new transaction to executor; dropping tx",
                                 );
-                                metrics::counter!(crate::metrics_init::GETH_COLLECTOR_TRANSACTIONS_DROPPED,
-                                    ROLLUP_ID_LABEL => chain_name.clone()).increment(1);
+                                metrics::counter!(
+                                    crate::metrics_init::TRANSACTIONS_DROPPED,
+                                    &[
+                                        (ROLLUP_ID_LABEL, chain_name.clone()),
+                                        (COLLECTOR_TYPE_LABEL, CollectorType::Geth.to_string())
+                                    ]
+                                ).increment(1);
                             }
                             Err(SendTimeoutError::Closed(_seq_action)) => {
                                 warn!(
@@ -239,8 +254,13 @@ impl Geth {
                                     "executor channel closed while sending transaction; dropping transaction \
                                      and exiting event loop"
                                 );
-                                metrics::counter!(crate::metrics_init::GETH_COLLECTOR_TRANSACTIONS_DROPPED,
-                                    ROLLUP_ID_LABEL => chain_name.clone()).increment(1);
+                                metrics::counter!(
+                                    crate::metrics_init::TRANSACTIONS_DROPPED,
+                                    &[
+                                        (ROLLUP_ID_LABEL, chain_name.clone()),
+                                        (COLLECTOR_TYPE_LABEL, CollectorType::Geth.to_string())
+                                    ]
+                                ).increment(1);
                                 break Err(eyre!("executor channel closed while sending transaction"));
                             }
                         }
--- a/crates/astria-composer/src/collectors/grpc.rs
+++ b/crates/astria-composer/src/collectors/grpc.rs
@@ -23,8 +23,8 @@ use tonic::{
 
 use crate::{
     collectors::{
-        CollectorType,
         EXECUTOR_SEND_TIMEOUT,
+        GRPC,
     },
     executor,
     metrics_init::{
@@ -70,7 +70,7 @@ impl GrpcCollectorService for Grpc {
             crate::metrics_init::TRANSACTIONS_RECEIVED,
             &[
                 (ROLLUP_ID_LABEL, rollup_id.to_string()),
-                (COLLECTOR_TYPE_LABEL, CollectorType::Grpc.to_string())
+                (COLLECTOR_TYPE_LABEL, GRPC.to_string())
             ]
         )
         .increment(1);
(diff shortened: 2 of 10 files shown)


crates/astria-composer/src/executor/bundle_factory/mod.rs
```diff
@@ -154,6 +169,16 @@ impl BundleFactory {
             }
             Err(SizedBundleError::NotEnoughSpace(seq_action)) => {
                 // if the bundle is full, flush it and start a new one
+                metrics::counter!(crate::metrics_init::BUNDLES_TOTAL_COUNT).increment(1);
+
+                #[allow(clippy::cast_precision_loss)]
```
bharath-123, line 174: this looks a bit odd. I am not sure if there is a much more cleaner way to do this?

noot: you can to `self.curr_bundle.get_size().try_into()` or perhaps `TryInto::<f64>::try_into(self.curr_bundle.get_size())`  since you might need to specify the type   - do you expect the current bundle size to always fit into a f64 or no? same for the other values being casted

bharath-123: it should fit into a 64 byte number for sure!

crates/astria-composer/src/metrics_init.rs
```diff
@@ -0,0 +1,234 @@
...
+
+use metrics::{
+    describe_counter,
+    describe_gauge,
+    describe_histogram,
+    Unit,
+};
+
+/// Labels
+pub(crate) const ROLLUP_ID_LABEL: &str = "rollup_id";
```
SuperFluffy, line 13: I feel like this can be removed for the time being.

crates/astria-composer/src/metrics_init.rs
```diff
@@ -0,0 +1,234 @@
...
+);
+
+pub const BUNDLES_OUTGOING_BYTES: &str =
+    concat!(env!("CARGO_CRATE_NAME"), "_bundles_outgoing_bytes");
+
+pub const BUNDLES_NOT_DRAINED: &str = concat!(env!("CARGO_CRATE_NAME"), "_bundles_not_drained");
+
+pub const BUNDLES_DRAINED: &str = concat!(env!("CARGO_CRATE_NAME"), "_bundles_drained");
+
+pub const BUNDLES_TOTAL_COUNT: &str = concat!(env!("CARGO_CRATE_NAME"), "_bundles_total_count");
```
SuperFluffy, line 227: This is closely related to the number of bundles submitted to the sequencer. Remove

crates/astria-composer/src/metrics_init.rs
```diff
@@ -0,0 +1,234 @@
...
+    env!("CARGO_CRATE_NAME"),
+    "_bundles_outgoing_transactions_count"
+);
+
+pub const BUNDLES_OUTGOING_BYTES: &str =
+    concat!(env!("CARGO_CRATE_NAME"), "_bundles_outgoing_bytes");
+
+pub const BUNDLES_NOT_DRAINED: &str = concat!(env!("CARGO_CRATE_NAME"), "_bundles_not_drained");
+
+pub const BUNDLES_DRAINED: &str = concat!(env!("CARGO_CRATE_NAME"), "_bundles_drained");
```
SuperFluffy, line 225: This is not really useful either, it's not a runtime metric.

crates/astria-composer/src/metrics_init.rs
```diff
@@ -0,0 +1,234 @@
...
+
+pub const BUNDLES_OUTGOING_TRANSACTIONS_COUNT: &str = concat!(
+    env!("CARGO_CRATE_NAME"),
+    "_bundles_outgoing_transactions_count"
+);
+
+pub const BUNDLES_OUTGOING_BYTES: &str =
+    concat!(env!("CARGO_CRATE_NAME"), "_bundles_outgoing_bytes");
+
+pub const BUNDLES_NOT_DRAINED: &str = concat!(env!("CARGO_CRATE_NAME"), "_bundles_not_drained");
```
SuperFluffy, line 223: remove

crates/astria-composer/src/metrics_init.rs
```diff
@@ -0,0 +1,234 @@
...
+
+pub const BUNDLES_DRAINED: &str = concat!(env!("CARGO_CRATE_NAME"), "_bundles_drained");
+
+pub const BUNDLES_TOTAL_COUNT: &str = concat!(env!("CARGO_CRATE_NAME"), "_bundles_total_count");
+
+pub const BUNDLES_TOTAL_BYTES: &str = concat!(env!("CARGO_CRATE_NAME"), "_bundles_total_bytes");
+
+pub const BUNDLES_TOTAL_TRANSACTIONS_COUNT: &str = concat!(
```
SuperFluffy, line 231: Very highly correlated to the number of bundles submitted to the sequencer. Remove

crates/astria-composer/src/metrics_init.rs
```diff
@@ -0,0 +1,234 @@
...
+pub const BUNDLES_OUTGOING_BYTES: &str =
+    concat!(env!("CARGO_CRATE_NAME"), "_bundles_outgoing_bytes");
+
+pub const BUNDLES_NOT_DRAINED: &str = concat!(env!("CARGO_CRATE_NAME"), "_bundles_not_drained");
+
+pub const BUNDLES_DRAINED: &str = concat!(env!("CARGO_CRATE_NAME"), "_bundles_drained");
+
+pub const BUNDLES_TOTAL_COUNT: &str = concat!(env!("CARGO_CRATE_NAME"), "_bundles_total_count");
+
+pub const BUNDLES_TOTAL_BYTES: &str = concat!(env!("CARGO_CRATE_NAME"), "_bundles_total_bytes");
```
SuperFluffy, line 229: high correlation to the number of submitted bundles

crates/astria-composer/src/metrics_init.rs
```diff
@@ -0,0 +1,234 @@
...
+         lifetime"
+    );
+}
+
+pub const GETH_COLLECTOR_TRANSACTIONS_COLLECTED: &str = concat!(
+    env!("CARGO_CRATE_NAME"),
+    "_geth_collector_transactions_collected"
+);
+
+pub const GETH_COLLECTOR_TRANSACTIONS_DROPPED: &str = concat!(
```
SuperFluffy, line 153: The distinction between transactions dropped from grpc vs geth collector is not very interesting - both feed into the pipeline, so just the overall `TRANSACTIONS_DROPPED` is interesting. If you want to distinguish here, add a different label.

crates/astria-composer/src/metrics_init.rs
```diff
@@ -0,0 +1,234 @@
...
+    );
+    describe_histogram!(
+        BUNDLES_TOTAL_TRANSACTIONS_COUNT,
+        Unit::Count,
+        "The total number of transactions in finished bundles constructed over the composer's \
+         lifetime"
+    );
+}
+
+pub const GETH_COLLECTOR_TRANSACTIONS_COLLECTED: &str = concat!(
```
SuperFluffy, line 148: IMO this should be `TRANSACTIONS_COLLECTED`, but potentially distinguished with a label.

crates/astria-composer/src/metrics_init.rs
```diff
@@ -0,0 +1,234 @@
...
+        "The number of times we have failed to fetch the nonce"
+    );
+    describe_histogram!(
+        NONCE_FETCH_LATENCY,
+        Unit::Milliseconds,
+        "The latency of nonce fetch"
+    );
+    describe_gauge!(CURRENT_NONCE, Unit::Count, "The current nonce");
+    describe_histogram!(
+        TRANSACTION_SUBMISSION_LATENCY,
```
SuperFluffy, line 89: I find the distinction between `bundle` and `transaction` in this context very confusing. We should either talk about "bundle (of transcations)" or just "transactions" everywhere.

crates/astria-composer/src/executor/bundle_factory/mod.rs
```diff
@@ -81,6 +83,9 @@ impl SizedBundle {
         let seq_action_size = estimate_size_of_sequence_action(&seq_action);
 
         if seq_action_size > self.max_size {
+            metrics::gauge!(crate::metrics_init::TRANSACTIONS_DROPPED_TOO_LARGE,
```
SuperFluffy, line 86: IMO this metric is too deep down in the stack. I would emit it at the level that is handling this error

crates/astria-composer/src/executor/mod.rs
```diff
@@ -174,6 +174,14 @@ impl Executor {
     /// Create a future to submit a bundle to the sequencer.
     #[instrument(skip_all, fields(nonce.initial = %nonce))]
     fn submit_bundle(&self, nonce: u32, bundle: SizedBundle) -> Fuse<Instrumented<SubmitFut>> {
+        #[allow(clippy::cast_precision_loss)]
+        metrics::histogram!(crate::metrics_init::BUNDLES_OUTGOING_BYTES)
+            .record(bundle.get_size() as f64);
```
SuperFluffy, line 179: These should be consistently emitted after a submission was successful

SuperFluffy:
High level feedback: I think the metrics are too granular at this stage.

crates/astria-composer/src/executor/bundle_factory/mod.rs
SuperFluffy: `clippy::cast_precision_loss` is fine. Just add a comment above why you allow it.

We never want to fail on metrics. It's not material

crates/astria-composer/src/collectors/geth.rs
```diff
@@ -185,6 +188,10 @@ impl Geth {
             .await
             .wrap_err("failed to subscribe eth client to full pending transactions")?;
 
+        metrics::histogram!(crate::metrics_init::GETH_COLLECTOR_CONNECTION_LATENCY,
```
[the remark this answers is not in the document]
bharath-123: we might have geth collector reconnects. but that too may not be too frequent hmm

crates/astria-composer/src/collectors/geth.rs
```diff
@@ -148,7 +150,7 @@ impl Geth {
             status,
             url,
             shutdown_token,
-            ..
+            chain_name,
```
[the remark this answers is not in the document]
bharath-123: sure will use the `leak`

crates/astria-composer/src/collectors/grpc.rs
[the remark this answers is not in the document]
bharath-123: the labels won't be useful for bundling related metrics but only sequence action related metrics. i ll add it for the dropped part.

crates/astria-composer/src/metrics_init.rs
bharath-123: prometheus could scrape these metrics since the composer won't shutdown immediately. but I can live with this info being in tracing for now.

crates/astria-composer/src/metrics_init.rs
bharath-123: not necessarily right? some bundles might be failed to be submitted because of some error with the sequencer(like bundle too large or so). It would be better if we can know how many have been created and how many have been totally submitted. I think we can just let go of the outgoing_bytes stuff and just have succesfully submitted bundles.

crates/astria-composer/src/metrics_init.rs
bharath-123: i think we should keep it for metrics to segregate the number of sequence actions submitted per rollup.

crates/astria-composer/src/collectors/geth.rs
```diff
@@ -204,24 +211,34 @@ impl Geth {
                             fee_asset_id: default_native_asset_id(),
                         };
 
+                        metrics::counter!(crate::metrics_init::GETH_COLLECTOR_TRANSACTIONS_COLLECTED,
+                            ROLLUP_ID_LABEL => chain_name.clone()).increment(1);
```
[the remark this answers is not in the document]
bharath-123: will do it in a future PR

crates/astria-composer/src/executor/mod.rs
bharath-123: cool, i ll emit them whenever there is a successful submission.

crates/astria-composer/src/metrics_init.rs
bharath-123: I have done the following:
Modify transaction_submission_latency to sequencer_submission_latency and transaction_failure_count to sequencer_failure_count. With this, the Transaction word is now only reserved for rollup transactions

crates/astria-composer/src/executor/bundle_factory/mod.rs
bharath-123: Have moved it to a higher level.

crates/astria-composer/src/collectors/geth.rs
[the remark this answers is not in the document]
bharath-123: removed for now. I do feel like with frequent re-connects we might want the data but we can add that sometime later.

itamarreif (approved):
Very nice - matches my understanding based on the doc we made. I tend to agree with Bharath that keeping track of both the total counts (of bytes, bundles, sequencer actions) and the submission counts is useful since we might want to explicitly see the difference there.

itamarreif (approved):

itamarreif (approved):
Looks good and matches my understanding based on the doc we made last week. I also tend to agree with Bharath's decision to maintain both the total counts and submission counts, since we'd be interested in monitoring the difference between the two.

crates/astria-composer/src/collectors/grpc.rs
```diff
@@ -59,18 +66,53 @@ impl GrpcCollectorService for Grpc {
             fee_asset_id: default_native_asset_id(),
         };
 
+        metrics::counter!(
+            crate::metrics_init::TRANSACTIONS_COLLECTED,
+            &[
+                (ROLLUP_ID_LABEL, rollup_id.to_string()),
```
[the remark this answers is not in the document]
bharath-123: this was part of a refactor by jordan to call things chain_id instead of rollup_id. it was just more natural to do this for now.

SuperFluffy (changes requested):
Reading the name of these metrics I am confused about what they represent. Their descriptions also don't provide much clarity.

Also, I recommend removing the concept of a "bundle" entirely, because that's a composer internal concept that keeps leaking out. This concept can be entirely covered by (without introducing new nomenclature):

1. Submissions to sequencer (keyword `submission` or `sequencer_submission`)
2. Rollup transactions (keyword `transactions` or `rollup_transactions`)

The metrics would then have clearer names like the ones below (also, `x_per_y` is something that we are using in sequencer-relayer):

1. `transactions_per_submission`
2. `total_transactions_submitted`
3. `bytes_per_submission`
4. `total_bytes_submitted`


There are a couple of highly correlated metrics which descriptions that make them sound the same. These groups should be rolled into 1 metric each:

1. `TRANSACTIONS_COLLECTED`, `TRANSACTIONS_RECEIVED`, `TRANSACTIONS_FORWARDED`; these should be flattened into 1.
2. `SEQUENCER_SUBMISSION_FAILURE_COUNT`, `BUNDLES_SUBMISSION_FAILURE_COUNT`.

crates/astria-composer/src/metrics_init.rs
```diff
@@ -0,0 +1,119 @@
...
+        Unit::Count,
+        "The total number of rollup transactions sent to the sequencer over the composer's \
+         lifetime. These can include failed submissions also"
+    );
+    describe_histogram!(
+        TOTAL_BYTES_SUBMITTED,
+        Unit::Bytes,
+        "The total bytes sent to the sequencer over the composer's lifetime. These can include \
+         failed submissions also"
+    );
```
bharath-123, line 84: Note: These include failed submissions too. It ll be useful to know the difference b/w how many transactions and bytes were successfully submitted vs those which were not.

crates/astria-composer/src/metrics_init.rs
```diff
@@ -0,0 +1,119 @@
...
+
+pub const SEQUENCER_SUBMISSION_FAILURE_COUNT: &str = concat!(
+    env!("CARGO_CRATE_NAME"),
+    "_sequencer_submission_failure_count"
+);
+
+pub const TRANSACTIONS_PER_SUBMISSION: &str =
+    concat!(env!("CARGO_CRATE_NAME"), "_transaction_per_submission");
+
+pub const BYTES_PER_SUBMISSION: &str = concat!(env!("CARGO_CRATE_NAME"), "_bytes_per_submission");
```
bharath-123, line 114: From the total count and sum values of the histogram generated by these metrics, we can derive the total number of transactions and bytes submitted to the sequencer over the composer's lifetime.

crates/astria-composer/src/executor/mod.rs
```diff
@@ -553,6 +590,22 @@ impl Future for SubmitFut {
...
+                            metrics::gauge!(crate::metrics_init::CURRENT_NONCE)
+                                .set(*this.nonce + 1);
+
+                            // allow: the number of bytes in a bundle will be large enough to not
+                            // cause a precision loss
+                            #[allow(clippy::cast_precision_loss)]
+                            metrics::histogram!(crate::metrics_init::BYTES_PER_SUBMISSION)
+                                .record(this.bundle.get_size() as f64);
+
+                            // allow: a bundle will have at least 1 transaction which will not cause
```
SuperFluffy, line 603: The precision loss is not important for small `usize`, but for large. But the main reason for this is:
```suggestion
                            // allow: precision loss is unlikely (values too small) but also unimportant in histograms
```

crates/astria-composer/src/executor/mod.rs
```diff
@@ -553,6 +590,22 @@ impl Future for SubmitFut {
                     Ok(rsp) => {
                         let tendermint::abci::Code::Err(code) = rsp.code else {
                             info!("sequencer responded with ok; submission successful");
+
+                            metrics::gauge!(crate::metrics_init::CURRENT_NONCE)
+                                .set(*this.nonce + 1);
+
+                            // allow: the number of bytes in a bundle will be large enough to not
```
SuperFluffy, line 597: The precision loss is not important for small `usize`, but for large. But the main reason for this is:
```suggestion
                            // allow: precision loss is unlikely (values too small) but also unimportant in histograms
```

crates/astria-composer/src/executor/mod.rs
```diff
@@ -553,6 +590,22 @@ impl Future for SubmitFut {
                     Ok(rsp) => {
                         let tendermint::abci::Code::Err(code) = rsp.code else {
                             info!("sequencer responded with ok; submission successful");
+
+                            metrics::gauge!(crate::metrics_init::CURRENT_NONCE)
```
SuperFluffy, line 594: If I read this right `CURRENT_NONCE` is only set after the `WaitingForSend` fut completes (or fails). IMO it should be set at the beginning of each attempt to submit a future.

Please move the emission of this metric into `submit_tx` - that way you will get the gauge also after `WaitingForNonce` completes and a new submission is attempted.

crates/astria-composer/src/collectors/mod.rs
```diff
@@ -1,9 +1,30 @@
...
+        Formatter,
+    },
+    time::Duration,
+};
+
 const EXECUTOR_SEND_TIMEOUT: Duration = Duration::from_millis(500);
 
-use std::time::Duration;
+#[derive(Debug)]
+pub(crate) enum CollectorType {
```
SuperFluffy, line 15: This has no advantage over using ` const COLLECTOR_TYPE: &str` in `geth` and `grpc`, respectively. Just do the const, it's simpler.

crates/astria-composer/src/collectors/geth.rs
```diff
@@ -206,24 +214,45 @@ impl Geth {
                             fee_asset_id: default_native_asset_id(),
                         };
 
+                        metrics::counter!(
+                            crate::metrics_init::TRANSACTIONS_RECEIVED,
+                            &[
+                                (ROLLUP_ID_LABEL, chain_name.clone()),
+                                (COLLECTOR_TYPE_LABEL, CollectorType::Geth.to_string())
```
SuperFluffy, line 221: There is no advantage using a `CollectorType` enum over a hard coded `const COLLECTOR_TYPE: &str = "geth";`

Just replace this with a (module-private) const that's also `'static` so that you don't have to allocate here.

crates/astria-composer/src/collectors/grpc.rs
```diff
@@ -59,18 +66,44 @@ impl GrpcCollectorService for Grpc {
             fee_asset_id: default_native_asset_id(),
         };
 
+        metrics::counter!(
+            crate::metrics_init::TRANSACTIONS_RECEIVED,
+            &[
+                (ROLLUP_ID_LABEL, rollup_id.to_string()),
+                (COLLECTOR_TYPE_LABEL, CollectorType::Grpc.to_string())
```
SuperFluffy, line 73: Similar to my comment for `geth`: replace this with a `const`

crates/astria-composer/src/executor/bundle_factory/mod.rs
```diff
@@ -106,11 +106,21 @@ impl SizedBundle {
...
+        self.curr_size
+    }
+
     /// Consume self and return the underlying buffer of actions.
     pub(super) fn into_actions(self) -> Vec<Action> {
         self.buffer
     }
 
+    /// Returns the number of sequence actions in the bundle.
+    pub(super) fn no_of_seq_actions(&self) -> usize {
```
SuperFluffy, line 120: Hate the nomenclature, but let's go with it because the other parts of `bundle_factory` (we have `into_actions` and other methods, so let's not introduce `seq` because this might imply a difference between the `actions` in `into_actions` and `seq_actions`):
```suggestion
    pub(super) fn no_of_actions(&self) -> usize {
```

Also IMO this is cleaner to avoid abbreviations (write once, read many times):
```suggestion
    pub(super) fn actions_count(&self) -> usize {
```
or
```suggestion
    pub(super) fn number_of_actions(&self) -> usize {
```

crates/astria-composer/src/executor/mod.rs
```diff
@@ -345,15 +363,17 @@ impl Executor {
...
                         );
+
                         nonce = new_nonce;
                         bundles_drained += 1;
                     }
                     Err(error) => {
-                        error!(bundle = %telemetry::display::json(&SizedBundleReport(&bundle)),
-                                %error, "failed submitting bundle to sequencer during shutdown; \
-                                    aborting shutdown");
+                        error!(bundle =  %telemetry::display::json(&SizedBundleReport(&bundle)),
```
SuperFluffy, line 371: Manual format becasue Rustfmt seems to fail:
```suggestion
                        error!(
                            bundle =  %telemetry::display::json(&SizedBundleReport(&bundle)),
                            %error,
                            "failed submitting bundle to sequencer during shutdown; \
                             aborting shutdown",
                        );
```

crates/astria-composer/src/executor/mod.rs
```diff
@@ -521,6 +557,7 @@ pin_project! {
 impl Future for SubmitFut {
     type Output = eyre::Result<u32>;
 
+    #[allow(clippy::too_many_lines)]
```
SuperFluffy, line 560: Can you make a tracking issue for us to refactor this? So we don't lose track, i.e.:
```suggestion
    // allow: refactor this future github.com/astriaorg/astria/issues/{number}
    #[allow(clippy::too_many_lines)]
```

SuperFluffy (approved):
Looks great, thank you!

Before you merge, please move construction of the `CURRENT_NONCE` gauge into `submit_tx` - that way we will always have an accurate reading of the nonce that's actually used, and we then only have to emit the metric at a single point.

Also the `CollectorType` is unnecessary - just use a simpler const.

crates/astria-composer/src/executor/bundle_factory/mod.rs
bharath-123: actions_count is good!

PR #935, opened 2024-04-09
fix(sequencer): set revision number from chain id in `init_chain`

noot:
## Summary
set revision number from chain id in `init_chain` 

## Background
we need to set the chain's revision number from the last number of the chain id, if applicable. otherwise, it's set to 0. see https://ibc.cosmos.network/main/ibc/overview#ibc-client-heights

## Changes
- set revision number from chain id in `init_chain` by parsing final integer of chain id if it's in the form `name-<revision-number>`

## Testing
unit tests

### DIFF
--- a/crates/astria-sequencer/src/api_state_ext.rs
+++ b/crates/astria-sequencer/src/api_state_ext.rs
@@ -381,17 +381,10 @@ mod test {
 
     use std::collections::HashMap;
 
-    use astria_core::{
-        sequencer::v1::{
-            asset::Id,
-            block::{
-                Deposit,
-                SequencerBlock,
-            },
-            Address,
-            RollupId,
-        },
-        Protobuf,
+    use astria_core::sequencer::v1::{
+        asset::Id,
+        block::Deposit,
+        Address,
     };
     use cnidarium::StateDelta;
     use rand::Rng;
--- a/crates/astria-sequencer/src/state_ext.rs
+++ b/crates/astria-sequencer/src/state_ext.rs
@@ -55,9 +55,8 @@ pub(crate) trait StateReadExt: StateRead {
             bail!("revision number not found in state");
         };
 
-        let Ok(bytes): Result<[u8; 8], _> = bytes.try_into() else {
-            bail!("failed turning raw revision number bytes into u64; not 8 bytes?");
-        };
+        let bytes = TryInto::<[u8; 8]>::try_into(bytes)
+            .map_err(|b| anyhow::anyhow!("expected 8 revision number bytes but got {}; this is a bug", b.len()))?;
 
         Ok(u64::from_be_bytes(bytes))
     }
@@ -179,16 +178,6 @@ pub(crate) trait StateReadExt: StateRead {
 
 impl<T: StateRead> StateReadExt for T {}
 
-fn revision_number_from_chain_id(chain_id: &str) -> u64 {
-    let parts = chain_id.split("-").collect::<Vec<_>>();
-    if let Some(revision_number) = parts.last() {
-        if let Ok(revision_number) = revision_number.parse::<u64>() {
-            return revision_number;
-        }
-    };
-    0
-}
-
 #[async_trait]
 pub(crate) trait StateWriteExt: StateWrite {
     #[instrument(skip(self))]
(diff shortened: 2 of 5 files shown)


crates/astria-sequencer/src/state_ext.rs
```diff
@@ -46,8 +47,19 @@ pub(crate) trait StateReadExt: StateRead {
...
-        Ok(0)
+        let Some(bytes) = self
+            .get_raw(REVISION_NUMBER_KEY)
+            .await
+            .context("failed to read raw revision number from state")?
+        else {
+            bail!("revision number not found in state");
+        };
+
+        let Ok(bytes): Result<[u8; 8], _> = bytes.try_into() else {
```
SuperFluffy, line 58: You can report the actual number of bytes (`map_err` with `anyhow` instead of `context` to get access to to the `Vec` returned by `try_from`: https://doc.rust-lang.org/stable/std/primitive.array.html#method.try_from-5)

```rust
let bytes = TryInto::<[u8; 8]>::try_into(bytes)
    .map_err(|b| anyhow!("expected 8 revision number bytes but got{}", b.len()))?;
```

crates/astria-sequencer/src/state_ext.rs
```diff
@@ -167,11 +179,31 @@ pub(crate) trait StateReadExt: StateRead {
 
 impl<T: StateRead> StateReadExt for T {}
 
+fn revision_number_from_chain_id(chain_id: &str) -> u64 {
+    let parts = chain_id.split("-").collect::<Vec<_>>();
```
SuperFluffy, line 183: You don't need this allocation, you can use this iterator adaptor: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.last

Your line below then becomes:

```rust
if let Some(rev_number) = chain_id.split("-").last() {
```

crates/astria-sequencer/src/state_ext.rs
```diff
@@ -167,11 +179,31 @@ pub(crate) trait StateReadExt: StateRead {
 
 impl<T: StateRead> StateReadExt for T {}
 
+fn revision_number_from_chain_id(chain_id: &str) -> u64 {
```
SuperFluffy, line 182: If I supply a chain ID `123`, won't this function parse the wrong thing? Alternatively, what about using regex groups to be stricter about how this is parsed?

Can you add unit tests for the cases `123` and `123-123`?

crates/astria-sequencer/src/state_ext.rs
noot: removed in favour of regex

crates/astria-sequencer/src/state_ext.rs
noot: removed in favour of regex and added tests

crates/astria-sequencer/src/api_state_ext.rs
```diff
@@ -381,17 +381,10 @@ mod test {
...
-                SequencerBlock,
-            },
-            Address,
-            RollupId,
-        },
-        Protobuf,
+    use astria_core::sequencer::v1::{
+        asset::Id,
+        block::Deposit,
+        Address,
```
joroshiba, line 387: Why do we have changes here, when no change in the rest of file?

joroshiba (approved):

crates/astria-sequencer/src/api_state_ext.rs
noot: compiler was giving me unused import warnings so i just removed them

crates/astria-sequencer/src/state_ext.rs
```diff
@@ -244,18 +269,56 @@ pub(crate) trait StateWriteExt: StateWrite {
 
 impl<T: StateWrite> StateWriteExt for T {}
 
+fn revision_number_from_chain_id(chain_id: &str) -> u64 {
```
SuperFluffy, line 272: However unlikely, I think this could do with a `anyhow::Result<Option<u64>>` instead of panicking.

For the error case (u64 overflow): `init_chain` allows returning an error, so returning it is not a big deal.

For `Option::None`: the code should emit a event that no revision was found and is hence assumed 0

SuperFluffy (commented):
Just a comment. Probably not a big deal but nice to have.

crates/astria-sequencer/src/state_ext.rs
noot: init_chain erroring is effectively the same as panicking as the node will shut down and not be able to initialize. will add a log for the no revision number case tho

PR #937, opened 2024-04-09
fix(composer): add capacity to bundle factory

itamarreif:
## Summary
This adds logic for setting the `BundleFactory`'s `finished_queue` capacity and for the `Executor` to stop pulling transactions from the `serialized_rollup_transactions` channel while its `BundleFactory` is full

## Background
The `Executor`'s `BundleFactory` has a queue of `SizedBundle`s that it builds as transactions are collected. This was previously uncapped, potentially leading to OOM problems (e.g. k8s shutting down the process). 

## Changes
- Added config variable for `finished_queue_capacity`
- Added variable to constructors for `Executor` and `BundleFactory`
- `BundleFactory::try_push` can now return `BundleFactoryError:FinishedQueueFull`
- The `BundleFactory` now tracks whether it is full or not.
- `Executor` will only pull from the channel if it has available capacity in its `BundleFactory`.

## Testing
- sanity check to see that the `BundleFactory` is indeed marked full when it should be
- test that `BundleFactory::next_finished().pop()` on a full bundle sets it to no longer full
- test that `BundleFactory::pop_now` on a full bundle sets it to no longer full

## Breaking Changelist
- `Executor` will only pull from the channel if it has available capacity in its `BundleFactory`. This will cause the collectors' `try_send` to time out when back pressure accumulates in the system, dropping the collected transactions.

## Related Issues

closes


crates/astria-composer/local.env.example
```diff
@@ -47,6 +47,13 @@ ASTRIA_COMPOSER_MAX_SUBMIT_INTERVAL_MS=2000
 # key and nonce bytes
 ASTRIA_COMPOSER_MAX_BYTES_PER_BUNDLE=200000
 
+# Max amount of `SizedBundle`s to allow to accrue in the `BundleFactory`'s finished queue.
+# ASTRIA_COMPOSER_FINISHED_QUEUE_CAPACITY * ASTRIA_COMPOSER_MAX_BYTES_PER_BUNDLE (e.g.
+# 10 * 200KB) is the limit on how much memory the finished bundle queue can consume.
+# This should be lower than the resource limit enforced by Kubernetes on the pod, defined here:
+# https://github.com/astriaorg/astria/blob/622d4cb8695e4fbcd86456bd16149420b8acda79/charts/evm-rollup/values.yaml#L276
+ASTRIA_COMPOSER_FINISHED_QUEUE_CAPACITY=10
```
bharath-123, line 55: you should add this in the charts directory for CI. https://github.com/astriaorg/astria/blob/0124b5723d980dabb9e9928eca3c6436f785122b/charts/evm-rollup/templates/configmap.yaml#L83

bharath-123: 10 might be a bit too small? i think we can calibrate the numbers during testing.

crates/astria-composer/src/executor/bundle_factory/mod.rs
```diff
@@ -153,14 +184,34 @@ impl BundleFactory {
...
+                    // if the finished queue is full after the flush mark it as such
+                    if self.finished.len() == self.finished_queue_capacity {
+                        self.full = true;
+                    }
+
+                    let sized_seq_action = self.curr_bundle.push_check(seq_action).expect(
+                        "seq_action should not be larger than max bundle size, this is a bug",
+                    );
+                    self.curr_bundle.push(sized_seq_action);
+                    trace!(
```
bharath-123, line 203: make this an info! trace?

crates/astria-composer/src/executor/bundle_factory/mod.rs
```diff
@@ -111,10 +123,19 @@ impl SizedBundle {
     }
 }
 
+pub(super) struct SizedSequenceAction(SequenceAction);
```
bharath-123, line 126: why do we require an additional struct? i don't see any methods defined for it per say?

crates/astria-composer/src/executor/bundle_factory/mod.rs
```diff
@@ -111,10 +123,19 @@ impl SizedBundle {
...
 #[derive(Debug, thiserror::Error)]
 pub(super) enum BundleFactoryError {
     #[error("sequence action is larger than the max bundle size. seq_action size: {size}")]
     SequenceActionTooLarge { size: usize, max_size: usize },
+    #[error(
+        "finished bundle queue is full and the sequence action does not fit in the current bundle"
+    )]
+    FinishedQueueFull {
+        curr_bundle_size: usize,
+        seq_action: SequenceAction,
```
bharath-123, line 137: i think we should return the current size of the finished queue too here.

crates/astria-composer/src/executor/bundle_factory/mod.rs
```diff
@@ -189,10 +240,17 @@ impl BundleFactory {
...
             .pop_front()
             .or_else(|| Some(self.curr_bundle.flush()))
-            .unwrap_or(SizedBundle::new(self.curr_bundle.max_size))
+            .unwrap_or(SizedBundle::new(self.curr_bundle.max_size));
+        self.full = false;
+        bundle
+    }
+
+    pub(super) fn full(&self) -> bool {
+        self.full
```
bharath-123, line 253: can we just use `self.finished.len() >= self.finished_queue_capacity` here? rather than maintaining a seperate `self.full` variable?

crates/astria-composer/src/executor/bundle_factory/mod.rs
```diff
@@ -153,14 +184,34 @@ impl BundleFactory {
...
+                    Err(BundleFactoryError::FinishedQueueFull {
+                        curr_bundle_size: self.curr_bundle.curr_size,
+                        seq_action,
+                    })
+                } else {
+                    self.finished.push_back(self.curr_bundle.flush());
+                    // if the finished queue is full after the flush mark it as such
+                    if self.finished.len() == self.finished_queue_capacity {
+                        self.full = true;
+                    }
```
bharath-123, line 197: I think we can just remove this logic here. If we define `full()` with `self.finished.len() > self.finished_queue_capacity`

crates/astria-composer/src/executor/bundle_factory/mod.rs
```diff
@@ -153,14 +184,34 @@ impl BundleFactory {
...
+
+                    let sized_seq_action = self.curr_bundle.push_check(seq_action).expect(
+                        "seq_action should not be larger than max bundle size, this is a bug",
+                    );
```
bharath-123, line 201: why are we doing this again? i think we enter this code block after succeeding this check right?

crates/astria-composer/src/executor/bundle_factory/mod.rs
```diff
@@ -189,10 +240,17 @@ impl BundleFactory {
...
+            .finished
             .pop_front()
             .or_else(|| Some(self.curr_bundle.flush()))
-            .unwrap_or(SizedBundle::new(self.curr_bundle.max_size))
+            .unwrap_or(SizedBundle::new(self.curr_bundle.max_size));
+        self.full = false;
+        bundle
+    }
+
+    pub(super) fn full(&self) -> bool {
```
bharath-123, line 252:
```suggestion
    pub(super) fn is_full(&self) -> bool {
```

crates/astria-composer/src/executor/bundle_factory/tests.rs
```diff
@@ -230,10 +238,55 @@ mod bundle_factory_tests {
...
+        bundle_factory.try_push(seq_action0.clone()).unwrap();
+
+        // push another sequence action that is <100 bytes total to force the current bundle to
+        // flush
+        let seq_action1 = SequenceAction {
+            rollup_id: RollupId::new([1; ROLLUP_ID_LEN]),
+            data: vec![1; 100 - ROLLUP_ID_LEN],
+            fee_asset_id: default_native_asset_id(),
+        };
+        bundle_factory.try_push(seq_action1).unwrap();
```
bharath-123, line 261: can we assert that the test results to an `Ok(())`. otherwise any other changes in code might result in a panic

crates/astria-composer/src/executor/bundle_factory/mod.rs
itamarreif: We decided to make this event a `trace!` a while back because it would be incredibly noisy otherwise.

crates/astria-composer/src/executor/bundle_factory/mod.rs
itamarreif: `SizedBundle.push()` requires a `SizedSequenceAction`s rather than a regular `SequenceAction` to ensure that the `SequenceAction` went through `SizedBundle.push_check()` before being pushed.

crates/astria-composer/src/executor/bundle_factory/mod.rs
itamarreif: If its full then its size is the capacity passed by config - this should be logged by the service on startup.

crates/astria-composer/src/executor/bundle_factory/mod.rs
itamarreif: The `BundleFactory` can have a full `finished` queue but still have space in its `curr_bundle`.

crates/astria-composer/src/executor/bundle_factory/mod.rs
itamarreif: The `BundleFactory` can have a full `finished` queue but still have space in its `curr_bundle`.

crates/astria-composer/src/executor/bundle_factory/mod.rs
itamarreif: The first `push_check` is on the original `curr_bundle`. If the seq action doesn't fit in it then curr is flushed into the `finished` queue and replaced with a fresh `SizedBundle`. The seq action is then pushed into the new `curr_bundle`. The push into a `SizedBundle` the checks in `push_check` have to be done first. This isn't different from what we had before - those checks happened at the start of the push call anyways.

crates/astria-composer/src/executor/bundle_factory/tests.rs
itamarreif: thats what `unwrap` does here.

crates/astria-composer/src/executor/mod.rs
```diff
@@ -250,8 +255,8 @@ impl Executor {
                     }
                 }
 
-                // receive new seq_action and bundle it
-                Some(seq_action) = self.serialized_rollup_transactions.recv() => {
+                // receive new seq_action and bundle it. will not pull from the channel if `bundle_factory` is full
+                Some(seq_action) = self.serialized_rollup_transactions.recv(), if !bundle_factory.full() => {
                     let rollup_id = seq_action.rollup_id;
                     if let Err(e) = bundle_factory.try_push(seq_action) {
                             warn!(
```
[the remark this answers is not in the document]
itamarreif: this condition will run before `Some(seq_action)` is pulled from the channel. this means that the channel won't be pulled on while `bundle_factory` is full. this will cause `try_send` calls in the collectors to timeout, and that's where transactions will be dropped.

crates/astria-composer/src/executor/bundle_factory/tests.rs
bharath-123: yeah but i think `unwrap` panics right? an assertion feels much more nicer if the test fails again?

crates/astria-composer/src/executor/mod.rs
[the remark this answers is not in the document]
bharath-123: ah got it. that makes sense.

crates/astria-composer/src/executor/bundle_factory/tests.rs
```diff
@@ -250,10 +303,60 @@ mod bundle_factory_tests {
...
+
+        // assert that the bundle factory has one bundle in the finished queue, that the factory is
+        // full and that err was returned
+        assert!(matches!(
+            full_err,
+            Err(BundleFactoryError::FinishedQueueFull {
+                curr_bundle_size: _,
+                seq_action: _
+            })
+        ));
```
Fraser999, line 346: I think this should be removed as we tested for this above, and it's just noise now in this test.

crates/astria-composer/src/executor/bundle_factory/mod.rs
```diff
@@ -73,9 +73,13 @@ impl SizedBundle {
         }
     }
 
-    /// Buffer `seq_action` into the bundle. If the bundle won't fit `seq_action`, flush `buffer`,
-    /// returning it, and start building up a new buffer using `seq_action`.
-    fn push(&mut self, seq_action: SequenceAction) -> Result<(), SizedBundleError> {
+    /// Buffer `seq_action` into the bundle. Requires calling `push_check` first on `seq_action` to
+    /// validate its size
```
Fraser999, line 77: This comment is out of date now.

crates/astria-composer/src/executor/bundle_factory/mod.rs
```diff
@@ -153,12 +176,31 @@ impl BundleFactory {
...
+                        seq_action,
+                    })
+                } else {
+                    self.finished.push_back(self.curr_bundle.flush());
+                    // if the finished queue is full after the flush mark it as such
+                    if self.finished.len() == self.finished_queue_capacity {
+                        self.full = true;
+                    }
+
+                    self.curr_bundle.try_push(seq_action).expect(
```
Fraser999, line 192: I still think `expect` here is a little fragile as per [my previous comment](https://github.com/astriaorg/astria/pull/937#discussion_r1562507169).  We could log an `error!` here in case `try_push` returns an error, and just return `BundleFactoryError::SequenceActionTooLarge`

bharath-123 (approved):
LGTM. PR needs a rebase

crates/astria-composer/src/executor/bundle_factory/mod.rs
```diff
@@ -115,6 +118,14 @@ impl SizedBundle {
 pub(super) enum BundleFactoryError {
     #[error("sequence action is larger than the max bundle size. seq_action size: {size}")]
     SequenceActionTooLarge { size: usize, max_size: usize },
+    #[error(
+        "finished bundle queue is full and the sequence action does not fit in the current bundle"
```
SuperFluffy, line 122: The error message should contain all the mentioned quantities, like `curr_bundle_size`, `finished_queue_capacity` etc

charts/evm-rollup/values.yaml
```diff
@@ -51,7 +51,16 @@ config:
...
     # - "SoftAndFirm" -> blocks are pulled from both the sequencer and DA
     executionCommitLevel: 'SoftAndFirm'
+    # Max bytes to encode into a single sequencer `SignedTransaction`, not including signature,
+    # public key, nonce. This is the sum of the sizes of all the `SequenceAction`s. Should be
+    # set below the sequencer's max block size to allow space for encoding, signature, public
+    # key and nonce bytes
     maxBytesPerBundle: 200000
+    # Max amount of bundles that can be in the submission queue.
+    # ASTRIA_COMPOSER_BUNDLE_QUEUE_CAPACITY * ASTRIA_COMPOSER_MAX_BYTES_PER_BUNDLE (e.g.
+    # 40 * 200KB = 8GB) is the limit on how much memory the finished bundle queue can consume.
```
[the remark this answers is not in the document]
itamarreif: oh yeah this should be 400, my bad

crates/astria-composer/src/executor/bundle_factory/mod.rs
itamarreif: the sequence action is already checked in the first `try_push`.

if it's a `SequenceActionTooLarge` error (which happens regardless of the state of `curr_bundle`) it won't reach this branch. getting back a `SequenceActionTooLarge` on the second call can't happen since it was already checked.

since its being pushed to a newly created `SizedBundle` and the `SizedBundle`'s max size is the same as an individual `SequenceAction`'s max size, `NotEnoughSpace` can't happen either.

crates/astria-composer/src/executor/bundle_factory/mod.rs
itamarreif: thanks for catching this

crates/astria-composer/local.env.example
```diff
@@ -47,6 +47,13 @@ ASTRIA_COMPOSER_MAX_SUBMIT_INTERVAL_MS=2000
 # key and nonce bytes
 ASTRIA_COMPOSER_MAX_BYTES_PER_BUNDLE=200000
 
+# Max amount of finished bundles that can be in the submission queue.
+# ASTRIA_COMPOSER_BUNDLE_QUEUE_CAPACITY * ASTRIA_COMPOSER_MAX_BYTES_PER_BUNDLE (e.g.
+# 40 * 200KB=8GB) is the limit on how much memory the finished bundle queue can consume.
```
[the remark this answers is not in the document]
itamarreif: changed to 400

crates/astria-composer/src/executor/bundle_factory/mod.rs
Fraser999: > if it's a `SequenceActionTooLarge` error (which happens regardless of the state of `curr_bundle`) it won't reach this branch.

That's true just now, but if someone in the future swaps the order of the checks to
```rust
if self.curr_size + seq_action_size > self.max_size {
    return Err(SizedBundleError::NotEnoughSpace(seq_action));
}
if seq_action_size > self.max_size {
    return Err(SizedBundleError::SequenceActionTooLarge(seq_action));
}
```
then we could end up panicking here.

I won't belabour the point though.  There's probably little chance that `SizedBundle::try_push` would actually get changed like that (since then we'd never actually hit the `SequenceActionTooLarge` path), and even if it did, we do at least have a test which would expose the panic.

itamarreif: If anything, were most likely to get rid of this bundle factory entirely in the near future in favor of using a db

charts/evm-rollup/values.yaml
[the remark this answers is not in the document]
itamarreif: LOL yes my bad again, was doing this rushed.

Fraser999 (approved):

WafflesVonMaple (approved):

PR #939, opened 2024-04-10
feat(proto): create sequencerblockapis v1alpha1

noot:
## Summary
creates a new buf package `astria.sequencerblock.v1alpha1`. this package contains block.proto, celestia.proto, merkle.proto and service.proto from my proto update PR (https://github.com/astriaorg/astria/pull/830) which are all related to the sequencer block API.

## Background
we want to split the block API and transaction API protos into different buf packages.

## Changes
- creates a new buf package `astria.sequencerblock.v1alpha1`
- contains block.proto, celestia.proto, merkle.proto and service.proto (updated for ABCI v0.38, however merkle.proto is unchanged from astria.sequencer.v1)

## Testing
n/a

## Related Issues

related to https://github.com/astriaorg/astria/issues/679 and will update https://github.com/astriaorg/astria/pull/830 after this is merged


joroshiba (commented):
We should add something to the release to publish, I can create the starter repo in buf for it to be pushed to.

joroshiba (approved):

proto/sequencerblockapis/astria/sequencerblock/v1alpha1/celestia.proto
```diff
@@ -0,0 +1,49 @@
...
+
+// The metadata of a sequencer block that is submitted to celestia.
+//
+// It is created by splitting a `astria.sequencer.v1alpha.SequencerBlock` into a
+// `CelestiaSequencerBlob` (which can be thought of as a header), and a sequence ofj
+// `CelestiaRollupBlob`s.
+//
+// The original sequencer block (and in turn CometBFT block) can be identified by the
+// block hash calculated from `header`.
+message CelestiaSequencerBlob {
```
SuperFluffy, line 34: I was thinking about renaming this to `HeaderBlob` or something like that to make it purpose clear. What do you think?

proto/sequencerblockapis/astria/sequencerblock/v1alpha1/block.proto
```diff
@@ -0,0 +1,136 @@
...
+import "google/protobuf/timestamp.proto";
+
+// `RollupTransactions` are a sequence of opaque bytes together with a 32 byte
+// identifier of that rollup.
+//
+// The binary encoding is understood as an implementation detail of the
+// services sending and receiving the transactions.
+message RollupTransactions {
+  // The 32 bytes identifying a rollup. Usually the sha256 hash of a plain rollup name.
+  bytes id = 1;
```
SuperFluffy, line 16: Can we name this `rollup_id` to make the other names? (like `rollup_ids_proof` in the block, and `rollup_id` in `CelestiaRollupBlob`)

crates/astria-core/src/generated/astria.sequencerblock.v1alpha1.rs
SuperFluffy, line 1: This requires changes to the proto-generator tool:

Can you amend the tool such the following code is generated for all new protos?

1. [`tonic_build::Builder::bytes`](https://docs.rs/tonic-build/latest/tonic_build/struct.Builder.html#method.bytes), probably with argument (please double check) `&[".astria.sequencerblock.v1alpha1"]`
2. [`pbjson_build::Builder::build`](https://docs.rs/pbjson-build/latest/pbjson_build/struct.Builder.html#method.build), also with `&[".astria.sequencerblock.valpha1"]`.

I have a long-standing task to do this but never got around to it for the code generated from our other specs.

proto/sequencerblockapis/astria/sequencerblock/v1alpha1/merkle.proto
SuperFluffy, line 1: I wonder if this should be in `sequencerblock` API or outside of it (similar to our custom `Uint128` type). No strong opinion either way.

proto/sequencerblockapis/astria/sequencerblock/v1alpha1/block.proto
```diff
@@ -0,0 +1,136 @@
...
+message SequencerBlockHeader {
+  // the cometbft chain ID of the sequencer chain
+  string chain_id = 1;
+  // the height of this sequencer block
+  uint64 height = 2;
+  // the timestamp of this sequencer block
+  google.protobuf.Timestamp time = 3;
+  // the data_hash of the sequencer block (merkle root of all transaction hashes)
+  bytes data_hash = 4;
+  // the cometbft proposer address of the sequencer block
```
SuperFluffy, line 61: Is this already affected our discussion re `bech32`? If yes, then we should probably change this to `string` because of the protobuf <> json mapping

proto/sequencerblockapis/astria/sequencerblock/v1alpha1/block.proto
```diff
@@ -0,0 +1,136 @@
...
+// corresponding to that within the respective rollup's data.
+//
+// A `Deposit` notifies a rollup that funds were locked to some account on the sequencer,
+// however it's up to the rollup what to do with that info.
+message Deposit {
+  // the address the funds were locked in on the sequencer.
+  // this is required as initializing an account as a bridge account
+  // is permissionless, so the rollup consensus needs to know and enshrine
+  // which accounts it accepts as valid bridge accounts.
+  bytes bridge_address = 1;
```
SuperFluffy, line 82: As above, should this be `string` because of bech32?

proto/sequencerblockapis/astria/sequencerblock/v1alpha1/merkle.proto
joroshiba: I think this could be its own package or partnered with the uint128 in a "primitives" package

noot: yeah, i'm fine with moving it to primitives, as it could eventually be used for tx proofs as well. i can make a follow up

proto/sequencerblockapis/astria/sequencerblock/v1alpha1/block.proto
noot: answered in slack - we can change this later as it's v1alpha1, and also other protos in sequencer would need to change as well, so would rather do it all at once in another PR

proto/sequencerblockapis/astria/sequencerblock/v1alpha1/celestia.proto
noot: maybe `CelestiaSequencerHeaderBlob`?

PR #940, opened 2024-04-10
fix(sequencer): move begin_block ibc stuff to end_block

noot:
don't review yet!! this is a hack fix.

## Summary
maybe fix running a full node - during proposal phase tx execution, we haven't called begin_block before executing the txs, so the state is different. this is not the proper fix, but might fix full nodes on dusk 4. the actual fix is to call the begin_block stuff inside prepare/process_proposal before tx execution.

## Changes
-  move begin_block ibc stuff to end_block

## Testing
dusk 4 full node ?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -738,6 +738,7 @@ dependencies = [
  "astria-merkle",
  "astria-telemetry",
  "async-trait",
+ "base64 0.21.7",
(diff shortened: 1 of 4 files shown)


noot:
closing b/c this needs to be built off the 0.10.1 release

PR #945, opened 2024-04-11
fix(sequencer)!: fix state setup to be consistent before transaction execution

noot:
## Summary
we had an issue with syncing blocks with a full node due to the state not being set up the same for tx execution in the proposal phase vs not in the proposal phase. previously, this did not cause issues as txs were not dependent on the state set up when beginning the block, but with the introduction of ibc actions, they now are. 

## Background
see above

## Changes
- refactor the sequencer app to have a fn `pre_execute_transactions` which must be called before transaction execution in any phase. this function is then called in `prepare/process_proposal`, which sets up the state to be the same as when a block is executed via only `finalize_block`.
- rename some functions for clarity (deliver_tx -> execute_transaction, execute_block_data -> execute_transactions_before_finalization)

## Testing
unit tests + we should run it on devnet

## Breaking Changelist
- state is now set up differently than before during prepare/process proposal

### DIFF
--- a/crates/astria-sequencer/src/app.rs
+++ b/crates/astria-sequencer/src/app.rs
@@ -146,43 +146,6 @@ pub(crate) struct App {
     app_hash: AppHash,
 }
 
-/// Relevant data of a block being executed.
-#[derive(Debug, Clone)]
-struct BlockData {
-    txs: Vec<bytes::Bytes>,
-    misbehavior: Vec<tendermint::abci::types::Misbehavior>,
-    height: tendermint::block::Height,
-    time: tendermint::Time,
-    next_validators_hash: Hash,
-    proposer_address: account::Id,
-}
-
-impl From<abci::request::FinalizeBlock> for BlockData {
-    fn from(finalize_block: abci::request::FinalizeBlock) -> Self {
-        Self {
-            txs: finalize_block.txs,
-            misbehavior: finalize_block.misbehavior,
-            height: finalize_block.height,
-            time: finalize_block.time,
-            next_validators_hash: finalize_block.next_validators_hash,
-            proposer_address: finalize_block.proposer_address,
-        }
-    }
-}
-
-impl From<abci::request::PrepareProposal> for BlockData {
-    fn from(prepare_proposal: abci::request::PrepareProposal) -> Self {
-        Self {
-            txs: prepare_proposal.txs,
-            misbehavior: prepare_proposal.misbehavior,
-            height: prepare_proposal.height,
-            time: prepare_proposal.time,
-            next_validators_hash: prepare_proposal.next_validators_hash,
-            proposer_address: prepare_proposal.proposer_address,
-        }
-    }
-}
-
 impl App {
     pub(crate) fn new(snapshot: Snapshot) -> Self {
         tracing::debug!("initializing App instance");
@@ -281,7 +244,7 @@ impl App {
         self.update_state_for_new_round(&storage);
 
         let txs = self
-            .pre_execute_block(prepare_proposal.into())
+            .pre_execute_transactions(prepare_proposal.into())
             .await
             .context("failed to prepare for executing block")?;
 
@@ -359,7 +322,7 @@ impl App {
         };
 
         let txs = self
-            .pre_execute_block(block_data)
+            .pre_execute_transactions(block_data)
             .await
             .context("failed to prepare for executing block")?;
 
@@ -408,7 +371,7 @@ impl App {
     ///
     /// Returns the transactions which were successfully decoded and executed
     /// in both their [`SignedTransaction`] and raw bytes form.
-    #[instrument(name = "App::execute_block_before_finalization", skip_all, fields(
+    #[instrument(name = "App::execute_transactions_before_finalization", skip_all, fields(
         tx_count = txs.len()
     ))]
     async fn execute_transactions_before_finalization(
@@ -505,7 +468,10 @@ impl App {
 
     // sets up the state for execution of the block's transactions.
     // set the current height and timestamp, and calls `begin_block` on all components.
-    async fn pre_execute_block(
+    //
+    // this *must* be called anytime before a block's txs are executed, whether it's
+    // during the proposal phase, or finalize_block phase.
+    async fn pre_execute_transactions(
         &mut self,
         block_data: BlockData,
     ) -> anyhow::Result<Vec<bytes::Bytes>> {
@@ -596,21 +562,15 @@ impl App {
             self.update_state_for_new_round(&storage);
         }
 
-        // this function returns the txs inside `finalize_block` back to us unmodified.
-        let txs = self
-            .pre_execute_block(finalize_block.into())
-            .await
-            .context("failed to execute block")?;
-
         ensure!(
-            txs.len() >= 2,
+            finalize_block.txs.len() >= 2,
             "block must contain at least two transactions: the rollup transactions commitment and
              rollup IDs commitment"
         );
 
         // cometbft expects a result for every tx in the block, so we need to return a
         // tx result for the commitments, even though they're not actually user txs.
-        let mut tx_results: Vec<ExecTxResult> = Vec::with_capacity(txs.len());
+        let mut tx_results: Vec<ExecTxResult> = Vec::with_capacity(finalize_block.txs.len());
         tx_results.push(ExecTxResult {
             ..Default::default()
         });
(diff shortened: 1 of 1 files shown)


crates/astria-sequencer/src/app.rs
```diff
@@ -426,24 +459,21 @@ impl App {
...
-            block::Header,
-        };
-
-        use crate::transaction::InvalidNonce;
-
+        block_data: BlockData,
+    ) -> anyhow::Result<Vec<bytes::Bytes>> {
         let data_hash = astria_core::sequencerblock::v1alpha1::block::merkle_tree_from_data(
-            finalize_block.txs.iter().map(std::convert::AsRef::as_ref),
+            block_data.txs.iter().map(std::convert::AsRef::as_ref),
```
Lilyjjo, line 476: If I'm reading things right, the generated `data_hash` could differ when called from `prepare_proposal()` than other places since it's using the unfiltered set of transactions vs filtered. I didn't see any of the begin_block() component implementations using it so I don't think it's going to cause problems as-is, but this could cause issues in the future.

noot: good point, this `data_hash` is only used in the `begin_block` struct below which is passed into the components, but none of them actually use the `data_hash` so this does nothing, i'll remove it!

crates/astria-sequencer/src/app.rs
```diff
@@ -785,11 +779,11 @@ impl App {
     /// successfully.
     ///
     /// Note that `begin_block` is now called *after* transaction execution.
```
Lilyjjo, line 781: I don't think this is true anymore 
```suggestion

```

crates/astria-sequencer/src/app.rs
```diff
@@ -537,38 +519,72 @@ impl App {
...
+        let chain_id: tendermint::chain::Id = self
+            .state
+            .get_chain_id()
+            .await
+            .context("failed to get chain ID from state")?
+            .try_into()
+            .context("invalid chain ID")?;
+
+        // set the current proposer
+        self.current_proposer = Some(finalize_block.proposer_address);
```
Lilyjjo, line 572: I think this is just used in `end_block()` which we call from this function. If there isn't a reason why don't want to change the `end_block()` signature, it might be cleaner to just pass it into that function and not set/unset it in this one.

crates/astria-sequencer/src/app.rs
```diff
@@ -315,10 +319,42 @@ impl App {
...
+            .pre_execute_transactions(block_data)
             .await
-            .context("transactions failed to decode and execute")?;
+            .context("failed to prepare for executing block")?;
+
+        // we don't care about the cometbft max_tx_bytes here, as cometbft would have
+        // rejected the proposal if it was too large.
+        // however, we should still validate the other constraints, namely
+        // the max sequenced data bytes.
+        let mut block_size_constraints = BlockSizeConstraints::new(usize::MAX)
```
SuperFluffy, line 342: Maybe add a new constructor `BlockSizeConstraints::unlimited_cometbft` or something like that?

crates/astria-sequencer/src/app.rs
```diff
@@ -315,10 +319,42 @@ impl App {
...
+            .execute_transactions_before_finalization(txs, &mut block_size_constraints)
+            .await
+            .context("failed to execute transactions")?;
+
+        // all txs in the proposal should be deserializable and executable
+        // if any txs were not deserializeable or executable, they would not have been
+        // returned by `execute_block_data`, thus the length of `txs_to_include`
+        // will be shorter than that of `txs`.
+        ensure!(
+            txs_to_include.len() == expected_txs_len,
```
SuperFluffy, line 355: Can we require that `execute_transactions_before_finalization` only return with success if all transactions are executable and deserializable instead of adding this extra check?

crates/astria-sequencer/src/app.rs
```diff
@@ -442,88 +493,19 @@ impl App {
...
-                );
-                self.execution_result.insert(tx_hash, Err(e));
-            }
-        }
-        tx_hash
-    }
-
-    #[instrument(name = "App::finalize_block", skip_all)]
-    pub(crate) async fn finalize_block(
+    // sets up the state for execution of the block's transactions.
```
SuperFluffy, line 500: Doc comment intended? I.e. `/// sets up` (triple instead of double slashes)

crates/astria-sequencer/src/app.rs
```diff
@@ -442,88 +493,19 @@ impl App {
...
+    // set the current height and timestamp, and calls `begin_block` on all components.
+    //
+    // this *must* be called anytime before a block's txs are executed, whether it's
+    // during the proposal phase, or finalize_block phase.
+    async fn pre_execute_transactions(
```
SuperFluffy, line 505: Add instrumentation?

crates/astria-sequencer/src/app.rs
```diff
@@ -537,38 +519,72 @@ impl App {
...
+
+    #[instrument(name = "App::finalize_block", skip_all)]
+    pub(crate) async fn finalize_block(
+        &mut self,
+        finalize_block: abci::request::FinalizeBlock,
+        storage: Storage,
+    ) -> anyhow::Result<abci::response::FinalizeBlock> {
+        let chain_id: tendermint::chain::Id = self
+            .state
+            .get_chain_id()
```
SuperFluffy, line 565: Should this method maybe just return a valid chain ID so that it need not be parsed again?

SuperFluffy (commented):
I left some comments here and there.

Truth be told, I don't feel qualified reviewing this PR in terms of business logic because I lack understanding of its internal flow.

Lilyjjo (approved):
Besides the small comments I left, this looks good to me!

crates/astria-sequencer/src/app.rs
noot: `execute_transactions_before_finalization` is also called in `prepare_proposal`, so it can't error if not all txs are executable. it's simplest to have this check here i think

crates/astria-sequencer/src/app.rs
```diff
@@ -442,133 +507,103 @@ impl App {
...
 
+        Ok(())
+    }
+
+    /// Executes the given block, but does not write it to disk.
+    ///
+    /// `commit` must be called after this to write the block to disk.
+    ///
+    /// This is called by cometbft after the block has already been
+    /// finalized by the network.
```
joroshiba, line 572: nit: duplication of finalize here kinda confusing, more technically correct to say "consensus reached" perhaps?

crates/astria-sequencer/src/app.rs
```diff
@@ -442,133 +507,103 @@ impl App {
             );
         }
 
+        self.execution_results = Some(execution_results);
         Ok((signed_txs, validated_txs))
```
joroshiba, line 511: nit: feels kinda overloaded here with the mutation of property on self and then also returning a tuple. I can't think of a great fix right now, but it feels a tad off.

crates/astria-sequencer/src/app.rs
```diff
@@ -577,37 +612,60 @@ impl App {
...
+                time,
+                next_validators_hash: finalize_block.next_validators_hash,
+                proposer_address,
+            };
+
+            self.pre_execute_transactions(block_data)
+                .await
+                .context("failed to execute block")?;
+
+            for tx in finalize_block.txs.iter().skip(2) {
```
joroshiba, line 630: nit: can we put a comment above about why the skip 2 is there?

crates/astria-sequencer/src/app.rs
```diff
@@ -2306,6 +2311,149 @@ mod test {
         assert_eq!(deposits[0], expected_deposit);
     }
 
+    // it's a test, so allow a lot of lines
+    #[tokio::test]
+    #[allow(clippy::too_many_lines)]
+    async fn app_execution_results_match_proposal_vs_after_proposal() {
```
joroshiba, line 2317: Should we have a test, perhaps using an snapshot to validate that when we only call finalize block on something (with actual transactions) that we reach the same result as what we generate by doing prepare/process/finalize?

joroshiba (approved):
Approved, noting I think we should add a test to confirm that going through block sync yields same results as going through a consensus round.

noot:
@joroshiba i added a unit test for it, at the bottom of `app.rs`.

crates/astria-sequencer/src/app.rs
noot: yeah it's more like this (returning the tuple) to prevent unnecessary allocations

crates/astria-sequencer/src/app.rs
noot: this test already tests that, it tests only finalize_block (full node), prepare_proposal, then process_proposal, then finalize_block (proposer node), and process_proposal then finalize_block (non proposer validator) and ensures the results are the same. and it uses both normal and bridge txs

Issue #884, opened 2024-03-31
Use genesis RPC to derive Celestia namespace from chain ID


SuperFluffy:
At the moment Conductor uses the `latest_commit` RPC (i.e. `"method": "commit"` with parameter `"height": null"`) to obtain Sequencer's CometBFT chain ID and from it derive the v0 Celestia namespace. It should instead be calling `"method": "genesis"` because it's simpler and does not rely on the particular current state of the chain.

Issue #184, opened 2023-07-19
Add timeouts to all RPC calls


SuperFluffy:
Right now neither conductor nor sequencer-relayer are timing out if their calls to celestia or cometbft/tendermint take too long. This can potentially deadlock the entire application.

These calls should all have a timeout and resubmit logic so that data does not get lost when servers or the network misbehave.

Issue #94, opened 2023-06-16
write integration tests against the `Conductor` type


SuperFluffy:

noot:
closing as the tracking issue is closed

SuperFluffy:
Reopened the issue because this is still relevant. It was decided to remove it from the tracking issue due to time constraints. It should have been addressed in the meantime but wasn't.

PR #946, opened 2024-04-12
fix(conductor): robust Celestia blob fetch, verify, convert

SuperFluffy:
## Summary
Fixes conductor so it works in the presence of duplicate sequencer blobs or rollup blobs fetched from Celestia.

## Background
Duplicate sequencer block information can end up on Celestia. If a single Sequencer block ends up as a duplicate at the same Celestia height (for example, because two Sequencer nodes both relayed to Celestia), then Conductor can potentially stop functioning correctly because of code present as of commit [`cd38d1d5458fd561696590417b647f1844d328f2`](https://github.com/astriaorg/astria/blob/cd38d1d5458fd561696590417b647f1844d328f2/crates/astria-conductor/src/celestia/mod.rs#L586-L589):
```rust
    ensure!(
        rollup_blobs.len() <= 1,
        "received more than one celestia rollup blob for the given namespace and height"
    );
```
This causes Conductor to drop reconstruction of a Sequencer block information from blobs fetched from Celestia, even though it would be perfectly acceptable to go on.

## Changes
- Split the celestia `Reader` object into a crate-public exposed Reader and an internal `RunningReader`. `Reader::run_until_stopped` itself initializes the `RunningReader` with information it has to await at runtime before delegating to it.
- Determine the Sequencer chain ID (required to get its namespace) form the `/genesis` endpoint (instead of `/latest_block`).
- Remove the stream of new blocks reconstructed from Celestia blobs
- Replace the stream with a `JoinMap` (one task per Celestia height)
- Each task performs the following steps
    1. fetch Celestia blobs for a given height, sequencer namespace, rollup namespace
    2. decode the blobs according to their namespace (sequencer header or rollup item)
    3. verify all sequencer headers blobs against sequencer
    4. reconstruct rollup information by matching the verified sequencer blobs to the rollup item blobs

## Testing
Blackbox tests are implemented for the following cases (more will be implemented in a followup):

1. conductor fetches a block with sequencer height 2 from celestia height 1 and executes it
2. conductor fetches blocks with sequencer height 2 from celestia height 1, sequencer height 3 from celestia height 2, and executes both in succession
3. the remote rollup starts at block number 5. Conductor fetches sequencer heights 5 and 6 from celestia heights 1 and 2, but only forwards sequencer height 6.
4. the remote rollup has celestia height 4 in its genesis. Conductor starts fetching at that block number.

## Related Issues
closes https://github.com/astriaorg/astria/issues/94
closes https://github.com/astriaorg/astria/issues/184 (together with work done in https://github.com/astriaorg/astria/pull/769)
closes https://github.com/astriaorg/astria/issues/884
closes https://github.com/astriaorg/astria/issues/947
closes https://github.com/astriaorg/astria/issues/948

### DIFF
--- a/crates/astria-conductor/src/celestia/block_verifier.rs
+++ b/crates/astria-conductor/src/celestia/block_verifier.rs
@@ -1,119 +1,74 @@
 use std::collections::HashMap;
 
-use astria_eyre::eyre::{
-    self,
-    bail,
-    ensure,
-    WrapErr as _,
-};
-use celestia_client::CelestiaSequencerBlob;
 use ed25519_consensus::{
     Signature,
     VerificationKey,
 };
 use prost::Message;
 use sequencer_client::{
-    tendermint,
+    tendermint::{
+        self,
+        block::Height,
+    },
     tendermint_rpc,
-    Client as _,
-    HttpClient,
 };
-use tracing::instrument;
 
-/// `BlockVerifier` is verifying blocks received from celestia.
-#[derive(Clone)]
-pub(super) struct BlockVerifier {
-    client: HttpClient,
-}
+#[derive(Debug, thiserror::Error)]
+pub(super) enum QuorumError {
+    #[error("commit height mismatch; expected `{expected_height}`, got `{actual_height}`")]
+    CommitHeightMismatch {
+        expected_height: Height,
+        actual_height: Height,
+    },
 
-impl BlockVerifier {
-    pub(super) fn new(client: HttpClient) -> Self {
-        Self {
-            client,
-        }
-    }
+    #[error(
+        "commit voting power is greater than total voting power: {commit_voting_power} > \
+         {total_voting_power}"
+    )]
+    CommitVotingPowerExceedsTotal {
+        commit_voting_power: u64,
+        total_voting_power: u64,
+    },
 
-    #[instrument(skip_all, fields(
-        height.in_blob = %blob.height(),
-        block_hash.in_blob = %telemetry::display::base64(&blob.block_hash()),
-    ))]
-    pub(super) async fn verify_blob(&self, blob: &CelestiaSequencerBlob) -> eyre::Result<()> {
-        Verify::at_height(self.client.clone(), blob.height())
-            .await
-            .wrap_err("failed getting the required objects to verify blob")?
-            .verify(blob)
-    }
-}
-struct Verify {
-    block_hash: tendermint::Hash,
-    commit_header: tendermint::block::signed_header::SignedHeader,
-}
+    #[error("commit contained an empty signature field for validator `{validator}`")]
+    EmptySignature { validator: tendermint::account::Id },
 
-impl Verify {
-    async fn at_height(
-        client: HttpClient,
-        height: tendermint::block::Height,
-    ) -> eyre::Result<Self> {
-        use futures::TryFutureExt as _;
-        ensure!(
-            height != tendermint::block::Height::from(0u32),
-            "cannot validate sequencer blocks at height 0",
-        );
-        // the validators at height h-1 vote for the block at height h.
-        let prev_height: tendermint::block::Height = (height.value() - 1).try_into().wrap_err(
-            "failed converting decremented height tendermint value to tendermint height type",
-        )?;
-        let (block, commit, validator_set) = tokio::try_join!(
-            client
-                .block(height)
-                .map_err(|e| eyre::Report::new(e).wrap_err("failed fetching sequencer block")),
-            client
-                .commit(height)
-                .map_err(|e| eyre::Report::new(e).wrap_err("failed fetching commit")),
-            client
-                .validators(prev_height, tendermint_rpc::Paging::Default)
-                .map_err(|e| eyre::Report::new(e).wrap_err("failed fetching validator set")),
-        )?;
-        let block_hash = block.block_id.hash;
-        let commit_header = commit.signed_header;
-        Self::new(block_hash, commit_header, &validator_set)
-    }
+    #[error(
+        "commit voting power is less than 2/3 of total voting power: {commit_voting_power} <= 2/3 \
+         {total_voting_power}"
+    )]
+    NoQuorum {
+        commit_voting_power: u64,
+        total_voting_power: u64,
+    },
 
-    fn new(
-        block_hash: tendermint::Hash,
-        commit_header: tendermint::block::signed_header::SignedHeader,
-        validator_set: &tendermint_rpc::endpoint::validators::Response,
-    ) -> eyre::Result<Self> {
-        ensure!(
-            block_hash == commit_header.header.hash(),
-            "block hash stored in commit header does not match block hash of sequencer block",
-        );
-        ensure_commit_has_quorum(
-            &commit_header.commit,
-            validator_set,
-            &commit_header.header.chain_id,
-        )
-        .wrap_err("unable to verify that commit had quorum")?;
-        Ok(Self {
-            block_hash,
-            commit_header,
-        })
-    }
+    #[error("validator `{validator}` was present in commit but not in validator set")]
+    NoSuchValidator { validator: tendermint::account::Id },
 
-    fn verify(&self, blob: &CelestiaSequencerBlob) -> eyre::Result<()> {
-        ensure!(
-            &self.commit_header.header.chain_id == blob.cometbft_chain_id(),
-            "expected cometbft chain ID `{}`, got {}",
-            self.commit_header.header.chain_id,
-            blob.cometbft_chain_id(),
-        );
-        ensure!(
-            blob.block_hash() == self.block_hash.as_bytes(),
-            "block hash in blob does not match block hash of sequencer block",
-        );
+    #[error(
+        "failed to recreate signature for validator from validator set to verify vote signature"
+    )]
+    Signature(#[source] ed25519_consensus::Error),
 
-        Ok(())
-    }
+    #[error("total voting power overflowed u64")]
+    TotalVotingPowerOverflowed,
+
+    #[error(
+        "validator `{expected}` in commit does not address `{actual}` derived from signature in \
+         validator set"
+    )]
+    ValidatorAddressMismatch {
+        expected: tendermint::account::Id,
+        actual: tendermint::account::Id,
+    },
+
+    #[error(
+        "failed to recreate public key for validator from validator set to verify vote signature"
+    )]
+    VerificationKey(#[source] ed25519_consensus::Error),
+
+    #[error("failed to verify vote signature")]
+    VerifyVoteSignature(#[source] ed25519_consensus::Error),
 }
 
 /// This function ensures that the given Commit has quorum, ie that the Commit contains >2/3 voting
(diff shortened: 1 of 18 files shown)


crates/astria-conductor/src/celestia/block_verifier.rs
SuperFluffy, line 1: This code only retains the `ensure_commit_has_quorum` logic. Other changes:

1. Typed out all error conditions
2. Replaced plain arithmetic by saturating arithmetic.

crates/astria-conductor/src/celestia/mod.rs
SuperFluffy, line 1: This file should be reviewed "from scratch" and not based on the diff.

crates/astria-conductor/tests/blackbox/helpers/mod.rs
SuperFluffy, line 1: Cointains new boiler plate to mount mocks and construct the right objects to be returned by these mocks.

crates/astria-core/src/sequencer/v1/test_utils.rs
```diff
@@ -27,16 +27,33 @@ pub fn make_cometbft_block() -> tendermint::Block {
...
 ///
 /// If the proposer address is not set it will be generated from the signing key.
 #[derive(Default)]
 pub struct ConfigureCometBftBlock {
+    pub chain_id: Option<String>,
     pub height: u32,
     pub proposer_address: Option<tendermint::account::Id>,
     pub signing_key: Option<ed25519_consensus::SigningKey>,
     pub rollup_transactions: Vec<(RollupId, Vec<u8>)>,
+    pub unix_timestamp: UnixTimeStamp,
```
SuperFluffy, line 56: This replaces the call to `Time::now()` - having non-deterministic bits in tests is bad because that changes the block hashes.

crates/astria-conductor/src/celestia/mod.rs
SuperFluffy, line 1: This file should probably be read from-scratch due to the large amount of changes.

noot:
does this close #94 also?

SuperFluffy:
> does this close #94 also?

Yes, indeed it does. Needs more test coverage but the PR can be closed

crates/astria-conductor/src/celestia/mod.rs
```diff
@@ -112,13 +116,19 @@ impl GetSequencerHeight for ReconstructedBlock {
     }
 }
 
+pub(super) struct ReconstructedBlocks {
+    celestia_height: u64,
+    sequencer_namespace: Namespace,
+    rollup_namespace: Namespace,
```
noot, line 122: shouldn't these always be known since they're set on startup?

crates/astria-conductor/src/celestia/mod.rs
```diff
@@ -130,138 +140,173 @@ pub(crate) struct Reader {
...
+    /// Tasks reconstructing Sequencer block information from Celestia blobs.
+    tasks: JoinMap<u64, eyre::Result<ReconstructedBlocks>>,
+
+    latest_heights: LatestHeightStream,
+
+    enqueued_block: Fuse<BoxFuture<'static, Result<u64, SendError<ReconstructedBlock>>>>,
+
+    celestia_head_height: Option<u64>,
+    celestia_next_height: u64,
+    celestia_reference_height: u64,
```
noot, line 198: can you add comments for these?

crates/astria-conductor/src/celestia/mod.rs
```diff
@@ -130,138 +140,173 @@ pub(crate) struct Reader {
...
+    celestia_client: CelestiaClient,
+
+    /// The channel used to send messages to the executor task.
+    executor: executor::Handle<StateIsInit>,
+
+    /// Token to listen for Conductor being shut down.
+    shutdown: CancellationToken,
+
+    /// Tasks reconstructing Sequencer block information from Celestia blobs.
+    tasks: JoinMap<u64, eyre::Result<ReconstructedBlocks>>,
```
noot, line 190:
```suggestion
    reconstruction_tasks: JoinMap<u64, eyre::Result<ReconstructedBlocks>>,
```

tasks is a bit vague :p

crates/astria-conductor/src/celestia/verify.rs
```diff
@@ -0,0 +1,342 @@
...
+                    "verification of sequencer blob failed; dropping it"
+                );
+            }
+        }
+    }
+
+    VerifiedBlobs {
+        celestia_height,
+        header_blobs: verified_header_blobs,
+        rollup_blobs,
```
noot, line 114: these look like theyre not getting verified? they appear to be verified in the caller in mod.rs in the call to `reconstruct_blocks_from_verified_blobs` so maybe rename this to `verify_headers_and_filter_decoded_blobs`

crates/astria-conductor/src/celestia/reconstruct.rs
```diff
@@ -0,0 +1,201 @@
...
+            .reverse_index
+            .get(&rollup.sequencer_block_hash())?
+            .iter()
+            .position(|key| {
+                let header = self
+                    .map
+                    .get(key)
+                    .expect("keys stored in the reverse index must exist in the map");
+                verify_rollup_blob_against_sequencer_blob(rollup, header)
+            })?;
```
noot, line 166: this returns the index of the first matching key, right? would there ever be a case where there are multiple keys in the `reverse_index.get(&rollup.sequencer_block_hash())` `Vec<Key>` that match the rollup blob?

crates/astria-conductor/src/celestia/reconstruct.rs
```diff
@@ -0,0 +1,201 @@
...
+        }
+    }
+}
+
+struct BlockHashAndHeightToHeader {
+    // The sequencer header blobs grouped by (block_hash, sequencer_height)
+    map: HashMap<Key, CelestiaSequencerBlob>,
+
+    // A reverse index of block_hash -> [(block_hash, sequencer_height)]
+    reverse_index: HashMap<[u8; 32], Vec<Key>>,
```
noot, line 111: why is the value a vec? wouldn't a specific block hash always be for a specific sequencer height?

crates/astria-conductor/src/celestia/reconstruct.rs
```diff
@@ -0,0 +1,201 @@
...
+        }
+    }
+
+    let remaining_headers = headers.map;
+
+    for header_blob in remaining_headers.into_values() {
+        if header_blob.contains_rollup_id(rollup_id) {
+            warn!(
+                block_hash = %base64(&header_blob.block_hash()),
+                "sequencer header blob contains the target rollup ID, but no matching rollup blob was found; the blob",
```
noot, line 77: is this missing something at the end?

noot, line 77: also, if this case happens, we should try to fetch the corresponding rollup blob (as it should eventually be posted), is there already logic for that?

crates/astria-conductor/src/celestia/reconstruct.rs
```diff
@@ -0,0 +1,201 @@
...
+};
+
+use super::{
+    verify::VerifiedBlobs,
+    ReconstructedBlock,
+};
+
+/// Reconstructs block information from verified blocks.
+///
+/// The reconstructed blocks contain a block hash, tendermint header (both extracted from
```
noot, line 22: it contains a `SequencerBlockHeader` now, no more tendermint :D

crates/astria-conductor/src/celestia/reconstruct.rs
```diff
@@ -0,0 +1,201 @@
...
+/// sequencer header blobs), and transactions (from rollup blobs, if present).
+///
+/// This process works in the following way:
+/// 1. Header blobs are placed in a map using a key `(block_hash, height)`.
+///    - header blobs with the same `(block_hash, height)` are dropped.
```
noot, line 27: height is sequencer height, not celestia height right? as per the comment on `reverse_index`

crates/astria-conductor/src/celestia/verify.rs
noot: or update the function comment to say it just verifies header blobs

crates/astria-conductor/src/celestia/reconstruct.rs
SuperFluffy: Yeah, that's what `fetch` does.

crates/astria-conductor/src/celestia/reconstruct.rs
```diff
@@ -0,0 +1,201 @@
...
+///    - A Sequencer block is reconstructed from the the first such match.
+///    - All all other sequencer header blobs with the same block hash but different height are
+///      dropped (it is assumed that sequencer blobs of different height cannot have the same block
+///      hash).
+///    - A rollup blob that has no matching sequencer header blob is dropped.
+/// 3. The remaining Sequencer blobs (i.e. those that had no matching rollup blob) are checked for
+///    `rollup_id`:
+///    - if they contained `rollup_id` they are dropped (as they should have had a matching blob but
+///      none was found).
+///    - if they did not contain `rollup_id` a Sequencer block is reconstructed.
```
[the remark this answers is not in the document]
SuperFluffy: yeah

crates/astria-conductor/src/celestia/reconstruct.rs
noot: sorry, i meant the rollup blob might not have already been posted, but it might be posted at a later celestia block (based on the variance). is that handled here? or are we assuming the rollup blob ends up in the same block as the sequencer blob always (based on the relayer impl)?

crates/astria-conductor/src/celestia/reconstruct.rs
SuperFluffy: The idea was that there could be multiple `(block_hash, sequencer_height)` pairs with the same `block_hash` but different `sequencer_height`s.

crates/astria-conductor/src/celestia/reconstruct.rs
SuperFluffy: I commented above: this code looks like its defending against something that should be caught during the `verify` phase.

noot (approved):
logic looks good! could you add tests also for cases where multiple rollup blobs for the same rollup are posted for the same sequencer blob (the case that the pr fixes?) however the blobs would have to differ in data (otherwise their DA commitment would be the same) so one would be invalid and probably filtered out during verification.

crates/astria-conductor/src/celestia/reconstruct.rs
SuperFluffy: Right - no. We are requiring that the blobs are "atomic" in that a single sequencer block must be available in a single Celestia height.

We can probable relax this requirement and build a map of `block_hash -> rollup_blob`, but that will require a rework.

The variance refers to two subsequent sequencer heights being available within a certain window of Celestia heights.

crates/astria-conductor/src/celestia/mod.rs
SuperFluffy: Yes, they are - I used them for reporting, but that seems redundant now (because the namespaces are always part of the `FetchConvertAndVerify::execute` instrumented future. Removed.

crates/astria-conductor/src/celestia/mod.rs
```diff
@@ -299,398 +333,228 @@ impl Reader {
...
-    }
 
-    fn variance(&self) -> u32 {
-        self.variance
+    fn cache_reconstructed_blocks(&mut self, reconstructed: ReconstructedBlocks) {
+        for block in reconstructed.blocks {
+            let block_hash = block.block_hash;
+            let celestia_height = block.celestia_height;
+            let sequencer_height = block.sequencer_height().value();
+            if let Err(e) = self.block_cache.insert(block) {
```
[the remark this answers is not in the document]
SuperFluffy: yes, indeed

crates/astria-conductor/src/celestia/verify.rs
SuperFluffy: I renamed it to just `verify_header_blobs` and adjusted the doc comment on the function.

crates/astria-conductor/src/celestia/reconstruct.rs
SuperFluffy: The verify phase is now constructing a map `block_hash -> CelestiaSequencerBlob` so that I could simplify this code.

crates/astria-conductor/src/celestia/reconstruct.rs
SuperFluffy: The verify phase is now constructing a map `block_hash -> CelestiaSequencerBlob` so that I could simplify this code.

crates/astria-conductor/src/celestia/reconstruct.rs
SuperFluffy: Reworded. Talking about "Sequencer headers" now.

crates/astria-conductor/src/celestia/reconstruct.rs
SuperFluffy: removed any mention of heights from this reconstruction logic (no longer relevant because it was absorbed by `verify`).

crates/astria-conductor/src/celestia/reconstruct.rs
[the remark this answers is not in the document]
SuperFluffy: added "empty block execution" as a term to that doc comment

SuperFluffy:
> could you add tests also for cases where multiple rollup blobs for the same rollup are posted for the same sequencer blob (the case that the pr fixes?) however the blobs would have to differ in data (otherwise their DA commitment would be the same) so one would be invalid and probably filtered out during verification.

Agree we should have that. This PR is already getting too big. I have opened https://github.com/astriaorg/astria/issues/976 and will follow-up there.

PR #950, opened 2024-04-13
feat(charts): hermes chart

joroshiba:
## Summary
Adds a hermes relayer chart which can be utilized to connect two IBC compatible chains for relaying.

Note: I have not been able to get the local connection working yet but have confirmed against celestia mocha.


WafflesVonMaple (approved):
madman

PR #953, opened 2024-04-15
refactor(composer): Use config-like Builders instead of `new` Constructors

bharath-123:
## Summary
Add config-like Builders for Executor, GethCollector and GrpcServer.

## Background
Currently, Executor, GethCollector and GrpcServer components are initialized by a `new()` constructor. As we keep adding more features, the variables passed to the constructor can increase which can harm readability. 

This PR aims to increase readability by maintaining seperate Builders for each component.

## Changes
- Create a `Builder` for `Executor` crate
- Create a `GethCollectorBuilder` for `GethCollector` struct
- Create a `GrpcServerBuilder` for `GrpcServer`.
- Remove `new()` method from Executor, GethCollector and GrpcServer

## Testing
Since this is a refactor and not a functional change, I ran the unit tests and blackbox successfully which should be fine.


crates/astria-composer/src/executor/builder.rs
```diff
@@ -0,0 +1,75 @@
...
+use tokio_util::sync::CancellationToken;
+
+use crate::{
+    executor,
+    executor::Status,
+};
+
+pub(crate) struct Builder {
+    pub(crate) sequencer_url: String,
+    pub(crate) private_key: SecretString,
```
bharath-123, line 30: We were initially passing them as references but I chose to avoid references here so that we don't have to deal with lifetime related stuff. Also String and SecretString should be fairly lightweight to clone.

crates/astria-composer/src/collectors/geth.rs
```diff
@@ -64,17 +64,17 @@ const WSS_UNSUBSCRIBE_TIMEOUT: Duration = Duration::from_secs(2);
 pub(crate) struct Geth {
     // Chain ID to identify in the astria sequencer block which rollup a serialized sequencer
     // action belongs to. Created from `chain_name`.
-    rollup_id: RollupId,
+    pub(super) rollup_id: RollupId,
```
bharath-123, line 67: We had to increase the visibility for the GethCollectorBuilder.

crates/astria-composer/src/grpc.rs
```diff
@@ -19,39 +19,19 @@ use tokio::{
...
 
 /// Listens for incoming gRPC requests and sends the Rollup transactions to the
 /// Executor. The Executor then sends the transactions to the Astria Shared Sequencer.
 ///
 /// It implements the `GrpcCollectorServiceServer` rpc service and also the tonic health service
 pub(crate) struct GrpcServer {
-    listener: TcpListener,
-    grpc_collector: collectors::Grpc,
-    shutdown_token: CancellationToken,
+    pub(super) listener: TcpListener,
```
SuperFluffy, line 29: Make this private again.

crates/astria-composer/src/grpc_server_builder.rs
```diff
@@ -0,0 +1,41 @@
...
+use tokio::net::TcpListener;
+use tokio_util::sync::CancellationToken;
+
+use crate::{
+    collectors,
+    executor,
+    grpc::GrpcServer,
+};
+
+pub(crate) struct Builder {
```
SuperFluffy, line 16: Move this builder into `grpc.rs` or `mkdir grpc && git mv grpc.rs grpc/mod.rs && grpc_server_builder.rs grpc/builder.rs`

crates/astria-composer/src/grpc_server_builder.rs
```diff
@@ -0,0 +1,41 @@
...
+    pub(crate) async fn build(self) -> eyre::Result<GrpcServer> {
+        let Self {
+            grpc_addr,
+            executor,
+            shutdown_token,
+        } = self;
+
+        let listener = TcpListener::bind(grpc_addr)
+            .await
+            .wrap_err("failed to bind grpc listener")?;
```
SuperFluffy, line 32:
```suggestion
            .wrap_err("failed to bind socket address")?;
```

crates/astria-composer/src/collectors/geth.rs
SuperFluffy: Undo this. The point of making these builders is to keep the implementation details of these objects private.

SuperFluffy (changes requested):
This has the right direction, but defining new `geth_collector_builder`/`grpc_server_builder` modules is wonky. Just move the types they create into the geth/grpc modules (or nest them under those modules). You can then refer to these builders as `grpc::Builder` and `geth::Builder`

crates/astria-composer/src/collectors/geth.rs
bharath-123: In the `collectors` module, we can have geth and grpc as seperate modules.

crates/astria-composer/src/executor/builder.rs
```diff
@@ -0,0 +1,74 @@
...
+
+use astria_core::sequencer::v1::{
+    transaction::action::SequenceAction,
+    Address,
+};
+use astria_eyre::{
+    eyre,
+    eyre::{
+        eyre,
+        Context,
```
SuperFluffy, line 11: `Context` is a compatibility trait used for migrating anyhow to eyre and will soon be feature-gated:
```suggestion
        WrapErr as ),
```

crates/astria-composer/src/executor/builder.rs
```diff
@@ -0,0 +1,74 @@
...
+    },
+};
+use ed25519_consensus::SigningKey;
+use ethers::core::k256::elliptic_curve::zeroize::Zeroize;
```
SuperFluffy, line 15: We don't use `ethers` but `secrecy` for this.

crates/astria-composer/src/executor/builder.rs
```diff
@@ -0,0 +1,74 @@
...
+
+use crate::{
+    executor,
+    executor::Status,
+};
+
+pub(crate) struct Builder {
+    pub(crate) sequencer_url: String,
+    pub(crate) private_key: SecretString,
+    pub(crate) block_time: u64,
```
SuperFluffy, line 31: Either keep the `block_time_ms` naming from config or make the type a `Duration`

crates/astria-composer/src/executor/builder.rs
```diff
@@ -0,0 +1,74 @@
...
+        } = self;
+        let sequencer_client = sequencer_client::HttpClient::new(sequencer_url.as_str())
+            .wrap_err("failed constructing sequencer client")?;
+        let (status, _) = watch::channel(Status::new());
+        let mut private_key_bytes: [u8; 32] = hex::decode(private_key.expose_secret())
+            .wrap_err("failed to decode private key bytes from hex string")?
+            .try_into()
+            .map_err(|_| eyre!("invalid private key length; must be 32 bytes"))?;
+        let sequencer_key = SigningKey::from(private_key_bytes);
+        private_key_bytes.zeroize();
```
SuperFluffy, line 53: As noted above: use the secrecy::zeroize::Zeroise` trait, not the one from ethers.

crates/astria-composer/src/executor/mod.rs
```diff
@@ -34,11 +34,7 @@ use futures::{
 };
 use pin_project_lite::pin_project;
 use prost::Message as _;
-use secrecy::{
-    ExposeSecret as _,
-    SecretString,
-    Zeroize as _,
-};
+use secrecy::Zeroize as _;
```
SuperFluffy, line 37: Here the right trait is used

crates/astria-composer/src/collectors/geth.rs
```diff
@@ -83,7 +83,7 @@ pub(crate) struct Status {
 }
 
 impl Status {
-    fn new() -> Self {
+    pub(crate) fn new() -> Self {
```
SuperFluffy, line 86:
```suggestion
    fn new() -> Self {
```

SuperFluffy (approved):
Thanks for the cleanup, this does make it quite a bit clearer where individual parameters go.

Please address the comments before merging.

crates/astria-composer/src/executor/builder.rs
bharath-123: will keep it block_time_ms for now.

PR #958, opened 2024-04-16
chore(proto)!: fully split sequencerapis and remove

joroshiba:
## Summary
Finishes the split up of the former `sequencer` apis into separate modules & repositories. `primitives` becomes the core that any other module can rely on, while `protocolapis` is shapes specifically needed for submitting transactions or interacting with the state of the astria blockchain. `sequencerblock` is information needed only for execution of already sequenced data.

## Background
The ultimate goal is to decouple the write path and read path of executing, such that future changes to transactions shapes are not necessarily breaking changes for dependent rollups. 

## Changes
- Removed sequencer protos & astria-core::sequencer::v1
- Added types to primitive and created protocol packages.

## Testing
CI/CD

## Breaking Changelist
- This is a breaking change for everything dependent on the sequencer protos package.


.github/workflows/release.yml
```diff
@@ -27,16 +27,16 @@ jobs:
           draft: ${{ github.ref_name != 'main'}}
           input: "proto/composerapis"
           buf_token: ${{ secrets.BUF_TOKEN }}
-      - name: "Upload Sequencer API Protos"
+      - name: "Upload Primitive API Protos"
```
joroshiba, line 30: Note: I will need to manually push some of these crates for the first time in order to tests to pass. I will push once I have review that covers the protos

aajimal (approved):

Fraser999 (approved):

SuperFluffy (commented):
Just brainstorming here for the design of `astria-core`: it probably makes sense to have `primitive/v1`, `sequencerblocks/v1` and similar modules closely match the protobuf packages.

Yet items like `RollupId` and `Address` are not "real" in that they don't have a matching representation in our protobuf specs.

I am thus thinking if these should be placed under a `astria-core::common` module that is explicitly not tracking any of the protobuf packages?

Or alternatively, should we create the following protobuf messages under `astria.primitive` (or actually, `astria.protocol`)?

```protobuf
message RollupId {
    bytes inner = 1;
}

message Address {
    bytes inner = 1;
}
```

joroshiba:
@SuperFluffy 

We could certainly do this with RollupId, i think it makes sense and could provide optionality in the future like providing a string and having the bytes be generated.

I have plans for an address protobuf which would support bech32 encoding as well. It will be a follow up PR. 

I would prefer not to do these implementations on this large PR. Have purposefully broken out larger changes into follow ups.

The only proto message change outside of splitting packages on here is changing action number details.

tools/protobuf-compiler/src/main.rs
SuperFluffy, line 1: Can the `bytes` and `pbjson` generade code also be done for the other newly defined types or does that create too much churn in this PR? If yes, then we can leave it to a followup.

joroshiba: Done

noot (approved):
looks good!

PR #965, opened 2024-04-17
feat(sequencer): add initial set of metrics to sequencer

### LINKED ISSUE #942: sequencer: add initial metrics
Currently the sequencer does not emit metrics to enable visibility into what is happening internally. To prepare for mainnet, we should fix this.

Lilyjjo:
## Summary
Adds metrics to the sequencer's app and mempool services, mostly around transaction exclusion metrics. Will need to add more metrics around 0.38 but this is the first batch.

## Background
Many metrics are already tracked by the sequencer's cometbft component. I created a [page in notion](https://www.notion.so/astria-org/Sequencer-Metrics-a615cadbdf0942789ae2104966b42c12?pvs=4) detailing what is already tracked and the new things tracked. 

## Changes
- Add initial set of metrics to the sequencer's app and mempool services. 
### Sequencer

| Metric Type | Metric Name                                             | Description                                                                                                       |
| --------------- | ------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------- |
| Count       | prepare_proposal_excluded_transactions_decode_failure   | The number of transactions that have been excluded from blocks due to failing to decode                           |
| Count       | prepare_proposal_excluded_transactions_cometbft_space   | The number of transactions that have been excluded from blocks due to running out of space in the cometbft block  |
| Count       | prepare_proposal_excluded_transactions_sequecner_space  | The number of transactions that have been excluded from blocks due to running out of space in the sequencer block |
| Count       | prepare_proposal_excluded_transactions_failed_execution | The number of transactions that have been excluded from blocks due to failing to execute                          |
| Gauge       | prepare_proposal_excluded_transactions                  | The number of excluded transactions in a proposal being prepared                                                  |
| Histogram       | process_proposal_deposits                   | The number of deposit transactions in a proposal                                                  |
| Count       | process_proposal_skipped_proposal                       | The number of times our submitted prepared proposal was skipped in process_proposal                               |
| Histogram       | process_proposal_transactions                       | The number of transactions in a proposal
| TODO        | Metrics for 0.38x version of CometBFT                   | Issue here: #966                                                                                                      |

### Mempool

| Metric Type | Metric Name                                          | Description                                                                                                                                 |
| ----------- | ---------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------- |
| Count       | check_tx_removed_too_large                           | The number of transactions that have been removed from the mempool due to being too large                                                   |
| Count       | check_tx_removed_failed_stateless                    | The number of transactions that have been removed from the mempool due to failing the stateless check                                       |
| Count       | check_tx_removed_stale_nonce                         | The number of transactions that have been removed from the mempool due to having a stale nonce                                              |
| Count       | check_tx_removed_account_balance                     | The number of transactions that have been removed from the mempool due to having not enough account balance                                 |

## Testing

Enabled metrics by setting the local env var `ASTRIA_SEQUENCER_NO_METRICS` to `false` and by watching the target metrics populate on the port prometheus watches via http://127.0.0.1:9000/

## Issues 
closes #942


crates/astria-sequencer/src/metrics_init.rs
```diff
@@ -0,0 +1,173 @@
...
+    );
+
+    describe_histogram!(
+        PROCESS_PROPOSAL_TRANSACTIONS,
+        Unit::Count,
+        "The number of transactions in the process_proposal phase"
+    );
+}
+
+// We configure buckets for manually, in order to ensure Prometheus metrics are structured as a
```
Fraser999, line 106:
```suggestion
// We configure buckets manually, in order to ensure Prometheus metrics are structured as a
```

crates/astria-sequencer/src/metrics_init.rs
```diff
@@ -0,0 +1,173 @@
...
+// Histogram, rather than as a Summary. These values are loosely based on guesses
+// and may need to be updated over time.
+pub const HISTOGRAM_BUCKETS: &[f64; 5] = &[25.0, 50.0, 100.0, 200.0, 1000.0];
```
Fraser999, line 109: I'm not seeing where this value is actually being used.

Fraser999 (commented):
Almost all looks good to me.  The only thing stopping me from approving is that I'm not able to see how the histogram buckets are being specified.

crates/astria-sequencer/src/app.rs
```diff
@@ -291,6 +292,7 @@ impl App {
                 self.executed_proposal_hash = process_proposal.hash;
                 return Ok(());
             }
+            metrics::gauge!(metrics_init::PROCESS_PROPOSAL_SKIPPED_PROPOSAL).increment(1);
```
noot, line 295: this should be equal to the number of `prepare_proposal` calls, maybe making a metric for `prepare_proposal` would be clearer?

crates/astria-sequencer/src/app.rs
```diff
@@ -372,6 +380,10 @@ impl App {
 
             // don't include tx if it would make the cometBFT block too large
             if !block_size_constraints.cometbft_has_space(tx.len()) {
+                metrics::counter!(
+                    metrics_init::PREPARE_PROPOSAL_EXCLUDED_TRANSACTIONS_COMETBFT_SPACE
+                )
+                .increment(1);
```
noot, line 386: is this reset every block or is it a total over time?

crates/astria-sequencer/src/metrics_init.rs
```diff
@@ -0,0 +1,173 @@
...
+        "The number of transactions that have been removed from the mempool due to failing to \
+         decode from bytes into a protobuf"
+    );
+
+    describe_counter!(
+        CHECK_TX_REMOVED_FAILED_TO_DECODE_SIGNED_TRANSACTION,
+        Unit::Count,
+        "The number of transactions that have been removed from the mempool due to failing to \
+         decode from a protobuf into a valid signed transaction"
+    );
```
noot, line 76: what use cases do you see for these? people can submit arbitrary invalid bytes which would increase these

crates/astria-sequencer/src/app.rs
Lilyjjo: the counters accumulate over time in general

crates/astria-sequencer/src/metrics_init.rs
Lilyjjo: I added them for completeness with the other counters. The nonce and balance ones are fun to watch go up. I can remove them if you don't think they're needed

Lilyjjo: removed them!

crates/astria-sequencer/src/app.rs
Lilyjjo: I left them as counters because I didn't want to add variables to the function to track them per block production, the function is long already.

noot (approved):

PR #968, opened 2024-04-17
fix(sequencer): discard duplicate transactions in prepare_proposal

### LINKED ISSUE #895: sequencer: ensure duplicate txs passed via `prepare_proposal` are not executed more than once
the sequencer does not check that the tx list in `prepare_proposal` does not contain duplicates, this will cause consensus errors as one might execute and the next will fail during the proposal but will succeed for all other nodes as the duplicate will be removed. flagged by zellic. we should remove any duplicates during `prepare_proposal`

Lilyjjo:
## Summary
Sequnecer's `prepare_proposal()` should remove duplicate transactions from cometBFT's initial transaction set. CometBFT shouldn't pass the sequencer duplicate transactions, but if it ever does it'd cause consensus failures due to how we track execution results internally. 

## Background
We use a `HashMap<transaction_hash,execution_result>` internally to record the results of transactions. If we execute the same transaction twice, it'd succeed the first time and then fail the second time. The failure's exeuction would overwrite the success. This would cause discrepancies between the final execution results of `prepare_proposal()` vs `process_proposal()` and `finalize_block()`, as the initially passing execution result would exist in the other execution results.

## Changes
`prepare_proposal()` now filters out duplicate transactions.

## Testing
Added unit test.

## Metrics
Will add to this PR once initial sequencer's metrics PR lands. 

## Related Issues
closes #895

### DIFF
--- a/crates/astria-sequencer/src/app.rs
+++ b/crates/astria-sequencer/src/app.rs
@@ -1,5 +1,8 @@
 use std::{
-    collections::VecDeque,
+    collections::{
+        HashSet,
+        VecDeque,
+    },
     sync::Arc,
 };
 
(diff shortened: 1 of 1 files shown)


Lilyjjo:
Closing because other refactors made this one no longer needed (see: #895)

PR #973, opened 2024-04-18
feat!: require chain id in transactions

joroshiba:
## Summary
Adds the concept of `TransactionParams` to `UnsignedTransaction` as a body that can grow to include many different parameters which much pass for a transaction to be valid, and uses this structure to add a `chain_id` to ensure that transactions from one Astria network cannot be replayed on another.

## Background
There are likely to be additional parameters we can add onto transactions (such a expiration or things relating to auction bids), and we needed to add the chain id parameter.

## Changes
- Modify Transaction format
- Add chain id param and checks.

## Testing
CI/CD


aajimal (approved):

crates/astria-composer/local.env.example
```diff
@@ -27,6 +27,9 @@ ASTRIA_COMPOSER_API_LISTEN_ADDR="0.0.0.0:0"
 # Address of the RPC server for the sequencer chain
 ASTRIA_COMPOSER_SEQUENCER_URL="http://127.0.0.1:26657"
 
+# Chain ID of the sequencer chain which transactions are submitted to.
+ASTRIA_COMPOSER_SEQUENCER_CHAIN_ID="astria-dev-1"
```
SuperFluffy, line 31: IMO remove in favor of querying sequencer's `/genesis` endpoint to extract the cometbft chain ID from there. This is how conductor is doing it and I would like this to be consistent between services.

crates/astria-sequencer/src/transaction/mod.rs
```diff
@@ -157,6 +173,21 @@ pub(crate) async fn execute<S: StateWriteExt>(
         .await
 }
 
+#[derive(Debug)]
+pub(crate) struct InvalidChainId(pub(crate) String);
```
SuperFluffy, line 177: why `pub(crate)`?

SuperFluffy (changes requested):
The chain ID is already available when connecting to sequencer by querying it's ABCI `/genesis` endpoint, which is what conductor is doing.

I would like our services to be consistent in how they query the chain ID and use a single source of truth (the mentioned endpoint).

However, if you prefer to explicitly use a configuration parameter for this (for example to guard against accidentally connecting to the wrong sequencer network), then please still query `/genesis` and check that `configured_chain_id == remote_chain_id` to have composer *before* it receives or polls any transactions.

crates/astria-core/src/protocol/abci.rs
```diff
@@ -16,6 +16,7 @@ impl AbciErrorCode {
     pub const INVALID_NONCE: Self = Self(4);
     pub const TRANSACTION_TOO_LARGE: Self = Self(5);
     pub const INSUFFICIENT_FUNDS: Self = Self(6);
+    pub const INVALID_CHAIN_ID: Self = Self(7);
```
SuperFluffy, line 19: Please add an entry to `AbciErrorCode::info` below

crates/astria-core/src/protocol/abci.rs
SuperFluffy, line 1: This should also get a matching entry in the `AbciErrorCode` protobuf enum

SuperFluffy (approved):
Approving after syncing offline. We decided to a) require Sequencer chain IDs be set in config, and b) the configured Sequencer chain IDs be checked against the `/genesis` endpoint. However, this will and can be done in follow PRs (for both composer and conductor).

I am approving this PR. Before merge, please address my comments re the ABCI error codes:

1. add entries to the `Display` and `From` impls
2. add matching protobuf ABCI error codes (note: apparently these have been removed https://github.com/astriaorg/astria/pull/869 for some reason)

crates/astria-sequencer/src/transaction/mod.rs
joroshiba: Admittedly I just matched the InvalidNonce error publicity since these should be consumed similarly.

crates/astria-core/src/protocol/abci.rs
joroshiba: There is no longer a matching protobuf enum

joroshiba:
Follow up issue: https://github.com/astriaorg/astria/issues/986

crates/astria-sequencer/src/transaction/mod.rs
```diff
@@ -241,10 +272,20 @@ impl ActionHandler for UnsignedTransaction {
         state: &S,
         from: Address,
     ) -> anyhow::Result<()> {
+        // Transactions must match the chain id of the node.
+        let chain_id = state.get_chain_id().await?;
+        ensure!(
+            self.params.chain_id == chain_id.as_str(),
+            InvalidChainId(self.params.chain_id.clone())
+        );
+
```
noot, line 281: you can remove this - the mempool check is sufficient, no txs with invalid chain ids will make it here. the nonce check below is needed for strict equality before execution

crates/astria-sequencer/src/app.rs
```diff
@@ -2111,6 +2182,52 @@ mod test {
         );
     }
 
+    #[tokio::test]
+    async fn app_deliver_tx_invalid_chain_id() {
```
noot, line 2186: move/change this test to test the mempool check (after you delete the check_stateful check this test will fail)

crates/astria-sequencer/src/transaction/mod.rs
joroshiba: I think we need to protect against a validator running their node with a no op mempool. IE someone can modify code for their validator grab alternate transactions and propose?

Issue #837, opened 2024-03-19
conductor: get blocks between commitments on restart


joroshiba:
On restart of conductor, if soft and firm states do not match we cannot update the firm commitments correctly, because the block cache has been cleared. The `GetBlock` and `BatchGetBlock` APIs of execution API can be utilized for this. 

On startup we call: `GetGenesisInfo` and `GetCommitmentState`. If after getting commitment state `soft` and `firm` block do not match we need to use either of the above Block fetching methods to grab the rollup block hashes to allow the updating for firm commitment.

PR #980, opened 2024-04-19
feat(conductor)!: fetch blocks pending finalization 

SuperFluffy:
## Summary

Adds the ability for Conductor to sync its connected
rollup if firm commitments trail soft commitments.

## Background
Conductor uses the
`astria.execution.v1alpha2.BatchGetBlocks` API to
retrieve all blocks in the range `firm+1..=soft`
from the rollup node. For example, if the firm
number is 3 and soft number is 10, then all
soft-committed blocks 4 through 10 (inclusive) are
fetched.

Previously conductor would fail try to re-execute
the firm blocks as it didn't have their
corresponding rollup-hashes, causing the rollup
node to reject the execution.

## Changes

1. Conductor now matches rollup block numbers
rollup blocks (previously, rollup block hashes,
which it doesn't have on a fresh start).
2. Conductor uses these block numbers to update
the commitment state upon receiving a pending
firm block.
3. Reworked the channel tracking the rollup state
in the executor task: since Conductor must now be
able to in both directions between Sequencer
heights and Rollup numbers, updating the tracked
commitment state is now a fallible operation. This
required an abstraction on top of the
watch-channel containing the state.
4. changed how the next expected Sequencer height
is calculated from the Rollup number (for both
firm and soft commitments). It is now
sequencer_genesis_block_height + block_number + 1
(previously without + 1, which was confusing).

## Testing

1. Added 2 blackbox tests for soft-and-only mode,
one for testing the pending-block fetch.
2. Removed the executor tests as they are now covered
by the blackbox tests.

## Breaking Changelist

1. The calculation of the next expected Sequencer
height is now one more than previous. This is a
breaking change in the sense that rollup + conductor
need to started with the same view of
rollup number <> sequencer height in mind. A rollup
that was started with an older conductor will not work
with this conductor.

## Related Issues

closes https://github.com/astriaorg/astria/issues/837


crates/astria-conductor/src/executor/mod.rs
```diff
@@ -437,8 +513,60 @@ impl Executor {
...
+        fields(
+            firm_number = self.state.firm_number(),
+            soft_number = self.state.soft_number(),
+        )
+    )]
+    async fn populate_blocks_pending_finalization(
+        &mut self,
+        mut client: Client,
+    ) -> eyre::Result<()> {
+        if self.firm_blocks.is_none() || self.soft_blocks.is_none() {
```
joroshiba, line 527: should this be an error case? Why would we call this before setting soft and firm block?

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -437,8 +513,60 @@ impl Executor {
...
+        );
+        let blocks = client
+            .batch_get_blocks(range_of_missing_heights)
+            .await
+            .wrap_err("failed getting blocks for not yet finalized firm heights")?;
+
+        info!(
+            pending_blocks = %json(&reporting::ReportBlocks(&blocks)),
+            "received blocks pending finalization",
+        );
```
joroshiba, line 559: this could be a lot of blocks? We probably don't want to have an info log with all of them? IE if variance window is 100, 600 blocks?

crates/astria-conductor/src/executor/reporting.rs
```diff
@@ -0,0 +1,20 @@
...
+};
+
+pub(super) struct ReportBlocks<'a, T>(pub(super) &'a [T]);
+
+impl<'a> Serialize for ReportBlocks<'a, astria_core::execution::v1alpha2::Block> {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        let mut seq = serializer.serialize_seq(Some(self.0.len()))?;
```
joroshiba, line 14: again concerned about length here

joroshiba (changes requested):
only minors, but concerned about the logging

crates/astria-conductor/tests/blackbox/soft_and_firm.rs
```diff
@@ -0,0 +1,318 @@
...
+            update_commitment_state_soft.wait_until_satisfied(),
+        ),
+    )
+    .await
+    .expect(
+        "conductor should have executed the soft block and updated the soft commitment state \
+         within 1000ms",
+    );
+
+    // mount_get_filtered_sequencer_block!(
```
noot, line 265: remove

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -108,84 +142,94 @@ impl Handle<StateIsInit> {
...
-    consider_commitment_spread: bool,
+    /// Tracks the status of the execution chain.
+    state: StateSender,
 
     /// Tracks executed blocks as soft commitments.
     ///
     /// Required to mark firm blocks received from celestia as executed
     /// without re-executing on top of the rollup node on top of the rollup node..
-    blocks_pending_finalization: HashMap<[u8; 32], Block>,
+    blocks_pending_finalization: HashMap<u32, Block>,
```
noot, line 232: is this rollup_block_height to block?

crates/astria-conductor/src/executor/tests.rs
```diff
@@ -1,554 +1,27 @@
...
-        client,
-    }
-}
-
-fn make_rollup_data(data: &str) -> RawRollupData {
-    RollupData::SequencedData(data.as_bytes().to_vec()).into_raw()
-}
-
-#[tokio::test]
-async fn firm_blocks_at_expected_heights_are_executed() {
```
noot, line 275: why are all these deleted?

crates/astria-conductor/tests/blackbox/soft_and_firm.rs
```diff
@@ -0,0 +1,318 @@
...
+        Duration::from_millis(1000),
+        join3(
+            execute_block.wait_until_satisfied(),
+            update_commitment_state_soft.wait_until_satisfied(),
+            update_commitment_state_firm.wait_until_satisfied(),
+        ),
+    )
+    .await
+    .expect(
+        "conductor should have executed the soft block and updated the soft commitment state \
```
noot, line 139:
```suggestion
        "conductor should have executed the block and updated the soft and firm commitment state \
```

crates/astria-conductor/src/executor/mod.rs
joroshiba: Would be good to have a comment above on what the height is for next person through.

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -272,25 +328,23 @@ impl Executor {
     /// Panics if the `u32` underlying the celestia block variance tracked in the state could
     /// not be converted to a `usize`. This should never happen on any reasonable architecture
     /// that Conductor will run on.
-    fn calculate_max_spread(&self) -> usize {
-        usize::try_from(self.state.borrow().celestia_block_variance())
+    fn calculate_max_spread(&mut self) -> usize {
```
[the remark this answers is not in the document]
SuperFluffy: This actually begs the question: why batch-request all these blocks in the first place? We could just as well request the parent-block from the rollup as we go instead of maintaing a cache.

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -369,52 +430,67 @@ impl Executor {
...
+                    "found pending block; updating state but not not re-executing it"
+                );
+                Update::OnlyFirm(block)
+            } else {
+                info!(
+                    block_number,
+                    "did not find pending block; executing firm block"
+                );
+                // The parent hash of the next block is the hash of the block at the current head.
+                let parent_hash = self.state.firm_hash();
```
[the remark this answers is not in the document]
SuperFluffy: The entire logic is faulty: if `sort > firm` and there is no soft block number, then that's a hard error and the block must not be executed.

As you note, this code is not new but can be traced back to the original implementation last year. I am going to address this in a follow-up PR.

(Actually, this begs the question: why batch-request all these blocks in the first place? We could just as well request the parent-block from the rollup as we go instead of maintaining a cache. I am going to leave the implementation as it is for now, but just food for thought.)

crates/astria-conductor/src/executor/mod.rs
SuperFluffy: `firm_blocks` and `soft_blocks` are the channels that are created during executor construction.

We use these as proxies for which "mode" executor operates under.

crates/astria-conductor/src/executor/mod.rs
SuperFluffy: Yes, this was confusing. Changed to:
```
    /// Tracks rollup blocks by their rollup block numbers.
    ///
    /// Required to mark firm blocks received from celestia as executed
    /// without re-executing on top of the rollup node.
```

crates/astria-conductor/src/executor/mod.rs
SuperFluffy: You are right, this is not actionable information. I have removed the `pending_blocks` field entirely. The `info!` event now leans on the `start` and `end` block numbers that are part of the `populate_blocks_pending_finalization` span.

crates/astria-conductor/src/executor/reporting.rs
SuperFluffy: Removed as no longer used.

SuperFluffy:
> only minors, but concerned about the logging

Addressed all comments. In particular I added the following new code:

1. `executor::Client::batch_get_blocks` now ensures that the rollup response does not violate the "batch" contract (no holes, no dupes, no extra blocks outside the requested range)
2. the event no longer emits all blocks that the batch-get-block RPC returned. It only emits a message now (start and end numbers are already part of the span that the event is under)

noot (approved):
looks good!

crates/astria-conductor/src/executor/mod.rs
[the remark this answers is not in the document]
SuperFluffy: https://github.com/astriaorg/astria/issues/1011

joroshiba (approved):
I don't think fixing the contract code is necessarily blocking, but it can be easily simplified and I don't think validates the actual contract. In this instance we are never requesting dupes, but we aren't validating order matches. Makes it less useful from a "make sure my rollup is behaiving correctly" sense.

crates/astria-conductor/src/executor/client.rs
```diff
@@ -168,3 +183,119 @@ impl Client {
...
+    )]
+    Holes(HashSet<u32>),
+    #[error(
+        "extra blocks for numbers: [{}]",
+        itertools::join(.0, ", "),
+    )]
+    Extras(HashSet<u32>),
+}
+
+fn ensure_batch_get_blocks_is_correct(
```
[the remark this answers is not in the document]
SuperFluffy: This was a good point. I have changed simplified the code accordingly and amended the tests.

crates/astria-conductor/src/executor/client.rs
```diff
@@ -168,3 +183,119 @@ impl Client {
...
+    mut blocks: Vec<RawBlock>,
+    requested_numbers: RangeInclusive<u32>,
+) -> Result<Vec<RawBlock>, BatchGetBlocksError> {
+    blocks.sort_unstable_by_key(|block| block.number);
```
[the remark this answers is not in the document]
SuperFluffy: Contract is now actually checked! :)

Issue #913, opened 2024-04-03
sequencer: bridge withdrawal handler


noot:
we need to implement some sort of bridge withdrawal handler process that will:
1. watch for transfers to a specific account on the evm
2. submit a corresponding withdrawal (transfer out of bridge account) on the sequencer

this is functionally similar to the composer in that it watches a rollup and submits txs to the sequencer. however, it differs in that it watches for transfers/potentially events instead of the mempool, and it submits `Transfer` actions instead of `Sequence` actions.

noot:
closed by #984

PR #984, opened 2024-04-19
feat(bridge-withdrawer): PoC astria-bridge-withdrawer implementation

### LINKED ISSUE #1109: bridge-withdrawer: Add event derivation logic for sequencer-native withdrawals
The rollup event -> sequencer action derivation logic is out of date with the updated bridge contracts as of pr #1101.

itamarreif:
## Summary
implement `ethereum` and `submitter` modules for astria-bridge-withdrawer.

## Background
required to implement rollup bridge withdrawals.

## Changes

ethereum:
- implement `AstriaWithdrawer.sol` contract, put inside `ethereum/` dir as a foundry project
- implement `ethereum` module which contains a `Watcher` that watches the contract for `Withdrawal` events
- implement event to action conversion
- the events are sent to a `Batcher` which batches all events by block number and sends a batch of actions to a sequencer handler 

sequencer: 
- submit txs to sequencer when a batch of actions is received

## Testing
unit tests

manual testing:
1. run sequencer+cometbft as normal (`just run` + `just run-cometbft` in astria-sequencer)
2. run anvil as normal (`anvil` in a terminal)
3.  init the bridge account: 
```sh
export SEQUENCER_PRIVATE_KEY=2bd806c97f0e00af1a1fc3328fa763a9269723c8db8fac4f93af71db186d6e90
./target/debug/astria-cli sequencer init-bridge-account --sequencer-url=http://localhost:26657 --rollup-name=astria --sequencer.chain-id=astria
```
4. `just copy-env` in `astria-bridge-withdrawer`, put the private key from above into a file, update .env to use that 
5. go to `astria-bridge-withdrawer/ethereum`, `cp local.env.example .env`
6. deploy the withdrawer contract:
```sh
source .env
forge create --rpc-url $RPC_URL --private-key $PRIVATE_KEY src/AstriaWithdrawer.sol:AstriaWithdrawer
```
7. run this twice, makes 2 withdrawals (forces anvil to make 2 blocks) note: the withdrawer will only submit the withdraw for the first tx, as it's waiting for the next block before submitting the second and anvil doesn't auto-build blocks by default
```sh
forge script script/AstriaWithdrawer.s.sol:AstriaWithdrawerScript \
   --rpc-url $RPC_URL --broadcast --sig "withdrawToSequencer()" -vvvv
```
8.  in `astria-bridge-withdrawer`, update `ASTRIA_BRIDGE_WITHDRAWER_ETHEREUM_CONTRACT_ADDRESS` to be the contract address deployed in step 6.
9. `just run` and you should see a withdrawal being submitted:
```sh
2024-05-24T19:36:22.952210Z DEBUG astria_bridge_withdrawer::withdrawer::submitter: fetched latest nonce nonce=1
2024-05-24T19:36:23.003380Z DEBUG astria_bridge_withdrawer::withdrawer::submitter: signed transaction tx_hash=faad567d4f71ba20bdd9a1f679c4c3875c4b172c96d236155d52eb7ca02cd389
2024-05-24T19:36:23.003742Z DEBUG submit_tx: astria_bridge_withdrawer::withdrawer::submitter: submitting signed transaction to sequencer nonce=1 transaction.hash=faad567d4f71ba20bdd9a1f679c4c3875c4b172c96d236155d52eb7ca02cd389
2024-05-24T19:36:24.049943Z  INFO astria_bridge_withdrawer::withdrawer::submitter: withdraw batch successfully executed. sequencer.block=881 sequencer.tx_hash=FAAD567D4F71BA20BDD9A1F679C4C3875C4B172C96D236155D52EB7CA02CD389 rollup.height=2
```

## Related Issues

 #913

closes #1109


noot:
note to self: add a config value for asset precision

crates/astria-core/src/protocol/transaction/v1alpha1/action.rs
```diff
@@ -694,26 +694,26 @@ enum MintActionErrorKind {
 ///
 /// It also contains a `return_address` field which may or may not be the same as the signer
 /// of the packet. The funds will be returned to the `return_address` in the case of a timeout.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq, Eq)]
```
SuperFluffy, line 697: All fields should be private.

crates/astria-bridge-withdrawer/src/withdrawer/submitter/signer.rs
```diff
@@ -0,0 +1,34 @@
...
+pub(super) struct SequencerSigner {
+    pub(super) address: Address,
+    pub(super) signing_key: SigningKey,
+}
+
+impl SequencerSigner {
+    /// Construct a `SequencerSigner` from a file.
+    ///
+    /// The file should contain a hex-encoded ed25519 secret key.
+    pub(super) fn from_path<P: AsRef<Path>>(path: P) -> eyre::Result<Self> {
```
SuperFluffy, line 22:
```suggestion
    pub(super) fn try_from_path<P: AsRef<Path>>(path: P) -> eyre::Result<Self> {
```

crates/astria-bridge-withdrawer/src/withdrawer/submitter/signer.rs
```diff
@@ -0,0 +1,34 @@
...
+};
+
+use astria_core::crypto::SigningKey;
+use astria_eyre::eyre::{
+    self,
+    eyre,
+};
+use sequencer_client::Address;
+
+pub(super) struct SequencerSigner {
```
SuperFluffy, line 13: This type does not in fact sign anything as far as I can see? I would expect there to be a method `SequencerSigner::sign`.

crates/astria-bridge-withdrawer/src/withdrawer/submitter/mod.rs
```diff
@@ -0,0 +1,286 @@
...
+use super::{
+    batch::Batch,
+    state,
+};
+
+mod builder;
+mod signer;
+
+pub(super) struct Handle {
+    pub(super) batches_tx: mpsc::Sender<Batch>,
```
SuperFluffy, line 51: Make this private. No point having a handle if it just reveals its contents. Also, please don't overload abbreviations. In other parts of this crate `_tx` is used as `transaction`, whereas here it's used as `sending`

crates/astria-bridge-withdrawer/src/withdrawer/submitter/builder.rs
```diff
@@ -0,0 +1,99 @@
...
+    pub(crate) shutdown_token: CancellationToken,
+    pub(crate) sequencer_key_path: String,
+    pub(crate) sequencer_chain_id: String,
+    pub(crate) cometbft_endpoint: String,
+    pub(crate) state: Arc<State>,
+}
+
+impl Builder {
+    /// Instantiates an `Submitter`.
+    pub(crate) async fn build(self) -> eyre::Result<(super::Submitter, super::Handle)> {
```
SuperFluffy, line 32: 1. There are multiple errors in here that don't have context
2. This constructor should ideally not be async

crates/astria-bridge-withdrawer/src/withdrawer/submitter/builder.rs
```diff
@@ -0,0 +1,99 @@
...
+        } = self;
+
+        let signer = super::signer::SequencerSigner::from_path(sequencer_key_path)?;
+        info!(address = %telemetry::display::hex(&signer.address), "loaded sequencer signer");
+        let (batches_tx, batches_rx) = tokio::sync::mpsc::channel(BATCH_QUEUE_SIZE);
+
+        let sequencer_cometbft_client = sequencer_client::HttpClient::new(&*cometbft_endpoint)
+            .context("failed constructing cometbft http client")?;
+
+        let actual_chain_id = get_sequencer_chain_id(sequencer_cometbft_client.clone()).await?;
```
SuperFluffy, line 48: This could probably be moved into a "runtime intializer" that runs before the `select!` loop

This would remove the need to make this method async.

SuperFluffy (commented):
I have skimmed the PR and left a few early comments.

Please provide blackbox tests for this service.

crates/astria-core/src/protocol/transaction/v1alpha1/action.rs
noot: all the other actions have pub fields - we should probably make constructors for them instead? but a bit out of scope for here

crates/astria-bridge-withdrawer/ethereum/local.env.example
```diff
@@ -0,0 +1,20 @@
+# anvil account 0
+PRIVATE_KEY=0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80
```
joroshiba, line 2: How are these env vars used with contract deployment? Is this foundry env var? Generally we want to put private keys in files.

crates/astria-bridge-withdrawer/local.env.example
```diff
@@ -0,0 +1,71 @@
...
+# The address of the AstriaWithdrawer contract on the evm rollup.
+ASTRIA_BRIDGE_WITHDRAWER_ETHEREUM_CONTRACT_ADDRESS="0x"
+
+# The rpc endpoint of the evm rollup.
+ASTRIA_BRIDGE_WITHDRAWER_ETHEREUM_RPC_ENDPOINT="ws://127.0.0.1:8545"
+
+# The socket address at which the bridge service will server healthz, readyz, and status calls.
+ASTRIA_BRIDGE_WITHDRAWER_API_ADDR=127.0.0.1:2450
+
+# Set to true to enable prometheus metrics.
```
joroshiba, line 49:
```suggestion
# Set to true to disable prometheus metrics.
```

crates/astria-bridge-withdrawer/src/withdrawer/batch.rs
```diff
@@ -0,0 +1,304 @@
...
+};
+use astria_eyre::eyre::{
+    self,
+    OptionExt,
+    WrapErr as _,
+};
+use ethers::types::{
+    TxHash,
+    U64,
+};
```
joroshiba, line 25: I think it would be a bit more clear further down if these types weren't "used" and instead fully specify the type path below. I was confused about why `u64` was capitalized further down and it's good to have clarity on what type of txHash we are talking about (it's an EVM tx hash not a cometbft one)

crates/astria-core/src/protocol/transaction/v1alpha1/action.rs
joroshiba: can we make a follow up issue here? seems a general code quality thing

joroshiba (changes requested):
Some general comments on source of withdrawal contract code, only change I really want to see on PR though is refactoring the github ci test logic. I want to be able to isolate the tests using all the foundry stuff in case it fails on us again.

crates/astria-bridge-withdrawer/src/main.rs
```diff
@@ -0,0 +1,81 @@
...
+        .set_force_stdout(cfg.force_stdout)
+        .set_pretty_print(cfg.pretty_print)
+        .filter_directives(&cfg.log);
+
+    if !cfg.no_metrics {
+        telemetry_conf = telemetry_conf
+            .metrics_addr(&cfg.metrics_http_listener_addr)
+            .service_name(env!("CARGO_PKG_NAME"));
+    }
+    metrics_init::register();
```
SuperFluffy, line 40: Please register metrics the same way as all other services

.github/workflows/test.yml
```diff
@@ -111,21 +137,15 @@ jobs:
...
-          --all-features \
-          --all-targets
       - name: Run tests
         timeout-minutes: 20
         run: |
           pip install solc-select 
           solc-select install 0.8.21
           solc-select use 0.8.21
-          cargo nextest run --archive-file=archive.tar.zst -- --include-ignored
+          cargo nextest run --package astria-bridge-withdrawer -- --include-ignored
```
joroshiba, line 148:
```suggestion
          cargo nextest run --package astria-bridge-withdrawer -- --include-ignored
```

crates/astria-bridge-withdrawer/ethereum/local.env.example
SuperFluffy, line 1: Where is this file used?

crates/astria-bridge-withdrawer/ethereum/local.env.example
SuperFluffy: I agree. We should never leak private keys in plain text.

joroshiba (approved):

crates/astria-bridge-withdrawer/local.env.example
```diff
@@ -0,0 +1,71 @@
...
+
+# If set to any non-empty value removes ANSI escape characters from the pretty
+# printed output. Note that this does nothing unless
+# `ASTRIA_BRIDGE_WITHDRAWER_PRETTY_PRINT` is set to `true`.
+NO_COLOR=
+
+# Address of cometbft/tendermint to request new block heights.
+# 127.0.0.1:26657 is the default socket address at which cometbft
+# serves RPCs.
+ASTRIA_BRIDGE_WITHDRAWER_COMETBFT_ENDPOINT="http://127.0.0.1:26657"
```
SuperFluffy, line 25: Let's be clearer about this. This reads like it's the withdrawer's endpoint.
```suggestion
ASTRIA_BRIDGE_WITHDRAWER_SEQUENCER_COMETBFT_ENDPOINT="http://127.0.0.1:26657"
```

crates/astria-bridge-withdrawer/src/withdrawer/mod.rs
```diff
@@ -0,0 +1,330 @@
...
+    api,
+    config::Config,
+};
+
+mod batch;
+mod ethereum;
+mod state;
+mod submitter;
+
+pub struct Service {
```
SuperFluffy, line 43: Call this `WithdrawerBridge` following all our other services.

crates/astria-bridge-withdrawer/src/withdrawer/ethereum/test_utils.rs
```diff
@@ -0,0 +1,68 @@
...
+/// Returns the contract address, provider, wallet, and anvil instance.
+///
+/// # Panics
+///
+/// - if the contract cannot be found in the expected path
+/// - if the contract cannot be compiled
+/// - if the provider fails to connect to the anvil instance
+/// - if the contract fails to deploy
+#[allow(dead_code)]
+pub(crate) async fn deploy_astria_withdrawer()
```
SuperFluffy, line 24: Delete. We don't want dead code.

crates/astria-bridge-withdrawer/src/withdrawer/ethereum/watcher.rs
```diff
@@ -0,0 +1,493 @@
...
+pub(crate) struct Watcher {
+    contract: AstriaWithdrawer<Provider<Ws>>,
+    event_tx: mpsc::Sender<(WithdrawalEvent, LogMeta)>,
+    batcher: Batcher,
+    state: Arc<State>,
+    shutdown_token: CancellationToken,
+}
+
+impl Watcher {
+    pub(crate) async fn new(
```
SuperFluffy, line 54: Make this non-async and split out the construction code from the run-time initialization code.

The goal is to ensure that the service can shut down cleanly when getting a shutdown signal

crates/astria-bridge-withdrawer/src/withdrawer/ethereum/watcher.rs
```diff
@@ -0,0 +1,493 @@
...
+    let mut stream = events.stream().await.unwrap().with_meta();
+
+    while let Some(item) = stream.next().await {
+        if let Ok((event, meta)) = item {
+            event_tx
+                .send((WithdrawalEvent::Sequencer(event), meta))
+                .await
+                .wrap_err("failed to send sequencer withdrawal event; receiver dropped?")?;
+        } else if let Err(e) = item {
+            return Err(e).wrap_err("failed to read from event stream; event stream closed?");
```
SuperFluffy, line 168: Please simplify this. `item.wrap_err(...)?`

crates/astria-bridge-withdrawer/src/withdrawer/ethereum/watcher.rs
```diff
@@ -0,0 +1,493 @@
...
+
+    while let Some(item) = stream.next().await {
+        if let Ok((event, meta)) = item {
+            event_tx
+                .send((WithdrawalEvent::Ics20(event), meta))
+                .await
+                .wrap_err("failed to send ics20 withdrawal event; receiver dropped?")?;
+        } else if let Err(e) = item {
+            return Err(e).wrap_err("failed to read from event stream; event stream closed?");
+        }
```
SuperFluffy, line 195: As above, please simplify this.

crates/astria-bridge-withdrawer/src/withdrawer/ethereum/watcher.rs
```diff
@@ -0,0 +1,493 @@
...
+            rollup_height: 0,
+        };
+
+        loop {
+            select! {
+                () = self.shutdown_token.cancelled() => {
+                    info!("batcher shutting down");
+                    break;
+                }
+                item = self.event_rx.recv() => {
```
SuperFluffy, line 243: This `select!` statement should break if `self.event_rx.recv() == None` because that means the channel is closed.

crates/astria-bridge-withdrawer/src/withdrawer/ethereum/watcher.rs
```diff
@@ -0,0 +1,493 @@
...
+}
+
+// converts an ethereum address string to an `ethers::types::Address`.
+// the input string may be prefixed with "0x" or not.
+fn address_from_string(s: &str) -> Result<ethers::types::Address> {
+    let s = s.strip_prefix("0x").unwrap_or(s);
+    let bytes = hex::decode(s).wrap_err("failed to parse ethereum address as hex")?;
+    let address: [u8; 20] = bytes
+        .try_into()
+        .map_err(|_| eyre!("invalid length for ethereum address, must be 20 bytes"))?;
```
SuperFluffy, line 285: report the actual number of bytes received

crates/astria-core/src/protocol/transaction/v1alpha1/action.rs
SuperFluffy: Gonna let it pass here because the other actions also have their fields exposed. But yeah, in general I prefer to not leak that stuff. We should address this in a follow, prior to publishing this crate.

crates/astria-bridge-withdrawer/src/withdrawer/submitter/mod.rs
```diff
@@ -0,0 +1,322 @@
...
+            self.state.set_last_rollup_height_submitted(rollup_height);
+            self.state.set_last_sequencer_height(rsp.height.value());
+            self.state.set_last_sequencer_tx_hash(rsp.hash);
+            Ok(())
+        }
+    }
+}
+
+/// Queries the sequencer for the latest nonce with an exponential backoff
+#[instrument(name = "get_latest_nonce", skip_all, fields(%address))]
```
SuperFluffy, line 180: name matches the function name. You can remove that attribute.

SuperFluffy (dismissed):

SuperFluffy:
I would like a simple blackbox test here that shows that we can spin up the service, receive an event, and see it come out on the other side.

crates/astria-bridge-withdrawer/src/withdrawer/ethereum/test_utils.rs
noot: this isn't dead code, will remove the allow

crates/astria-bridge-withdrawer/ethereum/local.env.example
noot: this is anvil account 0 so it's a known private key. i can remove tho

crates/astria-bridge-withdrawer/src/withdrawer/batch.rs
noot: moved to the ethereum crate as this logic is eth-specific

crates/astria-bridge-withdrawer/ethereum/local.env.example
noot: this is used to deploy the ethereum contract and make calls to it - see the readme in this crate

crates/astria-bridge-withdrawer/src/withdrawer/mod.rs
noot: that cause a clippy name repetitions error since this module is already called `withdrawer` and this isn't a bridge, so i'd prefer not to name it that

noot:
@SuperFluffy comments are addressed, blackbox test will be done in a PR shortly

joroshiba (approved):
Re-approving noting this is blocking further development and blackbox tests are actively being built, we have a tight timeline can't delay on the tests.

PR #985, opened 2024-04-19
chore(sequencer): update justfile and testnet script

Lilyjjo:
## Summary
Updated the sequencer's justfile and testnet.sh script to work again with latest version of sequencer.

TODO: have someone test on their linux machine. I had to edit the script to work on a mac and tried to make the commands work on both but I'm unsure if I did it right.

## Testing
`just run-testnet` : to run testnet locally 
`just clean-testnet` : to cleanup testnet

### DIFF
--- a/crates/astria-sequencer/scripts/testnet.sh
+++ b/crates/astria-sequencer/scripts/testnet.sh
@@ -3,6 +3,15 @@
 set -o errexit -o nounset
 set -x
 
+# Function to perform sed in-place editing in a cross-platform way
+edit_in_place() {
+    case "$(uname)" in
+        Linux*) sed -i "$@" ;;
+        Darwin*) sed -i '' "$@" ;;
+        *) echo "Unsupported OS" >&2; exit 1 ;;
+    esac
+}
+
 : "${OUT_DIR:=sequencer_testnet}"
 : "${NUM_VALIDATORS:=3}"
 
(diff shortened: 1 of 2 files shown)


crates/astria-sequencer/scripts/testnet.sh
```diff
@@ -16,27 +25,38 @@ for i in $(seq 0 "$((NUM_VALIDATORS - 1))"); do
...
 
     # allow peers from the same IP
-    sed -i 's/allow_duplicate_ip = false/allow_duplicate_ip = true/' $OUT_DIR/node$i/config/config.toml
-    sed -i 's/addr_book_strict = true/addr_book_strict = false/' $OUT_DIR/node$i/config/config.toml
+    edit_in_place 's/allow_duplicate_ip = false/allow_duplicate_ip = true/' $OUT_DIR/node$i/config/config.toml
+    edit_in_place 's/addr_book_strict = true/addr_book_strict = false/' $OUT_DIR/node$i/config/config.toml
+
+    # update genesis file with app state
+    ../../target/debug/astria-sequencer-utils --genesis-app-state-file=test-genesis-app-state.json --destination-genesis-file=$OUT_DIR/node$i/config//genesis.json --chain-id=astria
+    sed -i'.bak' 's/timeout_commit = "1s"/timeout_commit = "2s"/g' ~/.cometbft/config/config.toml
```
noot, line 46: should this use `edit_in_place` also?

crates/astria-sequencer/scripts/testnet.sh
```diff
@@ -16,27 +25,38 @@ for i in $(seq 0 "$((NUM_VALIDATORS - 1))"); do
...
 # start sequencer application for each node
 for i in $(seq 0 "$((NUM_VALIDATORS - 1))"); do
-    APP_PORT=$((26660 + $i))
-    ../../target/debug/astria-sequencer --genesis-file=test-genesis.json --listen-addr=127.0.0.1:$APP_PORT &> $OUT_DIR/node$i/sequencer.log &
+    # edit default .env files for multiple validators 
+    DB_PATH="/tmp/astria_db_$i"
+    ASTRIA_SEQUENCER_LISTEN_ADDR=127.0.0.1:$((26660 + $i))
+    ASTRIA_SEQUENCER_GRPC_ADDR=127.0.0.1:$((8080 + $i))
+
+    export ASTRIA_SEQUENCER_DB_FILEPATH=$DB_PATH \
```
joroshiba, line 56: We have the `local.example.env` file for this,`just copy-env` and such to configure them. I'd prefer not to have more ways that env vars are set which can create confusion.

noot (commented):
script works for me! two things when testing locally:
1. the logs of both the app and cometbft get output to my terminal when running it, but they should get redirected to the logfile. are you seeing this also?
2. `pkill astria-sequencer` doesn't actually kill the sequencer process for me anymore, i think you'd need to get the pid of all `astria-sequencer` processes and kill them from that

crates/astria-sequencer/scripts/testnet.sh
noot: we'd have to manually make a new .env file for each node in that case

crates/astria-sequencer/scripts/testnet.sh
Lilyjjo: Since it has the '.bak' thing it doesn't need to. The command is for when sed doesn't have the first argument

Lilyjjo:
(1) and (2) might be machine dependent, the output gets directed for me and the pkill works. The original pkill for me didn't work on mac tho

crates/astria-sequencer/scripts/testnet.sh
Lilyjjo: I'd have to have the script edit those variables even with copying. This file is pretty brittle in general

noot (approved):
approving, i'll make a follow up for the small linux fixes!

crates/astria-sequencer/scripts/testnet.sh
noot: yeah just leave as-is, this is fine

PR #987, opened 2024-04-19
chore(proto)!: create wrapper types for `RollupId` and `Account`

joroshiba:
## Summary
Creates wrapper proto types for `RollupId` and `Account`, which can allow us to add more representations and acceptable formats for them as non breaking changes in the future.

## Background
We were using `Account` and `RollupId` as simple raw bytes, this means any change would be a breaking change, now we can do non breaking changes such as offering bech32 address versions.

## Changes
- Switched to wrapper types for all references of RollupId and Account

## Testing
CI/CD

## Metrics

## Breaking Changelist
- Changes core transaction protos

### DIFF
--- a/crates/astria-core/src/sequencerblock/v1alpha1/celestia.rs
+++ b/crates/astria-core/src/sequencerblock/v1alpha1/celestia.rs
@@ -237,7 +237,7 @@ impl CelestiaRollupBlob {
         } = self;
         raw::CelestiaRollupBlob {
             sequencer_block_hash: sequencer_block_hash.to_vec(),
-            rollup_id: rollup_id.to_vec(),
+            rollup_id: Some(rollup_id.to_raw()),
             transactions,
             proof: Some(proof.into_raw()),
         }
@@ -254,8 +254,11 @@ impl CelestiaRollupBlob {
             transactions,
             proof,
         } = raw;
+        let Some(rollup_id) = rollup_id else {
+            return Err(CelestiaRollupBlobError::field_not_set("rollup_id"));
+        };
         let rollup_id =
-            RollupId::try_from_vec(rollup_id).map_err(CelestiaRollupBlobError::rollup_id)?;
+            RollupId::try_from_raw(rollup_id).map_err(CelestiaRollupBlobError::rollup_id)?;
         let sequencer_block_hash = sequencer_block_hash
             .try_into()
             .map_err(|bytes: Vec<u8>| CelestiaRollupBlobError::sequencer_block_hash(bytes.len()))?;
@@ -435,7 +438,7 @@ impl UncheckedCelestiaSequencerBlob {
         }?;
         let rollup_ids: Vec<_> = rollup_ids
             .into_iter()
-            .map(RollupId::try_from_vec)
+            .map(RollupId::try_from_raw)
             .collect::<Result<_, _>>()
             .map_err(CelestiaSequencerBlobError::rollup_ids)?;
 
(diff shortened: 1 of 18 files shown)


joroshiba:
May need to move pushing protos to only on merge to main.

The failing test I will update after another merge which makes this a snapshot test, avoid conflicts here.

crates/astria-core/src/sequencerblock/v1alpha1/celestia.rs
```diff
@@ -435,7 +438,7 @@ impl UncheckedCelestiaSequencerBlob {
         }?;
         let rollup_ids: Vec<_> = rollup_ids
             .into_iter()
-            .map(RollupId::try_from_vec)
+            .map(|raw| RollupId::try_from_raw(&raw))
```
SuperFluffy, line 441: Very minor comment:

You could express this more elegantly (similar to before) like so:
```rust
let rollup_ids: Vec<_> = rollup_ids
    .iter()
    .map(RollupId::try_from_raw)
```

crates/astria-sequencer/src/grpc/sequencer.rs
```diff
@@ -89,7 +89,7 @@ impl SequencerService for SequencerServer {
         let rollup_ids = request
             .rollup_ids
             .into_iter()
-            .map(RollupId::try_from_vec)
+            .map(|raw| RollupId::try_from_raw(&raw))
```
SuperFluffy, line 92: Similar to my previous comment, using `request.rollup_ids.iter().map(RollupId::try_from_vec)` would make this a tad more elegant.

SuperFluffy (approved):
I like this change. I was pondering if we should make it. Seeing it now I agree that we should.

This makes the protobuf more complex because we no longer have direct `bytes` fields flattened into the various messages that use addresses and rollup IDs. A big annoyance is the forced use of `Some(...)` everywhere an address/rollup ID is used.

However, this is strongly outweighed by the advantages:

1. in our rust code, there will be one single source of truth for the display-formatting of rollup IDs and addresses. We will have to manually implement `serde::Serialize`/`serde::Deserialize` for these protobuf types and exclude them from the protobuf compiler tool (context: at the moment higher-level Rust types delegate to the generated Rust types - this means that if these types contain rollup IDs or addresses, these are formatted as base64 using the pbjson-generated serde traits. On the other hand, our hand written `RollupId`/`Address` types also contain hand-written serde types that format these as base64. This is a potential source of breakage).
2. I believe this removes a potential source of errors for 3rd parties interacting with our APIs. Instead of passing around a bunch of bytes, the use of `RollupId` and `Address` is now made explicit.

.github/workflows/release.yml
```diff
@@ -69,7 +83,7 @@ jobs:
       contents: write
     runs-on: ubuntu-latest
     needs: run_checker
-    if: needs.run_checker.outputs.run_release_charts == 'true'
+    if: github.ref_name == 'main' && needs.run_checker.outputs.run_release_charts == 'true'
```
joroshiba, line 86: I realized while cleaning up the above that this would be deploying in merge queue, we should only ever run on main.

.github/workflows/release.yml
```diff
@@ -14,13 +14,21 @@ jobs:
     runs-on: ubuntu-latest
     environment: BUF
     needs: run_checker
-    if: needs.run_checker.outputs.run_release_proto == 'true' && github.repository_owner == 'astriaorg'
+    if: github.event_name != 'merge_group' && needs.run_checker.outputs.run_release_proto == 'true' && github.repository_owner == 'astriaorg'
```
joroshiba, line 17: Merge group branch naming is very weird, and offers no real benefit to push to buf here. Whereas pushing PRs enables testing and testing in other repos.

WafflesVonMaple (approved):

noot (approved):

PR #990, opened 2024-04-22
fix(composer): account for fee asset id while estimating sequence action size

bharath-123:
## Summary
Account for the length of fee_asset_id while estimating the sequence action size in composer.

## Background
In Composer, we create bundles with the incoming txs collected from the gRPC and Geth collectors. 
These bundles have to be appropriately sized so that we don't end up sending a tx to the sequencer which is much bigger than the block size or the max tx size accepted by the sequencer. 
To do this, we have a variable `ASTRIA_COMPOSER_MAX_BYTES_PER_BUNDLE` which is the max bytes allowed in a bundle. 
Currently, this was not considering the length of the fee_asset_id field of a SequenceAction. When a large inflow of txs are sent to the composer which causes the bundles to be completely filled, the submission of the bundle to the sequencer fails because the unaccounted fee_asset_id fields cause the tx size to go beyond the accepted tx size by the sequencer. 
It fails with the below error:
`2024-04-19T07:57:14.930902Z  WARN run_until_stopped:submit_bundle: astria_composer::executor: sequencer rejected the transaction; the bundle is likely lost abci.code=5 abci.log="transaction size too large; allowed: 256000 bytes, got 305106" address=Uw2IlR9YdowCe09slJO3kwRg+KQ= nonce.initial=1`

To fix this, we now account for the fee_asset_id len while estimating the sequence action size.

## Changes
- Create a new var `FEE_ASSET_ID_LEN` in astria-core primitives.
- Account for `FEE_ASSET_ID_LEN` in `estimate_size_of_sequence_action` method.

## Testing
We tested this in 2 ways:
1. Unit tests were updated to account for fee_asset_id and they successfully pass
2. Running a load test where a lot of txs were sent to Composer. Before this fix, the txs were being dropped because the tx being sent to the sequencer was too big.

### DIFF
--- a/crates/astria-composer/src/executor/bundle_factory/mod.rs
+++ b/crates/astria-composer/src/executor/bundle_factory/mod.rs
@@ -214,5 +214,9 @@ impl<'a> NextFinishedBundle<'a> {
 
 /// The size of the `seq_action` in bytes, including the rollup id.
 fn estimate_size_of_sequence_action(seq_action: &SequenceAction) -> usize {
-    seq_action.data.len() + ROLLUP_ID_LEN + FEE_ASSET_ID_LEN
+    seq_action
+        .data
+        .len()
+        .saturating_add(ROLLUP_ID_LEN)
+        .saturating_add(FEE_ASSET_ID_LEN)
 }
--- a/crates/astria-composer/src/executor/bundle_factory/mod.rs
+++ b/crates/astria-composer/src/executor/bundle_factory/mod.rs
@@ -11,6 +11,7 @@ use std::{
 use astria_core::{
     primitive::v1::{
         RollupId,
+        FEE_ASSET_ID_LEN,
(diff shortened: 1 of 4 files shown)


noot (approved):

crates/astria-composer/src/executor/bundle_factory/mod.rs
```diff
@@ -213,5 +214,5 @@ impl<'a> NextFinishedBundle<'a> {
 
 /// The size of the `seq_action` in bytes, including the rollup id.
 fn estimate_size_of_sequence_action(seq_action: &SequenceAction) -> usize {
-    seq_action.data.len() + ROLLUP_ID_LEN
+    seq_action.data.len() + ROLLUP_ID_LEN + FEE_ASSET_ID_LEN
```
SuperFluffy, line 217: Please make these saturating additions

Issue #715, opened 2024-01-30
sequencer: ensure failed txs are removed from the mempool


noot:
problem: the cometbft mempool does not remove txs from the mempool until it sees that it was included in a block (ok or failed). however, due to including only successful txs in a block via `prepare_proposal`, `deliver_tx` is not called for txs that will fail (as it's only called for what's in the proposed block), so cometbft never knows that the tx failed, and keeps sending it to the app each block. this will also result in a bunch of txs that will never succeed staying in the mempool.

a partial solution for txs which do not include `sequence::Action`s is to include them in the block (success or fail), however txs which include `sequence::Action`s cannot be included in a block if they fail as they'll be "sequenced" for free. the only way to fix this completely is 1. fork cometbft, 2. make sequenced block data not available on the sequencer level somehow.

noot:
we will go with a config option for mempool TTL for now

noot:
```

noot:
closed by #1138

PR #992, opened 2024-04-22
fix(sequencer): don't store execution result of failed tx

noot:
## Summary
this was causing a cometbft consensus failure - we should only return tx results for txs included in the block, which are only txs which pass execution.

## Changes
- don't store execution result of failed tx, fixes consensus failure as it now returns the correct number of execution results.

## Testing
ran with cometbft v0.38.6

### DIFF
--- a/crates/astria-sequencer/src/app.rs
+++ b/crates/astria-sequencer/src/app.rs
@@ -483,20 +483,10 @@ impl App {
                 Err(e) => {
                     debug!(
                         transaction_hash = %telemetry::display::base64(&tx_hash),
+                        error = AsRef::<dyn std::error::Error>::as_ref(&e),
                         "failed to execute transaction, not including in block"
                     );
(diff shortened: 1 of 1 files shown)


Lilyjjo (approved):
We might need to add this back if we go the route of letting transactions fail (re: #715). Otherwise, LGTM!

joroshiba (approved):

Lilyjjo:
Also just to note, it's unclear if we should remove the error push in `finalize_block()` too. The cometBFT error is around a mismatch between the execution results length and number of returned transactions, so removing it there would probably also cause a consensus error. Considering that being in that state is already a protocol error I don't think it super matters.

noot:
@Lilyjjo , yeah, if there's an error result in `finalize_block` it's a protocol error so there will be a consensus failure no matter what the app returns, so i think it's ok to leave for now

PR #1001, opened 2024-04-24
feat(sequencer-relayer)!: provide filter for rollups

Fraser999:
## Summary
A new configuration env var has been added to allow filtering data submission to a subset of rollups.

## Background
This allows more flexibility in what is actually submitted to the Celestia chain.

## Changes
A new env var was added, mapped to a new field `Config::rollup_id_filter`.  This represents a list of rollups whose transactions should be included in blobs submitted to Celestia.  Rollups not in the list do not have their data submitted.  However, if the list is empty, no filtering is applied; all rollups have their data submitted.

The collection of filtered rollup IDs is passed down to the `write::conversion::convert` function where sequencer blocks are converted to blobs.  The blobs are filtered inside this function, partly since this is relatively early in the flow, but also so that the `ConversionInfo` struct can carry information about blobs being excluded.  This data is logged at the start of `write::submit_blobs` in an info-level log message.  It might be worthwhile adding metrics around the number of excluded rollups per submission, but I didn't see a clear benefit to that and it can be added in a follow-up PR if required.

## Testing
- Added unit tests covering parsing of the config string to a collection of `RollupId`s.
- ~Added a black box test where a sequencer block is constructed with several transactions, half from included rollups and half from excluded ones.  Existing black box tests ensure that when the filter is empty, no filtering is done.~ These all got removed in #963, and should be restored as part of #1008.

## Breaking Changelist
- Added `ASTRIA_SEQUENCER_RELAYER_ONLY_INCLUDE_ROLLUPS` which can be empty or can be a comma-separated list.


crates/astria-sequencer-relayer/tests/blackbox/helper.rs
```diff
@@ -307,13 +307,23 @@ impl TestSequencerRelayer {
     }
 }
 
+// allow: this is not performance-critical, with likely only one instance per test fixture.
+#[allow(clippy::large_enum_variant)]
+pub enum CometBftBlockToMount {
+    GoodAtHeight(u32),
+    BadAtHeight(u32),
+    Block(tendermint::Block),
+}
```
noot, line 316: just flagging that i updated these tests to use `SequencerBlock`s into of cometbft blocks, will need to be updated

noot (approved):
looks good!

crates/astria-sequencer-relayer/src/config.rs
```diff
@@ -38,18 +54,105 @@ pub struct Config {
...
+
+        // No entries: "".
+        assert!(create_filter("").unwrap().is_empty());
+    }
+
+    #[test]
+    fn should_fail_to_create_filter_from_bad_input() {
+        // Invalid base64 encoding.
+        let input = "CAgICAgICAgICAgICAgICAgICAgICAgICAgICAgICAg!";
+        assert!(create_filter(input).is_err());
```
SuperFluffy, line 152: My preference for the sort of tests is to `unwrap_err` instead of asserting on a boolean. If this fails then the person for whom this failed (i.e. `create_filter` returned with success) will have to start rewriting the test to figure out why this is `false`

crates/astria-sequencer-relayer/src/config.rs
```diff
@@ -38,18 +54,105 @@ pub struct Config {
...
+
+        // Invalid decoded length (31 bytes).
+        let input = BASE64_STANDARD.encode([0; 31]);
+        assert!(create_filter(&input).is_err());
```
SuperFluffy, line 156: Same as above, prefer `unwrap_err` over asserting on booleans for UX reasons.

crates/astria-sequencer-relayer/src/relayer/write/conversion.rs
```diff
@@ -36,23 +38,33 @@ pub(super) struct RollupInfo {
...
-pub(super) fn convert(block: SequencerBlock) -> eyre::Result<Converted> {
+/// Convert the given sequencer block into a collection of blobs and related metadata.
+///
+/// If `include` is empty, blobs from all rollups will be included.  Otherwise only blobs from the
+/// rollups specified in `include` will be included.
+// allow: we'd need static lifetime on a ref to avoid pass-by-value here.
+#[allow(clippy::needless_pass_by_value)]
+pub(super) fn convert(
+    block: SequencerBlock,
+    include: HashSet<RollupId>,
```
SuperFluffy, line 66: I understand the design decision here, but I am worried that this might quickly become confusing a few months down the line.

Instead of passing this `HashSet<RollupId>` around, could this be made into a newtype wrapper `IncludeRollup(HashSet<RollupId>)` with a single `IncludeRollup::include: rollup_id -> bool` that contains this logic?

Extra idea: if we make this `IncludeRollup<Arc<HashSet<RollupId>>` then this can be cheaply cloned (and the `needless_pass_by_value` avoided?).

crates/astria-sequencer-relayer/src/relayer/mod.rs
```diff
@@ -82,6 +84,9 @@ pub(crate) struct Relayer {
     // If this is set, only relay blocks to DA which are proposed by the same validator key.
     validator: Option<Validator>,
 
+    // The rollup IDs to include in submissions (all rollups if filter is empty).
+    rollup_id_filter: HashSet<RollupId>,
```
SuperFluffy, line 88: Could this be made a `Arc<HashSet<RollupId>>` since this will never be mutated again? This way we can avoid the (moderatly expensive) clones.

crates/astria-sequencer-relayer/local.env.example
```diff
@@ -44,6 +44,12 @@ ASTRIA_SEQUENCER_RELAYER_BLOCK_TIME=1000
 # Set to true to only relay blocks proposed by the validator key in `ASTRIA_SEQUENCER_RELAYER_VALIDATOR_KEY_FILE`.
 ASTRIA_SEQUENCER_RELAYER_RELAY_ONLY_VALIDATOR_KEY_BLOCKS=false
 
+# A comma-separated list of base64-encoded rollup IDs.
```
SuperFluffy, line 47: We should probably explicitly spell out that rollup IDs are 32 byte hashes, in case somebody is confused (maybe "often the sha256 hash of a rollup name").

SuperFluffy (commented):
I agree with the direction of this PR.

The only thing I am worried about is the relatively complex use of `filter.is_empty() || filter.contains(&rollup_id)` to test if a block should be permitted. Could we make this into a dedicated type with a single `filter -> bool` method instead?

This feels like the classic thing that is unclear a few months down (plus this gives us the chance to make all these clones cheaper).

crates/astria-sequencer-relayer/tests/blackbox/helper.rs
Fraser999: Thanks!  Unfortunately a recent PR deleted most of these tests, but #1008 will see them restored.  I'll merge our changes so when the tests are restored, we can reintroduce the test I initially added in this PR, but which has been deleted after merging `main`.

crates/astria-sequencer-relayer/src/relayer/write/conversion.rs
Fraser999: The newtype makes sense to me, but the `Arc` seems a little overkill given I'm expecting the set to be empty or at most have one or two entries, with the entries being `Copy` types.  It's also only being cloned once per sequencer block.  Saying that, I guess the `Arc` can't hurt, so I'll add it :)

The `needless_pass_by_value` still won't be avoided though, since Clippy wants me to change the signature to take a ref to the set, and the lifetime on it would need to be `'static` if I did that.

crates/astria-sequencer-relayer/tests/blackbox/helper.rs
Fraser999: Done in [81686cc](https://github.com/astriaorg/astria/pull/1001/commits/81686cc4a3d4e1540b2f594e06e622b6dc597dc8).

crates/astria-sequencer-relayer/src/config.rs
Fraser999: Done in [aa8957e](https://github.com/astriaorg/astria/pull/1001/commits/aa8957e8c1a7a2d49742e755f3c4b0617fbda1e9).

crates/astria-sequencer-relayer/src/config.rs
Fraser999: Done in [aa8957e](https://github.com/astriaorg/astria/pull/1001/commits/aa8957e8c1a7a2d49742e755f3c4b0617fbda1e9).

crates/astria-sequencer-relayer/src/relayer/write/conversion.rs
Fraser999: Done in [02b1276](https://github.com/astriaorg/astria/pull/1001/commits/02b1276267c6ac65b72c78191ebecdfae92e72d9).

crates/astria-sequencer-relayer/src/relayer/mod.rs
Fraser999: Done in [02b1276](https://github.com/astriaorg/astria/pull/1001/commits/02b1276267c6ac65b72c78191ebecdfae92e72d9).

crates/astria-sequencer-relayer/local.env.example
Fraser999: Done in [cb1238b](https://github.com/astriaorg/astria/pull/1001/commits/cb1238b701b520d1bdbd778f544777afb9242736).

SuperFluffy (approved):
Looks great, thank you for the PR.

I am not sure about the state of the blackbox since #963 removed a ton of that functionality.

Since #1008 is a follow-up PR that is imminent I would ask for having a special test specifically for using this filtering functionality.

Other that I only have some comments here and there and suggestions for maybe more explicit naming, none of which is blocking.

crates/astria-sequencer-relayer/src/config.rs
```diff
@@ -38,18 +55,122 @@ pub struct Config {
...
 
     const EXAMPLE_ENV: &str = include_str!("../local.env.example");
 
     #[test]
     fn example_env_config_is_up_to_date() {
         config::tests::example_env_config_is_up_to_date::<Config>(EXAMPLE_ENV);
     }
+
+    #[test]
+    fn should_create_filter() {
```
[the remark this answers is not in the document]
Fraser999: Yes, I generally favour that style too.  But in this case, most of the assertions are one-liners, so I didn't think it was worthwhile.

crates/astria-sequencer-relayer/src/config.rs
```diff
@@ -38,18 +55,122 @@ pub struct Config {
...
+        let rollup_ids: HashSet<_> = (0..10).map(|i| RollupId::new([i; 32])).collect();
+
+        // Normal form: "aaa,bbb,ccc".
```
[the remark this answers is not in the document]
Fraser999: I'm thinking probably not, as there's a much greater chance of the `IncludeRollup::new(&input).unwrap()` panicking than the parsing reporting success but actually returning a different set.  If the `unwrap` panics, we don't see the assertion message.

crates/astria-sequencer-relayer/local.env.example
```diff
@@ -45,6 +45,13 @@ ASTRIA_SEQUENCER_RELAYER_BLOCK_TIME=1000
 # Set to true to only relay blocks proposed by the validator key in `ASTRIA_SEQUENCER_RELAYER_VALIDATOR_KEY_FILE`.
 ASTRIA_SEQUENCER_RELAYER_RELAY_ONLY_VALIDATOR_KEY_BLOCKS=false
 
+# A comma-separated list of base64-encoded rollup IDs, often the sha256
+# hash of a rollup name.
+# Only data from these rollups will be submitted to Celestia.
+# If empty, no filtering will be applied, i.e. data from all
+# rollups will be submitted.
```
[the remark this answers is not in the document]
Fraser999: Done in [c745b3f](https://github.com/astriaorg/astria/pull/1001/commits/c745b3f359ac092260ddbeee83289debf9082373).

crates/astria-sequencer-relayer/local.env.example
```diff
@@ -45,6 +45,13 @@ ASTRIA_SEQUENCER_RELAYER_BLOCK_TIME=1000
...
+ASTRIA_SEQUENCER_RELAYER_ROLLUP_ID_FILTER=
```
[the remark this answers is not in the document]
Fraser999: Done in [8946615](https://github.com/astriaorg/astria/pull/1001/commits/89466159194d778fe103782251c4bb1920a73c6a).

joroshiba (approved):
I believe we will need to bump chart version value again, after my merge but looks good to me otherwise.

PR #1005, opened 2024-04-24
refactor(core): remove `SequencerBlock::try_from_cometbft`

### LINKED ISSUE #969: remove `SequencerBlock::try_from_cometbft` and `try_from_cometbft_header_and_data`
we can remove these functions as we should no longer construct `SequencerBlock`s from cometbft blocks (due to deposits existing). however these functions are still used in tests, so we should update those tests to construct `SequencerBlock`s differently.

noot:
## Summary
 remove `SequencerBlock::try_from_cometbft` and `try_from_cometbft_header_and_data`.

## Background
these functions are no longer legitimate with the addition of the `Deposit` rollup data type, which are not contained in a cometbft block. we're also moving towards using `SequencerBlock` everywhere instead of the cometbft types.

## Changes
- remove `SequencerBlock::try_from_cometbft` and `try_from_cometbft_header_and_data`.
- implement `ConfigureSequencerBlock` test util as a replacement for generating blocks for tests

## Testing
existing tests

## Related Issues

closes #969

### DIFF
--- a/crates/astria-sequencer-relayer/tests/blackbox/helper.rs
+++ b/crates/astria-sequencer-relayer/tests/blackbox/helper.rs
@@ -230,6 +230,11 @@ impl TestSequencerRelayer {
     }
 
     pub fn mount_block_response<const RELAY_SELF: bool>(&mut self, height: u32) -> BlockGuard {
+        use astria_core::primitive::v1::RollupId;
+        use rand::Rng;
+
+        let mut rng = rand::thread_rng();
+
         let proposer = if RELAY_SELF {
             self.account
         } else {
@@ -239,11 +244,17 @@ impl TestSequencerRelayer {
         let (tx, rx) = oneshot::channel();
 
         let block = ConfigureSequencerBlock {
+            block_hash: Some(rng.gen::<[u8; 32]>()),
             height,
             proposer_address: Some(proposer),
+            sequence_data: vec![(
+                RollupId::from_unhashed_bytes(b"some_rollup_id"),
+                vec![99u8; 32],
+            )],
             ..Default::default()
         }
         .make();
+
         let mut blocks = self.sequencer_server_blocks.lock().unwrap();
         blocks.push_back((tx, block.into_raw()));
         BlockGuard {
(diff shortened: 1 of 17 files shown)


joroshiba (approved):

crates/astria-sequencer-relayer/tests/blackbox/helper.rs
```diff
@@ -231,19 +230,31 @@ impl TestSequencerRelayer {
     }
 
     pub fn mount_block_response<const RELAY_SELF: bool>(&mut self, height: u32) -> BlockGuard {
+        use astria_core::primitive::v1::RollupId;
+        use rand::Rng;
+
+        let mut rng = rand::thread_rng();
```
SuperFluffy, line 236: Tests should ideally not have random components because that can become very confusing when debugging. Can you just replace this `block_hash` with a non-random `[u8; 32]`?

crates/astria-sequencer/src/api_state_ext.rs
```diff
@@ -382,63 +382,26 @@ impl<T: StateWrite> StateWriteExt for T {}
...
-            next_validators_hash: Hash::default(),
-            proposer_address: account::Id::try_from([0u8; 20].to_vec()).unwrap(),
-            time: Time::now(),
-            validators_hash: Hash::default(),
-            version: Version {
-                app: 0,
-                block: 0,
-            },
-        };
+        let block_hash: [u8; 32] = rng.gen();
```
SuperFluffy, line 401: Same here, if `block_hash` doesn't matter, than just use a constant value.

SuperFluffy (approved):
Good change. Happy to finally get rid of these cometbft blocks.

If you touch these tests anyways, can you please remove the randomly generated block hashes? Ideally tests never contain any random components - not a blocker here (and so I don't request changes as that's no different from what we had before), but it's good practice.

crates/astria-sequencer/src/api_state_ext.rs
noot: it matters here, since we're storing by block hash as the key in the db, so we need a unique key

PR #1006, opened 2024-04-25
feat(conductor, relayer): brotli compress data blobs

joroshiba:
## Summary
Uses brotli level 5 to compress data posted to celestia in relayer, and decompresses from read in conductor.

## Background
Data size is largest component of DA costs, compression can reduce costs of data posting. Based on research, and testing brotli level 5 provides us fast enough compression with a good compression ratio for encoded protobuf data.

## Changes
- compression with brotli in relayer
- decompression in conductor

## Testing
updated blackbox tests, will test with smoke test.


crates/astria-sequencer-relayer/src/relayer/write/conversion.rs
```diff
@@ -92,3 +120,22 @@ pub(super) fn convert(block: SequencerBlock) -> eyre::Result<Converted> {
         },
     })
 }
+
+fn brotli_compressed_bytes(data: &[u8]) -> eyre::Result<Vec<u8>> {
```
SuperFluffy, line 124: Nit: as for the method used inside conductor using a verb is more idiomatic:
```suggestion
fn brotli_compress_bytes(data: &[u8]) -> eyre::Result<Vec<u8>> {
```

crates/astria-sequencer-relayer/src/relayer/write/conversion.rs
```diff
@@ -92,3 +120,22 @@ pub(super) fn convert(block: SequencerBlock) -> eyre::Result<Converted> {
...
+    use std::io::Write as _;
+    let compression_params = BrotliEncoderParams {
+        quality: 5,
+        size_hint: data.len(),
+        ..Default::default()
+    };
+    let mut output = Vec::new();
```
SuperFluffy, line 131: An easy perf improvement might be `Vec::with_capacity(data.len())` (as an upper limit) or (maybe better because deterministic) a multiple of the buffer size set in `CompressorWriter::with_params`: i.e. `Vec::with_capacity(BUF_SIZE * 8)` with `BUF_SIZE = 4096`.

`Vec` grows exponentially. In it would reallocate like `0 -> 4096 -> 8192 -> 16384` (note that `Vec::new()` is a no-op, so in this case doing `Vec::with_capacity(4096)` or not does the same amount of work).

crates/astria-sequencer-relayer/src/relayer/write/conversion.rs
```diff
@@ -92,3 +120,22 @@ pub(super) fn convert(block: SequencerBlock) -> eyre::Result<Converted> {
...
+    {
+        let mut compressor =
+            brotli::CompressorWriter::with_params(&mut output, 4096, &compression_params);
+        compressor
+            .write_all(data)
+            .wrap_err("failed compressing data")?;
```
SuperFluffy, line 137: You can probably just `compressor.write_all(data)?` and have the function return an `io::Result<Vec<u8>>` because the context msg here is not very useful.

SuperFluffy (approved):
This is a very nice improvement.

I only have very minor nits, none of them blocking.

crates/astria-sequencer-relayer/src/metrics_init.rs
```diff
@@ -96,3 +108,10 @@ pub const SEQUENCER_HEIGHT_FETCH_FAILURE_COUNT: &str = concat!(
     env!("CARGO_CRATE_NAME"),
     "_sequencer_height_fetch_failure_count",
 );
+
+pub const TOTAL_ASTRIA_BLOB_DATA_SIZE_FOR_BLOCK: &str =
+    concat!(env!("CARGO_CRATE_NAME"), "_total_astria_blob_data_size");
```
Fraser999, line 113: There's a little inconsistency around the name/description here.  In the `describe_gauge!` we have `_DATA_SIZE_FOR_BLOCK` and "for a Astria **blob**", but here we have just `_data_size`.

crates/astria-sequencer-relayer/src/relayer/write/conversion.rs
```diff
@@ -78,11 +86,31 @@ pub(super) fn convert(block: SequencerBlock) -> eyre::Result<Converted> {
...
+        let compressed_blob = brotli_compressed_bytes(&raw_blob)
+            .wrap_err_with(|| format!("failed compressing rollup `{rollup_id}`"))?;
+        total_data_compressed_size += compressed_blob.len();
+        let blob = Blob::new(namespace, compressed_blob)
             .wrap_err_with(|| format!("failed creating blob for rollup `{rollup_id}`"))?;
         blobs.push(blob);
         rollups.push(info);
     }
+
+    let compression_ratio = total_data_uncompressed_size / total_data_compressed_size;
```
Fraser999, line 100: Might be nice to just do the casts to f64 here since we need to cast the result anyway.  Would give a slightly more accurate value I guess, especially if compression is not too effective.
```suggestion
    let compression_ratio = total_data_uncompressed_size as f64 / total_data_compressed_size as f64;
```

Fraser999 (approved):
LGTM too - just a couple of minor points.

crates/astria-sequencer-relayer/src/relayer/write/conversion.rs
SuperFluffy: Good catch, I missed that completely.

WafflesVonMaple (approved):

crates/astria-core/src/brotli.rs
```diff
@@ -0,0 +1,53 @@
...
+/// Returns an error if the compression fails.
+pub fn compress_bytes(data: &[u8]) -> Result<Vec<u8>, std::io::Error> {
+    let compression_params = BrotliEncoderParams {
+        quality: 5,
+        size_hint: data.len(),
+        ..Default::default()
+    };
+    // Capacity based on expecting best potential compression ratio of 8x (based on benchmarks)
+    // only would need to resize 2 times to reach worst case.
+    let mut output = Vec::with_capacity(data.len() / 8);
```
SuperFluffy, line 45: I think being overly optimistic doesn't buy us anything. We will likely incur at least 1 reallocation. Let's go with `data.len() / 4`?

SuperFluffy (approved):
Still looks great (especially now with @Fraser999 having caught the missing cast-to-f64 before calculating the ratio).

I also think this does not need extra tests past the blackbox tests in conductor (unit tests would just test if `brotli` indeed compresses/decompresses).

I believe the compression buffer is chosen to optimistically. The 8x compression factor is likely never reached, so we'd incur at least 1 resize always. IMO going with `Vec::with_capacity(data.len() / 4)` is more realistic.

But either way, this is not a blocker.

PR #1009, opened 2024-04-25
fix(sequencer): stateful check now ensures balance for total tx

### LINKED ISSUE #786: sequencer: fix stateful check to check for sufficient fees in entire tx, not just individual actions
txs shouldn't get to execution if the sender does not have sufficient balance for the summation of the fees in the entire tx.

Lilyjjo:
## Summary
Sequencer's stateful check wasn't ensuring that a transaction's entire fee consumption was covered by the balance, it was only checking per individual action. Now the stateful check will err if there aren't enough funds for the whole transaction.

## Background
We were checking this already in the mempool, but this check will cover the edge case of an account's earlier transaction consuming funds in the same block. 

## Testing
Added unit test.

## Related Issues
closes #786

### DIFF
--- a/crates/astria-sequencer/src/app.rs
+++ b/crates/astria-sequencer/src/app.rs
@@ -1410,6 +1410,95 @@ mod test {
     }
 
     #[tokio::test]
+    async fn app_stateful_check_fails_insufficient_total_balance() {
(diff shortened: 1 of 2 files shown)


noot (approved):
looks good!

Issue #1003, opened 2024-04-24
sequencer: make fees configurable via a transaction


noot:
store fees in state so that they're configurable by a transaction. additionally implement a sudo action which allows for fees to be updated.

PR #1017, opened 2024-04-26
refactor(sequencer): store fees for actions in app state

noot:
## Summary
store fees for actions in app state, fees are of the form `base_fee + byte_cost_multiplier * byte_size`.

for txs that don't write to DA, they have an implicit `byte_size` of zero, so they only have a base fee.

for txs that write to DA, they have a `byte_cost_multiplier` which is multiplied by the number of bytes written to DA for fee calculation.

## Background
 this is so eventually we can change fees via a tx.

## Changes
- store fees for actions in app state

## Testing
unit tests

## Related Issues
part of  #1003


crates/astria-sequencer/src/transaction/mod.rs
```diff
@@ -95,11 +104,12 @@ pub(crate) async fn check_balance_for_total_fees<S: StateReadExt + 'static>(
                     .or_insert(act.amount);
                 fees_by_asset
                     .entry(act.fee_asset_id)
-                    .and_modify(|amt| *amt += TRANSFER_FEE)
-                    .or_insert(TRANSFER_FEE);
+                    .and_modify(|amt| *amt += transfer_fee)
```
SuperFluffy, line 107:
```suggestion
                    .and_modify(|amt| *amt = amt.saturating_add(transfer_fee))
```

crates/astria-sequencer/src/transaction/mod.rs
```diff
@@ -113,14 +123,14 @@ pub(crate) async fn check_balance_for_total_fees<S: StateReadExt + 'static>(
                     .or_insert(act.amount());
                 fees_by_asset
                     .entry(*act.fee_asset_id())
-                    .and_modify(|amt| *amt += ICS20_WITHDRAWAL_FEE)
-                    .or_insert(ICS20_WITHDRAWAL_FEE);
+                    .and_modify(|amt| *amt += ics20_withdrawal_fee)
```
SuperFluffy, line 126:
```suggestion
                    .and_modify(|amt| *amt = amt.saturating_add(ics20_withdrawal_fee))
```

crates/astria-sequencer/src/transaction/mod.rs
```diff
@@ -113,14 +123,14 @@ pub(crate) async fn check_balance_for_total_fees<S: StateReadExt + 'static>(
...
+                    .or_insert(ics20_withdrawal_fee);
             }
             Action::InitBridgeAccount(act) => {
                 fees_by_asset
                     .entry(act.fee_asset_id)
-                    .and_modify(|amt| *amt += INIT_BRIDGE_ACCOUNT_FEE)
-                    .or_insert(INIT_BRIDGE_ACCOUNT_FEE);
+                    .and_modify(|amt| *amt += init_bridge_account_fee)
```
SuperFluffy, line 132:
```suggestion
                    .and_modify(|amt| *amt = amt.saturating_add(init_bridge_account_fee))
```

crates/astria-sequencer/src/transaction/mod.rs
```diff
@@ -129,8 +139,8 @@ pub(crate) async fn check_balance_for_total_fees<S: StateReadExt + 'static>(
                     .or_insert(act.amount);
                 fees_by_asset
                     .entry(act.fee_asset_id)
-                    .and_modify(|amt| *amt += TRANSFER_FEE)
-                    .or_insert(TRANSFER_FEE);
+                    .and_modify(|amt| *amt += transfer_fee)
```
SuperFluffy, line 142:
```suggestion
                    .and_modify(|amt| *amt = amt.saturating_add(transfer_fee))
```

crates/astria-sequencer/src/sequence/state_ext.rs
SuperFluffy, line 1: Add simple unit tests to this file in line with the other unit tests that were added over the last few weeks?

crates/astria-sequencer/src/bridge/bridge_lock_action.rs
```diff
@@ -58,6 +66,22 @@ impl ActionHandler for BridgeLockAction {
...
+        let transfer_fee = state
+            .get_transfer_base_fee()
+            .await
+            .context("failed to get transfer base fee")?;
+
+        let byte_cost_multiplier = state
+            .get_bridge_lock_byte_cost_multiplier()
+            .await
+            .context("failed to get byte cost multiplier")?;
+        let fee = byte_cost_multiplier * DEPOSIT_BYTE_LEN + transfer_fee;
```
SuperFluffy, line 82: Saturating arithmetic:
```suggest
        let fee = byte_cost_multiplier.saturating_mul(DEPOSIT_BYTE_LEN).saturating_add(transfer_fee);
```

Also, please add a unit tests for this calculation.

crates/astria-sequencer/src/bridge/bridge_lock_action.rs
```diff
@@ -26,6 +32,8 @@ use crate::{
     transaction::action_handler::ActionHandler,
 };
 
+pub(crate) const DEPOSIT_BYTE_LEN: u128 = std::mem::size_of::<Deposit>() as u128;
```
SuperFluffy, line 35: Don't use the Rust in-memory representation of types if you want deterministic results. There are no guarantees on ordering of fields or alignment paddding. You can get different results depending on Rust compiler or even architecture.

You can use `#[repr(C)]` to get a defined layout (and hence deterministic size), but even that is tricky if somebody changes the representation of one of `Deposit`'s fields.

For `Deposit` in partciular I notice that it contains `destination_chain_address: String`, the memory size of which is is 24 bytes on a 64bit host (64 bit pointer, 64 bit capacity and length), with the actual memory used being unbounded.

crates/astria-sequencer/src/bridge/bridge_lock_action.rs
```diff
@@ -78,6 +102,19 @@ impl ActionHandler for BridgeLockAction {
             .await
             .context("failed to execute bridge lock action as transfer action")?;
 
+        // the transfer fee is already deducted in `transfer_action.execute()`,
+        // so we just deduct the bridge lock byte multiplier fee.
+        let byte_cost_multiplier = state
+            .get_bridge_lock_byte_cost_multiplier()
+            .await
+            .context("failed to get byte cost multiplier")?;
+        let fee = byte_cost_multiplier * DEPOSIT_BYTE_LEN;
```
SuperFluffy, line 111:
```suggestion
        let fee = byte_cost_multiplier.saturating_mul(DEPOSIT_BYTE_LEN);
```
And unit test please.

crates/astria-sequencer/src/bridge/bridge_lock_action.rs
```diff
@@ -78,6 +102,19 @@ impl ActionHandler for BridgeLockAction {
...
+
+        state
+            .decrease_balance(from, self.fee_asset_id, fee)
+            .await
+            .context("failed to deduct fee from account balance")?;
+
         let rollup_id = state
             .get_bridge_account_rollup_id(&self.to)
             .await?
```
SuperFluffy, line 120: Just noticed the bare question mark (outside of the PR, but can you context please?)

SuperFluffy (changes requested):
Requesting changes due to the use of `std::mem::size_of`. The rest looks fine to me, but I would like have saturating arithmetic instead of bare `+` and `*`. Also, unit tests for  the following would be great:

1. all fee calculations
2. the byte length of `Deposit`
3. writing/reading from rocksdb


Orthogonal to this PR I have another concern:

We have no mechanism that ensure that the keys in hardcoded in the various `state_ext` modules/`StateExt` immplementations don't clash. I think it might make sense to list all of them in a central locations in a first step (`astria_sequencer::database_keys` maybe?). In a second step I want to think about how we can ensure at compile time that they don't clash.

crates/astria-sequencer/src/bridge/bridge_lock_action.rs
noot: good point on the `destination_chain_address`, i'll change this to calculate the byte size at runtime (similar to sequence actions)

crates/astria-sequencer/src/app.rs
```diff
@@ -1289,12 +1310,13 @@ mod test {
...
             value + 10u128.pow(19)
         );
+        let transfer_fee = app.state.get_transfer_base_fee().await.unwrap();
         assert_eq!(
             app.state
                 .get_account_balance(alice_address, native_asset)
                 .await
                 .unwrap(),
-            10u128.pow(19) - (value + TRANSFER_FEE),
+            10u128.pow(19) - (value + transfer_fee),
```
SuperFluffy, line 1319: We should really move this `tests` module into `src/app/tests.rs`. I wanted to comment several times now how I want saturating arithmetic. :-)

crates/astria-sequencer/src/bridge/init_bridge_account_action.rs
```diff
@@ -41,6 +38,8 @@ impl ActionHandler for InitBridgeAccountAction {
             "invalid fee asset",
         );
 
+        let fee = state.get_init_bridge_account_base_fee().await?;
```
SuperFluffy, line 41: context?

SuperFluffy (approved):
Looking good. Just once place I found that should get error context.

PR #1018, opened 2024-04-26
Relayer secret provider

aajimal:
## Summary
Adds support for using a secret provider to the sequencer-relayer chart

## Background
These changes are being made to make this component consistent with others that have already implemented this.

## Changes
- Add `SecretProviderClass`
- Update the `Values.config.celestiaAppPrivateKey` to be in the correct format
- Update ConfigMaps to reference the updated secret
- Update StatefulSet to optionally pull from secretProvider or devContent value

## Testing
- Ensure `helm template` actually renders without error
- Deploy to local dev-cluster with `Values.secretProvider.enabled=false`
- Will need to still coordinate + deploy to dev with `Values.secretProvider.enabled=true` assuming the secrets have been created

### DIFF
--- a/charts/sequencer-relayer/values.yaml
+++ b/charts/sequencer-relayer/values.yaml
@@ -42,7 +42,6 @@ config:
     secret:
       filename: "key.hex"
       resourceName: "projects/$PROJECT_ID/secrets/celestiaPrivateKey/versions/latest"
-      key: token
 ports:
   relayerRPC: 2450
   metrics: 9000
--- a/charts/sequencer-relayer/values.yaml
+++ b/charts/sequencer-relayer/values.yaml
@@ -42,6 +42,7 @@ config:
     secret: 
       filename: "key.hex"
       resourceName: "projects/$PROJECT_ID/secrets/celestiaPrivateKey/versions/latest"
+      key: token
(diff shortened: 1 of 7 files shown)


charts/sequencer-relayer/values.yaml
```diff
@@ -59,3 +61,27 @@ storage:
       size: "1Gi"
       persistentVolumeName: "sequencer-relayer-storage"
       path: "/data/sequencer-relayer-data"
+
+# When deploying in a production environment should use a secret provider
+# This is configured for use with GCP, need to set own resource names
+# and keys
+secretProvider:
+  enabled: true
```
joroshiba, line 69: Torn on enabling by default here, more technically correct for prod but need to have secret provider setup to use chart at all.

I think elsewhere we have disabled by default, stick to that for now?

joroshiba (approved):
approved with one thing we should change I think plus lint. you can merge first and then I'll update my pr on top of this one.

charts/sequencer-relayer/values.yaml
aajimal: Good call. Thanks for catching that. It's disabled now.

Issue #1011, opened 2024-04-25
conductor tries re-executing firm blocks if block number not cached


SuperFluffy:
As flagged in https://github.com/astriaorg/astria/pull/980#discussion_r1576846333, conductor tries executing a firm block if it cannot be found in its map of pending blocks. This is legacy code for making it run in various modes (soft and soft-and-firm). But when running in `soft-and-firm` mode is a hard error.

This error *should* not happen, especially with the work done in #980, but the logic should be rewritten nonetheless to guard against this edge case.

PR #1021, opened 2024-04-28
fix(conductor): only execute firm blocks if firm and soft block numbers match

SuperFluffy:
## Summary
Ensures that conductor only executes firm blocks if running in soft-and-firm mode, and if the rollup soft and firm block numbers are equal.

## Background
Previously, Conductor used the presence of a block in its cache [1] as the condition for whether it should execute a firm block against a rollup. This should never be observed in practice, but there is a better (less confusing, more maintainable way) of expressing the same condition:

When running in soft-and-firm mode, Conductor:
1. must only execute a block if the rollup's `firm number == soft number`, and then use the rollup response to advance the firm commitment state.
2. must otherwise only advance (without prior execution) the rollup's firm commitment state using the block pending finalization for the corresponding block number stored in its cache.

1: the cache of blocks pending finalization; only relevant in soft-and-firm mode.

## Changes
- Tightened the condition for when Conductor executes a firm bock against the rollup when running in soft-and-firm mode (only execute if soft number equals firm number)
- Makes the absence of a block in its cache a hard error when not executing but only advancing the commitment state
- Tries to fetch the missing block pending finalization from the rollup as a last ditch effort before bailing

## Testing
Added tests for determining if a firm block should be executed. Because the case where Conductor tries refetching the missing block is a fallback for a faulty edge case that should never be observed in practice this cannot be tested using blackbox tests.

## Related Issues

closes https://github.com/astriaorg/astria/issues/1011

### DIFF
--- a/crates/astria-conductor/src/conductor.rs
+++ b/crates/astria-conductor/src/conductor.rs
@@ -64,8 +64,7 @@ impl Conductor {
         // Spawn the executor task.
         let executor_handle = {
             let (executor, handle) = executor::Builder {
-                with_firm: !cfg.execution_commit_level.is_soft_only(),
-                with_soft: !cfg.execution_commit_level.is_firm_only(),
+                mode: cfg.execution_commit_level,
                 rollup_address: cfg.execution_rpc_url,
                 shutdown: shutdown.clone(),
             }
@@ -76,7 +75,7 @@ impl Conductor {
             handle
         };
 
-        if !cfg.execution_commit_level.is_firm_only() {
+        if cfg.execution_commit_level.is_with_soft() {
             let sequencer_grpc_client =
                 sequencer::SequencerGrpcClient::new(&cfg.sequencer_grpc_url)
                     .wrap_err("failed constructing grpc client for Sequencer")?;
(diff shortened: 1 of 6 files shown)


crates/astria-conductor/src/conductor.rs
```diff
@@ -76,7 +75,7 @@ impl Conductor {
             handle
         };
 
-        if !cfg.execution_commit_level.is_firm_only() {
+        if cfg.execution_commit_level.is_with_soft() {
```
SuperFluffy, line 78: Inverted the logic here. This was confusing.

crates/astria-conductor/src/conductor.rs
```diff
@@ -95,7 +94,7 @@ impl Conductor {
             tasks.spawn(Self::SEQUENCER, sequencer_reader.run_until_stopped());
         }
 
-        if !cfg.execution_commit_level.is_soft_only() {
+        if cfg.execution_commit_level.is_with_firm() {
```
SuperFluffy, line 97: Inverted the logic here too. Confusing.

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -447,29 +460,49 @@ impl Executor {
...
+            // cache - either due to being pre-populated at startup (via
+            // `populate_blocks_pending_finalization`), or during normal operation (as part of
+            // `execute_soft`).
+            // This code is here as a fall-back mechanism in case there is a bug.
+            error!(
+                block_number,
+                "pending block not found for block number in cache. THIS SHOULD NOT HAPPEN. \
+                 Trying to fetch the already-executed block from the rollup before giving up."
+            );
+            match client.clone().get_block(block_number).await {
```
SuperFluffy, line 489: @joroshiba Flagging this new logic here: we could forego the entire logic introduced in https://github.com/astriaorg/astria/pull/980 and just try and fetch the missing blocks (those pending finalization) when we actually need them.

Would make startup cheaper at the cost of extra network requests.

PR #1025, opened 2024-04-30
fix(sequencer): set current app hash properly when creating app

noot:
## Summary
set current app hash properly when creating app

## Background
this was causing issues if the node was restarted as the app hash would be empty, and this is written to state in ibc begin_block, causing an app hash mismatch on the next synced block

## Changes
- set current app hash properly when creating app

## Testing
local testnet with 4 validators and restarting one node, previously failed, now it works

### DIFF
--- a/crates/astria-sequencer/src/app.rs
+++ b/crates/astria-sequencer/src/app.rs
@@ -146,21 +146,30 @@ pub(crate) struct App {
 }
 
 impl App {
-    pub(crate) fn new(snapshot: Snapshot) -> Self {
+    pub(crate) async fn new(snapshot: Snapshot) -> anyhow::Result<Self> {
         tracing::debug!("initializing App instance");
+
+        let app_hash: AppHash = snapshot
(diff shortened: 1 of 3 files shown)


crates/astria-sequencer/src/app.rs
```diff
@@ -146,21 +146,30 @@ pub(crate) struct App {
 }
 
 impl App {
-    pub(crate) fn new(snapshot: Snapshot) -> Self {
+    pub(crate) async fn new(snapshot: Snapshot) -> anyhow::Result<Self> {
         tracing::debug!("initializing App instance");
 
+        let app_hash: AppHash = snapshot
```
joroshiba, line 152: will this fail if init_chain has not been called yet?

noot: no, it'll just return the empty root hash. confirmed that syncing from genesis with an empty state works

joroshiba (approved):

PR #1026, opened 2024-04-30
tests(sequencer): improvements to testnet script, make script posix compliant

### LINKED ISSUE #168: Multi sequencer testnet script is not posix compliant
The test script to create multiple sequencer nodes uses `&>` redirection which is a bash-ism and not posix compliant. Therefore there is no guarantee that it will work on systems that don't symlink sh to bash (or use default shells that don't understand `&>`).

Currently here: https://github.com/astriaorg/astria/blob/main/crates/astria-sequencer/scripts/testnet.sh

noot:
## Summary
as title says - improvements for the testnet script

## Changes
- make script posix compliant so output is redirected properly
- store sequencer app pids in a file for easy killing

## Testing
ran it locally on a linux machine

## Related Issues

closes #168


crates/astria-sequencer/justfile
```diff
@@ -28,18 +28,18 @@ run-testnet:
 
 [linux]
 stop-testnet:
-  pkill cometbft && pkill astria-sequencer
+  pkill cometbft && pkill -f astria-sequencer
```
Fraser999, line 31: The `-f` causes `pkill` to search the full command line rather than just the process name, which ends up killing not only any running `astria-sequencer` but also the shell which `just` spawns to run the command.  The full command which `just` executes is
```sh
sh -cu pkill cometbft && pkill -f astria-sequencer
```
So `pkill` kills that shell, causing the `just` execution to fail.

We can specify `pkill -f ^astria-sequencer` to ensure it's matching only on the start of the command.

Also, we can probably ignore the results of the `pkill` calls, meaning we can then make `stop-testnet` a dependency of `clean-testnet` again?

```suggestion
  pkill cometbft || true
  pkill -f ^astria-sequencer || true
```

If the `[macos]` variant of `pkill` supports `-f` in the same way, we could maybe deduplicate these sections?

crates/astria-sequencer/justfile
```diff
@@ -28,18 +28,18 @@ run-testnet:
...
 [linux] 
 run-testnet: 
   cargo build
-  ${OUT_DIR} ${NUM_VALIDATORS} ./scripts/testnet.sh
+  ./scripts/testnet.sh
```
Fraser999, line 35: With this change, the `[macos]` and `[linux]` versions are identical, so could probably be deduplicated?

crates/astria-sequencer/justfile
```diff
@@ -28,18 +28,18 @@ run-testnet:
...
 
 cometbft-logs-testnet NODE="node0":
   tail -f sequencer_testnet/{{ NODE }}/cometbft.log
 
 sequencer-logs-testnet NODE="node0":
   tail -f sequencer_testnet/{{ NODE }}/sequencer.log
 
-clean-testnet DIR="sequencer_testnet": stop-testnet
+clean-testnet DIR="sequencer_testnet":
```
Fraser999, line 43: As mentioned above, if we ignore the results of `pkill`, we could reinstate this, since it seems like we wouldn't want to clean the dirs if the testnet is still running?

crates/astria-sequencer/justfile
noot: good point, i was wondering why the justfile  was always failing lmao. `pkill -f ^astria-sequencer` didn't end up killing the sequencer processes for me as it wasn't an exact match. i changed this to be `pkill -f [a]stria-sequencer` which still regex-matches processes with `astria-sequencer` but not the shell as it no longer contains `astria-sequencer` :D

crates/astria-sequencer/justfile
noot: changed so the linux version is different

crates/astria-sequencer/justfile
noot: re-added it!

Fraser999 (approved):

SuperFluffy (approved):
Following @Fraser999 approval to unblock merging (codeowner/reviewer requirements)

PR #1029, opened 2024-04-30
wip: feat: Add parameterizable load tests to justfile

bharath-123:
## Summary
Add a load-test using astria's fork of [goomy-blobs](https://github.com/astriaorg/goomy-blob). This is akin to a smoke test.

## Background
Currently, we have a smoke test which just does an EOA transfer to check if the entire system works. We extend the tests to include a parameterizable load test where users can configure the number of wallets to spin up, the throughput of the txs and the number of txs to send in parallel.

We also run 3 types of load-test against the EVM. We run:

1. EOA txs
1. ERC20 txs
1. Large txs i.e generate txs where each one fills up the block easily,

To run the load test, we need to run:
`just deploy-load-test` to get the goomy-blob binary.
and
`just run-load-test` to run it. We can tweak the parameters in the justfile

## Changes
* Add the `deploy-load-test` cmd which deploys a load test by just installing the binary
* Add the `run-load-test` cmd which runs the load-test

TODO: Instead of downloading a binary, it ll be easier to just dockerize this and run the command with the docker image.


SuperFluffy:
I think big parts of this ended up in https://github.com/astriaorg/spamooor (including logic to hit sequencer). Closing, but feel free to reopen.

PR #1033, opened 2024-05-01
test(sequencer-relayer): reinstate black box tests

### LINKED ISSUE #1008: Restore black box tests removed from sequencer-relayer
We need to restore the black box tests removed as part of #963.  Ensure we also restore the single test added in [f236c79](https://github.com/astriaorg/astria/pull/1001/commits/f236c799e5603e25aa663c8b16468e89ff940d54#diff-d101a60a06ba8d0a04e99781aff64eaa686effd984628d5d5a30cdf0f5fdd501) and subsequently deleted when merging `main` into #1001.

They should be re-implemented using the new grpc-mock crate.

Fraser999:
## Summary
The black box tests were removed in a recent PR.  This PR reinstates them using the new mock gRPC framework `astria-grpc-mock`.

## Background
The tests would have needed heavy refactoring in #963, and we wanted to avoid adding a lot more code to that already-large PR.  We decided to temporarily delete them and reinstate them using `astria-grpc-mock` to mock responses from the Celestia app.

## Changes
The tests removed in #963 and a single test briefly added in #1001 then removed again have been restored.  I also added a new test to check the relayer shuts down in a timely manner.

Previously the tests leaned heavily on counts of blobs received by the mock Celestia app.  The current tests retain these checks, but also query the `/status` endpoint of the sequencer-relayer to confirm its state.  There was also a single test previously which checked the state of the postsubmit.json file.  I didn't think this was necessary given that we're querying the state in a more "black box" way now (via the http server), but I didn't remove the function to perform this check (`TestSequencerRelayer::assert_state_files_are_as_expected`) pending a decision on whether to reinstate that check or not inside the `later_height_in_state_leads_to_expected_relay` test.

As @SuperFluffy predicted, all the new protobuf packages added in #963 had to be added to the `pbjson_build` builder to generate the serde impls required for using them in `astria-grpc-mock`.  This accounts for the vast majority of the new LoC in this PR.

I made a few small changes to the mock framework:
- added `Mock::up_to_n_times` to avoid failures due to a single-use mock response being sent multiple times
- added `DynamicResponse` to support constructing mock responses which can be passed the relevant gRPC request
- changed `MockGuard::wait_until_satisfied` to take `self` by ref rather than value, since if it consumes self and `wait_until_satisfied` times out, the `MockGuard` panics in its `Drop` impl, meaning we don't get a good indication from e.g. a `tokio::time::timeout` as to what went wrong

Most of the new functionality lives in `MockCelestiaAppServer` and `TestSequencerRelayer`.  For the former, all gRPCs except for `BroadcastTx` and `GetTx` are set up to always return valid-enough responses - i.e. these don't need to be mounted individually in every test case.  In the case of `MockCelestiaAppServer`, the new functionality is a combination of support for mounting mock responses and functions with timeouts to query the `/status`, `/healthz` and `/readyz` http endpoints of the sequencer-relayer.

## Testing
These changes are tests.

Closes #1008.


crates/astria-sequencer-relayer/src/relayer/read.rs
```diff
@@ -233,11 +233,12 @@ async fn fetch_block(
     .await
     .wrap_err("retry attempts exhausted; bailing")?;
 
-    state.set_sequencer_connected(true);
-
-    let block = SequencerBlock::try_from_raw(block.into_inner())
-        .wrap_err("failed to parse raw proto block from grpc response")?;
+    let maybe_block = SequencerBlock::try_from_raw(block.into_inner())
+        .wrap_err("failed to parse raw proto block from grpc response");
+    state.set_sequencer_connected(maybe_block.is_ok());
```
Fraser999, line 238: This is the only change to production code, so worth highlighting :)

I think this is the correct behaviour, since if we return an error here, we exit the `select!` loop in `Relayer::run` and are hung until externally killed.  While in that state, I think it makes sense to have the sequencer noted as disconnected so that the `/healthz` endpoint reports degraded health.

crates/astria-sequencer-relayer/src/relayer/read.rs
```diff
@@ -233,11 +233,12 @@ async fn fetch_block(
...
 
+    let block = maybe_block?;
+    debug_assert_eq!(height, block.height());
```
Fraser999, line 241: Discussed this with @joroshiba - it's maybe overkill, since it's probably reasonable to expect that responses from the sequencer don't need to be checked.  (I stumbled on this when sending a mocked response with height 6 to a request for height 1, so the `debug_assert` might help keep the mocks slightly more realistic)

crates/astria-sequencer-relayer/tests/blackbox/main.rs
```diff
@@ -1,90 +1,432 @@
...
+        .await;
+    sequencer_relayer
+        .wait_for_latest_fetched_sequencer_height(1, 1_000)
+        .await;
+    sequencer_relayer
+        .wait_for_latest_observed_sequencer_height(1, 1_000)
+        .await;
+
+    assert_eq!(
+        sequencer_relayer.celestia_app_received_blob_count(),
```
SuperFluffy, line 64: This is very close to the original implementation (the celestia mock server using channels etc), but I think it might be nicer if we expressed this in terms of a mock (or series of mocks) together with `wait_until_satisfied`.

crates/astria-sequencer-relayer/tests/blackbox/helpers/test_sequencer_relayer.rs
```diff
@@ -0,0 +1,742 @@
...
+    }
+
+    /// Executes `future` within the specified duration, returning its result.
+    ///
+    /// If execution takes more than 80% of the allowed time, an error is logged before returning.
+    ///
+    /// # Panics
+    ///
+    /// Panics if execution takes longer than the specified duration.
+    pub async fn timeout_ms<F: Future>(
```
SuperFluffy, line 480: So the idea behind this is to grab the `/status` in case of timeouts and report that?

crates/astria-grpc-mock/src/mock.rs
```diff
@@ -45,6 +46,13 @@ impl Mock {
         }
     }
 
+    #[must_use = "a mock must be mounted on a server to be useful"]
+    pub fn up_to_n_times(mut self, n: u64) -> Self {
```
SuperFluffy, line 50: This is the original wiremock functionality that we hadn't yet ported over right?

crates/astria-grpc-mock/src/mock_server.rs
```diff
@@ -110,7 +110,7 @@ pub struct MockGuard {
 }
 
 impl MockGuard {
-    pub async fn wait_until_satisfied(self) {
+    pub async fn wait_until_satisfied(&self) {
```
SuperFluffy, line 113: What makes it necessary to take this by reference? The timeout?

I actually thought of it as a feature that it gets dropped at the end of the scope, immediately returning the panic/report. But I suppose that wasn't useful and/or blocking some other functionality?

crates/astria-grpc-mock/src/response.rs
```diff
@@ -54,6 +56,52 @@ impl Respond for DefaultResponse {
     }
 }
 
+pub fn dynamic_response<I, O, F>(responder: F) -> DynamicResponse<I, O, F>
```
SuperFluffy, line 59: I wonder if we can implement e.g.:
```rust
impl<I, O, F> Respond for F,
where
    F: Fn(&I) -> O,
    O: erased_serde::Serialize + prost::Name + Clone + 'static,
```
Maybe another indirection would be necessary like `MakeDynamicResponse` and then implemeent `Respond for MakeDynamicResponse`. Gonna think about how to do that.

crates/astria-sequencer-relayer/src/relayer/read.rs
SuperFluffy: Are we hanging until externally killed or do we just shut down gracefully?

crates/astria-sequencer-relayer/src/relayer/read.rs
SuperFluffy: The problem is of course that we are running these tests without `debug`. But I guess we can easily change this.

However, I actually think it would be fine to add an `ensure!(height == block.height(), "requested block at height `{height}` but received a block at height `{}`", block.height());` here and just stop hard if that's violated.

conductor is doing something similar now with the responses it gets from the rollup implemeting `astria.executionapi`

SuperFluffy (approved):
It's very nice to have these back. I also the additions like the custom `timeout_ms` that reports `/status` in case of failures (the contents of which we should probably expand on :-) ).

I believe some mocks that currently return guards should not do so (I am thinking of the ABCI info one and also the sequencer-block fetch; I might have missed one). In my mind, "intermittent" mocks should just be mounted without any extra requirements when and in what order they are picked up by relayer - only the final result (blocks posted to Celestia) should actually be awaited. This would leave the order and timing "inside" relayer opaque.

Also, I think the Celestia-App checks (e.g. how many blocks were received) could be replaced by mocks.

But I also realize that both requirements are not necessarily trivial and might be beyond the scope of this PR, which stays very close to the original hand-written mock. I think these would be good to cover in future PRs.

I think with this PR you can also remove `astria-celestia-client` and `astria-celestia-mock`?

crates/astria-sequencer-relayer/src/relayer/read.rs
Fraser999: Hanging [here](https://github.com/astriaorg/astria/blob/73852b9476889a989ab45258003d1270cc42d90a/crates/astria-sequencer-relayer/src/relayer/mod.rs#L225).  We could do `self.shutdown_token.cancel();` just before that if we do want to actually shut down?

crates/astria-sequencer-relayer/src/relayer/read.rs
Fraser999: Sure - will do.

tools/protobuf-compiler/src/main.rs
```diff
@@ -112,6 +112,19 @@ fn main() {
             ".astria.execution.v1alpha2",
             ".astria.primitive.v1",
             ".astria.sequencerblock.v1alpha1",
+            ".celestia.blob.v1",
```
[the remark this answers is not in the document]
Fraser999: The only thing stopping me from doing that was I didn't realise I could do that :D
I'll make that change.

crates/astria-sequencer-relayer/src/relayer/read.rs
SuperFluffy: I see - yeah. Good catch. We should do that. It's outside the scope of this PR I'd say, but `relayer` knows that if the read task fails no progress will be made. So it should just shut everything down.

crates/astria-sequencer-relayer/tests/blackbox/main.rs
```diff
@@ -1,90 +1,432 @@
...
+            2_000,
+            "waiting for get tx guard",
+            get_tx_guard.wait_until_satisfied(),
+        )
+        .await;
+
+    // Mount a bad block next, so the relayer will fail to fetch the block.
+    let _guard = sequencer_relayer.mount_abci_response(2).await;
+    let block_to_mount = SequencerBlockToMount::BadAtHeight(2);
+    let block_guard = sequencer_relayer
```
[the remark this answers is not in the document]
Fraser999: If we did the same for the abci one (i.e. not return a guard for it), I think that should work.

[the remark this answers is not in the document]
Fraser999: Actually, I take that back :)  Without the scoped guards being awaited, the test exits before the mocks have registered that they've been consumed, so they fail the test.

I think I can get away with making the abci one not scoped, and have two methods for the sequencer block - one scoped and the other not.

crates/astria-sequencer-relayer/tests/blackbox/helpers/mock_sequencer_server.rs
```diff
@@ -0,0 +1,153 @@
...
+
+        Mock::for_rpc_given(
+            GET_SEQUENCER_BLOCK_GRPC_NAME,
+            message_type::<GetSequencerBlockRequest>(),
+        )
+        .respond_with(constant_response(block))
+        .up_to_n_times(1)
+        .expect(1)
+        .with_name(debug_name)
+        .mount_as_scoped(&self.mock_server)
```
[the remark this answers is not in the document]
Fraser999: As commented above, I'll provide `mount_sequencer_block_response` which doesn't return a guard (used in almost every test) and `mount_sequencer_block_response_as_scoped` which does (used in the healthz test)

crates/astria-sequencer-relayer/tests/blackbox/main.rs
Fraser999: Agreed, but as mentioned above, I'll leave for a follow-up PR.

crates/astria-sequencer-relayer/tests/blackbox/helpers/test_sequencer_relayer.rs
Fraser999: Exactly.  I found it pretty useful when writing/debugging the tests.  I also wanted to have a way of seeing if I was close to bumping into any of the timeouts, hence the error message if we consume >80% of the timeout duration.

crates/astria-grpc-mock/src/mock.rs
Fraser999: Yup, indeed.

crates/astria-grpc-mock/src/mock_server.rs
Fraser999: If we had e.g.
```rust
tokio::time::timeout(duration, guard.wait_until_satisfied()).expect("shouldn't timeout");
```
then what was happening if `wait_until_satisfied` exceeded the duration was:
- `tokio::time::timeout` expired and cancelled the `wait_until_satisfied` future
- `wait_until_satisfied` had consumed the `MockGuard`, so when it was cancelled, the guard was dropped
- the `Drop` impl of the guard panicked due to unsatisfied condition
- we never get as far as executing `.expect("shouldn't timeout")` since we've panicked

I found this unexpected initially - I was expecting to see "shouldn't timeout" in the panic message, and then the report about the unsatisfied mock.  That what happens now by taking `&self` in `wait_until_satisfied`.

crates/astria-grpc-mock/src/response.rs
Fraser999: Yeah - that was my initial approach, but `I` and `O` are unconstrained type parameters so it doesn't fly :)

I played with this for a while but the explicit `dynamic_response` function seemed like the path of least resistance.  Maybe there's a cleaner way though?

crates/astria-sequencer-relayer/src/relayer/read.rs
Fraser999: Done in [d198996](https://github.com/astriaorg/astria/pull/1033/commits/d198996c67a81f2b6379f3c9f52d42a5ee96ba74).

tools/protobuf-compiler/src/main.rs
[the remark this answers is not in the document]
Fraser999: Done in [78fb6df](https://github.com/astriaorg/astria/pull/1033/commits/78fb6df0226d43bb9b4ba955b6b17677425bbae2).

crates/astria-sequencer-relayer/tests/blackbox/helpers/mock_sequencer_server.rs
```diff
@@ -0,0 +1,153 @@
...
+                ..Default::default()
+            }
+            .make(),
+            SequencerBlockToMount::Block(block) => block,
+        };
+
+        let mut block = block.into_raw();
+        if should_corrupt {
+            let header = block.header.as_mut().unwrap();
+            header.data_hash = [0; 32].to_vec();
```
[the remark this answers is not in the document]
Fraser999: Done in [c2aac68](https://github.com/astriaorg/astria/pull/1033/commits/c2aac68b17a141c4a214a904b307e1c904a98085).

crates/astria-sequencer-relayer/tests/blackbox/main.rs
```diff
@@ -1,90 +1,432 @@
...
+        last_written_sequencer_height: None,
+        only_include_rollups: HashSet::new(),
+    }
+    .spawn_relayer()
+    .await;
+
+    let _guard = sequencer_relayer.mount_abci_response(1).await;
+    let block_to_mount = SequencerBlockToMount::GoodAtHeight(1);
+    let _guard = sequencer_relayer
+        .mount_sequencer_block_response::<RELAY_ALL>(block_to_mount, "good block 1")
```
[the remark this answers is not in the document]
Fraser999: Done in [d64481e](https://github.com/astriaorg/astria/pull/1033/commits/d64481e2f743c2c929c324a86250213575922bb1).

crates/astria-sequencer-relayer/tests/blackbox/main.rs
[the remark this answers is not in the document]
Fraser999: Done in [d64481e](https://github.com/astriaorg/astria/pull/1033/commits/d64481e2f743c2c929c324a86250213575922bb1).

crates/astria-sequencer-relayer/tests/blackbox/helpers/mock_sequencer_server.rs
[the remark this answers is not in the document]
Fraser999: Done in [d64481e](https://github.com/astriaorg/astria/pull/1033/commits/d64481e2f743c2c929c324a86250213575922bb1).

Fraser999:
@SuperFluffy 
> I think with this PR you can also remove `astria-celestia-client` and `astria-celestia-mock`?

Done in [c9a5841](https://github.com/astriaorg/astria/pull/1033/commits/c9a584115841512e40817031aa35e0ec95631ec7).

PR #1034, opened 2024-05-02
feat(sequencer)!: implement bridge unlock action and derestrict transfers  

### LINKED ISSUE #983: update bridge accounts to allow incoming `Transfer`s
right now, bridge accounts only allow `BridgeLockAction`s to send funds into them. however, this is bad in the case that the asset accepted by the bridge account is not a sequencer fee payment token, in which case the bridge account will not have funds to pay withdrawal fees. we should update bridge accounts to allow incoming `Transfer`s as well, however a `Transfer` to a bridge account is a normal transfer that does not result in a `Deposit` event.

Lilyjjo:
## Summary
Implements bridge accounts' `BridgeUnlock` action and allows bridge accounts to transfer assets like normal.

## Background

We need a `BridgeUnlock` action to help with tracking/proving that withdraws for rollup specific events happened. This PR implements the ground work for that by adding a new action with a `memo:bytes` field that is currently unused. 

We also decided that it's unnecessary to have restrictions on the bridge account in terms of normal transfers in/out of the account. The bridge account is already a privileged actor and having the restrictions on transfer would make normal account maintenance (like gas funding) messy. 

## Changes
- implement `BridgeUnlockAction` for transferring out a bridge's bridged asset 
- derestricted the transfer function to let bridge account also act like normal accounts

## Testing
unit tests

## Breaking Changelist
- one new action is added: `BridgeUnlock`
- bridge account now can transfer assets in/out like normal sequencer accounts

## Related Issues
closes #983


Fraser999 (commented):
Just a few minor comments/questions.

crates/astria-sequencer/src/accounts/action.rs
```diff
@@ -96,6 +96,24 @@ impl ActionHandler for TransferAction {
                 .is_none(),
             "cannot send transfer to bridge account",
         );
+        // ensure that if the sender is a bridge account, that the asset being transferred is not
+        // the same asset as the bridge account's asset. The bridge account should be using
+        // the `BridgeUnlockAction` for that.
```
noot, line 101: i'm not sure if we need this logic - i think we should either not allow `Transfer`s out of bridge accounts at all, or allow them but not place any restrictions on them. this seems unnecessary, as if the bridge account owner wanted to transfer out the same asset but was restricted from using `Transfer`, they could still just use `BridgeUnlock` and leave the memo empty (or put something random)

I think we should remove this part for now and have a bigger discussion on what actions/accounting should look like

crates/astria-sequencer/src/bridge/bridge_unlock_action.rs
```diff
@@ -0,0 +1,198 @@
...
+        // ensure the from is a bridge account.
+        state
+            .get_bridge_account_rollup_id(&from)
+            .await
+            .context("failed to get bridge account rollup id")?
+            .ok_or_else(|| anyhow::anyhow!("bridge unlock must be sent from a bridge account"))?;
+
+        // ensure the bridge account's tracked asset is being transferred.
+        let allowed_asset_id = state
+            .get_bridge_account_asset_ids(&from)
```
noot, line 48: since this returns just 1 asset id now, can you rename the fn to `get_bridge_account_asset_id`?

proto/protocolapis/astria/protocol/transactions/v1alpha1/types.proto
```diff
@@ -181,3 +182,21 @@ message BridgeLockAction {
...
+//
+// It's the same as a `TransferAction` but with the added
+// `memo` field.
+message BridgeUnlockAction {
+  // the address of the bridge account to transfer to
+  astria.primitive.v1.Address to = 1;
+  // the amount to transfer
+  astria.primitive.v1.Uint128 amount = 2;
+  // the asset to be transferred
+  bytes asset_id = 3;
```
noot, line 197: don't think we need this field - since bridge accounts only allow one asset to be bridged, we know using `BridgeUnlockAction` must also withdraw that same asset

proto/protocolapis/astria/protocol/transactions/v1alpha1/types.proto
```diff
@@ -181,3 +182,21 @@ message BridgeLockAction {
...
   string destination_chain_address = 5;
 }
+
+// `BridgeUnlockAction` represents a transaction that transfers
+// funds from a bridge account to a sequencer account.
+//
+// It's the same as a `TransferAction` but with the added
+// `memo` field.
+message BridgeUnlockAction {
+  // the address of the bridge account to transfer to
```
noot, line 192:
```suggestion
  // the address to withdraw funds to
```

crates/astria-sequencer/src/bridge/bridge_unlock_action.rs
```diff
@@ -0,0 +1,198 @@
...
+
+        // this performs the same checks as a normal `TransferAction`,
+        // but without the check that prevents transferring from a bridge account,
+        // as we are explicitly transferring from a bridge account here.
+        transfer_check_stateful(&transfer_action, state, from).await
+    }
+
+    #[instrument(skip_all)]
+    async fn execute<S: StateWriteExt>(&self, state: &mut S, from: Address) -> Result<()> {
+        let transfer_action = TransferAction {
```
noot, line 64: the memo is unused, which is fine, when we decide on the memo structure we should also store that (or data extracted from it). maybe add a comment here about that?

crates/astria-sequencer/src/app/tests_breaking_changes.rs
```diff
@@ -238,5 +241,29 @@ async fn app_execute_transaction_with_every_action_snapshot() {
...
     app.execute_transaction(signed_tx).await.unwrap();
+
+    // execute BridgeUnlock action
+    let tx = UnsignedTransaction {
+        params: TransactionParams {
+            nonce: 0,
+            chain_id: "test".to_string(),
+        },
+        actions: vec![
+            BridgeUnlockAction {
```
noot, line 252: you can put this action into the big transaction above, just put it under the `BridgeLockAction` so that the `BridgeLockAction` is executed first

crates/astria-sequencer/src/app/tests_execute_transaction.rs
```diff
@@ -1072,3 +1038,80 @@ async fn app_stateful_check_fails_insufficient_total_balance() {
...
+            nonce: 0,
+            chain_id: "test".to_string(),
+        },
+        actions: vec![action.into()],
+    };
+
+    let signed_tx = tx.into_signed(&bridge_signing_key);
+    app.execute_transaction(signed_tx)
+        .await
+        .expect("failed while executing bridge unlock action");
```
noot, line 1108:
```suggestion
        .expect("executing bridge unlock action should succeed");
```

crates/astria-sequencer/src/app/tests_execute_transaction.rs
```diff
@@ -1072,3 +1038,80 @@ async fn app_stateful_check_fails_insufficient_total_balance() {
...
+    assert_eq!(
+        app.state
+            .get_account_balance(bridge_address, asset_id)
+            .await
+            .expect("failed while getting account balance"),
```
noot, line 1113:
```suggestion
            .expect("getting account balance should succeed"),
```

noot (approved):
nice, this looks good!

PR #1037, opened 2024-05-02
feat(sequencer)!: implement `FeeChangeAction` for the authority component

### LINKED ISSUE #1003: sequencer: make fees configurable via a transaction

noot:
## Summary
 implement `FeeChangeAction` for the authority component, which allows fees to be configured at chain runtime.

## Background
we want to allow for fees to be configured at chain runtime.

## Changes
-  implement the `FeeChangeAction` type

## Testing
unit tests

## Breaking Changelist
- app state can be changed by sending a `FeeChangeAction`.

## Related Issues

closes #1003


proto/protocolapis/astria/protocol/transactions/v1alpha1/types.proto
```diff
@@ -181,3 +182,22 @@ message BridgeLockAction {
   // will receive the bridged funds
   string destination_chain_address = 5;
 }
+
+message FeeChangeAction {
+  // note that the proto number ranges are doubled from that of `Action`.
+  // this to accomodate both `base_fee` and `byte_cost_multiplier` for each action.
+  oneof value {
+    // core protocol fees are defined on 1-20
+    bytes transfer_base_fee = 1;
```
joroshiba, line 191: Should these by bytes or typed as u128 or uint64?

It might be cleaner to have them all be uint128 and then because it's a message they are optional. could get rid of the `oneof`, can update as many as you want at once anything not set stays the same?

noot: updated to make each `oneof value` type `uint128` - i think it's better to leave it as a oneof for now, as changing multiple fees in one action can get more complex than just bundling a tx with multiple `FeeChangeAction`s in it.

joroshiba (approved):

crates/astria-sequencer/src/transaction/mod.rs
```diff
@@ -72,15 +75,20 @@ pub(crate) async fn check_balance_mempool<S: StateReadExt + 'static>(
     Ok(())
 }
 
-// Checks that the account has enough balance to cover the total fees for all actions in the
-// transaction.
+// Checks that the account has enough balance to cover the total fees and transferred values
+// for all actions in the transaction.
+//
+// allow too many lines because of the number of actions and the need to check each one.
+#[allow(clippy::too_many_lines)]
```
SuperFluffy, line 82: IMO this is a strong hint to have a bunch of free-standing functions for each of the actions, and delegate the checks there. I.e:

```rust
match action {
    Action::Transfer(act) => process_transfer_action(act, fees_by_asset)
    ...
}
```

crates/astria-sequencer/src/authority/action.rs
```diff
@@ -94,3 +98,183 @@ impl ActionHandler for SudoAddressChangeAction {
...
+    #[instrument(skip_all)]
+    async fn execute<S: StateWriteExt>(&self, state: &mut S, _: Address) -> Result<()> {
+        use crate::{
+            accounts::state_ext::StateWriteExt as _,
+            bridge::state_ext::StateWriteExt as _,
+            ibc::state_ext::StateWriteExt as _,
+            sequence::state_ext::StateWriteExt as _,
+        };
+
+        match self.fee_change {
```
SuperFluffy, line 129: do ever get events for this? I wonder if it might be warranted to emit `warn!` level events for each one of these changes.

noot: yeah, Jordan and I discussed this, it might be useful to emit ABCI events for certain things occurring for easier indexing in general. this would be one of those spots. will probably look over all of sequencer at some point to decide what to add

PR #1044, opened 2024-05-03
refactor(proto)!: update `SignedTransaction` to contain `Any` for transaction

### LINKED ISSUE #310: Verifying signed transactions is not reliable because protobuf serialization is not deterministic

noot:
## Summary
as title says

## Background
protobuf encoding is not deterministic, so we want to verify the signature over the exact same bytes signed.

## Changes
- update `SignedTransaction` to contain `Any` for transaction instead of `UnsignedTransaction`

## Testing
unit test

## Breaking Changelist
- the `SignedTransaction` type is changed
- tx hashes are no longer the same

## Related Issues

closes #310


crates/astria-core/src/protocol/transaction/v1alpha1/mod.rs
```diff
@@ -282,6 +327,10 @@ enum UnsignedTransactionErrorKind {
     Action(#[source] action::ActionError),
     #[error("`params` field is unset")]
     UnsetParams(),
+    #[error("invalid type URL")]
+    InvalidTypeUrl,
```
SuperFluffy, line 331: Useful to provide a bit more info:
```suggestion
    #[error("encountered invalid type URL when converting from `google.protobuf.Any`; got `{got}`, expected `{}`", UnsignedTransaction::type_url())]
    InvalidTypeUrl { got: String, },
```

crates/astria-core/src/protocol/transaction/v1alpha1/mod.rs
```diff
@@ -282,6 +327,10 @@ enum UnsignedTransactionErrorKind {
...
+    #[error("invalid bytes in Any")]
```
SuperFluffy, line 332: I would be a tad more specific about this. It might be that 
```suggestion
#[error("failed to decode `google.protobuf.Any` to `{}`", UnsignedTransaction::type_url())]
```

crates/astria-core/src/protocol/transaction/v1alpha1/mod.rs
```diff
@@ -282,6 +327,10 @@ enum UnsignedTransactionErrorKind {
...
+    InvalidAnyBytes(#[source] prost::DecodeError),
```
SuperFluffy, line 333:
```suggestion
    DecodeAny(#[source] prost::DecodeError),
```

crates/astria-core/src/protocol/transaction/v1alpha1/mod.rs
```diff
@@ -10,6 +10,9 @@ use super::raw;
 pub mod action;
 pub use action::Action;
 
+pub const UNSIGNED_TRANSACTION_TYPE_URL: &str =
+    "/astria.protocol.transaction.v1alpha1.UnsignedTransaction";
```
SuperFluffy, line 14: I am not generally opposed to doing this (to save some allocations), but we must ensure that const matches the generated type URL of the `prost::Name` trait. So either use `UnsignedTransaction::type_url()` everywhere, or alternatively use the hardcoded value here, but add a unit test like below.

```rust
#[test]
fn hardcoded_type_url_matches_generated_trait() {
    assert_eq!(UNSIGNED_TRANSACTION_TYPE_URL, &UnsignedTransaction::type_url(),);
}
```

Read my other comments re the use of `Name::type_url()` with this in mind - either use the trait method everywhere, or ensure that the const always stays valid.

SuperFluffy (commented):
I think we could amend the `SignedTransaction` type to contain the raw bytes that we signed over, i.e. do something like this:

```rust
struct SignedTransaction {
    signature: Signature,
    verification_key: VerificationKey,
    transaction: UnsignedTransaction,
    transaction_bytes: Bytes,
}
```

Because we don't allow mutation of the signed transaction we can ensure that the signature always stays valid.

crates/astria-core/src/protocol/transaction/v1alpha1/mod.rs
noot: updated to use `type_url()`

noot:
@SuperFluffy addressed comments! what's the point of keeping the tx bytes inside the `SignedTransaction`? there isn't any case right now where we would need to re-verify the signature, as `SignedTransaction` can only be constructed if the signature was valid.

SuperFluffy:
@noot 
> @SuperFluffy addressed comments! what's the point of keeping the tx bytes inside the `SignedTransaction`? there isn't any case right now where we would need to re-verify the signature, as `SignedTransaction` can only be constructed if the signature was valid.

My thinking is in similar vain to what @joroshiba had in their comment: it's not about the signature but about maintaining the exact same byte-representation when going `SignedTransaction <> raw::SignedTransaction`. Imagine this flow:

```
network bytes --1> raw Signed Transaction --2> checked Signed Transaction --3> raw Signed Transaction --4> network bytes
```

`raw Signed Transaction` contains `UnsignedTransaction` in the form of an `Any`.

Right now, if you do step 2, i.e. `raw Signed Transaction --2> checked Signed Transaction`, you throw away the original bytes that made up the unsigned transaction. So when you go `checked Signed Transaction --3> raw Signed Transaction` you re-encode you don't have a guarantee that you get the same bytes.

If you keep the original bytes around you can completely avoid the issue.

joroshiba (approved):

PR #1047, opened 2024-05-06
ci: test using depot instead of buildjet

joroshiba:
## Summary
Migrates us away from using buildjet for builds towards using depot. This reduces the number of third party services we rely on for tests, and is as fast if not faster than buildjet based on recent tests.


WafflesVonMaple (approved):

joroshiba:
closed in 7 days.

joroshiba:

.github/workflows/test.yml
Fraser999, line 1: There are a few instances of `cache-provider: "buildjet"` remaining in this file.

.github/workflows/test.yml
```diff
@@ -77,7 +77,7 @@ jobs:
           fi
 
   compiles:
-    runs-on: buildjet-4vcpu-ubuntu-2204
+    runs-on: depot-ubuntu-22.04-4
```
SuperFluffy, line 80: I think this, and the others, should be updated to `depot-ubuntu-24.04`

SuperFluffy (commented):
Commented on one of the lines: `depot-ubuntu-24.04` is available. I think we should bump to that at the same time.

.github/workflows/test.yml
joroshiba: The cache-provider is free and still generally useful I believe.

Using non github runners results in getting throttled on caching speeds.

Fraser999 (approved):
Approved, but as per @SuperFluffy's comment, would be good to use a more recent version of Ubuntu.

.github/workflows/test.yml
joroshiba: Completed in https://github.com/astriaorg/astria/pull/1047/commits/81a7c4492280530e2d834cce5edb6e1284392472

PR #1055, opened 2024-05-08
feat(sequencer)!: add fees to genesis state

noot:
## Summary
instead of having hard coded default fee values, the fees now must be set in the genesis file. this makes it more explicit as to how the network is configured.

## Background
related to https://github.com/astriaorg/astria/issues/1003

## Changes
- update genesis app state to contain required fees
- set these fees in state in `init_chain`

## Testing
unit tests

## Breaking Changelist
- genesis app state must now contain fees

## Related Issues

related to https://github.com/astriaorg/astria/issues/1003

### DIFF
--- a/crates/astria-sequencer/test-genesis-app-state.json
+++ b/crates/astria-sequencer/test-genesis-app-state.json
@@ -21,6 +21,14 @@
     "inbound_ics20_transfers_enabled": true,
     "outbound_ics20_transfers_enabled": true
   },
+  "fees": {
+    "transfer_base_fee": 12,
+    "sequence_base_fee": 32,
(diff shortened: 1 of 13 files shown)


crates/astria-sequencer/test-genesis-app-state.json
joroshiba, line 1: Can we add these fees here as well: https://github.com/astriaorg/astria/blob/main/charts/sequencer/files/cometbft/config/genesis.json#L34

joroshiba (approved):
Approved, but pretty sure smoke test will fail without the update to chart, commented below, should be in this PR.

PR #1057, opened 2024-05-08
tests(sequencer): add snapshot tests for app hash

### LINKED ISSUE #1051: sequencer: breaking changes unit tests
we should implement some basic unit tests that catch breaking changes in the sequencer (state tree changes)

noot:
## Summary
 add snapshot tests for app hash to catch breaking changes made to the app logic.

## Background
if changes are breaking, we should know about it. these tests should catch (most) breaking changes to the application state itself.

## Changes
- add snapshot tests for `init_chain`, `finalize_block`, and all actions except for IBC actions as they require extensive setup.

## Testing
no code was changed, just tests added

## Related Issues

closes #1051


Fraser999 (approved):
LGTM.

> thoughts on adding the snapshot directly to the unit tests that already exist there?

Seems like a good idea to me, but maybe best to wait for more opinions?

noot:
@Fraser999 ended up adding a snapshot test which tests every action!

joroshiba (approved):

PR #1059, opened 2024-05-08
feat(sequencer-utils): add subcommand to parse Celestia blob data

### LINKED ISSUE #1050: Provide functionality to parse blob data via `astria-sequencer-utils`
The tool should be updated as follows:
- move the existing functionality into a dedicated subcommand
- provide a new subcommand which expects a base64-encoded string as input, parses it as blob data and outputs it in a human-readable format

Fraser999:
## Summary
This provides a new subcommand in the `astria-sequencer-utils` tool to allow for decoding Celestia base-64 blobs containing Astria data.

## Background
It would be helpful to be able to visualize the data contained in the blobs as available in [the Celenium blob viewer](https://mocha.celenium.io/block/1726074?tab=transactions) (found by expanding a blob on that page).

## Changes
- The tool was restructured to make the existing command into a subcommand `copy-genesis-state`.  The existing functionality was not changed.
- A new subcommand `parse-blob` was added, best explained via the help output:
```
Parse blob data from an arg, a file, or stdin

Usage: astria-sequencer-utils parse-blob [OPTIONS] <BLOB|PATH>

Arguments:
  <BLOB|PATH>  Base64-encoded blob data, or a file containing this, or stdin if `-`

Options:
  -f, --format <FORMAT>  Configure formatting of output [default: display] [possible values: display, json]
  -v, --verbose          Display verbose output (e.g. displays full contents of transactions in rollup data)
  -h, --help             Print help
```
- The input data can be of the form of the new batched blob, or the old single-entry blob, and each can be either sequencer metadata or rollup data.  I don't think there's a way to know in advance of attempting to parse which of the four flavours we're dealing with, so the tool just tries each case until one successfully parses or fails if it can't parse as any of them.  I pretty strongly dislike this trial-and-error parsing approach - if there's a way to avoid it (other than embedding more data in the blobs themselves) I'd be happy to implement it!
- All binary fields in the output data (e.g. sequencer block hash, proposer address) are base-64 encoded.  It would be trivial to hex-encode any of these instead of, or as well as, base-64 encoding them.
- I added the `deprecated` feature to `clap` so it emits deprecation warnings, and fixed up the places where we were using deprecated functionality.

## Testing
There are happy-path integration tests for the CLI using [`assert_cmd`](https://lib.rs/crates/assert_cmd) and [`predicates`](https://lib.rs/crates/predicates).  These test the four possible types of input, and for each case assert that the various CLI option combos generate expected output.  The input and expected output data for these tests live in files in a new `tests/resources` folder to avoid cluttering the test source code with lengthy string literals.

## Related Issues
Closes #1050.


crates/astria-sequencer-relayer/src/relayer/write/conversion.rs
```diff
@@ -71,7 +71,7 @@ impl Submission {
     // significance.
     #[allow(clippy::cast_precision_loss)]
     pub(super) fn compression_ratio(&self) -> f64 {
-        self.compressed_size() as f64 / self.uncompressed_size() as f64
+        self.uncompressed_size() as f64 / self.compressed_size() as f64
```
noot, line 74: maybe we should add a comment saying this is now uncompressed:compressed?

crates/astria-sequencer-utils/src/blob_parser.rs
```diff
@@ -0,0 +1,797 @@
...
+    bytes::Bytes,
+    Message,
+};
+use serde::Serialize;
+
+#[derive(clap::Args, Debug)]
+pub struct Args {
+    /// Base64-encoded blob data
+    #[arg(value_name = "BLOB")]
+    input: String,
```
noot, line 64: is the blob just pasted in directly to this command? should there be an option for reading from a file perhaps?

crates/astria-sequencer-relayer/src/relayer/write/conversion.rs
SuperFluffy, line 74: Please make this change in a separate pr

crates/astria-core/src/sequencerblock/v1alpha1/celestia.rs
```diff
@@ -520,6 +520,11 @@ impl SubmittedMetadata {
         &self.header
     }
 
+    /// Returns the rollup IDs.
+    pub fn rollup_ids(&self) -> impl Iterator<Item = &RollupId> {
```
SuperFluffy, line 524: Should we introduce a type that wraps this iterator (so that we don't return an anonymous type)?

SuperFluffy (commented):
Two comments for now, gonna make a main review tomorrow

noot (approved):
looks good, cli worked for me :)

crates/astria-sequencer-relayer/src/relayer/write/conversion.rs
Fraser999: I moved this change to [a separate PR](https://github.com/astriaorg/astria/pull/1075) and added a comment there.

crates/astria-core/src/sequencerblock/v1alpha1/celestia.rs
Fraser999: I'm not sure I see a benefit to that.  Is it a stylistic thing?

I could always change the return type to `std::slice::Iter<RollupId>`, but that's a bit less canonical than `impl Iterator` I think - it couples the struct's interface a bit more tightly to its private internals.

SuperFluffy: The idea is to avoid impl trait in return position for public APIs (to not return anonymous types that is). It seems this actually never made it into the official Rust API guidelines. Here is a discussion that never lead anywhere, but I agree with burntsushi's point: https://github.com/rust-lang/api-guidelines/discussions/60

Fraser999: Hmm.  Can't say I disagree with the points in general, but otoh I think the arguments against returning impl Trait are particularly weak in this specific case.  It seems unlikely that a consumer of this crate would need to hold an iterator of rolllup IDs in a struct or make it an associated type of their own trait.

I think it's reasonably common to see `impl Iterator` as a return type, and the fact that the linked discussion has relatively few comments/upvotes points to this not being an issue in general.

Saying that, the only real objection I have to your original suggestion is purely that it seems unnecessary - it clutters the code and the rustdocs, i.e. it's a very weak objection given how small a change it would be.  So, I'll just drop in a new `RollupIdIter` :)

Fraser999: Done in [63a605c0](https://github.com/astriaorg/astria/pull/1059/commits/63a605c0a4c6c7006f22af34a00bb4bbf9fffec7).

crates/astria-sequencer-utils/src/blob_parser.rs
Fraser999: Good call - I'll add an option to get the input from a file too.

crates/astria-cli/src/cli/mod.rs
SuperFluffy, line 1: Changes to this file are probably necessary because of the clap bump?

Cargo.toml
```diff
@@ -54,7 +54,7 @@ base64-serde = "0.7.0"
 bytes = "1"
 celestia-tendermint = "0.32.1"
 celestia-types = "0.1.1"
-clap = "4"
+clap = "4.5.4"
```
SuperFluffy, line 57: Agree with this change. We should probably provide full versions on all crates we are using.

crates/astria-sequencer-utils/Cargo.toml
```diff
@@ -9,8 +9,27 @@ repository = "https://github.com/astriaorg/astria"
...
+rlp = "0.5.2"
 serde = { workspace = true }
 serde_json = { workspace = true }
 
+astria-core = { path = "../astria-core", features = ["brotli"] }
 astria-eyre = { path = "../astria-eyre" }
+astria-merkle = { path = "../astria-merkle" }
+
+[dev-dependencies]
+assert_cmd = "2.0.14"
```
SuperFluffy, line 34: Nice

crates/astria-core/src/sequencerblock/v1alpha1/celestia.rs
```diff
@@ -494,6 +494,17 @@ pub struct SubmittedMetadata {
     rollup_ids_proof: merkle::Proof,
 }
 
+/// An iterator over rollup IDs.
+pub struct RollupIdIter<'a>(std::slice::Iter<'a, RollupId>);
+
+impl<'a> Iterator for RollupIdIter<'a> {
+    type Item = &'a RollupId;
```
SuperFluffy, line 501: `RollupId` are `Copy`, so this could be a `type Item = RollupId;`

crates/astria-sequencer-utils/src/blob_parser.rs
SuperFluffy: To pile onto this, I think it would be nice if this could also read input from STDIN, maybe after a `--` flag.

SuperFluffy (approved):
This looks great. I love the use of `assert_cmd`

crates/astria-cli/src/cli/mod.rs
Fraser999: It's more down to the fact I added the `deprecated` feature to `clap` in the manifest of `astria-sequencer-utils`.  The author deprecated a bunch of stuff (maybe when moving to 3.0) and got serious push back from the community, so he added a `deprecated` feature without which the deprecations aren't applied.

Because we're using a workspace, the features of each dependency become a union of the features enabled by individual workspace targets.

I think it's worth having the feature enabled so we can keep up to date with upcoming `clap` changes, or at least are aware of them, but I guess I should have ticketed that for a later PR?

crates/astria-core/src/sequencerblock/v1alpha1/celestia.rs
Fraser999: Yeah - I've had a similar debate a bunch of times in the past :D  Just because a type _is_ `Copy` I don't know if that means we should favour copying over passing by ref.  It make sense for very small types, but less so for larger types.  The question then boils down to: is a 32-byte array small enough?

I guess I've landed on generally going for pass-by-ref since it's slightly more performant, but it's a weak preference outside of hot paths.

crates/astria-sequencer-utils/src/blob_parser.rs
Fraser999: Done in [f16e647](https://github.com/astriaorg/astria/pull/1059/commits/f16e647ec145aa761ed0c4b8a70227911a9d1fe6).

Fraser999:
@SuperFluffy - I added the input arg to handle reading from a file or from stdin if it's `"-"` (in [f16e647](https://github.com/astriaorg/astria/pull/1059/commits/f16e647ec145aa761ed0c4b8a70227911a9d1fe6)).  I think the single hyphen is maybe a little more common than the double?  It's 50% less typing too! :D

Fraser999:
@noot I updated the readme in [1d48a58](https://github.com/astriaorg/astria/pull/1059/commits/1d48a58cf7efb0ad50c3c4941bfccc7d2976264e).

Issue #986, opened 2024-04-19
validate communicating with correct sequencer chain


joroshiba:
We should add a check to conductor, composer, and sequencer relayer to validate that the sequencer RPC they communicate with is of the expected Astria sequencer network. 

It is possible that RPCs change, or are confusingly named. On startup the services should validate this. There should be an env var specifying the expected chain id and then it can be verified from genesis file. 

```
### Tasks
- [x] composer verify chain id on startup
- [x] sequencer-relayer verify sequencer chain id on startup
- [ ] conductor verify sequencer chain id on startup
```

PR #1063, opened 2024-05-09
feat(sequencer-relayer)!: add chain IDs for sequencer and Celestia to config env vars

### LINKED ISSUE #978: Add network ID of DA to env vars of sequencer-relayer
We want to avoid accidentally connecting to a wrong DA network.  Currently the network ID isn't explicitly specified by the operator of the sequencer relayer; the network is defined only by the endpoint chosen in `ASTRIA_SEQUENCER_RELAYER_CELESTIA_ENDPOINT`.

We should add an explicit configuration env var taking the ID of the relevant Celestia network.

This issue is blocked by #963, since that PR has logic whereby the network ID is retrieved from the Celestia app on startup.  We want to build the ID check into the handler for that response.

Fraser999:
## Summary
Added two new configuration options specifying the expected chain IDs of the sequencer and Celestia networks.

## Background
This is a guard rail to help avoid an operator accidentally connecting the relayer to the wrong network.

## Changes
The configured values are used in the initialization phase, where the relevant clients query the appropriate servers for their chain IDs.

The Celestia client was already retrieving this info prior to this PR.  Now it returns an error if the chain ID doesn't match, and this particular error variant causes the endless retry loop (and ultimately the process) to exit.

The sequencer client was not previously requesting this info, so it now does this in a somewhat similar manner to the Celestia client.  Likewise, a chain ID mismatch there causes the process to exit.

## Testing
A blackbox test for a mismatch in each of the chain IDs was added.

## Breaking Changelist
- added `ASTRIA_SEQUENCER_RELAYER_SEQUENCER_CHAIN_ID`
- added `ASTRIA_SEQUENCER_RELAYER_CELESTIA_CHAIN_ID`

## Related Issues

Part of #986.
Closes #978.


crates/astria-sequencer-relayer/src/relayer/mod.rs
```diff
@@ -288,6 +303,63 @@ impl Relayer {
...
+) -> Result<String, tendermint_rpc::Error> {
+    use sequencer_client::Client as _;
+
+    let response = sequencer_cometbft_client.status().await;
+    // trace-level logging, so using Debug format is ok.
+    #[cfg_attr(dylint_lib = "tracing_debug_field", allow(tracing_debug_field))]
+    {
+        trace!(?response);
+    }
+    response.map(|status_response| status_response.node_info.network.to_string())
```
noot, line 360: the fact that this var is called `network` and not `chain_id` as in other rpc responses (such as genesis and block) bothers me a bit :p I think it's fine to leave though, as it should be the same as the chain id

noot (approved):

crates/astria-sequencer-relayer/src/relayer/mod.rs
Fraser999: Yeah - it doesn't sit well with me either, but it's part of the tendermint API, so my hands are pretty much tied on this one I think.

crates/astria-sequencer-relayer/src/relayer/celestia_client/builder.rs
```diff
@@ -40,6 +43,9 @@ pub(in crate::relayer) enum BuilderError {
     /// The node info response was empty.
     #[error("the celestia node info response was empty")]
     EmptyNodeInfo,
+    /// Mismatch in Celestia chain ID.
+    #[error("mismatch in celestia chain id, configured id: `{configured}`, actual id: `{actual}`")]
```
SuperFluffy, line 47: I am frequently confusing myself when seeing `actual`. Maybe `received` to make it clear that this was read off the remote host?
```suggestion
    #[error("mismatch in celestia chain id, configured id: `{configured}`, received id: `{actual}`")]
```

crates/astria-sequencer-relayer/src/relayer/mod.rs
```diff
@@ -288,6 +303,63 @@ impl Relayer {
...
+            .wrap_err("retry attempts exhausted; bailing")?;
+
+    if actual_sequencer_chain_id == configured_sequencer_chain_id {
+        info!(sequencer_chain_id = %configured_sequencer_chain_id, "confirmed sequencer chain id");
+        return Ok(());
+    }
+    bail!(
+        "mismatch in sequencer chain id, configured id: `{configured_sequencer_chain_id}`, actual \
+         id: `{actual_sequencer_chain_id}`"
+    )
```
SuperFluffy, line 346: With `eyre::ensure!` this could become (changed the wording of the error message a bit to make clear where it came from):
```suggestion
    ensure!(
        actual_sequencer_chain_id == configured_sequencer_chain_id,
        "configured sequencer chain ID does not match remote; configured: `{configured_sequencer_chain_id}`, \
         remote: `{actual_sequencer_chain_id}`",
    );
    info!(sequencer_chain_id = %configured_sequencer_chain_id, "confirmed sequencer chain id");
    Ok(())
```

crates/astria-sequencer-relayer/tests/blackbox/helpers/test_sequencer_relayer.rs
```diff
@@ -594,6 +644,25 @@ impl TestSequencerRelayer {
...
     }
+
+    /// Mounts a `CometBFT` status response with the chain ID set as per
+    /// `TestSequencerRelayerConfig::sequencer_chain_id`.
+    async fn mount_cometbft_status_response(&self) {
+        use tendermint_rpc::endpoint::status;
+
+        let mut status_response: status::Response = serde_json::from_str(STATUS_RESPONSE).unwrap();
+        status_response.node_info.network =
+            tendermint::chain::Id::from_str(&self.actual_sequencer_chain_id).unwrap();
```
SuperFluffy, line 655:
```suggestion
            self.actual_sequencer_chain_id.parse().unwrap();
```

crates/astria-sequencer-relayer/tests/blackbox/helpers/test_sequencer_relayer.rs
```diff
@@ -9,6 +9,7 @@ use std::{
     io::Write,
     mem,
     net::SocketAddr,
+    str::FromStr,
```
SuperFluffy, line 12: We can avoid this by using `parse` below

SuperFluffy (approved):
Only minor comments. I tend to prefer `received` over `actual` nowadays because it requires less thinking. :)

crates/astria-sequencer-relayer/src/relayer/celestia_client/builder.rs
Fraser999: Done in [458f2a2](https://github.com/astriaorg/astria/pull/1063/commits/458f2a27ff3a3fe97880a3b049c4359c1078058f).

crates/astria-sequencer-relayer/src/relayer/mod.rs
Fraser999: Done in [458f2a2](https://github.com/astriaorg/astria/pull/1063/commits/458f2a27ff3a3fe97880a3b049c4359c1078058f), although I used the term "received" rather than "remote" to be consistent with the similar change above.

crates/astria-sequencer-relayer/tests/blackbox/helpers/test_sequencer_relayer.rs
Fraser999: Done in [458f2a2](https://github.com/astriaorg/astria/pull/1063/commits/458f2a27ff3a3fe97880a3b049c4359c1078058f).

crates/astria-sequencer-relayer/tests/blackbox/helpers/test_sequencer_relayer.rs
Fraser999: Done in [458f2a2](https://github.com/astriaorg/astria/pull/1063/commits/458f2a27ff3a3fe97880a3b049c4359c1078058f).

joroshiba (approved):
infra approval

PR #1065, opened 2024-05-11
fix(sequencer): fix ibc prefix conversion

noot:
## Summary
prefixes were not being converted correctly for bridge locks, i refactored the ibc prefix conversion to be in its own function and added unit tests for all cases.

## Background
found while testing incoming ibc transfers to a bridge account, the denom was not being prefixed before being compared to the allowed asset ID for the bridge account.

## Changes
- put all conversion from packet denom -> converted ibc'd denom into its own `convert_denomination` function
- this is now called before everything else in `execute_ics20_transfer`, so the correct converted denom should be used

## Testing
unit tests + tested it with a bridge lock via ibc

### DIFF
--- a/crates/astria-sequencer/src/ibc/ics20_transfer.rs
+++ b/crates/astria-sequencer/src/ibc/ics20_transfer.rs
@@ -244,7 +244,13 @@ impl AppHandlerExecute for Ics20Transfer {
         .await
         {
             Ok(()) => TokenTransferAcknowledgement::success(),
-            Err(e) => TokenTransferAcknowledgement::Error(e.to_string()),
+            Err(e) => {
+                tracing::debug!(
+                    error = AsRef::<dyn std::error::Error>::as_ref(&e),
+                    "failed to execute ics20 transfer"
+                );
+                TokenTransferAcknowledgement::Error(e.to_string())
+            }
         };
 
         let ack_bytes: Vec<u8> = ack.into();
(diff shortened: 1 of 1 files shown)


steezeburger (approved):

crates/astria-sequencer/src/ibc/ics20_transfer.rs
```diff
@@ -244,7 +244,13 @@ impl AppHandlerExecute for Ics20Transfer {
         .await
         {
             Ok(()) => TokenTransferAcknowledgement::success(),
-            Err(e) => TokenTransferAcknowledgement::Error(e.to_string()),
+            Err(e) => {
+                tracing::debug!(
```
SuperFluffy, line 248: I wonder if this should be elevated to `info!`

crates/astria-sequencer/src/ibc/ics20_transfer.rs
```diff
@@ -373,6 +379,42 @@ async fn execute_ics20_transfer_bridge_lock<S: StateWriteExt>(
...
+
+    Ok(denom)
+}
+
+fn prefix_denomination(
+    packet_denom: Denom,
+    dest_port: &PortId,
+    dest_channel: &ChannelId,
+    is_refund: bool,
+) -> Denom {
```
SuperFluffy, line 410: looking at how you use this this might be a use case for `Cow<'a, Denom>` with `Cow::Borrowed(&'a Denom)` if `is_refund` and `Cow::Owned(Denom)` otherwise.

SuperFluffy (approved):
lgtm

I have to note that it's difficult for me to understand how this fits into ibc because I lack background knowledge.

PR #1067, opened 2024-05-13
feat(sequencer): query full denomination from asset ID

### LINKED ISSUE #1053: sequencer: query node for full denom from asset ID
implement a query endpoint that returns the stored denom given an asset ID. good for ux/block explorers

noot:
## Summary
implement ABCI query to get full denomination from an asset ID, if it exists in the latest sequencer state.

## Background
useful for block explorers/UIs as txs only have asset IDs, not the full denom.

## Changes
- implement ABCI query to get full denomination from an asset ID, if it exists in the latest sequencer state

## Testing
unit tests

## Related Issues

closes #1053


crates/astria-sequencer/src/asset/query.rs
```diff
@@ -0,0 +1,101 @@
...
+                info: AbciErrorCode::INTERNAL_ERROR.to_string(),
+                log: format!("failed to retrieve denomination: {err:?}"),
+                ..response::Query::default()
+            };
+        }
+    };
+
+    let payload = DenomResponse {
+        height: height as u64,
+        denom: denom.clone(),
```
Fraser999, line 62:
```suggestion
        height,
        denom,
```

crates/astria-sequencer/src/asset/query.rs
```diff
@@ -0,0 +1,101 @@
...
+    }
+    .into_raw()
+    .encode_to_vec()
+    .into();
+
+    let height = u32::try_from(height).expect("height must fit into a u32");
+    response::Query {
+        code: tendermint::abci::Code::Ok,
+        key: request.path.clone().into_bytes().into(),
```
Fraser999, line 71:
```suggestion
        key: request.path.into_bytes().into(),
```

crates/astria-sequencer/src/asset/query.rs
```diff
@@ -0,0 +1,101 @@
...
+
+    let payload = DenomResponse {
+        height: height as u64,
+        denom: denom.clone(),
+    }
+    .into_raw()
+    .encode_to_vec()
+    .into();
+
+    let height = u32::try_from(height).expect("height must fit into a u32");
```
Fraser999, line 68: Probably not really an issue, but this `try_from` is more restrictive than `Height::try_from` where it only requires the value to be not more than `i64::MAX`.

```suggestion
    let height = tendermint::block::Height::try_from(height).expect("height must fit into an i64");
```

Fraser999 (commented):
Nothing critical, but a few suggestions.

I don't think a server should be returning a full stack trace as part of an error response message, hence I suggested changing the formatting of the anyhow errors from debug to alternate `{:#}` to only include the display-formatted source chain.  This might not be what we do elsewhere, so feel free to ignore.

Also, it might be nice to have some sad-path tests for the new endpoint, but that's not a blocker IMO.

noot:
@Fraser999 addressed comments - I changed the error returned to be display-formatted, but in the account query module, we return the debug formatted errors. do you think we should change it there as well?

Fraser999:
@noot - yes, I would vote for changing them all to return the alternate display form.  There's still one debug one lingering in query.rs on line 95 btw :)

crates/astria-sequencer/src/asset/query.rs
```diff
@@ -0,0 +1,101 @@
...
+            };
+        }
+    };
+
+    let denom = match snapshot.get_ibc_asset(asset_id).await {
+        Ok(denom) => denom,
+        Err(err) => {
+            return response::Query {
+                code: AbciErrorCode::INTERNAL_ERROR.into(),
+                info: AbciErrorCode::INTERNAL_ERROR.to_string(),
```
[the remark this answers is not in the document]
SuperFluffy: I agree. IMO we should define a specific error code for this and stick to it. We have u32 integers available. :)

IMO the easiest way to go about it is to either change `get_ibc_asset` to return `Result<Option<Denom>>`, or use a typed error inside `get_ibc_asset` and then try to downcast to it here.

(BTW, we used to have a protobuf enumeration for these ABCI codes; what happened to it?)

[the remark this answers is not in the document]
noot: > (BTW, we used to have a protobuf enumeration for these ABCI codes; what happened to it?)

it wasn't actually used in the code anywhere, so it was deleted at some point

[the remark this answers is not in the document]
noot: @SuperFluffy @Fraser999 updated `get_ibc_asset` to return `Result<Option<Denom>>` and also have a specific error code for if the value isn't found

Fraser999 (approved):

joroshiba (approved):
api approval

proto/protocolapis/astria/protocol/asset/v1alpha1/types.proto
```diff
@@ -0,0 +1,9 @@
+syntax = "proto3";
+
+package astria.protocol.asset.v1alpha1;
+
+// A response containing the denomination given an asset ID.
+message DenomResponse {
+  uint64 height = 2;
```
SuperFluffy, line 7: should this start at 1?

SuperFluffy (approved):
looks fine, but the `DenomResponse` starting at number 1 is throwing me off?

proto/protocolapis/astria/protocol/asset/v1alpha1/types.proto
SuperFluffy: Yeah, they should all start at 1. Too late now because that would be a breaking change. :shrug:

PR #1075, opened 2024-05-15
chore(sequencer-relayer): change compression ratio calculation

Fraser999:
## Summary
Inverts the compression ratio calculation.

## Background
It's normal to express a compression ratio as uncompressed:compressed, but were reporting it as the inverse of this in logs and metrics (despite the metric's description claiming it was uncompressed:compressed).

## Changes
As per summary.

## Testing
Added a unit test.

## Metrics
No new metric added, but the value of `astria_sequencer_relayer_compression_ratio_for_astria_block` should now generally be > 1 rather than < 1.

### DIFF
--- a/crates/astria-sequencer-relayer/src/relayer/write/conversion.rs
+++ b/crates/astria-sequencer-relayer/src/relayer/write/conversion.rs
@@ -67,11 +67,12 @@ impl Submission {
         )
     }
 
+    /// The ratio of uncompressed blob size to compressed size.
     // allow: used for metric gauges, which require f64. Precision loss is ok and of no
     // significance.
     #[allow(clippy::cast_precision_loss)]
     pub(super) fn compression_ratio(&self) -> f64 {
-        self.compressed_size() as f64 / self.uncompressed_size() as f64
+        self.uncompressed_size() as f64 / self.compressed_size() as f64
     }
 
     pub(super) fn compressed_size(&self) -> usize {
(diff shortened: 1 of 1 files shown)


crates/astria-sequencer-relayer/src/relayer/write/conversion.rs
```diff
@@ -575,4 +576,12 @@ mod tests {
         let payload = input.try_into_payload().unwrap();
         assert_eq!(2, payload.num_blobs());
     }
+
+    #[tokio::test]
+    async fn should_compress() {
+        let mut next_submission = NextSubmission::new(include_all_rollups());
+        next_submission.try_add(block(1)).unwrap();
+        let submission = next_submission.take().await.unwrap();
+        assert!(submission.compression_ratio() > 1.0);
```
SuperFluffy, line 585: I mean, this is not guaranteed always. In principle the compression *could* make things worse.

I don't think a unit test here is strictly necessary.

SuperFluffy (approved):
Looks good. I think we can forego the unit test altogether because a) the compression ratio can actually be smaller 1, and b) this code is just a visual crutch.

crates/astria-sequencer-relayer/src/relayer/write/conversion.rs
Fraser999: Yeah - fair enough.  It did seem a little overkill to me too, although if we updated something in the future which caused this test to start failing, we could always change the test to use data which is guaranteed to compress.

Either way - I'll remove the test, as it's unlikely we'll accidentally cause a regression in the `compression_ratio` method.

Fraser999: Removed in [210daca](https://github.com/astriaorg/astria/pull/1075/commits/210daca1a6ea79387378df6959ca0e6d41a133c7).

PR #1078, opened 2024-05-16
feat(charts): Add support for forma genesis config

mycodecrafting:
## Summary
Changes to charts to support Forma


charts/evm-rollup/files/genesis/geth-genesis.json
```diff
@@ -26,6 +26,9 @@
         "terminalTotalDifficulty": 0,
         "terminalTotalDifficultyPassed": true,
         "ethash": {},
+        {{- if .Values.config.rollup.genesis.forma }}
```
joroshiba, line 29: Should we make this more generic and just add a "additional genesis" field that's converted to json?

mycodecrafting: that sounds better.

joroshiba (approved):

PR #1080, opened 2024-05-17
feat(conductor): respect shutdown signals during init

SuperFluffy:
## Summary
Conductor now respects shutdown signals it receives during init.

## Background
Conductor's task ignored shutdowns while still initializing. This meant that Conductor would hang for up to 30 seconds.

## Changes
- refactor conductor's constituent long-running tasks to separate initialization and running
- listen for the shutdown signal in all of conductor's tasks

## Testing
Run conductor with endpoints that hang indefinitely and sending it SIGTERM. Observe that conductor shuts down quickly.

The main operation of conductor is unaffected on the happy path: all blackbox tests run to completion.

A proper test for the shutdown logic will be implemented in a follow-up refactor similar to https://github.com/astriaorg/astria/pull/889

### DIFF
--- a/crates/astria-conductor/src/executor/mod.rs
+++ b/crates/astria-conductor/src/executor/mod.rs
@@ -250,7 +250,7 @@ pub(crate) struct Executor {
 impl Executor {
     #[instrument(skip_all, err)]
     pub(crate) async fn run_until_stopped(mut self) -> eyre::Result<()> {
-        tokio::select!(
+        select!(
             () = self.shutdown.clone().cancelled_owned() => {
                 info!(
                     "received shutdown signal while initializing task; \
--- a/crates/astria-conductor/src/executor/mod.rs
+++ b/crates/astria-conductor/src/executor/mod.rs
@@ -214,7 +214,11 @@ impl Handle<StateIsInit> {
     }
 }
 
+/// The private inner executor implementing all the logic.
 pub(crate) struct Executor {
+    /// The execution client driving the rollup.
+    client: Client,
+
     /// The mode under which this executor (and hence conductor) runs.
     mode: CommitLevel,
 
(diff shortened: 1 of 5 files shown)


crates/astria-conductor/src/executor/mod.rs
```diff
@@ -241,32 +242,29 @@ pub(crate) struct Executor {
...
     /// without re-executing on top of the rollup node.
     blocks_pending_finalization: HashMap<u32, Block>,
+
+    /// The maximum permitted spread between firm and soft blocks.
+    max_spread: Option<usize>,
 }
 
 impl Executor {
-    #[instrument(skip_all)]
+    #[instrument(skip_all, err)]
```
noot, line 251: what does the `err` addition do?

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -536,20 +556,23 @@ impl Executor {
     }
 
     #[instrument(skip_all)]
-    async fn set_initial_node_state(&mut self, client: Client) -> eyre::Result<()> {
+    async fn set_initial_node_state(&mut self) -> eyre::Result<()> {
         let genesis_info = {
-            let mut client = client.clone();
-            async move {
-                client
+            // let mut client = self.client.clone();
```
noot, line 561: remove

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -536,20 +556,23 @@ impl Executor {
...
                     .await
                     .wrap_err("failed getting genesis info")
             }
         };
         let commitment_state = {
-            let mut client = client.clone();
-            async move {
-                client
+            // let mut client = self.client.clone();
+            // async move {
```
noot, line 572: remove

noot (approved):

crates/astria-conductor/src/executor/mod.rs
SuperFluffy: Done in  59dbca

crates/astria-conductor/src/executor/mod.rs
SuperFluffy: Sorry, overlooked this: this generates an error event under the span created by the instrumented method.

PR #1081, opened 2024-05-17
chore: add `clippy::arithmetic-side-effects` lint and fix resulting warnings

### LINKED ISSUE #981: Enable `clippy::arithmetic-side-effects`
We should enable this lint in CI by adding `--warn clippy::arithmetic-side-effects` to [.github/workflows/test.yml](https://github.com/astriaorg/astria/blob/6d25f2c7e1f869a9a0a9174eb0c94229b2cc0f93/.github/workflows/test.yml#L156).

We can add `clippy::arithmetic_side_effects` to the existing list of allows in [crates/astria-core/src/generated/mod.rs](https://github.com/astriaorg/astria/blob/6d25f2c7e1f869a9a0a9174eb0c94229b2cc0f93/crates/astria-core/src/generated/mod.rs#L1-L5).

It's ok to add `allow`s for this lint in test code, but warnings from production code should generally be fixed without resorting to ignoring the lint.

Fraser999:
## Summary
Added a clippy lint to CI to check for arithmetic side-effects (e.g. overflows).

## Background
This is to harden the production codebase.

## Changes
- Added `--warn clippy::arithmetic-side-effects` to the clippy CI target.
- Fixed resulting warnings.

## Testing
No new tests added.

## Related Issues
Closes #981.


crates/astria-composer/src/executor/bundle_factory/mod.rs
```diff
@@ -87,16 +87,21 @@ impl SizedBundle {
...
+            .ok_or_else(|| SizedBundleError::NotEnoughSpace(seq_action.clone()))?;
+
+        if new_size > self.max_size {
             return Err(SizedBundleError::NotEnoughSpace(seq_action));
         }
 
         self.rollup_counts
             .entry(seq_action.rollup_id)
-            .and_modify(|count| *count += 1)
+            .and_modify(|count| *count = count.saturating_add(1))
```
Fraser999, line 101: This one seemed highly unlikely to overflow to me, so I went with `saturating_add` mainly to avoid complicating the code, but I can change to return an error if anyone thinks it's a realistic possibility.

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -728,7 +728,7 @@ fn does_block_response_fulfill_contract(
         ExecutionKind::Firm => state.firm_number(),
         ExecutionKind::Soft => state.soft_number(),
     };
-    let expected = current + 1;
+    let expected = current.saturating_add(1);
```
Fraser999, line 731: I'm not sure how we handle assigning a height when creating a new block if the latest is at height `u32::MAX`.  So `saturating_add` might be the wrong call here?  It's possibly better to do `checked_add().unwrap()` if we don't expect to ever reach that height?

crates/astria-sequencer/src/bridge/state_ext.rs
```diff
@@ -243,7 +243,7 @@ pub(crate) trait StateWriteExt: StateWrite {
     #[instrument(skip(self))]
     async fn put_deposit_event(&mut self, deposit: Deposit) -> Result<()> {
         let nonce = self.get_deposit_nonce(deposit.rollup_id()).await?;
-        self.put_deposit_nonce(deposit.rollup_id(), nonce + 1);
+        self.put_deposit_nonce(deposit.rollup_id(), nonce.wrapping_add(1));
```
Fraser999, line 246: Would this too maybe be better as a `checked_add().unwrap()`?

crates/astria-composer/src/executor/mod.rs
```diff
@@ -386,9 +390,14 @@ impl Executor {
...
 
+        let number_of_submitted_bundles = if let Some(value) = bundles_drained {
+            value.to_string()
+        } else {
+            format!("more than {}", u64::MAX)
+        };
         if bundles_to_drain.is_empty() {
             info!(
-                number_of_submitted_bundles = bundles_drained,
+                %number_of_submitted_bundles,
```
SuperFluffy, line 400: I think `Value for String`, so `%` can be omitted

crates/astria-composer/src/executor/mod.rs
```diff
@@ -598,7 +607,7 @@ impl Future for SubmitFut {
                             metrics::histogram!(crate::metrics_init::TRANSACTIONS_PER_SUBMISSION)
                                 .record(this.bundle.actions_count() as f64);
 
-                            return Poll::Ready(Ok(*this.nonce + 1));
+                            return Poll::Ready(Ok(this.nonce.wrapping_add(1)));
```
SuperFluffy, line 610: Hm, if we ever run into this particular issue then things are going extremely badly. Maybe even just panic here?

crates/astria-conductor/src/block_cache.rs
```diff
@@ -49,7 +49,10 @@ impl<T> BlockCache<T> {
     /// Returns the next sequential block if it exists in the cache.
     pub(crate) fn pop(&mut self) -> Option<T> {
         let block = self.inner.remove(&self.next_height)?;
-        self.next_height += 1;
+        self.next_height = self
+            .next_height
+            .checked_add(1)
+            .expect("block height should not be `u64::MAX`");
```
SuperFluffy, line 55:
```suggestion
            .expect("block height must not exceed `u64::MAX`");
```

crates/astria-conductor/src/executor/mod.rs
SuperFluffy: IMO add an a variant to `ContractViolation` (i.e. make it an enum) for this case, so `current.checked_add(1).ok_or(...)`

crates/astria-sequencer-relayer/src/relayer/write/conversion.rs
```diff
@@ -159,6 +180,8 @@ pub(super) enum TryIntoPayloadError {
          could not be constructed"
     )]
     NoSequencerNamespacePresent,
+    #[error("the payload is too large")]
+    PayloadSize,
```
SuperFluffy, line 184: This could give the impression that there is a defined and reasonable upper limit somewhere, leaving readers guessing:
```suggestion
    #[error("the payload size exceeded u64::MAX bytes")]
    PayloadSizeOverflowed,
```

crates/astria-sequencer/src/app/mod.rs
```diff
@@ -602,7 +602,7 @@ impl App {
         txs: Vec<SignedTransaction>,
         block_size_constraints: &mut BlockSizeConstraints,
     ) -> anyhow::Result<()> {
-        let mut excluded_tx_count = 0;
+        let mut excluded_tx_count = 0.0_f64;
```
SuperFluffy, line 605: Extremely nit
```suggestion
        let mut excluded_tx_count = 0_f64;
```

crates/astria-sequencer/src/bridge/state_ext.rs
SuperFluffy: Maybe not a panic, but I think this should definitely be a `checked_add` because if nonces exceed that we have a problem.
```suggestion
        self.put_deposit_nonce(
            deposit.rollup_id(),
            nonce.checked_add(1).context("nonce overflowed")?
        );
```

crates/astria-sequencer/src/mempool.rs
```diff
@@ -20,7 +20,9 @@ pub(crate) struct TransactionPriority {
 
 impl TransactionPriority {
     fn nonce_diff(&self) -> u32 {
```
SuperFluffy, line 22: Not part of this PR, but could you move this type's constructor (`TransactionPriority::new`) into this impl block? First thing I did was look for it and then realized it's so far down.

crates/astria-sequencer/src/mempool.rs
```diff
@@ -20,7 +20,9 @@ pub(crate) struct TransactionPriority {
...
-        self.transaction_nonce - self.current_account_nonce
+        self.transaction_nonce
+            .checked_sub(self.current_account_nonce)
+            .expect("transaction nonce should not be less than current account nonce")
```
SuperFluffy, line 25:
```suggestion
            .expect("invariant violated: transaction nonce should not be less than current account nonce")
```

crates/astria-sequencer/src/mint/action.rs
```diff
@@ -47,8 +48,11 @@ impl ActionHandler for MintAction {
             .get_account_balance(self.to, native_asset)
             .await
             .context("failed getting `to` account balance")?;
+        let new_balance = to_balance
+            .checked_add(self.amount)
+            .ok_or_else(|| anyhow!("`to` balance would overflow"))?;
```
SuperFluffy, line 53: `anyhow` (different from `eyre`) implements `Context for Option<T>`:
```suggestion
            context("`to` balance would overflow")?;
```

crates/astria-sequencer/src/proposal/commitment.rs
```diff
@@ -27,7 +27,7 @@ impl GeneratedCommitments {
     /// which can be used as the block's transactions.
     #[must_use]
     pub(crate) fn into_transactions(self, mut tx_data: Vec<Bytes>) -> Vec<Bytes> {
-        let mut txs = Vec::with_capacity(tx_data.len() + 2);
```
SuperFluffy, line 30: This is an optimization I think. I think it's reasonable to `tx_data.len().saturating_add(2)` here. If this ends up too small then the `Vec::push` calls below will panic (not that this will ever be a problem because we won't have enough memory...)

crates/astria-composer/src/executor/bundle_factory/mod.rs
```diff
@@ -87,16 +87,21 @@ impl SizedBundle {
             return Err(SizedBundleError::SequenceActionTooLarge(seq_action));
         }
 
-        if self.curr_size + seq_action_size > self.max_size {
+        let new_size = self
+            .curr_size
+            .checked_add(seq_action_size)
```
SuperFluffy, line 92: IMO this one extra test here is redundant and we can just do a `saturating_add` here: if `new_size = u64::MAX` and `self.max_size = u64::MAX`, then `new_size > self.max_size == false` always.

SuperFluffy (approved):
Thank you so much for having dug through the codebase and added these everywhere.

I left a few comments here and there, but I think I mostly agree with the choice of panics vs errors. I tried to tighten the error messages or panics here and there to make them more explicit. Maybe have another pass over them to make sure a reader isn't left guessing?

crates/astria-composer/src/executor/mod.rs
Fraser999: Strings are `Debug` formatted by default, so we need the `%` to force `Display`.

crates/astria-sequencer/src/mempool.rs
Fraser999: I have removed this constructor in #1092, so it won't be an issue soon :)

crates/astria-conductor/src/executor/mod.rs
Fraser999: Done in [44c99af](https://github.com/astriaorg/astria/pull/1081/commits/44c99affc52d66ee3134fa1bbc23f66d21bb632a).

crates/astria-sequencer/src/bridge/state_ext.rs
Fraser999: Done in [44c99af](https://github.com/astriaorg/astria/pull/1081/commits/44c99affc52d66ee3134fa1bbc23f66d21bb632a).

crates/astria-composer/src/executor/mod.rs
Fraser999: Done in [44c99af](https://github.com/astriaorg/astria/pull/1081/commits/44c99affc52d66ee3134fa1bbc23f66d21bb632a).

crates/astria-conductor/src/block_cache.rs
Fraser999: Done in [44c99af](https://github.com/astriaorg/astria/pull/1081/commits/44c99affc52d66ee3134fa1bbc23f66d21bb632a).

crates/astria-sequencer/src/app/mod.rs
Fraser999: Done in [44c99af](https://github.com/astriaorg/astria/pull/1081/commits/44c99affc52d66ee3134fa1bbc23f66d21bb632a).

crates/astria-sequencer/src/mempool.rs
Fraser999: Done in [44c99af](https://github.com/astriaorg/astria/pull/1081/commits/44c99affc52d66ee3134fa1bbc23f66d21bb632a).

crates/astria-sequencer/src/mint/action.rs
Fraser999: Done in [44c99af](https://github.com/astriaorg/astria/pull/1081/commits/44c99affc52d66ee3134fa1bbc23f66d21bb632a).

crates/astria-sequencer/src/proposal/commitment.rs
Fraser999: Done in [44c99af](https://github.com/astriaorg/astria/pull/1081/commits/44c99affc52d66ee3134fa1bbc23f66d21bb632a).

crates/astria-composer/src/executor/bundle_factory/mod.rs
Fraser999: Good point!  Done in [44c99af](https://github.com/astriaorg/astria/pull/1081/commits/44c99affc52d66ee3134fa1bbc23f66d21bb632a).

joroshiba (approved):
approved on infra side

PR #1082, opened 2024-05-17
feat(sequencer): update sequencer API to contain methods required by conductor/relayer

### LINKED ISSUE #957: update sequencer API to contain methods required by conductor/relayer
conductor and relayer currently rely on both cometbft and the sequencer gRPC API, which is bad, as we only want to rely on the sequencer gRPC API. we should update the sequencer API to contain the methods or similar functionality that's currently provided by cometbft to conductor/relayer. (then in another issue, actually remove the dependence on cometbft from conductor/relayer).

noot:
## Summary
update sequencer API to contain methods required by conductor/relayer

## Background
ideally conductor/relayer only need the sequencer gRPC API endpoint, not the cometbft one as well. this also allows us to eventually have the sequencer API be a full proxy of all the cometbft endpoints.

## Changes
* implement `get_chain_id`, `get_abci_info`, `get_commit` and `get_validators` in the sequencer gRPC service 
* update sequencer to know the cometbft rpc addr, as it uses the cometbft JSON-RPC API to fetch commit/validators for `get_commit` and `get_validators` respectively. it needs this as it doesn't store the commit or full validator info in its own state

## Testing
manual testing so far - unit tests/blackbox tests would require a cometbft endpoint or mock cometbft JSON-RPC server

## Related Issues
closes #957


proto/sequencerblockapis/astria/sequencerblock/v1alpha1/service.proto
```diff
@@ -4,6 +4,9 @@ package astria.sequencerblock.v1alpha1;
 
 import "astria/primitive/v1/types.proto";
 import "astria/sequencerblock/v1alpha1/block.proto";
+import "astria_vendored_sequencerblock/tendermint/abci/types.proto";
+import "astria_vendored_sequencerblock/tendermint/types/types.proto";
+import "astria_vendored_sequencerblock/tendermint/types/validator.proto";
```
joroshiba, line 9: is there any reason these can't be named the same as they are in protocol `astria_vendored`? Does this create clashes?

charts/sequencer/templates/configmaps.yaml
```diff
@@ -60,6 +60,7 @@ data:
   ASTRIA_SEQUENCER_ENABLE_MINT: "false"
   # Socket address for gRPC server
   ASTRIA_SEQUENCER_GRPC_ADDR: "0.0.0.0:{{ .Values.ports.sequencerGRPC }}"
+  ASTRIA_SEQUENCER_COMETBFT_RPC_ADDR: "http://127.0.0.1:{{ .Values.ports.cometBFTRPC }}"
```
joroshiba, line 63: nit: add to the `else` section of `if not .Values.global.dev` below instead of here, makes it clear we have breaking changes when doing a new main release.

crates/astria-sequencer/src/grpc/sequencer.rs
```diff
@@ -25,13 +39,33 @@ use crate::{
...
+
+    async fn validate_height(&self, request_height: u64) -> Result<(), Status> {
+        let snapshot = self.storage.latest_snapshot();
+        let curr_block_height = snapshot.get_block_height().await.map_err(|e| {
+            Status::internal(format!("failed to get block height from storage: {e}"))
+        })?;
+
+        if curr_block_height < request_height {
+            return Err(Status::invalid_argument(
+                "requested height is greater than current block height",
```
Fraser999, line 64: Should we also include the current height in the error response?

crates/astria-sequencer/src/grpc/sequencer.rs
```diff
@@ -152,6 +168,102 @@ impl SequencerService for SequencerServer {
...
+        let height: u32 = request
+            .height
+            .try_into()
+            .map_err(|_| Status::invalid_argument("height is too large to fit into an u32"))?;
+
+        let resp = self
+            .cometbft_client
+            .validators(height, Paging::Default)
+            .await
+            .map_err(|e| Status::internal(format!("failed to get commit from cometbft: {e}")))?;
```
Fraser999, line 232:
```suggestion
            .map_err(|e| Status::internal(format!("failed to get validators from cometbft: {e}")))?;
```

crates/astria-sequencer/src/grpc/sequencer.rs
```diff
@@ -152,6 +168,102 @@ impl SequencerService for SequencerServer {
...
+        self.validate_height(request.height).await?;
+
+        let height: u32 = request
+            .height
+            .try_into()
+            .map_err(|_| Status::invalid_argument("height is too large to fit into an u32"))?;
+
+        let resp = self
+            .cometbft_client
+            .validators(height, Paging::Default)
```
Fraser999, line 230: Should we be requesting `Paging::All` here, or maybe we need to implement paging on our server too?

crates/astria-sequencer/src/grpc/sequencer.rs
```diff
@@ -152,6 +168,102 @@ impl SequencerService for SequencerServer {
...
+            .root_hash()
+            .await
+            .map_err(|e| Status::internal(format!("failed to get app hash from storage: {e}")))?;
+
+        let response = ResponseInfo {
+            version: env!("CARGO_PKG_VERSION").to_string(),
+            app_version: 1,
+            last_block_height: u32::try_from(block_height)
+                .expect("block height must fit into u32")
+                .into(),
```
Fraser999, line 260:
```suggestion
            last_block_height: i64::try_from(block_height).expect("block height must fit into i64"),
```

SuperFluffy:
High level feedback on the quasi-cometbft RPCs here (namely `get_chain_id`, `get_abci_info`, `get_commit` and `get_validators`): I think we should have a distinct gRPC services that resemble those provided by cometbft. For example, for ABCI they are defined here: https://github.com/cometbft/cometbft/blob/main/proto/cometbft/abci/v1/service.proto

We don't need to support all of them for now, only the subset needed. But because we are in that ecosystem (and following the principle of least surprise), it makes sense to just follow their example.

crates/astria-sequencer/local.env.example
```diff
@@ -12,6 +12,11 @@ ASTRIA_SEQUENCER_ENABLE_MINT=false
 
 # Socket address for gRPC server
 ASTRIA_SEQUENCER_GRPC_ADDR="127.0.0.1:8080"
+
+# The address of the cometbft RPC server.
+# 127.0.0.1:26657 is the default socket address in comebft's `rpc.laddr` setting.
```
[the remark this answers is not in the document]
joroshiba: I hate that this is technically correct: https://github.com/astriaorg/astria/blob/main/charts/sequencer/files/cometbft/config/config.toml#L90

noot:
closing due to staleness, but we should probably do this

PR #1086, opened 2024-05-20
implement AstriaWithdrawer contract and ethereum watcher

noot:
## Summary
implement `ethereum` module for astria-bridge-withdrawer.

## Background
required to read from the astria withdrawal evm contract.

## Changes
- implement `AstriaWithdrawer.sol` contract, put inside `ethereum/` dir as a foundry project
- implement `ethereum` module which contains a `Watcher` that watches the contract for `Withdrawal` events
- the events are sent to a `Batcher` which batches all events by block number and sends a batch of events to a sequencer handler (to be implemented later)

## Testing
unit tests

## Related Issues

 #913


crates/astria-bridge-withdrawer/ethereum/out/AstriaWithdrawer.sol/AstriaWithdrawer.json
```diff
@@ -0,0 +1 @@
...e":"EmitStatement","src":"257:44:0"}]},"functionSelector":"0968f264","implemented":true,"kind":"function","modifiers":[],"name":"withdraw","nameLocation":"200:8:0","parameters":{"id":12,"nodeType":"ParameterList","parameters":[{"constant":false,"id":11,"mutability":"mutable","name":"memo","nameLocation":"224:4:0","nodeType":"VariableDeclaration","scope":23,"src":"209:19:0","stateVariable":false,"storageLocation":"calldata","typeDescriptions":{"typeIdentifier":"t_bytes_calldata_ptr","typeString":"bytes"},"typeName":{"id":10,"name":"bytes","nodeType":"ElementaryTypeName","src":"209:5:0","typeDescriptions":{"typeIdentifier":"t_bytes_storage_ptr","typeString":"bytes"}},"visibility":"internal"}],"src":"208:21:0"},"returnParameters":{"id":13,"nodeType":"ParameterList","parameters":[],"src":"247:0:0"},"scope":24,"stateMutability":"payable","virtual":false,"visibility":"external"}],"abstract":false,"baseContracts":[],"canonicalName":"AstriaWithdrawer","contractDependencies":[],"contractKind":"contract","fullyImplemented":true,"linearizedBaseContracts":[24],"name":"AstriaWithdrawer","nameLocation":"81:16:0","scope":25,"usedErrors":[],"usedEvents":[9]}],"license":"MIT or Apache-2.0"},"id":0}
```
noot, line 1: i committed this because otherwise we have to install foundry in ci which was painful last time - however this will need to be updated if the contract changes. thoughts?

itamarreif (approved):
Only thing that sticks out to me is back pressure:
- both the watcher->batcher and the batcher->submitter channels have capacity of 100 
- both use `send()` instead of `send_timeout()`
- submitter logic will use `broadcast_tx_commit` so might take a bit of time in the event of a congested mempool on the sequencer side
- if withdraws for 100 entire blocks clog up the batcher->submitter channel, the next 100 individual withdraws will clog up the watcher->batcher channel

doesn't seem like too big of a problem if we choose the right values for channel sizes (100 is probably fine) but figured it would be useful to flag and maybe clarify somewhere in the documentation

crates/astria-bridge-withdrawer/ethereum/out/AstriaWithdrawer.sol/AstriaWithdrawer.json
itamarreif: i agree - its nicer to have these committed rather than having to deal with with foundry every time.

joroshiba: Absolutely agreed here

crates/astria-bridge-withdrawer/ethereum/README.md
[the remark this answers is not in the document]
noot: i think `forge build` should automatically do it

crates/astria-bridge-withdrawer/ethereum/lib/forge-std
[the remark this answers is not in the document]
noot: yeah, this is just how foundry is set up, `forge build` should install everything. we don't use the forge-std deps right now but if we want to add solidity unit tests we will

noot:
> doesn't seem like too big of a problem if we choose the right values for channel sizes (100 is probably fine) but figured it would be useful to flag and maybe clarify somewhere in the documentation

yeah this is pretty arbitrary, i just picked a random number basically. i suppose the watcher->batcher channel will probably have more items than the batcher->submitter, so maybe the value for that should be higher. i can add a comment for now saying the values are arbitrary

> both use send() instead of send_timeout()

i think `send_timeout` would be kinda bad, since we want to guarantee 1 eth block of withdrawals -> 1 sequencer tx without missing any, if there's timeouts there's a chance a tx within the block could be missed? i'd prefer that if anything fails to send, we just error at this point. not against timeouts but depends how we implement tracking txs that have been submitted

PR #1089, opened 2024-05-21
chore(sequencer): remove unneeded allow attribute

SuperFluffy:
## Summary
Removes an unneeded clippy allow attribute. 

## Background
`#[allow(clippy::non_canonical_partial_ord_impl)]` is not doing anything because `PartialOrd` is in fact implemented in terms of `Ord`.


SuperFluffy:
Superceded by #1092

PR #1103, opened 2024-05-23
feat(cli): update for new charts, dusk-7 endpoints

joroshiba:
## Summary
Updates rust cli to use dusk-7. This includes updates to use latest charts + sequencer updates. One additional update of removing unneeded chain id from basic args.

## Background
Dusk-6 is being released, and cli needs to be updated to work with it

## Changes
- Add default resources which are on low end for chart
- Update passing of private keys to match new chart setup
- Utilize newest chart release
- Default to dusk-7 network
- Remove the unused chain id from basic account args.

## Testing
Manual testing + some CI updates


crates/astria-cli/src/types.rs
```diff
@@ -251,6 +255,81 @@ struct SequencerConfig {
...
+#[derive(Debug, PartialEq, Serialize, Deserialize)]
+struct Resource {
+    cpu: f32,
+    memory: String,
+}
+
+impl Resource {
+    fn default_low_request() -> Self {
+        Self {
+            cpu: 0.01,
```
Fraser999, line 306: We could maybe use the milliCPU form for these - the note section in [the docs](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-cpu) recommends that to help avoid providing an invalid value in the future.

crates/astria-cli/src/cli/rollup.rs
```diff
@@ -8,8 +8,8 @@ use color_eyre::eyre;
 use serde::Serialize;
 
 const DEFAULT_ROLLUP_CHART_PATH: &str =
-    "https://github.com/astriaorg/charts/releases/download/evm-rollup-0.11.1/evm-rollup-0.11.1.tgz";
-const DEFAULT_SEQUENCER_GRPC: &str = "https://grpc.sequencer.dusk-4.devnet.astria.org/";
+    "https://github.com/astriaorg/charts/releases/download/evm-rollup-0.18.1/evm-rollup-0.18.1.tgz";
```
Fraser999, line 11: We're at `evm-rollup-0.18.4` now - not sure if that matters here or not?

Fraser999 (approved):
I don't have a handle on whether the default values for `CoreResources` are good for our use case or not, but they _seem_ fine.

I guess these don't need to be configurable via the CLI?

Otherwise, this LGTM.

PR #1104, opened 2024-05-23
feat(sequencer)!: fees go to sudo poa

joroshiba:
## Summary
Gives block fees to POA authority, and updates some snapshot tests.

## Background
During POA phase of network, we want network fees to go to the authority account. This allows use to rebate fees to proposers while figuring out economics. 

## Changes
- Fees to sudo instead of proposer.
- Update default genesis for tests. On first pass the snapshot tests where not failing because the default proposer and the default sudo address were the same (just a zero address). Updated default genesis so this should fail in the future. If payee changes.
- Fix test for payee to now be a test for fees going to sudo address.

## Testing
CI. Updated snapshot test which was not working, manually verified that changing back to proposer as fee recipient will cause test failure.

## Breaking Changelist
- Changes receiver of fees for all blocks.


crates/astria-sequencer/src/app/mod.rs
```diff
@@ -1060,14 +1060,14 @@ impl App {
...
-                .get_account_balance(proposer_address, asset)
+                .get_account_balance(fee_recipient, asset)
                 .await
                 .context("failed to get proposer account balance")?;
             let new_balance = balance
                 .checked_add(amount)
                 .context("account balance overflowed u128")?;
             state_tx
-                .put_account_balance(proposer_address, asset, new_balance)
+                .put_account_balance(fee_recipient, asset, new_balance)
```
noot, line 1070: i think you can change these to use the `increase_balance` function

crates/astria-sequencer/src/app/test_utils.rs
```diff
@@ -101,8 +103,8 @@ pub(crate) async fn initialize_app_with_storage(
 
     let genesis_state = genesis_state.unwrap_or_else(|| GenesisState {
         accounts: default_genesis_accounts(),
-        authority_sudo_address: Address::from([0; 20]),
-        ibc_sudo_address: Address::from([0; 20]),
+        authority_sudo_address: address_from_hex_string(JUDY_ADDRESS),
+        ibc_sudo_address: address_from_hex_string(TED_ADDRESS),
```
joroshiba, line 107: I added addresses in here that were different because changing the fee recipient was NOT causing insta tests to fail. The proposer & the authority address where both the nil addresses.

noot (approved):

crates/astria-sequencer/src/app/test_utils.rs
noot: yeah this is one of those things snapshots wouldn't necessarily catch since it's not modfying the logic of any action, this is a good add

crates/astria-sequencer/src/app/tests_breaking_changes.rs
```diff
@@ -138,6 +138,7 @@ async fn app_finalize_block_snapshot() {
     app.finalize_block(finalize_block.clone(), storage.clone())
         .await
         .unwrap();
+    app.commit(storage.clone()).await;
```
joroshiba, line 141: the app hash is not updated until commit, we were just comparing the genesis app hashes lol.

crates/astria-sequencer/src/app/tests_breaking_changes.rs
```diff
@@ -119,7 +119,7 @@ async fn app_finalize_block_snapshot() {
     let deposits = HashMap::from_iter(vec![(rollup_id, vec![expected_deposit.clone()])]);
     let commitments = generate_rollup_datas_commitment(&[signed_tx.clone()], deposits.clone());
 
-    let timestamp = Time::now();
+    let timestamp = Time::unix_epoch();
```
joroshiba, line 122: timestamp was causing non-determinism in test

crates/astria-sequencer/src/app/mod.rs
```diff
@@ -747,13 +747,15 @@ impl App {
             .get_chain_id()
             .await
             .context("failed to get chain ID from state")?;
+        let sudo_address = self
```
noot, line 750: you can move this to the `end_block` call so you don't have to pass it as a parameter. `state_tx` implements StateRead

noot (approved):

crates/astria-sequencer/src/app/mod.rs
joroshiba: I figured it would be better if it failed earlier than later (shouldn't fail), and this provides the ability to add fork choice logic elsewhere. So switching back to proposer is a couple of lines of code here without modifying the end_block function signature.

PR #1113, opened 2024-05-24
fix(cli): fix tx submission result check

noot:
## Summary
previously, the command was not erroring if `check_tx` failed as only the `deliver_tx` response was checked. this PR fixes that and now checks both responses

## Changes
- check `check_tx` result 

## Testing
i ran it manually with txs that fail check_tx.


Fraser999 (approved):

crates/astria-cli/src/commands/sequencer.rs
```diff
@@ -501,6 +479,10 @@ async fn submit_transaction(
     let sequencer_key = SigningKey::from(private_key_bytes);
 
     let from_address = Address::from_verification_key(sequencer_key.verification_key());
+    println!(
+        "sending tx from address: {}",
+        hex::encode(from_address.to_vec())
+    );
```
SuperFluffy, line 485:
```suggestion
    println!("sending tx from address: {from_address}");
```
address implements `Display`

PR #1114, opened 2024-05-26
chore(charts): update evm chart to have better default request limits

joroshiba:
## Summary
Chart limits for conductor & composer were too low, and we have a new geth image with less noisy logs


WafflesVonMaple (approved):

WafflesVonMaple:
Lint nonsense but stamped

PR #1115, opened 2024-05-27
feat(conductor): add retry to execution API gRPCs

SuperFluffy:
## Summary
This adds retry logic to all execution API gRPCs from Conductor to the rollup.

## Background
If there was a hickup when communicating with the rollup, conductor would fail immediately. This adds retry logic all execution API calls. Because this cannot yet be done at the transport level due to type-level constraints on tonic and tower, this patch adds retries at the application leve.

## Changes
- Add retry logic to every execution v1alpha2 API call
- Introduce a `ExecutionApiRetryStrategy` implementing `tryhard::BackoffStrategy`
- The following gRPC tonic status codes will be retried:
    - Cancelled
    - Unknown
    - DeadlineExceeded
    - NotFound
    - ResourceExhausted
    - Aborted
    - Unavailable
    - DataLoss

## Testing
Add unit tests for the all codes. The happy-path blackbox tests are unaffected.

## Related Issues

closes https://github.com/astriaorg/astria/issues/624

### DIFF
--- a/crates/astria-conductor/src/executor/client.rs
+++ b/crates/astria-conductor/src/executor/client.rs
@@ -1,3 +1,5 @@
+use std::time::Duration;
+
 use astria_core::{
     execution::v1alpha2::{
         Block,
@@ -25,7 +27,12 @@ use tonic::transport::{
     Endpoint,
     Uri,
 };
-use tracing::instrument;
+use tracing::{
+    instrument,
+    warn,
+    Instrument,
+    Span,
+};
 
 /// A newtype wrapper around [`ExecutionServiceClient`] to work with
 /// idiomatic types.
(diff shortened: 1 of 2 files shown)


crates/astria-conductor/src/executor/client.rs
```diff
@@ -168,3 +216,172 @@ fn block_identifier(number: u32) -> raw::BlockIdentifier {
...
+            "failed executing RPC; retrying after after backoff"
+        );
+        futures::future::ready(())
+    }
+}
+
+fn retry_config()
+-> tryhard::RetryFutureConfig<tryhard::backoff_strategies::ExponentialBackoff, OnRetry> {
+    tryhard::RetryFutureConfig::new(u32::MAX)
+        .exponential_backoff(Duration::from_millis(100))
```
Fraser999, line 250: Should this be using the new `ExecutionApiRetryStrategy`?
```suggestion
fn retry_config() -> tryhard::RetryFutureConfig<ExecutionApiRetryStrategy, OnRetry> {
    tryhard::RetryFutureConfig::new(u32::MAX)
        .custom_backoff(ExecutionApiRetryStrategy {
            delay: Duration::from_millis(100),
        })
```

crates/astria-conductor/src/executor/client.rs
```diff
@@ -168,3 +216,172 @@ fn block_identifier(number: u32) -> raw::BlockIdentifier {
...
+    matches!(
+        status.code(),
+        Code::Cancelled
+            | Code::Unknown
+            | Code::DeadlineExceeded
+            | Code::NotFound
+            | Code::ResourceExhausted
+            | Code::Aborted
+            | Code::Unavailable
+            | Code::DataLoss
```
Fraser999, line 324: Not sure if this is worth retrying?  The docs say "Unrecoverable data loss or corruption" which sounds fairly terminal to me.

crates/astria-conductor/src/executor/client.rs
```diff
@@ -168,3 +216,172 @@ fn block_identifier(number: u32) -> raw::BlockIdentifier {
...
+    // Code::FailedPrecondition => no, failed precondition should not be retried unless the
+    //                             precondition is fixed, see [1]
+    // Code::Aborted => yes, although this applies to a read-modify-write sequence. We should
+    //                  implement this not per-request but per-request-sequence (for example,
+    //                  execute + update-commitment-state)
+    // Code::OutOfRange => no, we don't expect to send any out-of-range requests.
+    // Code::Unimplemented => no, no point retrying
+    // Code::Internal => no, this is a serious error on the backend; don't retry
+    // Code::Unavailable => yes, retry after backoff is desired
+    // Code::DataLoss => yes, unclear how this would happen; err on retrying
```
Fraser999, line 313: I initially thought this meant only retry once and return an err when retrying if we get the same error again! :D  Maybe it could be phrased "err on the side of retrying"?

crates/astria-conductor/src/executor/client.rs
SuperFluffy: Yeah, You saw my comment in the code below. I was not sure about this. But maybe you are right, I have removed it from the match.  40a9f51

crates/astria-conductor/src/executor/client.rs
SuperFluffy: Switched it to `=> no` in  40a9f51 :)

Fraser999 (approved):

PR #1120, opened 2024-05-29
feat(conductor): skip outdated block metadata

SuperFluffy:
## Summary
Conductor drops sequencer metadata items retrieved from Celestia if their height has already been executed/committed.

## Background
Conductor would try and unnecessarily verify Sequencer metadata retrieved from Celestia, putting a lot of pressure on the Sequencer node. Because Conductor is stateless, it fetches all Celestia blobs for the sequencer and rollup namespaces starting from the `celestia_base_block_height` configured the rollup. It would then verify all sequencer metadata items retrieved in this way against the Sequencer, even if the metadata was for old blocks (meaning blocks with a height below the next firm height expected by the rollup).

## Changes
- Check the Sequencer height of each sequencer metadata item received from Celestia against the next expected firm height; drop the item without verification if its less than the expected height

## Testing
- Cannot yet be tested programatically, but the blackbox tests have been updated to contain several sequencer blocks per Celestia blob. Running the following one can observe a specific line indicating that the logic works:
```
TEST_LOG=1 RUST_LOG=debug cargo test firm_only::skips_already_executed_heights
...
2024-05-29T13:00:48.704019Z  INFO execute:verify_metadata: astria_conductor::celestia::verify: dropping Sequencer metadata item without verifying against Sequencer because its height is below the next expected firm height next_expected_firm_sequencer_height=7 sequencer_height_in_metadata=6 celestia_height=1 rollup_namespace=AAAAAAAAAAAAAAAAAAAAAAAAACoqKioqKioqKio= sequencer_namespace=AAAAAAAAAAAAAAAAAAAAAAAAAKDcd7eO77LbDQs=
```

### DIFF
--- a/crates/astria-conductor/src/celestia/verify.rs
+++ b/crates/astria-conductor/src/celestia/verify.rs
@@ -50,6 +50,10 @@ use super::{
     block_verifier,
     convert::ConvertedBlobs,
 };
+use crate::executor::{
+    self,
+    StateIsInit,
+};
 
 pub(super) struct VerifiedBlobs {
     celestia_height: u64,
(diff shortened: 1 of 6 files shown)


crates/astria-conductor/tests/blackbox/helpers/mod.rs
```diff
@@ -473,7 +473,18 @@ fn make_config() -> Config {
 
 #[must_use]
 pub fn make_sequencer_block(height: u32) -> astria_core::sequencerblock::v1alpha1::SequencerBlock {
+    fn repeat_bytes_of_u32_as_array(val: u32) -> [u8; 32] {
```
SuperFluffy, line 476: We need a unique block hash for each sequencer height.

joroshiba (approved):

crates/astria-conductor/src/celestia/verify.rs
```diff
@@ -94,24 +98,37 @@ struct VerificationTaskKey {
...
-                .verify_metadata(blob)
-                .in_current_span(),
-        );
+        if blob.height().value() < next_expected_firm_sequencer_height {
+            info!(
+                next_expected_firm_sequencer_height,
+                sequencer_height_in_metadata = blob.height().value(),
+                "dropping Sequencer metadata item without verifying against Sequencer because its \
+                 height is below the next expected firm height"
+            );
```
noot, line 118: is this guaranteed safe? ie are these blobs already guaranteed to be executed/verified as soft blocks already and we're sure that they are the same as the ones already verified?

SuperFluffy: Note that this is checking against the next expected *firm* height - this is independent of the soft blocks.

PR #1121, opened 2024-05-29
feat(conductor, proto)!: celestia base heights in commitment state

joroshiba:
## Summary
Updates the execution api to have celestia base height as a part of commitment instead of genesis, requiring rollups to track the base celestia height as it increases. This enables conductor on resync to move much faster.

https://github.com/astriaorg/astria-geth/pull/22 contains an implementation which works with this PR.

## Background
Conductor resync needs to grab all data from genesis for the rollup currently, even if much of that data is previously executed.

## Changes
- Change execution API to have celestia base height on commitment
- Update conductor logic to update celestia base height to that of most recent firm commitment found in.

## Testing
Manually tested running linked PR in dev cluster, updated blackbox tests.

## Breaking Changelist
- Execution API breaking change, rollout will require a new image for any rollup utilizing.

### DIFF
--- a/crates/astria-conductor/src/executor/mod.rs
+++ b/crates/astria-conductor/src/executor/mod.rs
@@ -456,11 +456,10 @@ impl Executor {
         block.height = block.sequencer_height().value(),
     ))]
     async fn execute_firm(&mut self, block: ReconstructedBlock) -> eyre::Result<()> {
-        let celestia_height =
-            block.celestia_height.try_into().expect(
-                "block height overflow, this should not happen since tendermint heights are i64 \
-                 and never negative under the hood",
-            );
+        let celestia_height = block.celestia_height.try_into().expect(
+            "block height overflow, this should not happen since tendermint heights are i64 and \
+             never negative under the hood",
+        );
         let executable_block = ExecutableBlock::from_reconstructed(block);
         let expected_height = self.state.next_expected_firm_sequencer_height();
         let block_height = executable_block.height;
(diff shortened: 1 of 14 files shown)


SuperFluffy (approved):
Looks good. This is not strictly necessary, but I think I would like to set the first Celestia heright that is fetched on startup to `celestia_base_height.saturating_sub(celestia_variance)`.

Also, given that this is a breaking change anyway, can you also bump the celestia base height (and variance) to uint32/u64? It turns out that when interacting with most of their RPCs u64 is the expected width, and not u32.

joroshiba:
> Looks good. This is not strictly necessary, but I think I would like to set the first Celestia heright that is fetched on startup to celestia_base_height.saturating_sub(celestia_variance).

@SuperFluffy This is intentionally not the behavior here. The base is the lowest height to look, the variance defines how far above you can go. Long term we need better logic in conductor on setting the base height. Right now this is acceptable because we never post out of order.

quasystaty1 (approved):

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -59,6 +58,8 @@ use state::StateReceiver;
 
 use self::state::StateSender;
 
+type CelestiaHeight = u64;
```
joroshiba, line 61: note: doing this because it makes the update enums much more clear or what is being passed in.

proto/executionapis/astria/execution/v1alpha2/execution.proto
```diff
@@ -86,6 +84,8 @@ message CommitmentState {
   Block soft = 1;
   // Firm commitment is achieved when data has been seen in DA.
   Block firm = 2;
+  // The lowest block number of celestia chain to be searched for rollup blocks given current state
+  uint64 base_celestia_height = 3;
```
noot, line 88: is this equal to the celestia height the firm block was in? it seems so from the code but want to double check

noot (approved):

proto/executionapis/astria/execution/v1alpha2/execution.proto
joroshiba: Right now yes, long term not necessarily. This is ok right now because sequencer relayer posts such that astria blocks are in order in celestia blocks and we have a single relayer. Long term need to add more complex logic into conductor with timestamps and acceptable time smear.

Issue #943, opened 2024-04-10
change addresses to bech32


SuperFluffy:
Following cosmos conventions, we should addresses to bech32 format.

Because of the protobuf to json mapping this requires changing all protobuf messages containing addresses (of type `bytes` right now)  to be of type `string`.

PR #1124, opened 2024-05-30
feat(core, proto)!: add bech32m addresses

SuperFluffy:
## Summary
Adds bech32m addresses.

## Background
bech32m encoded addresses are the de-facto standard in cosmos. By using a `"astria"` prefix (in the human readable prefix, bech32 HRP sense) we align astria with the rest of the cosmos ecosystem.

## Changes
- Add a field `bech32m` to `astria.primitive.v1.Address`
- Deprecate `astria.primitive.v1.Address.inner`
- Change `std::fmt::Display for Address` to use bech32m encoding
- Serialize `Address` in terms of its protobuf-json mapping
- Update the `Address:try_from_raw` constructor to allow ingesting protobufs that have their `inner` or `bech32m` or populated. `inner` and `bech32m` are mutually exclusive, only one may be set.
- Protobufs with `inner` set are assumed to have prefix `"astria"`.
- Introduce an `AddressBuilder` type state builder that requires both a prefix and array/byte slice be set.
- Update all services to the new way of constructing addresses, but with the prefix `"astria"` set everywhere
- Add a `SignedTransaction::address` method that constructs an address from the verification key and the `chain_id` stored in the unsigned transaction's `transaction_params::chain_id`.
- Provide a builder for `TransactionParams` enforcing that chain IDs are have hrp compatible names.

## Testing
- update snapshot tests for address-as-json serialization
- add unit tests:
  - addresses with only bytes are accepted
  - addresses with only bech32m are accepted
  - addresses with both bytes and bech32m are are rejected
  - protobufs created from the common `Address` type only have `bech32m` populated
  - various tests to assert `Address` invariants (mainly maximum length of prefix + fixed-size address always being encodeable to a string)
  - updated all tests that somehow make use of addresses (mainly sequencer, bridge-withdrawer, composer)

## Breaking Changelist
This patch is marked breaking because the sequencer snapshot tests are triggered. They all contain actions that themselves contain protobuf address. This patch is backward but not forward compatible.

## Related Issues
closes https://github.com/astriaorg/astria/issues/943


crates/astria-core/src/primitive/v1/mod.rs
```diff
@@ -16,6 +19,14 @@ use crate::{
 };
 
 pub const ADDRESS_LEN: usize = 20;
+/// The human readable prefix of astria addresses (also known as bech32 HRP).
+pub const HUMAN_READABLE_ADDRESS_PREFIX: &str = "astria";
```
noot, line 23: for cosmos chains, this is configurable - would there be an easy way to configure this prefix based on genesis info for instance?

noot (approved):
looks good!

crates/astria-core/src/primitive/v1/mod.rs
SuperFluffy: Tough to do right now, punting.

crates/astria-sequencer/src/app/snapshots/astria_sequencer__app__tests_breaking_changes__app_execute_transaction_with_every_action_snapshot.snap
joroshiba, line 1: why are the app_hashes changing? I would expect the data hash to change but not the app_hash?

proto/primitives/astria/primitive/v1/types.proto
```diff
@@ -48,4 +48,6 @@ message RollupId {
 // using the first 20 bytes of the sha256 hash.
 message Address {
   bytes inner = 1;
+
+  string bech32m = 2;
```
joroshiba, line 52: Can we add some comments in here on the use within the protos?

crates/astria-sequencer/src/app/snapshots/astria_sequencer__app__tests_breaking_changes__app_execute_transaction_with_every_action_snapshot.snap
SuperFluffy: It contains a `SudoAddressChangeAction`, which itself contains `new_address: Address`. Hence the change.

proto/primitives/astria/primitive/v1/types.proto
SuperFluffy: Added comments on the fields.

SuperFluffy:
After the review comments I received from @joroshiba and @noot I went back to the drawing board and made the `Address` type prefix aware. This lead to some far reaching changes that I am listing here:

1. I have added a `deprecated = true` flag to the `astria::primitive::v1::Address::inner` field.
2. protobuf `Address`es with an `inner` field set are assumed to have `"astria"` as their prefix.
3. `Address::inner` and `Adress::bech32m` are mutually exclusive: only one may be set.
4. When going from the common `Address` type to the protobuf `Address` message type, only `bech32m` is populated, `inner` is left empty.

crates/astria-core/src/crypto.rs
```diff
@@ -147,28 +140,28 @@ impl VerificationKey {
     // Silence the clippy lint because the function body asserts that the panic
     // cannot happen.
     #[allow(clippy::missing_panics_doc)]
-    pub fn address(&self) -> &Address {
-        self.address.get_or_init(|| {
```
SuperFluffy, line 151: Removed the memoization for the time being. We can bring it back later.

crates/astria-core/src/primitive/v1/mod.rs
```diff
@@ -243,87 +247,238 @@ pub struct IncorrectRollupIdLength {
...
     fn as_ref(&self) -> &[u8] {
-        &self.0
+        &self.bytes
     }
 }
 
-impl From<[u8; ADDRESS_LEN]> for Address {
-    fn from(inner: [u8; ADDRESS_LEN]) -> Self {
-        Self(inner)
+// impl From<[u8; ADDRESS_LEN]> for Address {
```
noot, line 458: remove? i feel like it's kind of nice to have this though, using the astria prefix as default, but i get why you removed it

crates/astria-sequencer/src/app/mod.rs
```diff
@@ -961,7 +961,7 @@ impl App {
     /// Executes a signed transaction.
     #[instrument(name = "App::execute_transaction", skip_all, fields(
         signed_transaction_hash = %telemetry::display::base64(&signed_tx.sha256_of_proto_encoding()),
-        sender = %signed_tx.verification_key().address(),
+        sender = %telemetry::display::base64(&signed_tx.verification_key().address_bytes()),
```
noot, line 964: should this display as bech32?

noot, line 964: should this display as bech32?

crates/astria-core/src/primitive/v1/mod.rs
SuperFluffy: Ah, commented code strikes again.

I agree, it's nice to have it, but it's so easy to get wrong - and then out of a sudden you have the wrong prefix in the code.

We could provide utility constructors like `Address::astria_from_array` or something like that - but not in this PR.

crates/astria-sequencer/src/app/mod.rs
SuperFluffy: Can't right now because all we have is the sender (and hence the verification key), but no prefix, and hence no bech32m.

I am not sure how to address this at this point.

crates/astria-sequencer/src/mempool.rs
```diff
@@ -62,13 +62,16 @@ impl PartialOrd for TransactionPriority {
 pub(crate) struct EnqueuedTransaction {
     tx_hash: [u8; 32],
     signed_tx: Arc<SignedTransaction>,
+    address: Address,
```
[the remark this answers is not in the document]
SuperFluffy: So `signed_tx.verification_key().address()` was used somewhere in this mempool module - I would need to read the code again, but basically I have replaced all instances of `.address()` by this `Address`. Maybe that's actually unnecessary and the new `address_bytes()` would be sufficient.

PR #1125, opened 2024-05-30
fix(sequencer): unstick wrong sudo signer from mempool

Lilyjjo:
## Summary
Currently authority transactions that have the wrong sudo signer get stuck in the mempool. This is because the incorrectly signed transactions will fail to execute and will be removed from the `prepare_proposal()` block, will not be removed from the cometBFT mempool, and then re-added to the app's mempool during cometBFT's `CheckTx` flow. 

## Changes
- added to `CheckTx` flow verification of correct sudo signer for sudo actions

## Testing
Sending sudo txs via CLI

### DIFF
--- a/crates/astria-sequencer/src/app/mod.rs
+++ b/crates/astria-sequencer/src/app/mod.rs
@@ -538,7 +538,7 @@ impl App {
                 }
                 Err(e) => {
                     metrics::counter!(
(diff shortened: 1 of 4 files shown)


Lilyjjo:
Going with a different fix direction

PR #1129, opened 2024-05-31
feat: use macro to declare metric constants

SuperFluffy:
## Summary
Introduce a macro `declare_metric_const!` to avoid repetition and common typos when declaring string consts used as metric names.

## Background
Noticed some common errors in reviews and when writing metrics: mismatch between the const's name and the actual value. Missing underscores. Annoying repetiton of `env!("CARGO_CRATE_NAME")`;

## Changes
- Declare a macro `declare_metric_const!` and define all consts in terms of it.
- Add unit tests for all macros to ensure they don't accidentally break.
 
## Testing
- Unit tests for all const declaration
- Simple unit test for the macro

## Metrics
- `astria_composer_transactions_per_submission` replaces `astria_composer_transaction_per_submission`


crates/astria-composer/src/metrics_init.rs
```diff
@@ -73,31 +74,60 @@ pub fn register() {
...
+        SEQUENCER_SUBMISSION_FAILURE_COUNT,
+        SEQUENCER_SUBMISSION_LATENCY,
+        TRANSACTIONS_DROPPED,
+        TRANSACTIONS_DROPPED_TOO_LARGE,
+        TRANSACTIONS_PER_SUBMISSION,
+        TRANSACTIONS_RECEIVED,
+    };
 
-pub const TRANSACTIONS_PER_SUBMISSION: &str =
-    concat!(env!("CARGO_CRATE_NAME"), "_transaction_per_submission");
```
SuperFluffy, line 101: The macro (and test) caught this const being oddly named. This is now `astria_composer_transactions_per_submission`

crates/astria-telemetry/src/macros.rs
```diff
@@ -0,0 +1,58 @@
...
+/// ```
+/// use astria_telemetry::declare_metric_const;
+/// declare_metric_const!(pub const EXAMPLE_COUNTER);
+/// // Note that this example has `astria_telemetry` a a prefix because
+/// // this doctest is part of this crate.
+/// // In your case, use your crate's `CARGO_CRATE_NAME` as prefix.
+/// assert_eq!(EXAMPLE_COUNTER, "astria_telemetry_example_counter");
+/// ```
+#[macro_export]
+macro_rules! declare_metric_const {
```
Fraser999, line 31: Maybe `metric_name` might be a more succinct way of describing what the macro should be used for?  Or even `declare_metric_name` if you think the term "declare" is needed?

crates/astria-telemetry/Cargo.toml
```diff
@@ -13,6 +13,7 @@ homepage = "https://astria.org"
 [dependencies]
 base64 = { workspace = true, optional = true }
 base64-serde = { workspace = true, optional = true }
+const_format = { version = "0.2.32", optional = true }
```
Fraser999, line 16: I think it'd be better to not make this optional.  Having created the macro to avoid copy-paste errors/typos, I think it should be the default way to declare metric names going forwards.

Fraser999 (approved):
Nice upgrade!  I made a couple of suggestions, but both non-blocking.

I also think the unit tests aren't needed (except for the one in astria-telemetry) and could be removed since they're likely to become a source of frustration as we'll inevitably change metrics over time.

I was planning to add metrics snapshot testing of some sort in an upcoming PR, and I think if that's in place, these unit tests are really only testing the macro, which has its own unit test :)

crates/astria-telemetry/src/macros.rs
SuperFluffy: Renamed to `metric_name` in [e64ca43](https://github.com/astriaorg/astria/pull/1129/commits/e64ca43d377e20cad7621e12f11821b979e5dd14)

crates/astria-telemetry/Cargo.toml
SuperFluffy: Removed the `macros` feature and made this non-optional in [e64ca43](https://github.com/astriaorg/astria/pull/1129/commits/e64ca43d377e20cad7621e12f11821b979e5dd14)

SuperFluffy:
Rebased on top of recent main and used the new macro for the new bridge-withdrawer service

PR #1130, opened 2024-06-01
feature(charts): hermes chart fixes, bech32 updates, ibc bridge test

steezeburger:
## Summary
Fixes hermes helm chart

## Background
We want to run hermes via helm chart so we can automate e2e testing.

## Changes
* add 2nd funded celestia account
* use this account's key in hermes
* fix hermes chart values
* add new chart `bridge-test`
* add new justfile `dev/bridgetester.just`
* add `AstriaSequencerHrpPrefix` to evm rollup geth-genesis

## Testing
- `just deploy-cluster`
- `just deploy-ibc-test-infra`
  - deploys ingress controller
  - deploys celestia-local
  - deploys single astria sequencer
    - `./charts/sequencer/values.yaml`
    - `./dev/values/validators/single.yaml`
  - deploys evm rollup
    - `./charts/evm-rollup/values.yaml`
    - `./dev/values/rollup/dev.yaml`
    - `./dev/values/rollup/ibc-bridge-test.yaml`
  - deploy hermes
    - `./charts/hermes/values.yaml`
    - `./dev/values/hermes/local.yml`
- `just deploy-bridge-tester`
  - initContainer handles initializing bridge account on the sequencer
  - container runs scripts
    - `celestia-appd keys add`
    - `celestia-appd tx ibc-transfer transfer`
    - checks balance before and after and fails if ibc-transfer did not go through

## Metrics

## Breaking Changelist

## Related Issues

closes


dev/values/hermes/local.yml
```diff
@@ -27,38 +34,41 @@ chains:
       interval: 1s
     gasPrice: 1
     gasDenom: nria
-    trustThreshold: 2/3
+    trustThreshold: 1/3
```
joroshiba, line 37: this should probably be 2/3 to match tendermint?

dev/values/hermes/local.yml
```diff
@@ -27,38 +34,41 @@ chains:
...
+    gasPrice: 0.0026
+    gasMultiplier: 1.2
+    defaultGas: "1000000"
+    maxGas: "10000000"
     maxMsgNum: 30
-    maxTxSize: "180000"
+    maxTxSize: "2097152"
     clockDrift: 5s
     maxBlockTime: 30s
     trustThreshold: 1/3
```
joroshiba, line 74: Also should probably be 2/3

dev/values/hermes/local.yml
joroshiba, line 1: we should probably add a trustingPeriod of '14days' for both chains

dev/values/hermes/local.yml
steezeburger: Updated!
I got these values from https://github.com/astriaorg/hermes/blob/astria/config-astria-celestia.toml
Should I update there as well? Not 100% sure of the implication/meaning of trustThreshold

joroshiba: > The trust threshold defines what fraction of the total voting power of a known and trusted validator set is sufficient for a commit to be accepted going forward. - [hermes docs](https://hermes.informal.systems/documentation/commands/path-setup/clients.html?highlight=trust%20threshold#create-client)

Basically 2/3 is the threshold for CometBFT to accept a block so I don't think it makes sense to accept lower than that.

joroshiba (approved):
looks good enough to me, we can merge and then clean up

PR #1131, opened 2024-06-01
fix(sequencer): fix prefix removal when receiving source non-refund ics20 packet 

noot:
## Summary
follow up to https://github.com/astriaorg/astria/pull/851 which was an audit fix - this fixes the fix by removing the source prefix correctly, in the case assets we're the source of are transferred in. previously, the entire prefix was removed, would would have been invalid if the asset was prefixed by multiple hops, eg. `transfer/channel-1/transfer/channel-2/denom` would have become just `denom` when it should have become `transfer/channel-2/denom`.

## Background
audit fix

## Changes
- fix prefix removal when receiving source non-refund ics20 packet 

## Testing
unit tests


crates/astria-core/src/primitive/v1/asset.rs
```diff
@@ -83,6 +88,30 @@ impl Denom {
...
+    /// Create a new [`Denom`] with the given prefix removed.
+    ///
+    /// # Errors
+    ///
+    /// - if the denom does not have the given prefix.
+    pub fn remove_prefix(&self, prefix: &str) -> Result<Self, InvalidPrefixToRemove> {
+        let denom_trace = self
+            .to_string()
+            .strip_prefix(prefix)
+            .ok_or(InvalidPrefixToRemove {
```
SuperFluffy, line 101: Use a lazy constructor to avoid allocation on every single call (i.e. even if you don't end up hitting the error case):
```suggestion
            .ok_or_or_else(|| InvalidPrefixToRemove {
```

crates/astria-core/src/primitive/v1/asset.rs
```diff
@@ -83,6 +88,30 @@ impl Denom {
...
     }
+
+    /// Create a new [`Denom`] with the given prefix removed.
+    ///
+    /// # Errors
+    ///
+    /// - if the denom does not have the given prefix.
+    pub fn remove_prefix(&self, prefix: &str) -> Result<Self, InvalidPrefixToRemove> {
+        let denom_trace = self
+            .to_string()
```
SuperFluffy, line 99: Why allocate a new string prior to removing `strip_prefix`? Because one sometimes wants to remove a prefix that reaches "into" the `denom` part?

Nomenclature wise this leaves me pretty confused. Just looking at the definition of `Denom` I see a `prefix` and a `denom`. So why would a method called `remove_prefix` also remove parts of the `Denom`?

crates/astria-core/src/primitive/v1/asset.rs
```diff
@@ -68,6 +68,11 @@ impl Denom {
         self.prefix == prefix
     }
 
+    #[must_use]
+    pub fn is_prefixed_with(&self, prefix: &str) -> bool {
```
SuperFluffy, line 72: Nit: I think one would ask "is something prefixed by?"
```suggestion
    pub fn is_prefixed_by(&self, prefix: &str) -> bool {
```

crates/astria-core/src/primitive/v1/asset.rs
```diff
@@ -83,6 +88,30 @@ impl Denom {
     pub fn to_base_denom(&self) -> Self {
         Self::from_base_denom(&self.base_denom)
     }
+
+    /// Create a new [`Denom`] with the given prefix removed.
+    ///
+    /// # Errors
+    ///
+    /// - if the denom does not have the given prefix.
+    pub fn remove_prefix(&self, prefix: &str) -> Result<Self, InvalidPrefixToRemove> {
```
SuperFluffy, line 97: Can you add unit tests for this functionality?

crates/astria-sequencer/src/ibc/ics20_transfer.rs
```diff
@@ -488,7 +488,16 @@ async fn execute_ics20_transfer<S: StateWriteExt>(
...
         // strip the prefix from the denom, as we're back on the source chain
         // note: if this is a refund, this is a no-op.
-        let denom = unprefixed_denom.to_base_denom();
+        let denom = if is_refund {
+            unprefixed_denom
+        } else {
+            unprefixed_denom
+                .remove_prefix(&format!("{}/{}/", source_port, source_channel))
+                .expect(
+                    "denom must be prefixed by source_port/source_channel as it was checked in \
```
SuperFluffy, line 497: Is this actually correct?

This is the definition of `fn is_prefixed` (after the changes)
```rust
fn is_prefixed(source_port: &PortId, source_channel: &ChannelId, asset: &Denom) -> bool {
    let prefix = format!("{source_port}/{source_channel}/");
    asset.is_prefixed_with(&prefix)
}
```
But here is `Asset::is_prefixed_with`:
```rust
    pub fn is_prefixed_with(&self, prefix: &str) -> bool {
       self.prefix.starts_with(prefix)
    }
```
So this code is checking `Asset::prefix` if it starts with `"{source_port}/{source_channel}"`, but it actually removing removing `{source_port}/{source_channel}` from the concatenation `{Asset::prefix}/{Asset::denom}`.

SuperFluffy (commented):
There are some oddities in the code around removing the prefix. Can you add a few unit tests for the `remove_prefix` functionality (and for `is_prefixed_with`?)

crates/astria-sequencer/src/ibc/ics20_transfer.rs
noot: yes this is the correct functionality, for example if you have a token `{source_port}/{source_channel}/base_denom` then `is_prefixed` is true, and if this is source we want to remove `{source_port}/{source_channel}` from the whole denom. this is the same functionality as before, except now we handle the case where a token has multiple IBC hops, and is something like `{source_port}/{source_channel}/other_port/other_channel/base_denom`. previously, the entire prefix `{source_port}/{source_channel}/other_port/other_channel` would have been removed, which is wrong, we only want to remove the  `{source_port}/{source_channel}` prefix.

crates/astria-core/src/primitive/v1/asset.rs
noot: the `Denom` consists of a prefix and a base denom (I'll rename the `denom` field to `base_denom`), this is just a "denomination trace" broken down into parts (eg. `transfer/channel-1/utia` is broken down into `transfer/channel-1` and `utia`). so removing a prefix is removing some prefix from the entire denomination trace. is there a rename I could do to make this clearer?

noot:
@SuperFluffy  refactored `remove_prefix` and added unit tests!

crates/astria-core/src/primitive/v1/asset.rs
```diff
@@ -83,6 +88,48 @@ impl Denom {
...
+        let prefix_to_remove = prefix.trim_end_matches('/');
+        if prefix_to_remove == self.prefix {
+            return Ok(self.to_base_denom());
+        }
+
+        let new_prefix = self
+            .prefix
+            .clone()
```
Fraser999, line 109: I don't think we need the clone here

crates/astria-core/src/primitive/v1/asset.rs
```diff
@@ -83,6 +88,48 @@ impl Denom {
...
+            .strip_prefix(prefix_to_remove)
+            .ok_or_else(|| InvalidPrefixToRemove {
+                prefix: prefix.to_string(),
+                denom: self.to_string(),
+            })?
+            .trim_start_matches('/')
```
Fraser999, line 115: Should we maybe check at this point that the stripped prefix is either empty or starts with `/` and return an error if not?  Just now this would allow stripping e.g. `p0/p` from `p0/p1` leaving `1` which probably isn't desirable?

Either way - I guess a test for that case would be useful.

SuperFluffy (commented):
I think the way the prefix construction works this code will always problematic. I have rewritten it here: https://github.com/astriaorg/astria/pull/1162

PR #1134, opened 2024-06-01
refactor(sequencer): remove mint module

noot:
## Summary
remove the mint module and associated types.

## Background
we never used it, and not something we want in production anyways.

## Changes
- remove the mint module and associated types.

## Testing
n/a, code was removed

## Breaking changes
this removes the mint action, which is technically a breaking change, however it was gated under a feature flag in sequencer which was never enabled on any network afaik. so I don't think this change affects any running network.


proto/protocolapis/astria/protocol/transactions/v1alpha1/types.proto
```diff
@@ -47,17 +47,17 @@ message Action {
     astria_vendored.penumbra.core.component.ibc.v1.IbcRelay ibc_action = 21;
     Ics20Withdrawal ics20_withdrawal = 22;
 
-    // POA sudo actions are defined of 50-60
+    // POA sudo actions are defined on 50-60
     SudoAddressChangeAction sudo_address_change_action = 50;
     astria_vendored.tendermint.abci.ValidatorUpdate validator_update_action = 51;
     IbcRelayerChangeAction ibc_relayer_change_action = 52;
     FeeAssetChangeAction fee_asset_change_action = 53;
-    MintAction mint_action = 54;
```
SuperFluffy, line 55: Hm, if this were a non-alpha protobuf I would make this a 2-step process with 2 PRs:

1. first PR: set it as `MintAction mint_action = 54 [deprecated = true];
2. second PR: `reserved 54; // deprecated "mint_action"`.

I think it's ok to not make this a 2-step process because of the alpha-nature. But I think to be on the safe side, can you add a section that reads like this? When we go to `v1` (non-alpha) we can remove these.

```protobuf
// deprecated fields
reserved 54; // deprecated "mint_action"
reserved "mint_action";
```

SuperFluffy (approved):
Good to see this go. Just a comment on best-practices for deprecating proto message fields.

PR #1138, opened 2024-06-03
feat(sequencer): add ttl and invalid cache to app mempool

### LINKED ISSUE #979: Sequencer: can DOS the mempool with transactions that pass the mempool checks then fail execution/stateful checks
The sequencer was changed to filter out transactions that do not pass the stateless/stateful/execute steps in `prepare_proposal()`. This is causing transactions to get stuck inside of the mempool and to be repeatedly fed into the `prepare_proposal()` block where they are again filtered. Transactions that pass the mempool checks but not the stateful/execute steps exhibit this behavior (e.g. sudo actions with the wrong signing key will get stuck and loop until the signing key's nonce becomes stale because they pass the stateless/nonce checks but fail in the stateful check). 

We shouldn't allow this because it's an attack vector for people to fill up the mempool and cometbft `prepare_proposal()` blocks with these stuck transactions.

We either need a heuristic to kick out these transactions or to let them fail in the blocks like normal.

Lilyjjo:
## Summary
This PR adds to the App's mempool a way to signal to CometBFT when transactions should be removed from the CometBFT mempool. 

## Background
The CometBFT and App mempools currently get transactions that fail to execute stuck in them. This is because when a transaction fails to execute in `prepare_proposal()`, it doesn't get removed from the CometBFT mempool. CometBFT only clears out transactions if they either fail `handleCheckTx()` or are included in a block. Because of this, CometBFT will re-add the failed transaction to the App mempool during its `handleCheckTx()` maintenance, which will cause it to be fed to `prepare_proposal()` again.

We also need a way for full nodes to clear out these invalid transactions. Since these nodes don't run `prepare_proposal()`, we need an additional way, like a tx TTL, to signal when transactions can be dropped.

## Changes
- Added a transaction cache to the App's mempool which signals to CometBFT when a transaction should be removed.
- Tracks in the App's mempool when a transaction is first seen.
- Transactions that fail in `prepare_proposal()` get added to the App's removal cache.
- Transactions that are older than 10 minutes are added to the App's removal cache.

## Testing
Unit tests and local testing of invalid transactions. 

## Metrics
Added `CHECK_TX_REMOVED_FAILED_EXECUTION` counter to the sequencer's metrics. 
Added `CHECK_TX_REMOVED_EXPIRED` counter to the sequencer's metrics.

## Related
Closes #979

### DIFF
--- a/crates/astria-sequencer/Cargo.toml
+++ b/crates/astria-sequencer/Cargo.toml
@@ -67,6 +67,7 @@ config = { package = "astria-config", path = "../astria-config", features = [
   "tests",
 ] }
 insta = { workspace = true, features = ["json"] }
+tokio = { workspace = true, features = ["test-util"] }
 
 [build-dependencies]
 astria-build-info = { path = "../astria-build-info", features = ["build"] }
--- a/crates/astria-sequencer/src/mempool.rs
+++ b/crates/astria-sequencer/src/mempool.rs
@@ -3,8 +3,12 @@ use std::{
         self,
         Ordering,
     },
-    collections::HashMap,
+    collections::{
+        HashMap,
+        VecDeque,
+    },
     future::Future,
+    num::NonZeroUsize,
     sync::{
         Arc,
         OnceLock,
@@ -22,7 +26,13 @@ use astria_core::{
     },
 };
 use priority_queue::PriorityQueue;
-use tokio::sync::RwLock;
+use tokio::{
+    sync::RwLock,
+    time::{
+        Duration,
+        Instant,
+    },
+};
 use tracing::debug;
 
 type MempoolQueue = PriorityQueue<EnqueuedTransaction, TransactionPriority>;
@@ -34,6 +44,7 @@ type MempoolQueue = PriorityQueue<EnqueuedTransaction, TransactionPriority>;
 #[derive(Clone, Debug)]
 pub(crate) struct TransactionPriority {
     nonce_diff: u32,
+    time_first_seen: Instant,
 }
 
 impl PartialEq for TransactionPriority {
@@ -75,7 +86,11 @@ impl EnqueuedTransaction {
         }
     }
 
-    fn priority(&self, current_account_nonce: u32) -> anyhow::Result<TransactionPriority> {
+    fn priority(
+        &self,
+        current_account_nonce: u32,
+        time_first_seen: Option<Instant>,
+    ) -> anyhow::Result<TransactionPriority> {
         let Some(nonce_diff) = self.signed_tx.nonce().checked_sub(current_account_nonce) else {
             return Err(anyhow::anyhow!(
                 "transaction nonce {} is less than current account nonce {current_account_nonce}",
@@ -85,6 +100,7 @@ impl EnqueuedTransaction {
 
         Ok(TransactionPriority {
             nonce_diff,
+            time_first_seen: time_first_seen.unwrap_or(Instant::now()),
         })
     }
 
@@ -117,6 +133,62 @@ impl std::hash::Hash for EnqueuedTransaction {
     }
 }
 
+#[derive(Debug, Clone)]
+pub(crate) enum RemovalReason {
+    Expired,
+    FailedPrepareProposal(String),
+}
+
+const TX_TTL: Duration = Duration::from_secs(600); // 10 minutes 
+const REMOVAL_CACHE_SIZE: usize = 4096;
+
+/// `RemovalCache` is used to signal to `CometBFT` that a
+/// transaction can be removed from the `CometBFT` mempool.
+///
+/// This is useful for when a transaction fails execution or when a transaction
+/// has expired in the app's mempool.
+#[derive(Clone)]
+pub(crate) struct RemovalCache {
+    cache: HashMap<[u8; 32], RemovalReason>,
+    remove_queue: VecDeque<[u8; 32]>,
+    max_size: NonZeroUsize,
+}
+
+impl RemovalCache {
+    fn new(max_size: NonZeroUsize) -> Self {
+        Self {
+            cache: HashMap::new(),
+            remove_queue: VecDeque::with_capacity(max_size.into()),
+            max_size,
+        }
+    }
+
+    /// returns Some(RemovalReason) if transaction is cached and
+    /// removes the entry from the cache at the same time
+    fn remove(&mut self, tx_hash: [u8; 32]) -> Option<RemovalReason> {
+        self.cache.remove(&tx_hash)
+    }
+
+    /// adds the transaction to the cache
+    fn add(&mut self, tx_hash: [u8; 32], reason: RemovalReason) {
+        if self.cache.contains_key(&tx_hash) {
+            return;
+        };
+
+        if self.remove_queue.len() == usize::from(self.max_size) {
+            // make space for the new transaction by removing the oldest transaction
+            let removed_tx = self
+                .remove_queue
+                .pop_front()
+                .expect("cache should contain elements");
+            // remove transaction from cache if it is present
+            self.cache.remove(&removed_tx);
+        }
+        self.remove_queue.push_back(tx_hash);
+        self.cache.insert(tx_hash, reason);
+    }
+}
+
 /// [`Mempool`] is an internally-synchronized wrapper around a prioritized queue of transactions
 /// awaiting execution.
 ///
@@ -129,51 +201,92 @@ impl std::hash::Hash for EnqueuedTransaction {
 /// - transaction expiration
 #[derive(Clone)]
 pub(crate) struct Mempool {
-    inner: Arc<RwLock<MempoolQueue>>,
+    queue: Arc<RwLock<MempoolQueue>>,
+    comet_bft_removal_cache: Arc<RwLock<RemovalCache>>,
+    tx_ttl: Duration,
 }
 
 impl Mempool {
     #[must_use]
     pub(crate) fn new() -> Self {
         Self {
-            inner: Arc::new(RwLock::new(MempoolQueue::new())),
+            queue: Arc::new(RwLock::new(MempoolQueue::new())),
+            comet_bft_removal_cache: Arc::new(RwLock::new(RemovalCache::new(
+                NonZeroUsize::try_from(REMOVAL_CACHE_SIZE)
+                    .expect("Removal cache cannot be zero sized"),
+            ))),
+            tx_ttl: TX_TTL,
         }
     }
 
     /// returns the number of transactions in the mempool
     #[must_use]
     pub(crate) async fn len(&self) -> usize {
-        self.inner.read().await.len()
+        self.queue.read().await.len()
     }
 
     /// inserts a transaction into the mempool
     ///
-    /// note: if the tx already exists in the mempool, it's overwritten with the new priority.
+    /// note: the oldest timestamp from found priorities is maintained.
     pub(crate) async fn insert(
         &self,
         tx: SignedTransaction,
         current_account_nonce: u32,
     ) -> anyhow::Result<()> {
         let enqueued_tx = EnqueuedTransaction::new(tx);
-        let priority = enqueued_tx.priority(current_account_nonce)?;
-        let tx_hash = enqueued_tx.tx_hash;
-        self.inner.write().await.push(enqueued_tx, priority);
-        tracing::trace!(
-            tx_hash = %telemetry::display::hex(&tx_hash),
-            "inserted transaction into mempool"
-        );
+        let fresh_priority = enqueued_tx.priority(current_account_nonce, None)?;
+        Self::update_or_insert(&mut *self.queue.write().await, enqueued_tx, &fresh_priority);
+
         Ok(())
     }
 
     /// inserts all the given transactions into the mempool
+    ///
+    /// note: the oldest timestamp from found priorities for an `EnqueuedTransaction` is maintained.
     pub(crate) async fn insert_all(&self, txs: Vec<(EnqueuedTransaction, TransactionPriority)>) {
-        self.inner.write().await.extend(txs);
+        let mut queue = self.queue.write().await;
+
+        for (enqueued_tx, priority) in txs {
+            Self::update_or_insert(&mut queue, enqueued_tx, &priority);
+        }
+    }
+
+    /// inserts or updates the transaction in a timestamp preserving manner
+    ///
+    /// note: updates the priority using the `possible_priority`'s nonce diff.
+    fn update_or_insert(
+        queue: &mut PriorityQueue<EnqueuedTransaction, TransactionPriority>,
+        enqueued_tx: EnqueuedTransaction,
+        possible_priority: &TransactionPriority,
+    ) {
+        let oldest_timestamp = queue.get_priority(&enqueued_tx).map_or(
+            possible_priority.time_first_seen,
+            |prev_priority| {
+                possible_priority
+                    .time_first_seen
+                    .min(prev_priority.time_first_seen)
+            },
+        );
+
+        let priority = TransactionPriority {
+            nonce_diff: possible_priority.nonce_diff,
+            time_first_seen: oldest_timestamp,
+        };
+
+        let tx_hash = enqueued_tx.tx_hash;
+        if queue.push(enqueued_tx, priority).is_none() {
+            // emit if didn't already exist
+            tracing::trace!(
+                tx_hash = %telemetry::display::hex(&tx_hash),
+                "inserted transaction into mempool"
+            );
+        }
     }
 
     /// pops the transaction with the highest priority from the mempool
     #[must_use]
     pub(crate) async fn pop(&self) -> Option<(EnqueuedTransaction, TransactionPriority)> {
-        self.inner.write().await.pop()
+        self.queue.write().await.pop()
     }
 
     /// removes a transaction from the mempool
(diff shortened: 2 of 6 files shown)


crates/astria-sequencer/src/mempool.rs
```diff
@@ -62,13 +67,18 @@ impl PartialOrd for TransactionPriority {
...
 impl EnqueuedTransaction {
-    fn new(signed_tx: SignedTransaction) -> Self {
+    fn new(signed_tx: SignedTransaction, ttl: i64) -> Self {
         Self {
             tx_hash: signed_tx.sha256_of_proto_encoding(),
             signed_tx: Arc::new(signed_tx),
+            ttl: Time::now()
+                .unix_timestamp()
+                .checked_add(ttl)
+                .expect("overflow in enqueued transaction ttl"),
```
noot, line 81: i'd prefer having a `timeout_time` passed in, which is a unix timestamp of when the tx expires, over a ttl. it'll be easier to test and compute client-side when the exact timeout is going to be before submitting

Lilyjjo: would the time be decided by the transaction signer? I'm a little hesitant to let users add an expiration if they aren't signing over the expiration. future block builders wouldn't have to respect non-signed data if they didn't want to

crates/astria-sequencer/src/mempool.rs
```diff
@@ -114,6 +128,74 @@ impl std::hash::Hash for EnqueuedTransaction {
...
+            self.time_added
+                .insert(tx_hash, Time::now().unix_timestamp());
+            return;
+        };
+        if self.size == self.max_size {
+            // make space for the new transaction by removing the oldest transaction
+            let removed_tx = self
+                .remove_queue
+                .pop_front()
+                .expect("cache should contain elements");
```
Fraser999, line 186: I don't think we need to panic here - maybe an error log if we get `None`?

Alternatively, if we changed `max_size` to be a `std::num::NonZeroUsize` and changed `if self.size == self.max_size` to `if self.remove_queue.len() == usize::from(self.max_size)` it would be reasonable to leave the `expect` since it could never be hit.

crates/astria-sequencer/src/service/mempool.rs
```diff
@@ -187,6 +187,17 @@ async fn handle_check_tx<S: StateReadExt + 'static>(
         };
     };
 
+    if mempool.check_invalid(tx_hash).await {
+        mempool.remove(tx_hash).await;
+        metrics::counter!(metrics_init::CHECK_TX_REMOVED_INVALID).increment(1);
+        return response::CheckTx {
+            code: AbciErrorCode::INVALID_PARAMETER.into(),
```
Fraser999, line 194: Not sure if this is a good code - maybe we need a new one for this case?

crates/astria-sequencer/src/mempool.rs
```diff
@@ -126,68 +208,107 @@ impl std::hash::Hash for EnqueuedTransaction {
...
     #[must_use]
     pub(crate) async fn len(&self) -> usize {
-        self.inner.read().await.len()
+        self.queue.read().await.len()
     }
 
     /// inserts a transaction into the mempool
     ///
-    /// note: if the tx already exists in the mempool, it's overwritten with the new priority.
+    /// note: if the tx already exists in the mempool it's a noop
```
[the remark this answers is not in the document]
Lilyjjo: I didn't want the original TTL to be overwritten, the `insert_all()` takes Enqueued transactions that already have a TTL set. I need to better document / assert this!

crates/astria-sequencer/src/mempool.rs
noot: okay right, since this doesn't actually add it to the tx type, this is fine

crates/astria-sequencer/src/mempool.rs
```diff
@@ -114,6 +128,74 @@ impl std::hash::Hash for EnqueuedTransaction {
...
+/// ttl has expired. This is useful for users who want to re-run a transaction
+/// that failed to execute or to re-add a transaction that has expired.
+#[derive(Clone)]
+pub(crate) struct InvalidCache {
+    cache: HashSet<[u8; 32]>,
+    time_added: HashMap<[u8; 32], i64>,
+    remove_queue: VecDeque<[u8; 32]>,
+    max_size: usize,
+    time_to_live: i64,
+    size: usize,
```
noot, line 149: is this the current size of the cache? could you add a comment for this?

crates/astria-sequencer/src/mempool.rs
```diff
@@ -114,6 +128,74 @@ impl std::hash::Hash for EnqueuedTransaction {
...
+    }
+
+    fn cached(&self, tx_hash: [u8; 32]) -> bool {
+        // the tx is known and entry hasn't expired
+        self.cache.contains(&tx_hash)
+            && (Time::now().unix_timestamp()
+                <= self.time_added[&tx_hash]
+                    .checked_add(self.time_to_live)
+                    .expect("overflowed ttl add"))
+    }
```
noot, line 171: it might be nicer to store `expiry_time` instead of `time_added` when a tx is inserted, that way you don't need to do the add + expect here each time

crates/astria-sequencer/src/mempool.rs
```diff
@@ -114,6 +128,74 @@ impl std::hash::Hash for EnqueuedTransaction {
...
+        if self.size == self.max_size {
+            // make space for the new transaction by removing the oldest transaction
+            let removed_tx = self
+                .remove_queue
+                .pop_front()
+                .expect("cache should contain elements");
+            self.cache.remove(&removed_tx);
+            self.time_added.remove(&removed_tx);
+        } else {
+            self.size = self.size.checked_add(1).expect("cache size overflowed");
```
noot, line 190: do we need a `self.size` variable over just checking `self.remove_queue.len()` or len of the hashmap?

crates/astria-sequencer/src/mempool.rs
```diff
@@ -114,6 +128,74 @@ impl std::hash::Hash for EnqueuedTransaction {
...
+            cache: HashSet::new(),
+            time_added: HashMap::new(),
+            remove_queue: VecDeque::with_capacity(max_size),
+            max_size,
+            time_to_live,
+            size: 0,
+        }
+    }
+
+    fn cached(&self, tx_hash: [u8; 32]) -> bool {
```
noot, line 164: this funtion returns `true` if the tx is invalidated but not expired right? could you add a comment for that or maybe change the function name to `cached_and_not_expired` for clarity?

crates/astria-sequencer/src/mempool.rs
noot: or maybe have separate functions for `is_cached` and `is_expired`?

crates/astria-sequencer/src/mempool.rs
```diff
@@ -126,68 +208,107 @@ impl std::hash::Hash for EnqueuedTransaction {
 /// - transaction expiration
 #[derive(Clone)]
 pub(crate) struct Mempool {
-    inner: Arc<RwLock<MempoolQueue>>,
+    queue: Arc<RwLock<MempoolQueue>>,
+    set: Arc<RwLock<HashSet<[u8; 32]>>>,
+    invalid_cache: Arc<RwLock<InvalidCache>>,
+    tx_ttl: i64,
```
noot, line 214: do we need to store this if it's already a constant? it appears to always be set to the same value

crates/astria-sequencer/src/mempool.rs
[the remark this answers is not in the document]
noot: you could put the original ttl into the current tx and overwrite the queued value with the new priority

crates/astria-sequencer/src/mempool.rs
```diff
@@ -126,68 +208,107 @@ impl std::hash::Hash for EnqueuedTransaction {
...
     #[must_use]
     pub(crate) async fn pop(&self) -> Option<(EnqueuedTransaction, TransactionPriority)> {
-        self.inner.write().await.pop()
+        let tx = self.queue.write().await.pop();
+
+        if let Some((tx, priority)) = tx {
+            self.set.write().await.remove(&tx.tx_hash);
+            Some((tx, priority))
+        } else {
+            tx
```
noot, line 281:
```suggestion
            None
```

a bit more explicit

crates/astria-sequencer/src/mempool.rs
noot: is this the right behaviour? based on the logic for adding txs to the invalid cache and when to remove them from cometbft, should this return true if the tx is expired?

crates/astria-sequencer/src/service/mempool.rs
noot: yeah, a code specifically for this would be good

crates/astria-sequencer/src/mempool.rs
Lilyjjo: the comment was ambiguous, it was returning true if it was expired in the app's mempool but not expired in the cache. refactored to make it more clear

Lilyjjo:
Changes made since last reviews:
- switched time accounting to `tokio::time::{Instant, Duration}`
- renamed `InvalidCache` to `RemovalCache` and renamed Mempool’s use of it to `comet_bft_removal_cache` to try to better document what the cache is and how it is being used 
- simplified the `RemovalCache` by removing the redundant `HashSet` and `size` struct members
- removed the `HashSet` added to the mempool in the last write
- added tracking to `RemovalCache` as to why the transaction was added and surface error in CheckTx to user
- added `AbciErrorCode`s and metrics for expired transactions and failing to execute transactions

crates/astria-sequencer/Cargo.toml
```diff
@@ -50,7 +50,7 @@ serde_json = { workspace = true }
 sha2 = { workspace = true }
 tendermint-proto = { workspace = true }
 tendermint = { workspace = true }
-tokio = { workspace = true, features = ["rt", "tracing"] }
+tokio = { workspace = true, features = ["rt", "tracing", "test-util"] }
```
noot, line 53: if the `test-util` feature is just for tests, move it under `dev-dependencies`

crates/astria-sequencer/src/mempool.rs
```diff
@@ -117,6 +133,81 @@ impl std::hash::Hash for EnqueuedTransaction {
...
+pub(crate) struct TransactionExpired;
+
+impl std::fmt::Display for TransactionExpired {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "Transaction expired")
+    }
+}
+
+const TX_TTL: Duration = Duration::from_secs(600); // 10 minutes 
+const REMOVAL_CACHE_SIZE: NonZeroUsize = unsafe { NonZeroUsize::new_unchecked(4096) }; // TODO make configuration variable
```
noot, line 146: why not just use `usize`?

crates/astria-sequencer/src/mempool.rs
```diff
@@ -117,6 +133,81 @@ impl std::hash::Hash for EnqueuedTransaction {
...
+    }
+}
+
+const TX_TTL: Duration = Duration::from_secs(600); // 10 minutes 
+const REMOVAL_CACHE_SIZE: NonZeroUsize = unsafe { NonZeroUsize::new_unchecked(4096) }; // TODO make configuration variable 
+const TX_REMOVAL_TTL: Duration = Duration::from_secs(30); // 30 seconds
+
+/// `RemovalCache` is a cache used for signaling to `CometBFT` when a
+/// transaction can be removed from the `CometBFT` mempool outside of the
+/// `CheckTx` checks.
```
noot, line 151: this is checked in `check_tx` right? there isn't a way to tell cometbft to remove something from the mempool without cometbft calling `check_tx` on it

crates/astria-sequencer/src/mempool.rs
```diff
@@ -117,6 +133,81 @@ impl std::hash::Hash for EnqueuedTransaction {
...
+/// This is useful for when a transaction fails execution or when a transaction
+/// has expired in the app's mempool.
+///
+/// The cache places a TTL on the entries, allowing users
+/// to request their transaction to be processed again after the
+/// ttl has expired. This is useful for users who want to re-run a transaction
+/// that failed to execute or to re-add a transaction that has expired.
+#[derive(Clone)]
+pub(crate) struct RemovalCache {
+    cache: HashMap<[u8; 32], (Instant, Arc<anyhow::Error>)>,
```
noot, line 162: for the error, the only two options right now are expired and failed execution right? i'd prefer to change the `Arc<anyhow::Error>` to an enum, that would be clearer as to which removal reasons exist

crates/astria-sequencer/src/mempool.rs
```diff
@@ -117,6 +133,81 @@ impl std::hash::Hash for EnqueuedTransaction {
...
+            cache: HashMap::new(),
+            remove_queue: VecDeque::with_capacity(max_size.into()),
+            max_size,
+            time_to_live,
+        }
+    }
+
+    /// returns Some(Error) if transaction is cached, where the Error is
+    /// the reason why it was cached. None is returned if the transaction
+    /// is not cached
```
noot, line 180:
```suggestion
    /// the reason why it was cached. None is returned if the transaction
    /// is not cached or is not expired
```

crates/astria-sequencer/src/mempool.rs
```diff
@@ -157,23 +255,33 @@ impl Mempool {
...
+            );
+        }
+
         Ok(())
     }
 
     /// inserts all the given transactions into the mempool
     pub(crate) async fn insert_all(&self, txs: Vec<(EnqueuedTransaction, TransactionPriority)>) {
-        self.inner.write().await.extend(txs);
+        self.queue.write().await.extend(txs);
```
noot, line 278: this isn't the same behaviour as `insert` as it doesn't preserve the timestamp if it already exists

crates/astria-sequencer/src/mempool.rs
Lilyjjo: Fraser recommended this instead I'm guessing for configuration safety purposes, but I can change back to just a `usize`

crates/astria-sequencer/src/mempool.rs
Lilyjjo: My b, yeah the comment should read 'other CheckTX checks'

crates/astria-sequencer/src/mempool.rs
Lilyjjo: I was trying to bubble up the reason for failure to the user, but it gets stuck at 'failed execution' or 'failed stateful checks'. I can make it an enum if you don't think that the `CheckTx` failure case should be more specific than 'failed execution'

crates/astria-sequencer/src/mempool.rs
Lilyjjo: I was considering a tx being 'cached' as being both not expired and in the cache, but I can change

crates/astria-sequencer/src/mempool.rs
Lilyjjo: You're right. I was just relying on the behavior that we're currently only using this function to re-insert transactions that we just popped. Do you have a preference of me adding a comment that explains that or to do a similar process as in `insert()`?

Lilyjjo:
Logic changes since last review:
- made `insert_all()` preserve timestamp like `insert()`
- made RemovalCache errors into enum

crates/astria-sequencer/src/mempool.rs
noot: i think doing a similar process as to `insert()` makes the most sense!

crates/astria-sequencer/src/mempool.rs
noot: i see, yeah i got confused from the variable naming, a comment on that would help!

noot (approved):

crates/astria-sequencer/src/mempool.rs
noot: sorry i got confused about the tx removal ttl vs the tx in mempool ttl. this makes sense!

crates/astria-sequencer/src/mempool.rs
```diff
@@ -117,6 +130,98 @@ impl std::hash::Hash for EnqueuedTransaction {
     }
 }
 
+/// inserts or updates the transaction in a timestamp preserving manner
+fn update_or_insert(
```
Lilyjjo, line 134: @Fraser999 is this an okay place to put a function like this? The rust CLI suggested I move it out of the Mempool struct's implementation because it wasn't using the `self` variable

Fraser999: Absolutely.  Another choice would be to leave it inside the `impl Mempool` but just don't have `self` as a parameter.  You have to call it like `Self::update_or_insert` in that case.  It can make sense to do that in cases like this where you have more than struct in scope which _could_ use this function, but it's really only intended to be used by the `Mempool` struct.  But for a private function like this, it's pretty much down to personal preference :)

You _could_ simplify the signature though by just taking `queue: &mut PriorityQueue` and at callsites passing `&mut *self.queue.write().await`.  Definitely not a big deal though :)

crates/astria-sequencer/src/mempool.rs
```diff
@@ -72,6 +84,7 @@ impl EnqueuedTransaction {
             tx_hash: signed_tx.sha256_of_proto_encoding(),
             signed_tx: Arc::new(signed_tx),
             address,
+            time_first_seen: Instant::now(),
```
Fraser999, line 87: This is probably better suited to live in the `TransactionPriority` struct since it might be considered in the future as part of the prioritisation algo, and also that would keep `EnqueuedTransaction`'s hash a genuine way to determine equality/ordering.

crates/astria-sequencer/src/mempool.rs
```diff
@@ -117,6 +130,98 @@ impl std::hash::Hash for EnqueuedTransaction {
...
+        queue.change_priority(&enqueued_tx, priority);
+    } else {
+        // doesn't exist
+        let tx_hash = enqueued_tx.tx_hash;
+        queue.push(enqueued_tx, priority);
+        tracing::trace!(
+            tx_hash = %telemetry::display::hex(&tx_hash),
+            "inserted transaction into mempool"
+        );
+    }
```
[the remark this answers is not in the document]
Lilyjjo: rewrote in moving the `time_first_seen` into the `TransactionPriority` struct

Lilyjjo:
Changes since last reviews:
- Removed TTL from `RemovalCache`.
- Moved `time_first_seen` out of the `EnqueuedTransaction` struct into the `TransactionPriority` struct.

Fraser999 (approved):

PR #1140, opened 2024-06-03
Chore(charts): update sequencer faucet chart

quasystaty1:
## Summary
Sequencer faucet needs an update to use latest protos.

## Changes
* version update for latest `sequencer` and `sequencer-faucet` charts
* adds new `sequencerChainId` value, defaults to the default chart sequencer chain Id
* decreases genesis accounts balance allocation, preventing balance overflow of sudo address
*  adjust other variables

## Testing
Tested against local k8 cluster and dev-network.

## Breaking Changelist
* sequencer faucet now takes a `sequencerChainId` as required by latest proto for sending a transfer action.


charts/sequencer/values.yaml
```diff
@@ -38,14 +38,14 @@ config:
     # Note large balances must be strings support templating with the u128 size account balances
     genesisAccounts:
     - address: 1c0c490f1b5528d8173c5de46d131160e4b2c0c3
-      balance: "340282366920938463463374607431768211455"
+      balance: "333333333333333333"
```
joroshiba, line 41: What is the reason for update here?

quasystaty1: fees are now being redirected to fee recipient, causing an overflow on fee recipient balance account. Took the value used in dusk-7.

joroshiba (approved):

PR #1142, opened 2024-06-04
feat(sequencer)!: implement bridge sudo and withdrawer addresses

noot:
## Summary
we want bridge accounts to be "upgradable"; ie. we want to be able to change the key which can control the bridge account without changing the bridge address from the rollup's point of view. with this change, we have a sudo key and a withdrawer key for each bridge account, which is set when the bridge account is initialized. the sudo key is able to change the sudo key and withdrawer key. the withdrawer key is the only key allowed to withdraw funds from the bridge account.

## Background
see above

## Changes
- change `InitBridgeAccountAction` to contain a `sudo_address` and `withdrawer_address` which are set in state
- change `BridgeUnlockAction` to have a `bridge_address` field which is the bridge address withdrawn from; it also checks if the signer of the tx is the bridge `withdrawer_address` and fail if not 
- implement `BridgeSudoChangeAction` which allows the bridge sudo key to change the bridge sudo and/or withdrawer keys
- `Ics20Withdrawal`s from bridge now also do a withdrawer key check like `BridgeUnlockAction`; `bridge_address` was added to the action in the case that the withdrawer differs from the bridge address
- disallow `Transfer` from a bridge account, as  `BridgeUnlockAction` or `Ics20Withdrawal` must be used. 

## Testing
unit tests

## Breaking Changelist
- `InitBridgeAccount` stores two new keys in state: the bridge sudo key and the bridge withdrawer key
- a new action, `BridgeSudoChangeAction` is added
- this is breaking on the sequencer state machine level.
- however it should NOT be breaking on the protobuf level as nothing was removed, only new fields/messages were added, and the new fields added are allowed to be unset. if the fields are unset, it gets converted to a `None` in the native rust type.


proto/protocolapis/astria/protocol/transactions/v1alpha1/types.proto
```diff
@@ -160,6 +171,11 @@ message InitBridgeAccountAction {
   bytes asset_id = 2;
   // the asset used to pay the transaction fee
   bytes fee_asset_id = 3;
+  // the address corresponding to the key which has sudo capabilities;
+  // ie. can change the sudo and withdrawer addresses for this bridge account.
+  astria.primitive.v1.Address sudo_address = 4;
+  // the address corresponding to the key which can withdraw funds from this bridge account.
```
joroshiba, line 177: Can we note in comments what happens if they are left a null?

crates/astria-sequencer/src/app/tests_breaking_changes.rs
```diff
@@ -247,18 +234,52 @@ async fn app_execute_transaction_with_every_action_snapshot() {
...
+    let signed_tx = Arc::new(tx.into_signed(&bridge_signing_key));
+    app.execute_transaction(signed_tx).await.unwrap();
+
+    let tx = UnsignedTransaction {
+        params: TransactionParams {
+            nonce: 1,
+            chain_id: "test".to_string(),
+        },
+        actions: vec![
+            BridgeLockAction {
```
joroshiba, line 262: Why relocate the lock + sudo into this one tx? It feels like a weird flow to lock + unlock in the same tx?

crates/astria-sequencer/test-genesis-app-state.json
```diff
@@ -27,6 +27,7 @@
     "sequence_byte_cost_multiplier": 1,
     "init_bridge_account_base_fee": 48,
     "bridge_lock_byte_cost_multiplier": 1,
+    "bridge_sudo_change_fee": 24,
```
joroshiba, line 30: does this make this a hard fork requiring network regen change?

In general actually, we probably need an answer for this. We should be able to add actions with fees without regenesis of network

proto/protocolapis/astria/protocol/transactions/v1alpha1/types.proto
```diff
@@ -195,6 +211,21 @@ message BridgeUnlockAction {
...
+  astria.primitive.v1.Address bridge_address = 1;
+  // the new sudo address; unchanged if unset
+  astria.primitive.v1.Address new_sudo_address = 2;
+  // the new withdrawer address; unchanged if unset
+  astria.primitive.v1.Address new_withdrawer_address = 3;
+  // the asset used to pay the transaction fee
+  bytes fee_asset_id = 4;
 }
 
 message FeeChangeAction {
```
joroshiba, line 231: I believe we need a new entry for FeeChangeAction

crates/astria-sequencer/src/bridge/state_ext.rs
```diff
@@ -57,9 +57,12 @@ impl From<&asset::Id> for AssetId {
 struct Fee(u128);
 
 const BRIDGE_ACCOUNT_PREFIX: &str = "bridgeacc";
+const BRIDGE_ACCOUNT_SUDO_PREFIX: &str = "bsudo";
+const BRIDGE_ACCOUNT_WITHDRAWER_PREFIX: &str = "bwithdrawer";
```
joroshiba, line 61: we use `bridge` as prefix for all other  bridger related things. Should probably be consistent?

crates/astria-sequencer/src/bridge/state_ext.rs
```diff
@@ -220,6 +272,17 @@ pub(crate) trait StateReadExt: StateRead {
         Ok(fee)
     }
 
+    #[instrument(skip(self))]
+    async fn get_bridge_sudo_change_fee(&self) -> Result<u128> {
```
joroshiba, line 276: This behaivior means you just won't be able to do the action unless fee is set, so if we get fee action updated you can't use these features until we set the fee right? That's a kinda nifty way to activate new actions.

crates/astria-sequencer/src/ibc/ics20_withdrawal.rs
```diff
@@ -51,6 +53,49 @@ fn withdrawal_to_unchecked_ibc_packet(
...
+    // the `get_bridge_account_withdrawer_address` step will fail.
+    let bridge_address = action.bridge_address.unwrap_or(from);
+
+    let Some(withdrawer) = state
+        .get_bridge_account_withdrawer_address(&bridge_address)
+        .await
+        .context("failed to get bridge withdrawer")?
+    else {
+        bail!("bridge address must have a withdrawer address set");
+    };
```
joroshiba, line 89: Would this cause sync failures with old withdrawals? If the withdrawer isn't set should we just assume it's the same as bridge address?

crates/astria-bridge-withdrawer/src/withdrawer/ethereum/convert.rs
joroshiba, line 1: Should one of these have a non none BridgeAddress?

crates/astria-bridge-withdrawer/priv.key
joroshiba, line 1: how is this used?

crates/astria-core/src/protocol/transaction/v1alpha1/action.rs
```diff
@@ -1305,6 +1412,123 @@ enum BridgeUnlockActionErrorKind {
...
+    }
+
+    /// Convert from a raw, unchecked protobuf [`raw::BridgeSudoChangeAction`].
+    ///
+    /// # Errors
+    ///
+    /// - if the `bridge_address` field is not set
+    /// - if the `bridge_address` field is invalid
+    /// - if the `new_sudo_address` field is invalid
+    /// - if the `new_withdrawer_address` field is invalid
```
joroshiba, line 1456: fee_asset_id can also be invalid?

crates/astria-sequencer/src/accounts/action.rs
```diff
@@ -85,6 +86,15 @@ impl ActionHandler for TransferAction {
         state: &S,
         from: Address,
     ) -> Result<()> {
+        ensure!(
+            state
+                .get_bridge_account_rollup_id(&from)
+                .await
+                .context("failed to get bridge account rollup id")?
+                .is_none(),
```
joroshiba, line 94: This logic seems funky. What happens in `get_bridge_account_rollup_id` if the account is not a bridge account? Is that just an `Ok(None)`? The context here makes it sounds like we failed to find a bridge account  rollup for the address, and then we want there to be none.

Also why use the get rollup id vs just checking for the bridge account existence?

crates/astria-sequencer/src/bridge/bridge_sudo_change_action.rs
```diff
@@ -0,0 +1,181 @@
...
+        );
+
+        // check that the sender of this tx is the authorized sudo address for the bridge account
+        let Some(sudo_address) = state
+            .get_bridge_account_sudo_address(&self.bridge_address)
+            .await
+            .context("failed to get bridge account sudo address")?
+        else {
+            // TODO: if the sudo address is unset, should we still allow this action
+            // if the sender if the bridge address itself?
```
joroshiba, line 47: Probably otherwise you can never update bridge accounts created previous to this change?

crates/astria-sequencer/src/bridge/bridge_unlock_action.rs
```diff
@@ -28,29 +30,48 @@ impl ActionHandler for BridgeUnlockAction {
...
             .context("failed to get bridge's asset id, must be a bridge account")?;
 
+        // check that the sender of this tx is the authorized withdrawer for the bridge account
+        let Some(withdrawer_address) = state
+            .get_bridge_account_withdrawer_address(&bridge_address)
+            .await
+            .context("failed to get bridge account withdrawer address")?
+        else {
+            bail!("bridge account does not have an associated withdrawer address");
+        };
```
joroshiba, line 50: similar to above, what about old bridge accounts and during sync this will fail for old unlocks

crates/astria-sequencer/src/bridge/component.rs
```diff
@@ -24,6 +24,7 @@ impl Component for BridgeComponent {
     async fn init_chain<S: StateWriteExt>(mut state: S, app_state: &Self::AppState) -> Result<()> {
         state.put_init_bridge_account_base_fee(app_state.fees.init_bridge_account_base_fee);
         state.put_bridge_lock_byte_cost_multiplier(app_state.fees.bridge_lock_byte_cost_multiplier);
+        state.put_bridge_sudo_change_fee(app_state.fees.bridge_sudo_change_fee);
```
joroshiba, line 27: It seems like this will break if we don't have a fee set in genesis so node could not sync with old genesis?

joroshiba (changes requested):

crates/astria-sequencer/src/app/tests_breaking_changes.rs
noot: the stateful check requires the bridge account to be initialized already for these actions, so i had to put `InitBridgeAccountAction` into a separate tx. there isn't a reason not to put them in one tx? putting them in separate txs will still have the same state changes

crates/astria-sequencer/test-genesis-app-state.json
noot: yes, this would be a hard fork. without hard fork logic this is a regenesis

crates/astria-sequencer/src/bridge/state_ext.rs
noot: it just adds more bytes to the db which i want to avoid, maybe we should change all of these

crates/astria-sequencer/src/bridge/state_ext.rs
noot: yess that is true!

crates/astria-sequencer/src/ibc/ics20_withdrawal.rs
noot: bridge accounts initialized before these changes would not have sudo/withdrawer set, so they would have to submit a `BridgeSudoChange` action to set them before this would work. i originally implemented it like you said but i didn't like it as it's better to force the addresses to be set

crates/astria-bridge-withdrawer/src/withdrawer/ethereum/convert.rs
noot: right now it doesn't matter since the withdrawer doesn't have the sudo/withdrawer separation of keys

crates/astria-bridge-withdrawer/priv.key
noot: accident will remove

crates/astria-sequencer/src/accounts/action.rs
noot: the context is just checking if `get_bridge_account_rollup_id` returned an error. if no rollup id is found for the bridge account, that means the bridge account does not exist. there's no explicit method for checking "existence" other than this

crates/astria-sequencer/src/bridge/bridge_sudo_change_action.rs
noot: i left this as-is because i wasn't sure if we were going to hard-fork or regenesis. if we regenesis, better to leave this

crates/astria-sequencer/src/bridge/bridge_unlock_action.rs
noot: same as above, user would need to set their sudo/withdrawer addrs before using

joroshiba (approved):
Approved, with note that this will be major breaking change, requiring regensis and a dusk-8. Moving forward if we add actions with new fees I think we should simply disallow the action to be run if a fee has not been set. This provides a mechanism through which we can add a new action without changing the genesis.

Issue #1100, opened 2024-05-23
Ensure all metrics are registered on executable startup


Fraser999:
Currently across all our binaries, individual metrics are only registered on their first usage, i.e. when we record that metric for the first time.  This is not ideal, as some metrics (especially ones recording failures) remain invisible via the metrics endpoint throughout the lifetime of the session.

We should register all metrics during startup, potentially holding the returned metric object in a new struct which can be cloned to all places in the codebase required to record metrics.  This would be more efficient than attempting to register the given metric every time we want to record it.

We should also ensure [the call to `register_metrics`](https://github.com/astriaorg/astria/blob/d85d5945c882d66bb91fbe868b9b8358d2a2bfed/crates/astria-telemetry/src/lib.rs#L348) occurs after [the call to `metrics_builder.install()`](https://github.com/astriaorg/astria/blob/d85d5945c882d66bb91fbe868b9b8358d2a2bfed/crates/astria-telemetry/src/lib.rs#L350) as it has no effect otherwise.

PR #1144, opened 2024-06-04
chore: register all metrics during startup

### LINKED ISSUE #1027: composer: Fix `clippy::too_many_lines` warning for the metrics_init::registry method
The metrics_init::registry method is too long according to the `too_many_lines` lint. This method mainly just registers metrics. To avoid this lint, we should figure out the best way to shorten this method.

Fraser999:
## Summary
Metrics in all five binaries are registered early in runtime flow.

## Background
For a given binary, we want the metrics http endpoint to always yield data for every metric relevant to that binary, even if some of them have not been modified since startup.  The `metrics::describe_` macros (currently called during startup) don't register the provided metrics - instead individual metrics are only registered on the first execution of the relevant `counter!`, `gauge!` or `histogram!` macro.

Further, the descriptions were never being applied, since in `astria-telemetry` the call to these macros was made before the metrics builder was installed.

## Changes
Similar changes were made to `astria-bridge-withdrawer`, `astria-composer`, `astria-conductor`, `astria-sequencer` and `astria-sequencer-relayer`:
- `metrics_init` module renamed to `metrics`, and used to hold a `Metrics` struct containing the relevant metrics objects.
- a static ref to a `Metrics` struct is created early in the flow (but after the call to telemetry's `Config::try_init` where the metrics builder installs the global registry) and passed to whichever objects/functions require it.  This approach will ultimately allow us to choose whether to use the global metrics registry or a local one, making the metrics much more testable in the future.
- removed telemetry `Config`'s `metric_buckets` since it wasn't really useful (it applies the bucket limits to all histograms, regardless of type), and `register_metrics` since it's also no longer used.
- changed the units of several `_latency` histograms from ms to seconds, since they were actually recording seconds. This is a non-issue though, as the units are ignored in Prometheus.
- added a description for `astria_conductor_sequencer_blocks_metadata_verified_per_celestia_fetch` as this was missing.

## Testing
Manually tested that all metrics for a couple of the binaries are available in full as soon as the binary is started.  Programmatic testing can be added in follow-up PRs.

## Metrics
- Changed the type of `astria_composer_transactions_dropped_too_large` from gauge to counter.  We previously had `describe_counter!` to describe it and it was only ever being incremented by 1, but it was accessed via `gauge!` which seemed like a mistake, so I changed its type to `Counter`.
- `astria_composer_transactions_received` and `astria_composer_transactions_dropped` each used to have two labels; one to ID the rollup and the other specifying collector type.  For the geth collector, the `rollup_id` label used the rollup's chain name, whereas for the grpc collector it used the rollup ID for the `rollup_id` label.  I changed these two metrics to always provide three labels; `collector_type`, `rollup_id` and `rollup_chain_name`.

## Related Issues
Closes #1027 and #1100.


crates/astria-composer/src/executor/mod.rs
```diff
@@ -422,9 +424,9 @@ impl Executor {
 async fn get_latest_nonce(
     client: sequencer_client::HttpClient,
     address: Address,
+    metrics: &Metrics,
 ) -> eyre::Result<u32> {
     debug!("fetching latest nonce from sequencer");
-    metrics::counter!(crate::metrics_init::NONCE_FETCH_COUNT).increment(1);
```
Fraser999, line 427: I moved this to inside `tryhard::retry_fn` on line 457, since it seems from the description ("The number of times we have attempted to fetch the nonce") like we want to count attempts rather than successes.

If we do mean success counts, I can revert this change and we could consider renaming the metric and description to reflect that?

crates/astria-conductor/src/conductor.rs
```diff
@@ -94,6 +96,9 @@ impl Conductor {
     /// actors could not be spawned (executor, sequencer reader, or data availability reader).
     /// This usually happens if the actors failed to connect to their respective endpoints.
     pub fn new(cfg: Config) -> eyre::Result<Self> {
+        static METRICS: OnceLock<Metrics> = OnceLock::new();
```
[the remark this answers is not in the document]
Fraser999: The intent here is to actively avoid creating a _global_ static, since that's what ties our hands during testing.

I agree the approach in this PR is fairly intrusive, but by passing a static ref down to the callsites we can instantiate multiple tests concurrently and provide each test instance with its own exclusive `Metrics` struct, avoiding munging the metrics between tests.

As an aside, I considered just passing the `Metrics` struct everywhere by value, cloning as required.  That's feasible with the implementation in this PR.  The individual metrics are all just wrappers around an `Arc`, so relatively cheap to clone.  But not as cheap as a static ref, so I opted for that (the Composer's `Metrics` is probably the most expensive to clone, holding five `HashMap`s of counters and another eight individual metrics).

crates/astria-composer/src/executor/mod.rs
SuperFluffy: I think the original idea was not to count the individual GET requests but to count the "process" of getting a nonce.

Having said that, this metric feels a tad useless because it either closely tracks the number of failures/attempts (as it's doing now), or the number of data points added to the histogram.

I am actually leaning toward removing it altogether...

SuperFluffy (approved):

PR #1149, opened 2024-06-05
feat(withdrawer): bridged ERC20 token withdrawals

### LINKED ISSUE #924: sequencer: ERC20 deposits and bridging
it would be good to be able to bridge tokens from the sequencer to the evm which are represented as ERC20s (or some other non-native token), as currently, bridged funds are minted as the native token.

considerations:
- gas cost (paid on the l2 from the sender's account? what if they have no native token?)
- how does the rollup know the contract/methods associated with the bridged asset? how does mint permissioning work for a bridged token?

noot:
## Summary
implement withdrawals of ERC20 tokens that are of type `AstriaBridgeableERC20` (see implemented contract).

## Background
we want to be able to withdraw ERC20 tokens that are bridged to a rollup.

## Changes
- implement `AstriaBridgeableERC20` which is a standard `ERC20` contract with additional functionality for minting (not used by the withdrawer, implemented/tested here https://github.com/astriaorg/astria-geth/pull/20) as well as functionality for withdrawing
- implement `IAstriaWithdrawer` which is implemented by both `AstriaWithdrawer` and `AstriaMintableERC20`. 
- the withdrawer now interacts with a `IAstriaWithdrawer`. both native assets and ERC20 withdrawals have the same event signatures, so no additional code was needed for the withdrawer itself.
- to use the withdrawer with an `AstriaBridgeableERC20`, the `ASTRIA_BRIDGE_WITHDRAWER_ETHEREUM_CONTRACT_ADDRESS` is set to some `AstriaMintableERC20`, and `ASTRIA_BRIDGE_WITHDRAWER_ROLLUP_ASSET_DENOMINATION` is set to the rollup asset's denomination as represented on the sequencer. for example, if the asset is represented on the sequencer is `transfer/channel-1/usdc`, that is the rollup asset denomination. the `name/symbol` of the ERC20 contract are not relevant.
- also update the build script to write the generated abigen contract bindings to files, this is easier for debugging and unit testing.

## Testing
unit tests

## Related Issues

closes #924


crates/astria-bridge-withdrawer/ethereum/src/AstriaBridgeableERC20.sol
```diff
@@ -0,0 +1,68 @@
+// SPDX-License-Identifier: MIT or Apache-2.0
```
joroshiba, line 1: This probably applies to the regular withdrawer as well, but what happens if you don't have enough funds to burn everything you requested?

crates/astria-bridge-withdrawer/ethereum/src/AstriaBridgeableERC20.sol
```diff
@@ -0,0 +1,68 @@
...
+    function mint(address _to, uint256 _amount)
+        external
+        onlyBridge
+    {
+        _mint(_to, _amount);
+        emit Mint(_to, _amount);
+    }
+
+    function withdrawToSequencer(uint256 _amount, address _destinationChainAddress)
+        external
```
joroshiba, line 54: why is the native asset contract `external payable` and this one only `external`?

crates/astria-bridge-withdrawer/src/withdrawer/ethereum/generated/astria_withdrawer_interface.rs
joroshiba, line 1: nit: add new generated file type locations into `.gitattributes` file, we should probably do this with the json attribute files as well?

crates/astria-bridge-withdrawer/build.rs
```diff
@@ -1,4 +1,32 @@
+use ethers::contract::Abigen;
```
joroshiba, line 1: should we be using something from alloy instead?

crates/astria-bridge-withdrawer/src/withdrawer/ethereum/watcher.rs
```diff
@@ -430,7 +435,7 @@ mod tests {
 
     #[tokio::test]
     #[ignore = "requires foundry and solc to be installed"]
```
joroshiba, line 437: doesn't need solc anymore right?

crates/astria-bridge-withdrawer/ethereum/src/AstriaBridgeableERC20.sol
noot: payable means you can send the native asset to the contract using that function, we don't want that here

crates/astria-bridge-withdrawer/build.rs
noot: don't think alloy has abigen yet

crates/astria-bridge-withdrawer/src/withdrawer/ethereum/watcher.rs
noot: no, will change the message

crates/astria-bridge-withdrawer/ethereum/src/AstriaBridgeableERC20.sol
noot: the tx will revert, the tokens are fine. same for native assets

joroshiba (approved):

PR #1157, opened 2024-06-05
fix(withdrawer): use block subscription in batcher; send to destination_chain_address

noot:
## Summary
fixes found during testing:
1. use block subscription in batcher. previously, txs were only sent out of the batcher when a new event occurred with a higher block number, which is wrong. it should have been sending upon any new block being built.
2. set recipient in `BridgeUnlock` to `destination_chain_address` instead of event `sender`.

## Background
bugs!!

## Changes
- use block subscription in batcher and send event batch upon new block being received
-  set recipient in `BridgeUnlock` to `destination_chain_address` instead of event `sender`.

## Testing
unit tests


crates/astria-bridge-withdrawer/src/withdrawer/ethereum/convert.rs
```diff
@@ -95,7 +95,7 @@ fn event_to_bridge_unlock(
         transaction_hash,
     };
     let action = BridgeUnlockAction {
-        to: event.sender.to_fixed_bytes().into(),
+        to: event.destination_chain_address.to_fixed_bytes().into(),
```
joroshiba, line 98: Should we do validation on the length like we do with the sender for return address with ICS withdrawals?

crates/astria-bridge-withdrawer/src/withdrawer/ethereum/watcher.rs
```diff
@@ -250,6 +260,32 @@ impl Batcher {
...
+                                    .wrap_err("failed to send batched events; receiver dropped?")?;
+                            }
+
+                            curr_batch = Batch {
+                                actions: Vec::new(),
+                                rollup_height: block_number.as_u64(),
+                            };
+                        }
+                    } else {
+                        warn!("block stream closed; shutting down batcher");
```
joroshiba, line 285: should this be a warn or an error?

crates/astria-bridge-withdrawer/src/withdrawer/ethereum/watcher.rs
```diff
@@ -250,6 +260,32 @@ impl Batcher {
                     info!("batcher shutting down");
                     break;
                 }
+                block = block_rx.next() => {
```
joroshiba, line 263: with adding this, should we remove the functionality which sends on receiving an event that occurs in a new block. It seems very duplicative? New events push into the batches, new blocks send them?

crates/astria-bridge-withdrawer/src/withdrawer/ethereum/watcher.rs
noot: i'll change to error

crates/astria-bridge-withdrawer/src/withdrawer/ethereum/watcher.rs
noot: i think we need to leave the event logic which pushes on new block, for example, if we get an event that's for a newer block than the current batch before we get the notification of the new block. then, the batch will end up containing events from multiple block numbers, which is not what we want

PR #1158, opened 2024-06-05
feat(sequencer): implement `bridge/account_last_tx_hash` abci query

### LINKED ISSUE #1107: sequencer: Add endpoint for last block height in which a transaction was made for bridge account
Following the bridge withdrawer discussion, we want an optional endpoint (gated by a flag on startup or something) that would provide the last height in which a transaction was made by a bridge account. This is for making syncing the withdrawer much easier.

noot:
## Summary
 implement `bridge/account_last_tx_hash` abci query which returns the last executed tx from the given bridge account.

## Background
required for the bridge withdrawer's sync logic (or rather, makes it much simpler to implement).

## Changes
- if a bridge account makes a tx, store this tx hash in the nonconsensus storage.
- implement `bridge/account_last_tx_hash` abci query which queries the nonconsensus storage for this tx hash given some address. 
- TODO: we can make this a config value, so only nodes that are configured to store this data store it. 

## Testing
unit tests

## Related Issues

closes #1107

### DIFF
--- a/crates/astria-sequencer/src/service/info/mod.rs
+++ b/crates/astria-sequencer/src/service/info/mod.rs
@@ -64,6 +64,12 @@ impl Info {
                 crate::asset::query::allowed_fee_asset_ids_request,
             )
             .context("invalid path: `asset/allowed_fee_asset_ids`")?;
+        query_router
+            .insert(
+                "bridge/account_last_tx_hash/:address",
+                crate::bridge::query::bridge_account_last_tx_hash_request,
+            )
+            .context("invalid path: `bridge/account_last_tx_hash/:address`")?;
         Ok(Self {
             storage,
             query_router,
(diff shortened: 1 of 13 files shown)


crates/astria-sequencer/src/service/info/mod.rs
```diff
@@ -64,6 +64,12 @@ impl Info {
                 crate::asset::query::allowed_fee_asset_ids_request,
             )
             .context("invalid path: `asset/allowed_fee_asset_ids`")?;
+        query_router
+            .insert(
+                "bridge/account_last_tx_hash/:address",
```
joroshiba, line 69: Not blocking here, because this matches existing pattern but I'd prefer to have resource name style path based queries is `/resource-type/{id}/sub-resource-type/{id}`. In this instance that would be `bridge/{address}/last_tx_hash` so it's clear who owns what property hierarchically. The `last_tx_hash` is a property of the address, which is a bridge.

This would apply to all our queries though.

joroshiba (approved):

PR #1161, opened 2024-06-07
feat(sequencer): implement refund to rollup logic upon ics20 transfer refund

### LINKED ISSUE #1110: sequencer: Add logic for handling reversion of IBC-based rollup withdrawals
ICS20 withdrawals currently don't have a method of refunding rollup users in case of failure. This is needed for bridge accounts to process IBC-based rollup-user withdrawals (e.g. rollup->celestia).

noot:
## Summary
implement refund to rollup logic upon ics20 transfer refund, if the transfer originally can from an Ics20Withdrawal that corresponded to a rollup withdrawal

## Background
we want to users to get their funds back in on the rollup automatically if they bridge out via IBC and it fails. currently, if bridging out from the rollup via IBC fails, their funds are returned on the sequencer, which would require the user to then manually do another `BridgeLock` if they want their funds back on the rollup. 

## Changes
- implement logic in the ics20 transfer module which detects if a transfer is a refund originating from a rollup withdrawal, and if it is, transfers the funds to the specified bridge account and emits a `Deposit` event. 
- update the withdrawer to be aware of the bridge account on the sequencer which corresponds to the rollup asset it's bridging, and it needs to place this in the withdraw memo in case of failure.

## Testing
unit tests

## Related Issues

closes #1110


crates/astria-bridge-withdrawer/local.env.example
```diff
@@ -35,7 +35,11 @@ ASTRIA_BRIDGE_WITHDRAWER_SEQUENCER_KEY_PATH=/path/to/priv_sequencer_key.json
 ASTRIA_BRIDGE_WITHDRAWER_FEE_ASSET_DENOMINATION="nria"
 
 # The asset denomination being withdrawn from the rollup.
-ASTRIA_BRIDGE_WITHDRAWER_ROLLUP_ASSET_DENOMINATION="nria",
+ASTRIA_BRIDGE_WITHDRAWER_ROLLUP_ASSET_DENOMINATION="nria"
+
+# The hex-encoded bridge address corresponding to the bridged rollup asset on the sequencer.
+# Should match the bridge address in the geth rollup's bridge configuration for that asset.
+ASTRIA_BRIDGE_WITHDRAWER_SEQUENCER_BRIDGE_ADDRESS=""
```
noot, line 42: i want to add these as immutable vars to the bridge contract instead of config values, will do in a follow up.

joroshiba (approved):

PR #1162, opened 2024-06-07
fix(core, sequencer): prefix removal source non-refund ics20 packet

SuperFluffy:
## Summary
Rewrites the IBC denomation construction by parsing its prefix into segments. This fixes the incorrect removal of prefxies for source non-refund ics20 packets.

## Background
follow up to https://github.com/astriaorg/astria/pull/851 which was an audit fix - this fixes the fix by removing the source prefix correctly, in the case assets we're the source of are transferred in. previously, the entire prefix was removed, would would have been invalid if the asset was prefixed by multiple hops, eg. transfer/channel-1/transfer/channel-2/denom would have become just denom when it should have become transfer/channel-2/denom.

## Changes
- refactor `astria_core::primitive::1::asset::Denom` to express its prefix as a list of segments
- ensure that only full prefix segments are removed
- make `Denom` construction a fallible operation by parsing it
- removed a bunch of constructors as they were only used in tests and not otherwise

## Testing
Provide exhaustive units.

## Related Issues
Replaces https://github.com/astriaorg/astria/pull/1131

### DIFF
--- a/crates/astria-bridge-withdrawer/src/withdrawer/ethereum/convert.rs
+++ b/crates/astria-bridge-withdrawer/src/withdrawer/ethereum/convert.rs
@@ -131,9 +131,8 @@ fn event_to_ics20_withdrawal(
     let sender = event.sender.to_fixed_bytes();
     let denom = rollup_asset_denom.clone();
 
-    let (_, channel) = denom
-        .prefix()
-        .rsplit_once('/')
+    let channel = denom
+        .channel()
         .ok_or_eyre("denom must have a channel to be withdrawn via IBC")?;
 
     let memo = Ics20WithdrawalFromRollupMemo {
@@ -189,12 +188,14 @@ fn calculate_packet_timeout_time(timeout_delta: Duration) -> eyre::Result<u64> {
 
 #[cfg(test)]
 mod tests {
+    use asset::default_native_asset;
+
     use super::*;
     use crate::withdrawer::ethereum::astria_withdrawer_interface::SequencerWithdrawalFilter;
 
     #[test]
     fn event_to_bridge_unlock() {
-        let denom = Denom::from("nria".to_string());
+        let denom = default_native_asset();
         let event_with_meta = EventWithMetadata {
             event: WithdrawalEvent::Sequencer(SequencerWithdrawalFilter {
                 sender: [0u8; 20].into(),
@@ -238,7 +239,7 @@ mod tests {
 
     #[test]
     fn event_to_bridge_unlock_divide_value() {
-        let denom = Denom::from("nria".to_string());
+        let denom = default_native_asset();
         let event_with_meta = EventWithMetadata {
             event: WithdrawalEvent::Sequencer(SequencerWithdrawalFilter {
                 sender: [0u8; 20].into(),
@@ -283,7 +284,7 @@ mod tests {
 
     #[test]
     fn event_to_ics20_withdrawal() {
-        let denom = Denom::from("transfer/channel-0/utia".to_string());
+        let denom = "transfer/channel-0/utia".parse::<Denom>().unwrap();
         let destination_chain_address = "address".to_string();
         let event_with_meta = EventWithMetadata {
             event: WithdrawalEvent::Ics20(Ics20WithdrawalFilter {
--- a/crates/astria-core/src/primitive/v1/asset.rs
+++ b/crates/astria-core/src/primitive/v1/asset.rs
@@ -1,18 +1,39 @@
-use std::{
-    fmt,
-    fmt::{
-        Display,
-        Formatter,
-    },
+use std::fmt::{
+    self,
+    Display,
+    Formatter,
+    Write,
 };
 
 /// The default sequencer asset base denomination.
 pub const DEFAULT_NATIVE_ASSET_DENOM: &str = "nria";
 
 /// Returns the default sequencer asset ID.
+// allow: parsing single-segment assets is unit tested
+#[allow(clippy::missing_panics_doc)]
+// allow: used in many places already
+#[allow(clippy::module_name_repetitions)]
+#[must_use]
+pub fn default_native_asset() -> Denom {
+    DEFAULT_NATIVE_ASSET_DENOM
+        .parse::<Denom>()
+        .expect("parsing a single segment string must work")
+}
+
+/// Returns the default sequencer asset ID.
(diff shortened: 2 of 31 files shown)


crates/astria-bridge-withdrawer/src/withdrawer/ethereum/convert.rs
```diff
@@ -129,9 +129,8 @@ fn event_to_ics20_withdrawal(
     let sender = event.sender.to_fixed_bytes().into();
     let denom = rollup_asset_denom.clone();
 
-    let (_, channel) = denom
-        .prefix()
-        .rsplit_once('/')
+    let channel = denom
+        .channel()
```
SuperFluffy, line 133: Instead of doing this ad-hoc I created a method for this.

crates/astria-core/src/primitive/v1/asset.rs
```diff
@@ -59,61 +69,163 @@ impl Denom {
     }
 
     #[must_use]
-    pub fn prefix(&self) -> &str {
-        &self.prefix
+    pub fn channel(&self) -> Option<&str> {
+        if self.prefix_segments.len() < 2 {
```
SuperFluffy, line 73: I am not sure if this is always true. But I went with how this was set in `astria-bridge-withdrawer`

crates/astria-core/src/primitive/v1/asset.rs
```diff
@@ -59,61 +69,162 @@ impl Denom {
...
+        self.prefix_segments.last().map(|seg| &**seg)
     }
 
     #[must_use]
     pub fn prefix_is(&self, prefix: &str) -> bool {
-        self.prefix == prefix
+        // For the current implementation, the number of slashes in a prefix is equal to the number
+        // of segments + 1.
+        // So a "path/to" would be a prefix to `"path/to/denom"`, but `"path/to/"` would not
+        let number_of_slashes = prefix.chars().filter(|c| *c == '/').count();
```
SuperFluffy, line 84: This just short-circuits to avoid having to walk over the entire slice.

crates/astria-core/src/primitive/v1/asset.rs
```diff
@@ -1,18 +1,39 @@
...
 #[must_use]
 pub fn default_native_asset_id() -> Id {
-    Denom::from_base_denom(DEFAULT_NATIVE_ASSET_DENOM).id()
+    DEFAULT_NATIVE_ASSET_DENOM
+        .parse::<Denom>()
+        .expect("parsing a single segment string must work")
+        .id()
+}
+
+struct DenomFmt<'a> {
```
SuperFluffy, line 34: We use this for constructing the id ad-hoc and for formatting.

crates/astria-core/src/primitive/v1/asset.rs
```diff
@@ -137,6 +252,13 @@ impl Id {
         Self(hash.into())
     }
 
+    fn from_denom_fmt(denom_fmt: &DenomFmt<'_>) -> Self {
```
SuperFluffy, line 255: This is just a quick fix and should be replaced by a non-allocating implementation.

Also, the `Id::from_denom` above is should actually take a `Denom`, not a `str`.

Fraser999 (approved):
LGTM - just a few nitpicks.

crates/astria-core/src/protocol/account/v1alpha1/mod.rs
```diff
@@ -48,21 +70,40 @@ impl raw::BalanceResponse {
...
+    //     BalanceResponse::try_from_raw(&self)
+    // }
+
+    // /// Converts a protobuf [`raw::BalanceResponse`] to an astria
+    // /// native [`BalanceResponse`] by allocating a new [`v1alpha::BalanceResponse`].
+    // #[must_use]
+    // pub fn try_to_native(&self) -> Result<BalanceResponse, BalanceResponseError> {
+    //     BalanceResponse::try_from_raw(self)
+    // }
+}
```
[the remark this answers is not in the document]
SuperFluffy: Yeah, addressed in  291126

crates/astria-core/src/protocol/asset/v1alpha1/mod.rs
```diff
@@ -50,19 +71,19 @@ impl raw::DenomResponse {
...
-    #[must_use]
-    pub fn into_native(self) -> DenomResponse {
-        DenomResponse::from_raw(&self)
-    }
+    // /// Converts a protobuf [`raw::DenomResponse`] to an astria
+    // /// native [`DenomResponse`].
+    // #[must_use]
+    // pub fn try_into_native(self) -> Result<DenomResponse, DenomResponseError> {
+    //     DenomResponse::try_from_raw(&self)
+    // }
```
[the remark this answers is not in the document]
SuperFluffy: Yes, address by 291126

crates/astria-core/src/primitive/v1/asset.rs
```diff
@@ -1,18 +1,39 @@
...
+    self,
+    Display,
+    Formatter,
+    Write,
 };
 
 /// The default sequencer asset base denomination.
 pub const DEFAULT_NATIVE_ASSET_DENOM: &str = "nria";
 
 /// Returns the default sequencer asset ID.
```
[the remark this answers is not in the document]
SuperFluffy: Changed to ``[`Denom`]`` in  8390b4

crates/astria-core/src/primitive/v1/asset.rs
```diff
@@ -1,18 +1,39 @@
...
+#[must_use]
+pub fn default_native_asset() -> Denom {
+    DEFAULT_NATIVE_ASSET_DENOM
+        .parse::<Denom>()
+        .expect("parsing a single segment string must work")
+}
+
+/// Returns the default sequencer asset ID.
+// allow: parsing single-segment assets is unit tested
+#[allow(clippy::missing_panics_doc)]
```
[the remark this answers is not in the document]
SuperFluffy: I removed the entire `default_native_asset_id` function in favor of just doing `default_native_asset().id()` in 291126

crates/astria-core/src/primitive/v1/asset.rs
noot: this should be the second prefix segment actually (the channel closest to the start of the denom trace)

PR #1165, opened 2024-06-08
feat(bridge-withdrawer): sync logic

### LINKED ISSUE #1108: bridge-withdrawer: Add sync logic
As discussed in the context of the bridge withdrawer design, we need to add logic for syncing the withdrawer to the current state of the rollup and sequencer. The withdrawer needs to:
1. look at sequencer state to figure out what the last batch of withdrawals it successfully processed is
2. figure out the corresponding rollup block height
3. start processing from there onwards

itamarreif:
## Summary
This checks the last sequencer transaction made by the bridge account for the corresponding rollup block height in order to initialize the watcher's event stream from the right block

## Background
The bridge withdrawer needs to know what the last finalized processed block was.

## Changes
- [x] submitter startup will fetch the last finalized tx and parse the rollup block out of it
- [x] watcher gets the initial rollup block height to listen to from the submitter's startup

## Testing
The sync is part of the startup sequence, which has existing tests already in place. This PR mounts the additional mock guards required by the sync process to ensure all the existing tests still work.

closes #1108


noot (approved):
lgtm!

PR #1167, opened 2024-06-09
chore: bump msrv and run clippy

noot:
## Summary
latest rust is now 1.79 and also our code was using items supported in rust 1.76 and the ci is using 1.76.0, so the msrv should be bumped:
```
error: current MSRV (Minimum Supported Rust Version) is `1.73.0` but this item is stable since `1.76.0`
   --> crates/astria-conductor/src/celestia/convert.rs:151:10
    |
151 |           .inspect_err(|err| {
    |  __________^
152 | |             info!(
153 | |                 error = err as &StdError,
154 | |                 target = SubmittedMetadataList::full_name(),
155 | |                 "failed decoding blob bytes; dropping the blob",
156 | |             );
157 | |         })
    | |__________^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#incompatible_msrv
```

also ran clippy and fixed the various lint issues.


Fraser999 (approved):
LGTM.  Should we bump [the Rust version for clippy in CI](https://github.com/astriaorg/astria/blob/2d79f7defffae8be5ea656da78bff8711c675794/.github/workflows/test.yml#L176) to 1.78.0 so we don't regress?

crates/astria-composer/src/config.rs
```diff
@@ -22,6 +22,7 @@ pub struct Config {
     /// The chain ID of the sequencer chain
     pub sequencer_chain_id: String,
 
+    #[allow(clippy::doc_markdown)]
```
SuperFluffy, line 25: Replace in favor of putting ticks around it, i.e.:
```suggestion
    /// A list of `<rollup_name>::<url>` pairs
```

crates/astria-sequencer-relayer/src/validator.rs
```diff
@@ -13,7 +13,7 @@ use tracing::instrument;
 #[derive(Clone, Debug)]
 pub(crate) struct Validator {
     /// The tendermint validator account address; defined as
-    /// Sha256(verification_key)[..20].
+    /// Sha256(`verification_key`)[..20].
```
SuperFluffy, line 16: That's going to look very strange:
```suggestion
    /// `Sha256(verification_key)[..20]`.
```

SuperFluffy (approved):
We should have an msrv check running on commit to ensure the toml msrv is correct.

Changes look good, except for a few markdown uses that could be made more code-y :)

aajimal (approved):

PR #1173, opened 2024-06-10
chore(chart)!: update default cometbft config vals

joroshiba:
## Summary
Updates default CometBFT configuration, and genesis.

## Background
We have been running on mostly defaults without much thought to it, gave a more thorough review.

## Breaking Changelist
- Changes genesis parameters and thus cannot be used with existing networks without adding ability to make more configurable w/ these as defaults.

### DIFF
--- a/charts/sequencer/files/cometbft/config/config.toml
+++ b/charts/sequencer/files/cometbft/config/config.toml
@@ -169,10 +169,10 @@ experimental_close_on_slow_client = false
 timeout_broadcast_tx_commit = "10s"
 
 # Maximum size of request body, in bytes
-max_body_bytes = 1000000
+max_body_bytes = 1_000_000
 
 # Maximum size of request header, in bytes
-max_header_bytes = 1048576
+max_header_bytes = 1_048_576
 
 # The path to a file containing certificate that is used to create the HTTPS server.
 # Might be either absolute path or path related to CometBFT's config directory.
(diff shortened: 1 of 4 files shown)


charts/sequencer/files/cometbft/config/config.toml
```diff
@@ -297,15 +297,15 @@ broadcast = true
...
 
 # Maximum number of transactions in the mempool
-size = 5000
+size = 2000
 
 # Limit the total size of all txs in the mempool.
 # This only accounts for raw transactions (e.g. given 1MB transactions and
 # max_txs_bytes=5MB, mempool will only accept 5 transactions).
-max_txs_bytes = 1073741824
+max_txs_bytes = 100_000_000
```
noot, line 305: why the change here?

charts/sequencer/files/cometbft/config/genesis.json
```diff
@@ -46,12 +46,12 @@
...
     "block": {
-      "max_bytes": "22020096",
+      "max_bytes": "1000000",
       "max_gas": "-1"
     },
     "evidence": {
-      "max_age_duration": "172800000000000",
-      "max_age_num_blocks": "100000",
+      "max_age_duration": "1209600000000000",
+      "max_age_num_blocks": "4000000",
```
noot, line 54: same here

charts/sequencer/files/cometbft/config/config.toml
quasystaty1: Decreasing mempool size from 1GB to 100 MB, better default configs, same for other values.

joroshiba:
Fixed as a part of  https://github.com/astriaorg/astria/pull/1193

PR #1175, opened 2024-06-10
feat(composer)!: Add chain_id check on executor build

eoroshiba:
## Summary
Added check to ensure the executor's configured chain_id and the sequencer client's actual chain_id match.

## Background
This check ensures that the composer is communicating with the sequencer RPC on the correct network.

## Changes
Added chain_id check to the composer executor builder.

Made executor builder async to accommodate for async communication with sequencer client.

Moved nonce guard declaration from mock startup to individual tests in executor tests.

Added function to mount a status response with a given chain_id to the mock. This is so the executor builder can gather the status response without throwing an HTTP 404 error.

## Testing
Added executor test to validate the failure of an executor build with a chain_id mismatch.

## Related Issues
Part of #986


crates/astria-composer/src/executor/builder.rs
```diff
@@ -36,7 +38,7 @@ pub(crate) struct Builder {
 }
 
 impl Builder {
-    pub(crate) fn build(self) -> eyre::Result<(super::Executor, executor::Handle)> {
+    pub(crate) async fn build(self) -> eyre::Result<(super::Executor, executor::Handle)> {
```
SuperFluffy, line 41: Please don't make the builder async. If one needs config-like input, then move this into the `run`/`run_until_stopped` function and run it with a retry.

crates/astria-composer/tests/blackbox/helper/mock_sequencer.rs
```diff
@@ -57,3 +59,62 @@ pub async fn mount_abci_query_mock(
...
+    Mock::given(body_partial_json(json!({"method": "status"})))
+        .respond_with(ResponseTemplate::new(200).set_body_json(response))
+        .up_to_n_times(1)
+        .expect(1..)
+        .named("CometBFT status")
+        .mount_as_scoped(server)
+        .await
+}
+
+const STATUS_RESPONSE: &str = r#"
```
SuperFluffy, line 82: Please construct this object ad-hoc instead of a fixed string.

crates/astria-composer/tests/blackbox/helper/mock_sequencer.rs
```diff
@@ -57,3 +59,62 @@ pub async fn mount_abci_query_mock(
         .mount_as_scoped(server)
         .await
 }
+
+async fn mount_cometbft_status_response(
+    server: &MockServer,
+    mock_sequencer_chain_id: &str,
+) -> MockGuard {
```
SuperFluffy, line 66: No need to return the mock guard into the builder. There isn't actually something to assert on other that the mock guard was triggered. Just mount it non-scoped and don't return a guard.

SuperFluffy (changes requested):

crates/astria-composer/tests/blackbox/helper/mock_sequencer.rs
eoroshiba: Since status::Response has no default configuration, I'm not sure how to get around doing an fixed declaration of the response somewhere. As of now, I've just moved the construction of the response inside the mount function. Also worth noting that this is copied from the status response mount for the sequencer relayer, so any change made here may need to be reflected there as well. Do you have a preferred solution in mind for the fixed strings?

crates/astria-composer/src/composer.rs
```diff
@@ -240,6 +240,7 @@ impl Composer {
         // run the collectors and executor
         spawn_geth_collectors(&mut geth_collectors, &mut geth_collector_tasks);
 
+        let _chain_id_result = executor.check_chain_ids().await.wrap_err("chain_id mismatch");
```
SuperFluffy, line 243: This performs a check but there is no actual effect on the rest of the code. Effectively, this just performs a get and wraps the return value into a `Result`. You need to have some more logic (like returning early) to have an effect.

Having said that, move this into `Executor::run_until_stopped`. The composer task acts more like a collection of sub tasks, while the sub tasks should be self-contained (including verification of chain IDs).

crates/astria-composer/src/executor/mod.rs
```diff
@@ -415,6 +419,23 @@ impl Executor {
 
         reason.map(|_| ())
     }
+
+    // check for mismatched configured chain_id and sequencer client chain_id
+    pub(crate) async fn check_chain_ids(&self) -> eyre::Result<()> {
+        let client_response = self
+            .sequencer_client
+            .status()
+            .await
```
SuperFluffy, line 428: This future should be wrapped in a retry because composer is likely to start way earlier then sequencer and cometbft. This means that you will a bunch of crash-look-backoff errors in kubernetes.

See conductor for doing essentially the same thing: https://github.com/astriaorg/astria/blob/c31c10b6db41b19fdce8e4b172363f855d88be23/crates/astria-conductor/src/celestia/mod.rs#L645

crates/astria-composer/src/executor/mod.rs
```diff
@@ -415,6 +419,23 @@ impl Executor {
...
+            .wrap_err("failed to retrieve sequencer network status")?;
+        let client_chain_id = client_response.node_info.network.to_string();
+        let configured_chain_id = self.sequencer_chain_id.clone();
```
SuperFluffy, line 431: You can avoid the clones/allocations. The `Genesis` return value has a `chain_id` field that provides a `as_str()` method:
https://docs.rs/tendermint/0.37.0/tendermint/genesis/struct.Genesis.html
https://docs.rs/tendermint/0.37.0/tendermint/chain/id/struct.Id.html#method.as_str

The check below would then become:
```rust
ensure!(
    &self.sequencer_chain_id == genesis.chain_id.as_str(), "<text>"
);
```

crates/astria-composer/tests/blackbox/helper/mock_sequencer.rs
SuperFluffy: You just have to manually fill in the blanks. :)

If you use the genesis RPC instead of status (which I think is more reasonable), then you can just copy what we have in conductor:
https://github.com/astriaorg/astria/blob/c31c10b6db41b19fdce8e4b172363f855d88be23/crates/astria-conductor/tests/blackbox/helpers/mod.rs#L287

eoroshiba: Should be resolved in #[f3c96ff]. Thanks for all the feedback :)

crates/astria-composer/src/composer.rs
eoroshiba: Resolved in f3c96ff

crates/astria-composer/src/executor/mod.rs
```diff
@@ -415,6 +419,23 @@ impl Executor {
 
         reason.map(|_| ())
     }
+
+    // check for mismatched configured chain_id and sequencer client chain_id
+    pub(crate) async fn check_chain_ids(&self) -> eyre::Result<()> {
+        let client_response = self
+            .sequencer_client
+            .status()
```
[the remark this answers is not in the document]
eoroshiba: Resolved in f3c96ff.

crates/astria-composer/src/executor/mod.rs
```diff
@@ -415,6 +419,23 @@ impl Executor {
 
         reason.map(|_| ())
     }
+
+    // check for mismatched configured chain_id and sequencer client chain_id
+    pub(crate) async fn check_chain_ids(&self) -> eyre::Result<()> {
```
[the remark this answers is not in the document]
eoroshiba: Done in f3c96ff.

crates/astria-composer/src/executor/mod.rs
eoroshiba: Resolved in f3c96ff.

crates/astria-composer/src/executor/mod.rs
eoroshiba: Done in f3c96ff.

crates/astria-composer/src/executor/builder.rs
```diff
@@ -15,7 +15,7 @@ use astria_core::{
 use astria_eyre::eyre::{
     self,
     eyre,
-    WrapErr as _,
```
joroshiba, line 18: unclear as to why this change is here to me

crates/astria-composer/src/executor/tests.rs
```diff
@@ -395,8 +468,18 @@ async fn two_seq_actions_single_bundle() {
...
+        "accounts/nonce",
+        NonceResponse {
+            height: 0,
+            nonce: 0,
+        },
+    )
+    .await;
     let status = executor.subscribe();
 
+    // executor.check_chain_ids().await.unwrap();
```
joroshiba, line 482: remove

crates/astria-composer/src/executor/tests.rs
```diff
@@ -395,8 +468,18 @@ async fn two_seq_actions_single_bundle() {
...
 
+    let nonce_guard = mount_nonce_query_mock(
+        &sequencer,
+        "accounts/nonce",
+        NonceResponse {
+            height: 0,
+            nonce: 0,
+        },
+    )
+    .await;
```
joroshiba, line 479: nit: this is repeated a number of times perhaps a helper that does exactly this `mount_default_nonce_query_guard` function?

crates/astria-composer/src/executor/mod.rs
```diff
@@ -415,6 +423,41 @@ impl Executor {
...
+                },
+            );
+        let client_genesis: tendermint::Genesis =
+            tryhard::retry_fn(|| self.sequencer_client.genesis())
+                .with_config(retry_config)
+                .await
+                .wrap_err("failed to retrieve sequencer genesis after many attempts")?;
+        ensure!(
+            self.sequencer_chain_id == client_genesis.chain_id.as_str(),
+            "mismatch in configured chain_id: {0} and sequencer chain_id: {1}",
```
joroshiba, line 455: nit: this is accurate but concerned the message might not give enough context information to operator to understand how configuration is wrong.

Something like: 
`mismatch in chain id: config specifies {0}, but sequencer rpc is for {1}`.

crates/astria-composer/src/executor/mod.rs
```diff
@@ -415,6 +423,41 @@ impl Executor {
 
         reason.map(|_| ())
     }
+
+    // check for mismatched configured chain_id and sequencer client chain_id
+    pub(crate) async fn ensure_configured_chain_id_matches_remote(&self) -> eyre::Result<()> {
```
joroshiba, line 428: perhaps simplify with `get_sequencer_chain_id` and migrate the ensure to the `run_until_stopped` the `get_sequencer_chain_id` function could also not be a method taking self but a function taking just a sequencer client (see `get_latest_nonce` as example) ?

The function title is a reason this can fail, but it can also fail for errors in grabbing genesis, feels like the function has too many responsibilities to me, don't think the ensure necessarily needs to be it's own method if its "grab the data needed then ensure it's good"

crates/astria-composer/src/executor/builder.rs
```diff
@@ -48,6 +48,7 @@ impl Builder {
         } = self;
         let sequencer_client = sequencer_client::HttpClient::new(sequencer_url.as_str())
             .wrap_err("failed constructing sequencer client")?;
+
```
joroshiba, line 51: nit: don't think this file needs any changes?

crates/astria-composer/src/executor/mod.rs
```diff
@@ -192,6 +196,10 @@ impl Executor {
     /// An error is returned if connecting to the sequencer fails.
     #[instrument(skip_all, fields(address = %self.address))]
     pub(super) async fn run_until_stopped(mut self) -> eyre::Result<()> {
+        let _chain_id_result = self
```
joroshiba, line 199: I don't believe this needs to be assigned to anything as written, but bigger note down on mild update which can be made

crates/astria-composer/src/executor/builder.rs
eoroshiba: Should be resolved with be387af.

crates/astria-composer/src/executor/tests.rs
eoroshiba: Done in be387af.

crates/astria-composer/src/executor/tests.rs
eoroshiba: Done in be387af.

crates/astria-composer/src/executor/mod.rs
eoroshiba: Done in be387af.

crates/astria-composer/src/executor/mod.rs
eoroshiba: Done in b8d148f.

crates/astria-composer/src/executor/builder.rs
eoroshiba: Fixed in b8d148f.

crates/astria-composer/src/executor/mod.rs
eoroshiba: Resolved in b8d148f.

crates/astria-composer/src/executor/builder.rs
SuperFluffy: Undo the change, use `WrapErr as _`

crates/astria-composer/src/executor/mod.rs
```diff
@@ -199,6 +203,16 @@ impl Executor {
     /// An error is returned if connecting to the sequencer fails.
     #[instrument(skip_all, fields(address = %self.address))]
     pub(super) async fn run_until_stopped(mut self) -> eyre::Result<()> {
+        let remote_chain_id = self
```
SuperFluffy, line 206: IMO, factor this into a method `async fn setup` or `async fn init` or `async fn pre_run_checks` and perform the checks there.

From inside `run_until_stopped` you can then do something like below so that your task is a good citizen and doesn't just ignore the shutdown signal during setup.


```rust
select!(
    biased;
        info!("received shutdown signal while running initialization routines; exiting");
        return Ok(());
    }

    res = self.setup() => {
        if let Err(e) = res {
            bail!("runtime initialization failed");
        }
);
```

crates/astria-composer/src/executor/mod.rs
```diff
@@ -417,6 +431,35 @@ impl Executor {
 
         reason.map(|_| ())
     }
+
+    // check for mismatched configured chain_id and sequencer client chain_id
```
SuperFluffy, line 435: This comment is wrong. You are not performing that check in here.

Also, if you want to add doc comments, use triple backticks: `/// <doc-string>`

crates/astria-composer/src/executor/mod.rs
```diff
@@ -417,6 +431,35 @@ impl Executor {
...
+                    );
+                    futures::future::ready(())
+                },
+            );
+        let client_genesis: tendermint::Genesis =
+            tryhard::retry_fn(|| self.sequencer_client.genesis())
+                .with_config(retry_config)
+                .await
+                .wrap_err("failed to retrieve sequencer genesis after many attempts")?;
+        Ok(client_genesis.chain_id.to_string())
```
SuperFluffy, line 461: The conversion is unnecessary, just return `genesis.chain_id`

crates/astria-composer/src/executor/tests.rs
```diff
@@ -183,6 +186,53 @@ async fn mount_broadcast_tx_sync_seq_actions_mock(server: &MockServer) -> MockGu
         .await
 }
 
+/// Mounts a `CometBFT` status response with a specified mock sequencer chain id
+async fn mount_genesis(server: &MockServer, mock_sequencer_chain_id: &str) {
```
SuperFluffy, line 190: For a followup: Sucks that we need this here. We should just rip out all of these wrongly placed pseudo-blackbox tests and just rely on `tests/blackbox`.

crates/astria-composer/src/executor/mod.rs
```diff
@@ -199,6 +203,16 @@ impl Executor {
     /// An error is returned if connecting to the sequencer fails.
     #[instrument(skip_all, fields(address = %self.address))]
     pub(super) async fn run_until_stopped(mut self) -> eyre::Result<()> {
+        let remote_chain_id = self
+            .get_sequencer_chain_id()
+            .await
+            .wrap_err("failed obtain sequencer chain_id")?;
+        ensure!(
+            self.sequencer_chain_id == remote_chain_id,
```
SuperFluffy, line 211: You `remote_chain_id.as_str()` from https://docs.rs/tendermint/latest/tendermint/chain/id/struct.Id.html (important due to my suggestion to just return `tendermint::chain::Id` withot allocating).

crates/astria-composer/src/executor/mod.rs
```diff
@@ -199,6 +203,16 @@ impl Executor {
...
+            "mismatch in configured chain_id: config specifies {0}, but sequencer rpc is for {1}",
```
SuperFluffy, line 212: You don't need the numbers, and you can rely on interpolation (also I recommend backticks to make it easier for a human to parse the text). I also suggest to remove mention of "config" - it's not not wrong, but this part of the code has no concept of the actual config, so I'd leave it more generic:
```suggestion
            "mismatch in Sequencer chain IDs: expected `{}`, but Sequencer returned `{remote_chain_id}`",
            self.sequencer_chain_id,
```

crates/astria-composer/src/executor/mod.rs
```diff
@@ -40,7 +41,10 @@ use futures::{
 use pin_project_lite::pin_project;
 use prost::Message as _;
 use sequencer_client::{
-    tendermint_rpc::endpoint::broadcast::tx_sync,
+    tendermint_rpc::{
+        endpoint::broadcast::tx_sync,
+        Client,
```
SuperFluffy, line 46:
```suggestion
        Client as _,
```

crates/astria-composer/src/executor/tests.rs
```diff
@@ -480,3 +538,33 @@ async fn two_seq_actions_single_bundle() {
         );
     }
 }
+
+/// Test to check that executor's configured sequencer chain id and sequencer's actual chain id
+/// match
+#[tokio::test]
+async fn should_exit_if_mismatch_sequencer_chain_id() {
```
SuperFluffy, line 545: So this test is dangerous because it could silently break: somebody might break the chain-ID check in a refactor, but then subsequently add code that still causes this method to exit with an error. Then, under test conditions, everything looks peachy, even though we have a problem.

IMO the most straightforward way to fix this is:

1. refactor the check into a free-standing function like so:

```rust
enum EnsureChainIdError {
    // <other error variants from getting genesis, etc>
    #[error("expected chain ID `{expected}`, but received `{actual}`)]
    WrongChainId { expected: String, actual: tendermint::chain::Id, }
}

fn ensure_chain_id_is_correct(client: tendermint::HttpClient, expected_chain_id: String) -> Result<(), EnsureChainIdError> {
    // code to get the chain ID
    if remote_chain_id != expected_chain_id {
        return Err(WrongChainId { expected: expected_chain_id, actual: remote_chain_id});
    }
    Ok(())
}
```

You can then actually check the function without setting up the rest, which makes things a bit easier, i.e. you could check it like so:
```rust
match err {
    EnsureChainIdError::WrongChainId { expected, actual } => {
        assert_eq!(expected, <configured-expected>);
        assert_eq!(actual, <configured-actual>);
    }
    other => panic!("expected `EnsureChainIdError::WrongChainId`, but got `{other}`),
}
```


If you still want to test `run_until_stopped` (like above, just some more code):

```rust
// <code as it's written right now>
let err = exector.run_until_stopped().await.expect_err("should exit with an error when reading a bad chain ID, but exited with success");

for cause in err.chain() {
    if let Some(err) = cause.downcast_ref::<EnsureChainIdError>() {
        match err {
            EnsureChainIdError::WrongChainId { expected, actual } => {
                assert_eq!(expected, <configured-expected>);
                assert_eq!(actual, <configured-actual>);
            }
            other => panic!("expected `EnsureChainIdError::WrongChainId`, but got `{other}`),
     }
}
```

crates/astria-composer/src/executor/tests.rs
```diff
@@ -319,8 +374,10 @@ async fn bundle_triggered_by_block_timer() {
     .build()
     .unwrap();
 
+    let nonce_guard = mount_default_nonce_query_mock(&sequencer).await;
     let status = executor.subscribe();
 
+    // executor.check_chain_ids().await.unwrap();
```
SuperFluffy, line 380: remove?

crates/astria-composer/src/executor/tests.rs
```diff
@@ -225,8 +277,10 @@ async fn full_bundle() {
...
```
SuperFluffy, line 283: remove?

SuperFluffy (commented):
Code is moving in the right direction, thank you for bearing with me and working on it. :)

I have left a bunch of comments on what what I'd like to see changed to test this.

The tests are currently not set up in the most ideal way, but short of making the exit reason publicly observable I don't see a way for you to test it other than you did.

I would only ask for you to type out the returned errors so that the tests won't stop silently breaking (and yet still passing).

crates/astria-composer/src/executor/mod.rs
eoroshiba: `chain_id` is of type `tendermint::chain::Id`. I may be mistaken, but I believe we'll have to de a conversion at some point for the comparison. Would you prefer to return the raw chain ID and do the conversion in `setup()`?

crates/astria-composer/src/executor/tests.rs
eoroshiba: Would this be an issue for later or would it be better to do that in this PR?

crates/astria-composer/src/executor/mod.rs
SuperFluffy: You don't need to convert at all. It provides this: https://docs.rs/tendermint/latest/tendermint/chain/id/struct.Id.html#method.as_str

crates/astria-composer/src/executor/tests.rs
SuperFluffy: No, that's why I prefixed it with "for a followup" :-)

crates/astria-composer/src/executor/builder.rs
```diff
@@ -12,7 +12,7 @@ use astria_core::{
 use astria_eyre::eyre::{
     self,
     eyre,
-    WrapErr as _,
+    Context,
```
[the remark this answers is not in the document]
eoroshiba: Resolved in 8523e04.

crates/astria-composer/src/executor/mod.rs
eoroshiba: Implemented in 8523e04.

crates/astria-composer/src/executor/mod.rs
eoroshiba: Apologies for the confusion, I saw your suggestion to compare using `as_str()` below and then deleted my previous comment. Should be resolved in 8523e04 :)

crates/astria-composer/src/executor/tests.rs
eoroshiba: Ah, I see. I was confused as to whether it meant followup in this PR or not. Thanks for the clarification! I also changed the comment in 8523e04 to reflect now mounting the genesis as opposed to a status response.

crates/astria-composer/src/executor/mod.rs
eoroshiba: Fixed in 8523e04.

crates/astria-composer/src/executor/mod.rs
eoroshiba: Resolved in 8523e04.

crates/astria-composer/src/executor/mod.rs
eoroshiba: Resolved in 8523e04.

crates/astria-composer/src/executor/tests.rs
eoroshiba: Resolved in 8523e04.

crates/astria-composer/src/executor/tests.rs
eoroshiba: Removed in 8523e04.

crates/astria-composer/src/executor/tests.rs
eoroshiba: Removed in 8523e04.

eoroshiba:
@SuperFluffy I should be thanking you for bearing with me! Recent commits should implement all the changes you requested.

crates/astria-composer/src/executor/mod.rs
```diff
@@ -417,6 +445,56 @@ impl Executor {
...
+            return Err(EnsureChainIdError::WrongChainId {
+                expected: self.sequencer_chain_id.clone(),
+                actual: remote_chain_id,
+            });
+        }
+        Ok(())
+    }
+
+    /// Fetch chain id from the sequencer client
+    async fn get_sequencer_chain_id(&self) -> eyre::Result<tendermint::chain::Id> {
```
SuperFluffy, line 471:
```suggest
    async fn get_sequencer_chain_id(&self) -> Result<tendermint::chain::Id, sequener_client::tendermint_rpc::Error> {
```

crates/astria-composer/src/executor/mod.rs
```diff
@@ -94,7 +99,19 @@ pub(crate) use builder::Builder;
...
 
 type StdError = dyn std::error::Error;
-
+#[derive(Debug, thiserror::Error)]
+pub(crate) enum EnsureChainIdError {
+    // <other error variants from getting genesis, etc>
+    #[error("failed to obtain sequencer chain ID")]
+    FailedToGetChainId,
+    #[error("failed to obtain sequencer genesis")]
+    FailedToGetGenesis,
```
SuperFluffy, line 108: Commented below, remove the `FailedtoGetGenesis` variant, keep the `FailedtoGetChainId`

crates/astria-composer/src/executor/mod.rs
```diff
@@ -94,7 +99,19 @@ pub(crate) use builder::Builder;
 const BUNDLE_DRAINING_DURATION: Duration = Duration::from_secs(16);
 
 type StdError = dyn std::error::Error;
-
+#[derive(Debug, thiserror::Error)]
+pub(crate) enum EnsureChainIdError {
+    // <other error variants from getting genesis, etc>
```
SuperFluffy, line 104: remove that comment :) That was just for not typing out everything in my review.

crates/astria-composer/src/executor/mod.rs
```diff
@@ -94,7 +99,19 @@ pub(crate) use builder::Builder;
...
+    #[error("failed to obtain sequencer chain ID")]
+    FailedToGetChainId,
```
SuperFluffy, line 106: Don't throw away the error:
```suggestion
    GetChainId(#[source] sequencer_client::tendermint_rpc::Error),
```
^ check if that's the right type. I think it is

Note on the variant: because this already is an error it's recommended to not repeat "failed" or "error" in the variants.

crates/astria-composer/src/executor/mod.rs
```diff
@@ -94,7 +99,19 @@ pub(crate) use builder::Builder;
 const BUNDLE_DRAINING_DURATION: Duration = Duration::from_secs(16);
 
 type StdError = dyn std::error::Error;
-
+#[derive(Debug, thiserror::Error)]
+pub(crate) enum EnsureChainIdError {
+    // <other error variants from getting genesis, etc>
+    #[error("failed to obtain sequencer chain ID")]
```
SuperFluffy, line 105:
```suggestion
    #[error("failed to obtain sequencer chain ID after multiple retries")]
```

crates/astria-composer/src/executor/mod.rs
```diff
@@ -199,6 +216,17 @@ impl Executor {
...
     pub(super) async fn run_until_stopped(mut self) -> eyre::Result<()> {
+        select!(
+            biased;
+            () = self.shutdown_token.cancelled() => {
+                info!("received shutdown signal while running initialization routines; exiting");
+                return Ok(());
+            }
+
+            res = self.pre_run_checks() => {
+                res?;
```
SuperFluffy, line 227:
```suggestion
                res.wrap_err("required pre-run checks failed)?;
```

crates/astria-composer/src/executor/tests.rs
```diff
@@ -481,3 +556,67 @@ async fn two_seq_actions_single_bundle() {
...
+    // ensure that the chain id check function fails
+    let chain_id_check_err = executor
+        .ensure_chain_id_is_correct()
+        .await
+        .expect_err("executor::ensure_chain_id_is_correct() should return an error");
+    assert_chain_id_err(
+        &chain_id_check_err,
+        &cfg.sequencer_chain_id,
+        Id::try_from("bad-chain-id".to_string()).unwrap(),
+    );
```
SuperFluffy, line 598: You can remove this block. You are already testing below, so this doesn't add anything of value.

crates/astria-composer/src/executor/tests.rs
```diff
@@ -112,15 +115,34 @@ async fn setup() -> (MockServer, MockGuard, Config, NamedTempFile) {
         grpc_addr: "127.0.0.1:0".parse().unwrap(),
         fee_asset: "nria".parse().unwrap(),
     };
-    (server, startup_guard, cfg, keyfile)
+    (server, cfg, keyfile)
+}
+
+/// Assert that given error is of correct type and contains the expected chain IDs.
+fn assert_chain_id_err(
```
SuperFluffy, line 122: To get a better error diagnostic:
```suggestion
#[track_caller]
fn assert_chain_id_err(
```

SuperFluffy (approved):
Looks good. Thank you very much!

I have left a few comments, primarily to not just throw away the error in `get_sequencer_chain_id`. Please address these before merging.

crates/astria-composer/src/executor/tests.rs
eoroshiba: Done in 33ae83d.

crates/astria-composer/src/executor/tests.rs
eoroshiba: Done in 33ae83d.

crates/astria-composer/src/executor/mod.rs
eoroshiba: Done in 33ae83d.

crates/astria-composer/src/executor/mod.rs
eoroshiba: Suggestion committed in 33ae83d.

crates/astria-composer/src/executor/mod.rs
eoroshiba: Done in 33ae83d.

crates/astria-composer/src/executor/mod.rs
eoroshiba: Done in 33ae83d.

crates/astria-composer/src/executor/tests.rs
```diff
@@ -112,15 +115,37 @@ async fn setup() -> (MockServer, MockGuard, Config, NamedTempFile) {
...
+    match err {
+        EnsureChainIdError::WrongChainId {
+            expected,
+            actual,
+        } => {
+            assert_eq!(*expected, configured_expected);
+            assert_eq!(*actual, *configured_actual);
+        }
+        other @ EnsureChainIdError::GetChainId(_) => {
+            panic!("expected `EnsureChainIdError::WrongChainId`, but got '{other}'")
```
SuperFluffy, line 137: Last change before merge:
```suggestion
            panic!("expected `EnsureChainIdError::WrongChainId`, but got '{other:?}'")
```
^ I'd have done it myself but don't have push rights for your fork.

PR #1178, opened 2024-06-12
chore(bridge-withdrawer): add missing errors and clean up names

itamarreif:
## Summary
add missing errors and clean up names

## Background
Various clean up tasks needed after rushing code through over the weekend.

## Changes
- added error messages
- renamed `Service` to `BridgeWithdrawer` to match the rest of the monorepo


crates/astria-bridge-withdrawer/src/bridge_withdrawer/submitter/signer.rs
```diff
@@ -26,8 +26,9 @@ impl SequencerKey {
     ///
     /// The file should contain a hex-encoded ed25519 secret key.
     pub(super) fn try_from_path<P: AsRef<Path>>(path: P) -> eyre::Result<Self> {
-        let hex = fs::read_to_string(path)?;
-        let bytes: [u8; 32] = hex::decode(hex.trim())?
+        let hex = fs::read_to_string(path).wrap_err("failed to read sequencer key from path")?;
+        let bytes: [u8; 32] = hex::decode(hex.trim())
+            .wrap_err("failed to decode hex")?
             .try_into()
             .map_err(|_| eyre!("invalid private key length; must be 32 bytes"))?;
```
Fraser999, line 33: It would be nice to include the `path.as_ref().display()` in all three error contexts here.

Fraser999 (approved):
One non-blocking nitpick, otherwise LGTM.

PR #1181, opened 2024-06-13
refactor(core): parse ics20 denoms as ibc or trace prefixed variants

SuperFluffy:
## Summary
Rewrites the ics20 parsing logic to distinguish between `ibc/` and `<path>` prefixed denoms.

## Background
While https://github.com/astriaorg/astria/pull/1162 was an improvement of what we had before, it was still problematic in that the parsing logic did not distinguish between denom of the form `[port/channel...]/base` and `ibc/hash`, so that a lot of the logic in sequencer was done ad-hoc. This patch actually distinguishes between both forms, allowing sequencer to enforce which type of denom it stores in its database, and which type it processes further.

## Changes
- Rewrite `astria_core::primitive::v1::Denom` as an enum with variants `TracePrefixed` and `IbcPrefixed`

## Testing
Adapt and expands unit tests, update sequencer tests.


crates/astria-bridge-withdrawer/src/withdrawer/ethereum/convert.rs
```diff
@@ -132,7 +135,8 @@ fn event_to_ics20_withdrawal(
     let denom = rollup_asset_denom.clone();
 
     let channel = denom
-        .channel()
+        .as_trace_prefixed()
+        .and_then(TracePrefixed::last_channel)
         .ok_or_eyre("denom must have a channel to be withdrawn via IBC")?;
```
SuperFluffy, line 140: Noticing this: the service does not actually enforce any form of the provided ics20 denom otherwise, but this error seems to suggest that `ibc/<hash>` denoms are not valid?

If that's the case the watcher builder can just parse the config input as `TracePrefixed` to test this at startup.

crates/astria-sequencer/src/ibc/ics20_transfer.rs
```diff
@@ -424,7 +413,7 @@ async fn execute_ics20_transfer<S: StateWriteExt>(
     // the IBC packet should have the address as a bech32 string
     let recipient = Address::try_from_bech32m(&recipient).context("invalid recipient address")?;
 
-    let is_prefixed = is_prefixed(source_port, source_channel, &unprefixed_denom);
+    let is_prefixed = denom_trace.has_exact_path(&format!("{source_port}/{source_channel}"));
```
SuperFluffy, line 416: As above, is the exact match maybe too strict? Would it be enough to check if it just starts with `{source_port}/{source_channel}`?

crates/astria-core/src/primitive/v1/asset/denom.rs
```diff
@@ -0,0 +1,554 @@
...
+        };
+        ibc
+    }
+
+    /// Unwraps the inner trace prefixed denom.
+    ///
+    /// # Panics
+    /// Panics if the self value equals [`Self::IbcPrefixed`].
+    #[must_use]
+    pub fn unwrap_trace(self) -> TracePrefixed {
```
noot, line 72:
```suggestion
    pub fn unwrap_trace_prefix(self) -> TracePrefixed {
```

for consistency w above

SuperFluffy: Fixed in 7b8ce7f (but used `unwrap_trace_prefixed`)

crates/astria-core/src/primitive/v1/asset/denom.rs
```diff
@@ -0,0 +1,554 @@
...
+    type Err = ParseTracePrefixedError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        if !s.is_ascii() {
+            return Err(Self::Err::not_ascii());
+        }
+        if s.as_bytes().iter().any(u8::is_ascii_whitespace) {
+            return Err(Self::Err::whitespace());
+        }
+        let (base_denom, segments) = match s.rsplit_once('/') {
```
noot, line 344: i think this is reversed, should be `let (segments, base_denom) = match s.rsplit_once('/')` since the base denom is always after the last slash

noot: can you add a unit test for this?

crates/astria-core/src/primitive/v1/asset/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+    }
+}
+
+impl Display for Id {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        use base64::{
+            display::Base64Display,
+            prelude::BASE64_STANDARD,
+        };
+        Base64Display::new(self.as_ref(), &BASE64_STANDARD).fmt(f)
```
noot, line 88: wondering if this should be hex to match how ibc/ prefixed denoms are displayed

crates/astria-core/src/primitive/v1/asset/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+
+impl Id {
+    #[must_use]
+    pub fn new(arr: [u8; 32]) -> Self {
+        Self(arr)
+    }
+
+    /// Constructs an ID by hashing `s` without checking if `s` is a valid denom.
+    #[must_use]
+    pub fn from_str_unchecked(s: &str) -> Self {
```
noot, line 47:
```suggestion
    pub fn from_denom_str_unchecked(s: &str) -> Self {
```

this is clearer imo, although the comment does explain this also

crates/astria-sequencer/src/ibc/ics20_transfer.rs
```diff
@@ -169,30 +172,25 @@ impl AppHandlerCheck for Ics20Transfer {
...
+        convert_denomination_if_ibc_prefixed(&mut state, denom)
             .await
-            .context("failed to get denom trace from asset id")?
-            .context("denom for given asset id not found in state")?;
-    }
+            .context("failed to convert denomination if ibc/ prefixed")?
+    };
 
-    let is_source = !is_prefixed(source_port, source_channel, &denom);
+    let is_source = !denom.has_exact_path(&format!("{source_port}/{source_channel}"));
```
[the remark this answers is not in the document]
noot: it's too strict, it should just check if the start of the trace (first port/channel) matches

crates/astria-core/src/primitive/v1/asset/denom.rs
SuperFluffy: Admittedly, the order is confusing and I will change it, but the important bit is the body of the match statement.

This is indirectly unit tested through the parse-logic below. But I will add a more explicit unit test that actually checks the fields being set correctly.

crates/astria-core/src/primitive/v1/asset/mod.rs
SuperFluffy: So the reason why I removed the `fn from_denom` is because it took an unstructured string but gave the impression that more was going on. I wanted to remove all mention of `denom` from the method because there is no extra checking for whether an actual denom is passed in.

I think I should move the type into the `denom` module so that one refers to it as `denom::Id`.

crates/astria-core/src/primitive/v1/asset/mod.rs
SuperFluffy: Yes, that makes sense. Does `Id` ever materialize in any of our protos? I think it doesn't right?

crates/astria-sequencer/src/ibc/ics20_transfer.rs
noot: yeah, should just check the start

noot (changes requested):
overall the changes look good! but have a few comments on the logic that should be addressed.

crates/astria-core/src/primitive/v1/asset/mod.rs
noot: yeah, i guess the ambiguity in the naming is that it's uncertain whether the string parameter is the encoded ID, or a denom trace. but the comment explains it so it's fine

crates/astria-core/src/primitive/v1/asset/mod.rs
noot: it does actually, for `fee_asset_id` in the transactions for example. do you want to leave as base64 then?

crates/astria-sequencer/src/ibc/ics20_transfer.rs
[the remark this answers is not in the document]
SuperFluffy: I have removed the `has_exact_path` method and replaced it with `TracePrefixed::starts_with_str`, which is much more forgiving.

 010b7e

crates/astria-sequencer/src/ibc/ics20_transfer.rs
SuperFluffy: As above, replaced `TracePrefixed::has_exact_path` replaced by `TracePrefixed::starts_with_str`  in 010b7e

crates/astria-core/src/primitive/v1/asset/denom.rs
SuperFluffy: Switched the var. Looked at the unit tests: the `pop_path` test covers all unit test needs for the parse logic I think.

 010b7e

crates/astria-core/src/primitive/v1/asset/mod.rs
SuperFluffy: Yeah, we should try to not have a different display between pbjson and our Display impls.

SuperFluffy: I did however add a `serde(transparent)` attribute on the `asset::Id` struct.  d6d136

crates/astria-sequencer/src/accounts/state_ext.rs
```diff
@@ -89,10 +89,7 @@ pub(crate) trait StateReadExt: StateRead {
                 // TODO: this is jank, just have 1 denom type.
                 // TODO: Getting the base denom out of a native asset should not need a parse.
                 balances.push(AssetBalance {
-                    denom: native_asset
-                        .base_denom()
-                        .parse()
-                        .context("failed to parse a denom's base as a denom; this is a problem")?,
+                    denom: native_asset.clone(),
```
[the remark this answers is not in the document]
SuperFluffy: done in  3d1d5d0 :)

crates/astria-core/src/primitive/v1/asset/mod.rs
SuperFluffy: Leaving it at `from_str_unchecked` then :)

crates/astria-core/src/primitive/v1/asset/denom.rs
```diff
@@ -168,25 +168,95 @@ impl TracePrefixed {
...
+            return true;
+        }
+        let mut had_trailing_slash = false;
+        let s = s
+            .strip_suffix('/')
+            .inspect(|_| had_trailing_slash = true)
+            .unwrap_or(s);
+        if s.is_empty() {
+            return false;
+        }
```
noot, line 219: the comment says `Empty string is always true` but this returns false

noot (approved):
looks good to me!

PR #1186, opened 2024-06-17
chore(core, proto)!: remove non-bech32m address bytes

SuperFluffy:
## Summary
Removes the non-bech32m address bytes field.

## Background
The change introduced in https://github.com/astriaorg/astria/pull/1124 is a breaking change that ripples through the rest of the Astria stack. While in principle we can keep the non-bech32m address around for backward compatibility, in practice it's cleaner to regenesis.

## Changes
- Remove `astria.primitive.v1.Address.inner`, reserving its number and name.

## Testing
Updates and removed all unit tests. All blackbox tests that use addresses in some form still work.

## Breaking Changelist
Since this is protobuf breaking it is also network breaking.

### DIFF
--- a/crates/astria-core/src/primitive/v1/mod.rs
+++ b/crates/astria-core/src/primitive/v1/mod.rs
@@ -7,7 +7,6 @@ use base64::{
     display::Base64Display,
     prelude::BASE64_STANDARD,
 };
-use bytes::Bytes;
 use sha2::{
     Digest as _,
     Sha256,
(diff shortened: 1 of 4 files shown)


crates/astria-core/src/primitive/v1/mod.rs
```diff
@@ -507,10 +486,9 @@ where
 
 #[cfg(test)]
 mod tests {
-    use bytes::Bytes;
+    use insta::assert_json_snapshot;
```
joroshiba, line 489: Doesn't look like any new usage added here?

SuperFluffy: Yeah, that was a strange import after rebasing. Removed.

noot (approved):

joroshiba (approved):

PR #1189, opened 2024-06-17
feat(sequencer): implement abci query for bridge account info

### LINKED ISSUE #1118: sequencer: query if account is bridge account
we should add a query for if an account is a bridge account and if so what its asset is.

noot:
## Summary
 implement abci query for bridge account info, so users can tell if an account is a bridge account or not, and its relevant info if it is.

## Background
needed for good UX

## Changes
-  implement abci query for bridge account info, so users can tell if an account is a bridge account or not, and its relevant info if it is
- update sequencer client respectively

## Testing
unit tests, sequencer client

## Related Issues
closes #1118


crates/astria-sequencer/src/bridge/query.rs
SuperFluffy, line 1: Can you please provide unit tests for these query handlers? Since we have written them as free standing functions unit testing should be relatively straight forward.

proto/protocolapis/astria/protocol/bridge/v1alpha1/types.proto
```diff
@@ -2,9 +2,19 @@ syntax = "proto3";
...
 // if it exists.
 message BridgeAccountLastTxHashResponse {
   uint64 height = 2;
   optional bytes tx_hash = 3;
 }
+
+message BridgeAccountInfoResponse {
+  uint64 height = 2;
+  astria.primitive.v1.RollupId rollup_id = 3;
+  optional bytes asset_id = 4;
```
SuperFluffy, line 17: What does it mean if this field is not set?

Also, I wonder if we should start using IBC denoms (i.e. strings) in their place - so this would be `ibc/<hash>`. I think it would make things a bit neater.

proto/protocolapis/astria/protocol/bridge/v1alpha1/types.proto
```diff
@@ -2,9 +2,19 @@ syntax = "proto3";
...
+import "astria/primitive/v1/types.proto";
+
 // A response containing the last tx hash given some bridge address,
 // if it exists.
 message BridgeAccountLastTxHashResponse {
   uint64 height = 2;
   optional bytes tx_hash = 3;
 }
+
+message BridgeAccountInfoResponse {
```
SuperFluffy, line 14: Small doc for what this is supposed to be the response to? I think this should mention ABCI because there is not (yet?) a matching grpc service.

crates/astria-core/src/protocol/bridge/v1alpha1/mod.rs
```diff
@@ -76,3 +83,167 @@ enum BridgeAccountLastTxHashResponseErrorKind {
...
+            height: self.height,
+            rollup_id: Some(info.rollup_id.into_raw()),
+            asset_id: Some(info.asset_id.get().to_vec()),
+            sudo_address: Some(info.sudo_address.into_raw()),
+            withdrawer_address: Some(info.withdrawer_address.into_raw()),
+        }
+    }
+}
+
+impl raw::BridgeAccountInfoResponse {
```
SuperFluffy, line 187: It was pretty annoying having to update all these `.into_native`/`from_native` methods. I think I introduced these utilities prematurely and we should get rid of them/not introduce them for now (also since thehy are only used in `astria-sequencer-client`).

proto/protocolapis/astria/protocol/bridge/v1alpha1/types.proto
noot: it means the account is not a bridge account - will add a comment to clarify

proto/protocolapis/astria/protocol/bridge/v1alpha1/types.proto
SuperFluffy, line 1: Should these live under `protocol`? They don't seem to be protocol-relevant?

proto/protocolapis/astria/protocol/bridge/v1alpha1/types.proto
```diff
@@ -2,9 +2,24 @@ syntax = "proto3";
...
 message BridgeAccountLastTxHashResponse {
   uint64 height = 2;
   optional bytes tx_hash = 3;
 }
+
+// A response to the `bridge/account_info` ABCI query
+// containing the account information given some bridge address,
+// if it exists.
+message BridgeAccountInfoResponse {
+  uint64 height = 2;
```
SuperFluffy, line 19: Why is this not indexed starting at 1? Is field number 1 reserved for something else?

crates/astria-sequencer/src/bridge/query.rs
```diff
@@ -15,6 +18,153 @@ use crate::{
...
+            log: info.into(),
+            ..response::Query::default()
+        };
+    }
+
+    let err = err.unwrap();
+    response::Query {
+        code: code.into(),
+        info: code.to_string(),
+        log: format!("{info}: {err:#}"),
```
SuperFluffy, line 39: This should not be pretty-printed.

crates/astria-sequencer/src/bridge/query.rs
```diff
@@ -15,6 +18,153 @@ use crate::{
     state_ext::StateReadExt as _,
 };
 
+fn error_query_response(
+    err: Option<anyhow::Error>,
+    code: AbciErrorCode,
+    info: &str,
+) -> response::Query {
+    if err.is_none() {
```
SuperFluffy, line 26: You can simplify the body of this function considerably by doing:

```rust
let log = match err {
    Some(err) => format!("{info}: {err:?}"),
    None => info.into(),
};
response::Query {
    code: code.into(),
    info: code.to_string(),
    log,
    ..response::Query::default()
}
```

crates/astria-sequencer/src/bridge/query.rs
```diff
@@ -15,6 +18,153 @@ use crate::{
     state_ext::StateReadExt as _,
 };
 
+fn error_query_response(
```
SuperFluffy, line 21: I like this helper.

crates/astria-sequencer/src/bridge/query.rs
```diff
@@ -15,6 +18,153 @@ use crate::{
...
+}
+
+pub(crate) async fn bridge_account_info_request(
+    storage: Storage,
+    request: request::Query,
+    params: Vec<(String, String)>,
+) -> response::Query {
+    use astria_core::protocol::bridge::v1alpha1::BridgeAccountInfoResponse;
+
+    let address = match preprocess_request(&params) {
```
SuperFluffy, line 127: I think you can apply the `?` operator here because this method returns a `response::Query`?

Also, other queries check the address prefix against the globally configured address prefixes, see the `astria_sequencer::address` submodule.

Should this also do that?

crates/astria-sequencer/src/bridge/query.rs
```diff
@@ -15,6 +18,153 @@ use crate::{
     state_ext::StateReadExt as _,
 };
 
+fn error_query_response(
+    err: Option<anyhow::Error>,
+    code: AbciErrorCode,
+    info: &str,
```
SuperFluffy, line 24: IMO rename this as `msg` or something because the `info` field in the query response (which is the human readable `code`) and this `info` argument tripped me up

SuperFluffy: Also to add to this: I think you can forego the `info` field entirely and lean on anyhow. If you encounter a `None`, you can either `None.context("something was not set").map_err(|e| error_to_query_response(e, <code>)`, or just `match thing_that_can_none() { None => ... }` as before.

proto/protocolapis/astria/protocol/bridge/v1alpha1/types.proto
noot: we can move them to a different protobuf package, can do in a follow-up

proto/protocolapis/astria/protocol/bridge/v1alpha1/types.proto
noot: this matches the original response types which i think you wrote :p
```
// A response containing the balance of an account.
message BalanceResponse {
  uint64 height = 2;
  repeated AssetBalance balances = 3;
}

// A response containing the current nonce for an account.
message NonceResponse {
  uint64 height = 2;
  uint32 nonce = 3;
}
```

crates/astria-sequencer/src/bridge/query.rs
noot: > I think you can apply the ? operator here because this method returns a response::Query?

doesn't work since this function doesn't return a result, just a `Query` :/

> Should this also do that?

yes, will add

proto/protocolapis/astria/protocol/bridge/v1alpha1/types.proto
SuperFluffy: yeah, starting later is indeed wrong. I wasn't aware that these are with us for a year already. :-S

In this PR, let them start at 1. The others we can fix in the followup breaking these request-response types out into a separate proto package.

crates/astria-sequencer/src/bridge/query.rs
SuperFluffy: > doesn't work since this function doesn't return a result, just a Query :/

Ah, yeah. :S

SuperFluffy (approved):

PR #1190, opened 2024-06-17
refactor(bridge-withdrawer)!: refactor startup to a separate subtask and remove balance check from startup

### LINKED ISSUE #1229: Remove balance check from startup in bridge-withdrawer
Remove the balance check from the startup logic. Submitting a transaction will simply fail the app-side mempool if there is insufficient balance, and the service will restart as a result. Doing this in the startup process only complicates operating the service since it requires funding the withdrawer prior to running it.

itamarreif:
## Summary
This refactors some of the startup logic into a separate startup subtask. It also removes the minimum expected balance check from the startup logic, as that is not necessary. 

## Background
The original startup logic had the `Submitter` running parts of the startup routine that weren't relevant to it. this moves that logic out to the `Startup` task.

The balance check is removed because if a sequencer transaction with insufficient balance is submitted it will just fail execution and the service will restart. Requiring the check in the startup logic results in clunky operation of the service, since launching it before funding it will cause it to crash.

## Changes
- Introduce `Startup` object and task
- `Watcher` and `Submitter` now wait for the `Startup` to update the global state object with startup info before running their startup logic, whereas before we just had the `Watcher` wait on the `Submitter` to send a message on the `oneshot` channel.
- Cleaned up some of the tryhard configs with helper funcs to be less confusing

## Breaking Changes
- Remove the balance check from startup and the associated configs. issue linked below

## Testing
- Updated some of the `Submitter`-specific tests to reflect changes to the startup logic.

closes #1229


crates/astria-bridge-withdrawer/src/bridge_withdrawer/ethereum/watcher.rs
```diff
@@ -195,15 +199,19 @@ impl Watcher {
...
-        } = self
-            .submitter_handle
-            .recv_startup_info()
-            .await
-            .wrap_err("failed to get sequencer startup info")?;
+            starting_rollup_height,
+        } = select! {
+            () = self.shutdown_token.cancelled() => {
+                info!("watcher received shutdown signal while waiting for startup");
+                return Err(eyre!("watcher shutting down"));
```
noot, line 208: put the log message in the error

crates/astria-bridge-withdrawer/src/bridge_withdrawer/submitter/tests.rs
```diff
@@ -93,8 +91,8 @@ use crate::{
 
 const SEQUENCER_CHAIN_ID: &str = "test_sequencer-1000";
 const DEFAULT_LAST_ROLLUP_HEIGHT: u64 = 1;
-const DEFAULT_LAST_SEQUENCER_HEIGHT: u64 = 0;
-const DEFAULT_SEQUENCER_NONCE: u32 = 0;
+// const DEFAULT_LAST_SEQUENCER_HEIGHT: u64 = 0;
+// const DEFAULT_SEQUENCER_NONCE: u32 = 0;
```
noot, line 95: remove

crates/astria-bridge-withdrawer/src/bridge_withdrawer/submitter/tests.rs
```diff
@@ -237,43 +222,38 @@ async fn register_default_min_expected_fee_asset_balance_guard(
...
-            "tx_hash".to_string(),
-            register_default_last_bridge_tx_hash_guard(cometbft_mock).await,
-        ),
-        (
-            "last_bridge_tx".to_string(),
-            register_default_last_bridge_tx_guard(cometbft_mock).await,
-        ),
-    ])
-}
+// async fn _register_default_last_bridge_tx_hash_guard(cometbft_mock: &MockServer) -> MockGuard {
```
noot, line 225: remove

crates/astria-bridge-withdrawer/src/bridge_withdrawer/submitter/tests.rs
```diff
@@ -364,47 +344,47 @@ fn make_tx_commit_deliver_tx_failure_response() -> tx_commit::Response {
...
-        index: 0,
-        tx_result: ExecTxResult {
-            code: abci::Code::Ok,
-            ..ExecTxResult::default()
-        },
-        tx: tx.into_raw().encode_to_vec(),
-        proof: None,
-    }
-}
+// fn _make_last_bridge_tx_hash_response() -> BridgeAccountLastTxHashResponse {
```
noot, line 347: remove

crates/astria-bridge-withdrawer/src/bridge_withdrawer/startup.rs
```diff
@@ -0,0 +1,614 @@
...
+            self.state.clone(),
+            self.address,
+        )
+        .await
+        .wrap_err("failed to get latest balance")?;
+        let fee_asset_balance = fee_asset_balances
+            .balances
+            .into_iter()
+            .find(|balance| balance.denom.id() == self.expected_fee_asset_id)
+            .ok_or_eyre("withdrawer's account does not have the minimum balance of the fee asset")?
```
noot, line 262:
```suggestion
            .ok_or_eyre("withdrawer's account balance of the fee asset is zero")?
```

crates/astria-bridge-withdrawer/src/bridge_withdrawer/startup.rs
```diff
@@ -0,0 +1,614 @@
...
+            .balance;
+        ensure!(
+            fee_asset_balance >= self.expected_min_fee_asset_balance,
+            "sequencer key does not have a sufficient balance of the fee asset"
```
noot, line 266:
```suggestion
            "withdrawer account does not have a sufficient balance of the fee asset"
```

for consistency w above

crates/astria-bridge-withdrawer/src/bridge_withdrawer/startup.rs
```diff
@@ -0,0 +1,614 @@
...
+    /// 4. The memo of the last transaction by the bridge account could not be parsed
+    async fn get_starting_rollup_height(&mut self) -> eyre::Result<u64> {
+        let signed_transaction = self
+            .get_last_transaction()
+            .await
+            .wrap_err("failed to get the bridge account's last sequencer transaction")?;
+        let starting_rollup_height = if let Some(signed_transaction) = signed_transaction {
+            rollup_height_from_signed_transaction(&signed_transaction).wrap_err(
+                "failed to extract rollup height from last transaction by the bridge account",
+            )?
```
noot, line 365:
```suggestion
            rollup_height_from_signed_transaction(&signed_transaction).wrap_err(
                "failed to extract rollup height from last transaction by the bridge account",
            )? + 1
```

this returns the last submitted rollup height, but we want to return the next rollup height to process, which is last submitted + 1.

noot (commented):
overall looks good, only some minor comments!

crates/astria-bridge-withdrawer/src/bridge_withdrawer/startup.rs
itamarreif: good catch, my bad

crates/astria-bridge-withdrawer/src/bridge_withdrawer/submitter/builder.rs
```diff
@@ -1,69 +1,50 @@
...
-            .await
-            .wrap_err("failed to get startup info from submitter. channel was dropped.")
-    }
-
     pub(crate) async fn send_batch(&self, batch: Batch) -> eyre::Result<()> {
         self.batches_tx
             .send(batch)
             .await
-            .wrap_err("failed to send batch")
+            .wrap_err("failed submitter_handleto send batch")
```
noot, line 37:
```suggestion
            .wrap_err("failed to send batch")
```

crates/astria-bridge-withdrawer/src/bridge_withdrawer/ethereum/watcher.rs
```diff
@@ -52,19 +52,21 @@ use crate::bridge_withdrawer::{
             WithdrawalEvent,
         },
     },
+    startup,
+    startup::WatcherInfo as StartupInfo,
```
SuperFluffy, line 56: Avoid aliases if they don't help clarity.

crates/astria-bridge-withdrawer/src/bridge_withdrawer/startup.rs
```diff
@@ -0,0 +1,613 @@
...
+        asset::v1alpha1::AllowedFeeAssetIdsResponse,
+        bridge::v1alpha1::BridgeAccountLastTxHashResponse,
+        transaction::v1alpha1::Action,
+    },
+};
+use astria_eyre::eyre::{
+    self,
+    ensure,
+    eyre,
+    Context as _,
```
SuperFluffy, line 19: Use `WrapErr` not `Context`

crates/astria-bridge-withdrawer/src/bridge_withdrawer/startup.rs
```diff
@@ -0,0 +1,613 @@
...
+    ContextCompat,
```
SuperFluffy, line 20: `eyre` actively discourages the use of this and only provides this as a crutch to move from anyhow, see here: https://docs.rs/eyre/latest/eyre/trait.ContextCompat.html#why-doesnt-eyre-impl-wraperr-for-option

Please change this.

crates/astria-bridge-withdrawer/src/bridge_withdrawer/startup.rs
```diff
@@ -0,0 +1,613 @@
...
+    OptionExt as _,
```
SuperFluffy, line 21: This one on the other hand is cool. I didn't know about it.

crates/astria-bridge-withdrawer/src/bridge_withdrawer/startup.rs
```diff
@@ -0,0 +1,613 @@
...
+};
+use prost::Message as _;
+use sequencer_client::{
```
SuperFluffy, line 24: Avoid mixing the sequencer_client reexports and the types imported from `astria-core.` Reexporting all these types from `astria-sequencer-client` was probably an error.

crates/astria-bridge-withdrawer/src/bridge_withdrawer/startup.rs
```diff
@@ -0,0 +1,613 @@
...
+    expected_fee_asset_id: asset::Id,
+    expected_min_fee_asset_balance: u128,
+}
+
+impl Startup {
+    pub(super) async fn run(mut self) -> eyre::Result<()> {
+        let shutdown_token = self.shutdown_token.clone();
+
+        let startup_task = tokio::spawn(async {
+            self.confirm_sequencer_config()
```
SuperFluffy, line 166: Should just spawn these 2 async tasks concurrently and then `try_join` them

crates/astria-bridge-withdrawer/src/bridge_withdrawer/startup.rs
```diff
@@ -0,0 +1,613 @@
...
+            Some(memo.block_number.as_u64())
+        }
+        Action::Ics20Withdrawal(action) => {
+            let memo: Ics20WithdrawalFromRollupMemo = serde_json::from_str(&action.memo)
+                .wrap_err("failed to parse memo from last transaction by the bridge account")?;
+            Some(memo.block_number)
+        }
+        _ => None,
+    }
+    .expect("action is already checked to be either BridgeUnlock or Ics20Withdrawal");
```
SuperFluffy, line 409: This is kind of far away from where the actual check happens. Easy to accidentally mess up when refactoring. Please just `bail!` here instead, I don't think one needs to panic.

SuperFluffy (changes requested):
I find the design of `WatcherInfo` and `SubmitterInfo` unnecessarily complicated with their use of `oneshot`. I understand why you went for them (the tasks should just grab that info once), but it seems unnecessarily heavy handed to just crash the app if they did it twice?

Could it be replaced by a [watch channel](https://docs.rs/tokio/latest/tokio/sync/watch/index.html) and all the individual values just put into a flat "state" file, i.e. something like:

```rust
struct State {
    starting_rollup_height: u64,
    fee_asset: asset::Denom,
    chain_id: String,
}
```

You could then design a handle to this state like so:

```rust
#[derive(Clone)]
pub(crate) struct StateHandle {
    rx: watch::Receiver<Option<State>>,
}

impl StateHandle {
    pub(crate) async fn get_state(&self) -> eyre::Result<State> {
        let state = self.rx.wait_for(Option::is_some).await.wrap_err("the startup task failed")?;
        Ok((*state).expect("the previous line guarantes that the state is intialized").clone())
} 
```

crates/astria-bridge-withdrawer/src/bridge_withdrawer/startup.rs
SuperFluffy, line 1: A note on al the retry-configs in this crate: they should probably be replaced to only retry on network errors. If some data could not be deserialized, then retrying  is hopeless.

crates/astria-bridge-withdrawer/src/bridge_withdrawer/startup.rs
itamarreif: this causes a cascade of compilation errors where functions expect e.g. `sequencer_client::Address` instead of `core::generate::Address`

crates/astria-bridge-withdrawer/src/bridge_withdrawer/startup.rs
itamarreif: Cleaned these up a bit by moving the retry config to a helper func (also realized that every other config had different delays configured...).

The only place where a `tendermint_rpc::Error` was mixed up for a deserialization error was for `get_latest_balance`

crates/astria-bridge-withdrawer/src/bridge_withdrawer/startup.rs
SuperFluffy: You should almost never use the types defined in `generate`. Use the ones from, for example, `astria_core::primitiv`, not `astria_core::generated::primitive`

crates/astria-bridge-withdrawer/src/bridge_withdrawer/ethereum/watcher.rs
```diff
@@ -641,23 +648,23 @@ mod tests {
 
         let value = 1_000_000_000.into();
         let recipient = crate::astria_address([1u8; 20]);
-
         let bridge_address = crate::astria_address([1u8; 20]);
-        let denom = default_native_asset();
+        let denom = "transfer/channel-0/utia".parse::<Denom>().unwrap();
```
noot, line 652:
```suggestion
        let denom = "nria".parse::<Denom>().unwrap();
```

arbitrary anyways since it's a test, but this matches what was there before

crates/astria-bridge-withdrawer/src/bridge_withdrawer/ethereum/watcher.rs
```diff
@@ -873,6 +853,34 @@ mod tests {
...
+            submitter_handle: submitter::Handle::new(batch_tx),
+            sequencer_address_prefix: crate::ASTRIA_ADDRESS_PREFIX.into(),
+        }
+        .build()
+        .unwrap();
+
+        tokio::task::spawn(watcher.run());
+
+        // make another tx to trigger anvil to make another block
+        send_ics20_withdraw_transaction(&contract, value, recipient).await;
```
noot, line 882: remove

crates/astria-bridge-withdrawer/src/bridge_withdrawer/ethereum/watcher.rs
```diff
@@ -998,6 +980,34 @@ mod tests {
...
+            submitter_handle: submitter::Handle::new(batch_tx),
+            sequencer_address_prefix: crate::ASTRIA_ADDRESS_PREFIX.into(),
+        }
+        .build()
+        .unwrap();
+
+        tokio::task::spawn(watcher.run());
+
+        // make another tx to trigger anvil to make another block
+        send_sequencer_withdraw_transaction_erc20(&contract, value, recipient).await;
```
noot, line 1009: remove

crates/astria-bridge-withdrawer/src/bridge_withdrawer/ethereum/watcher.rs
```diff
@@ -1106,6 +1090,34 @@ mod tests {
...
+            submitter_handle: submitter::Handle::new(batch_tx),
+            sequencer_address_prefix: crate::ASTRIA_ADDRESS_PREFIX.into(),
+        }
+        .build()
+        .unwrap();
+
+        tokio::task::spawn(watcher.run());
+
+        // make another tx to trigger anvil to make another block
+        send_ics20_withdraw_transaction_astria_bridgeable_erc20(&contract, value, recipient).await;
```
noot, line 1119: remove

crates/astria-bridge-withdrawer/src/bridge_withdrawer/state.rs
```diff
@@ -39,17 +41,24 @@ macro_rules! forward_setter {
...
 #[derive(Clone, Debug, Default, PartialEq, Eq, serde::Serialize)]
+// This lint is allowed because there are 4 boolean fields that are necessary to track here.
+#[allow(clippy::struct_excessive_bools)]
 pub(crate) struct StateSnapshot {
+    startup_info: Option<startup::Info>,
+
     watcher_ready: bool,
     submitter_ready: bool,
 
+    cometbft_connected: bool,
```
noot, line 61: what is this for? right now, the only connection to the sequencer is via cometbft?

crates/astria-bridge-withdrawer/src/bridge_withdrawer/submitter/tests.rs
```diff
@@ -139,22 +132,20 @@ impl TestSubmitter {
         let cometbft_mock = MockServer::start().await;
         let sequencer_cometbft_endpoint = format!("http://{}", cometbft_mock.address());
 
-        // withdrawer state
+        // startup info
```
noot, line 135: remove

crates/astria-bridge-withdrawer/src/bridge_withdrawer/submitter/tests.rs
```diff
@@ -193,38 +179,26 @@ impl TestSubmitter {
...
         let mut submitter = Self::setup().await;
-        submitter.startup_and_spawn().await;
+        submitter.startup().await;
         submitter
     }
 }
 
-async fn register_default_chain_id_guard(cometbft_mock: &MockServer) -> MockGuard {
-    register_genesis_chain_id_response(SEQUENCER_CHAIN_ID, cometbft_mock).await
+async fn _register_default_chain_id_guard(cometbft_mock: &MockServer) -> MockGuard {
```
noot, line 189: are these all unused now? if so, remove them

crates/astria-bridge-withdrawer/src/bridge_withdrawer/state.rs
itamarreif: this is my bad, had some git troubles and this slipped past me

crates/astria-bridge-withdrawer/src/bridge_withdrawer/startup.rs
```diff
@@ -0,0 +1,564 @@
...
+            rx,
+        }
+    }
+
+    pub(super) async fn get_info(&mut self) -> eyre::Result<Info> {
+        let state = self
+            .rx
+            .wait_for(|state| state.get_startup_info().is_some())
+            .await
+            .wrap_err("")?;
```
SuperFluffy, line 113: Some context?

crates/astria-bridge-withdrawer/src/bridge_withdrawer/startup.rs
```diff
@@ -0,0 +1,564 @@
...
+                    error = error as &dyn std::error::Error,
+                    retry_message,
+                );
+                futures::future::ready(())
+            },
+        )
+}
+
+fn make_sequencer_retry_config(
+    state: Arc<State>,
```
SuperFluffy, line 532: Honestly think this is unnecessary when you run the startup logic. Either this returns successfully or it doesn't. If it returns successfully you know that you are connected.

crates/astria-bridge-withdrawer/src/bridge_withdrawer/startup.rs
```diff
@@ -0,0 +1,564 @@
...
+    let res = res?.wrap_err("failed to deserialize the latest balance response from CometBFT");
+
+    // set cometbft as connected if received a response
+    state.set_sequencer_connected(res.is_ok());
+
+    res
+}
+
+fn make_cometbft_retry_config(
+    state: Arc<State>,
```
SuperFluffy, line 503: As for the other one - this is essentially a oneshot method at startup - if it returns you know that you are "connected". Make this simpler be removing it.

crates/astria-bridge-withdrawer/src/bridge_withdrawer/startup.rs
```diff
@@ -0,0 +1,564 @@
...
+                }
+            }
+        }
+    })
+    .with_config(make_cometbft_retry_config(
+        state.clone(),
+        "attempt to get latest balance from CometBFT; retrying after backoff".to_string(),
+    ))
+    .await
+    .wrap_err("failed to get latest balance from Sequencer after a lot of attempts");
```
SuperFluffy, line 492: Apply the `?` at the end of this line

crates/astria-bridge-withdrawer/src/bridge_withdrawer/state.rs
SuperFluffy: Should probably disambiguate between `sequencer_cometbft_connected` and `sequencer_grpc_connected` or something like that. As it stands, this is a bit confusing.

noot: @itamarreif can you remove this then?

crates/astria-bridge-withdrawer/src/bridge_withdrawer/state.rs
```diff
@@ -39,17 +41,23 @@ macro_rules! forward_setter {
...
+    [set_startup_info <- startup::Info],
     [set_sequencer_connected <- bool],
     [set_last_rollup_height_submitted <- u64],
     [set_last_sequencer_height <- u64],
     [set_last_sequencer_tx_hash <- tendermint::Hash],
 );
 
 #[derive(Clone, Debug, Default, PartialEq, Eq, serde::Serialize)]
+// This lint is allowed because there are 4 boolean fields that are necessary to track here.
+#[allow(clippy::struct_excessive_bools)]
```
noot, line 53: remove this also after removing `cometbft_connected`

noot (approved):
approving but please address comments before merge!

crates/astria-bridge-withdrawer/local.env.example
```diff
@@ -37,9 +37,6 @@ ASTRIA_BRIDGE_WITHDRAWER_SEQUENCER_ADDRESS_PREFIX=astria
 # The fee asset denomination to use for the bridge account's transactions.
 ASTRIA_BRIDGE_WITHDRAWER_FEE_ASSET_DENOMINATION="nria"
 
-# The minimum expected balance of the fee asset in the bridge account.
-ASTRIA_BRIDGE_WITHDRAWER_MIN_EXPECTED_FEE_ASSET_BALANCE=1000000
```
SuperFluffy, line 41: Removing this is a breaking change but it's not marked in the PR title nor explained in the PR message.

crates/astria-bridge-withdrawer/src/config.rs
```diff
@@ -18,8 +18,6 @@ pub struct Config {
     pub sequencer_key_path: String,
     // The fee asset denomination to use for the bridge account's transactions.
     pub fee_asset_denomination: asset::Denom,
-    // The minimum expected balance of the fee asset in the bridge account.
-    pub min_expected_fee_asset_balance: u64,
```
SuperFluffy, line 22: Mentioned in the example config, reproduced here:
Removing this is a breaking change but it's not marked in the PR title nor explained in the PR message.

crates/astria-bridge-withdrawer/src/bridge_withdrawer/mod.rs
```diff
@@ -264,26 +288,48 @@ struct Shutdown {
...
+                Err(_) => {
+                    error!(
+                        timeout_secs = limit.as_secs(),
+                        "startup did not shut down within timeout; killing it"
+                    );
+                    startup_task.abort();
+                }
+            }
+        } else {
+            info!("startup task was already dead");
```
SuperFluffy, line 330: IMO this provides no useful information. Remove

crates/astria-bridge-withdrawer/src/bridge_withdrawer/startup.rs
```diff
@@ -0,0 +1,490 @@
...
+    sequencer_cometbft_client: sequencer_client::HttpClient,
+    sequencer_bridge_address: Address,
+    expected_fee_asset: asset::Denom,
+}
+
+impl Startup {
+    pub(super) async fn run(mut self) -> eyre::Result<()> {
+        let shutdown_token = self.shutdown_token.clone();
+
+        let startup_task = tokio::spawn({
```
SuperFluffy, line 130: Why spawn here? Just create the async block without an extra spawn.

crates/astria-bridge-withdrawer/src/bridge_withdrawer/startup.rs
```diff
@@ -0,0 +1,490 @@
...
+    /// - `self.sequencer_chain_id` matches the value returned from the sequencer node's genesis
+    /// - `self.fee_asset_id` is a valid fee asset on the sequencer node
+    /// - `self.sequencer_key.address` has a sufficient balance of `self.fee_asset_id`
+    ///
+    /// # Errors
+    ///
+    /// - `self.chain_id` does not match the value returned from the sequencer node
+    /// - `self.fee_asset_id` is not a valid fee asset on the sequencer node
+    /// - `self.sequencer_key.address` does not have a sufficient balance of `self.fee_asset_id`.
+    async fn confirm_sequencer_config(&mut self) -> eyre::Result<()> {
```
SuperFluffy, line 183: Does this have to be mutable?

crates/astria-bridge-withdrawer/src/bridge_withdrawer/startup.rs
```diff
@@ -0,0 +1,490 @@
...
+                    error = error as &dyn std::error::Error,
+                    retry_message,
+                );
+                futures::future::ready(())
+            },
+        )
+}
+
+fn make_sequencer_retry_config(
+    retry_message: String,
```
SuperFluffy, line 462: This can just be a `&'static str`

crates/astria-bridge-withdrawer/src/bridge_withdrawer/startup.rs
```diff
@@ -0,0 +1,490 @@
...
+}
+
+#[instrument(skip_all)]
+async fn get_allowed_fee_asset_ids(
+    client: sequencer_client::HttpClient,
+    state: Arc<State>,
+) -> eyre::Result<AllowedFeeAssetsResponse> {
+    let res = tryhard::retry_fn(|| client.get_allowed_fee_assets())
+        .with_config(make_sequencer_retry_config(
+            "attempt to get allowed fee assets from Sequencer; retrying after backoff".to_string(),
```
SuperFluffy, line 426: IMO unnecessary `.to_string()`

crates/astria-bridge-withdrawer/src/bridge_withdrawer/startup.rs
```diff
@@ -0,0 +1,490 @@
...
+}
+
+#[instrument(skip_all)]
+async fn get_sequencer_chain_id(
+    client: sequencer_client::HttpClient,
+    state: Arc<State>,
+) -> eyre::Result<tendermint::chain::Id> {
+    let genesis: tendermint::Genesis = tryhard::retry_fn(|| client.genesis())
+        .with_config(make_cometbft_retry_config(
+            "attempt to get genesis from CometBFT; retrying after backoff".to_string(),
```
SuperFluffy, line 409: Mentioned elsewhere: remove the `to_string`, it's IMO unnecessary

crates/astria-bridge-withdrawer/src/bridge_withdrawer/startup.rs
```diff
@@ -0,0 +1,490 @@
...
+
+#[instrument(skip_all)]
+async fn get_tx(
+    client: sequencer_client::HttpClient,
+    state: Arc<State>,
+    tx_hash: tendermint::Hash,
+) -> eyre::Result<tx::Response> {
+    let res = tryhard::retry_fn(|| client.tx(tx_hash, false))
+        .with_config(make_cometbft_retry_config(
+            "attempt to get transaction from CometBFT; retrying after backoff".to_string(),
```
SuperFluffy, line 392: Remove `to_string`

crates/astria-bridge-withdrawer/src/bridge_withdrawer/startup.rs
```diff
@@ -0,0 +1,490 @@
...
+             attempts",
+        );
+
+    state.set_sequencer_connected(res.is_ok());
+
+    res
+}
+
+#[instrument(skip_all)]
+async fn get_tx(
```
SuperFluffy, line 385: Can you give this a more verbose name? `get_sequencer_transaction_at_hash`? I mentioned it before, but this crate uses `tx` to mean:

1. a sending channel
2. an ethereum transaction
3. a sequencer (cometbft) transaction

crates/astria-bridge-withdrawer/src/bridge_withdrawer/startup.rs
```diff
@@ -0,0 +1,490 @@
...
+async fn get_bridge_account_last_transaction_hash(
+    client: sequencer_client::HttpClient,
+    state: Arc<State>,
+    address: Address,
+) -> eyre::Result<BridgeAccountLastTxHashResponse> {
+    let res = tryhard::retry_fn(|| client.get_bridge_account_last_transaction_hash(address))
+        .with_config(make_sequencer_retry_config(
+            "attempt to fetch last bridge account's transaction hash from Sequencer; retrying \
+             after backoff"
+                .to_string(),
```
SuperFluffy, line 371: unnecessary allocation

crates/astria-bridge-withdrawer/src/bridge_withdrawer/startup.rs
```diff
@@ -0,0 +1,490 @@
...
+        let startup_task = tokio::spawn({
+            let state = self.state.clone();
+            async move {
+                self.confirm_sequencer_config()
+                    .await
+                    .wrap_err("failed to confirm sequencer config")?;
+                let starting_rollup_height = self
+                    .get_starting_rollup_height()
+                    .await
+                    .wrap_err("failed to get next rollup block height")?;
```
SuperFluffy, line 139: I think these could run concurrently?

crates/astria-bridge-withdrawer/src/bridge_withdrawer/ethereum/watcher.rs
```diff
@@ -192,15 +195,20 @@ impl Watcher {
...
-        } = self
-            .submitter_handle
-            .recv_startup_info()
-            .await
-            .wrap_err("failed to get sequencer startup info")?;
+            starting_rollup_height,
+            ..
+        } = select! {
+            () = self.shutdown_token.cancelled() => {
+                info!("watcher received shutdown signal while waiting for startup");
```
SuperFluffy, line 204: So this will lead to confusing errors: either this is an error or not. IMO there is no need to write an info event if it's covered by the error anyway.

crates/astria-bridge-withdrawer/src/bridge_withdrawer/startup.rs
```diff
@@ -0,0 +1,490 @@
...
+                &*last_transaction.tx,
+            )
+            .wrap_err("failed to convert transaction data from CometBFT to proto")?;
+
+        let tx = SignedTransaction::try_from_raw(proto_tx)
+            .wrap_err("failed to convert transaction data from proto to SignedTransaction")?;
+
+        info!(
+            last_bridge_account_tx.hash = %telemetry::display::hex(&tx_hash),
+            last_bridge_account_tx.height = i64::from(last_transaction.height),
```
SuperFluffy, line 268: The conversion is not necessary, just print it with `%last_trasanction.height`

crates/astria-bridge-withdrawer/src/bridge_withdrawer/startup.rs
```diff
@@ -0,0 +1,490 @@
...
+            last_transaction.tx_result.code == tendermint::abci::Code::Ok,
+            "last transaction by the bridge account failed to execute. this should not happen in \
+             the sequencer logic."
+        );
+
+        let proto_tx =
+            astria_core::generated::protocol::transaction::v1alpha1::SignedTransaction::decode(
+                &*last_transaction.tx,
+            )
+            .wrap_err("failed to convert transaction data from CometBFT to proto")?;
```
SuperFluffy, line 261:
```suggestion
            .wrap_err_format(|| format!(
                "failed to decode data in Sequencer CometBFT transaction as `{}`",
                astria_core::generated::protocol::transaction::v1alpha1::SignedTransaction::full_name(),
            )?;
```

crates/astria-bridge-withdrawer/src/bridge_withdrawer/startup.rs
```diff
@@ -0,0 +1,490 @@
...
+
+        let tx = SignedTransaction::try_from_raw(proto_tx)
+            .wrap_err("failed to convert transaction data from proto to SignedTransaction")?;
```
SuperFluffy, line 264:
```suggestion
            .wrap_err_with(|| format!("failed to verify {}", astria_core::generated::protocol::transaction::v1alpha1::SignedTransaction::full_name()))?;
```

crates/astria-bridge-withdrawer/src/bridge_withdrawer/startup.rs
```diff
@@ -0,0 +1,490 @@
...
+                Ok(())
+            }
+        });
+
+        tokio::select!(
+            () = shutdown_token.cancelled() => {
+                Err(eyre!("startup was cancelled"))
+            }
+            res = startup_task => {
+                match res {
```
SuperFluffy, line 159: IMO this unwrapping is not necessary if you just run the async function without the extra `tokio::spawn`

crates/astria-bridge-withdrawer/src/bridge_withdrawer/startup.rs
```diff
@@ -0,0 +1,490 @@
...
+                    Ok(Ok(())) => Ok(()),
+                    Ok(Err(err)) => {
+                        error!(%err, "startup failed");
```
SuperFluffy, line 162: This log is IMO not necesary. This will be printed one level above again I think.

SuperFluffy (commented):
Code-wise this is mergeable.

Where I am blocking is on the removal of the min expected fee asset balance. That's actual functionality being removed without any explanation. I can approve this if there is a good reason for it - or if you just remove these lines from the PR

crates/astria-bridge-withdrawer/local.env.example
itamarreif: it is mentioned in the PR and there was an issue for it that I forgot to link.

The balance check was unnecessary and removing it resolved some of your previous comments regarding exponential backoffs and serialization errors.

crates/astria-bridge-withdrawer/src/bridge_withdrawer/startup.rs
itamarreif: `get_tx` is the name of this in the tendermint client, but sure.

SuperFluffy (approved):

quasystaty1 (approved):

PR #1191, opened 2024-06-18
chore(sequencer-relayer): add timeout to gRPCs to Celestia app

### LINKED ISSUE #1187: Add timeouts to sequencer-relayer RPC calls

Fraser999:
## Summary
Added a five second timeout to all gRPCs targeting the Celestia app.

## Background
Currently if the Celestia app doesn't respond to any gRPC in a timely manner, the relayer's submission flow is blocked by this.  We want to avoid this by not waiting indefinitely for any gRPC.

## Changes
- Added timeout to the gRPC channel.  This is applied to all requests sent via that channel.  We can move to more fine-grained timeouts (different per request) in the future if required.

## Testing
- Added a unit test.  I opted for this approach over black box testing for a couple of reasons: we can leverage tokio's `time::advance` to avoid the test running for the actual duration of the timeout, and we have a better chance to inspect the actual error returned by the RPC to ensure it's a timeout variant.

## Related Issues
Closes #1187.

### DIFF
--- a/crates/astria-sequencer-relayer/src/relayer/celestia_client/tests.rs
+++ b/crates/astria-sequencer-relayer/src/relayer/celestia_client/tests.rs
@@ -1,19 +1,9 @@
-use astria_core::generated::cosmos::{
-    base::abci::v1beta1::TxResponse,
-    tx::v1beta1::{
-        BroadcastTxResponse,
-        GetTxResponse,
-    },
-};
+use astria_core::generated::cosmos::base::abci::v1beta1::TxResponse;
 use celestia_types::{
     blob::Commitment,
     nmt::Namespace,
 };
 use prost::bytes::Bytes;
-use tonic::{
-    Response,
-    Status,
-};
 
 use super::*;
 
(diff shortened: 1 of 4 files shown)


crates/astria-sequencer-relayer/src/relayer/celestia_client/tests.rs
```diff
@@ -1,12 +1,103 @@
...
+    },
+    Request,
+    Response,
+    Status,
+};
 
 use super::*;
 
+#[tokio::test(start_paused = true)]
+async fn get_tx_should_timeout() {
```
SuperFluffy, line 36: At its core, this tests the timeout inside `Client::get_tx` and thus if the underlying timeout logic of tonic's channel works. This limits the usefulness of this test (because we should always be able to rely on the correct behavior of our building blocks).

SuperFluffy (approved):
I agree with the patch, but I am not a fan of the test (described as a comment above).

What would make for a useful test is externally observable behavior, such as a request timing out and a second request being issued. This would test relayer's expected behavior in light of failure.


I don't see a reason to block this PR on these grounds (because adding this blackbox test requires more work on the mock grpc server), but I would be fine if the unit test was removed.

crates/astria-sequencer-relayer/src/relayer/celestia_client/tests.rs
Fraser999: As discussed, I agree on removing this test, with a view to add black box tests in the near future for sad path cases like gRPC errors, including timeouts.

PR #1196, opened 2024-06-19
feat(sequencer): implement transaction fee query

### LINKED ISSUE #1071: sequencer: `estimate_fee` api endpoint
implement a sequencer API endpoint which returns the fee for a given transaction (based on the current head state)

noot:
## Summary
implement transaction fee ABCI query method. calculates the fee for the transaction based on the current state.

## Background
useful for UX and fee estimation.

## Changes
- implement transaction fee ABCI query method which calculates the fee for the transaction based on the current state
- update sequencer-client respectively

## Testing
unit tests

## Related Issues

closes #1071


crates/astria-sequencer/src/transaction/query.rs
```diff
@@ -0,0 +1,106 @@
...
+        }
+    };
+
+    let fees = match get_fees_for_transaction(&tx, &snapshot).await {
+        Ok(fees) => fees,
+        Err(err) => {
+            return response::Query {
+                code: AbciErrorCode::INTERNAL_ERROR.into(),
+                info: AbciErrorCode::INTERNAL_ERROR.to_string(),
+                log: format!("failed calculating fees for provided transaction: {err:?}"),
```
Fraser999, line 52:
```suggestion
                log: format!("failed calculating fees for provided transaction: {err:#}"),
```

crates/astria-sequencer/src/transaction/query.rs
```diff
@@ -0,0 +1,106 @@
...
+        height,
+        fees,
+    };
+
+    let payload = resp.into_raw().encode_to_vec().into();
+
+    let height = tendermint::block::Height::try_from(height).expect("height must fit into an i64");
+    response::Query {
+        code: 0.into(),
+        key: request.path.clone().into_bytes().into(),
```
Fraser999, line 70:
```suggestion
        key: request.path.into_bytes().into(),
```

crates/astria-sequencer/src/transaction/query.rs
```diff
@@ -0,0 +1,106 @@
...
+    }
+}
+
+fn preprocess_request(request: &request::Query) -> Result<UnsignedTransaction, response::Query> {
+    let tx = match RawUnsignedTransaction::decode(&*request.data) {
+        Ok(tx) => tx,
+        Err(err) => {
+            return Err(response::Query {
+                code: AbciErrorCode::INTERNAL_ERROR.into(),
+                info: AbciErrorCode::INTERNAL_ERROR.to_string(),
```
Fraser999, line 83: This doesn't seem like an internal server error - it points to the request being malformed.  I'm not sure if any of the existing `AbciErrorCode` variants really fit.  We need something like a `BAD_REQUEST` (equivalent to HTTP 400 code) variant I think.

crates/astria-sequencer/src/transaction/query.rs
```diff
@@ -0,0 +1,106 @@
...
+            });
+        }
+    };
+
+    let tx = match UnsignedTransaction::try_from_raw(tx) {
+        Ok(tx) => tx,
+        Err(err) => {
+            return Err(response::Query {
+                code: AbciErrorCode::INTERNAL_ERROR.into(),
+                info: AbciErrorCode::INTERNAL_ERROR.to_string(),
```
Fraser999, line 95: Same for this case - more like a `BAD_REQUEST` error?

Fraser999 (approved):

crates/astria-core/src/protocol/abci.rs
```diff
@@ -37,6 +38,7 @@ impl AbciErrorCode {
             8 => "the requested value was not found".into(),
             9 => "the transaction expired in the app's mempool".into(),
             10 => "the transaction failed to execute in prepare_proposal()".into(),
+            11 => "the request was malformed".into(),
```
SuperFluffy, line 41: Maybe "the request body" or "request payload"?

crates/astria-core/src/protocol/transaction/v1alpha1/mod.rs
```diff
@@ -470,6 +473,83 @@ impl TransactionParams {
...
+    fn asset(inner: asset::ParseDenomError) -> Self {
+        Self(TransactionFeeResponseErrorKind::Asset(inner))
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+enum TransactionFeeResponseErrorKind {
+    #[error("`fee` field is unset")]
+    UnsetFee,
+    #[error("failed to parse asset denom")]
```
SuperFluffy, line 549: Maybe add the field, i.e. `failed to parse ok asset listed in the ``.assets`` field`

crates/astria-core/src/protocol/transaction/v1alpha1/mod.rs
```diff
@@ -470,6 +473,83 @@ impl TransactionParams {
...
+    Asset(#[from] asset::ParseDenomError),
```
SuperFluffy, line 550: Better just `#[source]` because I think we don't use the from conversion ever

crates/astria-sequencer/src/transaction/checks.rs
```diff
@@ -57,17 +57,15 @@ pub(crate) async fn check_balance_mempool<S: StateReadExt + 'static>(
...
 
-// Checks that the account has enough balance to cover the total fees and transferred values
-// for all actions in the transaction.
-pub(crate) async fn check_balance_for_total_fees<S: StateReadExt + 'static>(
+pub(crate) async fn get_fees_for_transaction<S: StateReadExt + 'static>(
     tx: &UnsignedTransaction,
-    from: Address,
     state: &S,
-) -> anyhow::Result<()> {
+) -> anyhow::Result<HashMap<asset::IbcPrefixed, u128>> {
```
SuperFluffy, line 68: Fine for doing this in a follow-up: cant we request all the data bits in this method concurrently instead of one after the other?

crates/astria-sequencer/src/transaction/checks.rs
```diff
@@ -148,7 +132,64 @@ pub(crate) async fn check_balance_for_total_fees<S: StateReadExt + 'static>(
...
+            | Action::Ibc(_)
+            | Action::IbcRelayerChange(_)
+            | Action::FeeAssetChange(_)
+            | Action::FeeChange(_) => {
+                continue;
+            }
+        }
+    }
+
+    for (asset, total_fee) in cost_by_asset {
```
SuperFluffy, line 192: Could fire these up concurrently too

proto/protocolapis/astria/protocol/transactions/v1alpha1/types.proto
```diff
@@ -246,3 +246,14 @@ message FeeChangeAction {
     astria.primitive.v1.Uint128 ics20_withdrawal_base_fee = 40;
   }
 }
+
+// Response to a transaction fee ABCI query.
+message TransactionFeeResponse {
+  uint64 height = 2;
```
SuperFluffy, line 252: Index from 1?

crates/astria-sequencer/src/transaction/checks.rs
```diff
@@ -164,16 +205,10 @@ pub(crate) async fn check_balance_for_total_fees<S: StateReadExt + 'static>(
 }
 
 fn transfer_update_fees(
```
SuperFluffy, line 207: Why is this body updated? Probably missing something, but from a cursory glance this looks orthogonal to the new functionality?

crates/astria-sequencer/src/transaction/checks.rs
```diff
@@ -197,16 +232,10 @@ async fn sequence_update_fees<S: StateReadExt>(
 }
 
 fn ics20_withdrawal_updates_fees(
-    asset: &asset::Denom,
```
SuperFluffy, line 200: Same here - is this functionality no longer necessary or wrong?

crates/astria-sequencer/src/transaction/checks.rs
```diff
@@ -233,37 +262,21 @@ fn bridge_lock_update_fees(
...
         .entry(act.asset.to_ibc_prefixed())
         .and_modify(|amt| *amt = amt.saturating_add(expected_deposit_fee))
         .or_insert(expected_deposit_fee);
 }
 
-async fn bridge_unlock_update_fees<S: StateReadExt>(
-    state: &S,
-    bridge_address: Address,
-    amount: u128,
+fn bridge_unlock_update_fees(
```
SuperFluffy, line 271: Same here - why all that change?

SuperFluffy (commented):
I left comments looking to understand why various code was updated (updating fees etc), even the PR just aims to read (not write)?

crates/astria-sequencer/src/transaction/checks.rs
noot: the function checking "fees" before was actually checking the total value used by the transaction, so it was summing fees and value transfers. however, the fees endpoint wants to return only fees used by the tx, so i had to separate these

crates/astria-sequencer/src/transaction/checks.rs
noot: yes - can make an issue

proto/protocolapis/astria/protocol/transactions/v1alpha1/types.proto
noot: we should probably change all of the response types if we want to index from 1, maybe we can do that when we move them to their own package?

crates/astria-sequencer/src/transaction/checks.rs
noot: same as above

crates/astria-sequencer/src/transaction/checks.rs
noot: same as above

crates/astria-sequencer/src/transaction/checks.rs
```diff
@@ -57,17 +57,15 @@ pub(crate) async fn check_balance_mempool<S: StateReadExt + 'static>(
     state: &S,
 ) -> anyhow::Result<()> {
     let signer_address = crate::address::base_prefixed(tx.verification_key().address_bytes());
-    check_balance_for_total_fees(tx.unsigned_transaction(), signer_address, state).await?;
+    check_balance_for_total_fees_and_transfers(tx.unsigned_transaction(), signer_address, state)
+        .await?;
```
SuperFluffy, line 61: context

crates/astria-sequencer/src/transaction/checks.rs
```diff
@@ -148,7 +132,64 @@ pub(crate) async fn check_balance_for_total_fees<S: StateReadExt + 'static>(
...
+}
+
+// Checks that the account has enough balance to cover the total fees and transferred values
+// for all actions in the transaction.
+pub(crate) async fn check_balance_for_total_fees_and_transfers<S: StateReadExt + 'static>(
+    tx: &UnsignedTransaction,
+    from: Address,
+    state: &S,
+) -> anyhow::Result<()> {
+    let mut cost_by_asset = get_fees_for_transaction(tx, state).await?;
```
SuperFluffy, line 145: context

crates/astria-sequencer/src/transaction/mod.rs
```diff
@@ -198,7 +199,7 @@ impl ActionHandler for UnsignedTransaction {
         ensure!(curr_nonce == self.nonce(), InvalidNonce(self.nonce()));
 
         // Should have enough balance to cover all actions.
-        check_balance_for_total_fees(self, from, state).await?;
+        check_balance_for_total_fees_and_transfers(self, from, state).await?;
```
SuperFluffy, line 202: context

SuperFluffy (approved):
approve for API.

few comments on missing error context

PR #1197, opened 2024-06-19
feat(conductor)!: add sequencer chain_id validation

eoroshiba:
## Summary
Added configured sequencer chain_id check against remote.

## Background
This is to prevent the conductor from connecting to the wrong sequencer network.

## Changes
Added sequencer chain ID configuration option for specifying expected sequencer chain ID.

Added check to ensure configured and remote chain IDs match within sequencer reader run_until_stopped() function.

Added function to retrieve the remote sequencer chain ID and check against configured chain ID. This function retrieves the remote chain ID from the sequencer network genesis file.

## Testing
Added new sequencer test file to allow access to sequencer without needing to mount anything other than the genesis file. This includes a test to ensure failure on mismatched chain IDs.

## Breaking Changelist
-  Added ASTRIA_CONDUCTOR_SEQUENCER_CHAIN_ID

## Related Issues
Part of #986


crates/astria-conductor/src/sequencer/tests.rs
```diff
@@ -0,0 +1,110 @@
...
+
+use crate::{
+    config::CommitLevel,
+    executor,
+    metrics::Metrics,
+    sequencer,
+    sequencer::CancellationToken,
+};
+
+async fn mount_genesis(mock_http: &MockServer, sequencer_chain_id: &str) {
```
noot, line 16: can you use the existing `helpers::mount_genesis` function instead of this? since this is only  used in the chain-id mismatch test, you can just set the `sequencer_chain_id` in the builder to a different value than the default set in `helpers::mount_genesis`.

eoroshiba: I might be mistaken, but I believe `helpers::mount_genesis` is only available to `tests/blackbox`. I created the test file within `src/sequencer` so that the functionality of the chain ID check could be tested without doing all the other necessary mounts/calls for the full conductor. Would you prefer to move this to the `tests/blackbox` folder and perform the mounts so that it can utilize `helpers::mount_genesis`?

PR #1201, opened 2024-06-20
feat(sequencer)!: allow configuring base address prefix

SuperFluffy:
## Summary
This patch allows configuring the base address prefix in sequencer's genesis file and enforces this prefix in all actions.

Similar to the native fee, the base prefix is set once during `init-chain` and is then available globally and never changed.

## Background
https://github.com/astriaorg/astria/pull/1124 changed astria addresses from opaque bytes to bech32m addresses, which themselves are encoded as strings on the wire. Where previously sequencer implicitly assumed that all addresses were `"astria"` prefixed bech32m addresses, this is now enforced.

## Changes
- Update sequencer `GenesisState` to use the canonical serde implementation of all `astria-core` types.
- Add a field `.address_prefixes.base` to the `GenesisState`.
- Ensure that all addresses in `GenesisState` are consistenst with the prefix.
- Inject the genesis base address prefix into sequencer's global state.
- Add a check for the global address prefix to all sequencer actions that carry a `astria.primitive.v1.core.Address` (this is done via `AppHandler::check_stateless`). These are specifically:
    - `TransferAction`
    - `SudoAddressChangeAction`
    - `BridgeLockAction`
    - `BridgeSudoChangeAction`
    - `BridgeUnlockAction`
    - `InitBridgeAccountAction`
    - `IbcRelayerChangeAction`
- Slightly change all storage keys that use addresses to be explicit about using bytes and provide snapshot tests (specifically in `accounts::state_ext`, `bridge::state_ext`, `ibc::state_ext`)
- Removed: unsigned transaction parameters no longer require a bech32m chain ID name
- Update composer, bridge-withdrawer to have a configurable address prefix:
    - `ASTRIA_BRIDGE_WITHDRAWER_SEQUENCER_ADDRESS_PREFIX`
    - `ASTRIA_COMPOSER_SEQUENCER_ADDRESS_PREFIX`
- Update all `astria-cli` commands to take a configurable prefix.

## Testing
+ Update all tests and ensure they still run.
+ Add tests for genesis construction.
+ Add genesis snapshot tests.
+ Add snapshots tests for IBC denoms (to accompany the genesis snapshot)

## Breaking Changelist
This change is breaking at the network level requiring sequencer re-genesis. Non-bech32m addresses in actions are now rejected, as well as addresses that have a prefix other then the one set at genesis. In addition, the storage key for all collections involving an address have changed.


crates/astria-sequencer/src/accounts/state_ext.rs
```diff
@@ -37,20 +37,20 @@ struct Fee(u128);
 const ACCOUNTS_PREFIX: &str = "accounts";
 const TRANSFER_BASE_FEE_STORAGE_KEY: &str = "transferfee";
 
-fn storage_key(address: &str) -> String {
+fn storage_key(address: &Address) -> String {
```
SuperFluffy, line 40: This fundamentally changes access to storage, segregating it by prefixes. Is this correct?

SuperFluffy: After syncing with the team: this is not correct and not what we want. The storage access must not care about the prefix, only about the bytes of the address (and hence the first 20 bytes of an account's public key).

crates/astria-bridge-withdrawer/src/bridge_withdrawer/ethereum/convert.rs
```diff
@@ -119,6 +122,8 @@ fn event_to_bridge_unlock(
     Ok(Action::BridgeUnlock(action))
 }
 
+// FIXME: Get this to work for now, but replace this with a builder.
```
noot, line 125: i'm refactoring/removing this in my next PR, so this is fine to leave

crates/astria-bridge-withdrawer/src/bridge_withdrawer/submitter/signer.rs
```diff
@@ -1,55 +1,107 @@
...
 
-impl SequencerKey {
-    /// Construct a `SequencerKey` from a file.
+pub(crate) struct SequencerKeyBuilder {
+    path: Option<PathBuf>,
+    prefix: Option<String>,
+}
+
+impl SequencerKeyBuilder {
+    /// Sets the path from which a the sequencey must be read.
```
noot, line 31:
```suggestion
    /// Sets the path from which a the sequencer key must be read.
```

crates/astria-core/src/primitive/v1/asset/denom.rs
```diff
@@ -509,6 +516,72 @@ impl FromStr for IbcPrefixed {
...
+                        S: Serializer,
+                    {
+                        serializer.collect_str(self)
+                    }
+                }
+            )*
+        }
+    }
+
+    impl_serde!(super::Denom, super::TracePrefixed, super::IbcPrefixed);
```
noot, line 554: are these supposed to be in this PR? seems somewhat unrelated

crates/astria-sequencer/src/accounts/snapshots/astria_sequencer__accounts__state_ext__test__snapshots-2.snap
```diff
@@ -0,0 +1,5 @@
+---
+source: crates/astria-sequencer/src/accounts/state_ext.rs
+expression: nonce_storage_key(address)
+---
+accounts/1c0c490f1b5528d8173c5de46d131160e4b2c0c3/nonce
```
noot, line 5: shouldn't this be a bech32 address, not hex? same for the balance snapshot

crates/astria-sequencer/src/app/tests_breaking_changes.rs
```diff
@@ -62,10 +62,32 @@ use crate::{
...
-    genesis::GenesisState,
+    genesis::{
+        AddressPrefixes,
+        UncheckedGenesisState,
+    },
     proposal::commitment::generate_rollup_datas_commitment,
 };
 
+/// XXX: This should be expressed in terms of `crate::app::test_utils::unchecked_genesis_state` to
+/// be consistent everywhere. `get_alice_sining_key` already is, why not this??
```
noot, line 73:
```suggestion
/// be consistent everywhere. `get_alice_signing_key` already is, why not this??
```

crates/astria-sequencer/src/genesis.rs
```diff
@@ -4,29 +4,157 @@ use astria_core::primitive::v1::{
...
     pub(crate) allowed_fee_assets: Vec<asset::Denom>,
     pub(crate) fees: Fees,
 }
 
-#[derive(Debug, Deserialize)]
+#[derive(Debug, thiserror::Error)]
+// allow: this error is only seen at chain init and never after so perf impact of too large enum
+// variants is negligible
+#[allow(clippy::result_large_err)]
+pub(crate) enum VerifiyGenesisError {
```
noot, line 36:
```suggestion
pub(crate) enum VerifyGenesisError {
```

crates/astria-sequencer/src/address/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+
+    static BASE_PREFIX: OnceLock<String> = OnceLock::new();
+
+    pub(crate) fn initialize_base_prefix(base_prefix: &str) -> anyhow::Result<()> {
+        // construct a dummy address to see if we can construct it; fail otherwise.
+        try_construct_dummy_address_from_prefix(base_prefix)
+            .context("failed constructing a dummy address from the provided prefix")?;
+
+        BASE_PREFIX.set(base_prefix.to_string()).expect(
+            "THIS IS A BUG: the base prefix was already more than once; it must only be \
```
joroshiba, line 53: nit: copy here is a bit off? Perhaps:

> THIS IS A BUG: attempted to set the base prefix more than once. It should be set once during `InitChain`. It cannot be initialized twice or concurrently from more than one task/thread.

crates/astria-sequencer/src/address/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+        );
+
+        Ok(())
+    }
+
+    pub(crate) fn get_base_prefix() -> &'static str {
+        BASE_PREFIX
+            .get()
+            .expect(
+                "the base prefix must have been set during chain init; if not set, the chain was \
```
joroshiba, line 65:
```suggestion
                "the base prefix must have been set during `InitChain`; if not set, the chain was \
```

This is the official ABCI method unlikely to change, might be easier to trace when called out.

noot (approved):
looks good to me!

crates/astria-sequencer/src/address/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+    ensure!(
+        get_base_prefix() == address.prefix(),
+        "address has prefix `{}` but only `{}` is permitted",
+        address.prefix(),
+        crate::address::get_base_prefix(),
+    );
+    Ok(())
+}
+
+#[cfg(not(test))]
```
joroshiba, line 37: Is there a way to have one test that trips this?

crates/astria-sequencer/src/authority/action.rs
```diff
@@ -74,6 +74,12 @@ impl ActionHandler for tendermint::validator::Update {
 
 #[async_trait::async_trait]
 impl ActionHandler for SudoAddressChangeAction {
+    async fn check_stateless(&self) -> Result<()> {
+        crate::address::ensure_base_prefix(&self.new_address)
+            .context("desired new sudo address is incorrect")?;
```
joroshiba, line 79: perhaps "desired new sudo address has incorrect prefix"?

crates/astria-sequencer/src/bridge/bridge_lock_action.rs
```diff
@@ -34,6 +34,11 @@ use crate::{
 
 #[async_trait::async_trait]
 impl ActionHandler for BridgeLockAction {
+    async fn check_stateless(&self) -> Result<()> {
+        crate::address::ensure_base_prefix(&self.to).context("destination address is invalid")?;
```
joroshiba, line 38: should this specify how it is invalid?

crates/astria-sequencer/src/bridge/bridge_sudo_change_action.rs
```diff
@@ -24,6 +24,22 @@ use crate::{
 
 #[async_trait::async_trait]
 impl ActionHandler for BridgeSudoChangeAction {
+    async fn check_stateless(&self) -> Result<()> {
+        crate::address::ensure_base_prefix(&self.bridge_address)
+            .context("bridge address is invalid")?;
```
joroshiba, line 29: same as previous, this is specifically wrong base prefix

crates/astria-sequencer/src/bridge/state_ext.rs
```diff
@@ -1112,4 +1144,16 @@ mod test {
             "nonce should have been deleted also"
         );
     }
+
+    #[test]
+    fn snapshots() {
```
joroshiba, line 1149: nit: more descriptive test name perhaps `storage_keys_have_not_changed` or something like that?

crates/astria-sequencer/src/ibc/state_ext.rs
```diff
@@ -422,4 +439,13 @@ mod test {
             "set balance for channel/asset pair not what was expected"
         );
     }
+
+    #[test]
+    fn snapshots() {
```
joroshiba, line 444: same nit on test name

joroshiba (commented):
A general note on composer & withdrawer is that we could get the prefix from genesis fetch.

Some nits through out but mostly looks good, didn't comment on each repeat but the stateless checks all just say "invalid" when they are specifically invalid because of wrong prefix.

Note that the ibc name stuff seems unrelated, is that artifact from other work?

crates/astria-core/src/primitive/v1/asset/denom.rs
SuperFluffy: Yeah, this is needed because `GenesisState` relies entirely on derived serde impls contains straight `asset::Denom` (either directly as in `allowed_fee_assets: Vec<asset::Denom>` or through the sequencer-defined `Account`).

crates/astria-core/src/primitive/v1/asset/denom.rs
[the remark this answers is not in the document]
SuperFluffy: Notes in another comment: the serde impls are required for sequencer's `GenesisState`. `IbcPrefixed::new` is a reasonable (and minimal) utility.

crates/astria-bridge-withdrawer/src/bridge_withdrawer/submitter/signer.rs
SuperFluffy: Addressed in bbbe762

crates/astria-sequencer/src/app/tests_breaking_changes.rs
SuperFluffy: in bbbe762

crates/astria-sequencer/src/genesis.rs
SuperFluffy: Hah, good catch. Auto-complete spread it everywhere. bbbe762

crates/astria-sequencer/src/accounts/state_ext.rs
```diff
@@ -705,4 +723,19 @@ mod test {
             .await
             .expect_err("should not be able to subtract larger balance than what existed");
     }
+
+    #[test]
+    fn snapshots() {
```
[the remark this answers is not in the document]
SuperFluffy: The setup is a bit lengthy, but this is the suggested way to just group all snapshots together.

crates/astria-sequencer/src/bridge/bridge_sudo_change_action.rs
SuperFluffy: Clarified the error in:  1634af

crates/astria-sequencer/src/bridge/bridge_lock_action.rs
SuperFluffy: Clarified the error in:  1634af

crates/astria-sequencer/src/authority/action.rs
SuperFluffy: Clarified the error in:  1634af

crates/astria-sequencer/src/bridge/state_ext.rs
SuperFluffy: Good idea. Done for this and the other snapshot tests in  5fd2ae5

crates/astria-sequencer/src/address/mod.rs
SuperFluffy: I don't think I can actually actively trigger this unfortunately. :(

I have an idea how to rework this (and the native asset global) in a follow up though.

crates/astria-sequencer/src/ibc/state_ext.rs
SuperFluffy: Done in 5fd2ae5

crates/astria-cli/src/cli/sequencer.rs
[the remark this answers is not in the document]
SuperFluffy: Good idea. Using `"astria"` as the default value so that these commands are the same:

```
> cargo run -p astria-cli -- sequencer address bech32m --bytes 00d75b270542084a54fcf0d0f6eab0402982d156
astria1qrt4kfc9ggyy548u7rg0d64sgq5c952kzk9tg9

> cargo run -p astria-cli -- sequencer address bech32m --prefix astria --bytes 00d75b270542084a54fcf0d0f6eab0402982d156
astria1qrt4kfc9ggyy548u7rg0d64sgq5c952kzk9tg9
```

crates/astria-cli/src/cli/sequencer.rs
```diff
@@ -96,16 +101,29 @@ pub struct BasicAccountArgs {
...
 pub struct TransferArgs {
     // The address of the Sequencer account to send amount to
-    pub(crate) to_address: SequencerAddressArg,
+    pub(crate) to_address: Address,
     // The amount being sent
     #[arg(long)]
     pub(crate) amount: u128,
+    /// The bech32m prefix that will be used for constructing addresses using the private key
+    #[arg(long)]
+    pub(crate) prefix: String,
```
joroshiba, line 126: is this just for logging? Doesn't seem necessary?

SuperFluffy: It's needed to construct an address and get the latest nonce - although if you follow the calls in `sequencer_client::SequencerClientExt`, then ultimately it only uses the bytes contained in the address. All of this looks pretty bad.

I am updating the queries in sequencer to parse the addresses from the provided paths.

SuperFluffy: Turns out the queries didn't need updating - just `astria-sequencer-client` would have been broken because sequencer was already updated.

I have updated all commands that take `--prefix` to use `"astria"` as the default.

crates/astria-sequencer/src/app/snapshots/astria_sequencer__app__tests_breaking_changes__app_execute_transaction_with_every_action_snapshot.snap
SuperFluffy, line 1: This and the other snapshots have changed because the base prefix is now written to app storage.

charts/sequencer/values.yaml
```diff
@@ -37,15 +39,15 @@ config:
...
-    - address: 1c0c490f1b5528d8173c5de46d131160e4b2c0c3
-      balance: "333333333333333333"
-    - address: 34fec43c7fcab9aef3b3cf8aba855e41ee69ca3a
-      balance: "333333333333333333"
-    - address: 60709e2d391864b732b4f0f51e387abb76743871
-      balance: "333333333333333333"
-      # NOTE - the following address matches the privKey that funds the sequencer-faucet
-    - address: 00d75b270542084a54fcf0d0f6eab0402982d156
-      balance: "333333333333333333"
+      - address: 1c0c490f1b5528d8173c5de46d131160e4b2c0c3
```
noot, line 42: just to confirm - does the genesis support both hex and bech32 now?

joroshiba: So the chart is indifferent, when you run in dev mode it builds the old genesis that is intended to work with the latest release. When you run in `dev` mode it builds the new genesis which works on this PR and latest after this is merged.

crates/astria-sequencer/src/address/mod.rs
joroshiba: Can we get a follow up issue?

joroshiba (approved):

charts/sequencer/values.yaml
SuperFluffy: Yeah to chime in: this is only for CI. The sequencer genesis itself only supports proper bech32m addresses.

SuperFluffy:
@joroshiba https://github.com/astriaorg/astria/issues/1208 for a follow-up for the `cfg(not(test))` business. It doens't explicitly mention it, but that change would make the compile-time shenanigans unnecessary.

PR #1202, opened 2024-06-20
feat(sequencer)!: implement `Ics20TransferDepositMemo` format for incoming ics20 transfers to bridge accounts

noot:
## Summary
 implement `Ics20TransferDepositMemo` format and handling for incoming ics20 transfers to bridge accounts, which requires the memo be of a specific format instead of assuming the entire memo is the destination chain address.

## Background
the memo being just a destination chain address is not good for UX and could lead to error.

## Changes
- implement `Ics20TransferDepositMemo` type which contains the destination `rollup_address`
- add stateless check that the `FungibleTokenPacketData` (which is the ics20 transfer packet itself) is not empty or too long
-  parsing and handling of the memo for incoming ics20 transfers

## Testing
unit tests

## Breaking Changelist
- the validation of the ICS20 memo is different, so this will break previously sent ICS20 transfers to bridge accounts.

### DIFF
--- a/crates/astria-core/src/bridge.rs
+++ b/crates/astria-core/src/bridge.rs
@@ -14,7 +14,10 @@ pub struct Ics20WithdrawalFromRollupMemo {
     pub block_number: u64,
     #[cfg_attr(
         feature = "serde",
-        serde(serialize_with = "crate::serde::base64_serialize")
+        serde(
+            serialize_with = "crate::serde::base64_serialize",
+            deserialize_with = "crate::serde::base64_deserialize_array"
+        )
     )]
     pub transaction_hash: [u8; 32],
 }
--- a/crates/astria-core/src/bridge.rs
+++ b/crates/astria-core/src/bridge.rs
@@ -12,6 +12,10 @@ pub struct Ics20WithdrawalFromRollupMemo {
     pub memo: String,
     pub bridge_address: Address,
     pub block_number: u64,
+    #[cfg_attr(
+        feature = "serde",
+        serde(serialize_with = "crate::serde::base64_serialize")
+    )]
     pub transaction_hash: [u8; 32],
 }
 
@@ -21,11 +25,10 @@ pub struct Ics20WithdrawalFromRollupMemo {
 #[cfg_attr(
     feature = "serde",
     derive(serde::Serialize),
-    derive(serde::Deserialize)
+    derive(serde::Deserialize),
+    serde(rename_all = "camelCase", deny_unknown_fields)
 )]
-#[serde(rename_all = "camelCase", deny_unknown_fields)]
 pub struct Ics20TransferDepositMemo {
     /// the destination address for the deposit on the rollup
-    #[serde(rename = "rollupAddress")]
     pub rollup_address: String,
 }
--- a/crates/astria-core/src/bridge.rs
+++ b/crates/astria-core/src/bridge.rs
@@ -1,5 +1,7 @@
 use crate::primitive::v1::Address;
 
+/// Memo format for a ICS20 withdrawal from the rollup which is sent to
+/// an external IBC-enabled chain.
 #[derive(Debug, Clone)]
 #[cfg_attr(
     feature = "serde",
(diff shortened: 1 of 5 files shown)


crates/astria-sequencer/src/ibc/ics20_transfer.rs
```diff
@@ -560,12 +575,16 @@ async fn execute_ics20_transfer_bridge_lock<S: StateWriteExt>(
...
 
+    // assert memo is valid
+    let deposit_memo: Ics20TransferDepositMemo =
+        serde_json::from_str(&memo).context("failed to parse memo as Ics20TransferDepositMemo")?;
+
     ensure!(
-        !destination_address.is_empty(),
+        !deposit_memo.rollup_address.is_empty(),
         "packet memo field must be set for bridge account recipient",
     );
```
joroshiba, line 585: We can probably add additional validation on length? Max 256 length?

joroshiba (approved):

crates/astria-sequencer/src/ibc/ics20_transfer.rs
noot: since we already check the entire packet length, additional validation here just put constraints on the rollup address format. however we can probably assume rollup addresses aren't >256 bytes so will add

crates/astria-core/src/bridge.rs
```diff
@@ -12,3 +14,18 @@ pub struct Ics20WithdrawalFromRollupMemo {
     pub block_number: u64,
     pub transaction_hash: [u8; 32],
```
SuperFluffy, line 15: This makes for a very ugly json. Use base64 for the hash?

crates/astria-core/src/bridge.rs
```diff
@@ -12,3 +14,18 @@ pub struct Ics20WithdrawalFromRollupMemo {
...
+
+/// Memo format for a ICS20 transfer to Astria which is sent to a
+/// bridge account, which will then be deposited into the rollup.
+#[derive(Debug, Clone)]
+#[cfg_attr(
+    feature = "serde",
+    derive(serde::Serialize),
+    derive(serde::Deserialize)
+)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
```
SuperFluffy, line 26: I got hung up on this, but it turns out that `camelCase` is the actually correct pbjson mapping, whereas we use `preserve_proto_field_names` in the protobuf generated. I don't recall why I did that, but I shouldn't have done that.

Also a note on `deny_unknown_fields`: this makes schema evolution extremely tricky and will require a breaking change on every change to this memo. Not that it's wrong - just want to flag it.

crates/astria-core/src/bridge.rs
```diff
@@ -12,3 +14,18 @@ pub struct Ics20WithdrawalFromRollupMemo {
...
+pub struct Ics20TransferDepositMemo {
+    /// the destination address for the deposit on the rollup
+    #[serde(rename = "rollupAddress")]
```
SuperFluffy, line 29: The rename here is unnecessary becase `rename_all = "camelCase"` should cover this already.

crates/astria-sequencer/src/ibc/ics20_transfer.rs
```diff
@@ -560,12 +578,21 @@ async fn execute_ics20_transfer_bridge_lock<S: StateWriteExt>(
         return Ok(());
     }
 
+    // assert memo is valid
+    let deposit_memo: Ics20TransferDepositMemo =
+        serde_json::from_str(&memo).context("failed to parse memo as Ics20TransferDepositMemo")?;
+
     ensure!(
-        !destination_address.is_empty(),
+        !deposit_memo.rollup_address.is_empty(),
```
SuperFluffy, line 586: Does this really need extra checking? We can't really know if the destination address is valid or not: what if I put `"."` or `"_"` or `"\n"` in there?

SuperFluffy (commented):
`Ics20TransferDepositMemo` should have a snapshot test for its json serialization (as well as the other types defined in `core::bridge`, but that can be left to a follow-up). This is because it's a public API and to ensure that we don't accidentally break our consumers.

Orthgonal to this: I brought it up in yesterday's meeting, but I think I would like to define all these objects via protobuf: even if we don't ever send these over the wire protobuf-encoded, we still have an easier way of enforcing a consistent json format.

crates/astria-sequencer/src/ibc/ics20_transfer.rs
noot: it must not be empty, otherwise there is no chance it's a valid rollup address, we can't check for every case obviously but nonempty is pretty simple

crates/astria-core/src/bridge.rs
joroshiba: TX hashes should probably be in hex, since that is how they are looked up in cometbft

PR #1205, opened 2024-06-24
tests(core): add ibc memo type snapshot tests

noot:
## Summary
follow up to #1202  - add snapshot tests for the ibc memo types.

## Background
these are protocol-breaking if they change.

## Changes
- add ibc memo type snapshot tests

### DIFF
--- a/crates/astria-core/Cargo.toml
+++ b/crates/astria-core/Cargo.toml
@@ -59,5 +59,6 @@ brotli = ["dep:brotli"]
 [dev-dependencies]
 insta = { workspace = true, features = ["json"] }
 rand = { workspace = true }
+serde_json = { workspace = true }
 tempfile = { workspace = true }
 astria-core = { path = ".", features = ["serde"] }
--- a/crates/astria-core/Cargo.toml
+++ b/crates/astria-core/Cargo.toml
@@ -59,6 +59,5 @@ brotli = ["dep:brotli"]
 [dev-dependencies]
 insta = { workspace = true, features = ["json"] }
 rand = { workspace = true }
-serde_json = { workspace = true }
 tempfile = { workspace = true }
 astria-core = { path = ".", features = ["serde"] }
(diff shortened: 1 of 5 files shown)


crates/astria-core/Cargo.toml
```diff
@@ -59,5 +59,6 @@ brotli = ["dep:brotli"]
 [dev-dependencies]
 insta = { workspace = true, features = ["json"] }
 rand = { workspace = true }
+serde_json = { workspace = true }
```
SuperFluffy, line 62: remove

crates/astria-core/src/bridge.rs
```diff
@@ -35,3 +35,34 @@ pub struct Ics20TransferDepositMemo {
...
+            bridge_address: Address::builder()
+                .array([99; 20])
+                .prefix(ASTRIA_ADDRESS_PREFIX)
+                .try_build()
+                .unwrap(),
+            block_number: 1,
+            transaction_hash: [88; 32],
+        };
+
+        insta::assert_json_snapshot!(serde_json::to_string(&memo).unwrap());
```
SuperFluffy, line 57: remove the string construction - `assert_json_snapshot` does that for you

crates/astria-core/src/bridge.rs
```diff
@@ -35,3 +35,34 @@ pub struct Ics20TransferDepositMemo {
...
+    }
+
+    #[test]
+    fn ics20_transfer_deposit_memo_snapshot() {
+        let memo = Ics20TransferDepositMemo {
+            rollup_address: "some_rollup_address".to_string(),
+        };
+
+        insta::assert_json_snapshot!(serde_json::to_string(&memo).unwrap());
```
SuperFluffy, line 66: as above, just pass in `memo` without `serde_json`

SuperFluffy (changes requested):

crates/astria-core/src/bridge.rs
noot: done

SuperFluffy (approved):
Looks great, thanks!

PR #1206, opened 2024-06-25
fix(sequencer-relayer): avoid hanging while waiting for submitter task to return

### LINKED ISSUE #1095: (sequencer-relayer): exit process if reader task ends
Currently if the reader task exits the endless loop, it hangs [here](https://github.com/astriaorg/astria/blob/3921313173b7ba7df6ca4c26ff2d290651621bfe/crates/astria-sequencer-relayer/src/relayer/mod.rs#L225) awaiting the submitter task to exit.

To address this, we should:
- create a newtype wrapping a tokio `CancellationToken` which doesn't implement `Clone`, but does provide `child_token`
- replace our current use of naked `CancellationToken`s with this newtype
- in each place we currently clone the token, instead spawn a child token
- before awaiting the submitter task, cancel the child token

Fraser999:
## Summary
This fixes a bug whereby the relayer hangs during shutdown due to the submitter task not being instructed to shut down.

## Background
Bugfix.

## Changes
- Rather than cloning a single cancellation token, we now have the following hierarchy:
    - `SequencerRelayer::new` constructs a `ShutdownHandle` which wraps the top-level cancellation token.  This handle is held as a member variable of `SequencerRelayer` and is used to cancel all relayer child tasks if the API task exits. A clone of the `ShutdownHandle`  is also held in `main` and is used to cancel all tasks if a `SIGTERM` is received.
    - A child of `ShutdownHandle`'s token is provided to the relayer task, named `relayer_shutdown_token`.  It is never actively cancelled - only ever via the parent token.
    - A child of `relayer_shutdown_token` is provided to the submitter task and also held as a member of `Relayer`.  It is actively cancelled after the relayer's main `select!` loop has exited if the submitter task is still running.

## Testing
Manual test only.  I could see no way to artificially cause the relayer's `select!` loop to exit.  We could potentially try dropping in some `#[cfg(test)]` gated code in the loop to force this, but that would also mean we couldn't use the blackbox tests for it. 

## Related Issues
Closes #1095.

### DIFF
--- a/crates/astria-sequencer-relayer/src/relayer/mod.rs
+++ b/crates/astria-sequencer-relayer/src/relayer/mod.rs
@@ -78,7 +78,12 @@ use crate::{
 
 pub(crate) struct Relayer {
     /// A token to notify relayer that it should shut down.
-    shutdown_token: CancellationToken,
+    // allow: want the prefix to disambiguate between this token and `submitter_shutdown_token`.
+    #[allow(clippy::struct_field_names)]
+    relayer_shutdown_token: CancellationToken,
+
+    /// A child token of `relayer_shutdown_token` to notify the submitter task to shut down.
+    submitter_shutdown_token: CancellationToken,
 
     /// The configured chain ID of the sequencer network.
     sequencer_chain_id: String,
@@ -124,7 +129,7 @@ impl Relayer {
             .wrap_err("failed reading submission state from files")?;
 
         select!(
-            () = self.shutdown_token.cancelled() => return Ok(()),
+            () = self.relayer_shutdown_token.cancelled() => return Ok(()),
             init_result = confirm_sequencer_chain_id(
                 self.sequencer_chain_id.clone(),
                 self.sequencer_cometbft_client.clone()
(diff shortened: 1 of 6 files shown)


crates/astria-sequencer-relayer/src/relayer/mod.rs
```diff
@@ -167,11 +172,13 @@ impl Relayer {
             select!(
                 biased;
 
-                () = self.shutdown_token.cancelled() => {
+                () = self.relayer_shutdown_token.cancelled() => {
                     info!("received shutdown signal");
                     break Ok("shutdown signal received");
                 }
 
+                _ = &mut submitter_task => break Err(eyre!("Celestia submission task returned")),
```
Fraser999, line 180: This is an unfortunate side-effect of changing to use child cancellation tokens - the submitter task (child) is no longer able to cancel the relayer task (parent).  Since the submitter task can exit with error during init, we need to poll for that here or else we hang in this `select!` loop.

This also means we need to `fuse` the submitter task and outside the loop we need to check whether it's already terminated or not before awaiting it, or else we'd hang there.

There are other potential approaches to address that, but I couldn't see anything cleaner than this (except for my still-preferred approach of a single cancellation token for all tasks rather than this hierarchy).

crates/astria-sequencer-relayer/tests/blackbox/helpers/test_sequencer_relayer.rs
```diff
@@ -607,6 +608,7 @@ impl TestSequencerRelayer {
             .and_then(|value: serde_json::Value| serde_json::from_value::<ZPage>(value).ok())
             .map_or("unknown".to_string(), |zpage| zpage.status);
 
+            error!("timed out; context: `{context}`, state: `{state}`, healthz: `{healthz}`");
```
Fraser999, line 611: This helps diagnose the point at which the test fails, since we can get several further log lines after this point before the panic is printed.

noot (approved):

PR #1209, opened 2024-06-26
feat(proto, sequencer)!: use full IBC ICS20 denoms instead of IDs

SuperFluffy:
## Summary
Replaces the bare asset ID bytes by full IBC ICS20 denoms.

## Background
Bare denom IDs in public APIs are difficult to understand. While IBC ICS20 denoms are prescribed to either contain strings of the form `port/channel/denom` or `ibc/<sha256-hex>`, due to protobuf-to-json mapping protobuf `bytes` are always encoded as base64. With https://github.com/astriaorg/astria/pull/1181 we parse denoms into either of the two forms and can thus allow users to provide assets as protobuf `string`s which are then parsed by sequencer.

In addition, since denominations of the form `port/channel/denom` can always be converted to `ibc/<sha256>`, this allows for more elegant fetching of denom-related data from storage.

## Changes
- Replace protobuf message fields from bytes to string (mostly in transaction/action types):
  - `bytes asset_id -> string asset`
  - `bytes fee_asset_id -> string fee_asset`
  - `repeated bytes fee_asset_ids -> repeated string fee_assets` (and rename `AllowedFeeAssetIdsResponse` to `AllowedFeeAsssetsRersponse`)
- Update all affected checked `astria-core` types to parse the asset strings to `astria_core::primitive::v1::asset::Denom`s instead of `[u8; 32]` arrays.
- Remove the `astria_core::primitive::v1::asset::Id` type
- Remove all notion of a default native asset type from astria_core (all services and tests are now expected to have their own notion of asset types; either through config/genesis or by defining their tests accordingly)
- Update all sequencer `StateWriteExt` and `StateWriteRead` types that read or write assets to disk:
  - instead of writing `asset::Id`s (in practice `[u8; 32]`), they now write `asset::IbcPrefixed`
  - make all trait methods generic over a parameter `TAsset: Into<asset::IbcPrefixed>` which allows fetching data using `astria::Denom`s, `astria::TracePrefixed`, or `astria::IbcPrecixed`
- update all storage keys that used IDs to use ibc-prefixed denoms instead
- provide snapshot tests for all such storage keys 
- update all `astria-cli` subcommands that construct actions with assets to have arguments `--asset` or `--fee-asset` (using `"nria"` as the default if not provided)
- update astria-composer to have an config env var `ASTRIA_COMPOSER_FEE_ASSET`

## Testing
All tests that involve fees were updated and pass again.

## Breaking Changelist
- This is a network breaking change because all transaction/action messages that took bytes now take strings


crates/astria-composer/local.env.example
```diff
@@ -69,6 +69,9 @@ ASTRIA_COMPOSER_METRICS_HTTP_LISTENER_ADDR="127.0.0.1:9000"
 # The address at which the gRPC collector and health services are listening.
 ASTRIA_COMPOSER_GRPC_ADDR="0.0.0.0:0"
 
+# The asset to use for paying for transactions submitted to sequencer.
+ASTRIA_COMPOSER_FEE_ASSET="nria"
```
SuperFluffy, line 73: Note that this sets the payment fee for all actions to `nria` - the same as was previously hard coded. This might not be exactly what we want to do in the future.

crates/astria-composer/src/executor/bundle_factory/mod.rs
```diff
@@ -264,6 +275,7 @@ impl<'a> NextFinishedBundle<'a> {
 
 /// The size of the `seq_action` in bytes, including the rollup id.
 fn estimate_size_of_sequence_action(seq_action: &SequenceAction) -> usize {
+    const FEE_ASSET_ID_LEN: usize = 32;
```
SuperFluffy, line 278: Note here: this is likely wrong now - and also unnecessary. We know the the byte-length of the asset string.

crates/astria-composer/src/executor/bundle_factory/tests.rs
```diff
@@ -1,19 +1,22 @@
-#[cfg(test)]
```
SuperFluffy, line 1: The structure of these modules was extremely strange and needed to be updated.

crates/astria-sequencer/src/asset/query.rs
```diff
@@ -96,18 +99,16 @@ fn preprocess_request(params: &[(String, String)]) -> anyhow::Result<asset::Id,
             ..response::Query::default()
         });
     };
-    let asset_id = hex::decode(asset_id)
+    let asset = <[u8; 32]>::from_hex(asset_id)
```
SuperFluffy, line 102: I am torn on this one. The request still has the form `asset/denom/<hex>`. So `asset/denom/ibc/<hex>` isn't allowed, and neither is `asset/denom/port/channel/base`.

crates/astria-sequencer/src/accounts/snapshots/astria_sequencer__accounts__state_ext__tests__storage_keys_have_not_changed.snap
joroshiba, line 1: What is the upside of this over the sha32 as a key?

crates/astria-sequencer/src/accounts/state_ext.rs
```diff
@@ -49,12 +48,11 @@ impl<'a> std::fmt::Display for StorageKey<'a> {
...
-    format!(
-        "{}/balance/{}",
-        StorageKey(&address),
-        asset.encode_hex::<String>()
-    )
+fn balance_storage_key<TAsset: Into<asset::IbcPrefixed>>(
+    address: Address,
+    asset: TAsset,
+) -> String {
+    format!("{}/balance/{}", StorageKey(&address), asset.into(),)
```
noot, line 55: i would prefer the storage keys continue to use the 32 byte asset ID - adding the `ibc/` prefix in front of every asset ID seems unnecessary and just adds extra data to storage, i don't really see a benefit to adding the prefix and it introduces ibc-specific naming conventions into the internal db which seems weird

crates/astria-bridge-withdrawer/src/bridge_withdrawer/submitter/mod.rs
```diff
@@ -87,7 +87,8 @@ pub(super) struct Submitter {
     signer: SequencerKey,
     sequencer_chain_id: String,
     startup_tx: oneshot::Sender<SequencerStartupInfo>,
-    expected_fee_asset_id: asset::Id,
+    expected_fee_asset: asset::Denom,
+    expected_fee_asset_ibc: asset::denom::IbcPrefixed,
```
noot, line 91: aren't these are the same (ibc-prefixed is just a hash of the denom)? why do we need both here?

crates/astria-cli/src/commands/sequencer.rs
```diff
@@ -111,7 +107,7 @@ pub(crate) async fn get_balance(args: &BasicAccountArgs) -> eyre::Result<()> {
 
     println!("Balances for address: {}", args.address);
     for balance in res.balances {
-        println!("    asset ID: {}", balance.denom.id());
+        println!("    asset ID: {}", balance.denom.to_ibc_prefixed());
```
noot, line 110: this is somewhat confusing as it'll print the native asset as IBC-prefixed when it isn't an IBC'd asset - maybe just leave as the 32-byte ID?

crates/astria-composer/src/executor/bundle_factory/mod.rs
noot: the max size should be 36 (32-byte id plus `ibc/` prefix`), if the denom string is >32 bytes it must get converted to the `ibc/` prefixed form

crates/astria-core/src/protocol/transaction/v1alpha1/action.rs
```diff
@@ -369,7 +369,7 @@ pub struct SequenceAction {
     pub rollup_id: RollupId,
     pub data: Vec<u8>,
     /// asset to use for fee payment.
-    pub fee_asset_id: asset::Id,
+    pub fee_asset: asset::Denom,
```
noot, line 372: we should enforce for all actions that if the denom trace is >32 bytes, it must be in `ibc/` prefixed form. ie don't accept traces that are excessively long as it's a DOS vector

crates/astria-cli/src/commands/sequencer.rs
SuperFluffy: IMO we should be consistent everywhere when communicating assets to users: as long as the string that you put into an action as an asset adheres to our ibc ics20 spec, then it will be accepted by sequencer.

Similarly, CLI tools should also return a valid denom so that you can just take it and paste it without having to know that you need to prefix it with `ibc/`.

What do you think if I change the line to read `asset IBC form:`?

crates/astria-core/src/protocol/transaction/v1alpha1/action.rs
SuperFluffy: Do you mean we should enforce this when parsing `string fee_asset -> asset::Denom`? Technically the upper limit for protobuf payloads is 4MB and sequencer just rejects payloads >200KB, but I can add an additional limit on the type of course.

crates/astria-sequencer/src/accounts/snapshots/astria_sequencer__accounts__state_ext__tests__storage_keys_have_not_changed.snap
SuperFluffy: Are you referring to the `ibc/<sha32>` bit as a suffix?

crates/astria-sequencer/src/accounts/state_ext.rs
SuperFluffy: I have addressed this (and also all other instances where assets are used) in e968da4

SuperFluffy:
@noot I have changed all keys that in some form employ ibc-prefixed assets to use the hex encoded sha256 in e968da4, in a new module `crate::storage_keys::hunks::Asset`.

crates/astria-bridge-withdrawer/src/bridge_withdrawer/submitter/mod.rs
SuperFluffy: You are right: it's enough to construct the `_ibc` one once in the setup method (so that it's not reconstructed in the setup's 2 loops):  3a8066b

crates/astria-sequencer/src/accounts/snapshots/astria_sequencer__accounts__state_ext__tests__storage_keys_have_not_changed.snap
SuperFluffy: I am not 100% if this referred the `ibc/<sha256>` bit, but in e968da4 I am again using only the sha256 bytes. I hope this resolves it, if not - please reopen.

crates/astria-cli/src/commands/sequencer.rs
SuperFluffy: Following this line of thought, I have decided to just remove it completely for now: sequencer right now returns the `ibc/<sha256>` right now, so the CLI tool doesn't need to print it twice. We can follow up retouching the tool to return both ibc and full trace-prefixed. ac4fe3

crates/astria-composer/src/executor/bundle_factory/mod.rs
SuperFluffy: The effective size is actually 68 bytes: hex-formatting doubles the size of 32 bytes to 64 (+4 due to `ibc/`). We need to decide if we prefer legibility or saving 36 bytes (note that from the PoV of sequencer-relayer this likely does not matter due to compression).

SuperFluffy: I have addressed this in [981930e](https://github.com/astriaorg/astria/pull/1209/commits/981930ef11dfc3f1db2501ef54b016fbb91d4888).

The size estimator overall was silly: we can use `prost::Message::encoded_len` to get the actual true length of the action as it becomes a protobuf. (there is a non-zero cost to go `sequence action -> proto sequence action -> proto len`, but that's likely negligible for a service like composer).

noot (approved):
looks good, great change!

joroshiba (approved):
approval for infra components

crates/astria-composer/src/executor/bundle_factory/tests.rs
```diff
@@ -110,60 +80,61 @@ mod sized_bundle {
...
-        bundle.try_push(seq_action2).unwrap();
-
-        let mut settings = Settings::new();
-        settings.set_sort_maps(true);
-
-        settings.bind(|| {
-            assert_json_snapshot!(bundle.rollup_counts);
-        });
-    }
+    // #[test]
```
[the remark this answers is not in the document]
SuperFluffy: I ended up removing the snapshot tests inside composer in  f06975d. There really is no point to have it. It's contents and formatting are dictated by the display implementations of `HashMap`, `RollupId` and `usize`, and their json formatting isn't part of any public API as far as I can see.

PR #1211, opened 2024-06-26
refactor(sequencer): fix prepare proposal metrics

Lilyjjo:
## Summary
The sequencer's `prepare_proposal()` metrics got shifted to `process_proposal()` across some refactors.

## Changes
Put metrics back in proper places and removed unneeded metric.

## Testing
Ran locally 

## Metrics
Removed the `PREPARE_PROPOSAL_EXCLUDED_TRANSACTIONS_DECODE_FAILURE` metric as the mempool should be filtering these transactions out.


crates/astria-sequencer/src/app/mod.rs
```diff
@@ -562,6 +563,8 @@ impl App {
                         // removed from the mempool in `update_mempool_after_finalization`.
                         txs_to_readd_to_mempool.push((enqueued_tx, priority));
                     } else {
+                        failed_tx_count = failed_tx_count.saturating_add(1);
```
Lilyjjo, line 566: I'm not really sure what this log is for. Do we use it for alters somewhere?

noot: pretty sure it's only used for logging at the moment, but useful to have

noot (approved):

Issue #1212, opened 2024-06-26
sequencer: local testing file `test-genesis-app-state.json` is brittle


Lilyjjo:
The local testing file `astria-sequencer/test-genesis-app-state.json` is brittle and there are no tests to ensure that it doesn't break.

We should reword the sequencer's local testing to not rely on this hardcoded file.

PR #1213, opened 2024-06-26
fix(sequencer): update out of sync local testing file

Lilyjjo:
## Summary
We use a hardcoded genesis file for the local testing of the sequencer. It broke with the recent shift to the astria prefixed bech32m addresses. 

## Changes
Updated file to work again.

## Related Issues
#1212 : We should refactor this testing setup to not be so brittle.


Lilyjjo:
Closing as @SuperFluffy is already busy with a syncing fix

PR #1215, opened 2024-06-26
fix(bridge-withdrawer)!: fix nonce handling

itamarreif:
## Summary
Adds a check that waits for an empty mempool on startup and then uses the pending nonce instead of latest finalized nonce for transaction submission.

## Background
Before these changes we used the latest nonce, which could lead to data races if there are batches awaiting inclusion in the mempool. See https://github.com/astriaorg/astria/issues/1228 for more context.

## Changes
- Add `ASTRIA_BRIDGE_WITHDRAWER_SEQUENCER_GRPC_ENDPOINT` config
- Wait for mempool to be empty as part of startup
- Submitter connects to the grpc service as part of its startup
- Submitter uses pending nonces for batch submission instead of latest nonce

## Testing
Removed most testing; reworked in https://github.com/astriaorg/astria/issues/1232

## Metrics
- Punted: https://github.com/astriaorg/astria/issues/1272 

## Breaking Changelist
- Adds a required env var `ASTRIA_BRIDGE_WITHDRAWER_SEQUENCER_GRPC_ENDPOINT`

## Related Issues
closes https://github.com/astriaorg/astria/issues/1228


itamarreif:
This required calling the sequencer's gRPC service, which broke the tests. I am working on a separate PR that would add the black box tests (#1227) instead of directly fixing this.

PR #1224, opened 2024-06-28
feat(sequencer-utils): generate example genesis state

SuperFluffy:
## Summary
Provides a `generate-genesis-state` subcommand to sequencer-utils.

## Background
We had a `test-genesis.json` file committed to the repo that was not checked to be in-sync with the most recent version of sequencer's actual `GenesisState`. The subcommand `generate-genesis-state` removes the need to keep the file in sync because one can always generate one that matches the current format.

## Changes
- remove `astria-sequencer/test-genesis-app-state.json`
- move sequencer's `GenesisState` to the new `astria-core::sequencer` submodule. Because this config is effectively a public API, it is justified to be outside of the `astria-sequencer` binary crate.

## Testing
It is a very thin binary which can be run by hand. The specific format of `GenesisState` not changing is ensured by a snapshot test that was moved from `astria-sequencer` to `astria-core`.

## Related Issues
Closes https://github.com/astriaorg/astria/issues/1212


crates/astria-sequencer/README.md
```diff
@@ -101,10 +101,14 @@ abci-cli query --path=accounts/nonce/<ADDRESS> 0x00
 # initialize the node
 cometbft init
 
-# inside astria-sequencer, update the genesis file to include genesis
-# application state
-../../target/debug/astria-sequencer-utils \
-    --genesis-app-state-file=test-genesis-app-state.json \
+# generate a sequencer genesis file
+cargo run -p astria-sequencer-utils -- generate-genesis-state -o genesis-state-json
```
noot, line 105:
```suggestion
cargo run -p astria-sequencer-utils -- generate-genesis-state -o genesis-state.json
```

crates/astria-sequencer/README.md
```diff
@@ -101,10 +101,14 @@ abci-cli query --path=accounts/nonce/<ADDRESS> 0x00
...
+
+# Edit the generated `genesis-state.json` as desired
+
+# ipdate the cometbft genesis file to include the sequencer genesis application state
```
noot, line 109:
```suggestion
# update the cometbft genesis file to include the sequencer genesis application state
```

crates/astria-sequencer/src/lib.rs
```diff
@@ -9,7 +9,7 @@ mod build_info;
 pub(crate) mod component;
 pub mod config;
 pub(crate) mod fee_asset_change;
-pub(crate) mod genesis;
+// pub(crate) mod genesis;
```
noot, line 12: remove

noot (commented):
can you update the astria-sequencer justfile `run-cometbft` command as well? otherwise, looks good!

crates/astria-sequencer/src/lib.rs
SuperFluffy: Done in  0dfad0b2

noot (approved):

PR #1232, opened 2024-07-01
test(bridge withdrawer): remove unit tests and add blackbox tests

### LINKED ISSUE #1227: Replace unit tests with black box tests in bridge-withdrawer
The tests in the bridge withdrawer are currently unit tests and are provided separately for the `Watcher` and `Submitter` components. This makes adding tests and functionality a bit cumbersome, since the tests are fairly fragile.

itamarreif:
## Summary
changes made in #1215 required adding a mock gRPC server. Moving to black box tests was the next step planned so this will replace the invalidated unit tests with valid black box tests.

## Background

## Changes
* Add black box tests for native and erc20 sequencer and ics20 withdrawals
* test helpers for the mock cometbft server
* test helpers for the mock sequencer gRPC server
* test helpers for managing anvil & ethereum node
* remove the "unit tests" in `submitter/tests.rs`
* some logs i found useful to add while putting this together

## Testing
* native withdrawal to the sequencer works
* native withdrawal with ics20 works
* erc20 to the sequencer works
* erc20 with ics20 works

## Related Issues

closes #1227


crates/astria-core/Cargo.toml
```diff
@@ -28,6 +28,7 @@ celestia-tendermint = { workspace = true }
 ed25519-consensus = { version = "2.1.0", default-features = false, features = [
   "std",
 ] }
+ethers = { workspace = true }
```
joroshiba, line 31: I don't think we really want this as a dependency in astria-core? doesn't appear to be used here at all?

itamarreif: good catch - this is a remnant of something that was fixed by janis

SuperFluffy (approved):
Looks good, great to have these finally in! Thank you for the lift!

As discussed offline, instead of leaving review comments I have made a few changes to the PR, notably:

1. added `ignore` attributes to the tests that require anvil -  these now run if `--include-ignored` is specified and the normal tests no longer fail.
2. clippy flagged the `ethereum::Watcher::startup` method. I replaced it by `startup(self) -> FullyInitialized`, which itself includes has a method `FullyInitialized::run`. That makes the purpose of `startup` clearer.
3. I have removed a lot of tracing events from long lived tasks (they make sense when looking at the command line, but will never be observed in grafana)
4. I have removed events that notify that a specific line of code is running. This information is redundant on a telemetry platform. If desired, https://docs.rs/tracing-subscriber/0.3.18/tracing_subscriber/fmt/struct.SubscriberBuilder.html#method.with_span_events can be used.
5. Instead of the events in 4., I have instrumented the functions.

PR #1233, opened 2024-07-02
release(sequencer): 0.14.1 patch release

joroshiba:
Includes changes to abci query for assets to enabled bridge withdrawer usage.


joroshiba:
> Does the sequencer chart's `appVersion` need updated too?

I'll update the chart in seperate PR, once image is cut

PR #1235, opened 2024-07-03
chore(cli)!: remove unmaintained rollup subcommand

SuperFluffy:
## Summary
Removes the `rollup` subcommand from `astria-cli`.

## Background
Deployment related logic has been moved from the monorepo's `astria-cli` to https://github.com/astriaorg/astria-cli-go. The current functionality is unmtained and likely broken.

## Changes
- Remove the `rollup` subcommand from `astria-cli` and all related modules.

## Breaking Changelist
Marked as a breaking change because functionality was removed. In practice this should not matter because nobody should be actively using the tool (no promise of maintenance, likely broken on HEAD).


joroshiba (approved):
should we as follow up make the `sequencer` commands work at the top level/consider merging this and sequencer utils?

SuperFluffy:
> should we as follow up make the `sequencer` commands work at the top level/consider merging this and sequencer utils?

merging with sequencer utils: yes

top level: no; I want to add a `bridge` subcommand.

PR #1236, opened 2024-07-03
feat(sequencer): integrate connect oracle and vote extension logic

noot:
## Summary
integrate skip's [connect](https://github.com/skip-mev/connect) (formerly named `slinky`) oracle service into astria.

at a high level, connect consists of an oracle sidecar program, which interacts with a validator node to provide price data, and various cosmos-sdk modules. 

since astria isn't cosmos, the relevant cosmos modules (x/marketmap and x/oracle) were essentially ported into the `connect` module of the sequencer app, which consists of two components, `market_map` and `oracle`.

the sequencer app was updated to talk to the sidecar during the `extend_vote` phase of consensus to gather prices to put into a vote extension.

the vote extension validation logic, proposal logic, and finalization logic were also ported from connect. 

## Background
we want oracle data to be available to rollups (and maybe on the sequencer itself too?)

## Changes
* import relevant protos from connect and create native rust types for them
* update the sequencer genesis state to contain `market_map` and `oracle` values
* implement the `market_map` component for the sequencer
  * update the sequencer grpc service to support the market_map grpc service, which is required by the oracle sidecar to retrieve the market map from the chain
* implement the `oracle` component for the sequencer and the query service for this component
* implement `extend_vote` logic which gets the prices from the sidecar and turns them into a vote extension
* implement `verify_vote_extension` logic which performs basic validation on a vote extension during the consensus phase
* implement `prepare_proposal` logic which gathers the vote extensions from the previous block, prunes any invalid votes, and performs additional validation to create a valid set of VEs 
* implement `process_proposal` logic which validates the set of VEs proposed, checking signatures and that the voting power is >2/3 amongst other things
* implement `finalize_block` logic which writes the updated prices to state based on the committed vote extensions. skip uses stake-weighted median to calculate the final price, but we don't have stake-weighting yet, so i just took the median.
* TODO: implement the connect cosmos `Msg` types as sequencer actions (follow-up)
  * https://github.com/skip-mev/slinky/blob/158cde8a4b774ac4eec5c6d1a2c16de6a8c6abb5/proto/slinky/oracle/v1/tx.proto
  * https://github.com/skip-mev/slinky/blob/158cde8a4b774ac4eec5c6d1a2c16de6a8c6abb5/proto/slinky/marketmap/v1/tx.proto
* TODO: update `SequencerBlockHeader` to contain the extended commit info + a proof for it (also follow-up)
* TODO: implement the `DeltaCurrencyPairStrategy` - right now only the `DefaultCurrencyPairStrategy` is implemented. can also do in follow-up

## Testing
TODO: run this on a multi-validator network also

clone connect: https://github.com/skip-mev/connect/tree/main

install go 1.22

build and run connect:

```sh
make build
go run scripts/genesis.go --use-coingecko=true --temp-file=markets.json
./build/connect --market-config-path markets.json --port 8081
```

checkout `noot/slinky` branch of astria

run sequencer app and `ASTRIA_SEQUENCER_NO_ORACLE=false` in `.env`:

```sh
rm -rf /tmp/astria_db
rm -rf ~/.cometbft
just run
just run-cometbft
```

should see a sequencer log like:

```sh
astria_sequencer::sequencer: oracle sidecar is reachable
```

should see a connect log like:

```sh
{"level":"info","ts":"2024-07-02T14:33:46.318-0400","caller":"marketmap/fetcher.go:147","msg":"successfully fetched market map data from module; checking if market map has changed","pid":727051,"process":"oracle","fetcher":"marketmap_api"}
```

then, when blocks are made, should see logs like the following for each block:

```
2024-07-05T02:49:21.254163Z DEBUG handle_request:handle_process_proposal: astria_sequencer::service::consensus: proposal processed height=28 time=2024-07-05T02:49:19.143352683Z tx_count=3 proposer=7BE21CDEB6FDCC9299A51F44C6B390EA990E88CD hash=7rmdhOsaW2a0NCZUwSE5yqt2AVR3cOPgGGb4Bb0kpRM= next_validators_hash=F6N7YDQZKfXQld95iV0AmQKNa8DiAxrDnTAcn323QSU=
2024-07-05T02:49:21.310218Z DEBUG handle_request:handle_extend_vote:App::extend_vote: astria_sequencer::app::vote_extension: got prices from oracle sidecar; transforming prices prices_count=118
2024-07-05T02:49:21.323262Z DEBUG handle_request:handle_extend_vote:App::extend_vote: astria_sequencer::app::vote_extension: transformed price for inclusion in vote extension currency_pair="BTC/USD" id=0 price=5683583007
2024-07-05T02:49:21.326070Z DEBUG handle_request:handle_extend_vote:App::extend_vote: astria_sequencer::app::vote_extension: transformed price for inclusion in vote extension currency_pair="ETH/USD" id=1 price=3055069469
2024-07-05T02:49:21.384266Z DEBUG handle_request:finalize_block:App::finalize_block: astria_sequencer::app::vote_extension: applied price from vote extension currency_pair="BTC/USD" price=5683583007 hash=EEB99D84EB1A5B66B4342654C12139CAAB7601547770E3E01866F805BD24A513 height=28 time=2024-07-05T02:49:19.143352683Z proposer=7BE21CDEB6FDCC9299A51F44C6B390EA990E88CD
2024-07-05T02:49:21.384553Z DEBUG handle_request:finalize_block:App::finalize_block: astria_sequencer::app::vote_extension: applied price from vote extension currency_pair="ETH/USD" price=3055069469 hash=EEB99D84EB1A5B66B4342654C12139CAAB7601547770E3E01866F805BD24A513 height=28 time=2024-07-05T02:49:19.143352683Z proposer=7BE21CDEB6FDCC9299A51F44C6B390EA990E88CD
```

## Breaking Changelist
* the PR is breaking because it adds a new proposer-added transaction at the start of the block. now there are 3 special "txs" expected when before there were only 2.
* the genesis was updated with a new optional field `connect`, however as this field is optional, it is non-breaking with existing networks. 
* if the `connect` genesis field is set, the sequencer state will change, as the genesis state changes and new values are stored in state. however this does not affect syncing existing networks, as the genesis of the existing network can be used as-is.
* additionally, vote extension participation is optional - if <=2/3 validators participate, blocks are still finalized, just no new oracle data will be published.


crates/astria-sequencer/src/app/mod.rs
```diff
@@ -301,9 +321,59 @@ impl App {
...
+        // adjust max block size to account for extended commit info
         let mut block_size_constraints = BlockSizeConstraints::new(
-            usize::try_from(prepare_proposal.max_tx_bytes)
-                .context("failed to convert max_tx_bytes to usize")?,
+            max_tx_bytes
+                .checked_sub(extended_commit_info_bytes.len())
+                .expect(
+                    "extended_commit_info_bytes is shorter than max_tx_bytes, as it was checked \
+                     above",
+                ),
```
Fraser999, line 376: We could avoid the `expect` by just using the result of `checked_sub`:
```rust
        // adjust max block size to account for extended commit info
        let adjusted_max_tx_bytes = max_tx_bytes
            .checked_sub(extended_commit_info_bytes.len())
            .unwrap_or_else(|| {
                // zero the commit info if it's too large to fit in the block
                // for liveness.
                warn!(
                    extended_commit_info_bytes_len = extended_commit_info_bytes.len(),
                    max_tx_bytes,
                    "extended commit info is too large to fit in block; not including in block"
                );
                extended_commit_info_bytes.clear();
                max_tx_bytes
            });
        let mut block_size_constraints = BlockSizeConstraints::new(adjusted_max_tx_bytes)
            .context("failed to create block size constraints")?;
```

crates/astria-sequencer/src/app/mod.rs
```diff
@@ -338,8 +408,11 @@ impl App {
         // included in the block
         let res = generate_rollup_datas_commitment(&signed_txs_included, deposits);
 
+        // inject the extended commit info into the start of the block's txs
+        let mut txs = vec![extended_commit_info_bytes.into()];
+        txs.extend(res.into_transactions(included_tx_bytes));
```
Fraser999, line 413: Not a big deal, but we can avoid the `mut` variable and potentially two allocations by doing:
```suggestion
        let txs = std::iter::once(extended_commit_info_bytes.into())
            .chain(res.into_transactions(included_tx_bytes))
            .collect();
```
Or maybe `GeneratedCommitments::into_transactions` should take the `extended_commit_info_bytes` as an arg and we can just add it to the front of the Vec inside there?

Fraser999 (commented):
Initial review for now, but making sense so far :)

crates/astria-sequencer/src/config.rs
```diff
@@ -31,6 +31,13 @@ pub struct Config {
     pub metrics_http_listener_addr: String,
     /// Writes a human readable format to stdout instead of JSON formatted OTEL trace data.
     pub pretty_print: bool,
+    /// If the oracle is enabled. If true, the `oracle_grpc_addr` must be set.
+    /// Should be true for validator nodes and false for non-validator nodes.
+    pub oracle_enabled: bool,
```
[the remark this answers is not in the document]
noot: we got rid of `enable_mint` :p it would be disabled for non-validator nodes, yeah.

> Also also... do you think we should be more specific about the oracle name in case we add further ones in the future? As in, these could be renamed to slinky_xxx?

not sure we would add multiple sidecars at this point, but naming `slinky` might be more specific

crates/astria-sequencer/src/grpc/slinky.rs
```diff
@@ -0,0 +1,247 @@
...
+            })?;
+        let chain_id = snapshot
+            .get_chain_id()
+            .await
+            .map_err(|e| Status::internal(format!("failed to get chain id from storage: {e}")))?;
+
+        Ok(Response::new(MarketMapResponse {
+            market_map: market_map.map(astria_core::slinky::market_map::v1::MarketMap::into_raw),
+            last_updated,
+            chain_id: chain_id.to_string(), // TODO: is this the right chain id?
```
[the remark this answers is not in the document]
noot: will ask skip!

crates/astria-sequencer/src/grpc/slinky.rs
```diff
@@ -0,0 +1,247 @@
...
+            .map_err(|e| Status::internal(format!("failed to get state from storage: {e}")))?
+        else {
+            return Err(Status::not_found("currency pair state not found"));
+        };
+
+        Ok(Response::new(GetPriceResponse {
+            price: Some(state.price.into_raw()),
+            nonce: state.nonce,
+            id: state.id,
+            decimals: 0, // TODO: get this from the marketmap
```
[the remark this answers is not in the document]
noot: implemented

crates/astria-sequencer/src/grpc/slinky.rs
```diff
@@ -0,0 +1,247 @@
...
+
+        let snapshot = self.storage.latest_snapshot();
+        let mut prices = Vec::new();
+        for currency_pair in currency_pairs {
+            let Some(state) = snapshot
+                .get_currency_pair_state(&currency_pair)
+                .await
+                .map_err(|e| Status::internal(format!("failed to get state from storage: {e}")))?
+            else {
+                return Err(Status::not_found("currency pair state not found"));
```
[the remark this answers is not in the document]
noot: not sure what's better behaviour here, i suppose a request with a pair that doesn't exist would be invalid, so it's ok to error?

crates/astria-sequencer/src/grpc/slinky.rs
```diff
@@ -0,0 +1,247 @@
...
+            };
+            prices.push(GetPriceResponse {
+                price: Some(state.price.into_raw()),
+                nonce: state.nonce,
+                id: state.id,
+                decimals: 0, // TODO: get this from the marketmap
```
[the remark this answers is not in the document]
noot: implemented

crates/astria-sequencer/src/sequencer.rs
```diff
@@ -100,10 +116,39 @@ impl Sequencer {
                 .context("failed to initialize global address base prefix")?;
         }
 
+        let oracle_client = if config.oracle_enabled {
+            let uri: Uri = config
+                .oracle_grpc_addr
+                .parse()
+                .context("failed parsing oracle grpc address as Uri")?;
+            let endpoint = Endpoint::from(uri.clone());
```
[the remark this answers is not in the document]
noot: nice!

crates/astria-sequencer/src/app/mod.rs
noot: nice, that's a lot better!

crates/astria-sequencer/src/slinky/oracle/state_ext.rs
```diff
@@ -0,0 +1,289 @@
...
+                .context("failed to strip prefix from currency pair state key")?
+                .parse::<CurrencyPair>()
+                .context("failed to parse storage key suffix as currency pair")?;
+            currency_pairs.push(currency_pair);
+        }
+        Ok(currency_pairs)
+    }
+
+    #[instrument(skip(self))]
+    async fn get_next_currency_pair_id(&self) -> Result<u64> {
```
Fraser999, line 183: This method doesn't seem to be used anywhere.

crates/astria-sequencer/src/slinky/oracle/state_ext.rs
```diff
@@ -0,0 +1,289 @@
...
+        else {
+            return Ok(0);
+        };
+        let Id(next_currency_pair_id) =
+            Id::try_from_slice(&bytes).context("invalid next currency pair id bytes")?;
+        Ok(next_currency_pair_id)
+    }
+
+    #[instrument(skip(self))]
+    async fn get_price_for_currency_pair(
```
Fraser999, line 197: This one also seems to be unused.

crates/astria-sequencer/src/slinky/oracle/state_ext.rs
```diff
@@ -0,0 +1,289 @@
...
+        Ok(())
+    }
+
+    #[instrument(skip(self))]
+    fn put_currency_pair(&mut self, id: u64, currency_pair: CurrencyPair) -> Result<()> {
+        let bytes =
+            serde_json::to_vec(&currency_pair).context("failed to serialize currency pair")?;
+        self.put_raw(id_to_currency_pair_storage_key(id), bytes);
+        Ok(())
+    }
```
Fraser999, line 233: These two methods might be better changed to free functions since they're not used outside of the trait?  Something like:
```rust
fn put_currency_pair_id<W: StateWrite + ?Sized>(
    state_writer: &mut W,
    currency_pair: &CurrencyPair,
    id: u64,
) -> Result<()> {
    ...
}
```

crates/astria-sequencer/src/slinky/oracle/state_ext.rs
```diff
@@ -0,0 +1,289 @@
...
+pub(crate) trait StateWriteExt: StateWrite {
+    #[instrument(skip(self))]
+    fn put_currency_pair_id(&mut self, currency_pair: &CurrencyPair, id: u64) -> Result<()> {
+        let bytes = borsh::to_vec(&Id(id)).context("failed to serialize currency pair id")?;
+        self.put_raw(currency_pair_to_id_storage_key(currency_pair), bytes);
+        Ok(())
+    }
+
+    #[instrument(skip(self))]
+    fn put_currency_pair(&mut self, id: u64, currency_pair: CurrencyPair) -> Result<()> {
```
Fraser999, line 228: We could take `&CurrencyPair` to avoid cloning it when calling this method.

crates/astria-sequencer/src/slinky/oracle/state_ext.rs
```diff
@@ -0,0 +1,289 @@
...
+    #[instrument(skip(self))]
+    fn put_num_removed_currency_pairs(&mut self, num_removed_currency_pairs: u64) -> Result<()> {
+        let bytes = borsh::to_vec(&Count(num_removed_currency_pairs))
+            .context("failed to serialize number of removed currency pairs")?;
+        self.put_raw(NUM_REMOVED_CURRENCY_PAIRS_KEY.to_string(), bytes);
+        Ok(())
+    }
+
+    #[instrument(skip(self))]
+    fn put_currency_pair_state_and_price(
```
Fraser999, line 252: I'm not too clear on why we have separate entries for the state (which contains the price) and the price itself.

Seems like we put an initial value for state during `OracleComponent::init_chain`, and that value is never changed.  However, we have `get_currency_pair_state` which is called by `OracleService::get_price` and `OracleService::get_prices`.  So those two methods will always report the same initial price, nonce and id.

Meanwhile, the price value is updated via `apply_prices_from_vote_extensions`, but that price is never read.

crates/astria-sequencer/src/slinky/marketmap/state_ext.rs
```diff
@@ -0,0 +1,105 @@
...
+    #[instrument(skip(self))]
+    fn put_market_map(&mut self, market_map: MarketMap) -> Result<()> {
+        let bytes = serde_json::to_vec(&market_map).context("failed to serialize market map")?;
+        self.put_raw(MARKET_MAP_KEY.to_string(), bytes);
+        Ok(())
+    }
+
+    #[instrument(skip(self))]
+    fn put_market_map_last_updated_height(&mut self, height: u64) -> Result<()> {
+        // TODO: since this is an optimization, should it be in the nonconsensus store?
```
Fraser999, line 89: Can this comment be removed?

crates/astria-sequencer/src/slinky/marketmap/state_ext.rs
```diff
@@ -0,0 +1,105 @@
...
+pub(crate) trait StateWriteExt: StateWrite {
+    #[instrument(skip(self))]
+    fn put_market_map(&mut self, market_map: MarketMap) -> Result<()> {
+        let bytes = serde_json::to_vec(&market_map).context("failed to serialize market map")?;
+        self.put_raw(MARKET_MAP_KEY.to_string(), bytes);
+        Ok(())
+    }
+
+    #[instrument(skip(self))]
+    fn put_market_map_last_updated_height(&mut self, height: u64) -> Result<()> {
```
Fraser999, line 88: This method is never called.

crates/astria-sequencer/src/app/vote_extension.rs
Fraser999, line 1: I think we should have unit tests for much of the logic in this file, since a lot of it is not trivial IMO.

Fraser999 (changes requested):
Nice job here - lots of headway made in the PR.  I'm blocking for now as I think there are a couple of bugs though.

.gitignore
```diff
@@ -15,3 +15,4 @@ target/
 charts/*/charts
 
 crates/astria-bridge-withdrawer/src/withdrawer/ethereum/generated/
+crates/astria-sequencer/app-genesis-state.json
```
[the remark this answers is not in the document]
SuperFluffy: We shouldn't add artifacts like this to `.gitignore`

crates/astria-sequencer-utils/src/genesis_example.rs
```diff
@@ -46,7 +59,56 @@ fn charlie() -> Address {
         .unwrap()
 }
 
+// allow clippy here because this is an example genesis state, which is long.
+#[allow(clippy::too_many_lines)]
```
SuperFluffy, line 63: For readability, split this up into separate constructors.

crates/astria-sequencer/Cargo.toml
```diff
@@ -25,6 +29,8 @@ telemetry = { package = "astria-telemetry", path = "../astria-telemetry", featur
 
 anyhow = "1"
 borsh = { version = "1", features = ["derive"] }
+# `is_sorted_by` is available in rust 1.81.0, but we haven't updated our msrv yet
+is_sorted = "0.1.1"
```
SuperFluffy, line 33: it looks like we don't use this?

charts/sequencer/files/cometbft/config/genesis.json
```diff
@@ -41,13 +41,32 @@
...
+    ],
+    "market_map": {
+      "market_map": {
+        "markets": {}
+      },
+      "last_updated": 0,
+      "params": {
+        "market_authorities": [],
+        "admin": {
+          "bech32m": "astria1rsxyjrcm255ds9euthjx6yc3vrjt9sxrm9cfgm"
```
quasystaty1, line 53: Can maybe use templated value for admin address, also maybe move to `.Values.global.dev` else for breaking changes.
```suggestion
          "bech32m": {{ include "sequencer.address" .Values.genesis.authoritySudoAddress }}
```

charts/sequencer/templates/configmaps.yaml
```diff
@@ -73,6 +73,9 @@ data:
   OTEL_EXPORTER_OTLP_HEADERS: "{{ .Values.sequencer.otel.otlpHeaders }}"
   OTEL_EXPORTER_OTLP_TRACE_HEADERS: "{{ .Values.sequencer.otel.traceHeaders }}"
   OTEL_SERVICE_NAME: "{{ tpl .Values.sequencer.otel.serviceName . }}"
+  ASTRIA_SEQUENCER_ORACLE_ENABLED: "false"
```
quasystaty1, line 76: Should move under else statement to mark breaking changes

crates/astria-core/src/sequencerblock/v1alpha1/block.rs
```diff
@@ -713,6 +721,12 @@ impl SequencerBlock {
         let data_hash = tree.root();
 
         let mut data_list = data.into_iter();
+
+        // TODO: this needs to go into the block header
```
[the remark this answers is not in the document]
noot: yes, can make a ticket for this after the PR is in, would be part of propagating oracle data to rollups

crates/astria-sequencer/src/app/vote_extension.rs
```diff
@@ -0,0 +1,571 @@
...
+                    error = AsRef::<dyn std::error::Error>::as_ref(&e),
+                    validator = crate::address::base_prefixed(vote.validator.address).to_string(),
+                    "failed to verify vote extension; pruning from proposal"
+                );
+                vote.sig_info = Flag(tendermint::block::BlockIdFlag::Absent);
+                vote.extension_signature = None;
+                vote.vote_extension = vec![].into();
+            }
+        }
+        validate_vote_extensions(state, height, &extended_commit_info)
```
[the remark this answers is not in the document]
noot: we don't need to call it here, as if we're the proposer, we're the ones generating the extended commit for the block so we don't need to verify it. only other validators need to validate the extended commit proposed against the actual last commit.

crates/astria-sequencer/src/app/vote_extension.rs
```diff
@@ -0,0 +1,571 @@
...
+                a.validator
+                    .power
+                    .partial_cmp(&b.validator.power)
+                    .map(std::cmp::Ordering::reverse)
+            }
+        }),
+        "extended commit votes are not sorted by voting power",
+    );
+
+    for (i, vote) in extended_commit_info.votes.iter().enumerate() {
```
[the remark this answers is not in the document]
noot: `last_commit` shouldn't contain duplicate validator entries, so not sure that's necessary. will confirm my `last_commit` assumption

crates/astria-sequencer/src/app/vote_extension.rs
```diff
@@ -0,0 +1,571 @@
...
+        if vote.sig_info != Flag(tendermint::block::BlockIdFlag::Commit) {
+            continue;
+        }
+
+        submitted_voting_power =
+            submitted_voting_power.saturating_add(vote.validator.power.value());
+
+        let verification_key = validator_set
+            .get(vote.validator.address)
+            .context("validator not found")?
```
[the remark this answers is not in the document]
noot: this should error imo, as in the process_proposal phase it means the proposer proposed an invalid commit. in the prepare_proposal phase, this shouldn't error as we got the extended commit from our cometbft node, so that would mean there's a cometbft bug

crates/astria-sequencer/src/app/vote_extension.rs
```diff
@@ -0,0 +1,571 @@
...
+        state: &S,
+        height: u64,
+        last_commit: &CommitInfo,
+        extended_commit_info: &ExtendedCommitInfo,
+    ) -> anyhow::Result<()> {
+        if height == 1 {
+            // we're processing block 1, so nothing to validate (no last commit yet)
+            return Ok(());
+        }
+
```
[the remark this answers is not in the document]
noot: there isn't any commit/vote extensions for block 0, so there isn't any extended commit info proposed during block 1.  `last_commit` and `extended_commit_info` would both be empty, so the following functions would be no-ops

crates/astria-sequencer/src/app/mod.rs
```diff
@@ -302,11 +322,56 @@ impl App {
...
+                );
+                tendermint::abci::types::ExtendedCommitInfo {
+                    round,
+                    votes: Vec::new(),
+                }
+            }
+        };
+
+        let mut extended_commit_info_bytes =
+            ExtendedCommitInfo::from(extended_commit_info).encode_to_vec();
```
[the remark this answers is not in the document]
noot: yeah we could, could be an optimization follow-up!

crates/astria-sequencer/src/slinky/oracle/currency_pair_strategy.rs
```diff
@@ -0,0 +1,65 @@
...
+    pub(crate) async fn from_id<S: StateReadExt>(
+        state: &S,
+        id: u64,
+    ) -> anyhow::Result<Option<CurrencyPair>> {
+        state.get_currency_pair(id).await
+    }
+
+    pub(crate) fn get_encoded_price<S: StateReadExt>(
+        _state: &S,
+        _: &CurrencyPair,
```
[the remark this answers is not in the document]
noot: they're used for the delta currency pair strategy which is used by default in slinky but not yet implemented in this PR: https://github.com/skip-mev/connect/blob/793b2e874d6e720bd288e82e782502e41cf06f8c/abci/strategies/currencypair/delta.go however i'll remove the parameters for now and we can add them back later

crates/astria-sequencer/src/slinky/oracle/state_ext.rs
```diff
@@ -0,0 +1,289 @@
...
+};
+use futures::StreamExt as _;
+use tracing::instrument;
+
+const CURRENCY_PAIR_TO_ID_PREFIX: &str = "oraclecpid";
+const ID_TO_CURRENCY_PAIR_PREFIX: &str = "oracleidcp";
+const CURRENCY_PAIR_STATE_PREFIX: &str = "oraclecpstate";
+const CURRENCY_PAIR_PRICE_PREFIX: &str = "oraclecpprice";
+
+// TODO: should these values be in nonverifiable storage?
```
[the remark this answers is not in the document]
noot: slinky stores everything in the consensus/verified store, so going to remove this comment

crates/astria-sequencer/src/slinky/oracle/state_ext.rs
noot: not currently but will be used when the slinky actions are implemented, should i remove for now?

crates/astria-sequencer/src/slinky/oracle/state_ext.rs
noot: same as above

crates/astria-sequencer/src/slinky/oracle/state_ext.rs
noot: yep, it looks like skip has updated this to just store the state only: https://github.com/skip-mev/connect/blob/main/x/oracle/keeper/keeper.go#L196 will update this respectively

crates/astria-sequencer/src/slinky/marketmap/state_ext.rs
noot: this should be in the consensus store, will update

crates/astria-sequencer/src/slinky/marketmap/state_ext.rs
noot: it'll be used when actions are implemented - should i remove for now?

crates/astria-sequencer/src/app/vote_extension.rs
noot: agreed, will work on tests

crates/astria-sequencer/src/config.rs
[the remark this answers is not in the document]
SuperFluffy: Let's make this consistent with our other services and name this
```suggestion
    pub no_oracle: bool,
```

SuperFluffy (commented):
Leaving a few prelim comments for very high level things.

proto/vendored/cosmos_sdk/cosmos_proto/cosmos.proto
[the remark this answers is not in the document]
noot: removed this file

crates/astria-core/src/slinky/abci.rs
```diff
@@ -0,0 +1,32 @@
+pub mod v1 {
+    use std::collections::HashMap;
+
+    use crate::generated::slinky::abci::v1 as raw;
+
+    #[derive(Debug, Clone, PartialEq, Eq)]
+    pub struct OracleVoteExtension {
+        pub prices: HashMap<u64, Vec<u8>>,
```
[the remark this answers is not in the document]
noot: is there a way to change the proto compiler to use a `BTreeMap` instead of a `HashMap` when generating the proto `OracleVoteExtension`?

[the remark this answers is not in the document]
SuperFluffy: Yes! https://docs.rs/tonic-build/latest/tonic_build/struct.Builder.html#method.btree_map for tonic, and there is something similar for prost.

crates/astria-core/src/slinky/types.rs
```diff
@@ -0,0 +1,81 @@
...
+    #[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
+    #[derive(Debug, Clone, PartialEq, Eq, Hash)]
+    pub struct CurrencyPair {
+        base: String,
+        quote: String,
+    }
+
+    impl CurrencyPair {
+        #[must_use]
+        pub fn new(base: String, quote: String) -> Self {
```
SuperFluffy, line 13: I'd remove this and rather lean on the `CurrencyPair::from_raw(raw::CurrencyPair)` being the sole entrypoint for this (so one need not remember that the first arg is `base` and the second is `quote`). Since the `serde`  derive above will need From/TryFrom impls anyway you can use `from`/`into` to make it more ergonomic.

crates/astria-sequencer-utils/src/genesis_parser.rs
```diff
@@ -80,6 +80,14 @@ fn insert_app_state_and_chain_id(dst: &mut Value, app_state: &Value, chain_id: S
     let Value::Object(dst) = dst else {
         panic!("dst is not an object");
     };
+    dst.get_mut("consensus_params")
+        .expect("consensus_params field exists in cometbft genesis")
+        .as_object_mut()
+        .expect("consensus_params field is an object")
```
SuperFluffy, line 86:
```suggestion
        .expect("consensus_params field is a JSON object")
```

crates/astria-sequencer/src/slinky/oracle/state_ext.rs
SuperFluffy: IMO yes so that once it's used it can be understood with the actions.

crates/astria-sequencer/Cargo.toml
```diff
@@ -41,8 +47,10 @@ divan = { workspace = true, optional = true }
 futures = { workspace = true }
 hex = { workspace = true, features = ["serde"] }
 ibc-types = { workspace = true, features = ["with_serde"] }
+indexmap = { workspace = true }
 penumbra-ibc = { workspace = true, features = ["component", "rpc"] }
 metrics = { workspace = true }
+pbjson-types = { workspace = true }
```
SuperFluffy, line 53: You likely need this for timestamps?  Since this is integral to types inside `astria-core`, we can reexport it there so you don't need to declare the dependency here.

crates/astria-sequencer/src/slinky/oracle/component.rs
```diff
@@ -0,0 +1,71 @@
...
+        for currency_pair in &app_state.slinky().oracle().currency_pair_genesis {
+            let currency_pair_state = CurrencyPairState {
+                id: currency_pair.id(),
+                nonce: currency_pair.nonce(),
+                price: currency_pair.currency_pair_price().clone(),
+            };
+            state
+                .put_currency_pair_state(currency_pair.currency_pair(), currency_pair_state)
+                .context("failed to put currency pair")?;
+            tracing::info!(
```
SuperFluffy, line 38: Remove this here. This information will be available form the span created by `put_currency_pair_state`

SuperFluffy (commented):
Still working myself through the sequencer changes, but already leaving a bunch of comments here and there.

Two high level observations that apply to several definitions:

1. Please add the source and sha-like for all vendored slinky protos.
2. Turn the maps into btreemaps (so that their order is stable; I found it extremely useful to have that stability even though it has a performance impact).

crates/astria-core/src/slinky/market_map.rs
```diff
@@ -0,0 +1,435 @@
...
+        ///
+        /// # Errors
+        ///
+        /// - if any of the `market_authorities` addresses are invalid
+        /// - if the `admin` address is invalid
+        pub fn try_from_raw(raw: raw::Params) -> Result<Self, ParamsError> {
+            let market_authorities = raw
+                .market_authorities
+                .into_iter()
+                .map(|s| Address::from_str(&s))
```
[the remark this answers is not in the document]
noot: yeah, these are astria accounts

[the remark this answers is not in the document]
noot: they're permissioned accounts which can update the market parameters

crates/astria-core/src/slinky/types.rs
```diff
@@ -0,0 +1,81 @@
...
+        }
+    }
+
+    impl std::fmt::Display for CurrencyPair {
+        fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+            write!(f, "{}/{}", self.base, self.quote)
+        }
+    }
+
+    impl std::str::FromStr for CurrencyPair {
```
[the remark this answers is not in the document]
noot: this is used in `transform_oracle_service_prices`: 
```
        let currency_pair = currency_pair_id
            .parse()
            .context("failed to parse currency pair")?;
```

charts/sequencer/values.yaml
```diff
@@ -227,6 +228,7 @@ ports:
   cometbftRpc: 26657
   cometbftMetrics: 26660
   sequencerABCI: 26658
+  # TODO: change to 9090 if running with oracle sidecar
```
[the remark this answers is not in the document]
noot: yeah, the sidecar can use `--port`, will update this comment

crates/astria-core/src/slinky/types.rs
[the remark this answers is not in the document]
SuperFluffy: Oh boy. Why would they not type it out and send over their own protos but let us parse this by hand? :(

crates/astria-core/src/slinky/types.rs
```diff
@@ -0,0 +1,94 @@
...
+        type Err = CurrencyPairParseError;
+
+        fn from_str(s: &str) -> Result<Self, Self::Err> {
+            s.split_once('/')
```
SuperFluffy, line 65: Can you please use a regex for this instead? Something like `([a-zA-Z]+)/([a-zA-Z])` two define two groups - or just `alnum` if it can contain digits too.

Does slinky define what can be included in these pairs anywhere?

crates/astria-core/src/slinky/types.rs
```diff
@@ -0,0 +1,94 @@
...
+        }
+    }
+
+    impl From<CurrencyPair> for raw::CurrencyPair {
+        fn from(currency_pair: CurrencyPair) -> Self {
+            currency_pair.into_raw()
+        }
+    }
+
+    impl std::fmt::Display for CurrencyPair {
```
SuperFluffy, line 55: Please provide unit tests for to go `to_string <> parse <> to_string`

crates/astria-sequencer/src/app/mod.rs
```diff
@@ -794,15 +904,34 @@ impl App {
         }
 
         ensure!(
-            finalize_block.txs.len() >= 2,
-            "block must contain at least two transactions: the rollup transactions commitment and
+            finalize_block.txs.len() >= 3,
+            "block must contain at least three transactions: the extended commit info, the rollup \
+             transactions commitment and
```
SuperFluffy, line 909: join this and the next line :)

SuperFluffy (commented):
Some more flyby comments.

crates/astria-core/src/slinky/types.rs
noot: the pairs themselves aren't defined anywhere afaik, but they should be two alphabetical strings with a `/` between.

crates/astria-sequencer/src/app/vote_extension.rs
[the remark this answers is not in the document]
Fraser999: I was meaning we should call it after the `if height == 1` code block.

crates/astria-sequencer/src/slinky/marketmap/state_ext.rs
Fraser999: I'd vote to remove, but not a big deal - certainly not a blocker :)

proto/protocolapis/astria/protocol/genesis/v1alpha1/types.proto
```diff
@@ -15,6 +17,7 @@ message GenesisAppState {
   IbcParameters ibc_parameters = 8;
   repeated string allowed_fee_assets = 9;
   Fees fees = 10;
+  SlinkyGenesis slinky = 11;
```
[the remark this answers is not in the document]
Fraser999: I was thinking using "slinky" everywhere (at least in all user-facing places like configs and protos) rather than the more generic "oracle" might be more future-proof in case we add more oracles down the line?  Or maybe we could use "slinky_oracle" or something to help readers unaware of slinky?

Fraser999 (commented):
No blockers remaining as far as I'm concerned, but still a few unresolved comments.

crates/astria-sequencer/src/app/mod.rs
```diff
@@ -793,15 +903,33 @@ impl App {
...
         ensure!(
-            finalize_block.txs.len() >= 2,
-            "block must contain at least two transactions: the rollup transactions commitment and
-             rollup IDs commitment"
+            finalize_block.txs.len() >= 3,
+            "block must contain at least three transactions: the extended commit info, the rollup \
+             transactions commitment and rollup IDs commitment"
         );
 
+        let extended_commit_info_bytes = &finalize_block.txs[0];
```
SuperFluffy, line 911: This is correct but prone to fail when refactored; this makes it a bit clearer for the reader what to look for:
```suggestion
        let extended_commit_info_bytes = finalize_block.get(0).expect("asserted length above");
```

crates/astria-sequencer/src/app/mod.rs
SuperFluffy, line 1: High level comment on `finalize_block` in this file: it's getting quite unwieldly and turning into an overlong method. We should refactor it so it becomes easier to read.

crates/astria-sequencer/src/app/vote_extension.rs
```diff
@@ -0,0 +1,649 @@
...
+            Err(e) => {
+                tracing::error!(error = %e, "failed to verify vote extension");
+                abci::response::VerifyVoteExtension::Reject
+            }
+        }
+    }
+}
+
+// see https://github.com/skip-mev/slinky/blob/5b07f91d6c0110e617efda3f298f147a31da0f25/abci/ve/utils.go#L24
+async fn verify_vote_extension<S: StateReadExt>(
```
SuperFluffy, line 121: OK, looking at how this is used: this function performs conflates normal expected operations (failing to decode bytes that are sent from external entities, ensuring that passed in data is correct), with potentially(?) catastrophic issues: getting information form state that should be there but isn't.

crates/astria-sequencer/src/app/vote_extension.rs
```diff
@@ -0,0 +1,649 @@
...
+            prices.len() <= MAXIMUM_PRICE_BYTE_LEN,
+            "encoded price length exceeded {MAXIMUM_PRICE_BYTE_LEN}"
+        );
+    }
+
+    Ok(())
+}
+
+// see https://github.com/skip-mev/slinky/blob/158cde8a4b774ac4eec5c6d1a2c16de6a8c6abb5/abci/ve/vote_extension.go#L290
+async fn transform_oracle_service_prices<S: StateReadExt>(
```
SuperFluffy, line 149: The body of this function should be a stream rather than a loop. Create a future for each (id, price) pair and collect into the `IndexMap` using `TryStreamExt::try_collect` from the futures crate.

crates/astria-sequencer/src/app/vote_extension.rs
```diff
@@ -0,0 +1,649 @@
...
+        state: &S,
+        height: u64,
+        mut extended_commit_info: ExtendedCommitInfo,
+    ) -> anyhow::Result<ExtendedCommitInfo> {
+        if height == 1 {
+            // we're proposing block 1, so nothing to validate
+            return Ok(extended_commit_info);
+        }
+
+        for vote in &mut extended_commit_info.votes {
```
SuperFluffy, line 199: This should be a stream too. By making this a sequential loop you incur the cost of every single await, which quickly sums up.

crates/astria-sequencer/src/app/vote_extension.rs
```diff
@@ -0,0 +1,649 @@
...
+    Ok(OracleVoteExtension {
+        prices: strategy_prices,
+    })
+}
+
+pub(crate) struct ProposalHandler;
+
+impl ProposalHandler {
+    // called during prepare_proposal
+    pub(crate) async fn prune_and_validate_extended_commit_info<S: StateReadExt>(
```
SuperFluffy, line 189: Two notes:

1. Can we give this a clearer naming? I have read the comments below and understand that this is called when the sequencer node acts as the proposer. But neither the comment nor the name make this very clear.
2. Please add instrumnetation (including the height).

crates/astria-sequencer/src/app/vote_extension.rs
```diff
@@ -0,0 +1,649 @@
...
+        state: &S,
+        height: u64,
+        mut extended_commit_info: ExtendedCommitInfo,
+    ) -> anyhow::Result<ExtendedCommitInfo> {
+        if height == 1 {
+            // we're proposing block 1, so nothing to validate
```
SuperFluffy, line 195: emit a log for this.

crates/astria-sequencer/src/app/vote_extension.rs
```diff
@@ -0,0 +1,649 @@
...
+
+    // called during process_proposal
+    pub(crate) async fn validate_extended_commit_info<S: StateReadExt>(
+        state: &S,
+        height: u64,
+        last_commit: &CommitInfo,
+        extended_commit_info: &ExtendedCommitInfo,
+    ) -> anyhow::Result<()> {
+        if height == 1 {
+            // we're processing block 1, so nothing to validate (no last commit yet)
```
SuperFluffy, line 230: add an event for this information.

crates/astria-sequencer/src/app/vote_extension.rs
```diff
@@ -0,0 +1,649 @@
...
+
+pub(crate) struct ProposalHandler;
+
+impl ProposalHandler {
+    // called during prepare_proposal
+    pub(crate) async fn prune_and_validate_extended_commit_info<S: StateReadExt>(
+        state: &S,
+        height: u64,
+        mut extended_commit_info: ExtendedCommitInfo,
+    ) -> anyhow::Result<ExtendedCommitInfo> {
```
SuperFluffy, line 193: To make the types clearer: this should return a `ValidatedExtendedCommitInfo`. It can just be a thin wrapper `struct ValidatedExtendedCommitInfo(ExtendedCommitInfo).`

crates/astria-sequencer/src/app/vote_extension.rs
```diff
@@ -0,0 +1,649 @@
...
+            .await
+            .context("failed to construct validator address with base prefix")?;
+
+        total_voting_power = total_voting_power.saturating_add(vote.validator.power.value());
+
+        if vote.sig_info == Flag(tendermint::block::BlockIdFlag::Commit)
+            && vote.extension_signature.is_none()
+        {
+            anyhow::bail!("vote extension signature is missing for validator {address}",);
+        }
```
SuperFluffy, line 280: this can be an `ensure!` macro

crates/astria-sequencer/src/app/vote_extension.rs
```diff
@@ -0,0 +1,649 @@
...
+
+        if vote.sig_info != Flag(tendermint::block::BlockIdFlag::Commit)
+            && !vote.vote_extension.is_empty()
+        {
+            anyhow::bail!("non-commit vote extension present for validator {address}",);
+        }
```
SuperFluffy, line 286: Also `ensure!`

crates/astria-sequencer/src/app/vote_extension.rs
```diff
@@ -0,0 +1,649 @@
...
+
+        if vote.sig_info != Flag(tendermint::block::BlockIdFlag::Commit)
+            && vote.extension_signature.is_some()
+        {
+            anyhow::bail!("non-commit extension signature present for validator {address}",);
+        }
```
SuperFluffy, line 292: `ensure!`

crates/astria-sequencer/src/service/consensus.rs
```diff
@@ -126,6 +137,10 @@ impl Consensus {
             bail!("database already initialized");
         }
 
+        println!(
+            "{}",
```
SuperFluffy, line 141: overlooked debugging code?

crates/astria-sequencer/src/slinky/marketmap/component.rs
```diff
@@ -0,0 +1,53 @@
...
+#[derive(Default)]
+pub(crate) struct MarketMapComponent;
+
+#[async_trait::async_trait]
+impl Component for MarketMapComponent {
+    type AppState = GenesisAppState;
+
+    #[instrument(name = "MarketMapComponent::init_chain", skip(state))]
+    async fn init_chain<S: StateWriteExt>(mut state: S, app_state: &Self::AppState) -> Result<()> {
+        // TODO: put market map authorites and admin in state;
```
SuperFluffy, line 26: Is there an issue to track this?

crates/astria-sequencer/src/slinky/marketmap/component.rs
```diff
@@ -0,0 +1,53 @@
...
+
+#[derive(Default)]
+pub(crate) struct MarketMapComponent;
+
+#[async_trait::async_trait]
+impl Component for MarketMapComponent {
+    type AppState = GenesisAppState;
+
+    #[instrument(name = "MarketMapComponent::init_chain", skip(state))]
+    async fn init_chain<S: StateWriteExt>(mut state: S, app_state: &Self::AppState) -> Result<()> {
```
SuperFluffy, line 25: Make this `S: StateWrite` and only import that `StateWriteExt as _` at the head of this file.

crates/astria-sequencer/src/slinky/marketmap/component.rs
```diff
@@ -0,0 +1,53 @@
...
+            .put_market_map(app_state.slinky().market_map().market_map.clone())
+            .context("failed to put market map")?;
+        state
+            .put_params(app_state.slinky().market_map().params.clone())
+            .context("failed to put params")?;
+        Ok(())
+    }
+
+    #[instrument(name = "MarketMapComponent::begin_block", skip(_state))]
+    async fn begin_block<S: StateWriteExt + 'static>(
```
SuperFluffy, line 39: `S: StateWrite`

crates/astria-sequencer/src/slinky/marketmap/component.rs
```diff
@@ -0,0 +1,53 @@
...
+        _state: &mut Arc<S>,
+        _begin_block: &BeginBlock,
+    ) -> Result<()> {
+        Ok(())
+    }
+
+    #[instrument(name = "MarketMapComponent::end_block", skip(_state))]
+    async fn end_block<S: StateWriteExt + 'static>(
```
SuperFluffy, line 47: `S: StateWrite`

crates/astria-sequencer/src/slinky/oracle/state_ext.rs
```diff
@@ -0,0 +1,278 @@
...
+                let currency_pair = serde_json::from_slice(&bytes)
+                    .context("failed to deserialize currency pair")?;
+                Ok(Some(currency_pair))
+            }
+            None => Ok(None),
+        }
+    }
+
+    #[instrument(skip_all)]
+    async fn get_currency_pair_mapping(&self) -> Result<HashMap<u64, CurrencyPair>> {
```
SuperFluffy, line 87: Instead of returning an object, implement this as a stream returning Items of type to make this more flexible `Result<CurrencyPair>` (or whatever). See https://github.com/astriaorg/astria/pull/1453 for how this can be done (this was ultimately merged into https://github.com/astriaorg/astria/pull/1408)

crates/astria-sequencer/src/slinky/oracle/state_ext.rs
```diff
@@ -0,0 +1,278 @@
...
+        let prefix = format!("{CURRENCY_PAIR_TO_ID_PREFIX}/");
+        let mut currency_pairs = HashMap::new();
+
+        let mut stream = std::pin::pin!(self.prefix_keys(&prefix));
+        while let Some(Ok(key)) = stream.next().await {
```
SuperFluffy, line 92: This is potentially a source of latency: use `futures::TryStreamExt::try_collect` on a `futures::StreamExt::buffered` adaptor to do this concurrently.

crates/astria-sequencer/src/slinky/oracle/state_ext.rs
```diff
@@ -0,0 +1,278 @@
...
+                let currency_pair_state = serde_json::from_slice(&bytes)
+                    .context("failed to deserialize currency pair state")?;
+                Ok(Some(currency_pair_state))
+            }
+            None => Ok(None),
+        }
+    }
+
+    #[instrument(skip_all)]
+    async fn get_all_currency_pairs(&self) -> Result<Vec<CurrencyPair>> {
```
SuperFluffy, line 160: As above, implement this as a stream and let the call site decide what to do with this. See https://github.com/astriaorg/astria/pull/1453 for how to do that.

proto/protocolapis/astria/protocol/genesis/v1alpha1/types.proto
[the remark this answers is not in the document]
SuperFluffy: Yeah, I have changed my mind. We should probably call this slinky because there is nothing generic about this. (Or should we use their new `*-connect` name already?)

SuperFluffy (commented):
Many more comments.

crates/astria-sequencer/src/slinky/oracle/currency_pair_strategy.rs
```diff
@@ -0,0 +1,60 @@
...
+    ) -> anyhow::Result<u128> {
+        ensure!(encoded_price.len() == 16, "invalid encoded price length");
+        let mut bytes = [0; 16];
+        bytes.copy_from_slice(encoded_price);
+        Ok(u128::from_be_bytes(bytes))
+    }
+
+    pub(crate) async fn get_max_num_currency_pairs<S: StateReadExt>(
+        state: &S,
+        is_proposal_phase: bool,
```
SuperFluffy, line 43: This low level function should not be aware of what happens at the higher level, I think. I looked at the `is_proposal_phase` function and had to follow it all the way here. IMO move this logic to the call-site.

crates/astria-sequencer/src/app/vote_extension.rs
```diff
@@ -0,0 +1,649 @@
...
+        let oracle_vote_extension = transform_oracle_service_prices(state, prices)
+            .await
+            .context("failed to transform oracle service prices")?;
+
+        Ok(abci::response::ExtendVote {
+            vote_extension: oracle_vote_extension.into_raw().encode_to_vec().into(),
+        })
+    }
+
+    pub(crate) async fn verify_vote_extension<S: StateReadExt>(
```
SuperFluffy, line 101: Is it intended that if the oracle is disabled, that this function is still called? If the oracle is disabled, `NO_ORACLE=true`, I would expect that a sequencer node completely abstains from participating in vote extensions.

I think it might be clearer if we separated the config in 1. participates in vote extensions, 2. only verifies vote extensions, 3. ignores vote extensions?

crates/astria-sequencer/src/app/mod.rs
```diff
@@ -297,11 +317,56 @@ impl App {
...
+                    "failed to generate extended commit info"
+                );
+                tendermint::abci::types::ExtendedCommitInfo {
+                    round,
+                    votes: Vec::new(),
+                }
+            }
+        };
+
+        let mut extended_commit_info_bytes =
```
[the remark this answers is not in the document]
noot: it needs to be mut as it's cleared if too large to fit into a block

crates/astria-sequencer/src/app/mod.rs
noot: #1562 addresses that :D

crates/astria-sequencer/src/slinky/marketmap/component.rs
noot: yes: https://github.com/astriaorg/astria/issues/1446

crates/astria-sequencer/src/slinky/oracle/currency_pair_strategy.rs
noot: this is how it's done in the original: https://github.com/skip-mev/connect/blob/793b2e874d6e720bd288e82e782502e41cf06f8c/abci/strategies/currencypair/default.go#L120 would it be clearer if i just moved this function to `vote_extension.rs`?

crates/astria-sequencer/src/app/vote_extension.rs
noot: if the oracle sidecar is disabled, this function should still be called as it's part of consensus, and validator nodes don't need the oracle sidecar to perform verification.

the cases are validator nodes which run the oracle sidecar, those that don't, and full nodes. all validator nodes need to perform verification for consensus reasons, however it's optional as to whether they want to run the sidecar which publishes prices to the network as vote extensions. validators shouldn't ignore vote extension verification, as that could result in them accepting an invalid block. full nodes don't care about vote extensions, to them the VEs are just another transaction in a block (the one the proposer now places first).

crates/astria-core/src/sequencerblock/v1alpha1/block.rs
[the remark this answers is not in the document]
noot: #1764

crates/astria-sequencer/src/app/vote_extension.rs
noot: will separate out the state read

charts/sequencer/files/cometbft/config/genesis.json
```diff
@@ -119,13 +119,31 @@
       {{- if $index }},{{- end }}
       {{ include "sequencer.address" $value }}
       {{- end }}
-    ]
+    ],
```
quasystaty1, line 122: nit: Should move under else statement marking genesis changes.

charts/sequencer/files/cometbft/config/genesis.json
```diff
@@ -119,13 +119,31 @@
...
+        "nextId": "0"
+      }
+    }
     {{- if not .Values.global.dev }}
     {{- else }}
     {{- end}}
   },
   "chain_id": "{{ .Values.genesis.chainId }}",
   "consensus_params": {
+    "abci": {
```
quasystaty1, line 144: nit: should move under else statement, marking breaking changes. You will have to add the closure

charts/sequencer/templates/configmaps.yaml
```diff
@@ -73,7 +73,10 @@ data:
   OTEL_EXPORTER_OTLP_HEADERS: "{{ .Values.sequencer.otel.otlpHeaders }}"
   OTEL_EXPORTER_OTLP_TRACE_HEADERS: "{{ .Values.sequencer.otel.traceHeaders }}"
   OTEL_SERVICE_NAME: "{{ tpl .Values.sequencer.otel.serviceName . }}"
+  ASTRIA_SEQUENCER_CONNECT_GRPC_ADDR: "http://127.0.0.1:8081"
```
quasystaty1, line 76: Should move new env vars to else statement.

charts/sequencer/templates/configmaps.yaml
```diff
@@ -73,7 +73,10 @@ data:
...
+  ASTRIA_SEQUENCER_ORACLE_CLIENT_TIMEOUT_MILLISECONDS: "1000"
```
quasystaty1, line 77: same here for configurable values

quasystaty1 (changes requested):
Should move values to mark breaking changes and have new values configurable with value file instead of constants in configmap.

charts/sequencer/files/cometbft/config/genesis.json
noot: is this breaking since it's not adding a new value? what's the correct syntax for this? do i move the whole `consensus_params` block into the else statement?

Fraser999 (approved):

PR #1242, opened 2024-07-06
fix(sequencer): store native asset ibc->trace mapping in init_chain

noot:
## Summary
we need to store the native asset ibc to "trace" mapping in the state, otherwise queries for the native asset using the ID will fail. for example `get_bridge_account_info` where the asset is the native asset fails right now

## Changes
-  store native asset ibc->trace mapping in `init_chain`

### DIFF
--- a/crates/astria-sequencer/src/app/mod.rs
+++ b/crates/astria-sequencer/src/app/mod.rs
@@ -69,6 +69,7 @@ use crate::{
     },
     address::StateWriteExt as _,
     api_state_ext::StateWriteExt as _,
+    asset::state_ext::StateWriteExt as _,
     authority::{
         component::{
             AuthorityComponent,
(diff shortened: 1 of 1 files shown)


crates/astria-sequencer/src/app/mod.rs
```diff
@@ -220,6 +222,15 @@ impl App {
...
 
         crate::asset::initialize_native_asset(genesis_state.native_asset_base_denomination());
+        let native_asset = crate::asset::get_native_asset();
+        if let Some(trace_native_asset) = native_asset.as_trace_prefixed() {
+            state_tx
+                .put_ibc_asset(trace_native_asset)
+                .context("failed to put native asset")?;
+        } else {
+            bail!("native asset must not be in ibc/<ID> form")
+        }
```
joroshiba, line 232: what's going on with indent here? it feels like there is a missing block here?

SuperFluffy: The indent is actually fine - github code wrapping makes this look a bit weird if your window is too narrow.

SuperFluffy (approved):
That was my oversight, thank you for fixing it.

I actually want to get rid of these globals (meaning the native asset one here and the base address prefix), and instead have a new `StateExt` where these are read everytime they are needed.

I believe this should have a negligible impact on performance, makes testing easier, and would avoid errors such this one.

SuperFluffy:
Added an exclamation mark, as in `fix(sequencer)!`, because this isi breaking

crates/astria-sequencer/src/app/mod.rs
joroshiba: Oh wow that was cursed. dammit github

PR #1248, opened 2024-07-08
chore(sequencer)!: add metrics

### LINKED ISSUE #1247: Provide new sequencer metrics related to `check_tx`.
To help diagnose issues in #1246, we should provide new metrics to expose any bottlenecks in the `check_tx` flow in the sequencer.

Fraser999:
## Summary
This adds further metrics to the sequencer.

## Background
This should help diagnose block production slowdown when the sequencer is stress-tested.

## Changes
- Added metrics (see below for list).
- Enabled `cnidarium` metrics.

Note that all histograms are still rendered as Prometheus summaries for now.  I have [an open PR](https://github.com/astriaorg/astria/pull/1192) which will make it simple to provide buckets for histograms, after which they will be rendered as true histograms.

## Testing
How are these changes tested?

## Metrics
- Added `astria_sequencer_check_tx_duration_seconds` histograms with the following labels:
    - `length check and parse raw tx`
    - `stateless check`
    - `nonce check`
    - `chain id check`
    - `balance check`
    - `check for removal`
    - `insert to app mempool`
- Added `astria_sequencer_actions_per_transaction_in_mempool` histogram
- Added `astria_sequencer_transaction_in_mempool_size_bytes` histogram
- Added `astria_sequencer_transactions_in_mempool_total` gauge
- Enabled `cnidarium_get_raw_duration_seconds` histogram
- Enabled `cnidarium_nonverifiable_get_raw_duration_seconds` histogram

## Related Issues
Closes #1247.

### DIFF
--- a/crates/astria-sequencer/src/metrics.rs
+++ b/crates/astria-sequencer/src/metrics.rs
@@ -1,3 +1,5 @@
+use std::time::Duration;
+
 use metrics::{
     counter,
     describe_counter,
@@ -11,6 +13,8 @@ use metrics::{
     Unit,
 };
 use telemetry::metric_names;
+
+const CHECK_TX_STAGE: &str = "stage";
 
 pub(crate) struct Metrics {
     prepare_proposal_excluded_transactions_cometbft_space: Counter,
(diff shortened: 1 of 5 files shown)


noot (approved):

crates/astria-sequencer/src/metrics.rs
```diff
@@ -212,6 +292,59 @@ impl Metrics {
...
+        #[allow(clippy::cast_precision_loss)]
+        self.actions_per_transaction_in_mempool.record(count as f64);
+    }
+
+    pub(crate) fn record_transaction_in_mempool_size_bytes(&self, count: usize) {
+        // allow: precision loss is unlikely (values too small) but also unimportant in histograms.
+        #[allow(clippy::cast_precision_loss)]
+        self.transaction_in_mempool_size_bytes.record(count as f64);
+    }
+
```
noot, line 343: these are set every time a tx is inserted (and is per-tx) right? would it be more/less useful to also have a mempool total?

PR #1260, opened 2024-07-10
(core, sequencer)!: generate serde traits impls for all protocol protobufs

SuperFluffy:
## Summary
This patch ensures extends our generated serde `Deserialize` and `Serialize` trait implementations to all protobuf messages defined in `astria.protocol`.

## Background
So far the protocol-specs did not have serde traits generated, but these are necessary to write out bridge-withdrawer actions as JSON.

A side-effect of this patch is that all of `proto/protocolapis/astria_vendored` is now also commited to the repository (they were previously filtered). Very likely this is not desired but should be addressed in a followup. It is not clear to the author of this patch in how far these types might be necessary.

## Changes
- Generate serde `Deserialize` and `Serialize` impls for Rust types generated from `astria.protocol` and `astria_vendored` protobufs
- Use a vendored `astria_vendored::tendermint::abci::ValidatorUpdate` instead of `tendermint_proto`: `tendermint_proto@v0.34` does not yet have serde impls for its types, which blocks us from having an easy way to use `pbjson` generated serde impls
- Introduce a new `ValidatorUpdate` action to replace `tendermint::validator::Update` to transform to/from the vendored protobuf type (again necessary to implement serde and foreign type restrictions)
- Update sequencer to use  the new `ValidatorUpdate` type

## Testing
All tests have been upgraded and still pass. Specifically, ingestion of validator updates happens through cometbft, which were already tested end-to-end.

## Breaking Changelist
- Marked as breaking even though none of the breaking change tests are affected: the serialization shape of the validator updates has changed. As they are commited to state using json, this should be breaking.

### DIFF
--- a/crates/astria-core/src/protocol/transaction/v1alpha1/action.rs
+++ b/crates/astria-core/src/protocol/transaction/v1alpha1/action.rs
@@ -24,10 +24,15 @@ use crate::{
 };
 
 #[derive(Clone, Debug)]
+#[cfg_attr(feature = "serde", derive(::serde::Deserialize, ::serde::Serialize))]
+#[cfg_attr(
+    feature = "serde",
+    serde(into = "raw::Action", try_from = "raw::Action")
+)]
 pub enum Action {
     Sequence(SequenceAction),
     Transfer(TransferAction),
-    ValidatorUpdate(tendermint::validator::Update),
+    ValidatorUpdate(ValidatorUpdate),
     SudoAddressChange(SudoAddressChangeAction),
     Ibc(IbcRelay),
     Ics20Withdrawal(Ics20Withdrawal),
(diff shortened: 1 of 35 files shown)


crates/astria-sequencer/src/authority/state_ext.rs
```diff
@@ -36,32 +36,48 @@ struct SudoAddress([u8; ADDRESS_LEN]);
 ///
 /// Contains a map of hex-encoded public keys to validator updates.
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
-pub(crate) struct ValidatorSet(BTreeMap<account::Id, validator::Update>);
+// #[derive(Clone, Debug, PartialEq, Eq)]
```
Fraser999, line 39: Left in by mistake?

crates/astria-core/src/protocol/transaction/v1alpha1/action.rs
```diff
@@ -24,10 +24,15 @@ use crate::{
 };
 
 #[derive(Clone, Debug)]
+#[cfg_attr(feature = "serde", derive(::serde::Deserialize, ::serde::Serialize))]
+#[cfg_attr(
+    feature = "serde",
+    serde(into = "raw::Action", try_from = "raw::Action")
+)]
```
Fraser999, line 31: These could probably be merged.  Same for `ValidatorUpdate` below.

Fraser999 (approved):
LGTM - just a couple of non-blocking nits.

crates/astria-core/src/protocol/transaction/v1alpha1/action.rs
SuperFluffy: You are right, that does look neater. Done for both types in  17d857fc. There are many other instances in the code base where this could be done, but not in thits PR.

PR #1261, opened 2024-07-10
feat(cli): add cmd to collect withdrawal events and submit as actions

SuperFluffy:
## Summary
Adds `bridge collect-withdrawals` and `bridge submit-withdrawals` subcommands to `astria-cli`.

## Background
The worker service `astria-bridge-withdrawer` is a closed system that collects withdrawal event from the rollup and submits them to the sequencer in a closed loop. But it can be desirable to inspect the withdrawal events generated by the bridge contracts, and then submit them manually. This functionality is now provided by `astria-cli`.

## Changes
- Add subcommands `bridge collect-withdrawal-s` to `astria-cli`, which has two different modes of operation, depending on whether `--to-rollup-height` is set:
    - if set, it fetches all blocks between `--from-rollup-height` and `--to-rollup-height` (inclusive), converts them to seuquencer actions, and then writes them to a file.
    - if not set, it fetches blocks from `--from-rollup-height` until SIGINT (Ctrl-C) is received. After the signal, the converted actions are written to a file.

## Testing
- Added `run-smoke-cli` in `charts/deploy.just` (@joroshiba)
- Added a `smoke-cli:` in `.github/workflows/docker-build.yaml` (@joroshiba)

### DIFF
--- a/crates/astria-cli/src/commands/bridge.rs
+++ b/crates/astria-cli/src/commands/bridge.rs
@@ -418,7 +418,7 @@ impl LogToIcs20WithdrawalAction {
             timeout_time: packet_timeout_time()
                 .wrap_err("failed to calculate packet timeout time")?,
             source_channel,
-            bridge_address: None,
+            bridge_address: Some(bridge_address),
         };
         Ok(Action::Ics20Withdrawal(action))
     }
@@ -459,7 +459,7 @@ impl LogToSequencerWithdrawalAction {
                 .ok_or_eyre("failed to divide amount by asset withdrawal multiplier")?,
             memo: serde_json::to_string(&memo).wrap_err("failed to serialize memo to json")?,
             fee_asset,
-            bridge_address: None,
+            bridge_address: Some(bridge_address),
         };
 
         Ok(Action::BridgeUnlock(action))
--- a/crates/astria-cli/src/commands/bridge.rs
+++ b/crates/astria-cli/src/commands/bridge.rs
@@ -53,25 +53,43 @@ use ethers::{
         H256,
     },
 };
+use futures::stream::BoxStream;
 use tokio::select;
 use tracing::warn;
 
 #[derive(Args, Debug)]
 pub struct CollectWithdrawalEvents {
+    /// The websocket endpoint of a geth compatible rollup.
     #[arg(long)]
     rollup_endpoint: String,
+    /// The eth address of the astria bridge contracts.
     #[arg(long)]
     contract_address: ethers::types::Address,
+    /// The start rollup height from which blocks will be checked for withdrawal events.
     #[arg(long)]
-    next_rollup_block_height: u64,
+    from_rollup_height: u64,
+    /// The end rollup height from which blocks will be checked for withdrawal events.
+    /// If not set, then this tool will stream blocks until SIGINT is received.
+    #[arg(long)]
+    to_rollup_height: Option<u64>,
+    /// The asset that will be used to pay the Sequencer fees (should the generated
+    /// actions be submitted to the Sequencer).
     #[arg(long, default_value = "nria")]
     fee_asset: asset::Denom,
+    /// The asset denomination of the asset that's withdrawn from the bridge.
     #[arg(long)]
     rollup_asset_denom: asset::Denom,
+    /// The bech32-encoded bridge address corresponding to the bridged rollup
+    ///  asset on the sequencer. Should match the bridge address in the geth
+    /// rollup's bridge configuration for that asset.
     #[arg(long)]
     bridge_address: Address,
+    /// The address prefix which will be used to construct the
+    /// return address should the withdrawal fail.
     #[arg(long, default_value = "astria")]
     sequencer_address_prefix: String,
+    /// The path to write the collected withdrawal events converted
+    /// to Sequencer actions.
     #[arg(long, short)]
     output: PathBuf,
 }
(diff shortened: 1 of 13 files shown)


crates/astria-cli/src/commands/bridge.rs
```diff
@@ -0,0 +1,484 @@
...
+use tracing::warn;
+
+#[derive(Args, Debug)]
+pub struct CollectWithdrawalEvents {
+    #[arg(long)]
+    rollup_endpoint: String,
+    #[arg(long)]
+    contract_address: ethers::types::Address,
+    #[arg(long)]
+    next_rollup_block_height: u64,
```
noot, line 66: it might be nice to also have a `to_block` option instead of always using the current head block by default

crates/astria-cli/src/commands/bridge.rs
```diff
@@ -0,0 +1,484 @@
...
+        let Some(current_rollup_block) = block_subscription.next().await else {
+            bail!("failed to get current rollup block from subscription")
+        };
+
+        let Some(current_rollup_block_height) = current_rollup_block.number else {
+            bail!(
+                "couldn't determine current rollup block height; value was not set on current on \
+                 most recent block",
+            );
+        };
```
noot, line 122: you don't use the subscription anywhere right? you can just get the latest block and use the number from that instead: https://docs.rs/ethers/latest/ethers/middleware/trait.Middleware.html#method.get_block

crates/astria-cli/src/commands/bridge.rs
```diff
@@ -0,0 +1,484 @@
...
+            memo: serde_json::to_string(&memo).wrap_err("failed to serialize memo to json")?,
+            fee_asset,
+            // note: this refers to the timeout on the destination chain, which we are unaware of.
+            // thus, we set it to the maximum possible value.
+            timeout_height: ibc_types::core::client::Height::new(u64::MAX, u64::MAX)
+                .wrap_err("failed to generate timeout height")?,
+            timeout_time: packet_timeout_time()
+                .wrap_err("failed to calculate packet timeout time")?,
+            source_channel,
+            bridge_address: None,
```
noot, line 421: use `Some(bridge_address)`

crates/astria-cli/src/commands/bridge.rs
```diff
@@ -0,0 +1,484 @@
...
+                .parse()
+                .wrap_err("failed to parse destination chain address")?,
+            amount: event
+                .amount
+                .as_u128()
+                .checked_div(asset_withdrawal_divisor)
+                .ok_or_eyre("failed to divide amount by asset withdrawal multiplier")?,
+            memo: serde_json::to_string(&memo).wrap_err("failed to serialize memo to json")?,
+            fee_asset,
+            bridge_address: None,
```
noot, line 462: also here: `Some(bridge_address)`

crates/astria-cli/src/commands/bridge.rs
```diff
@@ -0,0 +1,484 @@
...
+            transaction_hash,
+        };
+
+        let action = Ics20Withdrawal {
+            denom: rollup_asset_denom,
+            destination_chain_address: event.destination_chain_address,
+            // note: this is actually a rollup address; we expect failed ics20 withdrawals to be
+            // returned to the rollup.
+            // this is only ok for now because addresses on the sequencer and the rollup are both 20
+            // bytes, but this won't work otherwise.
```
noot, line 399: so this is not applicable anymore - we need to also fix this in the withdrawer. this will return funds back to the sequencer address, but we want them to be refunded back to the rollup. we can leave this to a follow-up. basically this should be set to the sender of the withdrawal tx on the rollup, which you should be able to extract this from the transaction data.

crates/astria-cli/src/commands/bridge.rs
SuperFluffy: The subscription gets used further down in the chained stream (not correct code, just for demonstration purposes):
```rust
let incoming_blocks = iter(start_height..current_hight).then(|height| providers.get_bock(height)),chain(block_subscription);
```
Doing it this way is import to not have the rollup advance between getting `current_height`, fetching blocks from `start` to `current_height`, and then going back to the stream.

From a first glance the ethers subscription buffers blocks so that we are not going to miss out

crates/astria-cli/src/commands/bridge.rs
SuperFluffy: So this is actually rollup-specific data, not sequencer-specific?

Should we then make this a breaking change and coll this field `bytes rollup_return_address = x;` in the protobuf?

crates/astria-cli/src/commands/bridge.rs
SuperFluffy: Done in  01da7859

crates/astria-cli/src/commands/bridge.rs
SuperFluffy: Done in  01da7859

crates/astria-cli/src/commands/bridge.rs
SuperFluffy: I have renamed this line to `from_rollup_height` and added an arg `to_rollup_height: Option<u64>` in [f513e17](https://github.com/astriaorg/astria/pull/1261/commits/f513e1783620d4e6cdfb54ca2111935a3f1685f1).

The stream is now bounded if `--to-rollup-height` is set, or unbounded otherwise (requiring ctrl-c to exit).

crates/astria-cli/src/commands/bridge.rs
SuperFluffy: Done in [8333ecf](https://github.com/astriaorg/astria/pull/1261/commits/8333ecfad8bec2d48d213fd27234242e3f1638dc) in after merging https://github.com/astriaorg/astria/pull/1266

joroshiba (approved):

PR #1266, opened 2024-07-12
fix(core, bridge, sequencer)!: dismabiguate return addresses

SuperFluffy:
## Summary
Ensures that bridge-withdrawer does not use a rollup address as a sequencer refund address.

## Background
Bridge-withdrawer was using an address on the ethereum rollup as the sequencer refund address in the actions it was submitting to sequencer. That was wrong.

## Changes
- add a field `rollup_refund_address` to `Ibc20WithdrawalFromRollupMemo` to make its purpose clear
- In bridge-withdrawer, set `refund_address` to the bridge account always (`refund_address` and `bridge_address` are now the same).
- Sequencer now assumes that for refunds of ics20 withdrawals initiated by a rollup `FungibleToketPacketData.sender` is now the bridge address to which funds should be refunded. This is consistent with the change above.
- Remove field `Ibc20WithdrawalFromRollupMemo.brigde_address` as it is no longer necessary.

## Testing
Tests were updated and pass. This patch relies on the tests exhaustively testing the ics20 logic.

## Breaking Changelist
- This change is network-breaking because `Ibc20WithdrawalFromRollupMemo` is used in Sequencer's protocol.

### DIFF
--- a/crates/astria-bridge-withdrawer/src/bridge_withdrawer/ethereum/convert.rs
+++ b/crates/astria-bridge-withdrawer/src/bridge_withdrawer/ethereum/convert.rs
@@ -56,7 +56,6 @@ pub(crate) fn event_to_action(
     rollup_asset_denom: asset::Denom,
     asset_withdrawal_divisor: u128,
     bridge_address: Address,
-    sequencer_address_prefix: &str,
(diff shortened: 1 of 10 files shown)


noot (approved):
nice, this looks great!

PR #1270, opened 2024-07-15
feat(cli, bridge-withdrawer)!: share code between cli and service

SuperFluffy:
## Summary
Moves common code from bridge-withdrawer and cli to `astria-bridge-contracts`.

## Background
The `bridge collect-withdrawals` subcommand of `astria-cli` and the `astria-bridge-withdrawer` have common logic that should use the same code. This patch ensures that both use the same codepaths to get contract withdraw events from the astria bridge contracts running on the rollup before converting them to astria Sequencer actions.

Writing this patch it was noticed that bridge-withdrawer with incorrectly set rollup asset denom would break and was fixed.

## Changes
- Define a `GetWithdrawalActions::get_for_block_hash` utility for fetching withdrawal events for a given block hash and converting them to astria sequencer actions in one go
- Refactor `astria-bridge-withdrawer` and `astria-cli` in terms of it
- Remove `bridge collect-withdrawals --rollup-asset-denom`
- Add `bridge collect-withdrawals --sequencer-asset-to-withdraw` (to disambiguate between withdrawal kinds)
- Add `bridge collect-withdrawals --ics20-asset-to-withdraw` (to disambiguate between withdrawal kinds)
- Require `ASTRIA_BRIDGE_WITHDRAWER_ROLLUP_ASSET_DENOM` to always contain a trace-prefixed ics20 asset denomination (a string of form `[<port>/<channel>/.../]base` because withdrawing `ibc/<hash>` denoms does not make sense for the bridge.

## Testing
Smoke tests still pass and run.

The overly heavy unit tests (which weren't really unit tests) in `astria-bridge-withdrawer` were removed in favor of future blackbox tests (pending in https://github.com/astriaorg/astria/issues/1227, which would have removed these eventually).

## Breaking Changelist
Neither of these are breaking in practice because the CLI `bridge` subcommand did not see a release yet, and `bridge-withdrawer` with `ibc/<hash>` rollup asset denoms was never used (and would likely not work)
- removed an argument on the CLI
- enforces a stricter config on the bridge-withdrawer


crates/astria-bridge-contracts/src/lib.rs
```diff
@@ -1,4 +1,654 @@
...
+    #[must_use]
+    pub fn set_sequencer_asset_to_withdraw(
+        self,
+        sequencer_asset_to_withdraw: Option<asset::Denom>,
+    ) -> Self {
+        Self {
+            sequencer_asset_to_withdraw,
+            ..self
+        }
+    }
```
itamarreif, line 199: why have both of these?

joroshiba (approved):
lgtm

crates/astria-bridge-contracts/src/lib.rs
SuperFluffy: Nice to maintain a builder chain. Could only have the `set_*` one, but I like to maintain synchrony with the other setters.

PR #1280, opened 2024-07-18
fix: abci error code

### LINKED ISSUE #1259: Change `AbciErrorCode` with value 0 to represent `OK`
Currently `AbciErrorCode(0)` is set as `UNSPECIFIED`.  We should change `AbciErrorCode(0)` to `OK` in line with expected abci behaviour.

We should also remove the `From<NonZeroU32>` impl as it's unused and allows for constructing unknown `AbciErrorCode`s.

Fraser999:
## Summary
This changes `AbciErrorCode` to no longer be able to represent tendermint Code 0.

## Background
Previously `0` was defined as `UNSPECIFIED` (in line with common practice for protobuf enum variant assignations).  However, in terms of ABCI codes, `0` should always represent success rather than an error case.

We also had an `impl From<NonZeroU32> for AbciErrorCode` which allowed for constructing arbitrary `AbciErrorCode`s with meaningless values in the context of the Astria codebase.  This impl was unused however.

## Changes
* Changed `AbciErrorCode` to wrap a `NonZeroU32` rather than a `u32` so that const tendermint `Code`s matching the `AbciErrorCode` variants can be created without using `unsafe` blocks.  All such unsafe code is restricted to the `AbciErrorCode` impl and is trivially checkable to be safe.
* Removed the `UNSPECIFIED` variant, making the `AbciErrorCode` name more meaningful, as it now only represents error cases.
* Removed the `From<NonZeroU32>` impl, meaning only the set of defined const values can be used.
* Replaced `impl From<AbciErrorCode> for Code` with `pub const fn value(self) -> NonZeroU32` so that const `Code`s can be used (useful when matching on `Code` values).

## Testing
N/A (all new code is trivial IMO).

## Breaking Changelist
* Technically removing the pub const `AbciErrorCode::UNSPECIFIED` breaks the astria-core crate's API, but as it's unpublished yet, and since that const was never actually used afaik, this should not be an issue.

## Related Issues
Closes #1259.


crates/astria-core/src/protocol/abci.rs
```diff
@@ -1,76 +1,49 @@
...
+            Self::INTERNAL_ERROR => "an internal server error occurred",
+            Self::INVALID_NONCE => "the provided nonce was invalid",
+            Self::TRANSACTION_TOO_LARGE => "the provided transaction was too large",
+            Self::INSUFFICIENT_FUNDS => "insufficient funds",
+            Self::INVALID_CHAIN_ID => "the provided chain id was invalid",
+            Self::VALUE_NOT_FOUND => "the requested value was not found",
+            Self::TRANSACTION_EXPIRED => "the transaction expired in the app's mempool",
+            Self::TRANSACTION_FAILED => "the transaction failed to execute in prepare_proposal()",
+            Self::BAD_REQUEST => "the request payload was malformed",
+            _ => "invalid error code: should be unreachable (this is a bug)",
```
SuperFluffy, line 44: This should report the actual number

crates/astria-core/src/protocol/abci.rs
```diff
@@ -1,76 +1,49 @@
...
+    pub const VALUE_NOT_FOUND: Self = Self(Code::Err(unsafe { NonZeroU32::new_unchecked(8) }));
+    pub const TRANSACTION_EXPIRED: Self = Self(Code::Err(unsafe { NonZeroU32::new_unchecked(9) }));
+    pub const TRANSACTION_FAILED: Self = Self(Code::Err(unsafe { NonZeroU32::new_unchecked(10) }));
+    pub const BAD_REQUEST: Self = Self(Code::Err(unsafe { NonZeroU32::new_unchecked(11) }));
 }
 
 impl AbciErrorCode {
+    /// Returns the wrapped [`Code`].
     #[must_use]
-    pub fn info(self) -> Cow<'static, str> {
```
SuperFluffy, line 28: I would leave this.

crates/astria-core/src/protocol/abci.rs
Fraser999: Happy to add the number, but there's currently no possible way to actually construct an invalid `AbciErrorCode`.  I guess we could add a `Deserialize` constructor in the future which supports constructing an invalid one or something though, so makes sense to add the value here.

Fraser999: Done in [5367a02](https://github.com/astriaorg/astria/pull/1280/commits/5367a0230be034ed20550471579909ba6746b604).

crates/astria-core/src/protocol/abci.rs
Fraser999: Done in [5367a02](https://github.com/astriaorg/astria/pull/1280/commits/5367a0230be034ed20550471579909ba6746b604).

noot (approved):

Issue #1278, opened 2024-07-17
Service panics if it cannot initialize


SuperFluffy:
If `astria_bridge_withdrawer::BridgeWithdrawer::new` fails, the service panics on this line:

https://github.com/astriaorg/astria/blob/bd34b3839c370bfc116c4eff639f094518759c28/crates/astria-bridge-withdrawer/src/main.rs#L56

It should emit an error event and an error code.

PR #1281, opened 2024-07-19
fix(bridge-withdrawer): don't panic on init

itamarreif:
## Summary
Handle errors from constructing the service and emit an error.

## Background
The service was set up to panic instead of emitting an error code and and message if initialization failed. 

## Changes
- Handle errors returned by service init rather than panicking and emit an event.

## Related Issues
closes [#1278](https://github.com/astriaorg/astria/issues/1278)


SuperFluffy (approved):
Thanks :)

PR #1285, opened 2024-07-22
refactor(core, proto)!: define bridge memos in proto

### LINKED ISSUE #1286: Move withdrawal memos into the protocol protobuf spec
The bridge and ics20 related memos are defined in `astria_core::bridge`. However, as they are protocol-critical, they should be defined in the `astria.protocol.v1alpha1` collection of protocol related packages. They need not be ever encoded as binary/protobuf, but having a single source of truth for all protocol-relevant types is desirable.

SuperFluffy:
## Summary
Defines the bridge unlock, ics20 withdrawal, and ics20 deposit memos as protobuf.

## Background
The various memos were read inside sequencer to inform the behaviour of the protocol, but were not defined in the canonical `astria.protocol.v1alpha1` protobufs. This patch defines the memo types in a new `astria.protocol.v1alpha1.memos` so that there is a single source of truth for all protocol-relevant objects. The memos are still only ever serialiazed as plaintext `JSON` and not as protobuf.

## Changes
* Define the following types:
  * `astria.protocol.v1alpha1.memos.BridgeUnlock`
  * `astria.protocol.v1alpha1.memos.Ics20WithdrawalFromRollup`
  * `astria.protocol.v1alpha1.memos.Ics20TransferDeposit`
* Remove the `astria_core::bridge` module
* Update `astria_bridge_contracts`, `astria_bridge_withdrawer`, and `astria_sequencer` to operate in terms of these types
* Update all memo fields containing bytes to be of type `string` so that the rollup-native formatting can be used (to make it easier to identify e.g. transactions and avoid the use of base64 for non-Astria byte slices)
* Various field names have been updated to clarify that their source is the rollup (e.g. `rollup_transaction_hash`, `rollup_return_address`).

## Testing
The changes in this code rely on smoke tests still working. Because the IBC tests require the latest (rust) astria-cli, the IBC tests were reworked.

## Breaking Changelist
* This is a protocl breaking change: the shape of the memo JSON objects has changed (object keys were renamed, the encoding of byte fields no longer requires base64).

## Related Issues
Closes #1286


crates/astria-sequencer/src/ibc/ics20_transfer.rs
```diff
@@ -581,20 +580,27 @@ async fn execute_ics20_transfer_bridge_lock<S: StateWriteExt>(
...
-        !deposit_memo.rollup_address.is_empty(),
+        !deposit_memo.rollup_deposit_address.is_empty(),
         "packet memo field must be set for bridge account recipient",
     );
 
     ensure!(
-        deposit_memo.rollup_address.len() <= MAX_ROLLUP_ADDRESS_BYTE_LENGTH,
+        deposit_memo.rollup_deposit_address.len() <= MAX_ROLLUP_ADDRESS_BYTE_LENGTH,
         "rollup address is too long: exceeds MAX_ROLLUP_ADDRESS_BYTE_LENGTH",
     );
```
joroshiba, line 594: should this validation now be handled by the core library?

crates/astria-sequencer/src/ibc/ics20_transfer.rs
```diff
@@ -581,20 +580,27 @@ async fn execute_ics20_transfer_bridge_lock<S: StateWriteExt>(
...
 
     // assert memo is valid
-    let deposit_memo: Ics20TransferDepositMemo =
+    let deposit_memo: memos::v1alpha1::Ics20TransferDeposit =
         serde_json::from_str(&memo).context("failed to parse memo as Ics20TransferDepositMemo")?;
 
     ensure!(
-        !deposit_memo.rollup_address.is_empty(),
+        !deposit_memo.rollup_deposit_address.is_empty(),
         "packet memo field must be set for bridge account recipient",
```
joroshiba, line 588: nit: issue before this pr, but this error message is inaccurate

proto/protocolapis/astria/protocol/memos/v1alpha1/types.proto
```diff
@@ -0,0 +1,51 @@
...
+  // format of the rollup when targeting plain text encoding.
+  string rollup_transaction_hash = 2;
+  // The return address on the rollup to which funds should returned in case of
+  // failure. This field exists so that the rollup can identify which account
+  // the returned funds originated from.
+  //
+  // This field is of type `string` so that it can be formatted in the preferred
+  // format of the rollup when targeting plain text encoding.
+  string rollup_return_address = 3;
+  string memo = 4;
```
joroshiba, line 40: should we have a comment on this memo field?

proto/protocolapis/astria/protocol/memos/v1alpha1/types.proto
joroshiba, line 1: It seems as though the only breaking change here is changing the strings used for the memos. Would be good to have some discussion over this. I'm not tied to what they are, it's good to have them be designed more.

Technically it is better to have as concise as necessary field names here, as each charachter is bytes we must include in txs and thus blocks, and we should expect briding txs to be some of our absolute most common txs.

.github/workflows/docker-build.yml
```diff
@@ -191,6 +191,10 @@ jobs:
         uses: helm/kind-action@v1
         with:
           install_only: true
+      - name: Install astria cli (rust)
+        run: just install-cli
+      - name: Fetch and install celestia-appd
+        run: just get-celestia-appd v1.9.0 Linux x86_64
```
SuperFluffy, line 197: This could be avoided if we make celestia-local into a stateful set. We could then `kubectl execute` from inside the celestia-local container, making this step unnecessary.

.github/workflows/docker-build.yml
```diff
@@ -204,27 +208,10 @@ jobs:
...
-          JOB_STATUS=$(kubectl get job bridge-tester-chart -n astria-dev-cluster -o jsonpath='{.status.succeeded}')
-          if [ "$JOB_STATUS" != "1" ]; then
-            printlogs
-          else
-            echo "IBC Transfer Test passed"
-          fi
+          just ibc-test deploy $TAG
+      - name: Run IBC ICS20 Transfer test
+        timeout-minutes: 3
+        run: just ibc-test run ./celestia ./celestia-appd
```
SuperFluffy, line 214: Similar to above: the arguments are required because we need to provide a celestia "home" directory, and the path to the celestia-appd executable. It would be neater if we could run straight from the container running celestia-appd.

crates/astria-sequencer/src/ibc/ics20_transfer.rs
SuperFluffy: I wanted that before but got pushback for it (here and for the one above).

IMO it's fine to just leave it as a stupid container type.

proto/protocolapis/astria/protocol/memos/v1alpha1/types.proto
SuperFluffy: I have changed the format of the hash and address types: before, the rust types were of type `[j8; N]` and serialized as base64, now they are opaque `string` so that the rollup can decide to write them in its favorite format.

No hard opinions on the names. I think this would be better addressed by compression of the data before being put into blocks, but maybe that would require too much of a lift. :-/

proto/protocolapis/astria/protocol/memos/v1alpha1/types.proto
SuperFluffy: Added in  b55fa71e

crates/astria-sequencer/src/ibc/ics20_transfer.rs
SuperFluffy: I think it's correct, albeit obtuse: this check ensures that the address be non-empty.

joroshiba: it doesn't validate that the memo field isn't empty though? It validates that the `rollup_desposit_address` is filled?

It used to be the case that we just used whatever was in the memo as the rollup address before we added shape. I don't think we updated the error message at that time.

proto/protocolapis/astria/protocol/memos/v1alpha1/types.proto
joroshiba: I think a long term option would be that we could upgrade to accept json OR base64 formatted pb wire format through the memo. Less human readable, would offer the wire format elimination of names space savings.

Reviewing names, I think these do offer enhanced clarity.

joroshiba (approved):
Approved, I think in general the smoke tests could use cleanup but that can be follow up.

crates/astria-sequencer/src/ibc/ics20_transfer.rs
SuperFluffy: Ah, I see what you mean now. I didn't read the text carefully enough. Updated in  b83d6c59

.github/workflows/docker-build.yml
SuperFluffy: Yes, this indeed has that issue, which running directly inside a stateful set wouldn't have. Followup sounds good. I will create an issue.

charts/deploy.just
```diff
@@ -225,6 +227,16 @@ init-rollup-bridge rollupName=defaultRollupName evmDestinationAddress=evm_destin
...
 
 
+init-ibc-bridge privateKey asset feeAsset rollupName=defaultRollupName:
+  astria-cli sequencer init-bridge-account \
+    --rollup-name {{ rollupName }} \
+    --private-key {{ privateKey }} \
+    --sequencer.chain-id {{ sequencer_chain_id }} \
+    --sequencer-url {{ sequencer_rpc_url }} \
+    --fee-asset {{ feeAsset }} \
+    --asset {{ asset }}
```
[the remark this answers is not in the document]
SuperFluffy: It's a partial duplication. `feeAsset` and `asset` are hardcoded in `init-rollup-bridge`, and there is an additional lock command to lock to the rollup. Could be refactored.

justfile
```diff
@@ -14,6 +13,22 @@ docker-build crate tag=default_docker_tag:
 install-cli:
   cargo install --path ./crates/astria-cli --locked
 
+# version is a celestia-app release. Example: v1.9.0
+# operatingSystem is Linux or Darwin
+# machineHardwareName is arm64 or x86_64
+celestia_default_appd_dst := "."
+get-celestia-appd version operatingSystem machineHardwareName dst=celestia_default_appd_dst:
```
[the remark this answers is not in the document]
SuperFluffy: No, because we don't install it. `cargo install foo` puts foo in a well known spot (`~/.cargo/bin`) that is expexted to be in `PATH`. I don't really know where to put this binary because I don't have knowledge of the users PATH.

Either way, we can remove this entirely when moving to a stateful set.

PR #1291, opened 2024-07-23
fix(proto)!: Change execution API to use primitive RollupId

### LINKED ISSUE #1287: Execution API should use the RollupId primitive
The primitives protobuf spec defines a type `astria.primitives.v1.RollupId` which is used in most other protobuf specs. However, 
`astria.execution.v1alpha2.GenesisInfo.rollup_id` was overlooked and it is still of type `bytes`. `astria.execution.v1alpha2.GenesisInfo` should be updated in terms of it, as well as its corresponding domain type constructor, `astria_core::execution::v1alpha2::RollupId::try_from_raw_raw`.

Note that this will likely also affect https://github.com/astriaorg/astria-geth in its `GenesisInfo` constructor.

## Links into the monorepo
https://github.com/astriaorg/astria/blob/a6d3d96bd483afb752f8a8a038f1e287e9fbf1d9/proto/primitives/astria/primitive/v1/types.proto#L41
https://github.com/astriaorg/astria/blob/a6d3d96bd483afb752f8a8a038f1e287e9fbf1d9/proto/executionapis/astria/execution/v1alpha2/execution.proto#L14
https://github.com/astriaorg/astria/blob/a6d3d96bd483afb752f8a8a038f1e287e9fbf1d9/crates/astria-core/src/execution/v1alpha2/mod.rs#L78-L91

## Corresponding line in astria-geth
https://github.com/astriaorg/astria-geth/blob/e89845ec26ef5bdb2de51603fc3caabe081acf6a/grpc/execution/server.go#L162-L166

ethanoroshiba:
## Summary
Changed execution API `rollup_id` to use primitive `RollupId` instead of bytes.

## Background
`RollupId` is defined in primitives protobuf specs, but `astria.execution.v1alpha2.GenesisInfo.rollup_id` was still using type bytes instead of the primitive type.

## Changes
- Changed execution `rollup_id` to be of type `RollupId`.
- Regenerated Rust sources with protobuf compiler.
- Changed affected conductor functions to utilize `RollupId` instead of bytes.
- Updated `evm-stack` and `evm-rollup` versions.

## Testing
With changes in `astria-geth`, passes smoke-test.

## Breaking Changelist
- Execution API changed, requires corresponding `astria-geth` [changes](https://github.com/astriaorg/astria-geth/pull/38) to function properly.
- Will require followup PR to update geth `devTag` to `latest` once geth PR is merged.

## Related Issues
With `astria-geth` [#38](https://github.com/astriaorg/astria-geth/pull/38), closes #1287


aajimal (approved):

charts/evm-rollup/values.yaml
```diff
@@ -10,7 +10,7 @@ images:
   geth:
     repo: ghcr.io/astriaorg/astria-geth
     tag: 0.13.0
-    devTag: latest
+    devTag: pr-38
```
aajimal, line 13: Just double-checking -- is this tag for all of development intentional?

ethanoroshiba: Was intentional since this PR is dependent on `astria-geth` [#38](https://github.com/astriaorg/astria-geth/pull/38). @joroshiba suggested waiting until this PR merges, updating geth PR with latest protos and merging, then making a monorepo cleanup PR to update `devTag` to `latest`.

joroshiba: Yeah, because we have proto changes here that needs geth update, it gets a little weird here:
1. get this pr submitted which pushes prod proto changes
2. update pr-38 of geth to utilize new main proto libraries, update and submit.
3. update mono repo back to latest.

A bit of chicken/egg problem here. My preference is to keep astria-geth main from relying on non-main protos. We should wait to submit PR here until pr-38 is reviewed and approved though so we don't end up out of sync.

crates/astria-core/src/execution/v1alpha2/mod.rs
```diff
@@ -81,8 +90,11 @@ impl Protobuf for GenesisInfo {
             sequencer_genesis_block_height,
             celestia_block_variance,
         } = raw;
-        let rollup_id =
-            RollupId::try_from_slice(rollup_id).map_err(Self::Error::incorrect_rollup_id_length)?;
+        let Some(rollup_id) = rollup_id else {
+            return Err(Self::Error::no_rollup_id());
+        };
+        let rollup_id = RollupId::try_from_slice(&rollup_id.inner)
+            .map_err(Self::Error::incorrect_rollup_id_length)?;
```
joroshiba, line 97:
```suggestion
        let rollup_id = RollupId::try_from_raw(&rollup_id)
            .map_err(Self::Error::incorrect_rollup_id_length)?;
```

rely on the `RollupId` existing implementation

crates/astria-core/src/execution/v1alpha2/mod.rs
```diff
@@ -104,7 +116,9 @@ impl Protobuf for GenesisInfo {
                  under the hood",
             );
         Self::Raw {
-            rollup_id: Bytes::copy_from_slice(rollup_id.as_ref()),
+            rollup_id: Some(RawRollupId {
+                inner: Bytes::copy_from_slice(&rollup_id.inner),
+            }),
```
joroshiba, line 121:
```suggestion
            rollup_id: Some(rollup_id.to_raw()),
```
you can rely on the existing `RollupId::to_raw()` implementation, instead of recreating here and then do not need to import the RawRollupId type

crates/astria-conductor/src/executor/tests.rs
```diff
@@ -46,7 +49,9 @@ fn make_state(
     }: MakeState,
 ) -> (StateSender, StateReceiver) {
     let genesis_info = GenesisInfo::try_from_raw(raw::GenesisInfo {
-        rollup_id: Bytes::copy_from_slice(ROLLUP_ID.as_ref()),
+        rollup_id: Some(RawRollupId {
+            inner: Bytes::copy_from_slice(ROLLUP_ID.as_ref()),
+        }),
```
joroshiba, line 54:
```suggestion
        rollup_id: ROLLUP_ID.to_raw(),
```

I believe `ROLLUP_ID` is already a type which can cleanly be converted, replace import above with the non raw.

crates/astria-conductor/src/executor/state.rs
```diff
@@ -379,7 +382,9 @@ mod tests {
 
     fn make_genesis_info() -> GenesisInfo {
         GenesisInfo::try_from_raw(raw::GenesisInfo {
-            rollup_id: vec![24; 32].into(),
+            rollup_id: Some(RawRollupId {
+                inner: vec![24; 32].into(),
+            }),
```
joroshiba, line 387: Could do something like....

```suggestion
            rollup_id: Some(RollupId::new(vec![24; 32]).to_raw()),
```

In general throughout trying to avoid importing the raw types as much as possible, rely on the core types and their existing conversions.

I'd probably create the rollup id as it's own variable and then just call `.to_raw()` on it here instead of all nested to keep a bit cleaner

crates/astria-core/src/primitive/v1/mod.rs
```diff
@@ -86,7 +86,7 @@ pub struct RollupId {
         feature = "serde",
         serde(serialize_with = "crate::serde::base64_serialize")
     )]
-    inner: [u8; 32],
+    pub inner: [u8; 32],
```
joroshiba, line 89: In general avoid exposing things as pub without good reason. With changes above we should be able to keep this non pub by using the existing functions.

joroshiba (changes requested):

crates/astria-core/src/primitive/v1/mod.rs
ethanoroshiba: Got it, I'll be sure to try to look for more ways around it in the future. Should be ready for re-review now. Thanks for the feedback!

crates/astria-conductor/tests/blackbox/helpers/macros.rs
```diff
@@ -96,7 +96,9 @@ macro_rules! genesis_info {
         $variance:expr $(,)?
     ) => {
         ::astria_core::generated::execution::v1alpha2::GenesisInfo {
-            rollup_id: ::bytes::Bytes::from($crate::ROLLUP_ID.to_vec()),
+            rollup_id: Some(::astria_core::generated::primitive::v1::RollupId {
+                inner: ::bytes::Bytes::from($crate::ROLLUP_ID.to_vec()),
+            }),
```
joroshiba, line 101:
```suggestion
            rollup_id: Some($crate::ROLLUP_ID.to_raw()),
```
does this work?

ethanoroshiba: Yup! Just committed

crates/astria-core/src/execution/v1alpha2/mod.rs
```diff
@@ -19,12 +19,18 @@ impl GenesisInfoError {
...
+        Self(GenesisInfoErrorKind::NoRollupId)
+    }
 }
 
 #[derive(Debug, thiserror::Error)]
 enum GenesisInfoErrorKind {
-    #[error("`rollup_id` field did not contain a valid rollup ID")]
+    #[error("`rollup_id` field contained an invalid rollup ID")]
     IncorrectRollupIdLength(IncorrectRollupIdLength),
+    #[error("no `rollup_id` provided")]
```
SuperFluffy, line 32: This should call out that the field was not set (technically the same, but most other similar errors in the codebase do it):
```suggestion
    #[error("`rollup_id` was not set")]
```

crates/astria-core/src/execution/v1alpha2/mod.rs
```diff
@@ -104,7 +113,7 @@ impl Protobuf for GenesisInfo {
                  under the hood",
             );
         Self::Raw {
-            rollup_id: Bytes::copy_from_slice(rollup_id.as_ref()),
+            rollup_id: Some(rollup_id.to_raw()),
```
SuperFluffy, line 116: Note that there is also `into_raw`. They do the same thing here, so it doesn't matter. But sometimes taking owernship allows some optimizations (like avoiding another allocation... if one remembers to implement it that way).

SuperFluffy (approved):
Looks great. Nothing to add to @joroshiba's review. Just a small nit in an error message.

crates/astria-core/src/execution/v1alpha2/mod.rs
ethanoroshiba: Got it! So for future reference if something is not a shared object then `into_raw()` may be a better choice?

joroshiba: I'll note that this is within a `to_raw` function. I have generally tried to keep the `to_raw` calling other `to_raw` and vice versa.

joroshiba (approved):

PR #1292, opened 2024-07-24
chore(bridge-withdrawer): cleanup nonce handling

### LINKED ISSUE #1272: Revisit metrics of PR #1215
#1215 added logic that has no metrics.

### LINKED ISSUE #1273: Clean up too many arguments in PR #1215
`process_batch` added in #1215 needs cleaning up.

### LINKED ISSUE #1274: Deduplicate get_pending_nonce in PR #1215
There are two functions to get the pending nonce of the bridge account in #1215. They should be merged.

ethanoroshiba:
## Summary
Deduplicated and added metrics to `get_pending_nonce()`, fixed too many arguments in `process_batch()`.

## Background
Following #1215, `get_pending_nonce()` was present in both the bridge-withdrawer and submitter, in which the submitter's `get_pending_nonce()` also did not implement metrics. The submitter's `process_batch()` function also had a Clippy exception for too many arguments.

## Changes
- Deleted `bridge_withdrawer/startup.rs: get_pending_nonce()` and its helper function `make_sequencer_grpc_retry_config`.
- Instead imported `submitter/mod.rs: get_pending_nonce()` and formatted accordingly.
- Added optional metrics argument to submitter `get_pending_nonce()` to allow for it to be used in the bridge-withdrawer startup.
- Added conditional metrics to `get_pending_nonce()`, as well as mutators within `Metrics`.
- Moved `process_batch()` inside `Submitter` to allow for sharing of members between the two, cutting down arguments and removing Clippy exception.

## Testing
Passes blackbox tests.

## Metrics
- Added `nonce_fetch_count`.
- Added `nonce_fetch_failure_count`.
- Added `nonce_fetch_latency`.

## Related Issues
Closes #1272
Closes #1273
Closes #1274


SuperFluffy:
Is this truly a breaking change? The PR title contains a `!`, but there is no entry in the PR body. (also this change does not sound like it would break anything)

ethanoroshiba:
> Is this truly a breaking change? The PR title contains a `!`, but there is no entry in the PR body. (also this change does not sound like it would break anything)

Certainly not, wasn't up to speed on labeling so thanks for the correction!

crates/astria-bridge-withdrawer/src/metrics.rs
```diff
@@ -26,6 +29,27 @@ impl Metrics {
         describe_gauge!(CURRENT_NONCE, Unit::Count, "The current nonce");
         let current_nonce = gauge!(CURRENT_NONCE);
 
+        describe_counter!(
+            NONCE_FETCH_COUNT,
+            Unit::Count,
+            "The number of times a fetch nonce request to sequencer was made"
```
Fraser999, line 35: This is a bit misleading, since it sounds like it also includes attempts where the fetch failed.  Maybe we should rephrase as something like "The number of times the nonce was fetched from the sequencer" (and we could move the call to `increment_nonce_fetch_count` to after the nonce has been fetched rather than before just to be accurate)

crates/astria-bridge-withdrawer/src/metrics.rs
```diff
@@ -26,6 +29,27 @@ impl Metrics {
...
+
+        describe_counter!(
+            NONCE_FETCH_FAILURE_COUNT,
+            Unit::Count,
+            "The number of failed attempts to fetch nonce from sequencer"
+        );
+        let nonce_fetch_failure_count = counter!(NONCE_FETCH_FAILURE_COUNT);
+
+        describe_histogram!(
+            NONCE_FETCH_LATENCY,
```
Fraser999, line 47: I think the metric name should include the units, but we can probably leave as is for now and review all the metrics' names in a later PR.

crates/astria-bridge-withdrawer/src/bridge_withdrawer/submitter/mod.rs
```diff
@@ -274,22 +271,22 @@ async fn submit_tx(
...
 }
 
-// TODO(https://github.com/astriaorg/astria/issues/1274): deduplicate here and in crate::bridge_withdrawer::startup
-async fn get_pending_nonce(
+pub(crate) async fn get_pending_nonce(
     client: sequencer_service_client::SequencerServiceClient<Channel>,
     address: Address,
     state: Arc<State>,
-    // metrics: &'static Metrics,
+    metrics: Option<&'static Metrics>,
```
Fraser999, line 278: I think we should be able to make this non-optional.  It will mean threading the `Metrics` through a few more places like the `Builder` and `Startup` types, but it shouldn't be a big lift I think.

The benefit is that we get metrics during the startup phase (which I think is the only time this is `None`) and that can help expose the problem if we we're unable to start due to repeatedly failing to fetch the nonce.  It's also a little more efficient for the normal flow after startup (which will be the vast majority of calls here usually) since the code isn't branching on an option which is eternally `Some` after startup.

Fraser999 (approved):

PR #1302, opened 2024-07-26
refactor(ci): execute ics20 test from statefulset celestia container

### LINKED ISSUE #1295: IBC smoke tests shouldn't require an additional binary to install
The tests introduced in https://github.com/astriaorg/astria/pull/1285 require a local installation of `celestia-appd` of a specific version. They further pollute the working directory or require changes to the host system (for example, because a celestia home directory has to be initialized with the private key of the account that makes the ibc transfer).

An alternative would be to run the transfer directly from the container running the celestia instance used for in the tests.

┆Issue Number: ENG-620

### LINKED ISSUE #1296: celestia-local should be a stateful set
charts/celestia-local is currently defined as a `Deployment`:
https://github.com/astriaorg/astria/blob/961294c82c2484423653fea1d690f57ec08cf2e8/charts/celestia-local/templates/deployment.yaml#L2

It is in fact more similar to a pet than cattle and should therefore be a `StatefulSet` instead. Doing this would also unblock https://github.com/astriaorg/astria/issues/1295 because no local copy of `celestia-appd` would be necessary as the transfer can be initiated from the stateful set container.

SuperFluffy:
## Summary
Updates the `charts/celestia-local` to spin up `celestia-appd` in a stateful set. Executes the ics20 transfer inside the celestia-app container inside the set.

## Background
https://github.com/astriaorg/astria/pull/1285 updates the ibc ics20 transfer test to require `celestia-appd` and an initialized celestia keystore to be present on the machine running the test. This is undesirable because it pollutes the host environment and creates potential for collisions. Turning `celestia-local` from a kubernetes deployment to a stateful set allows executing the ics20 transfer from inside the container.

## Changes
- Rename `charts/celestia-local/templates/deployment.yaml -> charts/celestia-local/templates/statefulsets.yaml`
- Changes `kind: Deployment -> kind: StatefulSet` in that file
- Update recipes in `charts/deploy.just` to use `kubectl rollout status` instead of `kubectl wait` (the latter does not seem to work with stateful sets)
- Use `kubectl exec` against the celestia-app container in the stateful set to initialize the transfer instead of a locally present `celestia-appd`

## Testing
This is an update to the ibc ics20 smoke test flow. The test still passes with the expected result.

## Related Issues
Closes #1296
Closes #1295

### DIFF
--- a/charts/celestia-local/Chart.yaml
+++ b/charts/celestia-local/Chart.yaml
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.5.0
+version: 0.6.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -193,8 +193,6 @@ jobs:
           install_only: true
       - name: Install astria cli (rust)
         run: just install-cli
-      - name: Fetch and install celestia-appd
-        run: just get-celestia-appd v1.9.0 Linux x86_64
       - name: Log in to GHCR
         uses: docker/login-action@v2
         with:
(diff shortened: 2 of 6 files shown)


charts/celestia-local/Chart.yaml
```diff
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.5.0
+version: 0.6.0
```
SuperFluffy, line 18: I don't have a good intution for this. Changing the kind of a rollout seems like a major change though.

joroshiba (approved):

PR #1305, opened 2024-07-26
feat(sequencer, core): Add fee reporting

### LINKED ISSUE #1293: report paid fees from sequencer
Currently we don't have any method by which we report fees. We don't fill out the gas wanted or used since we don't have a true concept of gas, it doesn't change in price and we can accept fees in different currencies.

I believe we need to have some reporting on the fees paid for block explorers though, otherwise they need to replicate our logic around fee calculation for actions. In order to support this we can utilize custom events for on the tx_results object of `FinalizeBlock` (see response type [here](https://github.com/cometbft/cometbft/blob/main/spec/abci/abci++_methods.md#finalizeblock), type `ExecTxResult` ([source](https://github.com/cometbft/cometbft/blob/main/spec/abci/abci++_methods.md#exectxresult)), fill out the eve

We could have one event per action on the ExecTxResult for entry of type `tx.fees` and it could have three `attributes` with the key values:

* key: asset
* key: value
* key: actionType

Additionally these tx results could be utilized when calculating fee payments instead of recalculating all fees at final execution time.

ethanoroshiba:
## Summary
Added event recording for fees on `Transfer` and `Sequence` actions.

## Background
Fees were previously not reported, and since Astria doesn't fill out gas wanted/used on `ExecTxResult`, there was no outside way of viewing a block or transaction's associated fees.

## Changes
- Added `get_action_type()` methods to `TransferAction` and `SequencerAction` to facilitate passing the kind of action to `get_and_increase_block_fees()`.
- Added construction and recording of ABCI events having kind `tx.fees` and attributes `asset`, `feeAmount`, and `actionType`.

## Testing
Added unit test to ensure `tx.fees` event is recorded correctly in the transaction state events.

## Related Issues
closes #1293


crates/astria-sequencer/src/app/tests_execute_transaction.rs
```diff
@@ -1042,3 +1042,38 @@ async fn app_execute_transaction_bridge_lock_unlock_action_ok() {
...
+    };
+
+    let signed_tx = Arc::new(tx.into_signed(&alice_signing_key));
+
+    let native_asset = get_native_asset();
+
+    let events = app.execute_transaction(signed_tx).await.unwrap();
+    let transfer_fee = app.state.get_transfer_base_fee().await.unwrap();
+    let fee_event =
+        crate::state_ext::construct_tx_fee_event(native_asset, transfer_fee, "Transfer");
```
Fraser999, line 1077: We could make `construct_tx_fee_event` private if it weren't for using it here.

It's probably about 50-50, but I think I'd have been inclined to favour keeping the function in the production code private and changing the test to something like:
```rust
    let event = events.first().unwrap();
    assert_eq!(event.kind, "tx.fees");
    assert_eq!(
        event.attributes[0],
        ("asset", get_native_asset().to_string()).index().into()
    );
    assert_eq!(
        event.attributes[1],
        ("feeAmount", transfer_fee.to_string()).index().into()
    );
    assert_eq!(
        event.attributes[2],
        ("actionType", "Transfer").index().into()
    );
```

Fraser999 (commented):
Generally looks good, but I made a few nitpicky suggestions :)

crates/astria-core/src/protocol/transaction/v1alpha1/action.rs
```diff
@@ -442,14 +442,20 @@ impl SequenceAction {
             fee_asset,
         })
     }
+
+    /// Returns action type as `&str` for fee event logging
+    #[must_use]
+    pub fn get_action_type(&self) -> &str {
+        "Sequence"
```
SuperFluffy, line 449: The actions are refinement types of types generated protobuf code. Please use https://docs.rs/prost/latest/prost/trait.Name.html instead (preferrably `Name::full_name()`, even though that allocates).

We can achieve this by updating the `astria_core::Protobuf` trait: it already has an associated type `Raw`. We should update it like so:

```rust
trait Protobuf {
    type Raw: prost::Name;

    fn full_name() -> String {
        Self::raw::full_name()
    }
}
```

crates/astria-core/src/protocol/transaction/v1alpha1/action.rs
```diff
@@ -516,6 +522,12 @@ impl TransferAction {
             fee_asset,
         })
     }
+
+    /// Returns action type as `&str` for fee event logging
+    #[must_use]
+    pub fn get_action_type(&self) -> &str {
```
SuperFluffy, line 528: As mentioned above. Don't define this an instead rely on `prost::Name::full_name()`

crates/astria-core/src/protocol/transaction/v1alpha1/action.rs
ethanoroshiba: I may be mistaken, but it looks like `Action` and its variants don't implement the `Protobuf` trait, so even with the changes to Protobuf, the `full_name()` function is not inherited. It does, however, look like the generated action types already implement `prost::Name::full_name()`, so I have changed `get_action_type()` to look like so:
```rust
/// Returns full name of action.
#[must_use]
pub fn full_name() -> String {
    raw::TransferAction::full_name()
}
```
Let me know if you think this is a satisfactory solution, or if you had something else in mind!

SuperFluffy: IMO just impl the trait for the types. We should probably just go ahead and do that for all the types anyway.

ethanoroshiba: I was going to ask, seemed like they all should probably implement the trait. Should I go ahead and do the implementation for all in this PR? Or would it be best to save that for a new ticket?

SuperFluffy: Fine to do it for the subset of actions that you need right now. Can you go ahead and create a ticket for the others?

ethanoroshiba: I'll do that. Thanks for the help!

Fraser999 (approved):

PR #1309, opened 2024-07-29
Revert "test(bridge withdrawer): remove unit tests and add blackbox tests"

joroshiba:
Reverts astriaorg/astria#1232

This PR broke main of bridge-withdrawer. Was not caught by tests due to faulty test setup which does not override bridge withdrawer correctly.


joroshiba:
fixed break with fix instead of revert

PR #1310, opened 2024-07-29
fix(charts, bridge): fix ci test

joroshiba:
## Summary
Fixes bridge withdrawer code which was broken + CI test setup issue which allowed broken test to get in.

## Background
We were not using the latest bridge withdrawer code correctly, this fixes that + fixes the code which broke main CI tests

### DIFF
--- a/charts/deploy.just
+++ b/charts/deploy.just
@@ -192,7 +192,7 @@ deploy-smoke-cli tag=defaultTag:
       --set postgresql.enabled=false \
       --set evm-bridge-withdrawer.enabled=false \
       --set evm-faucet.enabled=false > /dev/null
-  @just wait-for-rollup > /dev/null
+  @just wait-for-dev-rollup > /dev/null
 
 
 evm_destination_address := "0xaC21B97d35Bf75A7dAb16f35b111a50e78A72F30"
--- a/charts/deploy.just
+++ b/charts/deploy.just
@@ -193,6 +193,7 @@ deploy-smoke-cli tag=defaultTag:
       --set evm-bridge-withdrawer.enabled=false \
       --set evm-faucet.enabled=false > /dev/null
   @just wait-for-dev-rollup > /dev/null
+  @sleep 10
 
 
 evm_destination_address := "0xaC21B97d35Bf75A7dAb16f35b111a50e78A72F30"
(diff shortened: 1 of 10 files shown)


charts/deploy.just
```diff
@@ -187,9 +187,10 @@ deploy-smoke-cli tag=defaultTag:
     --create-namespace > /dev/null
   @just wait-for-sequencer > /dev/null
   @echo "Starting EVM rollup..." && helm install -n astria-dev-cluster astria-chain-chart ./charts/evm-stack -f dev/values/rollup/dev.yaml \
-      {{ if tag != '' { replace('--set evm-rollup.images.conductor.devTag=# --set composer.images.composer.devTag=# --set evm-bridge-withdrawer.images.evmBridgeWithdrawer.devTag=#', '#', tag) } else { '' } }} \
+      {{ if tag != '' { replace('--set evm-rollup.images.conductor.devTag=# --set composer.images.composer.devTag=#', '#', tag) } else { '' } }} \
       --set blockscout-stack.enabled=false \
       --set postgresql.enabled=false \
+      --set evm-bridge-withdrawer.enabled=false \
```
joroshiba, line 193: this is in the cli test, it shouldn't be running the bridge withdrawer

charts/evm-stack/Chart.yaml
```diff
@@ -29,10 +29,10 @@ dependencies:
     version: 0.1.0
     repository: "file://../evm-faucet"
     condition: evm-faucet.enabled
-  - name: bridge-withdrawer
+  - name: evm-bridge-withdrawer
```
joroshiba, line 32: Our overrides were over the wrong chart sha in tests. The name here wasn't following norms, the name here should match the chart name, fixed the chart instead of just changing values overrides.

crates/astria-bridge-withdrawer/src/bridge_withdrawer/startup.rs
```diff
@@ -409,12 +409,11 @@ async fn wait_for_empty_mempool(
                 futures::future::ready(())
             },
         );
-    let sequencer_client =
-        SequencerServiceClient::connect(format!("http://{}", sequencer_grpc_endpoint.clone()))
-            .await
-            .wrap_err_with(|| {
-                format!("failed to connect to sequencer at `{sequencer_grpc_endpoint}`")
-            })?;
+    let sequencer_client = SequencerServiceClient::connect(sequencer_grpc_endpoint.clone())
```
joroshiba, line 412: We shouldn't be forcing http on the url path here, uri protocol specified in the env var.

SuperFluffy (approved):
Rust is correct. I could have sworn I flagged this in a review, but maybe forgot.

This should be refactored so that there is a single point of failure when constructing this client. I will create a follow-up.

steezeburger (approved):

PR #1317, opened 2024-07-31
refactor(sequencer): remove global state

### LINKED ISSUE #1208: Remove global state in sequencer
Sequencer sets global state in https://github.com/astriaorg/astria/blob/167f72fc780fc8df5562930ff29d35a978b14992/crates/astria-sequencer/src/asset/mod.rs#L10 and, with PR #1201, similary for the address prefixes.

Since state is explicitly passed around to almost all parts of the sequencer through `Storage`, there is no need to set it globally. Instead, address prefixes and denoms could be read from rocksdb every time they are needed. Note that because the getters are async, this could have performance implications.

SuperFluffy:
## Summary
Removed all global from sequencer.

## Background
Upon init, sequencer stored its native asset and address base prefix at a fixed global memory locations. This did not yet lead to issues during runtime (because we don't mutate global after startup), it made testing unnecessarily difficult: sequencer relies heavily on unit tests, yet the tests were impure because they relied on the same global state.

The changes to the `mempool` module shows another issue: while the mempool could be wholly agnostic to the specific global state, heavy use of the global base prefix meant was less decoupled than it could be.

Because sequencer explicitly passes around state (in the form of state deltas, the data inside which are made accessible via various `StateReadExt` and `StateWriteExt` traits), we can read the configured base prefix and native assets from there.

The downside is that this patch introduces more database reads. However, given the amounts of data sequencer is dealing with this might not matter and can be alleviated by using a rocksdb Cache (or other mechanism) instead of relying on globals.

## Changes
- Remove global setters and writers in `crate::address` and `crate::assets`
- Update all code to read the base address prefix and native asset through `crate::address::StateReadExt` and `crate::address::assets::StateReadExt` instead.
- Update the mempool to be agnostic of address prefixes and use address bytes insteda: this change is in line with sequencer design considerations in that the address prefix only matters at the boundary while state reads are done via the address bytes underlying the bech32m addresses.
- Require that the sequencer genesis contains a `TracePrefixed` asset.

## Testing
Updated all tests to either use hard coded prefixes, native assets (if not performing reads or writes on state deltas), or write a base prefix and native asset to state before performing tests (if accessing state).

## Breaking Changelist
While the change to `astria-core` is technically breaking (because a field that was of type `asset::Denom` is now `asset::TracePrefixed`), in practice this change is not breaking because Sequencer required native assets to be non-ibc prefixed.

## Related Issues
Closes #1208


crates/astria-sequencer/src/accounts/state_ext.rs
```diff
@@ -36,12 +41,49 @@ struct Fee(u128);
 const ACCOUNTS_PREFIX: &str = "accounts";
 const TRANSFER_BASE_FEE_STORAGE_KEY: &str = "transferfee";
 
-struct StorageKey<'a>(&'a Address);
-impl<'a> std::fmt::Display for StorageKey<'a> {
+trait GetAddressBytes: Send + Sync {
```
Fraser999, line 44: I'd prefer if this were renamed `AddressBytes` and the method likewise, but not a big deal.

crates/astria-sequencer/src/sequencer.rs
```diff
@@ -89,18 +89,14 @@ impl Sequencer {
...
-            crate::assets::initialize_native_asset(&native_asset.to_string());
-            let base_prefix = snapshot
+                .context("failed to query state for native asset")?;
+            let _ = snapshot
                 .get_base_prefix()
                 .await
-                .context("failed to get address base prefix from storage")?;
-            crate::address::initialize_base_prefix(&base_prefix)
-                .context("failed to initialize global address base prefix")?;
+                .context("failed to query state for base prefix")?;
```
Fraser999, line 99: Is there any point to still doing these calls?

Fraser999 (approved):

crates/astria-sequencer/src/accounts/state_ext.rs
SuperFluffy: Addressing this as part of #1318

crates/astria-sequencer/src/sequencer.rs
SuperFluffy: I think you are right. I am going to remove these as part of addressing #1318

PR #1319, opened 2024-07-31
chore(core, proto): migrate byte slices from Vec to Bytes

### LINKED ISSUE #674: core: migrate byte slices from Vec to Bytes
This is to use zero-copy deserialization from protobuf encoded bytes to the generated Rust types, and to make conversion between the generated Rust types and idiomatic Astria types cheaper.

This issue has a few steps:

1. Update the [protobuf compilation tool](https://github.com/astriaorg/astria/blob/8fa15cec0e8b4039470f64ad9ced6442075bc5f6/tools/protobuf-compiler/src/main.rs#L66) to emit `Bytes` instead of `Vec<u8>`
2. (generate new code by running the tool)
3. Update all conversions in `astria-core` between the protobuf and native domain types (primarily all `try_from_raw`, `try_from_raw_ref`, `into_raw`, and `to_raw`).
4. Update all code that broke because it required `Vec<u8>` (I expect this primarily affecting tests).

For step 3 in particular care should be taken to not end up with conversions `Bytes -> Vec<u8> -> Bytes` because we have some instances where we convert `Bytes` to `Vec<u8>`.

ethanoroshiba:
## Summary
Changed protobuf, astria-core, and dependent code to utilize generic `Bytes` instead of the default `Vec<u8>` to utilize zero-copy deserialization from Protobuf and make type conversion cheaper.

## Background
Previously, some components were utilizing `Bytes` and some were utilizing `Vec<u8>` as the generated Rust types when converting from Protobuf `bytes`. 

## Changes
- Adjusted Protobuf compiler to use `Bytes` for all Astria paths.
- Edited conversions to/from `Raw` in `astria-core` to utilize `prost::bytes::Bytes`.
- Adjusted code which relied on `Vec<u8>`.

## Testing
Passing all tests.

## Related Issues
closes #674


crates/astria-core/src/sequencerblock/v1alpha1/block.rs
```diff
@@ -448,12 +454,14 @@ impl SequencerBlockHeader {
...
-                SequencerBlockHeaderError::incorrect_rollup_transactions_root_length(e.len())
+            rollup_transactions_root.chunk().try_into().map_err(|_| {
+                SequencerBlockHeaderError::incorrect_rollup_transactions_root_length(
+                    rollup_transactions_root.len(),
+                )
             })?;
 
-        let data_hash = data_hash.try_into().map_err(|e: Vec<_>| {
-            SequencerBlockHeaderError::incorrect_rollup_transactions_root_length(e.len())
+        let data_hash = data_hash.chunk().try_into().map_err(|_| {
```
Fraser999, line 463:
```suggestion
        let data_hash = data_hash.as_ref().try_into().map_err(|_| {
```

crates/astria-core/src/sequencerblock/v1alpha1/block.rs
```diff
@@ -844,8 +854,9 @@ impl SequencerBlock {
         } = raw;
 
         let block_hash = block_hash
+            .chunk()
```
Fraser999, line 857:
```suggestion
            .as_ref()
```

crates/astria-core/src/sequencerblock/v1alpha1/block.rs
```diff
@@ -1061,8 +1076,9 @@ impl FilteredSequencerBlock {
...
```
Fraser999, line 1079:
```suggestion
            .as_ref()
```

crates/astria-core/src/sequencerblock/v1alpha1/celestia.rs
```diff
@@ -258,8 +266,8 @@ impl SubmittedRollupData {
         };
         let rollup_id =
             RollupId::try_from_raw(&rollup_id).map_err(SubmittedRollupDataError::rollup_id)?;
-        let sequencer_block_hash = sequencer_block_hash.try_into().map_err(|bytes: Vec<u8>| {
-            SubmittedRollupDataError::sequencer_block_hash(bytes.len())
+        let sequencer_block_hash = sequencer_block_hash.chunk().try_into().map_err(|_| {
```
Fraser999, line 269:
```suggestion
        let sequencer_block_hash = sequencer_block_hash.as_ref().try_into().map_err(|_| {
```

crates/astria-core/src/sequencerblock/v1alpha1/block.rs
```diff
@@ -448,12 +454,14 @@ impl SequencerBlockHeader {
         .map_err(SequencerBlockHeaderError::time)?;
 
         let rollup_transactions_root =
-            rollup_transactions_root.try_into().map_err(|e: Vec<_>| {
-                SequencerBlockHeaderError::incorrect_rollup_transactions_root_length(e.len())
+            rollup_transactions_root.chunk().try_into().map_err(|_| {
```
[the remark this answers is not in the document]
ethanoroshiba: Should be done in d92a373.

crates/astria-core/src/protocol/bridge/v1alpha1/mod.rs
```diff
@@ -28,10 +33,14 @@ impl BridgeAccountLastTxHashResponse {
...
+                .map(|bytes| bytes.chunk().try_into())
                 .transpose()
-                .map_err(|bytes: Vec<u8>| {
-                    BridgeAccountLastTxHashResponseError::invalid_tx_hash(bytes.len())
+                .map_err(|_| {
+                    let Some(tx_hash) = raw.tx_hash else {
+                        return BridgeAccountLastTxHashResponseError::missing_tx_hash();
+                    };
+                    BridgeAccountLastTxHashResponseError::invalid_tx_hash(tx_hash.len())
                 })?,
```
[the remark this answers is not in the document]
ethanoroshiba: Done in d92a373.

crates/astria-conductor/src/celestia/reconstruct.rs
```diff
@@ -56,7 +57,12 @@ pub(super) fn reconstruct_blocks_from_verified_blobs(
                 celestia_height,
                 block_hash: header_blob.block_hash(),
                 header: header_blob.into_unchecked().header,
-                transactions: rollup.into_unchecked().transactions,
+                transactions: rollup
+                    .into_unchecked()
+                    .transactions
+                    .into_iter()
+                    .map(Bytes::into)
+                    .collect(),
```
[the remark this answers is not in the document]
ethanoroshiba: Managed to get a few layers deeper, but found no other way than Vec <> Byte conversion when it came to `RollupData`, since its `SequencedData` variant is of type `Vec<Bytes>`, but its `Deposit` variant is of type `Box<Deposit>`. Additionally, due to changing the type of `RollupTransactions`, the new commit was failing this test: https://github.com/astriaorg/astria/blob/d92a37366a01d190a92f578d763ec54c8b0d15cc/crates/astria-conductor/src/sequencer/reporting.rs#L79
I generated another snapshot, but let me know if you think this is appropriate. Thanks for the feedback :)

crates/astria-core/src/sequencerblock/v1alpha1/block.rs
```diff
@@ -744,8 +749,11 @@ impl SequencerBlock {
                     fee_asset: _,
                 }) = action
                 {
-                    let elem = rollup_datas.entry(rollup_id).or_insert(vec![]);
-                    let data = RollupData::SequencedData(data).into_raw().encode_to_vec();
+                    let elem = rollup_datas.entry(rollup_id).or_insert(vec![Bytes::new()]);
```
Fraser999, line 752:
```suggestion
                    let elem = rollup_datas.entry(rollup_id).or_insert(vec![]);
```

crates/astria-conductor/src/celestia/reconstruct.rs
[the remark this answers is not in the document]
Fraser999: Looks like this is sorted now, and the snapshots remain unchanged, which is what we want :)

Fraser999 (approved):
One call to `chunk()` has been missed (I unresolved the comment to flag it - it's in celestia.rs).  Otherwise LGTM - nice one.

crates/astria-composer/src/executor/tests.rs
```diff
@@ -13,7 +13,10 @@ use astria_core::{
 };
 use astria_eyre::eyre;
 use once_cell::sync::Lazy;
-use prost::Message;
+use prost::{
+    bytes::Bytes,
+    Message,
```
SuperFluffy, line 18: `Message as _`

crates/astria-conductor/src/executor/client.rs
```diff
@@ -117,14 +117,14 @@ impl Client {
...
-        transactions: Vec<Vec<u8>>,
+        transactions: Vec<Bytes>,
         timestamp: Timestamp,
     ) -> eyre::Result<Block> {
         use prost::Message;
 
         let transactions = transactions
             .into_iter()
-            .map(|tx| RollupData::decode(tx.as_slice()))
+            .map(|tx| RollupData::decode(tx.as_ref()))
```
SuperFluffy, line 127: I think you don't need to do that call here: just `map(RollupData::decode)` because `decode(impl Buf)`: https://docs.rs/prost/latest/prost/trait.Message.html#method.decode

crates/astria-core/src/protocol/bridge/v1alpha1/mod.rs
```diff
@@ -22,25 +24,27 @@ impl BridgeAccountLastTxHashResponse {
     ///
     /// - if the transaction hash is not 32 bytes
     pub fn try_from_raw(
-        raw: raw::BridgeAccountLastTxHashResponse,
+        raw: &raw::BridgeAccountLastTxHashResponse,
```
SuperFluffy, line 27: Please keep the argument taking ownership.

If you want an argument taking a reference, add `try_from_raw_ref`

crates/astria-core/src/protocol/bridge/v1alpha1/mod.rs
```diff
@@ -22,25 +24,27 @@ impl BridgeAccountLastTxHashResponse {
...
         Ok(Self {
             height: raw.height,
             tx_hash: raw
                 .tx_hash
-                .map(TryInto::<[u8; 32]>::try_into)
-                .transpose()
-                .map_err(|bytes: Vec<u8>| {
-                    BridgeAccountLastTxHashResponseError::invalid_tx_hash(bytes.len())
-                })?,
+                .clone()
```
SuperFluffy, line 33: This clone seems redundant.

crates/astria-core/src/protocol/test_utils.rs
```diff
@@ -148,7 +150,10 @@ impl ConfigureSequencerBlock {
             rollup_ids_root.to_vec(),
         ];
         data.extend(txs.into_iter().map(|tx| tx.into_raw().encode_to_vec()));
-
+        let data = data
+            .into_iter()
+            .map(|data| Bytes::copy_from_slice(&data))
```
SuperFluffy, line 155: Use `Bytes::from`. It can save an alloction.

crates/astria-core/src/sequencerblock/v1alpha1/block.rs
```diff
@@ -196,7 +199,7 @@ impl SequencerBlockError {
         Self(SequencerBlockErrorKind::NoRollupTransactionsRoot)
     }
 
-    fn incorrect_rollup_transactions_root_length(len: usize) -> Self {
+    fn incorrect_rollup_tx_root_length(len: usize) -> Self {
```
SuperFluffy, line 202: This should match the name of the error variant. Change back.

crates/astria-core/src/sequencerblock/v1alpha1/celestia.rs
```diff
@@ -1,3 +1,7 @@
+use bytes::{
+    Buf,
```
SuperFluffy, line 2: Import as `_`

crates/astria-core/src/sequencerblock/v1alpha1/celestia.rs
```diff
@@ -51,6 +55,10 @@ impl PreparedBlock {
                 proof,
                 ..
             } = rollup_txs.into_parts();
+            let transactions = transactions
+                .into_iter()
+                .map(|tx| Bytes::copy_from_slice(&tx))
```
SuperFluffy, line 60: I believe this can use `into` to potentially avoid allocation

SuperFluffy (commented):
I would like to see a couple of changes before this is merged: a few allocations can be avoided and a few function signatures shouldn't be changed

crates/astria-core/src/sequencerblock/v1alpha1/block.rs
ethanoroshiba: This was to avoid a clippy warning of too many lines in `SequencerBlock::try_from_block_info_and_data`, as the lengthened function name pushes the formatter to break the `map_err` in which it is called into 3 lines instead of one. Would it be best for me to try to refactor in this case and avoid a clippy exception? Thanks for the detailed feedback :)

SuperFluffy: Yeah, that seems to a hilarious way to appease the linter with the downside of breaking symmetry with all other methods. :-) The lint hitting that other function is probably correct. But I'm happy with you adding an `allow` directive on top and address it in a future PR.

crates/astria-core/src/primitive/v1/mod.rs
```diff
@@ -176,9 +177,11 @@ impl RollupId {
     /// # Errors
     ///
     /// Returns an error if the byte slice was not 32 bytes long.
-    pub fn try_from_vec(bytes: Vec<u8>) -> Result<Self, IncorrectRollupIdLength> {
-        let inner =
-            <[u8; ROLLUP_ID_LEN]>::try_from(bytes).map_err(|bytes| IncorrectRollupIdLength {
+    pub fn try_from_bytes(bytes: &Bytes) -> Result<Self, IncorrectRollupIdLength> {
```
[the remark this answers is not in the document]
ethanoroshiba: Removed in [9f2e333](https://github.com/astriaorg/astria/pull/1319/commits/9f2e3332e7d78690a0bb144788b07b7d9eb7c993)

Fraser999 (approved):

Issue #1321, opened 2024-07-31
chore: add instrumentation on all async calls


ethanoroshiba:
Mentioned in the Protocol Weekly on 7/25, instrumentation should be added on all async calls.

Things to keep in mind (via Janis):

> 1. we need to remove instrumentation from all long-running tasks (in our case, essentially everything that is called T::run or T::run_until_stopped
> 2. we also need to remove all events that run under these (so if you have a loop { select!((....)} construct, then there cannot be events in these
>             a. The reason is that forever spans are completely useless on an observability platforms; in the best case we will see them at the end of a run, in the worst case we will lose them completely because the gRPC payload is too big (because the span contains too many events).
> 3. while we want to have #[instrument(skip_all, err)], we don't necessarily want the errors to be emitted under an ERROR target. ERROR is for service-stopping issues, while WARN is for issues that can be handled (for example, through a retry).
>             a. using this example, this would be #[instrument(err(level = Level::WARN)]

```
### Tasks
- [x] bridge-withdrawer
- [x] composer
- [x] conductor
- [ ] sequencer
- [x] sequencer-relayer
```

┆Issue Number: ENG-670

joroshiba:
This issue is stale because it has been open 45 days with no activity. Remove stale label or this issue
be closed in 7 days.

PR #1324, opened 2024-07-31
chore(bridge-withdrawer): Add instrumentation to bridge-withdrawer

ethanoroshiba:
## Summary
Added instrumentation to `bridge-withdrawer`

## Background

## Changes
- Added instrumentation to all async function calls that are not long-lived.

## Related Issues
Part of #1321

### DIFF
--- a/crates/astria-bridge-withdrawer/src/bridge_withdrawer/submitter/mod.rs
+++ b/crates/astria-bridge-withdrawer/src/bridge_withdrawer/submitter/mod.rs
@@ -134,7 +134,7 @@ impl Submitter {
         Ok(())
     }
 
-    #[instrument(skip_all)]
+    #[instrument(skip_all, err)]
     async fn process_batch(
         &self,
         sequencer_grpc_client: SequencerServiceClient<Channel>,
@@ -224,7 +224,8 @@ impl Submitter {
     fields(
         nonce = tx.nonce(),
         transaction.hash = %telemetry::display::hex(&tx.sha256_of_proto_encoding()),
-    )
+    ),
+    err
 )]
 async fn submit_tx(
     client: sequencer_client::HttpClient,
@@ -279,6 +280,7 @@ async fn submit_tx(
     res
 }
 
+#[instrument(skip_all, err)]
 pub(crate) async fn get_pending_nonce(
     client: sequencer_service_client::SequencerServiceClient<Channel>,
     address: Address,
--- a/crates/astria-bridge-withdrawer/src/api.rs
+++ b/crates/astria-bridge-withdrawer/src/api.rs
@@ -20,6 +20,7 @@ use http::status::StatusCode;
 use hyper::server::conn::AddrIncoming;
 use serde::Serialize;
 use tokio::sync::watch;
+use tracing::instrument;
 
 use crate::bridge_withdrawer::StateSnapshot;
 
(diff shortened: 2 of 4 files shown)


crates/astria-bridge-withdrawer/src/bridge_withdrawer/mod.rs
```diff
@@ -245,6 +248,47 @@ impl BridgeWithdrawer {
     }
 }
 
+type TaskHandles = (
+    JoinHandle<eyre::Result<()>>,
+    Option<JoinHandle<eyre::Result<()>>>,
+    JoinHandle<eyre::Result<()>>,
+    JoinHandle<eyre::Result<()>>,
+);
```
Fraser999, line 256: Maybe cleaner to just create a struct, especially since most of the elements here are of the same type (harder that way to mix them up at the callsite of `spawn_tasks`).

crates/astria-bridge-withdrawer/src/bridge_withdrawer/submitter/mod.rs
```diff
@@ -182,25 +176,23 @@ impl Submitter {
...
-                "check_tx failure upon submitting transaction to sequencer"
-            ))
+            Err(eyre!(format!(
+                "abci.code = {},
+                abci.log = {},
+                rollup.height = {},
+                check_tx failure upon submitting transaction to sequencer: transaction failed to \
+                 be included in the mempool, aborting.",
+                check_tx_code, rsp.check_tx.log, rollup_height
+            )))
```
Fraser999, line 186: Not sure I'd try and force the sort of formatting `tracing` produces since this will all end up as a single string in the `message` field I believe.  Also, we don't want line breaks in the message string, and `eyre!` supports passing a format string and args (so need for the `format!` use).
```suggestion
            Err(eyre!(
                "check_tx failure upon submitting transaction to sequencer: transaction failed to \
                 be included in the mempool, aborting. abci.code = {check_tx_code}, abci.log = \
                 {}, rollup.height = {rollup_height}",
                rsp.check_tx.log
            ))
```

crates/astria-bridge-withdrawer/src/bridge_withdrawer/submitter/mod.rs
```diff
@@ -182,25 +176,23 @@ impl Submitter {
...
-                "deliver_tx failure upon submitting transaction to sequencer"
-            ))
+            Err(eyre!(format!(
+                "abci.code = {},
+                abci.log = {},
+                rollup.height = {},
+                deliver_tx failure upon submitting transaction to sequencer: transaction failed to \
+                 be executed in a block, aborting.",
+                deliver_tx_code, rsp.tx_result.log, rollup_height
+            )))
```
Fraser999, line 195:
```suggestion
            Err(eyre!(
                "deliver_tx failure upon submitting transaction to sequencer: transaction failed \
                 to be executed in a block, aborting. abci.code = {deliver_tx_code}, abci.log = \
                 {}, rollup.height = {rollup_height}",
                rsp.tx_result.log,
            ))
```

crates/astria-bridge-withdrawer/src/bridge_withdrawer/mod.rs
```diff
@@ -245,6 +252,47 @@ impl BridgeWithdrawer {
     }
 }
 
+pub struct TaskHandles {
```
Fraser999, line 255:
```suggestion
struct TaskHandles {
```

Fraser999 (approved):
One tiny nitpick, but otherwise LGTM.

PR #1337, opened 2024-08-01
perf(sequencer): add benchmark for prepare_proposal (ENG-660)

### LINKED ISSUE #1314: Provide faster feedback loop for assessing performance of sequencer
Investigation into sequencer slowdown under load is pointing towards block execution being a major contributor.

Currently this has been discovered via running spamoor, which requires running the composer and cometbft too.  We would benefit from a more targeted solution, e.g. a benchmark suite for the `Sequencer` struct or even the `App`.

┆Issue Number: ENG-660

Fraser999:
## Summary
Addition of a new benchmark target to assess the performance of the prepare_proposal method in sequencer's `App`.

## Background
Previous perf work has indicated this is a bottleneck.  However, making that determination was done via spamoor which is slightly convoluted to run.  Before working to improve the performance, we want to have a faster feedback loop on the effects of updates, hence the need for a benchmark which is easy to run and which isolates the slow function.

## Changes
- Added benchmark to `app` module.  Currently this has only one case: a mempool filled with transactions containing exclusively transfers.  This matches the shape of the data being sent when using spamoor.
- Added `benchmark` feature to enable sharing some of the existing test utils.

## Testing
This is a new test.

Example of running `cargo bench --features=benchmark -qp astria-sequencer app` on my Ryzen 7900X:
```
Timer precision: 10 ns
astria_sequencer                                fastest       │ slowest       │ median        │ mean          │ samples │ iters
╰─ app                                                        │               │               │               │         │
   ╰─ benchmarks                                              │               │               │               │         │
      ╰─ execute_transactions_prepare_proposal  11.63 s       │ 14.68 s       │ 12.74 s       │ 12.88 s       │ 10      │ 10
```

Since rebasing after #1317 has merged, the same run shows (as expected) a slowdown:
```
astria_sequencer                                fastest       │ slowest       │ median        │ mean          │ samples │ iters
╰─ app                                                        │               │               │               │         │
   ╰─ benchmarks                                              │               │               │               │         │
      ╰─ execute_transactions_prepare_proposal  14.49 s       │ 17 s          │ 16.52 s       │ 15.98 s       │ 8       │ 8
```

## Related Issues
Closes #1314.


crates/astria-sequencer/src/app/benchmarks.rs
```diff
@@ -0,0 +1,103 @@
...
+        let first_address = accounts.first().unwrap().address;
+        let unchecked_genesis_state = UncheckedGenesisState {
+            accounts,
+            address_prefixes,
+            authority_sudo_address: first_address,
+            ibc_sudo_address: first_address,
+            ibc_relayer_addresses: vec![],
+            native_asset_base_denomination: nria(),
+            ibc_params: IBCParameters::default(),
+            allowed_fee_assets: vec!["nria".parse().unwrap()],
```
noot, line 64:
```suggestion
            allowed_fee_assets: vec![nria().into()],
```

crates/astria-sequencer/src/app/benchmarks.rs
```diff
@@ -0,0 +1,103 @@
...
+
+#[divan::bench(max_time = MAX_TIME)]
+fn execute_transactions_prepare_proposal(bencher: divan::Bencher) {
+    let runtime = tokio::runtime::Builder::new_multi_thread()
+        .enable_all()
+        .build()
+        .unwrap();
+    let mut fixture = runtime.block_on(async { Fixture::new().await });
+    bencher
+        .with_inputs(|| BlockSizeConstraints::new(22_019_254).unwrap())
```
noot, line 90: what's this number from?

noot (approved):

crates/astria-sequencer/src/app/benchmarks.rs
Fraser999: It was taken from the actual value seen in `prepare_proposal.max_tx_bytes` when handling `prepare_proposal` during stress testing using spamoor.  I'll add a comment.

Fraser999: Added comment in [8436f24](https://github.com/astriaorg/astria/pull/1337/commits/8436f2406c97a13be5182a3f358332bce85de1c0).

Issue #1340, opened 2024-08-02
CheckTx should use the same code paths as regular block production


SuperFluffy:
┆Issue Number: ENG-686

joroshiba:
This issue is stale because it has been open 45 days with no activity. Remove stale label or this issue
be closed in 7 days.

joroshiba:
This issue is stale because it has been open 45 days with no activity. Remove stale label or this issue
be closed in 7 days.

joroshiba:
This issue was closed because it was stale

Issue #1318, opened 2024-07-31
Sequencer suffers from TOCTOU issues


SuperFluffy:
Bug found by auditors:

>
> Due to the way action handling is done, unexpected behaviors can be noted.
> 
> a transactions action's stateful checks are ran together in parallel. Then the actions are executed.
> ```
> Action 1 -> check_stateful()
> Action 2 -> check_stateful()
> ...
> Action 1 -> Execute()
> Action 2 -> Execute()
> ```
> This means that certain actions could have unforseen consequences. Imagine 2 BridgeSudoChangeAction actions which have been bundled into 1 tx, they would both execute even if the first action changes the sudo address because the stateful checks of both actions are executed before the execute
> ```
>     async fn check_stateless(&self) -> anyhow::Result<()> {
>         ensure!(!self.actions.is_empty(), "must have at least one action");
> 
>         for action in &self.actions {
>             match action {
> ```

┆Issue Number: ENG-666

SuperFluffy:
Some thoughts I have for this:

Because check_stateless, check_stateful, and execute are split and run one after the other, this inherently creates ace conditions. The way to fix this is to move the checks into execute. Given their serial nature this addresses the problem.

Looking at the current state of cnidarium, I find that they changed their naming in cndiarium_component::ActionHandler to the following:
check_stateless (as before)
check_historical (where they explicitly highlight the issue we are facing: time-of-check-to-time-of-use: https://github.com/penumbra-zone/penumbra/blob/6981aec13519d352be642f690350b5227657de4f/crates/cnidarium-component/src/action_handler.rs#L56-L59)
check_and_execute (where now stateful check && execution happen atomically, taking into account serial execution)

PR #1341, opened 2024-08-02
feat(sequencer)!: perform check tx like regular block building

SuperFluffy:
## Summary
This patch updates sequencer's mempool service to execute the same steps when serving an ABCI `CheckTx` request as it would during regular block building.

## Background
The current check-tx implementation was only performing a subset of the checks that would be performed during regular block construction (specifically, during  finalize-block). This means that while a check-tx would return a positive response, a subsequent and immediate execution would fail. Furthermore, with the changes in https://github.com/astriaorg/astria/issues/1318 the distinction between checks and execution has been lifted, moving the current check-tx implementation further away from actual business logic.

With this patch the check-tx logic now uses exactly the same code paths as for regular block construction.

This is done by creating a new `App` instance with a snapshot that never gets applied to the actual storage.

## Changes
- Rename `App::execute_transaction` to `App::deliver_tx`: this moves the nomenclature closer to cometbft (and penumbra)
- Introduce `App::deliver_tx_bytes` as a wrapper to the now private `App:deliver_tx`.
- Rewrite `crate::service::Mempool` in terms of `App::execute_transaction_bytes`
- Move checks for total signed transaction size to `App::deliver_tx_bytes`
- Change `crate::mempool::Mempool` to require `Arc<SignedTransaction>` in its methods: this is a relatively mild change because it wrapped its argument in an `Arc` anyway. This is useful because `App::execute_transaction_bytes` returns a `Arc<SignedTransacttion>`
- Drop metrics we no longer have access to (because transactionq checking now happens in a single opaque method)
- Provide convenience methods to downcast dynamic errors to types when possible and return abci responses
- Remove obsolete functions in `transaction::checks`

## Testing
ChecTx is now running the same logic as the rest of the app, which reduces the need for extra tests for the mempool service.

Tests were added for the following scenarios:
1. A valid transaction makes it into the mempool.
2. A transaction with a nonce ahead of the current one makes it into the mempool.

More test coverage is desirable but left to a future PR.

## Metrics
**Removed** (at present, we no longer have direct access to these)
- `CHECK_TX_REMOVED_FAILED_STATELESS`
- `CHECK_TX_REMOVED_STALE_NONCE`
- `CHECK_TX_REMOVED_ACCOUNT_BALANCE`
- `CHECK_TX_DURATION_SECONDS` with the following labels:
    - `CHECK_TX_STAGE => "stateless check"`
    - `CHECK_TX_STAGE => "nonce check"`
    - `CHECK_TX_STAGE => "chain id check"`
    - `CHECK_TX_STAGE => "balance check"`

**Added** (this is a coarser metrics that effectively encapsulates those above):
- `CHECK_TX_REMOVED_SPECULATIVE_DELIVER_TX`
- `CHECK_TX_DURATION_SECONDS` label:

**Renamed** (describes better what happened and contrasting to the speculative metric above):
- `CHECK_TX_REMOVED_FAILED_EXECUTION` to `CHECK_TX_REMOVED_FAILED_DELIVER_TX` 

## Breaking Changelist
- This is a breaking change because two different versions of sequencer will end up with different CheckTx results.

## Related Issues
Closes https://github.com/astriaorg/astria/issues/1340


crates/astria-sequencer/src/service/mempool.rs
```diff
@@ -280,21 +208,24 @@ async fn handle_check_tx<S: accounts::StateReadExt + address::StateReadExt + 'st
...
 
     metrics
         .record_check_tx_duration_seconds_insert_to_app_mempool(finished_check_removed.elapsed());
-    metrics.record_actions_per_transaction_in_mempool(actions_count);
-    metrics.record_transaction_in_mempool_size_bytes(tx_len);
+    metrics.record_actions_per_transaction_in_mempool(the_tx.actions().len());
+    metrics.record_transaction_in_mempool_size_bytes(bytes.len());
     metrics.set_transactions_in_mempool_total(mempool_len);
 
     response::CheckTx::default()
```
Lilyjjo, line 231: This file is really undertested in general. I think that at least transactions that fail execution due to too high of a nonce should be let into the mempool. There could be an argument that only stateful/stateless checks should be checked here and not execution, most mempools only filter against nonce and balance. We're different because we don't let failing transactions into blocks, but I'm a little concerned that requiring execution success is too strict.

Here's a test that should work with this PR showing the too high nonce failure: 

```suggestion
    response::CheckTx::default()
}
    
#[cfg(test)]
mod test {
    use std::sync::Arc;

    use astria_core::{
        crypto::{
            SigningKey,
            VerificationKey,
        },
        protocol::transaction::v1alpha1::{
            action::TransferAction,
            TransactionParams,
            UnsignedTransaction,
        },
        sequencer::{
            Account,
            AddressPrefixes,
            UncheckedGenesisState,
        },
    };
    use prost::Message as _;
    use rand::rngs::OsRng;
    use tendermint::abci::Code;

    use super::*;
    use crate::{
        app::test_utils::default_fees,
        mempool::Mempool,
        metrics::Metrics,
        test_utils::astria_address,
    };

    async fn new_setup(funded_key: Option<VerificationKey>) -> (Storage, Mempool) {
        let accounts = if funded_key.is_some() {
            vec![Account {
                address: crate::test_utils::astria_address(&funded_key.unwrap().address_bytes()),
                balance: 10u128.pow(19),
            }]
        } else {
            vec![]
        };
        let genesis_state = UncheckedGenesisState {
            accounts,
            address_prefixes: AddressPrefixes {
                base: crate::test_utils::ASTRIA_PREFIX.into(),
            },
            authority_sudo_address: crate::test_utils::astria_address(&[0; 20]),
            ibc_sudo_address: crate::test_utils::astria_address(&[0; 20]),
            ibc_relayer_addresses: vec![],
            native_asset_base_denomination: crate::test_utils::nria(),
            ibc_params: penumbra_ibc::params::IBCParameters::default(),
            allowed_fee_assets: vec!["nria".parse().unwrap()],
            fees: default_fees(),
        }
        .try_into()
        .unwrap();

        let storage = cnidarium::TempStorage::new().await.unwrap();
        let snapshot = storage.latest_snapshot();
        let mempool = Mempool::new();
        let metrics = Box::leak(Box::new(Metrics::new()));
        let mut app = App::new(snapshot, mempool.clone(), metrics).await.unwrap();
        app.init_chain(storage.clone(), genesis_state, vec![], "test".to_string())
            .await
            .unwrap();
        app.commit(storage.clone()).await;

        (storage.clone(), mempool)
    }

    #[tokio::test]
    async fn handle_check_tx_future_nonce_ok() {
        let signing_key_0 = SigningKey::new(OsRng);
        let signing_key_1 = SigningKey::new(OsRng);
        let (storage, mempool) = new_setup(Some(signing_key_0.verification_key())).await;

        let metrics = Box::leak(Box::new(Metrics::new()));

        // make valid transfer with higher nonce
        let value = 333_333;
        let tx = UnsignedTransaction {
            params: TransactionParams::builder()
                .nonce(1)
                .chain_id("test")
                .build(),
            actions: vec![
                TransferAction {
                    to: astria_address(&signing_key_1.address_bytes()),
                    amount: value,
                    asset: crate::test_utils::nria().into(),
                    fee_asset: crate::test_utils::nria().into(),
                }
                .into(),
            ],
        };

        let signed_tx = Arc::new(tx.into_signed(&signing_key_0));

        let request = request::CheckTx {
            tx: <astria_core::protocol::transaction::v1alpha1::SignedTransaction as Clone>::clone(
                &signed_tx,
            )
            .into_raw()
            .encode_to_vec()
            .into(),
            kind: request::CheckTxKind::New,
        };

        let res = handle_check_tx(request, storage, mempool, metrics).await;

        assert!(matches!(res.code, Code::Ok), "higher nonce should be ok");
    }
```

SuperFluffy: Check-tx is now trying to downcast the errors to `InvalidNonce`: if `InvalidNonce` is found, and if it's due to the nonce being too far ahead, the transaction is still put into the mempool.

joroshiba:
I don't think this is the direction we want to go in, but some of what this achieves should be handled. Functionally for performance reasons with the new mempool design we want to have different levels of checks, to avoid DDOS at entry:

1.  allowed into mempool (nonce can be eventually valid, chainid, basic structure)
2. move to pending mempool (nonce is currently valid/valid with others in pending, adequate account balance)
3. execution (if fails evicted from mempool)

I believe some of this requires some memoization across the transaction since transactions are atomic, we need each action to be statically valid but also need the set to be valid. This requires some consideration across the transaction especially with regard to balances. There are other potential failures within a multi action transaction, but this is the primary concern. 

For some edge cases around changing sudo accounts the solution is likely to disallow those actions from being in a bundle.

joroshiba:
closed in 7 days.

joroshiba:

PR #1343, opened 2024-08-02
fix(sequencer)!: fix block fees

### LINKED ISSUE #1333: Proposer is not paid correct block fees
Block fees are transferred to the proposer here:
https://github.com/astriaorg/astria/blob/c765408c97442549db04f7a8c46dd486e4f2be21/crates/astria-sequencer/src/app/mod.rs#L1064-L1076
And block fees are updated here:
https://github.com/astriaorg/astria/blob/c765408c97442549db04f7a8c46dd486e4f2be21/crates/astria-sequencer/src/assets/state_ext.rs#L197-L199
But `get_and_increase_block_fees()` is only called in Transfer and Sequence actions:
https://github.com/astriaorg/astria/blob/c765408c97442549db04f7a8c46dd486e4f2be21/crates/astria-sequencer/src/sequence/action.rs#L70
https://github.com/astriaorg/astria/blob/c765408c97442549db04f7a8c46dd486e4f2be21/crates/astria-sequencer/src/accounts/action.rs#L121
`get_and_increase_block_fees()` needs to be added to the other actions with associated fees so the proposer is paid the correct amount.

ethanoroshiba:
## Summary
Corrected block fee reporting so that proposer is paid the correct fees.

## Background
Previously, only `TransferAction` and `SequenceAction` called `get_and_increase_block_fees()`, the method for increasing the running fees for the block. In `end_block`, this was used for calculating how much to increase the proposer's balance meaning that they were only paid the fees for `Transfer` and `Sequence` actions.

## Changes
- Implemented `get_and_increase_block_fees()` in all actions which charge fees.

## Testing
Added unit tests for each action, with the exception of `Ics20Withdrawal`. This test should be implemented, but will require a more in-depth configuration.

## Breaking Changelist
- App hash changed, snapshot has been regenerated.

## Related Issues
closes #1333


SuperFluffy (dismissed):
<del>I'd like to take this PR into a different direction:

Instead of implementing fee-payments on a per-action basis, I would fee payments to happen at a higher level in `<SignedTransaction as ActionHandler>::check_and_execute`.

I am thinking of something like this:

1. define a `trait ActionFee` with a trait method `ActionFee::action_fee`
2. implement this for every action and the `enum Actions` collection.
3. define a `calculate_fees` method on `SignedTransaction` (note that this is a foreign type, so you might either need to make a free-standing function or create a trait just for that); `calculate_fees` just runs over all actions and sums up their fees.
4. pay the fee right before `SignedTransaction::check_and_execute` starts processing the individual transactions.

I know that fee calculations right now require access to the state. Ideally the fee calculations on an individual action can do without access to state.</del>


EDIT: Having had a second look at how fees are calculated, we use a base fee (that's fine) but across potentially many different `fee_assets` (not fine).

IMO this makes my suggestion above untenable, so that fees indeed must be paid on a per-action basis.

Note that the reason why this cannot be done prior to execution is the same as what required to fix #1318: fees cannot be paid before execution begins. Following scenario:

1. action A adds a new fee asset
2. action B adds funds to an account for that fee asset
3. action C uses the fee asset for that account to pay for the transaction.

crates/astria-sequencer/src/app/tests_execute_transaction.rs
```diff
@@ -1100,3 +1109,234 @@ async fn transaction_execution_records_fee_event() {
...
+        actions,
+    };
+    let signed_tx = Arc::new(tx.into_signed(&alice));
+    app.execute_transaction(signed_tx).await.unwrap();
+
+    let total_block_fees: u128 = app
+        .state
+        .get_block_fees()
+        .await
+        .expect("failed to get block fees")
```
noot, line 1146:
```suggestion
        .expect("can get block fees")
```

`expect` is used with the success case statement, ie. what you expect to happen. also, in this case, you can probably just use `unwrap` anyways, as it's clear what the issue is if `get_block_fees` fails

crates/astria-sequencer/src/app/tests_execute_transaction.rs
```diff
@@ -1100,3 +1109,234 @@ async fn transaction_execution_records_fee_event() {
...
+        .into_iter()
+        .map(|(_, fee)| fee)
+        .sum();
+    assert_eq!(total_block_fees, 1);
+}
+
+// TODO: Add test to ensure correct block fees for ICS20 withdrawal
```
noot, line 1342: will you add that here?

crates/astria-sequencer/src/app/tests_execute_transaction.rs
```diff
@@ -1100,3 +1109,234 @@ async fn transaction_execution_records_fee_event() {
...
+#[tokio::test]
+async fn ensure_correct_block_fees_bridge_lock() {
+    let alice = get_alice_signing_key();
+    let bridge = get_bridge_signing_key();
+    let bridge_address = astria_address(&bridge.address_bytes());
+    let rollup_id = RollupId::from_unhashed_bytes(b"testchainid");
+
+    let mut app = initialize_app(None, vec![]).await;
+    let mut state_tx = StateDelta::new(app.state.clone());
+    state_tx.put_transfer_base_fee(1).unwrap();
```
noot, line 1245:
```suggestion
    let transfer_base_fee = 1;
    state_tx.put_transfer_base_fee(transfer_base_fee).unwrap();
```

noot (approved):

crates/astria-sequencer/src/app/tests_execute_transaction.rs
ethanoroshiba: I attempted to do so but ran into a lot of difficulty setting up the IBC channel/ports, which is required for `check_and_execute()`. It looks like Janis ran into the same issue in #1345 because Penumbra doesn't make the `StateWriteExt` traits public. I may be missing something, though. If you have any suggestions it would be greatly appreciated! Also, thanks for the quick review and welcome back!

crates/astria-sequencer/src/ibc/ics20_withdrawal.rs
```diff
@@ -10,6 +10,7 @@ use astria_core::{
         Address,
     },
     protocol::transaction::v1alpha1::action,
+    Protobuf,
```
SuperFluffy, line 13: `Protobuf as _`

crates/astria-sequencer/src/bridge/bridge_lock_action.rs
```diff
@@ -9,6 +9,7 @@ use astria_core::{
         TransferAction,
     },
     sequencerblock::v1alpha1::block::Deposit,
+    Protobuf,
```
SuperFluffy, line 12: `Protobuf as _`

crates/astria-sequencer/src/app/tests_execute_transaction.rs
SuperFluffy: Yeah, gotta make a PR against penumbra.

crates/astria-sequencer/src/bridge/init_bridge_account_action.rs
```diff
@@ -98,7 +102,9 @@ impl ActionHandler for InitBridgeAccountAction {
             from,
             self.withdrawer_address.map_or(from, Address::bytes),
         );
-
+        state
+            .get_and_increase_block_fees(&self.fee_asset, fee, Self::full_name())
+            .await?;
```
SuperFluffy, line 107: context

crates/astria-sequencer/src/bridge/init_bridge_account_action.rs
```diff
@@ -17,7 +18,10 @@ use crate::{
     },
     address::StateReadExt as _,
     app::ActionHandler,
-    assets::StateReadExt as _,
+    assets::{
+        StateReadExt as _,
+        StateWriteExt,
```
SuperFluffy, line 23: import as `_`

crates/astria-sequencer/src/bridge/bridge_lock_action.rs
```diff
@@ -23,6 +24,7 @@ use crate::{
     },
     address::StateReadExt as _,
     app::ActionHandler,
+    assets::StateWriteExt,
```
SuperFluffy, line 27: `as  _`

crates/astria-sequencer/src/app/tests_execute_transaction.rs
SuperFluffy, line 1: Can you group the block fee tests into a separate module?

crates/astria-sequencer/src/app/tests_execute_transaction.rs
joroshiba: If we are leaving a TODO in after merge, create a follow up issue and link in code please :) ie

``` 
// TODO(https://github.com/astriaorg/astria/issues/<issue num>): add test for correct block fees
```

crates/astria-sequencer/src/app/tests_execute_transaction.rs
ethanoroshiba: Moved to new module in fbc99f0.

SuperFluffy (approved):
Looks great. We should restructure the tests so we don't end up with a plethora of `tests_*` but rather create a module structure. But that can be done in a followup.

Issue #1308, opened 2024-07-29
Ics20Withdrawal should reduce balance of bridge account when signed by withdrawer_address


joroshiba:
bug found by auditors:

> According to your bridge docs, withdrawer address have permission to withdraw on behalf of bridge address. It seems the Ics20Withdrawal action check if transaction sender is withdrawer address in check_stateful. But in execute, it decreases balance of sender from, which is not bridge address. This mean Ics20Withdrawal will try to decrease balance of withdrawer address

PR #1344, opened 2024-08-02
fix(sequencer)!: take funds from bridge in ics20 withdrawals

SuperFluffy:
## Summary
Fixes ics20 withdrawals by actually taking funds from the provided bridge address.

## Background
Prior to this change, funds were always taken from the signer of the `astria.protocol.v1alpha1.Ics20Withdrawal` action, even if its `bridge_address` field was set.

## Changes
- Uses `astria.protocol.v1alpha1.Ics20Withdrawal.bridge_address` to take funds in an ics20 withdrawal if set (and if it passes bridge checks)
- Remove unnecessary state accesses to check if enough funds are present: just try to decrease them. If they fail, then the entire action fails.

## Testing
All present tests have been updated to check if the correct withdrawal target is found.

There should be tests for the entire `Ics20Withdrawal::check_and_execute` flow. However, adding an ibc channel and port does not seem to be currently possible because `penumbra_ibc` does not export the channel `StateWriteExt` trait. This is left to a followup.

## Breaking Changelist
This is a breaking change because two different versions of sequencer will come to different conclusions on which account to take funds from.

## Related Issues
Closes https://github.com/astriaorg/astria/issues/1308

### DIFF
--- a/crates/astria-sequencer/src/ibc/ics20_withdrawal.rs
+++ b/crates/astria-sequencer/src/ibc/ics20_withdrawal.rs
@@ -57,13 +57,6 @@ fn withdrawal_to_unchecked_ibc_packet(
     )
 }
 
-// bridge address checks:
-// - if the sender of this transaction is not a bridge account, and the tx `bridge_address` field is
-//   None, don't need to do any bridge related checks as it's a normal user withdrawal.
-// - if the sender of this transaction is a bridge account, and the tx `bridge_address` field is
-//   None, check that the withdrawer address is the same as the transaction sender.
-// - if the tx `bridge_address` field is Some, check that the `bridge_address` is a valid bridge,
-//   and check that the withdrawer address is the same as the transaction sender.
 /// Establishes the withdrawal target.
 ///
 /// The function returns the following addresses under the following conditions:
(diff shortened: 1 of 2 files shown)


Fraser999 (approved):

crates/astria-sequencer/src/ibc/ics20_withdrawal.rs
```diff
@@ -59,30 +57,36 @@ fn withdrawal_to_unchecked_ibc_packet(
...
 }
 
-async fn ics20_withdrawal_check_stateful_bridge_account<S: StateRead>(
+// bridge address checks:
+// - if the sender of this transaction is not a bridge account, and the tx `bridge_address` field is
+//   None, don't need to do any bridge related checks as it's a normal user withdrawal.
+// - if the sender of this transaction is a bridge account, and the tx `bridge_address` field is
+//   None, check that the withdrawer address is the same as the transaction sender.
+// - if the tx `bridge_address` field is Some, check that the `bridge_address` is a valid bridge,
+//   and check that the withdrawer address is the same as the transaction sender.
```
noot, line 66: move this to above `check_and_execute`?

noot (approved):

crates/astria-sequencer/src/ibc/ics20_withdrawal.rs
SuperFluffy: This was shuffling around code. I forgot to remove it.

PR #1346, opened 2024-08-04
Genesis as proto

SuperFluffy:
## Summary

## Background

## Changes

## Testing
How are these changes tested?

## Metrics

## Breaking Changelist

## Related Issues

closes


proto/protocolapis/astria/protocol/genesis/v1alpha1/types.proto
```diff
@@ -0,0 +1,47 @@
+syntax = "proto3";
+
+package astria.protocol.genesis.v1alpha1;
+
+import "astria/primitive/v1/types.proto";
+
+message GenesisAppState {
+  string chain_id = 1;
```
SuperFluffy, line 8: This field is in preparation of a follow up refactoring sequencer to run `init-chain` with only the data provided by `GenesisAppState`.

Fraser999 (approved):
One tiny nitpick, otherwise LGTM.

charts/sequencer/files/cometbft/config/genesis.json
```diff
@@ -1,42 +1,42 @@
...
+    "addressPrefixes": {
       "base": "{{ .Values.genesis.addressPrefixes.base }}"
     },
     "accounts": [
       {{- range $index, $value := .Values.genesis.genesisAccounts }}
       {{- if $index }},{{- end }}
       {
         "address": {{ include "sequencer.address" $value.address }},
-        "balance": {{ toString $value.balance | replace "\"" "" }}
+        "balance": {{ include "sequencer.fee" ( toString $value.balance | replace "\"" "" ) }}
```
joroshiba, line 33: This change disallows some hight initial balances that were previously possible. Limited to uint64 balances effectively.

charts/sequencer/templates/_helpers.tpl
```diff
@@ -71,3 +71,7 @@ name: {{ .Values.moniker }}-sequencer-metrics
 {{/* New sequencer address */}}
 {{- define "sequencer.address"}}{ "bech32m": "{{ . }}" }
 {{- end }}
+
+{{/* uint64 fee converted to a astria proto Uint128 with only lo set */}}
+{{- define "sequencer.fee"}}{ "lo": {{ . }} }
```
joroshiba, line 76: I don't know that this helper name is accurate. `sequencer.toUint128Proto`?

proto/protocolapis/astria/protocol/genesis/v1alpha1/types.proto
joroshiba: With this change we wouldn't additionally add a new `chainId` field into the genesis `app_state` would we?

charts/sequencer/files/cometbft/config/genesis.json
joroshiba, line 1: I'll note that while I'm generally in favor of this change. I'm not a fan of the genesis file having camelCase for the info in `app_state` field and snake_case elsewhere. Is there anyway to have pbjson treat this one file differently? (I know camelCase is pbjson standard) 

According to the proto json standard: 

> Proto3 JSON parsers are required to accept both the converted lowerCamelCase name and the proto field name.

Maybe pbjson doesn't support that? I know it will default serialize this way but that doesn't seem to mean we need the inconsistency in the json that gets deserialized?

charts/sequencer/files/cometbft/config/genesis.json
SuperFluffy: Yes indeed. For our current use cases this is fine. A followup should fix this. I wonder if there is a way to call out to a helper script using helm?

charts/sequencer/files/cometbft/config/genesis.json
SuperFluffy: I dug through the code and you are right (thanks for flagging this, I wasn't aware): pbjson indeed allows reading lowerCamelCase and proto-name, for example:
```
"authoritySudoAddress" | "authority_sudo_address" => Ok(GeneratedField::AuthoritySudoAddress),
```

However, serialization is really annoying because there is no way to override this at a higher level for all child-types (fields) of a type.

charts/sequencer/templates/_helpers.tpl
SuperFluffy: Yeah, we'd ultimately end up doing that. I had a dislike for extra configuration data being passed in "from the outside". I would like all genesis settings relevant to the app/sequencer to be in one place.

We would then either enforce that the cometbft chain ID match the sequencer chain ID - or ignore cometbft.

crates/astria-core/src/protocol/genesis/v1alpha1.rs
```diff
@@ -0,0 +1,778 @@
...
+}
+
+#[derive(Debug, thiserror::Error)]
+#[error("failed ensuring invariants of {}", GenesisAppState::full_name())]
+enum GenesisAppStateErrorKind {
+    #[error("`accounts` field was invalid")]
+    Accounts { source: AccountError },
+    #[error("`address_prefixes` field was invalid")]
+    AddressPrefixes { source: AddressPrefixesError },
+    #[error("one or more addresses did not match the provided base prefix")]
```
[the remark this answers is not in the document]
Fraser999: Just seems odd using plural when we only provide a single address in the variant.

Not a big deal though - I'd stick with `one or more addresses` over `given addresses`, since we're not giving more than one address in the error :)

charts/sequencer/templates/_helpers.tpl
joroshiba: I don't think your comment is on the correct note here?

charts/sequencer/files/cometbft/config/genesis.json
joroshiba: We could have an init script which does the generation.

charts/sequencer/templates/_helpers.tpl
SuperFluffy: ODd, yeah. This was supposed to go on another comment. I remember explicitly putting it there.. :-/

charts/sequencer/files/cometbft/config/genesis.json
joroshiba: I think as long as deserialization is good it's not a major issue. The way we do it at least genesis has been hand built/built via helm. 

If we create something like the cosmos cli tool to fully build the genesis from a series of command line calls more complicated. I don't care so much if dynamically generated genesis files are ugly as the ones committed in code are uniform. Kinda nitty though.

proto/protocolapis/astria/protocol/genesis/v1alpha1/types.proto
SuperFluffy: Yeah, we'd ultimately end up doing that. I had a dislike for extra configuration data being passed in "from the outside". I would like all genesis settings relevant to the app/sequencer to be in one place.

We would then either enforce that the cometbft chain ID match the sequencer chain ID - or ignore cometbft.

joroshiba: The chain id is already part of the `InitChain` abci call, so this would be breaking expected abci conventions: https://github.com/cometbft/cometbft/blob/main/spec/abci/abci++_methods.md#initchain

This is inherently already a part of the call, I wouldn't want to break the abci convention here.

SuperFluffy: Agree, in my mind these fields would just be duplicates: sequencer would check if its genesis contains the same chain ID that cometbft reports.

crates/astria-core/src/protocol/genesis/v1alpha1.rs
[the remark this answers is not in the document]
SuperFluffy: I changed it to `"one of the provided addresses ..."`:  7569ca74

charts/sequencer/files/cometbft/config/genesis.json
SuperFluffy: I updated the genesis json to be snake_case (actually, to match the protobuf field names, which happen to be snake case): [60a5022](https://github.com/astriaorg/astria/pull/1346/commits/60a5022a7fb626540e5fdef10be0688137234385)

charts/sequencer/files/cometbft/config/genesis.json
SuperFluffy: I have created https://github.com/astriaorg/astria/issues/1388 as a followup.

charts/sequencer/templates/_helpers.tpl
SuperFluffy: Renamed the helper in [dfdff4d](https://github.com/astriaorg/astria/pull/1346/commits/dfdff4db0976afd50485d2d8b523f86526929987)

joroshiba (approved):
Approving, with assumption as discussed offline that the test fixes for astria-cli are moved into seperate PR before merge.

SuperFluffy:
> Approving, with assumption as discussed offline that the test fixes for astria-cli are moved into seperate PR before merge.

Created a followup in https://github.com/astriaorg/astria/issues/1392

PR #1361, opened 2024-08-12
refactor(composer): streamline geth and executor `run_until_stopped`

ethanoroshiba:
## Summary
Shortened and streamlined geth and executor `run_until_stopped()` functions to get rid of clippy exceptions and ensure events are emitted within spans.

## Background
#1326 removed instrumentation on these `run_until_stopped()` functions, which revealed a clippy warning for too many lines. Additionally, this made it so that logging inside these functions would not be emitted within any span.

## Changes
- Delegated many tasks to helper functions with instrumentation.
- Created new `utils` module to house a shared `report_exit_reason()` function.
- Moved `ensure_chain_id_is_correct()` and `get_latest_nonce()` to `init()` (previosuly `pre_run_checks()`)

## Testing
Passing all tests.

## Related Issues
Part of #1321

### DIFF
--- a/crates/astria-composer/src/collectors/geth.rs
+++ b/crates/astria-composer/src/collectors/geth.rs
@@ -147,7 +147,7 @@ impl Geth {
 
     /// Starts the collector instance and runs until failure or until
     /// explicitly closed
-    // Allow: Refactored in <insert GH here>. Current logging events will show as errors in otel,
+    // Allow: Refactored in https://github.com/astriaorg/astria/pull/1361. Current logging events will show as errors in otel,
     // refactor addresses this as well.
     #[allow(clippy::too_many_lines)]
     pub(crate) async fn run_until_stopped(self) -> eyre::Result<()> {
@@ -169,10 +169,10 @@ impl Geth {
             .geth_txs_received(&chain_name)
             .cloned()
             .unwrap_or_else(|| {
-                // error!(
-                //     rollup_chain_name = %chain_name,
-                //     "failed to get geth transactions_received counter"
-                // );
+                error!(
+                    rollup_chain_name = %chain_name,
+                    "failed to get geth transactions_received counter"
+                );
                 Counter::noop()
             });
         let txs_dropped_counter = metrics
--- a/crates/astria-composer/src/collectors/geth.rs
+++ b/crates/astria-composer/src/collectors/geth.rs
@@ -60,7 +60,7 @@ use crate::{
         Handle,
     },
     metrics::Metrics,
-    utils::report_exit,
+    utils::log_exit,
 };
 
 type StdError = dyn std::error::Error;
@@ -218,7 +218,7 @@ impl Geth {
             }
         };
 
-        report_exit(&reason);
+        log_exit(&reason);
 
         status.send_modify(|status| status.is_connected = false);
 
--- a/crates/astria-composer/src/collectors/geth.rs
+++ b/crates/astria-composer/src/collectors/geth.rs
@@ -285,8 +285,8 @@ async fn unsubscribe_wss_stream_handler(tx_stream: &SubscriptionStream<'_, Ws, T
         Ok(Err(err)) => {
             error!(error = %Report::new(err), "failed unsubscribing from the geth tx stream");
         }
-        Err(err) => {
-            error!(error = %Report::new(err), "timed out while unsubscribing from the geth tx stream");
+        Err(_) => {
+            error!("timed out while unsubscribing from the geth tx stream");
         }
     }
 }
--- a/crates/astria-composer/src/collectors/geth.rs
+++ b/crates/astria-composer/src/collectors/geth.rs
@@ -225,7 +225,7 @@ impl Geth {
         // if the loop exits with an error, we can still proceed with unsubscribing the WSS
         // stream as we could have exited due to an error in sending messages via the executor
         // channel.
-        unsubscribe_wss_stream_handler(&tx_stream).await;
+        unsubscribe_from_rollup(&tx_stream).await;
 
         reason.map(|_| ())
     }
--- a/crates/astria-composer/src/collectors/geth.rs
+++ b/crates/astria-composer/src/collectors/geth.rs
@@ -218,7 +218,7 @@ impl Geth {
             }
         };
 
-        report_exit_reason(reason.as_ref());
+        report_exit_reason(reason.as_deref());
 
         status.send_modify(|status| status.is_connected = false);
 
--- a/crates/astria-composer/src/collectors/geth.rs
+++ b/crates/astria-composer/src/collectors/geth.rs
@@ -202,7 +202,7 @@ impl Geth {
 
                         txs_received_counter.increment(1);
 
-                        if let Err(err) = executor_send_timeout_handler(
+                        if let Err(err) = forward_geth_tx(
                             &executor_handle,
                             seq_action,
                             tx_hash,
--- a/crates/astria-composer/src/collectors/geth.rs
+++ b/crates/astria-composer/src/collectors/geth.rs
@@ -28,10 +28,14 @@ use astria_eyre::eyre::{
     Report,
     WrapErr as _,
 };
-use ethers::providers::{
-    Provider,
-    ProviderError,
-    Ws,
+use ethers::{
+    providers::{
+        Provider,
+        ProviderError,
+        SubscriptionStream,
+        Ws,
+    },
+    types::Transaction,
 };
 use metrics::Counter;
 use tokio::{
@@ -43,15 +47,18 @@ use tokio::{
 };
 use tokio_util::sync::CancellationToken;
 use tracing::{
-    debug,
     error,
     info,
+    instrument,
     warn,
 };
 
 use crate::{
     collectors::EXECUTOR_SEND_TIMEOUT,
-    executor,
+    executor::{
+        self,
+        Handle,
+    },
     metrics::Metrics,
 };
 
@@ -147,9 +154,6 @@ impl Geth {
 
     /// Starts the collector instance and runs until failure or until
     /// explicitly closed
-    // Allow: Refactored in https://github.com/astriaorg/astria/pull/1361. Current logging events will show as errors in otel,
-    // refactor addresses this as well.
-    #[allow(clippy::too_many_lines)]
     pub(crate) async fn run_until_stopped(self) -> eyre::Result<()> {
         use ethers::providers::Middleware as _;
         use futures::stream::StreamExt as _;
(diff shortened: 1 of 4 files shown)


crates/astria-composer/src/executor/mod.rs
```diff
@@ -607,6 +567,76 @@ async fn submit_tx(
...
+    }
+}
+
+/// Handles timeout of shutdown process
+#[instrument(skip_all)]
+async fn bundle_drain_timout_handler(shutdown_logic: impl Future<Output = eyre::Result<()>>) {
+    match tokio::time::timeout(BUNDLE_DRAINING_DURATION, shutdown_logic).await {
+        Ok(Ok(())) => info!("executor shutdown tasks completed successfully"),
+        Ok(Err(error)) => error!(%error, "executor shutdown tasks failed"),
+        Err(error) => error!(%error, "executor shutdown tasks failed to complete in time"),
```
Fraser999, line 591: The tokio error here is an `Elapsed` instance, which [doesn't give any more useful info than our own string](https://github.com/tokio-rs/tokio/blob/694577fa8563dfe62878a4b3b4879a2199e3c409/tokio/src/time/error.rs#L111-L115).
```suggestion
        Err(error) => error!("executor shutdown tasks failed to complete in time"),
```

crates/astria-composer/src/executor/mod.rs
```diff
@@ -607,6 +567,76 @@ async fn submit_tx(
...
+#[instrument(skip_all, err)]
+fn process_result_update_nonce(nonce: &mut u32, rsp: eyre::Result<u32>) -> eyre::Result<()> {
+    match rsp {
+        Ok(new_nonce) => {
+            *nonce = new_nonce;
+            Ok(())
+        }
+        Err(error) => {
+            error!(%error, "failed submitting bundle to sequencer; aborting executor");
+            Err(error).wrap_err("failed submitting bundle to sequencer")
```
Fraser999, line 605:
```suggestion
            Err(error).wrap_err("failed submitting bundle to sequencer; aborting executor")
```

crates/astria-composer/src/collectors/geth.rs
```diff
@@ -241,64 +201,154 @@ impl Geth {
...
+    // this hang for too long
+    match tokio::time::timeout(WSS_UNSUBSCRIBE_TIMEOUT, tx_stream.unsubscribe()).await {
+        Ok(Ok(true)) => info!("unsubscribed from geth tx stream"),
+        Ok(Ok(false)) => warn!("failed to unsubscribe from geth tx stream"),
+        Ok(Err(err)) => {
+            error!(error = %Report::new(err), "failed unsubscribing from the geth tx stream");
+        }
+        Err(err) => {
+            error!(error = %Report::new(err), "timed out while unsubscribing from the geth tx stream");
+        }
```
Fraser999, line 290:
```suggestion
        Err(_) => {
            error!("timed out while unsubscribing from the geth tx stream");
        }
```

Fraser999 (commented):
A few nitpicks, but the moved logic all seems to be unchanged.

crates/astria-composer/src/executor/mod.rs
```diff
@@ -607,6 +567,76 @@ async fn submit_tx(
...
+        }
+    }
+}
+
+/// Checks if all bundles have been drained, warning if not
+#[instrument(skip_all)]
+fn check_if_bundles_drained(
```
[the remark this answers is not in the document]
ethanoroshiba: Name changed in 9961605

crates/astria-composer/src/executor/mod.rs
```diff
@@ -607,6 +567,76 @@ async fn submit_tx(
...
+/// Reports and logs exit reason
+#[instrument(skip_all, err)]
+fn report_exit(reason: &eyre::Result<&str>) -> eyre::Result<()> {
+    match &reason {
+        Ok(reason) => {
+            info!(reason, "shutting down");
+            Ok(())
+        }
+        Err(reason) => {
+            error!(%reason, "shutting down");
```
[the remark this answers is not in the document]
ethanoroshiba: Error log removed in 9961605. Thanks for the review!

Fraser999 (approved):

SuperFluffy (changes requested):
Looks mosly fine except for some oddities.

Requesting changes because an early return in the select loop skips shutdown logic.

crates/astria-composer/src/executor/mod.rs
```diff
@@ -607,6 +557,72 @@ async fn submit_tx(
     res
 }
 
+/// Reports and logs exit reason
+#[instrument(skip_all, err)]
+fn report_exit(reason: &eyre::Result<&str>) -> eyre::Result<()> {
```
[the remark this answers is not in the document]
ethanoroshiba: Created new `utils` module with this functionality. I also renamed the function to `log_exit`, since I thought this was more accurate and less likely to be confused with `composer::report_exit` which functions slightly differently.

crates/astria-composer/src/executor/mod.rs
```diff
@@ -212,15 +212,11 @@ impl Executor {
...
-    // Allow: Refactored in <insert GH here>. Current logging events will show as errors in otel,
-    // refactor addresses this as well.
-    #[allow(clippy::too_many_lines)]
     pub(super) async fn run_until_stopped(mut self) -> eyre::Result<()> {
         select!(
             biased;
             () = self.shutdown_token.cancelled() => {
-                info!("received shutdown signal while running initialization routines; exiting");
-                return Ok(());
+                return report_exit(&Ok("received shutdown signal while running initialization routines; exiting"));
```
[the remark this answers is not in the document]
ethanoroshiba: Added exit logging for other instances before the select loop is entered in e656f42.

crates/astria-composer/src/executor/mod.rs
```diff
@@ -256,13 +252,7 @@ impl Executor {
...
                 // process submission result and update nonce
                 rsp = &mut submission_fut, if !submission_fut.is_terminated() => {
-                    match rsp {
-                        Ok(new_nonce) => nonce = new_nonce,
-                        Err(error) => {
-                            error!(%error, "failed submitting bundle to sequencer; aborting executor");
-                            break Err(error).wrap_err("failed submitting bundle to sequencer");
-                        }
-                    }
+                    process_result_update_nonce(&mut nonce, rsp)?;
```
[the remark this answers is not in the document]
ethanoroshiba: Used break and also moved `block_timer.reset` inside the handler in e656f42

crates/astria-composer/src/collectors/geth.rs
```diff
@@ -241,64 +202,142 @@ impl Geth {
...
-            }
-        };
+        log_exit(&reason);
 
         status.send_modify(|status| status.is_connected = false);
 
         // if the loop exits with an error, we can still proceed with unsubscribing the WSS
         // stream as we could have exited due to an error in sending messages via the executor
         // channel.
+        unsubscribe_wss_stream_handler(&tx_stream).await;
```
SuperFluffy, line 228: This is overly descriptive while providing very limited semantic meaning to the reader:
```suggestion
        unsubscribe_from_rollup(&tx_stream).await;
```

crates/astria-composer/src/collectors/geth.rs
```diff
@@ -241,64 +202,142 @@ impl Geth {
...
-                            Err(SendTimeoutError::Closed(_seq_action)) => {
-                                warn!(
-                                    transaction.hash = %tx_hash,
-                                    "executor channel closed while sending transaction; dropping transaction \
-                                     and exiting event loop"
-                                );
-                                txs_dropped_counter.increment(1);
-                                break Err(eyre!("executor channel closed while sending transaction"));
-                            }
+                        if let Err(err) = executor_send_timeout_handler(
```
SuperFluffy, line 205: Easier to read:
```suggestion
                        if let Err(err) = forward_geth_tx(
```

crates/astria-composer/src/collectors/geth.rs
```diff
@@ -241,64 +202,142 @@ impl Geth {
...
+}
 
-        reason.map(|_| ())
+#[instrument(skip_all)]
+async fn unsubscribe_wss_stream_handler(tx_stream: &SubscriptionStream<'_, Ws, Transaction>) {
+    // give 2s for the websocket connection to be unsubscribed as we want to avoid having
+    // this hang for too long
+    match tokio::time::timeout(WSS_UNSUBSCRIBE_TIMEOUT, tx_stream.unsubscribe()).await {
+        Ok(Ok(true)) => info!("unsubscribed from geth tx stream"),
+        Ok(Ok(false)) => warn!("failed to unsubscribe from geth tx stream"),
```
SuperFluffy, line 273:
```
        Ok(Ok(false)) => warn!("geth responded to unsubscribe request but returned `false`"),
```

crates/astria-composer/src/executor/mod.rs
```diff
@@ -212,25 +214,26 @@ impl Executor {
...
+                return Ok(())
             }
 
             res = self.pre_run_checks() => {
+                log_exit(&res);
                 res.wrap_err("required pre-run checks failed")?;
             }
         );
         let mut submission_fut: Fuse<Instrumented<SubmitFut>> = Fuse::terminated();
         let mut nonce = get_latest_nonce(self.sequencer_client.clone(), self.address, self.metrics)
```
SuperFluffy, line 231: If we refactor this anway, please combine this `get_latest_nonce` with the `self.pre_run_checks()` above by creating something like a `Executor::init` method.

Right now `get_latest_nonce` ignores the shutdown token above and this PR looks like a good place to fix that.

crates/astria-composer/src/executor/mod.rs
```diff
@@ -212,25 +214,26 @@ impl Executor {
...
             .await
-            .wrap_err("failed getting initial nonce from sequencer")?;
+            .map_err(|err| {
+                log_exit(&Err(eyre!(err.to_string())));
```
SuperFluffy, line 234: Why turn this into a string? You can do:

```rust
get_latest_nonce().wrap_err("failed getting initial nonce from sequencer").inspect_err(|err| report_exit(err))
```

You will also need to redefine `report_exit` to be something like `fn report_exit(reason: Result<&str, &eyre::Report>)` - but then you can make use of `Result::as_ref` where necessary.

crates/astria-composer/src/executor/mod.rs
ethanoroshiba: Refactored such that `ensure_chain_id_is_correct()` and `get_latest_nonce()` are in an `Executor::init()` function, which will call `report_exit_reason()` on an Err result.

crates/astria-composer/src/executor/mod.rs
```diff
@@ -215,25 +214,18 @@ impl Executor {
     /// # Errors
     /// An error is returned if connecting to the sequencer fails.
     pub(super) async fn run_until_stopped(mut self) -> eyre::Result<()> {
-        select!(
+        let mut nonce = select!(
             biased;
             () = self.shutdown_token.cancelled() => {
-                log_exit(&Ok("received shutdown signal while running initialization routines; exiting"));
+                report_exit_reason(Ok(&"received shutdown signal while running initialization routines; exiting"));
```
SuperFluffy, line 220: `"foobar"` is of type `&str`, so you should be able to call this as `report_exit_reason(Ok("received shutdown ..."))`

crates/astria-composer/src/executor/mod.rs
```diff
@@ -293,9 +287,9 @@ impl Executor {
         // sequence actions
         self.serialized_rollup_transactions.close();
 
-        log_exit(&reason);
+        report_exit_reason(reason.as_ref());
```
SuperFluffy, line 290: I think you can do this to have a cleaner `report_exit_reason` impl:
```suggestion
        report_exit_reason(reason.as_deref());
```

crates/astria-composer/src/utils.rs
```diff
@@ -2,11 +2,9 @@ use astria_eyre::eyre::Report;
 use tracing::{
     error,
     info,
-    instrument,
 };
 
-#[instrument(skip_all)]
-pub(crate) fn report_exit_reason(reason: &Result<&str, Report>) {
+pub(crate) fn report_exit_reason(reason: Result<&&str, &Report>) {
```
SuperFluffy, line 7: Using `Result::as_deref` above this can become:
```suggestion
pub(crate) fn report_exit_reason(reason: Result<&str, &Report>) {
```

SuperFluffy (approved):
Looks great, thank you for your work. Made a suggestion to clean up `report_exit_reason`, but otherwise this is good to merge.

PR #1363, opened 2024-08-13
Refactor(charts): move celestia-node dependency up a level

### LINKED ISSUE #1362: Celestia-node chart incompatible with evm-stack
The celestia-node chart is incompatible with the new evm-stack chart break. 
The evm-stack can't access definition created within the node deployment duo to dependency level. The immediate and rational solution for this is to move the node chart to be a direct subchart of the evm-stack

┆Issue Number: ENG-710

quasystaty1:
## Summary
Moves `celestia-node` chart to be a direct dependency of the `evm-stack` 
## Background
With the new `evm-stack` wrapper chart, deploying a celestia-node failed duo to a dependency level issue.
## Changes
- moved `celestia-node` chart up a level from `evm-rollup -> evm-stack`
- now uses an HTTP server to share auth token with conductor service
- conductor script running within the pod

## Testing
- tested against a local kubernetes cluster with the latest components
- tested against devdev remote cluster, using a mocha-4 celestia-node

## Related Issues
closes #1362

### DIFF
--- a/charts/celestia-node/templates/statefulset.yaml
+++ b/charts/celestia-node/templates/statefulset.yaml
@@ -72,6 +72,24 @@ spec:
               subPath: {{ $secret.filename }}
             {{- end }}
             {{- end }}
+        {{- if .Values.config.tokenAuthLevel }}
+        - name: token-server
+          image: {{ .Values.images.tokenServer }}
+          command: [ "/bin/httpd", "-v", "-f", "-p", "{{ .Values.ports.tokenServer }}", "-h", "/celestia/token-server/" ]
+          ports:
+            - containerPort: {{ .Values.ports.tokenServer }}
+              name: token-svc
+          volumeMounts:
+            - mountPath: /celestia
+              name: {{ $label }}-vol
+              readOnly: true
+          startupProbe:
+            httpGet:
+              path: /
+              port: token-svc
+            failureThreshold: 30
+            periodSeconds: 5
+        {{- end }}
       volumes:
         # ------------ Startup scripts -------------
         - name: {{ $label }}-scripts-vol
--- a/charts/evm-rollup/templates/configmap.yaml
+++ b/charts/evm-rollup/templates/configmap.yaml
@@ -38,7 +38,7 @@ data:
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: {{ include "rollup.namespace" .  }}-conductor-scripts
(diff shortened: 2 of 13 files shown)


charts/celestia-node/templates/statefulset.yaml
```diff
@@ -72,6 +72,24 @@ spec:
               subPath: {{ $secret.filename }}
             {{- end }}
             {{- end }}
+        {{- if .Values.config.tokenAuthLevel }}
```
quasystaty1, line 75: Used for serving the auth token to conductor

charts/celestia-node/templates/statefulset.yaml
```diff
@@ -72,6 +72,24 @@ spec:
...
+            - mountPath: /celestia
+              name: {{ $label }}-vol
+              readOnly: true
+          startupProbe:
+            httpGet:
+              path: /
+              port: token-svc
+            failureThreshold: 30
+            periodSeconds: 5
+        {{- end }}
```
quasystaty1, line 92: Used for serving the auth token to conductor

charts/evm-rollup/templates/configmap.yaml
```diff
@@ -5,13 +5,12 @@ metadata:
...
-  {{- if (index .Values "celestia-node").enabled }}
-  TOKEN_SERVER_URL: "{{ include "celestiaNode.service.addresses.token" (index .Subcharts "celestia-node") }}"
-  ASTRIA_CONDUCTOR_CELESTIA_NODE_HTTP_URL: "{{ include "celestiaNode.service.addresses.rpc" (index .Subcharts "celestia-node") }}"
-  {{- else }}
+  {{- if .Values.config.celestia.node.enabled }}
   TOKEN_SERVER_URL: "{{ .Values.config.celestia.token }}"
-  ASTRIA_CONDUCTOR_CELESTIA_NODE_HTTP_URL: "{{ .Values.config.celestia.rpc }}"
+  {{- else }}
+  ASTRIA_CONDUCTOR_CELESTIA_BEARER_TOKEN: "{{ .Values.config.celestia.token }}"
   {{- end }}
```
quasystaty1, line 12: When deploying a celestia node to read from, fetch celestia auth token through the token server. Otherwise, use a raw value

joroshiba (commented):
I think we should angle towards not having.... I'm working on a PR which supports this through conductor more thorougly. 

The issue right now is that if a token is NOT set we also have issues because of the way conductor works.

PR #1284, opened 2024-07-19
chore: ibc e2e smoke test

steezeburger:
## Summary
Adds smoke test for end to end Celestia > Astria > EVM > back to Astria > Celestia bridging

## Background
Smoke tests r gud

## Changes
* Combines work from withdraw smoke tests and ibc bridging tests

## Testing
This is the PR for the testing

PR #1364, opened 2024-08-13
fix(sequencer): ICS 20 Withdrawals Missing current_block_timestamp

steezeburger:
## Summary
Quick hack to fix ics20 withdrawals for e2e ibc testing

## Background
During ibc testing, was experiencing error from Penumbra "Missing current_block_timestamp".

## Changes
- Sets block timestamp using Penumbra's `EpochManager` for now. Should 

## Testing
Tested as part of ibc e2e tests here https://github.com/astriaorg/astria/pull/1284

## Breaking Changelist
- not sure of implications here


dev/values/validators/single.yml
```diff
@@ -1,6 +1,12 @@
 global:
   namespaceOverride: astria-dev-cluster
 
+images:
+  sequencer:
+    repo: ghcr.io/astriaorg/sequencer
+    tag: "0.14.0"
+    devTag: sha-b4c5d1c
```
quasystaty1, line 8: Should this be a specific devTag sha instead of latest? can add a new `single-ics20.yml` for now instead or manually change when using.

steezeburger: It should be latest, but the image hasn't been built yet because the fix is in this pr

steezeburger:
Fixed by https://github.com/astriaorg/astria/pull/1389

PR #1366, opened 2024-08-14
fix(relayer): change `reqwest` for `isahc` in relayer blackbox tests (ENG-699)

### LINKED ISSUE #1354: Fix sequencer-relayer tests on macOS
┆Issue Number: ENG-699

Fraser999:
## Summary
This is a straight swap of [`isahc`](https://github.com/sagebind/isahc) in place of `reqwest` in the relayer's blackbox tests.  Other than a single timeout being increased, no logic was changed.

## Background
Blackbox tests were failing on macOS due to the poor performance of the `reqwest` messages.  The timeout for a `GET` was set to 100ms, and using `reqwest` I was seeing times around 120ms.  By swapping to `isahc`, times dropped to 2ms.

## Changes
- Swapped `isahc` in place of `reqwest` in the realyer's blackbox tests.
- Increased the timeout on a flaky test from 2s to 10s (the test occasionally takes just over 2s, less than 3s from several attempts locally).
- Added `[lints.rust]` to the relayer's manifest to squash some warnings.

## Testing
No new tests required, existing ones now pass on macOS.

## Related Issues
Closes #1354.

### DIFF
--- a/crates/astria-sequencer-relayer/Cargo.toml
+++ b/crates/astria-sequencer-relayer/Cargo.toml
@@ -80,7 +80,6 @@ hyper = { workspace = true }
 itertools = { workspace = true }
 once_cell = { workspace = true }
 rand_core = { version = "0.6", features = ["getrandom"] }
-reqwest = { workspace = true, features = ["json"] }
 tempfile = { workspace = true }
 tendermint-rpc = { workspace = true, features = ["http-client"] }
 tokio = { workspace = true, features = ["test-util"] }
@@ -88,9 +87,16 @@ tokio-stream = { workspace = true, features = ["net"] }
 wiremock = { workspace = true }
 
 assert-json-diff = "2.0.2"
+http = "0.2.7"
+isahc = { version = "1.7.2", features = ["json"] }
 tower = { version = "0.4.13" }
 tokio-test.workspace = true
 rand_chacha = "0.3.1"
 
 [build-dependencies]
 astria-build-info = { path = "../astria-build-info", features = ["build"] }
+
+[lints.rust]
+unexpected_cfgs = { level = "warn", check-cfg = [
+  'cfg(dylint_lib, values("tracing_debug_field"))',
+] }
--- a/crates/astria-sequencer-relayer/tests/blackbox/main.rs
+++ b/crates/astria-sequencer-relayer/tests/blackbox/main.rs
@@ -12,7 +12,7 @@ use helpers::{
     SequencerBlockToMount,
     TestSequencerRelayerConfig,
 };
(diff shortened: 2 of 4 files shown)


crates/astria-sequencer-relayer/Cargo.toml
```diff
@@ -80,17 +80,23 @@ hyper = { workspace = true }
...
+isahc = { version = "1.7.2", features = ["json"] }
 tower = { version = "0.4.13" }
 tokio-test.workspace = true
 rand_chacha = "0.3.1"
 
 [build-dependencies]
 astria-build-info = { path = "../astria-build-info", features = ["build"] }
+
+[lints.rust]
+unexpected_cfgs = { level = "warn", check-cfg = [
```
SuperFluffy, line 100: What is this for?

crates/astria-sequencer-relayer/tests/blackbox/main.rs
```diff
@@ -273,7 +273,7 @@ async fn should_filter_rollup() {
         .await;
     sequencer_relayer
         .timeout_ms(
-            2_000,
+            10_000,
```
SuperFluffy, line 276: This seems like it might just fix the tests itself (compared to swapping out reqwest for isahc)?

SuperFluffy (approved):
Interesting. Some years ago I observed that I'd get double digits millisecond latency on Linux compared to the macOS loopback device (which was in the low digit ms).

Do you know why isahc's latency is so much lower? Just experimenting?

crates/astria-sequencer-relayer/Cargo.toml
Fraser999: I was getting a bunch of warnings when building on macOS, and this was the fix suggested as part of the warning text.

crates/astria-sequencer-relayer/tests/blackbox/main.rs
Fraser999: It was more the calls to `TestSequencerRelayer::get_from_relayer_api` which were causing multiple tests to fail - specifically the 100ms timeout inside that method.

Even after fixing that (by using isahc) this test was occasionally taking just over 2 seconds on macOS.  I didn't check if we also see that on Linux.

Fraser999:
@SuperFluffy 

> Do you know why isahc's latency is so much lower? Just experimenting?

I was recommended `isahc` a few months back by a friend who had found it to be much faster than `reqwest`.  I believe the  thinking is that `reqwest` burns a lot of cycles in setting up connection pools, but I haven't actually looked into that.

PR #1367, opened 2024-08-14
fix(conductor): update for celestia-node v0.15.0

ethanoroshiba:
## Summary
Updated `fetch_blobs_with_retry()` to accept [Celestia Node v0.15.0](https://github.com/celestiaorg/celestia-node/releases/tag/v0.15.0)'s new `GetAll()` return format. This will still work with Node v0.14.1 as well.

## Background
celestia-node v0.15.0 now returns an empty list of blobs for `GetAll` if blobs are not found, which broke our current `jsonrpsee` error handling in conductor.

## Changes
- Updated `fetch_blobs_with_retry()` to accept the `ParseError` that Json RPC now throws when there are no blobs fetched.

## Testing
E2E tests working

### DIFF
--- a/crates/astria-conductor/src/celestia/fetch.rs
+++ b/crates/astria-conductor/src/celestia/fetch.rs
@@ -120,6 +120,7 @@ async fn fetch_blobs_with_retry(
             match client.blob_get_all(height, &[namespace]).await {
                 Ok(blobs) => Ok(blobs),
                 Err(err) if is_blob_not_found(&err) => Ok(vec![]),
+                Err(err) if is_null_blobs(&err) => Ok(vec![]),
                 Err(err) => Err(err),
             }
         }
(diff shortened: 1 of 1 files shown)


SuperFluffy (dismissed):
This checks if parsing failed and is extremely fragile.

Please try this branch instead: https://github.com/eigerco/lumina/pull/340

crates/astria-conductor/src/celestia/fetch.rs
```diff
@@ -119,7 +119,7 @@ async fn fetch_blobs_with_retry(
         async move {
             match client.blob_get_all(height, &[namespace]).await {
                 Ok(blobs) => Ok(blobs),
-                Err(err) if is_blob_not_found(&err) => Ok(vec![]),
+                Err(err) if is_null_blobs(&err) => Ok(vec![]),
```
joroshiba, line 122: I think we should keep the old check as well, so that celestia upgrade can happen seamlessly without requiring conductor upgrade in lockstep.

joroshiba (changes requested):

joroshiba (approved):

PR #1368, opened 2024-08-14
chore(sequencer): Add instrumentation

### LINKED ISSUE #1321: chore: add instrumentation on all async calls

### LINKED ISSUE #1331: chore: refactor execute_ics20_transfer()
This function needs a refactor to get rid of the clippy exception:

https://github.com/astriaorg/astria/blob/79531330196249e128cb7f46b2b3e14a95aff464/crates/astria-sequencer/src/ibc/ics20_transfer.rs#L370-L377

┆Issue Number: ENG-675

ethanoroshiba:
## Summary
Added instrumentation to `astria-sequencer`

## Background

## Changes
- Disabled `clippy::blocks_in_conditions` lint
- Added instrumentation to all async function calls that are not long-lived.
- Removed instrumentation from long-running functions.
- Minor refactoring of two functions to get rid of clippy exceptions and move logging within spans

## Related Issues
closes #1321


Fraser999 (commented):
I think we should be using `err(Debug)` throughout this PR since we're dealing with `anyhow::Result`s, and they are usually fairly unhelpful if just `Display` is used.

ethanoroshiba:
> I think we should be using `err(Debug)` throughout this PR since we're dealing with `anyhow::Result`s, and they are usually fairly unhelpful if just `Display` is used.

Reverted `Debug` instances for this PR, will make a followup to transition to `eyre` instead of `anyhow`

crates/astria-sequencer/src/accounts/query.rs
```diff
@@ -177,6 +181,7 @@ async fn get_snapshot_and_height(storage: &Storage, height: Height) -> Result<(S
     Ok((snapshot, height))
 }
 
+#[instrument(skip_all)]
```
SuperFluffy, line 184:
```suggestion
#[instrument(skip_all, err)]
```

crates/astria-sequencer/src/accounts/query.rs
```diff
@@ -36,6 +36,7 @@ use crate::{
     state_ext::StateReadExt as _,
 };
 
+#[instrument(skip_all, err)]
```
SuperFluffy, line 39: Let's add `fields(%asset)` here

crates/astria-sequencer/src/accounts/query.rs
```diff
@@ -154,6 +157,7 @@ pub(crate) async fn nonce_request(
     }
 }
 
+#[instrument(skip_all, err)]
```
SuperFluffy, line 160: let's add `fields(%height)`

.github/workflows/test.yml
```diff
@@ -211,7 +211,8 @@ jobs:
       - name: run pedantic clippy on workspace crates
         run: |
           cargo clippy --all-targets --all-features \
-          -- --warn clippy::pedantic --warn clippy::arithmetic-side-effects --deny warnings
+          -- --warn clippy::pedantic --warn clippy::arithmetic-side-effects \
+          --allow clippy::blocks_in_conditions --deny warnings
```
SuperFluffy, line 215: I am working on updating to rust 1.81.0, but I am running into some issues: https://github.com/astriaorg/astria/pull/1523

Just a headsup so that we can remove it later.

crates/astria-sequencer/src/accounts/state_ext.rs
```diff
@@ -208,7 +208,7 @@ pub(crate) trait StateReadExt: StateRead + crate::assets::StateReadExt {
         Ok(balance)
     }
 
-    #[instrument(skip_all)]
+    #[instrument(skip_all, err)]
```
SuperFluffy, line 211: Note on this `T: AddressBytes` and all methods that contain a similar argument: in https://github.com/astriaorg/astria/pull/1495 I have added logic to display the address. It would be very useful to get that context here.

SuperFluffy (commented):
Overall good PR, but I would like to expand this PR, adding fields where needed.

When writing https://github.com/astriaorg/astria/pull/1495 I started adding fields to the spans because without them the generated `err` events will be hard to understand.

This will inevitably create tension when reading logs on stdout, but to understand error context there is no way around.

crates/astria-sequencer/src/accounts/query.rs
ethanoroshiba: I tried this, and it doesn't work because the function's error type doesn't implement the display trait. Is there another way around this that you'd like to do?

crates/astria-sequencer/src/accounts/state_ext.rs
ethanoroshiba: Tried to add similar fields wherever valuable in 2b30bf2 :)

crates/astria-sequencer/src/accounts/state_ext.rs
```diff
@@ -262,7 +262,7 @@ pub(crate) trait StateWriteExt: StateWrite {
         Ok(())
     }
 
-    #[instrument(skip_all)]
+    #[instrument(skip_all, fields(address = %address.display_address()), err)]
```
SuperFluffy, line 265: What about the nonce? Let's add that too?

crates/astria-sequencer/src/accounts/state_ext.rs
```diff
@@ -325,7 +325,7 @@ pub(crate) trait StateWriteExt: StateWrite {
         Ok(())
     }
 
-    #[instrument(skip_all)]
+    #[instrument(skip_all, fields(%fee), err)]
```
SuperFluffy, line 328: `u128` implements `Valuable`, so the `%` is not necessary.

PR #1376, opened 2024-08-16
fix(bridge-withdrawer, cli, sequencer-client): migrate from `broadcast_tx_commit` to `broadcast_tx_sync`

### LINKED ISSUE #1283: Don't use broadcast_tx_commit in bridge-withdrawer
The `bridge-withdrawer` currently uses `broadcast_tx_commit` to submit transactions to the sequencer. We used this instead of `broadcast_tx_sync` in order to make sure the transaction containing a batch of withdrawals is finalized on the sequencer before the next batch is processed.

The [CometBFT RPC docs](https://docs.cometbft.com/main/rpc/#/Tx/broadcast_tx_commit) warn against using `broadcast_tx_commit` in production.

Instead, we should use `broadcast_tx_sync` and then check for the transaction finalizing with a backoff before processing the next batch.

ethanoroshiba:
## Summary
Deleted all usage of `broadcast_tx_commit` and instead implemented `broadcast_tx_sync`.

## Background
Previously, we relied on `broadcast_tx_commit` to submit transactions to the sequencer. [CometBFT RPC docs warn against this](https://docs.cometbft.com/main/rpc/#/Tx/broadcast_tx_commit) and instead favor using `broadcast_tx_sync`.

## Changes
- Deleted all instances of `broadcast_tx_commit`, replacing with `broadcast_tx_sync`.
- Added function `wait_for_tx_inclusion()` to probe the sequencer client for a new transaction of the given hash with backoff.

## Testing
- Added unit test to `sequencer-client` to ensure `wait_for_tx_inclusion()` works properly

## Related Issues
closes #1283


crates/astria-sequencer-client/src/extension_trait.rs
```diff
@@ -664,21 +670,52 @@ pub trait SequencerClientExt: Client {
...
     /// # Errors
     ///
-    /// - If calling the tendermint RPC endpoint fails.
-    async fn submit_transaction_commit(
+    /// - If the transaction is not found.
+    /// - If the transaction execution failed.
+    /// - If the transaction proof is missing.
+    #[allow(clippy::blocks_in_conditions)] // Allow: erroneous clippy warning. Should be fixed in Rust 1.81
+    #[instrument(skip_all, err)]
+    async fn wait_for_tx_inclusion(
```
Fraser999, line 682: It might be worth considering following a similar approach to that taken in [the sequencer-relayer for `confirm_submission`](https://github.com/astriaorg/astria/blob/f69ffe22a53b063984c83c5993798e249b39c46d/crates/astria-sequencer-relayer/src/relayer/celestia_client/mod.rs#L331-L378) which is essentially doing the same thing, but targeting Celestia rather than a sequencer.

The main differences are:
- it tries forever, hence the function doesn't return a `Result`
- when the server sends a success response, but indicates the tx hash wasn't found, the retry interval stays at the minimum
- because we can keep retrying fairly frequently, there's a logging strategy to avoid spamming the logs

If you decide to go that way, then bear in mind that the timing consts were chosen for a Celestia block time of 12 seconds, whereas the sequencer block times are closer to 2 seconds I believe.

ethanoroshiba: Thanks for the tip! I changed the logic to reflect the function you sent over. The only differences are:
- used millis instead of secs since the timeframe of this is much shorter
- kept the result return value. If we get a response which has a non-ok error code or doesn't include proof of the tx's inclusion in the block, I think it would probably still be good to throw an error.

crates/astria-bridge-withdrawer/src/bridge_withdrawer/submitter/mod.rs
```diff
@@ -262,21 +265,24 @@ async fn submit_tx(
...
     .await
     .wrap_err("failed sending transaction after 1024 attempts");
 
-    state.set_sequencer_connected(res.is_ok());
+    state.set_sequencer_connected(check_tx.is_ok());
 
     metrics.record_sequencer_submission_latency(start.elapsed());
 
-    res
+    let check_tx = check_tx?;
```
noot, line 282: check the `check_tx` code here, ie move this block here: `if let tendermint::abci::Code::Err(check_tx_code) = check_tx.code`

if `check_tx` returns an error code, then the next call to `wait_for_tx_inclusion` can never be successful.

crates/astria-sequencer-client/src/extension_trait.rs
```diff
@@ -664,21 +669,69 @@ pub trait SequencerClientExt: Client {
...
+            logged_at = Instant::now();
+        };
+
+        let mut sleep_millis = MIN_POLL_INTERVAL_MILLIS;
+        loop {
+            tokio::time::sleep(Duration::from_millis(sleep_millis)).await;
+            match self.tx(tx_hash, true).await {
+                Ok(tx) => {
+                    ensure!(
+                        tx.tx_result.code.is_ok() && tx.proof.is_some(),
```
noot, line 720:
```suggestion
            match self.tx(tx_hash, false).await {
                Ok(tx) => {
                    ensure!(
                        tx.tx_result.code.is_ok() 
```

don't need a tx proof

noot, line 720: also, you error here if the code is not ok, however the callers for this function also check for a not-ok code, which is redundant. either check `is_ok` here or in the callers, but not both. I think checking only in callers is sufficient as the calling functions can then decide how to handle the error code.

ethanoroshiba: Removed from `wait_for_tx_inclusion()` in 5812193 and added checks to callers which did not have them. Thanks for the review!

crates/astria-sequencer-client/src/extension_trait.rs
Fraser999: It looks like we're always returning `Ok`, so maybe the return type can be changed to just `tendermint_rpc::endpoint::tx::Response`?

ethanoroshiba: Should we also have a wrapper function with a timeout, similar to the one you sent? Seems like it may be problematic to introduce code which will try forever unchecked instead of erroring out at some point.

Fraser999 (approved):

PR #1387, opened 2024-08-21
chore(sequencer): migrate from `anyhow::Result` to `eyre::Result`

### LINKED ISSUE #1386: Transition sequencer to eyre results
Currently, the entire sequencer uses `anyhow::Result`, which provides limited error information in its `Display` trait. This is because Penumbra's `cnidarium` uses `anyhow`, but we should update all pieces which do not touch Penumbra to use `eyre` like the rest of our codebase. This will provide continuity and also allow for useful error emission after the added instrumentation in  #1368.

┆Issue Number: ENG-735

ethanoroshiba:
## Summary
Migrate all instances of `anyhow::Result` to `eyre::Result`.

## Background
Sequencer was using `anyhow::Result`, which provides an unhelpful `Display` impl and contrasts our error handling in the rest of the codebase. This change is to flush out our error handling in the sequencer, except for those parts which necessitate using `anyhow`.

## Changes
- Add eyre to sequencer's `cargo.toml`.
- Migrate all instances of `anyhow::Result` to `eyre::Result`, except for those that touch cnidarium directly.
- Create `anyhow_to_eyre()` and `eyre_to_anyhow()` helper functions for moving between the two without breaking the source chain.

## Testing
Added unit tests to ensure `eyre` and `anyhow` source chains are maintained when converting between one another.

## Related Issues
closes #1386


SuperFluffy (changes requested):
This approach breaks the error source chain, see proof below. (on a side note: please base the instrumentation fix in ENG-670 on this change from anyhow to eyre, not the other way round)

We need to find a way to turn `anyhow:Error` into a `Box<dyn std::error::Error>` and then wrap that in `eyre` (which sounds easier then it is in practice): https://docs.rs/anyhow/latest/anyhow/struct.Error.html#impl-From%3CError%3E-for-Box%3Cdyn+Error+%2B+Send+%2B+Sync%3E

Proof:

```
eyre-wrapped anyhow error:
Err(wrapped

Caused by:
    level 2

Location:
    src/main.rs:8:55)

eyre-wrapped anyhow error: broken source chain
0: wrapped
1: level 2
eyre error:
Err(wrapped

Caused by:
   0: level 2
   1: level 1
   2: level 0

Location:
    src/main.rs:6:33)
eyre-only error: functioning source chain
0: wrapped
1: level 2
2: level 1
3: level 0
```

Code used to generate this:

```rust
fn main() {
    let anyhow_err = Err::<(), _>(anyhow!("level 0").context("level 1").context("level 2"));
    let eyre_err = Err::<(), _>(eyre!("level 0").wrap_err("level 1").wrap_err("level 2"));

    let wrapped_anyhow_err = anyhow_err.map_err(|err| eyre!(err).wrap_err("wrapped"));
    println!("eyre-wrapped anyhow error:\n{wrapped_anyhow_err:?}\n");

    println!("eyre-wrapped anyhow error: broken source chain");
    for (i, source) in wrapped_anyhow_err.unwrap_err().chain().enumerate() {
        println!("{i}: {source}");
    }

    let wrapped_eyre_err = eyre_err.wrap_err("wrapped");
    println!("eyre error:\n{wrapped_eyre_err:?}");

    println!("eyre-only error: functioning source chain");
    for (i, source) in wrapped_eyre_err.unwrap_err().chain().enumerate() {
        println!("{i}: {source}");
    }
```

crates/astria-sequencer/Cargo.toml
```diff
@@ -17,6 +17,8 @@ benchmark = ["divan"]
 [dependencies]
 astria-core = { path = "../astria-core", features = ["server", "serde"] }
 astria-build-info = { path = "../astria-build-info", features = ["runtime"] }
+astria-eyre = { path = "../astria-eyre" }
```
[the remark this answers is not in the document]
ethanoroshiba: `anyhow` is still referenced directly in some implementations of Penumbra's traits, such as for `HostInterface`, so I'm not sure if we can remove the direct dependency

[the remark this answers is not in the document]
SuperFluffy: Yes, you are right - especially given our offline discussion about how to retain the source chain.

crates/astria-sequencer/src/utils.rs
```diff
@@ -16,9 +20,29 @@ impl<'a> std::fmt::Display for Hex<'a> {
     }
 }
 
+pub(crate) fn anyhow_to_eyre(anyhow_error: anyhow::Error) -> astria_eyre::eyre::Report {
+    let anyhow_result = Err::<(), _>(anyhow_error);
+    let boxed: Result<(), Box<dyn std::error::Error + Send + Sync>> =
+        anyhow_result.map_err(std::convert::Into::into);
+    let Err(err) = boxed else {
+        panic!("anyhow_to_eyre called on `Ok`")
+    };
```
SuperFluffy, line 29: This err -> result -> err dance is not necessary:
```suggestion
    let boxed: Box<dyn std::error::Error + Send + Sync> = anyhow::Error.into();
    eyre!(boxed)
```

crates/astria-sequencer/src/utils.rs
```diff
@@ -16,9 +20,29 @@ impl<'a> std::fmt::Display for Hex<'a> {
...
+    eyre!(err)
+}
+
+pub(crate) fn eyre_to_anyhow(eyre_error: astria_eyre::eyre::Report) -> anyhow::Error {
```
SuperFluffy, line 33: Have you checked that this conversion works and retains the source chain?

crates/astria-sequencer/src/utils.rs
```diff
@@ -16,9 +20,29 @@ impl<'a> std::fmt::Display for Hex<'a> {
...
+    let eyre_result = Err::<(), _>(eyre_error);
+    let boxed: Result<(), Box<dyn std::error::Error + Send + Sync>> =
+        eyre_result.map_err(std::convert::Into::into);
+    let Err(err) = boxed else {
+        panic!("eyre_to_anyhow called on `Ok`")
+    };
+    anyhow::anyhow!(err)
```
SuperFluffy, line 40: Similar to the eyre case above: the err -> result -> err is probably not necessary. :)

crates/astria-sequencer/src/ibc/ics20_transfer.rs
```diff
@@ -11,12 +11,7 @@
 //! [`AppHandlerExecute`] is used for execution.
 use std::borrow::Cow;
 
-use anyhow::{
-    bail,
-    ensure,
-    Context as _,
-    Result,
-};
+use anyhow::Context;
```
SuperFluffy, line 14:
```suggestion
use anyhow::Context as _;
```

SuperFluffy (commented):
This is almost ready to be approved. Thank you for going through all the code and updating it!

My only concern is with the conversion of eyre back to anyhow: have you checked that the source chain is maintained?

I would actually like to propose a unit test for both:

Do somehing like this:
```rust
let mut errs = &["foo", "bar", "baz", "qux"];
let anyhow_error = anyhow!(errs[0]).context(errs[1]).context(errs[2]);
let eyre_from_anyhow = anyhow_to_eyre(anyhow_err).wrap_err(errs[3]);

errs.reverse();
for (i, err) = eyre_from_anyhow.chain().enumerate() {
    assert_eq!(errs[i], &err.to_string());
}
```

And similarly for defining an `eyre_error` and converting it to anyhow.

crates/astria-sequencer/src/ibc/ics20_transfer.rs
```diff
@@ -394,33 +425,49 @@ async fn execute_ics20_transfer<S: ibc::StateWriteExt>(
...
+    if is_refund
+        && serde_json::from_str::<memos::v1alpha1::Ics20WithdrawalFromRollup>(&packet_data.memo)
+            .is_ok()
+    {
+        execute_withdrawal_refund_to_rollup(state, packet_data)
+            .await
+            .wrap_err("failed to execute rollup withdrawal refund")?;
+        return Ok(());
+    }
+
```
Fraser999, line 463: I think this should be removed - it's duplicated at line 416 above.

Fraser999 (approved):

crates/astria-eyre/Cargo.toml
```diff
@@ -8,8 +8,12 @@ edition = "2021"
...
 itoa = "1.0.10"
+anyhow = { version = "1", optional = true }
 
 [dev-dependencies]
 tracing = { workspace = true }
 tokio-test = { workspace = true }
 tracing-subscriber = "0.3.18"
+
+[features]
+anyhow_support = ["anyhow"]
```
SuperFluffy, line 19: This is not necessary. Remove the extra feature. Because rust dependencies and features (unfortunately) share the same namespace, you can activate it by using `cargo build --features anyhow`

crates/astria-eyre/src/lib.rs
```diff
@@ -83,3 +85,44 @@ fn write_value(err: &dyn Error, f: &mut core::fmt::Formatter<'_>) -> core::fmt::
     f.write_fmt(format_args!("\"{err}\""))?;
     Ok(())
 }
+
+#[cfg(feature = "anyhow_support")]
```
SuperFluffy, line 89: Put all of this under a module so you don't need to repeat `cfg(feature = ...)` all the time:

#[cfg(feature = "anyhow")]
pub use anyhow_conversion::{anyhow_to_eyre, eyre_to_anyhow};

```rust
#[cfg(feature = "anyhow")]
mod anyhow_conversion {
   pub fn anyhow_to_eyre() {}

   pub fn eyre_to_anyhow() {}
   
   #[cfg(test)]
    mod tests {
        // the tests
    }
}
```

crates/astria-sequencer/Cargo.toml
```diff
@@ -17,13 +17,13 @@ benchmark = ["divan"]
 [dependencies]
 astria-core = { path = "../astria-core", features = ["server", "serde"] }
 astria-build-info = { path = "../astria-build-info", features = ["runtime"] }
+astria-eyre = { path = "../astria-eyre", features = ["anyhow_support"] }
```
SuperFluffy, line 20: Please make a note here why this feature is necessary (because it's needed to implement the foreign trait `penumbra_ibc::component::HostInterface` in `crates/astria-sequencer/src/ibc/host_interface.rs`)

crates/astria-eyre/Cargo.toml
```diff
@@ -8,8 +8,12 @@ edition = "2021"
 [dependencies]
 eyre = "0.6"
 itoa = "1.0.10"
+anyhow = { version = "1", optional = true }
```
SuperFluffy, line 11: It's recommended to provide the exact version that you used when writing this code, i.e. the result of `cargo add anyhow --optional` in this case.

SuperFluffy (approved):
Thank you for going through the entire codebase and making these changes! I think we should go ahead and merge this.

I have left a few comments. Please address them before merging.

crates/astria-eyre/src/lib.rs
```diff
@@ -86,43 +89,44 @@ fn write_value(err: &dyn Error, f: &mut core::fmt::Formatter<'_>) -> core::fmt::
     Ok(())
 }
 
-#[cfg(feature = "anyhow_support")]
-pub fn anyhow_to_eyre(anyhow_error: anyhow::Error) -> eyre::Report {
-    let boxed: Box<dyn std::error::Error + Send + Sync> = anyhow_error.into();
-    eyre::eyre!(boxed)
-}
+#[cfg(feature = "anyhow")]
+pub mod anyhow_conversion {
```
SuperFluffy, line 93: Don't make this a public module. Just reexport the inner functions at the root (as you already do.

crates/astria-eyre/src/lib.rs
```diff
@@ -86,43 +89,44 @@ fn write_value(err: &dyn Error, f: &mut core::fmt::Formatter<'_>) -> core::fmt::
...
+        anyhow::anyhow!(boxed)
     }
 
-    #[test]
-    #[cfg(feature = "anyhow_support")]
-    fn eyre_to_anyhow_preserves_source_chain() {
-        let mut errs = ["foo", "bar", "baz", "qux"];
-        let eyre_error = eyre::eyre!(errs[0]).wrap_err(errs[1]).wrap_err(errs[2]);
-        let anyhow_from_eyre = crate::eyre_to_anyhow(eyre_error).context(errs[3]);
+    mod test {
```
SuperFluffy, line 107: As I said in the comment, change to this so that the module is only present in tests.

```rust
#[cfg(test)]
mod tests {
```

PR #1391, opened 2024-08-22
feat(sequencer, bridge-withdrawer)!: enforce withdrawals consumed

joroshiba:
## Summary
Adds state for bridge events and consumes the events as they occur. Updates proto for bridge unlock to no longer use memo for rollup information and instead include on main action. Renames `rollup_transaction_hash` to `rollup_withdrawal_event_id`.

## Background
There was only implicit stop against reusing a withdrawal event, this adds consumption of withdrawal events. While this is not strictly security enhancing, consuming the event adds in protocol protection against accidental double spend by the bridge operator. 

## Changes
- Proto updates for `BridgeUnlockAction` + `ICS20WithdrawalMemo`
- Add stateless checks on bridging unlock and withdrawals of correct information filled out
- Added state to enforce rollup event cannot be consumed twice. 
- Enforce that `Ics20WithdrawalAction` must have `bridge_address` set if making a bridge withdrawal

## Testing
smoke test + minor test updates

## Breaking Changelist
- New stateful information about rollup withdrawal events
- Adds new fields to `BridgeUnlockAction`

## Related Issues
closes https://github.com/astriaorg/astria/issues/1430


crates/astria-sequencer/src/bridge/bridge_unlock_action.rs
```diff
@@ -11,23 +11,43 @@ use astria_core::protocol::transaction::v1alpha1::action::{
...
 };
 
 #[async_trait::async_trait]
 impl ActionHandler for BridgeUnlockAction {
     async fn check_stateless(&self) -> Result<()> {
+        ensure!(self.amount > 0, "amount must be greater than zero",);
+        ensure!(self.memo.len() <= 64, "memo must not be more than 64 bytes");
+        ensure!(
+            !self.rollup_withdrawal_event_id.is_empty(),
+            "rollup transaction hash must be non-empty",
```
Fraser999, line 37:
```suggestion
            "rollup withdrawal event id must be non-empty",
```

crates/astria-sequencer/src/bridge/bridge_unlock_action.rs
```diff
@@ -11,23 +11,43 @@ use astria_core::protocol::transaction::v1alpha1::action::{
...
+        );
+        ensure!(
+            self.rollup_withdrawal_event_id.len() <= 64,
+            "rollup transaction hash must not be more than 64 bytes",
```
Fraser999, line 41:
```suggestion
            "rollup withdrawal event id must not be more than 64 bytes",
```

crates/astria-sequencer/src/bridge/bridge_unlock_action.rs
```diff
@@ -68,6 +100,11 @@ impl ActionHandler for BridgeUnlockAction {
         };
 
         check_transfer(&transfer_action, self.bridge_address, &state).await?;
+        state.put_withdrawal_event_block_for_bridge_account(
+            self.bridge_address.address_bytes(),
```
Fraser999, line 104:
```suggestion
            self.bridge_address,
```

crates/astria-sequencer/src/bridge/bridge_unlock_action.rs
```diff
@@ -60,6 +80,18 @@ impl ActionHandler for BridgeUnlockAction {
...
+            .get_withdrawal_event_block_for_bridge_account(
+                self.bridge_address,
+                &self.rollup_withdrawal_event_id,
+            )
+            .await
+            .context("failed to get withdrawal event height")?;
+        ensure!(
+            rollup_withdrawal_height.is_none(),
+            "withdrawal event already processed",
+        );
```
Fraser999, line 93: Maybe useful to include the rollup height in the error?
```suggestion
        if let Some(height) = rollup_withdrawal_height {
            bail!("withdrawal event already processed at rollup height {height}");
        }
```

crates/astria-sequencer/src/bridge/bridge_unlock_action.rs
```diff
@@ -258,31 +301,41 @@ mod tests {
...
+            rollup_withdrawal_event_id: "a-rollup-defined-hash".to_string(),
+        };
+        let bridge_unlock_second = BridgeUnlockAction {
+            to: to_address,
+            amount: transfer_amount,
+            fee_asset: asset.clone(),
+            memo: String::new(),
+            bridge_address,
+            rollup_block_number: 10,
+            rollup_withdrawal_event_id: "a-rollup-defined-hash".to_string(),
```
Fraser999, line 325:
```suggestion
        let bridge_unlock_second = BridgeUnlockAction {
            rollup_block_number: 10,
            ..bridge_unlock_first.clone()
```

crates/astria-sequencer/src/bridge/state_ext.rs
```diff
@@ -340,6 +354,25 @@ pub(crate) trait StateReadExt: StateRead + address::StateReadExt {
...
+        let Some(bytes) = self
+            .get_raw(&key)
+            .await
+            .context("failed reading raw withdrawal event from state")?
+        else {
+            return Ok(None);
+        };
+
+        let block_num = u64::try_from_slice(&bytes).context("invalid block number bytes")?;
+        Ok(Some(block_num))
```
Fraser999, line 373:
```suggestion
        let byte_array = <[u8; 8]>::try_from(bytes).map_err(|bytes| {
            anyhow!(
                "invalid block number bytes: expected 8, got {}",
                bytes.len()
            )
        })?;

        Ok(Some(u64::from_be_bytes(byte_array)))
```

crates/astria-sequencer/src/ibc/ics20_withdrawal.rs
```diff
@@ -106,6 +112,37 @@ async fn establish_withdrawal_target<S: StateRead>(
 impl ActionHandler for action::Ics20Withdrawal {
     async fn check_stateless(&self) -> Result<()> {
         ensure!(self.timeout_time() != 0, "timeout time must be non-zero",);
+        ensure!(self.amount() > 0, "amount must be greater than zero",);
+        if self.bridge_address.is_some() {
+            let parsed_bridge_memo: Ics20WithdrawalFromRollupMemo =
```
Fraser999, line 117: We're JSON-encoding a PB version of this type, whereas here we're decoding straight into our domain type.  The serde `try_from = "raw::Ics20WithdrawalFromRollup"` on the domain type handles that ok, but I still keep feeling uneasy when I see this.

crates/astria-sequencer/src/ibc/ics20_withdrawal.rs
```diff
@@ -106,6 +112,37 @@ async fn establish_withdrawal_target<S: StateRead>(
...
+                !parsed_bridge_memo.rollup_return_address.is_empty(),
+                "rollup return address must be non-empty",
+            );
+            ensure!(
+                parsed_bridge_memo.rollup_return_address.into_bytes().len() <= 256,
+                "rollup return address must be no more than 256 bytes",
+            );
+            ensure!(
+                !parsed_bridge_memo.rollup_withdrawal_event_id.is_empty(),
+                "rollup transaction hash must be non-empty",
```
Fraser999, line 131:
```suggestion
                "rollup withdrawal event id must be non-empty",
```

crates/astria-sequencer/src/ibc/ics20_withdrawal.rs
```diff
@@ -106,6 +112,37 @@ async fn establish_withdrawal_target<S: StateRead>(
...
+            );
+            ensure!(
+                parsed_bridge_memo
+                    .rollup_withdrawal_event_id
+                    .into_bytes()
+                    .len()
+                    <= 64,
+                "rollup transaction hash must be no more than 64 bytes",
```
Fraser999, line 139:
```suggestion
                "rollup withdrawal event id must be no more than 64 bytes",
```

crates/astria-sequencer/src/ibc/ics20_withdrawal.rs
```diff
@@ -128,6 +165,27 @@ impl ActionHandler for action::Ics20Withdrawal {
             state.ensure_base_prefix(bridge_address).await.context(
                 "failed to verify that bridge address address has permitted base prefix",
             )?;
+            let parsed_bridge_memo: Ics20WithdrawalFromRollupMemo =
+                serde_json::from_str(&self.memo)
+                    .context("failed to memo for ICS bound bridge withdrawal")?;
```
Fraser999, line 170:
```suggestion
                    .context("failed to decode memo for ICS bound bridge withdrawal")?;
```

crates/astria-sequencer/src/ibc/ics20_withdrawal.rs
```diff
@@ -128,6 +165,27 @@ impl ActionHandler for action::Ics20Withdrawal {
...
+                .get_withdrawal_event_block_for_bridge_account(
+                    self.bridge_address.map_or(from, Address::bytes),
+                    &parsed_bridge_memo.rollup_withdrawal_event_id,
+                )
+                .await
+                .context("failed to get withdrawal event height")?;
+            ensure!(
+                rollup_withdrawal_height.is_none(),
+                "withdrawal event already processed",
+            );
```
Fraser999, line 182:
```suggestion
            if let Some(height) = rollup_withdrawal_height {
                bail!("withdrawal event already processed at rollup height {height}");
            }
```

Fraser999 (commented):
Mostly nitpicks, but one issue in bridge/state_ext.rs causing the rollup block height to not be parsed correctly.  Not blocking the PR though since the actual value isn't critical to the flow, only whether it exists in storage or not.

crates/astria-sequencer/src/ibc/ics20_withdrawal.rs
joroshiba: I could either remove the domain type or modify the serialization on the other end, it uses the pb::name property when serializing for errors which is why it uses the raw type

Fraser999 (approved):

crates/astria-sequencer/src/ibc/ics20_withdrawal.rs
Fraser999: It's certainly not a blocker, and maybe I'm being overly pedantic.  But I guess I'd prefer if the domain type didn't derive `serde` traits, forcing us to always use the pb type for encoding to/decoding from JSON, and we use the domain type in the business logic.

No probs if you prefer to leave as is - I think we're doing similar things elsewhere.  We can always try and get more strict on the separation of pb and domain types in the future.

joroshiba: I went ahead and went back to how the memos were defined in domain type as reexporting, less code changes here a bit cleaner I think.

crates/astria-sequencer/src/bridge/bridge_unlock_action.rs
```diff
@@ -60,6 +77,17 @@ impl ActionHandler for BridgeUnlockAction {
...
 
+        let rollup_withdrawal_height = state
+            .get_withdrawal_event_block_for_bridge_account(
+                self.bridge_address,
+                &self.rollup_withdrawal_event_id,
+            )
+            .await
+            .context("failed to get withdrawal event height")?;
+        if let Some(height) = rollup_withdrawal_height {
+            bail!("withdrawal event already processed at rollup height {height}");
```
SuperFluffy, line 88:
```suggestion
            bail!("a withdrawal event with ID `{}` was already processed at rollup height {height}", self.rollup_withdrawal_event_id);
```

SuperFluffy (approved):
Discussed offline: `BridgeUnlockAction::check_and_execute` contains a check that no event with a given ID was previously processed, so a second action would not be able to overwrite the first. That addresses my issue.

Approving.

crates/astria-sequencer/src/bridge/bridge_unlock_action.rs
```diff
@@ -17,17 +17,34 @@ use crate::{
...
+        StateReadExt as _,
+        StateWriteExt as _,
+    },
     transaction::StateReadExt as _,
 };
 
 #[async_trait::async_trait]
 impl ActionHandler for BridgeUnlockAction {
     async fn check_stateless(&self) -> Result<()> {
+        ensure!(self.amount > 0, "amount must be greater than zero",);
```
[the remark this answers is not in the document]
joroshiba: It's just kinda pointless? I would prefer not wasting compute time on pointless actions :)

crates/astria-sequencer/src/bridge/state_ext.rs
```diff
@@ -419,6 +456,17 @@ pub(crate) trait StateWriteExt: StateWrite {
...
     }
 
+    #[instrument(skip_all)]
+    fn put_withdrawal_event_block_for_bridge_account<T: AddressBytes>(
+        &mut self,
+        address: T,
+        withdrawal_event_id: &str,
+        block_num: u64,
+    ) {
+        let key = bridge_account_withdrawal_event_storage_key(&address, withdrawal_event_id);
```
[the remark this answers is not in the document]
joroshiba: I have moved the check and set logic into a single function to avoid requirement higher up to be ordered and set correctly.

crates/astria-sequencer/src/bridge/state_ext.rs
```diff
@@ -457,14 +434,34 @@ pub(crate) trait StateWriteExt: StateWrite {
...
+            .await
+            .context("failed reading raw withdrawal event from state")?;
+        if let Some(bytes) = bytes {
+            let existing_block_num = u64::from_be_bytes(
+                bytes
+                    .try_into()
+                    .expect("all block numbers stored should be 8 bytes; this is a bug"),
+            );
+
+            return Err(anyhow!(
```
SuperFluffy, line 457: Replace this by the `anyhow::bail!` macro

crates/astria-sequencer/src/ibc/ics20_withdrawal.rs
```diff
@@ -69,47 +69,53 @@ fn withdrawal_to_unchecked_ibc_packet(
...
 
-    ensure!(
-        withdrawer == from.address_bytes(),
-        "sender does not match bridge withdrawer address; unauthorized"
-    );
+    // If the bridge address is not set, the sender must not be a bridge account.
+    if state
+        .is_a_bridge_account(from)
+        .await
+        .context("failed to get bridge account rollup id")?
```
SuperFluffy, line 108: Update this error message (here you don't know what the called method is actually doing):
```suggestion
        .context("failed to establish whether the sender is a bridge account")?
```

crates/astria-sequencer/src/ibc/ics20_withdrawal.rs
joroshiba: I have created an issue here for follow up

crates/astria-sequencer/src/ibc/ics20_withdrawal.rs
```diff
@@ -128,6 +161,26 @@ impl ActionHandler for action::Ics20Withdrawal {
             state.ensure_base_prefix(bridge_address).await.context(
                 "failed to verify that bridge address address has permitted base prefix",
             )?;
+            let parsed_bridge_memo: Ics20WithdrawalFromRollupMemo =
```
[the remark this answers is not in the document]
joroshiba: Added as a follow up TODO with linked issue

PR #1404, opened 2024-08-23
trusted block builder second iteration

bharath-123:
## Summary

## Background

## Changes

## Testing
How are these changes tested?

## Metrics

## Breaking Changelist

## Related Issues

closes


charts/composer/values.yaml
```diff
@@ -3,31 +3,30 @@ global:
...
+    repo: astria-composer
     tag: "0.8.2"
-    devTag: latest
+    devTag: tb
 
 config:
-  logLevel: "debug"
+  logLevel: "info"
   maxBytesPerBundle: 200000
+  maxBundleSize: 200000
```
bharath-123, line 17: `maxBundleSize` can be removed.

crates/astria-composer/src/config.rs
```diff
@@ -70,21 +65,12 @@ pub struct Config {
...
-            .split(',')
-            .filter(|s| !s.is_empty())
-            .map(|s| Rollup::parse(s).map(Rollup::into_parts))
-            .collect::<Result<HashMap<_, _>, _>>()
-    }
+    /// The URL of the execution API server
+    pub execution_api_url: String,
+
+    /// The maximum possible size of a bundle
+    pub max_bundle_size: usize,
```
bharath-123, line 73: This is not necessary, we can just hard code this in the system.

PR #1406, opened 2024-08-23
feat(ci): provide demo/preview environments

aajimal:
## Summary
Updates the celestia-local chart to allow helm-parameter overrides when creating PR preview environments.

## Background
To show how we can leverage a preview environments for: 
	- testing work in progress PR
	- providing a demo environment  to external users
	- Testing against a live cluster 

## Changes
- Update celestia-local chart to:
  - remove hardcoded values
  - allow specifying ingress params
  - ensuring PVC storage is namespace scoped
- Update sequencer and evm-stack components to support parameterized imagePullPolicy for ArgoCD
- Include ArgoCD ApplicationSet for PR preview envs (to be applied by admins)

## Testing
- Rendering locally for manual inspection
- Applied changes to argocd.dev.astria.org to test on live cluster
- Verify components are being deployed successfully 
- Confirm smoke-tests pass when pointing to deployed components


charts/celestia-local/templates/_helpers.tpl
```diff
@@ -0,0 +1,102 @@
...
+4
+{{- else if eq .Values.config.logLevel "trace" }}
+5
+{{- end }}
+{{- end }}
+
+{{/*
+Full image paths for Astria built images
+*/}}
+{{- define "celestia.image" -}}
```
quasystaty1, line 50: Should Astria built images be here?

WafflesVonMaple (approved):

charts/celestia-local/templates/_helpers.tpl
aajimal: Good catch! Thanks! Actually looks like none of this image stuff is used anywhere (not even in the celestia-node chart). I must have copied from another chart + did search/replace and forgot to remove references that aren't needed.

quasystaty1 (approved):

PR #1408, opened 2024-08-26
feat(sequencer): make mempool balance aware

Lilyjjo:
## Summary
Added logic to the mempool to make it's pending vs parked logic balance aware. Transactions will only be added to the pending queue if the signing account has enough balance to cover the cost of all of the transactions in the pending plus the new transaction. Transactions that exceed the account's available balance are put into parked. 

Note: this is a draft because it is missing the logic to re-cost all of the transactions after a `FeeAssetChange` or `FeeChange` action occur.

## Background
Previously the mempool did not have sufficient logic to deal with transaction costs. `CheckTx` would kick out any transaction that the account did not have enough balance to cover. This didn't account for the edge cases of:
1. A different in-flight transaction from another account funding the deficient account.
2. Sequential transactions that together exceed an account's balance but independently are fine. This causes unnecessary transaction failures in `prepare_proposal()`'s block construction. 

## Changes
TODO

## Testing
TODO

## Metrics
TODO

## Breaking Changelist
TODO

## Related Issues
TODO

closes

### DIFF
--- a/crates/astria-sequencer/src/accounts/state_ext.rs
+++ b/crates/astria-sequencer/src/accounts/state_ext.rs
@@ -1,10 +1,15 @@
+use std::collections::HashMap;
+
 use anyhow::{
     Context,
     Result,
 };
 use astria_core::{
     primitive::v1::{
-        asset,
+        asset::{
+            self,
+            IbcPrefixed,
+        },
         Address,
     },
     protocol::account::v1alpha1::AssetBalance,
@@ -69,8 +74,47 @@ fn nonce_storage_key<T: AddressBytes>(address: T) -> String {
 pub(crate) trait StateReadExt: StateRead + crate::assets::StateReadExt {
     #[instrument(skip_all)]
     async fn get_account_balances(&self, address: Address) -> Result<Vec<AssetBalance>> {
-        let prefix = format!("{}/balance/", StorageKey(&address));
+        let ibc_balances = self
+            .get_account_balances_ibc(address)
+            .await
+            .context("failed to grab ibc balances for account")?;
         let mut balances: Vec<AssetBalance> = Vec::new();
+
+        for (asset, balance) in ibc_balances {
+            let native_asset = self
+                .get_native_asset()
+                .await
+                .context("failed to read native asset from state")?;
+            if asset == native_asset.to_ibc_prefixed() {
+                balances.push(AssetBalance {
+                    denom: native_asset.into(),
+                    balance,
+                });
+                continue;
+            }
+
+            let denom = self
+                .map_ibc_to_trace_prefixed_asset(asset)
+                .await
+                .context("failed to get ibc asset denom")?
+                .context("asset denom not found when user has balance of it; this is a bug")?
+                .into();
+            balances.push(AssetBalance {
+                denom,
+                balance,
+            });
+        }
+
+        Ok(balances)
+    }
+
+    #[instrument(skip_all)]
+    async fn get_account_balances_ibc<T: AddressBytes>(
+        &self,
+        address: T,
+    ) -> Result<HashMap<IbcPrefixed, u128>> {
+        let prefix = format!("{}/balance/", StorageKey(&address));
+        let mut balances: HashMap<IbcPrefixed, u128> = HashMap::new();
 
         let mut stream = std::pin::pin!(self.prefix_keys(&prefix));
         while let Some(Ok(key)) = stream.next().await {
@@ -94,28 +138,7 @@ pub(crate) trait StateReadExt: StateRead + crate::assets::StateReadExt {
             let Balance(balance) =
                 Balance::try_from_slice(&value).context("invalid balance bytes")?;
 
-            let native_asset = self
-                .get_native_asset()
-                .await
-                .context("failed to read native asset from state")?;
-            if asset == native_asset.to_ibc_prefixed() {
-                balances.push(AssetBalance {
-                    denom: native_asset.into(),
-                    balance,
-                });
-                continue;
-            }
-
-            let denom = self
-                .map_ibc_to_trace_prefixed_asset(asset)
-                .await
-                .context("failed to get ibc asset denom")?
-                .context("asset denom not found when user has balance of it; this is a bug")?
-                .into();
-            balances.push(AssetBalance {
-                denom,
-                balance,
-            });
+            balances.insert(asset, balance);
         }
         Ok(balances)
     }
@@ -579,6 +602,70 @@ mod tests {
             .await
             .expect("retrieving account balances should not fail");
         assert_eq!(balances, vec![]);
+    }
+
+    #[tokio::test]
+    async fn get_account_balances_ibc() {
+        let storage = cnidarium::TempStorage::new().await.unwrap();
+        let snapshot = storage.latest_snapshot();
+        let mut state = StateDelta::new(snapshot);
+
+        // native account should work with ibc too
+        state.put_native_asset(&nria());
+
+        let asset_0 = state.get_native_asset().await.unwrap();
+        let asset_1 = asset_1();
+        let asset_2 = asset_2();
+
+        // also need to add assets to the ibc state
+        state
+            .put_ibc_asset(&asset_0.clone())
+            .expect("should be able to call other trait method on state object");
+        state
+            .put_ibc_asset(&asset_1.clone().unwrap_trace_prefixed())
+            .expect("should be able to call other trait method on state object");
+        state
+            .put_ibc_asset(&asset_2.clone().unwrap_trace_prefixed())
+            .expect("should be able to call other trait method on state object");
+
+        // create needed variables
+        let address = astria_address(&[42u8; 20]);
+        let amount_expected_0 = 1u128;
+        let amount_expected_1 = 2u128;
+        let amount_expected_2 = 3u128;
+
+        // add balances to the account
+        state
+            .put_account_balance(address, asset_0.clone(), amount_expected_0)
+            .expect("putting an account balance should not fail");
+        state
+            .put_account_balance(address, &asset_1, amount_expected_1)
+            .expect("putting an account balance should not fail");
+        state
+            .put_account_balance(address, &asset_2, amount_expected_2)
+            .expect("putting an account balance should not fail");
+
+        let balances = state
+            .get_account_balances_ibc(address)
+            .await
+            .expect("retrieving account balances should not fail");
+
+        assert_eq!(
+            balances.get(&asset_0.into()).expect("x"),
+            &amount_expected_0,
+            "returned value for ibc asset_0 does not match"
+        );
+        assert_eq!(
+            balances.get(&asset_1.into()).expect("x"),
+            &amount_expected_1,
+            "returned value for ibc asset_1 does not match"
+        );
+        assert_eq!(
+            balances.get(&asset_2.into()).expect("x"),
+            &amount_expected_2,
+            "returned value for ibc asset_2 does not match"
+        );
+        assert_eq!(balances.len(), 3, "should only return existing values");
     }
 
     #[tokio::test]
--- a/crates/astria-sequencer/src/mempool/mod.rs
+++ b/crates/astria-sequencer/src/mempool/mod.rs
@@ -1,9 +1,11 @@
+#[cfg(test)]
 mod benchmarks;
 mod transactions_container;
 
 use std::{
     collections::{
         HashMap,
+        HashSet,
         VecDeque,
     },
     future::Future,
@@ -11,7 +13,10 @@ use std::{
     sync::Arc,
 };
 
-use astria_core::protocol::transaction::v1alpha1::SignedTransaction;
+use astria_core::{
+    primitive::v1::asset::IbcPrefixed,
+    protocol::transaction::v1alpha1::SignedTransaction,
+};
 use tokio::{
     join,
     sync::{
@@ -37,7 +42,6 @@ pub(crate) enum RemovalReason {
     NonceStale,
     LowerNonceInvalidated,
     FailedPrepareProposal(String),
-    FailedCheckTx(String),
 }
 
 /// How long transactions are considered valid in the mempool.
@@ -159,18 +163,28 @@ impl Mempool {
         &self,
         tx: Arc<SignedTransaction>,
         current_account_nonce: u32,
+        current_account_balances: HashMap<IbcPrefixed, u128>,
+        transaction_cost: HashMap<IbcPrefixed, u128>,
     ) -> anyhow::Result<(), InsertionError> {
-        let timemarked_tx = TimemarkedTransaction::new(tx);
+        let timemarked_tx = TimemarkedTransaction::new(tx, transaction_cost);
 
         let (mut pending, mut parked) = self.acquire_both_locks().await;
 
-        // try insert into pending (will fail if nonce is gapped or already present)
-        match pending.add(timemarked_tx.clone(), current_account_nonce) {
-            Err(InsertionError::NonceGap) => {
+        // try insert into pending
+        match pending.add(
+            timemarked_tx.clone(),
+            current_account_nonce,
+            &current_account_balances,
+        ) {
+            Err(InsertionError::NonceGap | InsertionError::AccountBalanceTooLow) => {
                 // Release the lock asap.
                 drop(pending);
                 // try to add to parked queue
-                parked.add(timemarked_tx, current_account_nonce)
+                parked.add(
+                    timemarked_tx,
+                    current_account_nonce,
+                    &current_account_balances,
+                )
             }
             error @ Err(
                 InsertionError::AlreadyPresent
@@ -180,17 +194,22 @@ impl Mempool {
             ) => error,
             Ok(()) => {
                 // check parked for txs able to be promoted
-                let to_promote = parked.pop_front_account(
+                let to_promote = parked.find_promotables(
                     timemarked_tx.address(),
                     timemarked_tx
                         .nonce()
                         .checked_add(1)
                         .expect("failed to increment nonce in promotion"),
+                    &pending.remaining_account_balances(
+                        *timemarked_tx.address(),
+                        current_account_balances.clone(),
+                    ),
                 );
-                // Release the lock asap.
-                drop(parked);
+                // promote the transactions
                 for ttx in to_promote {
-                    if let Err(error) = pending.add(ttx, current_account_nonce) {
+                    if let Err(error) =
+                        pending.add(ttx, current_account_nonce, &current_account_balances)
+                    {
                         error!(
                             current_account_nonce,
                             "failed to promote transaction during insertion: {error:#}"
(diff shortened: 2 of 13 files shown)


crates/astria-sequencer/src/service/mempool.rs
```diff
@@ -326,7 +367,7 @@ async fn handle_check_tx<S: accounts::StateReadExt + address::StateReadExt + 'st
     let mempool_len = mempool.len().await;
 
     metrics
-        .record_check_tx_duration_seconds_insert_to_app_mempool(finished_check_removed.elapsed());
+        .record_check_tx_duration_seconds_insert_to_app_mempool(finished_fetch_tx_cost.elapsed());
```
Fraser999, line 370:
```suggestion
        .record_check_tx_duration_seconds_insert_to_app_mempool(finished_fetch_balances.elapsed());
```

Fraser999 (commented):
Partial review with some minor suggestions.

crates/astria-sequencer/src/accounts/state_ext.rs
```diff
@@ -69,8 +74,47 @@ fn nonce_storage_key<T: AddressBytes>(address: T) -> String {
...
+                denom,
+                balance,
+            });
+        }
+
+        Ok(balances)
+    }
+
+    #[instrument(skip_all)]
+    async fn get_account_balances_ibc<T: AddressBytes>(
```
noot, line 112:
```suggestion
    async fn get_account_balances<T: AddressBytes>(
```

the native asset can still be "ibc-prefixed" (sort of confusing but it is what it is lol) so prefer to not specific `_ibc` here

crates/astria-sequencer/src/accounts/state_ext.rs
```diff
@@ -69,8 +74,47 @@ fn nonce_storage_key<T: AddressBytes>(address: T) -> String {
 pub(crate) trait StateReadExt: StateRead + crate::assets::StateReadExt {
     #[instrument(skip_all)]
     async fn get_account_balances(&self, address: Address) -> Result<Vec<AssetBalance>> {
+        let ibc_balances = self
```
noot, line 77:
```suggestion
        let ibc_prefixed_balances = self
```

since there's the native asset, which is not ibc'd, but can be an ibc-prefixed denom

crates/astria-sequencer/src/mempool/mod.rs
```diff
@@ -267,34 +286,104 @@ impl Mempool {
...
+        // 3.) if there were no demotions, check if parked has transactions we can
+        //     promote
+
+        // TODO reprice transactions also if FeeAssetChange or FeeChange Actions were ran :0
+
+        let addresses: HashSet<[u8; 20]> = pending
+            .addresses()
+            .into_iter()
+            .chain(parked.addresses())
+            .collect();
```
noot, line 321: there could be duplicates of address here right? if it's in both pending and parked?  maybe sort and then dedup?

crates/astria-sequencer/src/mempool/mod.rs
```diff
@@ -267,34 +286,104 @@ impl Mempool {
...
+
+            // clean pending and parked of stale and expired
+            removed_txs.extend(pending.clean_account_stale_expired(address, current_nonce));
+            removed_txs.extend(parked.clean_account_stale_expired(address, current_nonce));
+
+            // get transactions to demote from pending
+            let demotion_txs = pending.find_demotables(address, &current_balances);
+
+            if demotion_txs.is_empty() {
+                // nothing to demote, check for transactions to promote
```
noot, line 354: why do we only promote txs if there's nothing to demote? shouldn't both occur either way?

crates/astria-sequencer/src/mempool/transactions_container.rs
```diff
@@ -159,6 +174,86 @@ impl PendingTransactionsForAccount {
     fn highest_nonce(&self) -> Option<u32> {
         self.txs.last_key_value().map(|(nonce, _)| *nonce)
     }
+
+    /// Removes and returns transactions that exceed the balances in `available_balances`.
+    fn find_demotables(
+        &mut self,
+        available_balances: &HashMap<IbcPrefixed, u128>,
```
noot, line 181:
```suggestion
        available_balances: HashMap<IbcPrefixed, u128>,
```

think you can just pass the map here and not clone below

crates/astria-sequencer/src/mempool/transactions_container.rs
```diff
@@ -159,6 +174,86 @@ impl PendingTransactionsForAccount {
...
+                        } else {
+                            // not enough balance, do not include
+                            break_flag = true;
+                            break;
+                        }
+                    }
+                    hash_map::Entry::Vacant(_) => {
+                        // not enough balance, do not include
+                        if *cost != 0 {
+                            break_flag = true;
```
noot, line 205: i think you can use `break 'label` here as to not use this: https://doc.rust-lang.org/rust-by-example/flow_control/loop/nested.html

crates/astria-sequencer/src/mempool/transactions_container.rs
```diff
@@ -159,6 +174,86 @@ impl PendingTransactionsForAccount {
...
+
+        for (nonce, tx) in &self.txs {
+            // ensure we have enough balance to cover inclusion
+            for (denom, cost) in tx.cost() {
+                match available_balances.entry(*denom) {
+                    hash_map::Entry::Occupied(mut entry) => {
+                        if cost <= entry.get() {
+                            // subtract cost
+                            let current_cost = entry.get_mut();
+                            *current_cost = current_cost.saturating_sub(*cost);
```
noot, line 195:
```suggestion
                            let current_balance = entry.get_mut();
                            *current_balance = current_balance.saturating_sub(*cost);
```

crates/astria-sequencer/src/mempool/transactions_container.rs
```diff
@@ -159,6 +174,86 @@ impl PendingTransactionsForAccount {
...
+
+        // return all keys higher than split target
+        self.txs.split_off(&split_at).into_values().collect()
+    }
+
+    /// Returns remaining balances after accounting for costs of contained transactions.
+    ///
+    /// Note: assumes that the balances in `current_account_balance` are large enough
+    /// to cover costs for contained transactions. Will log an error if this is not true
+    /// but will not fail.
```
noot, line 227: why not return an error? what cases would that error happen, developer error or other?

crates/astria-sequencer/src/mempool/transactions_container.rs
```diff
@@ -159,6 +174,86 @@ impl PendingTransactionsForAccount {
...
+            split_at = nonce.saturating_add(1);
+        }
+
+        // return all keys higher than split target
+        self.txs.split_off(&split_at).into_values().collect()
+    }
+
+    /// Returns remaining balances after accounting for costs of contained transactions.
+    ///
+    /// Note: assumes that the balances in `current_account_balance` are large enough
```
noot, line 225:
```suggestion
    /// Note: assumes that the balances in `current_account_balances` are large enough
```

crates/astria-sequencer/src/mempool/transactions_container.rs
```diff
@@ -189,6 +284,35 @@ impl TransactionsForAccount for PendingTransactionsForAccount {
...
+
+    fn has_balance_to_cover(
+        &self,
+        ttx: &TimemarkedTransaction,
+        current_account_balances: &HashMap<IbcPrefixed, u128>,
+    ) -> bool {
+        let mut remaining_account_balances = current_account_balances.clone();
+
+        // deduct costs from current account balances and return false if missing funds for any
+        for tx in self.txs.values().chain(std::iter::once(ttx)) {
```
noot, line 296: does this work if `ttx` has a nonce that isn't directly sequential to `self.txs`?

crates/astria-sequencer/src/mempool/transactions_container.rs
```diff
@@ -189,6 +284,35 @@ impl TransactionsForAccount for PendingTransactionsForAccount {
         // is equal to the account nonce
         self.txs().contains_key(&previous_nonce) || ttx.signed_tx.nonce() == current_account_nonce
     }
+
+    fn has_balance_to_cover(
+        &self,
+        ttx: &TimemarkedTransaction,
+        current_account_balances: &HashMap<IbcPrefixed, u128>,
+    ) -> bool {
+        let mut remaining_account_balances = current_account_balances.clone();
```
noot, line 293: maybe sum all the costs within the txs and then check against the map balances so you don't have to clone

crates/astria-sequencer/src/mempool/transactions_container.rs
```diff
@@ -200,17 +324,45 @@ pub(super) struct ParkedTransactionsForAccount<const MAX_TX_COUNT: usize> {
...
+                // ensure we have enough balance to cover promotions
+                for (denom, cost) in self.txs.get(nonce).unwrap().cost() {
+                    match available_balances.entry(*denom) {
+                        hash_map::Entry::Occupied(mut entry) => {
+                            if cost <= entry.get() {
+                                let remaining_balance = entry.get_mut();
+                                *remaining_balance = remaining_balance.saturating_sub(*cost);
+                            } else {
+                                // not enough balance, do not include
+                                break_flag = true;
```
noot, line 350: same here, `break 'label`

crates/astria-sequencer/src/mempool/transactions_container.rs
```diff
@@ -200,17 +324,45 @@ pub(super) struct ParkedTransactionsForAccount<const MAX_TX_COUNT: usize> {
...
 impl<const MAX_TX_COUNT: usize> ParkedTransactionsForAccount<MAX_TX_COUNT> {
     /// Returns contiguous transactions from front of queue starting from target nonce, removing the
-    /// transactions in the process.
+    /// transactions in the process. Will only return transactions if their cost is covered by the
+    /// `available_balances`.
     ///
     /// Note: this function only operates on the front of the queue. If the target nonce is not at
-    /// the front, an error will be logged and nothing will be returned.
-    fn pop_front_contiguous(
+    /// the front, nothing will be returned.
```
noot, line 331: why pass in the target nonce if the function only works at the front of the queue?

crates/astria-sequencer/src/mempool/mod.rs
Lilyjjo: They're collected into a HashSet which does the dedup by nature of the container. Happy to change into something else if you'd prefer

crates/astria-sequencer/src/mempool/mod.rs
Lilyjjo: If we demoted something from an account's pending queue that means there shouldn't be anything to promote from the account's parked queue. The pending queue's enforcement of sequential nonces is the main reason for this.

For example, if we have nonces `[0,1,2]` in pending and nonces `[3,4]` in parked , and `[2]` gets demoted from pending because of lack of account funds, then parked would be `[2,3,4]`.  During the promotion loop `2` would be ineligible for promotion due to lack of account funds, also making `3,4` ineligible because of the sequential nonce requirement.

crates/astria-sequencer/src/mempool/transactions_container.rs
Lilyjjo: It'd be a code logic error. It is possible to have too high costs in pending after transaction re-costing or a different node including a different transaction for the same account's nonce in a block, but we only call `get_remaining_balances()` after we run the demotion logic on pending which should make the transaction costs covered.

Fraser and I thought it would be better to let the mempool continue running if somewhere else in the code there was a bug that would make the above assumptions untrue instead of having the mempool crash. The idea was that someone would see the error logs and we'd know we need to go fix something 😬

crates/astria-sequencer/src/mempool/transactions_container.rs
Lilyjjo: Yes, the `.chain(x)` appends `x` as the last element in the iterator. It doesn't apply any sorting or deduping logic to the additional element. We only ever trigger this after ensuring that the nonce is sequential but it's not necessary

crates/astria-sequencer/src/mempool/transactions_container.rs
Lilyjjo: The target nonce is the next nonce that pending is expecting, I can add that to the documentation. It's not guaranteed that parked will contain a transaction with the target nonce and I was letting this function internalize that check instead of doing it explicitly. I can make it explicit if you'd prefer. 

There shouldn't be the scenario where parked and pending have overlapping nonces as we only add things to parked if their nonce isn't already in pending.

crates/astria-sequencer/src/mempool/mod.rs
```diff
@@ -267,34 +286,104 @@ impl Mempool {
...
+        let mut removed_txs = Vec::<([u8; 32], RemovalReason)>::new();
+
+        // To clean we need to:
+        // 1.) remove stale and expired transactions from both pending and parked
+        // 2.) check if we have transactions in pending which need to be demoted due
+        //     to balance decreases
+        // 3.) if there were no demotions, check if parked has transactions we can
+        //     promote
+
+        // TODO reprice transactions also if FeeAssetChange or FeeChange Actions were ran :0
```
Fraser999, line 315: Is this ticketed?

crates/astria-sequencer/src/mempool/transactions_container.rs
```diff
@@ -159,6 +174,77 @@ impl PendingTransactionsForAccount {
...
+                        *current_cost = current_cost.saturating_sub(*cost);
+                    }
+                    hash_map::Entry::Vacant(_) => {
+                        if *cost != 0 {
+                            error!("pending transactions has cost not in account balances");
+                        }
+                    }
+                }
+            }
+        }
```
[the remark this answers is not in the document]
Lilyjjo: learning that for loops are less cool than iterators in rust land haha

crates/astria-sequencer/src/mempool/transactions_container.rs
```diff
@@ -278,17 +433,30 @@ pub(super) trait TransactionsForAccount: Default {
         current_account_nonce: u32,
     ) -> bool;
 
+    /// Returns `Ok` if adding `ttx` would not break the balance precondition, i.e. enough
+    /// balance to cover all transactions in `CostsCovered` mode.
```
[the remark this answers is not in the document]
Lilyjjo: I was trying to do the naming similar to the `SequentialNonces` mode we have but for saying the container only contains transactions that are covered by the account's balances. On reflection `CostsCovered` isn't the same as the `SequentialNonces` case because we have to reclean to stay in `CostsCovered` when `SequentialNonces` is always true.

Happy to remove the comments if you think they are misleading.

crates/astria-sequencer/src/mempool/transactions_container.rs
```diff
@@ -189,6 +275,46 @@ impl TransactionsForAccount for PendingTransactionsForAccount {
...
         self.txs().contains_key(&previous_nonce) || ttx.signed_tx.nonce() == current_account_nonce
     }
+
+    fn has_balance_to_cover(
+        &self,
+        ttx: &TimemarkedTransaction,
+        current_account_balances: &HashMap<IbcPrefixed, u128>,
+    ) -> bool {
+        // build up transaction cost map
+        let mut transaction_costs = HashMap::<IbcPrefixed, u128>::new();
```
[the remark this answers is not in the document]
Lilyjjo: I included everything but the `CostDeductionError` part of the suggestion. It would be worthwhile to give users a comprehensive way to see their balances and pending/parked statuses outside of this PR. 

I think creating this error category here would be too heavy weight since the errors as is would only return partial information on how much funds are lacking

crates/astria-sequencer/src/accounts/state_ext.rs
```diff
@@ -69,8 +74,47 @@ fn nonce_storage_key<T: AddressBytes>(address: T) -> String {
...
+    }
+
+    #[instrument(skip_all)]
+    async fn get_account_balances_ibc_prefixed<T: AddressBytes>(
+        &self,
+        address: T,
+    ) -> Result<HashMap<IbcPrefixed, u128>> {
         let prefix = format!("{}/balance/", StorageKey(&address));
-        let mut balances: Vec<AssetBalance> = Vec::new();
+        let mut balances: HashMap<IbcPrefixed, u128> = HashMap::new();
```
SuperFluffy, line 117: This should be an `IndexMap` to provide a stable order.

crates/astria-sequencer/src/accounts/state_ext.rs
```diff
@@ -69,8 +74,47 @@ fn nonce_storage_key<T: AddressBytes>(address: T) -> String {
 pub(crate) trait StateReadExt: StateRead + crate::assets::StateReadExt {
     #[instrument(skip_all)]
     async fn get_account_balances(&self, address: Address) -> Result<Vec<AssetBalance>> {
```
SuperFluffy, line 76: Not a fan of this code duplication here and with `get_account_balances_ibc_prefixed`. Internally the "natural" representation is in terms of ibc-prefixed items (i.e. `ibc/<sha>`). So the standard way to look at denoms should also be that.

If you want to map the ibc-prefixed items to full trace-prefixed items then that should likely happen at the boundary.

SuperFluffy, line 76: I wish this simpler named function was done in terms of ibc prefixes because that's the "natural" representation inside sequencer (as it maps all trace prefixed items to ibc prefixed and only makes the reverse mapping at the API boundary).

crates/astria-sequencer/src/accounts/state_ext.rs
```diff
@@ -69,8 +74,47 @@ fn nonce_storage_key<T: AddressBytes>(address: T) -> String {
...
+        let ibc_prefixed_balances = self
+            .get_account_balances_ibc_prefixed(address)
```
SuperFluffy, line 78: On the one hand it's great that this method is built on top of the lower level `get_account_balances_ibc_prefixed`. On the other hand we could entirely avoid the allocation if we used the stream provided by cnidarium.

crates/astria-sequencer/src/mempool/transactions_container.rs
```diff
@@ -33,15 +40,17 @@ pub(super) struct TimemarkedTransaction {
     tx_hash: [u8; 32],
     time_first_seen: Instant,
     address: [u8; 20],
+    cost: HashMap<IbcPrefixed, u128>,
 }
 
 impl TimemarkedTransaction {
-    pub(super) fn new(signed_tx: Arc<SignedTransaction>) -> Self {
+    pub(super) fn new(signed_tx: Arc<SignedTransaction>, cost: HashMap<IbcPrefixed, u128>) -> Self {
```
SuperFluffy, line 47: `new` methods should almost always only contain one argument. I think in this case it would be acceptable to leave the method as is but provide a new inherent method `TimemarkedTransaction::set_cost_map(&mut self, cost_map: HashMap<IbcPrefixed, u128>)`

crates/astria-sequencer/src/mempool/transactions_container.rs
```diff
@@ -159,6 +196,71 @@ impl PendingTransactionsForAccount {
...
+    }
+
+    /// Returns remaining balances after accounting for costs of contained transactions.
+    ///
+    /// Note: assumes that the balances in `current_account_balances` are large enough
+    /// to cover costs for contained transactions. Will log an error if this is not true
+    /// but will not fail.
+    fn get_remaining_balances(
+        &self,
+        mut remaining_account_balances: HashMap<IbcPrefixed, u128>,
```
SuperFluffy, line 243: the fact that this takes an argument throws me off. Can there be a better name? Also, shouldn't this take `remaining_account_balances` as a `&mut`?

crates/astria-sequencer/src/accounts/state_ext.rs
Lilyjjo: We don't really need the balances sorted in the rest of the code that uses the `get_account_balances_ibc_prefixed()` function. I'd prefer to just sort the returned vec in either `get_account_balance()` or in the one function that calls it (`balance_request()`)
https://github.com/astriaorg/astria/blob/e9d5f0f0fc2f2449292aeeb07a1f2f7e7227a637/crates/astria-sequencer/src/accounts/query.rs#L36

SuperFluffy: The point is that you get the sorting for free (in the same order as the Vec).

Since indexmap is already a dependency this is doesn't have downsides.

crates/astria-sequencer/src/accounts/state_ext.rs
Lilyjjo: changed `get_account_balances()` to `get_account_balances_traced_prefixed()` and used the simpler name for `get_account_balances_ibc_prefixed()` 👍

crates/astria-sequencer/src/accounts/state_ext.rs
Lilyjjo: I can not use the inner function if you'd prefer. The `get_account_balances()` is only used for outside queries and there isn't much overall performance impact from the allocation (from what I can tell)

crates/astria-sequencer/src/mempool/transactions_container.rs
Lilyjjo: I'll add the internal function on the stacked PR [1411](https://github.com/astriaorg/astria/pull/1411) since we do change the cost var there

crates/astria-sequencer/src/mempool/transactions_container.rs
Lilyjjo: renamed `subtract_contained_costs` and fixed the `&mut`!

crates/astria-sequencer/src/mempool/mod.rs
Lilyjjo: it's written in: https://github.com/astriaorg/astria/pull/1411

crates/astria-sequencer/src/accounts/state_ext.rs
```diff
@@ -94,28 +144,7 @@ pub(crate) trait StateReadExt: StateRead + crate::assets::StateReadExt {
...
-                .map_ibc_to_trace_prefixed_asset(asset)
-                .await
-                .context("failed to get ibc asset denom")?
-                .context("asset denom not found when user has balance of it; this is a bug")?
-                .into();
-            balances.push(AssetBalance {
-                denom,
-                balance,
-            });
+            balances.insert(asset, balance);
```
SuperFluffy, line 147: The return value of this `insert` should be checked.

crates/astria-sequencer/src/mempool/transactions_container.rs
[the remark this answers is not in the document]
Fraser999: I guess if you're ok to remove just the " in `CostsCovered` mode" part, that would make more sense to me.

crates/astria-sequencer/src/mempool/transactions_container.rs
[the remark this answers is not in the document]
Fraser999: Ah, right.  I think I was originally expecting we could use this method in what is now named `subtract_contained_costs` and there we would have wanted to match on the error variants to log the appropriate thing, so `anyhow::Error` wouldn't have been good for that.

Seems like we actually don't use the returned error anywhere other than for checking `.is_err()` or `.is_ok()`, so the actual info added in the `anyhow!` calls won't be visible to anyone as things stand, but that's not a problem I guess :)

Lilyjjo:
Addressed all comments I didn't comment responses to, marking this ready for review again

crates/astria-sequencer/src/accounts/state_ext.rs
```diff
@@ -65,59 +69,131 @@ fn nonce_storage_key<T: AddressBytes>(address: T) -> String {
...
+}
+
+#[derive(Clone, Copy, Debug)]
+pub(crate) struct AssetBalance {
+    pub(crate) asset: asset::IbcPrefixed,
+    pub(crate) balance: u128,
+}
+
+pin_project! {
+    /// A stream of IBC prefixed assets for a given account.
```
SuperFluffy, line 108:
```suggestion
    /// A stream of IBC prefixed assets and their balances for a given account.
```

crates/astria-sequencer/src/accounts/state_ext.rs
```diff
@@ -65,59 +69,131 @@ fn nonce_storage_key<T: AddressBytes>(address: T) -> String {
...
+    ) -> AccountAssetBalancesStream<Self::PrefixRawStream> {
+        let prefix = format!("{}/balance/", StorageKey(&address));
+        AccountAssetBalancesStream {
+            underlying: self.prefix_raw(&prefix),
         }
-        Ok(balances)
+    }
+
+    #[instrument(skip_all)]
+    async fn get_account_balances(
```
SuperFluffy, line 183: This should probably not live in  `StateReadExt` since this is mempool-related logic.

I know you have the goal of keeping the concept of our state out of the mempol proper - I suggest you make a new sibling `mempool-state.rs`/`mod mempool_state` where you define a helper `fn get_account_balances<S: StateRead>(state: S, account: impl AddressBytes) -> HashMap<IbcPrefixed, u128>`.

This keeps both the mempool module and the `StateReadExt` impl clean from mempool requirements.

We should keep the `StateReadExt` implementations as thin wrappers around cnidarium. If a module (like mempool) needs extra structure on top of the stream, then that module should implement the logic there instead of here.

crates/astria-sequencer/src/accounts/state_ext.rs
```diff
@@ -65,59 +69,131 @@ fn nonce_storage_key<T: AddressBytes>(address: T) -> String {
...
+    ) -> anyhow::Result<std::collections::HashMap<asset::IbcPrefixed, u128>> {
+        use futures::TryStreamExt as _;
+        self.account_asset_balances(address)
+            .map_ok(
+                |crate::accounts::AssetBalance {
+                     asset,
+                     balance,
+                 }| (asset, balance),
+            )
+            .try_collect::<std::collections::HashMap<_, _>>()
```
SuperFluffy, line 195: Add a note there that this relies on the IBC prefixed assets coming out of the stream being unique.

This assumption is justified because the prefixes are extracted from (unique) keys (I certainly hope they are unique because otherwise that would be a weird database).

crates/astria-sequencer/src/accounts/state_ext.rs
```diff
@@ -264,11 +340,9 @@ impl<T: StateWrite> StateWriteExt for T {}
 
 #[cfg(test)]
 mod tests {
-    use astria_core::{
-        primitive::v1::Address,
-        protocol::account::v1alpha1::AssetBalance,
-    };
+    use astria_core::primitive::v1::Address;
     use cnidarium::StateDelta;
+    use futures::TryStreamExt;
```
SuperFluffy, line 345:
```suggestion
    use futures::TryStreamExt as _;
```

SuperFluffy (commented):
few comments on the merged asset/asset-balance stream

Fraser999 (approved):
LGTM now - optimistic approval assuming @SuperFluffy's latest comments are addressed.

crates/astria-sequencer/src/mempool/mod.rs
```diff
@@ -181,17 +197,22 @@ impl Mempool {
...
-                let to_promote = parked.pop_front_account(
+                let to_promote = parked.find_promotables(
                     timemarked_tx.address(),
                     timemarked_tx
                         .nonce()
                         .checked_add(1)
                         .expect("failed to increment nonce in promotion"),
+                    &pending.subtract_contained_costs(
+                        *timemarked_tx.address(),
+                        current_account_balances.clone(),
```
SuperFluffy, line 208: Is cloning the entire hashmap here necessary?

crates/astria-sequencer/src/mempool/transactions_container.rs
```diff
@@ -189,6 +287,19 @@ impl TransactionsForAccount for PendingTransactionsForAccount {
         // is equal to the account nonce
         self.txs().contains_key(&previous_nonce) || ttx.signed_tx.nonce() == current_account_nonce
     }
+
+    fn has_balance_to_cover(
+        &self,
+        ttx: &TimemarkedTransaction,
+        current_account_balances: &HashMap<IbcPrefixed, u128>,
+    ) -> bool {
+        let mut current_account_balances = current_account_balances.clone();
```
SuperFluffy, line 296: This is a pretty innocent looking functions where one does not expect a deep copy to happen.

Lilyjjo: Ah, good catch. Will change!

Lilyjjo: Actually, this was done to prevent doing copies of the `account_balances` for the parked transactions container, which is a noop for the `has_balance_to_cover` fn. We can avoid making unnecessary copies if we do the clone on the inside for the pending container

SuperFluffy:
Given this being a breaking charge, please mark this patch as breaking using `!`: `feat(sequencer)!: make mempool balance aware`.

Also, please add a section `# Breaking Changelist` explaining:

1. why changes to the mempool constitute a breaking change?
2. why changes to the mempool affect the app hash?

PR #1409, opened 2024-08-27
fix(cli, bridge-withdrawer): dont fail entire block due to bad withdraw event

### LINKED ISSUE #1251: Bridge withdrawer shouldnt panic on bad single withdrawal address
The bridge withdrawer currently panics and stops running if there is an executed withdrawal with a bad astria address in its memo. This means the withdrawer will always panic when trying to execute the block with the bad withdrawal. Instead, we need to simply  drop the withdrawal that fails the address deserialization and continue with the next event.

itamarreif:
## Summary

## Background

## Changes
- Separate the errors for getting logs from the conversion errors
- Conversion returns a `Result<Vec<Result<Action, WithdrawalConversionError>>, GetWithdrawalLogsError> ` instead of `Result<Vec<Action>, GetWithdrawalActionsError>`
- Handle failed conversions in the cli and withdrawer crates

## Testing
- Unit tests for the conversions will be added in a subsequent PR: 

## Breaking Changelist
- This fixes the bug described in #1251, where a single withdrawal which fails to be converted to an action will cause the entire rollup block's batch of withdrawals to fail, causing the withdrawer to be stuck.

## Related Issues

closes #1251

### DIFF
--- a/crates/astria-bridge-contracts/src/lib.rs
+++ b/crates/astria-bridge-contracts/src/lib.rs
@@ -332,7 +332,7 @@ where
     pub async fn get_for_block_hash(
         &self,
         block_hash: H256,
-    ) -> Result<Vec<Action>, GetWithdrawalActionsError> {
+    ) -> Result<Vec<Result<Action, GetWithdrawalActionsError>>, GetWithdrawalActionsError> {
         use futures::FutureExt as _;
         let get_ics20_logs = if self.configured_for_ics20_withdrawals() {
             get_logs::<Ics20WithdrawalFilter, _>(&self.provider, self.contract_address, block_hash)
@@ -358,7 +358,7 @@ where
         // XXX: The calls to `log_to_*_action` rely on only be called if `GetWithdrawalActions`
         // is configured for either ics20 or sequencer withdrawals (or both). They would panic
         // otherwise.
-        ics20_logs
+        Ok(ics20_logs
             .into_iter()
             .map(|log| self.log_to_ics20_withdrawal_action(log))
             .chain(
@@ -366,7 +366,7 @@ where
                     .into_iter()
                     .map(|log| self.log_to_sequencer_withdrawal_action(log)),
             )
-            .collect()
+            .collect())
     }
 
     fn log_to_ics20_withdrawal_action(
--- a/crates/astria-bridge-withdrawer/src/bridge_withdrawer/ethereum/watcher.rs
+++ b/crates/astria-bridge-withdrawer/src/bridge_withdrawer/ethereum/watcher.rs
@@ -359,7 +359,7 @@ async fn get_and_forward_block_events(
         .filter_map(|r| {
             r.map_err(|e| {
                 warn!(
-                    error = &e as &dyn std::error::Error,
+                    error = %eyre::Report::new(error),
                     "failed to convert withdrawal event to sequecner action"
                 );
             })
--- a/crates/astria-bridge-withdrawer/src/bridge_withdrawer/ethereum/watcher.rs
+++ b/crates/astria-bridge-withdrawer/src/bridge_withdrawer/ethereum/watcher.rs
@@ -359,8 +359,8 @@ async fn get_and_forward_block_events(
         .filter_map(|r| {
             r.map_err(|e| {
                 warn!(
-                    error = %eyre::Report::new(error),
-                    "failed to convert withdrawal event to sequecner action"
+                    error = %eyre::Report::new(e),
+                    "failed to convert rollup withdrawal event to sequencer action; dropping"
                 );
             })
             .ok()
--- a/crates/astria-bridge-withdrawer/src/bridge_withdrawer/ethereum/watcher.rs
+++ b/crates/astria-bridge-withdrawer/src/bridge_withdrawer/ethereum/watcher.rs
@@ -7,9 +7,12 @@ use astria_bridge_contracts::{
     GetWithdrawalActions,
     GetWithdrawalActionsBuilder,
 };
-use astria_core::primitive::v1::{
-    asset,
-    Address,
+use astria_core::{
+    primitive::v1::{
+        asset,
+        Address,
+    },
+    protocol::transaction::v1alpha1::Action,
 };
 use astria_eyre::{
     eyre::{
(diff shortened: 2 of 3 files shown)


joroshiba (commented):
LGTM

crates/astria-bridge-contracts/src/lib.rs
```diff
@@ -353,47 +353,61 @@ where
...
+    /// - If the log cannot be decoded as a sequencer withdrawal event.
+    /// - If the log does not contain a `recipient` field.
+    /// - If the memo cannot be encoded to json.
+    /// - If calculating the amount using the asset withdrawal divisor overflows.
+    ///
+    /// # Panics
+    /// - If getting a withdrawal event type that it was not configured for. For example, if the
+    ///   configuration was specified to only get sequencer withdrawal events but an ICS20
+    ///   withdrawal is received.
+    pub fn log_to_ics20_withdrawal_action(
```
SuperFluffy, line 386: I don't see why this needs to be a pub method? I don't think it's used anywhere outside this lib? You can then also remove the doc comment.

crates/astria-bridge-contracts/src/lib.rs
```diff
@@ -467,54 +495,97 @@ where
...
 
-impl GetWithdrawalActionsError {
+impl GetWithdrawalLogsError {
+    fn get_logs(source: GetLogsError) -> Self {
+        Self(GetWithdrawalLogsErrorKind::GetLogs(source))
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+enum GetWithdrawalLogsErrorKind {
```
SuperFluffy, line 507: Making this an enum seems unnecessary?

crates/astria-bridge-contracts/src/lib.rs
```diff
@@ -467,54 +495,97 @@ where
 
 #[derive(Debug, thiserror::Error)]
 #[error(transparent)]
-pub struct GetWithdrawalActionsError(GetWithdrawalActionsErrorKind);
+pub struct GetWithdrawalLogsError(GetWithdrawalLogsErrorKind);
```
SuperFluffy, line 498: I think you are leaking concepts that don't appear at the high level. Please change this back.

crates/astria-bridge-contracts/src/lib.rs
```diff
@@ -424,34 +438,48 @@ where
...
+    /// - If the log does not contain a transaction hash.
+    /// - If the log cannot be decoded as a sequencer withdrawal event.
+    /// - If the log does not contain a `recipient` field.
+    /// - If the memo cannot be encoded to json.
+    /// - If calculating the amount using the asset withdrawal divisor overflows.
+    ///
+    /// # Panics
+    /// - If getting a withdrawal event type that it was not configured for. For example, if the
+    ///   configuration was specified to only get sequencer withdrawal events but an ICS20
+    ///   withdrawal is received.
```
SuperFluffy, line 454: Remove this doc comment.

crates/astria-cli/src/commands/bridge/collect.rs
```diff
@@ -199,10 +199,21 @@ async fn block_to_actions(
         .await
         .wrap_err_with(|| {
             format!(
-                "failed getting actions for block; block hash: `{block_hash}`, block height: \
+                "failed getting logs for block; block hash: `{block_hash}`, block height: \
```
SuperFluffy, line 202: This is not correct. The `actions_fetcher` is promising to give you actions. Why is this talking about logs? That's confusing.

crates/astria-cli/src/commands/bridge/collect.rs
```diff
@@ -199,10 +199,21 @@ async fn block_to_actions(
...
                  `{rollup_height}`"
             )
-        })?;
+        })?
+        .into_iter()
+        .filter_map(|r| {
```
SuperFluffy, line 207: I am ambivalent about changing this logic in the CLI at this point. I am fine with adding a flag that filters out bad actions, but I would leave this to a followup.

SuperFluffy (changes requested):
This PR contains a lot of renames that I don't think are necessary (and partially confusing, talking about "logs" where "actions" are promised).

crates/astria-bridge-withdrawer/src/bridge_withdrawer/ethereum/watcher.rs
```diff
@@ -348,10 +351,21 @@ async fn get_and_forward_block_events(
...
         .get_for_block_hash(block_hash)
         .await
-        .wrap_err("failed getting actions for block")?;
+        .wrap_err("failed getting actions for block")?
+        .into_iter()
+        .filter_map(|r| {
+            r.map_err(|e| {
+                warn!(
+                    error = &e as &dyn std::error::Error,
+                    "failed to convert withdrawal event to sequecner action"
```
SuperFluffy, line 363:
```suggestion
                    "failed to convert rollup withdrawal event to sequencer action; dropping"
```

crates/astria-cli/src/commands/bridge/collect.rs
```diff
@@ -202,7 +202,10 @@ async fn block_to_actions(
                 "failed getting actions for block; block hash: `{block_hash}`, block height: \
                  `{rollup_height}`"
             )
-        })?;
+        })?
+        .into_iter()
+        .collect::<Result<_, _>>()?;
```
SuperFluffy, line 207: Not the most efficient, but I guess it doesn't matter in this case.

crates/astria-bridge-withdrawer/src/bridge_withdrawer/ethereum/watcher.rs
```diff
@@ -348,10 +351,21 @@ async fn get_and_forward_block_events(
...
+    let actions: Vec<Action> = actions_fetcher
         .get_for_block_hash(block_hash)
         .await
-        .wrap_err("failed getting actions for block")?;
+        .wrap_err("failed getting actions for block")?
+        .into_iter()
+        .filter_map(|r| {
+            r.map_err(|e| {
+                warn!(
+                    error = &e as &dyn std::error::Error,
```
SuperFluffy, line 362: Use our error encoding here:
```suggestion
                    error = %eyre::Report::new(error),
```

SuperFluffy (approved):
Nice, this is a much neater change. Thank you!

The message in the error event looks a bit wonky. See my suggestion.

PR #1410, opened 2024-08-27
feat(proto, core, sequencer)!: add traceability to rollup deposits

### LINKED ISSUE #1402: Add traceability to rollup deposit source
Currently the deposit events created have no information about deposit source, this poses an issue for tracing if your deposit has landed without simply monitoring the account balance. It additionally creates an issue in the EVM where all deposit txs of the same value to the same account for the same asset create the same tx hash, which is again subpar for monitoring.

In order to fix this we should propogate the tx hash source through the deposit, similar to how on rollup withdrawals we track the withdrawal event id. 

To do this we need to update the proto [`Deposit`](https://github.com/astriaorg/astria/blob/main/proto/sequencerblockapis/astria/sequencerblock/v1alpha1/block.proto#L76)  to include a `string source_transaction_hash` field, as well as updatess to the `astria-core` types and filling in during deposit creation.

After this is done as a follow up changes should be made to [astria-geth](https://github.com/astriaorg/astria-geth/) to consume the new information as a part of the deposit tx.

┆Issue Number: ENG-744

ethanoroshiba:
## Summary
Added `id_of_source_transaction` and `position_in_source_transaction` to `Deposit`.

## Background
Previous deposit events had no information about the deposit source, so rollups could not trace if their deposits had landed without checking the balance. Additionally, this could result in deposit txs having the same hash in the EVM.

## Changes
- Created primitive proto type `TransactionId` for identifying source transaction in deposits.
- Added `id_of_source_transaction` and `position_in_source_transaction` to `Deposit`.
- Included tx id in the ephemeral state via `current_source()` so that it can be accessed by actions.
- Added accessors/mutators for deposit index to `StateReadExt` and `StateWriteExt` traits in `transaction`.
- Updated `bridge_lock` and `ics20_transfer` actions to populate `Deposit` events with transaction id and index based on the state.

## Testing
Added test to ensure `position_in_source_transaction` is incremented correctly.

## Breaking Changelist
- App hash changed because of increasing block fees due to longer length of `Deposit`.
- Changed SequencerBlock API `Deposit` event, having two new mandatory fields `source_transaction_id` and `source_action_index`. `source_transaction_id` uses a new primitive proto type `TransactionId`.

## Related Issues
with https://github.com/astriaorg/astria-geth/pull/42, closes #1402


crates/astria-sequencer/src/bridge/bridge_lock_action.rs
```diff
@@ -72,12 +76,24 @@ impl ActionHandler for BridgeLockAction {
             .await
             .context("failed to get transfer base fee")?;
 
+        let transaction_hash = state
+            .get_current_source()
+            .ok_or(anyhow!("expected current source to be `Some`"))?
```
noot, line 81: the transaction source should always be set, right? if so, you can `expect("current transaction source must be set")`

crates/astria-sequencer/src/bridge/bridge_lock_action.rs
```diff
@@ -72,12 +76,24 @@ impl ActionHandler for BridgeLockAction {
...
+            .transaction_hash;
+        let source_transaction_index = state
+            .get_transaction_deposit_index()
+            .await
+            .context("failed to get transaction deposit index")?
+            .ok_or(anyhow!("expected transaction deposit index to be `Some`"))?;
```
noot, line 87: same here

noot (commented):
overall this looks good! i'm wondering if it makes more sense to include the `action_index` instead of the `deposit_index` in the transaction, ie. the index of the action within the tx. for example, ics20 transfers don't necessarily create a deposit, so if a transaction contains one ics20 transfer that does create a deposit and one that doesn't, it might be unclear as to which one generated the deposit.

ethanoroshiba:
> overall this looks good! i'm wondering if it makes more sense to include the `action_index` instead of the `deposit_index` in the transaction, ie. the index of the action within the tx. for example, ics20 transfers don't necessarily create a deposit, so if a transaction contains one ics20 transfer that does create a deposit and one that doesn't, it might be unclear as to which one generated the deposit.

@noot I think this makes sense, too. I just implemented the same thing for `tx.fees` events, so it would make sense to also have an `actionIndex` for `Deposit` actions.

crates/astria-sequencer/Cargo.toml
```diff
@@ -38,6 +38,7 @@ tower-http = { version = "0.4", features = ["cors"] }
 async-trait = { workspace = true }
 bytes = { workspace = true }
 divan = { workspace = true, optional = true }
+ethers = { workspace = true, features = ["ws"] }
```
joroshiba, line 41: We don't want to have this as a dependency. It's fine if the hex transaction hashes don't have "0x" in front of them.

crates/astria-sequencer/src/ibc/ics20_transfer.rs
```diff
@@ -1052,7 +1087,7 @@ mod test {
                 memo: String::new(),
                 rollup_block_number: 1,
                 rollup_return_address: "rollup-defined".to_string(),
-                rollup_transaction_hash: hex::encode([1u8; 32]),
+                rollup_transaction_hash: [1u8; 32].encode_hex_with_prefix(),
```
joroshiba, line 1090: the format here doesn't really matter, it's a string. Also a very seperate concern from the deposits

ethanoroshiba: Removed in 1fd6ad2. Thanks for taking a look over!

crates/astria-sequencer/Cargo.toml
ethanoroshiba: Removed and reverted back to `hex::encode` in 1fd6ad2.

SuperFluffy (changes requested):
This leaves too many implementation details to sequencer. We want to also have the astria cli have the same concept of what a transaction ID is (and potentially other parts of the stack).

IMO we should type out (in protobuf) a transaction ID type. Then provide a `SignedTransaction::id -> Id` method that generates an ID from it (replacing the current `sha256_from_protobuf`).

proto/sequencerblockapis/astria/sequencerblock/v1alpha1/block.proto
```diff
@@ -86,6 +86,11 @@ message Deposit {
   // the address on the destination chain which
   // will receive the bridged funds
   string destination_chain_address = 5;
+  // the hash of the transaction which initiated the deposit
+  string source_transaction_hash = 6;
```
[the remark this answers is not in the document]
ethanoroshiba: New type created in cf70451. Astria-geth has been updated as well. Thanks for the suggestion!

proto/primitives/astria/primitive/v1/types.proto
```diff
@@ -55,3 +55,10 @@ message Address {
   reserved 1;
   reserved "inner";
 }
+
+// A `TransactionId` is a unique identifier for a transaction.
+// It contains the hash of the transaction, to be included in
+// rollup deposit events for source tracking.
+message TransactionId {
+  string hash = 1;
```
SuperFluffy, line 63: Why string over `bytes`? So for identifiers (like rollup ID) that are `bytes` (and hence use base64 encoding).

proto/primitives/astria/primitive/v1/types.proto
```diff
@@ -55,3 +55,10 @@ message Address {
...
+}
```
SuperFluffy: newline

crates/astria-core/src/primitive/v1/mod.rs
```diff
@@ -488,6 +488,103 @@ where
     tree
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Hash, PartialOrd, Ord)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize))]
+#[cfg_attr(feature = "serde", serde(transparent))]
```
SuperFluffy, line 493: This should serialize like the protobuf type:

```suggestion
#[cfg_attr(
    feature = "serde",
    serde(try_from = "raw::TransactionId", into = "raw::TransactionId"),
)]
```

crates/astria-core/src/primitive/v1/mod.rs
```diff
@@ -488,6 +488,103 @@ where
...
+    ///
+    /// # Examples
+    /// ```
+    /// use astria_core::primitive::v1::TransactionId;
+    /// let hash = hex::encode([0u8; 32]);
+    /// let transaction_id = TransactionId::new(hash);
+    /// assert_eq!(hash, transaction_id.get());
+    /// ```
+    #[must_use]
+    pub const fn new(hash: String) -> Self {
```
SuperFluffy, line 509: Remove. One should not be able to construct arbitrary transaction IDs.

crates/astria-core/src/primitive/v1/mod.rs
```diff
@@ -488,6 +488,103 @@ where
     tree
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Hash, PartialOrd, Ord)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize))]
+#[cfg_attr(feature = "serde", serde(transparent))]
+pub struct TransactionId {
+    hash: String,
```
SuperFluffy, line 495: There is no need to make this a `String`. Keep this as a `[u8; 32]` and decide in the `Display` implementation what this should format to.

crates/astria-core/src/primitive/v1/mod.rs
SuperFluffy, line 1: High level feedback: transaction IDs should be checked to match a specification. The definitions here basically make it a newtype wrapper without extra constraints.

I think it should have exactly 2 entry point:

1. using the `SignedTransaction::id` method;
2. using `try_from_raw`/`try_from_raw_ref` when validating the wire protobuf type.

crates/astria-core/src/protocol/transaction/v1alpha1/mod.rs
```diff
@@ -102,6 +103,11 @@ impl SignedTransaction {
         Sha256::digest(bytes).into()
     }
 
+    #[must_use]
+    pub fn id(&self) -> TransactionId {
+        hex::encode(self.sha256_of_proto_encoding()).into()
```
SuperFluffy, line 108: With this we can remove `sha256_of_proto_encoding`. Just rename `s/sha256_of_proto_encoding/id` and let it return a `TransactionId`.

Also note my comment above: don't turn this into a hex-encoding string at this point.

crates/astria-core/src/primitive/v1/mod.rs
```diff
@@ -488,6 +488,103 @@ where
...
+        raw::TransactionId {
+            hash: self.hash,
+        }
+    }
+
+    /// Converts from protobuf type to rust type for a transaction ID.
+    #[must_use]
+    pub fn from_raw(raw: &raw::TransactionId) -> Self {
+        Self {
+            hash: raw.hash.clone(),
```
SuperFluffy, line 547: This does not ensure that the data that comes over the wire is indeed a hex encoded slice of 32 bytes.

crates/astria-core/src/primitive/v1/mod.rs
```diff
@@ -488,6 +488,103 @@ where
...
+    }
+}
+
+impl From<&TransactionId> for TransactionId {
+    fn from(value: &TransactionId) -> Self {
+        Self {
+            hash: value.hash.clone(),
+        }
+    }
+}
```
SuperFluffy, line 580: `From<&TransactionId> for TransactionId` is just `Clone`.

The `From<String>` and `From<&String>` impls are wrong (for the same reason as the constructor is wrong; the strings are not checked to actually contain a valid ID).

The `AsRef<String> for TransactionId` should be `AsRef<str>` (since `AsRef<String>` is pretty useless), but should IMO be removed entirely.

crates/astria-core/src/sequencerblock/v1alpha1/block.rs
```diff
@@ -1318,17 +1320,17 @@ impl Deposit {
         amount: u128,
         asset: asset::Denom,
         destination_chain_address: String,
-        source_transaction_hash: String,
-        source_transaction_index: u32,
+        transaction_id: TransactionId,
+        index_of_action: u32,
```
SuperFluffy, line 1324: Maybe the following?
```suggestion
        position_in_source_action: u32,
```

proto/sequencerblockapis/astria/sequencerblock/v1alpha1/block.proto
```diff
@@ -86,10 +86,11 @@ message Deposit {
...
   // will receive the bridged funds
   string destination_chain_address = 5;
-  // the hash of the transaction which initiated the deposit
-  string source_transaction_hash = 6;
+  // the transaction ID of the source action for the deposit, consisting
+  // of the transaction hash.
+  astria.primitive.v1.TransactionId transaction_id = 6;
   // index of the deposit's source action within its transaction
-  uint32 source_transaction_index = 7;
+  uint32 index_of_action = 7;
```
SuperFluffy, line 93: Maybe this?
```suggestion
  uint64 position_in_source_transaction  = 7;
```
Comment on `uint64` vs `uint32`: this is to make your life easier: `astria-core` enforces that it only compiles on 64 bit architectures, so you can safely convert from `usize` to `uint64` without having to worry. The size on the wire is the same (for any reasonable numbers) due to varint encoding.

proto/sequencerblockapis/astria/sequencerblock/v1alpha1/block.proto
```diff
@@ -86,10 +86,11 @@ message Deposit {
   // the address on the destination chain which
   // will receive the bridged funds
   string destination_chain_address = 5;
-  // the hash of the transaction which initiated the deposit
-  string source_transaction_hash = 6;
+  // the transaction ID of the source action for the deposit, consisting
+  // of the transaction hash.
+  astria.primitive.v1.TransactionId transaction_id = 6;
```
SuperFluffy, line 91: Just an idea?
```suggestion
    astria.primitive.v1.TransactionId id_of_source_transaction = 6;
```

crates/astria-sequencer/src/bridge/bridge_lock_action.rs
```diff
@@ -72,24 +72,24 @@ impl ActionHandler for BridgeLockAction {
...
-        let source_transaction_index = state
-            .get_transaction_action_index()
+            .transaction_id;
+        let index_of_action = state
+            .get_transaction_index_of_action()
             .await
-            .context("failed to get transaction action index")?
-            .expect("action index should be `Some`");
+            .context("failed to get transaction index of action")?
+            .expect("index_of_action should be `Some`");
```
SuperFluffy, line 83: I know you just adapted these panic messages, but they are kinda useless. If `index_of_action` is `None`, then you get the message "it should be `Some`". This kinda leaves it open as to why it should be set.

crates/astria-core/src/primitive/v1/mod.rs
ethanoroshiba: I tried to incorporate everything you said after reverting back to using `String` instead of `[u8; 32]`. There are a couple more "entry points" than just those two, but I believe all of them should be checked for being hex encoded (if applicable) and 32 bytes. A couple are required by `serde::Serialize` and `serde::Deserialize` and then a couple I put in just because they are useful (such as the from trait for `[u8; 32]`. Hopefully this addresses all the concerns, please let me know if there's anything else I need to do! I appreciate all the feedback

proto/primitives/astria/primitive/v1/types.proto
SuperFluffy: Discussed offline: in cosmos land hex encoded identifiers are standard for identifying transactions. Furthermore, base64 poses issues in URLs. So using `string` here is ultimately a nod toward end users.

crates/astria-core/src/primitive/v1/mod.rs
```diff
@@ -488,6 +490,104 @@ where
...
+
+impl std::fmt::Display for TransactionId {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.hash)
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum TransactionIdError {
```
SuperFluffy, line 584: Please follow the rest of the codebase and keep the error variants private. Define this:

```rust
#[error(transparent)]
pub struct TransactionIdErrror(TransactionIdErrorKind);

enum TransactionIdErrorKind {
}
```

crates/astria-core/src/primitive/v1/mod.rs
SuperFluffy:
```suggestion
    inner: [u8; 32],
```

proto/primitives/astria/primitive/v1/types.proto
SuperFluffy, line 63: IMO prefer to call this `string inner = 1;` or even `hex inner = 1;` Saying that this is a hash (of what?) is not really important.

Please add which restrictions this field has (namely, 32 bytes lower-hex encoded).

crates/astria-core/src/primitive/v1/mod.rs
SuperFluffy: Ideally there should be only 1 entry point. Other entry points should be implemented in terms of it.

Also the internal representation should be different than what comes over the wire, so the following definition is more semantically correct:

```rust
struct TransactionId {
    inner: [u8;TRANSACTION_ID_LENGTH],
}
```

crates/astria-core/src/primitive/v1/mod.rs
```diff
@@ -488,6 +490,104 @@ where
...
+        raw::TransactionId {
+            hash: self.hash.clone(),
+        }
+    }
+
+    /// Returns a copy of the transaction hash as a 32-byte array.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the transaction ID is not 32 bytes long. This should never happen, as the
```
SuperFluffy, line 528: There is no need for this if you represent this internally as `[u8; 32]`

crates/astria-core/src/primitive/v1/mod.rs
```diff
@@ -488,6 +490,104 @@ where
...
+    }
+}
+
+impl From<[u8; TRANSACTION_ID_LENGTH]> for TransactionId {
+    fn from(inner: [u8; TRANSACTION_ID_LENGTH]) -> Self {
+        Self {
+            hash: hex::encode(inner),
+        }
+    }
+}
```
SuperFluffy, line 575: I am skeptical about this.

crates/astria-core/src/primitive/v1/mod.rs
```diff
@@ -488,6 +490,104 @@ where
...
+
+    fn try_from(hash: String) -> Result<Self, Self::Error> {
+        let inner =
+            hex::decode(hash.clone()).map_err(|err| TransactionIdError::HexDecodeError {
+                source: err,
+            })?;
+        if inner.len() != TRANSACTION_ID_LENGTH {
+            return Err(TransactionIdError::IncorrectTransactionIdLength {
+                received: inner.len(),
+            });
```
SuperFluffy, line 561: These should recided in the `try_from_raw_ref` impl, but you can also the following trait to just do `<[u8; TRANSACTION_ID_LEN]>::from_hex(s)`: https://docs.rs/hex/0.4.3/hex/trait.FromHex.html

crates/astria-core/src/protocol/transaction/v1alpha1/mod.rs
```diff
@@ -609,7 +610,7 @@ mod test {
             transaction_bytes: unsigned.to_raw().encode_to_vec().into(),
         };
 
-        insta::assert_json_snapshot!(tx.sha256_of_proto_encoding());
+        insta::assert_json_snapshot!(tx.id().as_bytes());
```
SuperFluffy, line 613: Don't assert on the raw bytes, assert on the json representation of the identifier which will make this much nicer:
```suggestion
        insta::assert_json_snapshot!(tx.id());
```

crates/astria-composer/src/executor/mod.rs
```diff
@@ -689,7 +689,7 @@ impl Future for SubmitFut {
                     info!(
                         nonce.actual = *this.nonce,
                         bundle = %telemetry::display::json(&SizedBundleReport(this.bundle)),
-                        transaction.hash = %telemetry::display::hex(&tx.sha256_of_proto_encoding()),
+                        transaction.hash = tx.id().get(),
```
SuperFluffy, line 692: This would give you a very ugly string of the form `[2, 5, 3441, 331, ...]`. Instead, use the display implementation that you defined for the ID:

```suggestion
                        transaction.id = %tx.id(),
```

crates/astria-composer/src/executor/mod.rs
```diff
@@ -767,7 +767,7 @@ impl Future for SubmitFut {
                         info!(
                             nonce.resubmission = *this.nonce,
                             bundle = %telemetry::display::json(&SizedBundleReport(this.bundle)),
-                            transaction.hash = %telemetry::display::hex(&tx.sha256_of_proto_encoding()),
+                            transaction.hash = tx.id().get(),
```
SuperFluffy, line 770: As above:
```suggestion
                        transaction.id = %tx.id(),
```

crates/astria-bridge-withdrawer/src/bridge_withdrawer/submitter/mod.rs
```diff
@@ -221,7 +221,7 @@ fn report_exit(reason: eyre::Result<&str>) {
     skip_all,
     fields(
         nonce = tx.nonce(),
-        transaction.hash = %telemetry::display::hex(&tx.sha256_of_proto_encoding()),
+        transaction.hash = tx.id().get(),
```
SuperFluffy, line 224:
```suggestion
                        transaction.id = %tx.id(),
```

crates/astria-bridge-withdrawer/src/bridge_withdrawer/submitter/mod.rs
```diff
@@ -164,7 +164,7 @@ impl Submitter {
 
         // sign transaction
         let signed = unsigned.into_signed(signer.signing_key());
-        debug!(tx_hash = %telemetry::display::hex(&signed.sha256_of_proto_encoding()), "signed transaction");
+        debug!(tx_hash = &signed.id().get(), "signed transaction");
```
SuperFluffy, line 167: IMO call this an ID like in all other places:
```suggestion
        debug!(transaction.id = %signed.id(), "signed transaction");
```

crates/astria-bridge-withdrawer/src/bridge_withdrawer/startup.rs
```diff
@@ -480,7 +480,7 @@ fn rollup_height_from_signed_transaction(
     .expect("action is already checked to be either BridgeUnlock or Ics20Withdrawal");
 
     info!(
-        last_batch.tx_hash = %telemetry::display::hex(&signed_transaction.sha256_of_proto_encoding()),
+        last_batch.tx_hash = signed_transaction.id().get(),
```
SuperFluffy, line 483:
```suggestion
        last_batch.transaction_id = %signed_transaction.id(),
```

crates/astria-core/src/primitive/v1/mod.rs
```diff
@@ -488,6 +490,104 @@ where
     tree
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Hash, PartialOrd, Ord)]
```
SuperFluffy, line 493: The transaction ID should be `Copy` (which it can be if you change the internal type to `[u8; 32]`).

crates/astria-core/src/primitive/v1/mod.rs
```diff
@@ -22,6 +22,8 @@ pub const ADDRESS_LEN: usize = 20;
 
 pub const ROLLUP_ID_LEN: usize = 32;
 
+pub const TRANSACTION_ID_LENGTH: usize = 32;
```
SuperFluffy, line 25: Folllow the convention of the constant right above it.
```suggestion
pub const TRANSACTION_ID_LEN: usize = 32;
```

crates/astria-core/src/primitive/v1/mod.rs
```diff
@@ -488,6 +490,104 @@ where
...
+    serde(try_from = "raw::TransactionId", into = "raw::TransactionId")
+)]
+pub struct TransactionId {
+    hash: String,
+}
+
+impl TransactionId {
+    /// Returns the hex-encoded 32-byte transaction hash as a string.
+    #[must_use]
+    pub fn get(&self) -> String {
```
SuperFluffy, line 506: This would become `pub fn get(self) -> [u8; 32]`.

crates/astria-core/src/sequencerblock/v1alpha1/block.rs
```diff
@@ -1395,12 +1412,19 @@ impl Deposit {
         let rollup_id =
             RollupId::try_from_raw(&rollup_id).map_err(DepositError::incorrect_rollup_id_length)?;
         let asset = asset.parse().map_err(DepositError::incorrect_asset)?;
+        let Some(id_of_source_transaction) = id_of_source_transaction else {
+            return Err(DepositError::field_not_set("transaction_id"));
+        };
+        let id_of_source_transaction = TransactionId::try_from(id_of_source_transaction)
```
SuperFluffy, line 1418: Use `try_from_raw` or `try_from_raw_ref`

crates/astria-core/src/sequencerblock/v1alpha1/block.rs
```diff
@@ -1439,6 +1467,8 @@ enum DepositErrorKind {
     IncorrectRollupIdLength(#[source] IncorrectRollupIdLength),
     #[error("the `asset` field could not be parsed")]
     IncorrectAsset(#[source] asset::ParseDenomError),
+    #[error("error converting to `TransactionId`")]
```
SuperFluffy, line 1470: Don't leak the type that you are converting to here; the following is much more descriptive to a reader:
```suggestion
    #[error("field `id_of_source_transaction` was invalid")]
```

crates/astria-sequencer/src/accounts/state_ext.rs
```diff
@@ -244,9 +244,10 @@ pub(crate) trait StateWriteExt: StateWrite {
         self.put_account_balance(
             &address,
             asset,
-            balance
-                .checked_sub(amount)
-                .context("subtracting from account balance failed due to insufficient funds")?,
+            balance.checked_sub(amount).context(format!(
+                "subtracting from account balance failed due to insufficient funds; balance: \
+                 {balance}, subtraction amount: {amount}",
+            ))?,
```
SuperFluffy, line 250: It looks like this change has nothing to do with the rest of the PR? Can you move it to a new PR?

crates/astria-sequencer/src/api_state_ext.rs
```diff
@@ -412,6 +412,8 @@ mod test {
                 amount,
                 asset,
                 destination_chain_address,
+                [0; 32].into(),
```
SuperFluffy, line 415: I see why you needed the `From` impl. This entire construction here is iffy and should be reworked, but for the time being you should be able to side-step the issue by just populating a protobuf `raw::Deposit` and converting that.

crates/astria-core/src/sequencerblock/v1alpha1/block.rs
```diff
@@ -1314,13 +1321,17 @@ impl Deposit {
         amount: u128,
         asset: asset::Denom,
         destination_chain_address: String,
+        id_of_source_transaction: TransactionId,
```
SuperFluffy, line 1324: Comment on the containing `Deposit::new`: this constructor is becoming unwieldy. I am fine if you keep it as is for this PR, but can you remove `Deposit::new` in favor of `Deposit::try_from_raw` (which is mostly done in tests anyway)?

crates/astria-sequencer/src/mempool/transactions_container.rs
SuperFluffy, line 1: As mentioned offline: it would IMO be very elegant to rename `tx_hash -> transaction_id` and also change its type to `TranasctionId`.

crates/astria-sequencer/src/transaction/state_ext.rs
```diff
@@ -39,12 +53,53 @@ pub(crate) trait StateWriteExt: StateWrite {
...
+            index_of_action_storage_key().as_bytes().to_vec(),
+            borsh::to_vec(&index).expect("serialize index of action"),
+        );
+    }
+
+    #[instrument(skip_all)]
+    async fn increment_transaction_index_of_action(&mut self) -> Result<()> {
+        let index = self
+            .get_transaction_index_of_action()
+            .await?
```
SuperFluffy, line 69: context

crates/astria-sequencer/src/transaction/mod.rs
```diff
@@ -169,23 +169,22 @@ impl ActionHandler for SignedTransaction {
...
             .is_some()
         {
-            state.put_last_transaction_hash_for_bridge_account(
-                self,
-                &self.sha256_of_proto_encoding(),
-            );
+            state.put_last_transaction_id_for_bridge_account(self, &self.id());
         }
 
+        state.put_transaction_index_of_action(0);
```
SuperFluffy, line 181: Why does this need to be written to state?

crates/astria-sequencer/src/transaction/mod.rs
```diff
@@ -169,23 +169,22 @@ impl ActionHandler for SignedTransaction {
...
+
+        // Should have enough balance to cover all actions.
+        check_balance_for_total_fees_and_transfers(self, &state)
```
SuperFluffy, line 184: This looks like it has nothing to do with tra ing the rollup deposits?

crates/astria-sequencer/src/transaction/mod.rs
```diff
@@ -269,6 +268,10 @@ impl ActionHandler for SignedTransaction {
                     .await
                     .context("failed executing bridge sudo change")?,
             }
+            state
+                .increment_transaction_index_of_action()
```
SuperFluffy, line 272: As above, why is it necessary to write this to state?

Can this just reside in the ephemeral state instead?

crates/astria-sequencer/src/transaction/checks.rs
```diff
@@ -261,6 +268,7 @@ fn bridge_lock_update_fees(
     fees_by_asset: &mut HashMap<asset::IbcPrefixed, u128>,
     transfer_fee: u128,
     bridge_lock_byte_cost_multiplier: u128,
+    tx_index_of_action: &mut u64,
```
SuperFluffy, line 271: Why is this a mutable reference?

crates/astria-sequencer/src/transaction/checks.rs
```diff
@@ -100,6 +100,7 @@ pub(crate) async fn get_fees_for_transaction<S: StateRead>(
         .context("failed to get bridge sudo change fee")?;
 
     let mut fees_by_asset = HashMap::new();
+    let mut tx_index_of_action = 0;
     for action in &tx.actions {
```
SuperFluffy, line 104: Don't manually increment a counter. Use `tx.actions.iter().enumerate()` instead.

crates/astria-sequencer/src/transaction/checks.rs
```diff
@@ -119,12 +120,15 @@ pub(crate) async fn get_fees_for_transaction<S: StateRead>(
...
-                transfer_fee,
-                bridge_lock_byte_cost_multiplier,
-            ),
+            Action::BridgeLock(act) => {
+                bridge_lock_update_fees(
+                    act,
+                    &mut fees_by_asset,
+                    transfer_fee,
+                    bridge_lock_byte_cost_multiplier,
+                    &mut tx_index_of_action,
```
SuperFluffy, line 129: I don't understand the mutable reference to the index.

crates/astria-sequencer/src/transaction/state_ext.rs
```diff
@@ -39,12 +53,53 @@ pub(crate) trait StateWriteExt: StateWrite {
...
+        let index = index.checked_add(1).expect("increment index of action");
+        self.nonverifiable_put_raw(
+            index_of_action_storage_key().as_bytes().to_vec(),
+            borsh::to_vec(&index).expect("serialize index of action"),
+        );
+        Ok(())
+    }
+
+    #[instrument(skip_all)]
+    fn clear_transaction_index_of_action(&mut self) {
```
SuperFluffy, line 80: See my comment above in regards to the ephemeral transaction <> action context.

crates/astria-sequencer/src/mempool/transactions_container.rs
ethanoroshiba: Ah, yes I agree. I had reached out to Lily once I started implementing this, though, and she asked to hold off on making this change until later since it may interfere with her work in mempool.

crates/astria-sequencer/src/transaction/state_ext.rs
```diff
@@ -39,12 +53,53 @@ pub(crate) trait StateWriteExt: StateWrite {
     fn delete_current_source(&mut self) {
         self.object_delete(current_source());
     }
+
+    #[instrument(skip_all)]
+    fn put_transaction_index_of_action(&mut self, index: u64) {
```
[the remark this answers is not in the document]
ethanoroshiba: While the abstraction may be nice, I wonder how the actual implementation would look. Similar to how `TransactionContext` is constructed from the transaction itself, it seems to me that it's implied `ActionContext` would be constructed from the action itself, which I believe would be problematic because the action does not contain the context of its position within the transaction. If only the transaction has this context, then IMO it seems most fitting to include it in `TransactionContext`, but given that it has to be updated every action, that would require may more reads/writes of the `TransactionContext` object. That's why I had originally separated the index from it, but if you had a different idea for how to implement `ActionContext` I'd be open to it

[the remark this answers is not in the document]
SuperFluffy: You are right - the action does not have that context, the transaction does. So indeed it should be added there.

For the ephemeral cache there are actually no reads and writes. If you look at the implementation of `object_put` (or `put_object` or whatever it's called), then `TransactionContext` is put into a `HashMap<String, Box<dyn Any>>`, with the value being downcast to a `TransactionContext`. So this is strictly cheaper to writing and reading from the non-verifiable cache.

crates/astria-core/src/primitive/v1/mod.rs
```diff
@@ -490,46 +491,83 @@ where
...
     }
 
-    /// Returns a copy of the transaction hash as a 32-byte array.
+    /// Convert from a raw protobuf type to a rust type for a transaction ID.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the transaction ID buffer was not 32 bytes long or if it was not hex
+    /// encoded.
+    pub fn try_from_raw(raw: raw::TransactionId) -> Result<Self, TransactionIdError> {
```
SuperFluffy, line 539: You can implement this in terms of `try_from_raw_ref` because we don't need to clone anything. Less code duplication.

crates/astria-core/src/primitive/v1/mod.rs
```diff
@@ -490,46 +491,83 @@ where
...
+        Ok(Self {
+            inner,
+        })
+    }
+
+    /// Returns a copy of the 32-byte transaction hash as a hex-encoded string.
     #[must_use]
-    pub fn as_bytes(&self) -> [u8; TRANSACTION_ID_LENGTH] {
-        hex::decode(self.hash.clone()).unwrap().try_into().unwrap()
+    pub fn as_hex(&self) -> String {
```
SuperFluffy, line 569: Remove. You have already defined `Display`, which does exactly this when you do `to_string()`.

crates/astria-sequencer/src/transaction/state_ext.rs
```diff
@@ -27,6 +23,7 @@ fn current_source() -> &'static str {
 pub(crate) struct TransactionContext {
     pub(crate) address_bytes: [u8; ADDRESS_LEN],
     pub(crate) transaction_id: TransactionId,
+    pub(crate) position_in_source_transaction: u64,
```
SuperFluffy, line 26: One last change: make this `Option<u64>`

crates/astria-sequencer/src/transaction/state_ext.rs
```diff
@@ -55,51 +53,23 @@ pub(crate) trait StateWriteExt: StateWrite {
...
-        let index = self
-            .get_transaction_index_of_action()
-            .await?
-            .expect("index of action should be set before incrementing it");
-        let index = index.checked_add(1).expect("increment index of action");
-        self.nonverifiable_put_raw(
-            index_of_action_storage_key().as_bytes().to_vec(),
-            borsh::to_vec(&index).expect("serialize index of action"),
-        );
+    fn increment_position_in_source_transaction(&mut self) -> Result<()> {
```
SuperFluffy, line 56: Don't increment, but instead provide a method `set_index_of_current_action` (or something like that).

crates/astria-sequencer/src/transaction/mod.rs
```diff
@@ -269,9 +267,8 @@ impl ActionHandler for SignedTransaction {
                     .context("failed executing bridge sudo change")?,
             }
             state
-                .increment_transaction_index_of_action()
-                .await
-                .context("failed to increment index of action")?;
+                .increment_position_in_source_transaction()
```
SuperFluffy, line 270: Commenting here because I can't comment on the loop above: we can express this more elegantly if we do the following changes:

1. change `put_current_source -> put_transaction_context<T: Into<TransactionContext>>(value: T) -> TransactionContext`
2. Provide `impl<'a> From<&'a SignedTransaction> for TransactionContext`
3. Add `#[derive(Copy)]` on `TransactionContext` (it only contains `Copy` types I think)
4. now you can get the initial transaction context from the signed tx `let mut transaction_context = state.put_transaction_context(self)`;
5. And instead of using `increment_position_in_source_transaction` after every iteration, use a more elegant iterator like this:

```rust
for (i, action) in (0..).zip(self.actions().iter()) {
    transaction_context.position_in_source_transaction = i;
    self.put_transaction_context(transaction_context);
}
```

Note that this relies on `impl<T> From<T> for T` being always present.

crates/astria-sequencer/src/transaction/state_ext.rs
```diff
@@ -26,19 +36,35 @@ impl From<&SignedTransaction> for TransactionContext {
...
         self.object_put(current_source(), context);
+        context
     }
 
     fn delete_current_source(&mut self) {
         self.object_delete(current_source());
     }
+
+    #[instrument(skip_all)]
+    fn set_position_in_source_transaction(&mut self, val: u64) -> Result<()> {
```
SuperFluffy, line 60: I think this can be removed now.

crates/astria-sequencer/src/ibc/ics20_transfer.rs
```diff
@@ -625,12 +625,24 @@ async fn execute_deposit<S: ibc::StateWriteExt>(
         "asset ID is not authorized for transfer to bridge account",
     );
 
+    let transaction_id = state
+        .get_current_source()
+        .expect("current source should be set before executing action")
+        .transaction_id;
+    let index_of_action = state
+        .get_current_source()
```
SuperFluffy, line 633: This reads the current source twice for now good reason.

crates/astria-sequencer/src/transaction/state_ext.rs
```diff
@@ -26,19 +36,35 @@ impl From<&SignedTransaction> for TransactionContext {
...
+        let mut context = self
+            .get_current_source()
+            .context("failed to get current source")?;
+        context.position_in_source_transaction = Some(val);
+        self.object_put(current_source(), context);
+        Ok(())
+    }
 }
 
 pub(crate) trait StateReadExt: StateRead {
```
SuperFluffy, line 70: Commenting on the line below: rename `get_current_source` -> `get_transaction_context`

crates/astria-core/src/sequencerblock/v1alpha1/block.rs
SuperFluffy: `try_from_raw` please

crates/astria-sequencer/src/transaction/checks.rs
SuperFluffy: Recommenting: the intention was to do the below instead of manually driving the iterator
```suggestion
    for (i, act) in tx.actions().iter().enumerate() {
```

crates/astria-core/src/sequencerblock/v1alpha1/block.rs
SuperFluffy: Can you make an issue for this to fix in a followup?

crates/astria-core/src/primitive/v1/mod.rs
```diff
@@ -8,6 +8,7 @@ use base64::{
     prelude::BASE64_STANDARD,
 };
 use bytes::Bytes;
+use hex::FromHex;
```
SuperFluffy, line 11: nit: move this import into the `try_from_raw_ref`, and also `use hex::FromHex as _;`

SuperFluffy (approved):
Thank you for all the work and bearing with my comments. This looks great (and I think we have reduced the new lines by 200).

crates/astria-core/src/sequencerblock/v1alpha1/block.rs
ethanoroshiba: Do you mean to remove `Deposit::new` entirely? Or just change the instances which are in tests to use `Deposit::try_from_raw`? The `new()` method is used in some actions for creating deposits, so it seems to me like it may be of some use to keep it for these instances, and if we already have it then IMO we may just as well use it for tests. Apologies if I'm misunderstanding what you're asking

SuperFluffy: The problem is with the method itself: a constructor with 6+ arguments is just terrible and is usually replaced by a builder.

proto/sequencerblockapis/astria/sequencerblock/v1alpha1/block.proto
```diff
@@ -86,6 +86,11 @@ message Deposit {
   // the address on the destination chain which
   // will receive the bridged funds
   string destination_chain_address = 5;
+  // the transaction ID of the source action for the deposit, consisting
+  // of the transaction hash.
+  astria.primitive.v1.TransactionId id_of_source_transaction = 6;
```
joroshiba, line 91: nit: `source_transaction_id`  would match naming conventions more. Generally perfer not to have "of" in names

proto/sequencerblockapis/astria/sequencerblock/v1alpha1/block.proto
```diff
@@ -86,6 +86,11 @@ message Deposit {
...
+  // index of the deposit's source action within its transaction
+  uint64 position_in_source_transaction = 7;
```
joroshiba, line 93: nit: `source_transaction_index`?

SuperFluffy:
Please expand on the list of breaking changes explaining why this is breaking the app hash.

joroshiba (approved):

PR #1411, opened 2024-08-27
feat(sequencer): have app recost transactions on fee or asset change

Lilyjjo:
## Summary
This PR builds on top of the mempool balance aware logic to recost the mempool when a `FeeAssetChange` or `FeeChange` action are seen in a finalized block.

There needs to be a follow up PR to ensure that the `FeeChange` and `FeeAssetChange` actions are only ran at the end of the block to ensure that transactions are invalidated mid-block due to fee changes. 

## Background
The mempool needs to recost transactions if the fees are changed, as the original placement of the transaction in the `parked` or `pending` structure may be out of date. 

## Changes
- Added logic to flag when `FeeAssetChange` or `FeeChange` actions are included in a block.
- The mempool will now re-cost transactions when this happens.
- Changed the mempool to use the app's state directly instead of using getters

## Testing
Unit tests for the mempool and the app logic.

## Breaking changes
This doesn't have breaking changes, but the snapshot files were re-generated because I changed the public addresses for `CAROL_ADDRESS`, `BOB_ADDRESS`, and `JUDY_ADDRESS` in the `test_utils.rs` file. I needed access to their private keys to sign transactions for testing promotion flows on the app level. 

## Metrics
Added `MEMPOOL_RECOSTED` counter to provide metrics on how often fees are being changed.


crates/astria-sequencer/src/mempool/mod.rs
```diff
@@ -292,34 +292,38 @@ impl Mempool {
     /// All removed transactions are added to the CometBFT removal cache to aid with CometBFT
     /// mempool maintenance.
     #[instrument(skip_all)]
-    pub(crate) async fn run_maintenance<F1, O1, F2, O2>(
+    pub(crate) async fn run_maintenance<F1, O1, F2, O2, F3, O3>(
         &self,
         current_account_nonce_getter: F1,
         current_account_balances_getter: F2,
+        transaction_cost_getter: F3,
```
Fraser999, line 299: I feel we're maybe at the point where we'd be better to just pass a reference to `S: accounts::StateReadExt` here now rather than having three getters.

The getters can make writing tests simpler, but at this stage we're making the production code quite a bit more cumbersome I think.

crates/astria-sequencer/src/app/mod.rs
```diff
@@ -1000,6 +1014,16 @@ impl App {
...
 
+        // flag mempool for cleaning if we ran a fee change action
+        for action in signed_tx.actions() {
+            match action {
+                Action::FeeAssetChange(_) | Action::FeeChange(_) => {
+                    self.recost_mempool = true;
+                }
+                _ => {}
+            }
+        }
```
Fraser999, line 1025: We can make this a little more efficient:
```suggestion
        self.recost_mempool = self.recost_mempool || signed_tx
            .actions()
            .iter()
            .any(|action| matches!(action, Action::FeeAssetChange(_) | Action::FeeChange(_)));
```
This way, if `self.recost_mempool` is already `true` before this statement, we won't iterate the actions.  Also, `any` is short-circuiting, so it stops iterating on the first instance where `matches!` returns `true`.

Fraser999 (commented):
Initial review.

crates/astria-sequencer/src/mempool/mod.rs
Lilyjjo: I'm starting to rewrite the unit tests to reflect this change and I'm afraid that it's creating a too-strong coupling between the mempool and how the internals of the app state work. Are you against keeping the getters as is?

SuperFluffy: I think the mempool is very tightly coupled to the state-reads anyway. There might be some benefit in constructing a handler-based system similar to what axum and warp are doing with their type-erased `Handler`s, but that requires some careful planning.

I think at this stage stronger coupling might be acceptable.

What would speak against it? Spinning off mempool as a free-standing crate?

Fraser999: Yeah - I was thinking we could create a test-only `State` struct which implemented `accounts::StateReadExt` and `cnidarium::StateRead` (with most methods being `unimplemented!()` since most aren't used), but this won't fly as we already have a blanket impl for `accounts::StateReadExt`.

I guess we need to use an actual `Storage` instance, populate a state delta with the required data for the test, and use the state delta here.  Might be a bit of boilerplate, but I guess we can abstract that to some mempool test fixture struct which can hold the storage and delta and can have methods to allow the required data to be put in the delta.

Lilyjjo: Finished using an actual storage instance! It wasn't as bad as I thought in terms of testing

Fraser999 (approved):

Issue #1407, opened 2024-08-26
IBC transfers in Noble USDC must contain bech32 addresses


SuperFluffy:
Noble enforces addresses in fungible packet messages to be bech32 encoded, failing a transfer otherwise.

This assertion held in as far as all IBC chains were Cosmos-derived, which uses bech32 addresses. However, with non-Cosmos IBC chains that make use of bech32m addresses like Penumbra (and now Astria) this check needs to be worked around by providing a compatibility layer at least until noble relaxes their checks.

Noble rejecting non-bech32 addresses: https://github.com/circlefin/noble-fiattokenfactory/blob/738932cb316d06f587c49dfb11a50515cce657d9/x/fiattokenfactory/ante.go#L204-L206
Fix by penumbra: https://github.com/penumbra-zone/penumbra/pull/4787/files
Penumbra discussion of the issue: https://forum.penumbra.zone/t/usdc-withdrawal-issues-and-proposed-fix/53
Unmerged fix into

┆Issue Number: ENG-745

PR #1425, opened 2024-08-29
feat(proto, core, sequencer)!: permit bech32 compatible addresses

SuperFluffy:
## Summary
Enables bech32 compatible addresses when necessary.

## Background
Some IBC enabled chains enforce that ics20 transfer packets must contain bech32 compatible only (see https://github.com/astriaorg/astria/issues/1407 for more background). This patch adds a `ibc_compat` prefix to sequencer's genesis and a new field `bool Ics20Withdrawal.use_compat_address` to the withdrawal action to request a bech32 compatible address format when interacting with such chains.

## Changes
- Add field `astria.protocol.transactions.v1alpha1.Ics20Withdrawal.use_compat_address`
- Add field `astria.protocol.genesis.v1alpha1.AddressPrefixes.ibc_compat`
- Rewrite the `Address` domain type to be generic over the format; this allows bech32 and bech32m addresses when parsing and formatting
    - Note that all Astria domain types containing addresses now require `Address<Bech32m>` with `Bech32m` also being set as the default type is the parameter is absent
- this now enforces that all Astria addresses in Astria's domain types *must* be bech32m compatible (this was not enforced before because `bech32::decode` does not enforce either checksum as of `bech32@v0.11`).
- When executing an `astria.protocol.transactions.v1alpha1.Ics20Withdrawal` with field `use_compat_address == true`, sequencer will now construct a bech32 compatible address using the value stored at `prefixes/ibc-compat` in its state.
- When refunding a rollup due to a failed withdrawal, sequencer will now attempt to parse the sender as either a standard (non-compat) bech32m address, or a compat bech32 address.
- If a compat bech32 address was found during a refund, sequencer will convert it to a bech32m address with its base prefix for further processing. This ensures that the compat addresses are only ever constructed for outgoing and incoming fungible token packets.

## Testing
+ Updated sequencer tests for ics20 refunds for both compat and base/non-compat senders.
+ Updated all snapshot tests
+ Added a snapshot test for `Address<Bech32>`
+ Added unit tests for address parsing (to ensure that bech32 addresses cannot be parsaed as bech32m and vice versa)
+ Updated chart sequencer genesis to run smoke tests
+ IBC tests for noble are still work in progress and outside the scope of this PR.

## Breaking Changelist
- Sequencer gained a new genesis value (the ibc compat prefix), changing the app hash

## Related Issues
Closes https://github.com/astriaorg/astria/issues/1407


SuperFluffy:
Squashed and rebased ontop of recent main because of merge conflicts.

Fraser999 (approved):
Just a couple of tiny nitpicks.

charts/sequencer/values.yaml
```diff
@@ -26,6 +26,7 @@ genesis:
   genesisTime: ""  # '2023-09-22T17:22:35.092832Z'
   addressPrefixes:
     base: "astria"
+    ibcCompat: "astriacompat"
```
noot, line 29: do these necessarily have to be different, or is it just to easily differentiate between the two address formats?

joroshiba: I think it's useful for easy identification of the two formats.

joroshiba (approved):
Approval for infra/protos only

charts/sequencer/values.yaml
SuperFluffy: Yes indeed. This is to make identification easier. We could always first try one and then the other format, but I think this is much cleaner.

crates/astria-bridge-contracts/src/lib.rs
```diff
@@ -420,6 +420,9 @@ where
             timeout_time: timeout_in_5_min(),
             source_channel,
             bridge_address: Some(self.bridge_address),
+            // FIXME: this needs a way to determine when to use compat address
+            // https://github.com/astriaorg/astria/issues/1424
+            use_compat_address: false,
```
noot, line 425: we could potentially check if the denom contains the `usdc` string in it

crates/astria-core/src/protocol/transaction/v1alpha1/action.rs
```diff
@@ -787,6 +786,11 @@ pub struct Ics20Withdrawal {
     // if unset, and the transaction sender is a bridge account, the withdrawal is
     // treated as a bridge withdrawal (ie. the bridge account's withdrawer address is checked).
     pub bridge_address: Option<Address>,
+
+    // whether to use a bech32-compatible format of the `.return_address` when generating
+    // fungible token packets (as opposed to Astria-native bech32m addresses). This is
+    // necessary for chains like noble which enforce a strict bech32 format.
```
noot, line 792: why does the return address need to be bech32 compatible? if the withdrawal is returned, it'll be to astria. shouldn't this be for `destination_chain_address`?

crates/astria-sequencer/src/ibc/ics20_transfer.rs
```diff
@@ -513,31 +508,109 @@ async fn execute_ics20_transfer<S: ibc::StateWriteExt>(
...
 
-/// execute a refund of tokens that were withdrawn from a rollup to another
-/// IBC-enabled chain via the sequencer using an `Ics20Withdrawal`, but were not
-/// transferred to the destination IBC chain successfully.
+/// Execute a refund for a failed ics20 transfer from a rollup to a remote IBC chain.
+///
+/// A withdrawal of tokens from a rollup to a remote IBC chain is started with a
+/// `Ics20Withdrawal` action via a (rollup's) bridge account on the sequencer.
+///
+/// This functions then sends the tokens back to the rollup via a `Deposit` event,
```
noot, line 516:
```suggestion
/// This function then sends the tokens back to the rollup via a `Deposit` event,
```

crates/astria-bridge-contracts/src/lib.rs
SuperFluffy: @noot Yeah, I am thinking about how to do this. Can you add that to the issue I linked? We should probably come up with a PRD.

crates/astria-sequencer/src/ibc/ics20_withdrawal.rs
```diff
@@ -50,20 +52,42 @@ use crate::{
...
+        let ibc_compat_prefix = state.get_ibc_compat_prefix().await.context(
+            "need to construct bech32 compatible address for IBC communication but failed reading \
+             required prefix from state",
+        )?;
+        withdrawal
+            .return_address()
+            .to_prefix(&ibc_compat_prefix)
+            .context("failed to convert the address to the bech32 compatible prefix")?
+            .to_format::<Bech32>()
+            .to_string()
```
noot, line 69: same question here - why does the counterparty care about the return address (which is an astria address?)

crates/astria-core/src/protocol/transaction/v1alpha1/action.rs
SuperFluffy: noble indeed checks the counterparty return address - unless I am mistaken about terminologies.

I have summarized the findings in this issue: https://github.com/astriaorg/astria/issues/1407

Do you see how I could improve the comment to make this clearer?

charts/sequencer/values.yaml
noot: do we need to specify `ibcCompat` in the genesis or would it be sufficient to just append `compat` onto the base prefix?

crates/astria-core/src/protocol/transaction/v1alpha1/action.rs
noot: nvm my comment, i reread the original bug post and this makes sense

charts/sequencer/values.yaml
SuperFluffy: I am actually ambivalent here. I like to explicitly spell it out, but do you think there might be a use case where we send funds back to, e.g., an `"astriabridge"` prefixed address?

crates/astria-core/src/sequencerblock/v1alpha1/block.rs
```diff
@@ -1281,7 +1281,7 @@ impl FilteredSequencerBlockError {
 ///
 /// A [`Deposit`] is constructed whenever a [`BridgeLockAction`] is executed
 /// and stored as part of the block's events.
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq)]
```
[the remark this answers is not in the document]
SuperFluffy: Yes indeed. I was figuring out how to implement `PartialEq` for `Address<T>` and forgot to bring it back. Fixed in 19b4a0cd

proto/protocolapis/astria/protocol/genesis/v1alpha1/types.proto
```diff
@@ -23,7 +23,11 @@ message Account {
 }
 
 message AddressPrefixes {
+  // The base prefix used for most addresses sequencer addresses
```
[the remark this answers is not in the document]
SuperFluffy: 19b4a0cd

crates/astria-sequencer/src/ibc/ics20_transfer.rs
SuperFluffy: 19b4a0cd

Issue #1145, opened 2024-06-04
sequencer: cleanup/refactor of action fee checks


noot:
we added a `check_balance_for_total_fees` which checks the sender's balance for fees and asset transfers in an entire tx before executing it. however, we should also check the fee assets are valid. we can then also remove the fee/balance checks inside individual actions's `check_stateful` as it would have already been checked.

PR #1426, opened 2024-08-29
feat(sequencer): rework fee reporting

### LINKED ISSUE #1369: Rework fee reporting
## Issue
#1305 added fee reporting by way of `get_and_increase_block_fees()`, which was fixed in #1343 by adding a call to this function in all actions which require fees. This setup requires passing ABCI events all the way through to `finalize_block()`, where the events are retroactively placed into `ExecTxResult`s. We should rework this such that the result Txs are constructed at a lower level and passed back through the execution calls to `finalize_block()`.

*Note: this should only be done after #1332 is merged*

## Relevant Code
**Fee reporting in** `get_and_increase_block_fees`
https://github.com/astriaorg/astria/blob/c5364e801a193b799e71390f51c54b76a934c999/crates/astria-sequencer/src/assets/state_ext.rs#L208
https://github.com/astriaorg/astria/blob/c5364e801a193b799e71390f51c54b76a934c999/crates/astria-sequencer/src/assets/state_ext.rs#L231
**Construction of fee event**
https://github.com/astriaorg/astria/blob/c5364e801a193b799e71390f51c54b76a934c999/crates/astria-sequencer/src/assets/state_ext.rs#L50-L63
**Construction of** `ExecTxResult` **from ABCI event**
https://github.com/astriaorg/astria/blob/c5364e801a193b799e71390f51c54b76a934c999/crates/astria-sequencer/src/app/mod.rs#L817-L820

┆Issue Number: ENG-718

ethanoroshiba:
## Summary
Reworked fee reporting based on [new design](https://www.notion.so/astria-org/Sequencer-Fee-Reporting-6c211884b50e4d4e844dee9af95ff424?d=9030b989a38543c8b9a85a03ffe28825#826da45147304caa91d7f46835c44b51).

## Background
Fee reporting was previously handled in a very roundabout way, with a few issues:
- Fees were burned and paid out in separate places and points of execution. This separation made it easy in the future to change one functionality without correctly updating the other, and hence introduce bugs.
- `BridgeLock` and `BridgeUnlock` actions both instantiated and executed `TransferAction` s within their `check_and_execute()` methods. Because fees are recorded based on the name of the action type, this would incorrectly report associated lock and unlock fees as transfer fees.
- There was already a check in place to ensure the sender balance can cover the total fees: `check_balance_for_total_fees_and_transfers()`. Hence, we were calculating the cost of fees more than once, which should not be the case.

## Changes
- Removed all current fee reporting mechanisms.
- Added `astria-sequencer::transaction::fees`, which handles all fee calculation and reporting.
- Created function `get_and_report_fees()`, which calculates all fees for a given transaction and returns a `fees_by_asset` hash map and an optional `fee_payment_map` for processing fee payments. This allows dual functionality for fees to be checked based on a state that only implements `StateRead` and not `StateWrite`.
- Created function `pay_fees()`, which intakes a state which implements `StateWrite` along with a `fee_payment_map` hash map. This function pays all fees and records all fee events to the state.
- Added `actionIndex` attribute to `tx.fees` events so that outside observers can see which action actually correlates to the fee event and where it is in the block.
- App hash was changed, snapshot regenerated.

## Testing
- Created fairly exhaustive unit tests for every fee-bearing action, testing each's fees, events, fee changes, and payment execution.
- Created tests to ensure failure on missing sudo address, invalid fee asset type, missing source, and missing fees.
- Created test to ensure correct handling of mid-transaction sudo address change

## Related Issues
closes #1369 and #1145


crates/astria-sequencer/src/app/snapshots/astria_sequencer__app__tests_breaking_changes__app_execute_transaction_with_every_action_snapshot.snap
joroshiba, line 1: This snapshot test suggests that we are changing actual state of acocunt balances with this PR, which I don't believe should be happening?

ethanoroshiba: Would the snapshot also change if all the functionality regarding block fees was taken out? That information is no longer recorded to the state (and the functionality to do so was taken out as well), instead just the account balances are increased/decreased and the fee events recorded to the state.

ethanoroshiba: Hmm, that's very strange. My logic may be wrong somewhere down the line but for all transactions with fees I implemented tests which check the account balances on the back end after paying, so I think if something is wrong it's probably in the fee calculation itself?

crates/astria-sequencer/src/transaction/fees.rs
```diff
@@ -0,0 +1,683 @@
...
+        }
+        action_index = action_index
+            .checked_add(1)
+            .expect("action index overflowed");
+    }
+    if return_payment_map {
+        Ok((fees_by_asset, Some(fee_payment_map)))
+    } else {
+        Ok((fees_by_asset, None))
+    }
```
noot, line 246: the `fee_payment_map` is generated every time no matter what `return_payment_map` is - remove the `return_payment_map`  param (also from `check_balance_and_get_fees`) and just return the payment map everytime, the caller can decide to drop or not

crates/astria-sequencer/src/transaction/fees.rs
```diff
@@ -0,0 +1,683 @@
...
+            .await
+            .wrap_err("failed to decrease from balance")?;
+        state
+            .increase_balance(*to, fee_asset, fee_info.amt)
+            .await
+            .wrap_err("failed to increase to balance")?;
+        for event in &fee_info.events {
+            state.record(event.clone());
+        }
+    }
```
noot, line 268:
```suggestion
    for ((from, to, fee_asset), fee_info) in fee_payment_map {
        state
            .decrease_balance(from, &fee_asset, fee_info.amt)
            .await
            .wrap_err("failed to decrease from balance")?;
        state
            .increase_balance(to, &fee_asset, fee_info.amt)
            .await
            .wrap_err("failed to increase to balance")?;
        for event in fee_info.events {
            state.record(event);
        }
    }
```

crates/astria-sequencer/src/transaction/fees.rs
```diff
@@ -0,0 +1,683 @@
...
+        ],
+    )
+}
+
+#[instrument(skip_all)]
+fn warn_if_asset_type_not_allowed(asset: &asset::Denom, allowed_fee_assets: &[IbcPrefixed]) {
+    if !allowed_fee_assets.contains(&IbcPrefixed::from(asset)) {
+        warn!("Transaction execution will fail, asset type not allowed for fee payment: {asset}");
+    }
+}
```
noot, line 671: this should be a hard error, not just a log

crates/astria-sequencer/src/transaction/fees.rs
noot: nvm, i see you are using it to populate the map in the `_update_fees` functions. not really a fan of this method - i would prefer if each action had two separate functions, one for updating `fees_by_asset` and one for updating the payment map.

crates/astria-sequencer/src/transaction/fees.rs
```diff
@@ -0,0 +1,683 @@
...
+    pub(crate) events: Vec<Event>,
+}
+
+struct PaymentInfo<'a> {
+    from: Option<[u8; 20]>,
+    to: Option<[u8; 20]>,
+    allowed_fee_assets: &'a mut Vec<IbcPrefixed>,
+    fee_asset: asset::Denom,
+    fees_by_asset: &'a mut HashMap<asset::IbcPrefixed, u128>,
+    fee_payment_map: &'a mut HashMap<([u8; 20], [u8; 20], asset::Denom), FeeInfo>,
```
noot, line 77: make a struct for `([u8; 20], [u8; 20], asset::Denom)` so the fields can be named:
```
struct Payment {
    from: [u8; 20],
    to: [u8; 20],
    denom: asset::Denom,
}
```

crates/astria-sequencer/src/transaction/fees.rs
```diff
@@ -0,0 +1,683 @@
...
+            allowed_fee_assets.push(asset.clone().into());
+        }
+        FeeAssetChangeAction::Removal(asset) => {
+            allowed_fee_assets.retain(|a| *a != IbcPrefixed::from(asset));
+        }
+    }
+}
+
+fn handle_fee_change(act: &FeeChangeAction, current_fees: &mut Fees) {
+    match act.fee_change {
```
noot, line 558: i just realized this should also have a `BridgeUnlock` variant so the fees for `BridgeUnlock` can be updated, we should do this in a follow up

crates/astria-sequencer/src/app/snapshots/astria_sequencer__app__tests_breaking_changes__app_execute_transaction_with_every_action_snapshot.snap
noot: the reason this changes is because of the `app_execute_transaction_with_every_action_snapshot` test - it doesn't call `end_block` via `finalize_block`, just `commit`. it probably should have :/ previously, fees were being updated only in `end_block`, but now they're updated when the transaction is executed, so the state at the end of the test is different (even though this may not be a breaking change)

noot: we can check if this is breaking by updating the test to call `end_block` on main and this branch and seeing if it matches

crates/astria-sequencer/src/transaction/fees.rs
ethanoroshiba: Should this be a hard error even if the transaction is not being executed? This function may be called by `transaction_fee_request`, in which case I had thought it might be important to differentiate between warning when `get_and_report_tx_fees()` is called for a fee request, versus erroring when it is called during transaction execution.

crates/astria-sequencer/src/transaction/fees.rs
ethanoroshiba: I originally had it set up the way you're describing, but changed it for a couple reasons: 1) `get_and_report_tx_fees()` was over the line limit, and this was a way to cut down on line count quite significantly, and 2) some of the actions require fee calculation beyond just the base transfer fees, and this way the calculation would only need to be done once within the `_update_fees` functions and then passed to `increase_fees()`. If you think it's important to have it set up the other way, I can revert this and put a clippy allow in. Thanks for taking a look over!

crates/astria-sequencer/src/transaction/fees.rs
ethanoroshiba: Switched to struct use in 7c1e05c.

crates/astria-sequencer/src/app/snapshots/astria_sequencer__app__tests_breaking_changes__app_execute_transaction_with_every_action_snapshot.snap
ethanoroshiba: PR out for this now: https://github.com/astriaorg/astria/pull/1455

SuperFluffy:
I feel this change breaks with the current architecture of sequencer too much and makes it more difficult to reason about. I don't want to merge this in as is.

My main issue is that I can no longer understand the state transitions that a single action will cause just by looking at its `AppHandler` implementation. This includes the state transitions due to burning fees. There is also the issue of always ensuring that the state transitions as understood by the payment map match those performed by the actions.

I would rather there be one source of truth.

But to take a step back, this PR attempts to fix several things at the same time which I think could (and should be) handled in separate PRs (taking these from the PR description):

> Fees were burned and paid out in separate places and points of execution. This separation made it easy in the future to change one functionality without correctly updating the other, and hence introduce bugs.

I don't immediately see anything bad about this. The execution of an action should result in 2 state changes: 1. the decrease of a balance for a given asset on the payer side, and 2. an increase of a balance for a given asset on the payee side (or in this case, the receiver of a payment).

However, since the cost of an action should be defined by its contents, I think `check_and_execute` is the right place to perform these state transitions.

With that said, I can see how the actual implementation of how payments are made and reported could be changed to become clearer. But I think that can be done with a lighter touch then what this does. For example, by making use of the transaction context ID and action index that was recently added in a different PR.

> BridgeLock and BridgeUnlock actions both instantiated and executed TransferAction s within their check_and_execute() methods. Because fees are recorded based on the name of the action type, this would incorrectly report associated lock and unlock fees as transfer fees.

I agree that this is problematic and I stumbled over this myself. However, this particular problem should be addressed in a different PR.

> There was already a check in place to ensure the sender balance can cover the total fees: check_balance_for_total_fees_and_transfers(). Hence, we were calculating the cost of fees more than once, which should not be the case.

Can you point to that line? It does sound repetitive (and should indeed be handled on a per-action level).

ethanoroshiba:
Closing due to offline discussion. Will open new PR with a design which breaks less from the current sequencer logic.

PR #1428, opened 2024-08-29
fix(bridge-contracts): fix memo transaction hash encoding

### LINKED ISSUE #1427: Use the correct encoding for `ethers` hashes in the bridge's conversion logic
Essentially, we were using `ethers::H256::to_string()` to get the string for the `rollup_transaction_hash` value, which returns a shortened version of the hash. the string `0x1234...0x1234` is instead of the full hash (i.e. you would expect it to give you something like `0x1234567890123456789012345678901234567890123456789012345678901234`). The fix is to use `encode_hex` instead of `to_string`.

┆Issue Number: ENG-764

itamarreif:
## Summary
Use the correct encoding functions to populate memo field in the bridge's conversion logic from rollup withdrawal events to sequencer actions.

## Background
We were using `ethers::H256::to_string()` to get the string for the `rollup_transaction_hash` value, which returns a shortened version of the hash. the string `0x1234...0x1234` is instead of the full hash (i.e. you would expect it to give you something like `0x1234567890123456789012345678901234567890123456789012345678901234`).

## Changes
- Use the correct 

## Breaking Changelist
- The previously invalid memo data will now be populated with the correct information. This doesn't actually break anything because we also aren't actually using these hashes anywhere.

## Related Issues

closes #1427, #1471


noot (approved):

crates/astria-bridge-withdrawer/tests/blackbox/helpers/test_bridge_withdrawer.rs
```diff
@@ -30,7 +30,10 @@ use astria_core::{
         },
     },
 };
-use ethers::types::TransactionReceipt;
+use ethers::{
+    types::TransactionReceipt,
+    utils::hex::ToHexExt,
```
SuperFluffy, line 35: Import as:
```suggestion
    utils::hex::ToHexExt as _,
```

crates/astria-bridge-contracts/src/lib.rs
```diff
@@ -32,6 +32,7 @@ use ethers::{
         Log,
         H256,
     },
+    utils::hex::ToHexExt,
```
SuperFluffy, line 35:
```suggestion
    utils::hex::ToHexExt as _.,
```

SuperFluffy (approved):
Change looks fine, but please import the trait anonymously.

crates/astria-bridge-contracts/src/lib.rs
```diff
@@ -502,6 +516,11 @@ impl GetWithdrawalActionsError {
     fn log_without_transaction_hash(_log: &Log) -> Self {
         Self(GetWithdrawalActionsErrorKind::LogWithoutTransactionHash)
     }
+
+    // XXX: Somehow identify the log?
```
noot, line 520: don't think this error should ever happen, but maybe put the tx hash?

crates/astria-sequencer/src/bridge/bridge_unlock_action.rs
```diff
@@ -35,7 +35,7 @@ impl ActionHandler for BridgeUnlockAction {
             "rollup withdrawal event id must be non-empty",
         );
         ensure!(
-            self.rollup_withdrawal_event_id.len() <= 64,
+            self.rollup_withdrawal_event_id.len() <= 256,
             "rollup withdrawal event id must not be more than 64 bytes",
```
noot, line 39: update error message to `256`

PR #1444, opened 2024-09-04
chore: memoize `address_bytes` of verification key

### LINKED ISSUE #1351: Assess and memoize (if appropriate) `VerificationKey::address_bytes`
We should run an ad hoc test under semi-realistic conditions which counts how many times `VerificationKey::address_bytes` is a repeated call on a `VerificationKey` instance.  In other words, if we only ever call this once per instance, the total will be zero.  If we call it exactly thrice per instance, the total will be twice the number of instances.

Once we have this information, we can decide whether to implement memoization of the address bytes as a field in the verification key, or leave as is.

Fraser999:
## Summary
Added a lazily-initialized field `address_bytes` to `VerificationKey`.

## Background
Testing showed that `address_bytes` was called multiple times on a given `VerificationKey` (up to 11 times in some cases).  Each time the key's bytes were being hashed, so this change ensures that hashing only happens once for a given verification key instance.

Note that this was implemented previously in #1111 and was then reverted in #1124.  However, when reverted, the manual impls of `PartialEq`, `Eq`, `PartialOrd`, `Ord` and `Hash` were left as-is, as were the unit tests for these.  Hence this PR doesn't need to make any changes to these trait impls or tests.

## Changes
- Added `address_bytes: OnceLock<[u8; ADDRESS_LEN]>` to `VerificiationKey`.

## Testing
No new tests required.  

## Related Issues
Closes #1351.

### DIFF
--- a/crates/astria-core/src/crypto.rs
+++ b/crates/astria-core/src/crypto.rs
@@ -10,6 +10,7 @@ use std::{
         Hash,
         Hasher,
     },
+    sync::OnceLock,
 };
 
 use base64::{
@@ -78,6 +79,7 @@ impl SigningKey {
     pub fn verification_key(&self) -> VerificationKey {
         VerificationKey {
             key: self.0.verification_key(),
+            address_bytes: OnceLock::new(),
         }
     }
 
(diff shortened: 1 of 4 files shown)


SuperFluffy (commented):
High level comment on this memoization:

Is there a more explicit way to memoize the address bytes, for example by calculating `address_bytes` once and storing them near the call site?

This change is relatively minimal for the user (minus the `Copy` trait, although we don't seem to make ause of it).

But introducing syncing primitives into a relatively innocent newtype wrapper makes me uneasy.

SuperFluffy (approved):

Fraser999:
> Is there a more explicit way to memoize the address bytes, for example by calculating address_bytes once and storing them near the call site?
>
> This change is relatively minimal for the user (minus the Copy trait, although we don't seem to make ause of it).

I think doing that means the change would have an impact on callers, which isn't the intent here.

> But introducing syncing primitives into a relatively innocent newtype wrapper makes me uneasy.

Well, I guess for my money, the need for hand-rolled equality/ordering/hash traits is the biggest drawback here, rather than the fact the new field is a synchronization primitive.

PR #1447, opened 2024-09-04
feat(sequencer): report deposit events

### LINKED ISSUE #1440: create cometbft tx result event when rollup `Deposit` created
Right now we create the `Deposit` in the sequencer app but there is nothing in cometbft indicating this has occured. We should have a tx result event which contains the deposit associated with the tx.

┆Issue Number: ENG-772

ethanoroshiba:
## Summary
Added creation of ABCI deposit events and record them to the state to be included in `ExecTxResult`.

## Background
Deposit events were previously only stored app-side in `SequencerBlock`. This will allow for outside observation of the deposit events.

## Changes
- Added helper function for the creation of ABCI deposit events.
- Added recording of these events whenever deposits are created.

## Testing
Added test to ensure deposit events are correctly recorded and returned by `state.apply()`

## Related Issues
closes #1440


crates/astria-sequencer/src/app/tests_execute_transaction.rs
```diff
@@ -1119,3 +1126,59 @@ async fn app_execute_transaction_action_index_correctly_increments() {
...
+
+    signed_tx.check_and_execute(&mut state_tx).await.unwrap();
+    let events = &state_tx.apply().1;
+    for event in events {
+        if event.kind == "tx.deposit" {
+            assert_eq!(*event, expected_deposit_event);
+            return;
+        }
+    }
+    panic!("no deposit event found in events");
```
Fraser999, line 1183: Maybe a little more canonical - not a big deal:
```suggestion
    let event = events
        .iter()
        .find(|event| event.kind == "tx.deposit")
        .expect("should have deposit event");
    assert_eq!(*event, expected_deposit_event);
```

Fraser999 (approved):
One non-blocking comment.

PR #1463, opened 2024-09-06
feat(conductor): implement restart logic

### LINKED ISSUE #906: feat: conductor restart logic
Currently conductor needs to be started after a rollup node implementing execution API. This is bad for development flow as it doesn't allow hot restart etc of services. We should update this to do a couple of things. The first is a simpler implementation likely unblocks most workflows, the second allows for dev workflows with a hot reload type setup (ie I update my rollup binary and blockchain restarts seamlessly).

1) On initial startup, poll for execution layer to be ready. Log that it's not but do no restart.
2) If execution API dies, we can restart services internally. The Geth implementation now responds with a `PermissionDenied` if `GetGenesisInfo` and `GetCommitmentState` are not called before attempting block execution. This would also indicate a need to clear cache and restart internally. (This should be documented as required behaivior)

ethanoroshiba:
## Summary
Added logic to restart conductor tasks if it receives a `PermissionDenied` code from the execution layer.

## Background
Conductor previously needed to be restarted if a rollup was restarted. This change is meant to make it such that if a rollup goes down and comes back online, the conductor can restart internally and automatically as opposed to needing a manual restart.

## Changes
- Added wrapper task and corresponding handle around conductor to manage a restart, allowing the `InnerConductorTask` to shut down gracefully before restarting.
- Added error propagation within the conductor to ensure tasks do not silently fail.
- Added `RestartToken` to be returned by the `InnerConductorTask` on shutdown to indicate to the wrapper whether to restart.
- Added checks for task errors with `tonic::code::PermissionDenied` to initiate a restart.
- Implemented logic to check a returned error for a need to restart and function `restart()` to end the tasks and create new ones.

## Testing
Added blackbox test and helper functions to ensure conductor restarts after receiving a `PermissionDenied` status code.

## Breaking Changelist
Not  breaking, but we should have documentation that reflects the fact that rollups need to return a `PermissionDenied` error if they receive an `execute_block` gRPC before `get_genesis_info` or `get_commitment_state`. Similarly, if the type of code used for this changes, the conductor restart logic will need to be changed to reflect this. It's quite rigid right now.

## Related Issues
closes #906


crates/astria-conductor/src/conductor.rs
```diff
@@ -174,19 +187,29 @@ impl Conductor {
...
 
-            Some((name, res)) = self.tasks.join_next() => {
-                match flatten(res) {
-                    Ok(()) => Err(eyre!("task `{name}` exited unexpectedly")),
-                    Err(err) => Err(err).wrap_err_with(|| "task `{name}` failed"),
+                Some((name, res)) = self.tasks.join_next() => {
+                    match flatten(res) {
+                        Ok(()) => break Err(eyre!("task `{name}` exited unexpectedly")),
+                        Err(err) => {
+                            if check_for_restart(&err) {
```
SuperFluffy, line 202: The assumption here is that this works if the executor-task exits first, followed by the other tasks.

There is no guarantee however for that to happen: if executor has lengthy cleanup code after it exits its event loop while other the other tasks exit immediately, then this check is guaranteed to no longer work.

crates/astria-conductor/src/conductor.rs
```diff
@@ -245,6 +269,16 @@ impl Conductor {
         }
         info!("shutting down");
     }
+
+    #[instrument(skip_all, err)]
+    async fn restart(&mut self) -> eyre::Result<()> {
```
SuperFluffy, line 274: Couple notes on this logic:

1. if the `run_until_stopped` loop introduces logic before it enters its event loop, then this restart function will immediately break.
2. this is hard killing the other tasks without waiting for them to complete. The line `self.tasks = new_conductor.tasks` just aborts all the functions immediately. It's not unlikely that `self.task_shutdown.cancel()` was not yet received by the other tasks.
3. this code should allow the other tasks to exit gracefully before starting a new conductor.

SuperFluffy (changes requested):
Requesting changes because checking the error condition for a restart is flaky, and because the way a new conductor is spun up is prone to refactoring errors:

I would prefer if a conductor restart used the same `Conductor::new` into `Conductor::run_until_stopped` logic that is used to spin it up in the first place.

The easiest way to accomplish this might be to change Conductor into a wrapper around a conductor task like `Conductor(JoinHandle<ConductorImpl>)`, and then `ConductorImpl` signals the reason for its exit, letting the public-facing `Conductor` wrapper decide to restart or not.

crates/astria-conductor/src/conductor.rs
ethanoroshiba: Added checks for the `PermissionDenied` code for the remaining tasks in `shutdown()` in d6fde21.

crates/astria-conductor/src/conductor.rs
ethanoroshiba: Added wrapper around Conductor per your suggestions, allowing the inner conductor task to fully shut down and return `RestartToken::Restart` if it should be restarted, or `RestartToken::Shutdown` if not. Per d6fde21. Thanks for your help on this!

crates/astria-conductor/src/conductor.rs
```diff
@@ -36,32 +37,41 @@ use crate::{
     Config,
 };
 
+/// Token to signal whether the conductor should restart or shut down.
+pub enum RestartToken {
```
SuperFluffy, line 41: It should not be public, and we should probably reserve the use of the word `Token` for cancellation tokens:

```suggestion
enum RestartOrShutdown {
```

crates/astria-conductor/src/conductor.rs
```diff
@@ -77,20 +87,145 @@ impl Future for Handle {
...
+        let this = self.project();
+        let task = this
+            .task
+            .as_mut()
+            .expect("the Conductor handle must not be polled after shutdown");
+        task.poll_unpin(cx)
+    }
+}
+
+pub struct InnerConductorTask {
```
SuperFluffy, line 215: This should not be public. Only the `Conductor` object should be `pub`.

crates/astria-conductor/src/conductor.rs
SuperFluffy, line 1: I think `conductor.rs` should be split into a public facing module and its internal implementation, something like this:

```
conductor/
  mod.rs     // this contains `pub struct Conductor`
  impl.rs    // this contains the inner business logic
```

crates/astria-conductor/src/conductor.rs
```diff
@@ -77,20 +87,145 @@ impl Future for Handle {
...
+        self.conductor_inner_shutdown = new_handle.shutdown_token.clone();
+        self.handle = new_handle;
+    }
+
+    /// Initiates shutdown of all conductor tasks from the top down, ignoring a restart signal.
+    async fn shutdown(self) -> eyre::Result<()> {
+        self.conductor_inner_shutdown.cancel();
+        let shutdown_result = self.handle.await?;
+        shutdown_result?;
+        Ok(())
```
SuperFluffy, line 176: Too many early escapes.

crates/astria-conductor/src/conductor.rs
```diff
@@ -77,20 +87,145 @@ impl Future for Handle {
...
+}
+
+impl Conductor {
+    /// Creates a new `Conductor` from a [`Config`].
+    ///
+    /// # Errors
+    /// Returns an error if [`InnerConductorTask`] could not be created.
+    pub fn new(cfg: Config, metrics: &'static Metrics) -> eyre::Result<Self> {
+        let conductor = InnerConductorTask::new(cfg.clone(), metrics)?;
+        let conductor_inner_shutdown = conductor.shutdown.clone();
```
SuperFluffy, line 115: As noted above, this is redundant. It's better to create a child token and pass that into the inner conductor.

crates/astria-conductor/src/conductor.rs
```diff
@@ -77,20 +87,145 @@ impl Future for Handle {
...
+        let conductor_inner_handle = conductor.spawn();
```
SuperFluffy, line 116: IMO roll all of this into a single constructor and call it like `InnerConductorTask::spawn(cfg, child_token, metrics)`

crates/astria-conductor/src/conductor.rs
```diff
@@ -244,13 +400,33 @@ impl Conductor {
...
-fn report_exit(exit_reason: eyre::Result<&str>, message: &str) {
+fn report_exit(exit_reason: &eyre::Result<&str>, message: &str) {
     match exit_reason {
         Ok(reason) => info!(%reason, message),
         Err(reason) => error!(%reason, message),
     }
 }
+
+#[instrument(skip_all)]
+fn check_for_restart(err: &eyre::ErrReport) -> bool {
```
SuperFluffy, line 421: Can you create a a simple unit test with a tonic error somewhere in the source chain?

SuperFluffy (commented):
Changes are going in the right direction, thank you! I think we should split the public-facing conductor and the inner business logic by modules. There is quite a bit of duplicated naming for both objects, and going through the single module makes that confusing.

Also have a second look at the cancellation tokens. If you clone a token, then cancelling it will automatically cancel the other, which might not be what you intended. Instead, check out https://docs.rs/tokio-util/0.7.12/tokio_util/sync/struct.CancellationToken.html#method.child_token, which allows you to cancel tokens only at the level of the child token (or below).

crates/astria-conductor/src/conductor.rs
```diff
@@ -184,26 +320,40 @@ impl Conductor {
...
+
+        if should_restart {
+            return Ok(RestartToken::Restart);
+        }
+
+        if let RestartToken::Restart = shutdown_res {
+            return Ok(RestartToken::Restart);
+        }
+        exit_reason?;
+        Ok(RestartToken::Shutdown)
```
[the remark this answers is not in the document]
ethanoroshiba: Tried to streamline a fair bit in 4f605ea. Let me know if there's more to be done in this regard!

crates/astria-conductor/src/conductor.rs
```diff
@@ -77,20 +87,145 @@ impl Future for Handle {
...
+pub struct InnerConductorTask {
+    /// Token to signal to all tasks to shut down gracefully.
+    shutdown: CancellationToken,
+
     /// The different long-running tasks that make up the conductor;
     tasks: JoinMap<&'static str, eyre::Result<()>>,
 }
 
-impl Conductor {
+impl InnerConductorTask {
```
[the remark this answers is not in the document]
ethanoroshiba: I chose `ConductorInner` since I thought it was slightly more appropriate, and also to my mind `Impl` kind of implies a trait. But, if using `{MODULENAME}Impl` is standard I'm happy to change it

crates/astria-conductor/src/conductor.rs
```diff
@@ -167,12 +302,13 @@ impl Conductor {
         })
     }
 
-    /// Runs [`Conductor`] until it receives an exit signal.
+    /// Runs [`InnerConductorTask`] until it receives an exit signal.
     ///
     /// # Panics
     /// Panics if it could not install a signal handler.
-    async fn run_until_stopped(mut self) {
+    async fn run_until_stopped(mut self) -> eyre::Result<RestartToken> {
```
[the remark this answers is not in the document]
ethanoroshiba: I have it returning an error because otherwise if one of the conductor tasks fail it just shuts down and doesn't propagate the error. It does emit an error event, so I wouldn't call it totally silent, but it seemed like it didn't follow the same error-handling approach from the rest of the codebase. Happy to remove this if needed, though.

crates/astria-conductor/src/conductor.rs
ethanoroshiba: There's currently a blackbox test in `soft_and_firm()` to make sure the conductor restarts. Do you want to implement a unit test in addition to this? I originally had purposefully not done this because I remember you had said not to add any new unit tests to conductor for the chain ID checks, but happy to add one if needed.

SuperFluffy: Yeah, not a unit test for the restart logic of the full task (which the blackbox test covers) - but walking through a chain of errors is a good unit test.

crates/astria-conductor/src/conductor/inner.rs
```diff
@@ -30,38 +30,28 @@ use tracing::{
...
+    Metrics,
 };
 
-pin_project! {
-    /// A handle returned by [`Conductor::spawn`].
-    pub struct Handle {
-        shutdown_token: CancellationToken,
-        task: Option<tokio::task::JoinHandle<()>>,
-    }
+/// Token to signal whether the conductor should restart or shut down.
```
SuperFluffy, line 39:
```suggestion
/// Exit value of the inner conductor impl to signal to the outer task whether to restart or shutdown
```

crates/astria-conductor/src/conductor/mod.rs
```diff
@@ -0,0 +1,155 @@
...
+    }
+}
+
+/// A wrapper around [`ConductorInner`] that manages shutdown and restart of the conductor.
+pub struct Conductor {
+    /// Token to signal to all tasks to shut down gracefully.
+    shutdown_token: CancellationToken,
+
+    /// Handle for the inner conductor task.
+    handle: InnerHandle,
```
SuperFluffy, line 76: "handle` is very nondescriptive.
```suggestion
    inner: inner::ConductorImpl
```

crates/astria-conductor/src/conductor/inner.rs
```diff
@@ -215,18 +212,29 @@ impl Conductor {
     /// because kubernetes issues SIGKILL 30 seconds after SIGTERM, giving 5 seconds
     /// to abort the remaining tasks.
     #[instrument(skip_all)]
-    async fn shutdown(mut self) {
-        self.shutdown.cancel();
+    async fn shutdown(
+        mut self,
+        exit_reason: eyre::Result<&str>,
```
SuperFluffy, line 217: If you pass in the latest failed task here with it's `&'static str` name and and its shutdown reason/error, then you can test this immediately instead of needing to collect all the shutdown reasons and parsing them.

crates/astria-conductor/src/conductor/inner.rs
```diff
@@ -244,13 +252,57 @@ impl Conductor {
...
+        // If any tasks failed and don't warrant a restart, return their error
+        for task_res in finished_tasks {
+            if let Err(err) = task_res {
+                let Some(task_name) = task_failed(&err) else {
+                    continue;
+                };
+                if task_name == "executor" && check_for_restart(&err) {
+                    return Ok(RestartOrShutdown::Restart);
+                }
+                return Err(err);
```
SuperFluffy, line 266: As mentioned above, remove all of this. Especially avoid the parsing the error message. in `task_failed`

crates/astria-conductor/src/conductor/inner.rs
```diff
@@ -162,51 +154,56 @@ impl Conductor {
...
-    /// Runs [`Conductor`] until it receives an exit signal.
+    /// Runs [`ConductorInner`] until it receives an exit signal.
     ///
     /// # Panics
     /// Panics if it could not install a signal handler.
-    async fn run_until_stopped(mut self) {
+    async fn run_until_stopped(mut self) -> eyre::Result<RestartOrShutdown> {
         info_span!("Conductor::run_until_stopped").in_scope(|| info!("conductor is running"));
 
         let exit_reason = select! {
```
SuperFluffy, line 169: For my suggestion further down: you don't need to keep the type of this `exit_reason` the same.

You can just introduce an enum like this:

```rust
enum ExitReason {
    ShutdownSignal,
    TaskFailed { name: &'static str, error: eyre::Report, }
}
```

And then your `shutdown` logic can deal with this elegantly.

crates/astria-conductor/src/conductor/inner.rs
```diff
@@ -244,13 +272,50 @@ impl Conductor {
...
     match exit_reason {
-        Ok(reason) => info!(%reason, message),
-        Err(reason) => error!(%reason, message),
+        ExitReason::ShutdownSignal => info!("received shutdown signal, {}", message),
+        ExitReason::TaskFailed {
+            name: _,
+            error,
+        } => error!(%error, message),
+    }
+}
```
SuperFluffy, line 289: Can you bring back the previous structure?

```suggestion
        ExitReason::ShutdownSignal => info!(reason = "received shutdown signal", message),
        ExitReason::TaskFailed {
            name: task,
            error: reason,
        } => error!(%reason, %task, message),
    }
}
```

crates/astria-conductor/src/conductor/inner.rs
```diff
@@ -10,7 +10,7 @@ use astria_eyre::eyre::{
 };
 use itertools::Itertools as _;
 use pin_project_lite::pin_project;
-use sequencer_client::HttpClient;
+use tendermint_rpc::HttpClient;
```
SuperFluffy, line 13: Why is this necessary?

crates/astria-conductor/src/conductor.rs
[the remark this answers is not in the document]
SuperFluffy: It's common to use `Thing` in the public API, and `ThingImpl` (or `ThingInner`) for the implementation deatils.

I would have been ok with `ConductorInner` as I said :D I just took offense to using `InnerConductorTask`

SuperFluffy (approved):
This will be a nice feature to have. Thank you for implementing it!

PR #1472, opened 2024-09-09
fix(conductor): fix flaky soft_and_firm test

### LINKED ISSUE #1143: conductor blackbox test for soft-and-firm mode is flaky
<details>
<summary>Failure log</summary>

    running 11 tests
    test soft_only::skips_already_executed_heights ... ok
    test soft_only::simple ... ok
    test shutdown::shut_down_during_bringup ... ok
    test soft_only::requests_from_later_genesis_height ... ok
    test soft_only::submits_two_heights_in_succession ... ok
    test firm_only::fetch_from_later_celestia_height ... ok
    test firm_only::submits_two_heights_in_succession ... ok
    test firm_only::simple ... ok
    test soft_and_firm::missing_block_is_fetched_for_updating_firm_commitment ... ok
    test firm_only::skips_already_executed_heights ... ok
    test soft_and_firm::simple ... FAILED
    
    failures:
    
    ---- soft_and_firm::simple stdout ----
    thread 'soft_and_firm::simple' panicked at crates/astria-conductor/tests/blackbox/soft_and_firm.rs:140:6:
    conductor should have executed the block and updated the soft and firm commitment states within 1000ms: Elapsed(())
    note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
    thread 'tokio-runtime-worker' panicked at crates/astria-grpc-mock/src/mock_server.rs:91:17:
    Verifications failed:
    - Mock #4 for RPC update_commitment_state.
            Expected range of matching incoming requests: == 1
            Number of matched requests with valid mock responses: 0
            Number of matched requests with bad mock responses: 0
    
    
    Received requests:
    - Request #0
            RPC name: get_commitment_state
            Requests Protobuf as JSON
            {}
    - Request #1
            RPC name: get_genesis_info
            Requests Protobuf as JSON
            {}
    - Request #2
            RPC name: get_filtered_sequencer_block
            Requests Protobuf as JSON
            {
              "height": "3",
              "rollup_ids": [
                {
                  "inner": "KioqKioqKioqKioqKioqKioqKioqKioqKioqKioqKio="
                }
              ]
            }
    - Request #3
            RPC name: execute_block
            Requests Protobuf as JSON
            {
              "prev_block_hash": "AQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQ==",
              "transactions": [
                {
                  "sequenced_data": "aGVsbG9fd29ybGQ="
                }
              ],
              "timestamp": "1970-01-01T00:00:01.000000001+00:00"
            }
    - Request #4
            RPC name: update_commitment_state
            Requests Protobuf as JSON
            {
              "commitment_state": {
                "soft": {
                  "number": 2,
                  "hash": "AgICAgICAgICAgICAgICAgICAgICAgICAgICAgICAgICAgICAgICAgICAgICAgICAgICAgICAgICAgICAgICAg==",
                  "parent_block_hash": "AQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQ==",
                  "timestamp": "1970-01-01T00:00:01.000000001+00:00"
                },
                "firm": {
                  "number": 2,
                  "hash": "AgICAgICAgICAgICAgICAgICAgICAgICAgICAgICAgICAgICAgICAgICAgICAgICAgICAgICAgICAgICAgICAg==",
                  "parent_block_hash": "AQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQ==",
                  "timestamp": "1970-01-01T00:00:01.000000001+00:00"
                },
                "base_celestia_height": "1"
              }
            }
    - Request #5
            RPC name: get_filtered_sequencer_block
            Requests Protobuf as JSON
            {
              "height": "3",
              "rollup_ids": [
                {
                  "inner": "KioqKioqKioqKioqKioqKioqKioqKioqKioqKioqKio="
                }
              ]
            }
    
    
    
    failures:
        soft_and_firm::simple
    
    test result: FAILED. 10 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 2.23s
    
    error: test failed, to rerun pass `-p astria-conductor --test blackbox`</details>

## Explanation
The `soft_and_firm::simple` blackbox tests has 4 guards that are asserted:

1. execution
1. soft update commitment state
1. firm update commitment state

The problem arises when the firm block (via Celestia) is received before the soft block (from Sequencer), leading to step 2 being skipped. The error log above shows that one update commitment state was received, with the block number being set to `firm.number = soft.number = 2`. This is expected behavior and consistent with the soft-execution being skipped.

The fix consists in changing to soft-update-commitment to `.expect(0..=1)` (can be received at most 1 time, but can also be skipped).

ethanoroshiba:
## Summary
Fixed flakiness in conductor `soft_and_firm::simple` test.

## Background
Test would previously fail ~6% of the time due to receiving firm block before soft block, hence never fulfilling the soft update commitment mock.

## Changes
- Split `simple()` into 2 tests:
1. Tests for receiving soft block first, and then firm block (and enforcing this order).
2. Tests for receiving firm block first, then ignoring the later soft block response at the same height. The tests for receiving the soft block at the next height.
- made changes to `astria-grpc` to enable a delayed response

## Testing
All tests passing.

## Related Issues
closes #1143


Fraser999 (changes requested):
This still fails for me (I'm running the test in a loop on an Ubuntu box).

I think we might need to either enforce the ordering in the test somehow (could be complicated - I haven't dug into the code), or maybe the test is pointing to a failure of the production code to handle a legitimate edge case?

crates/astria-conductor/tests/blackbox/soft_and_firm.rs
```diff
@@ -23,21 +20,13 @@ use crate::{
...
+/// ignored).
 ///
-/// NOTE: there is a potential race condition in this test in that the information could be first
-/// retrieved from Celestia before Sequencer and executed against the rollup. In that case step 3.
-/// would be skipped (no soft commitment update).
+/// This is to test for the unlikely scenario of receiving a firm block from Celestia before the
+/// soft block from sequencer.
 #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
-async fn simple() {
+async fn simple_firm_only() {
```
SuperFluffy, line 29: Can you give this a more meaningful name? For example:
```suggestion
async fn soft_is_ignored_if_firm_arrives_first() {
```

crates/astria-conductor/tests/blackbox/soft_and_firm.rs
```diff
@@ -86,6 +82,110 @@ async fn simple() {
...
+///
+/// The following steps are most important:
+/// 1. a block at rollup number 1, sequencer height 2 is fetched from Sequencer
+/// 2. the block is executed against the rollup
+/// 3. the rollup's soft commitment state is updated to reflect the last execution
+/// 4. block information for rollup number 1, sequencer height 2 is reconstructed from Celestia
+///    height 1
+/// 5. the rollup's firm commitment state is updated (but without executing the block)
+#[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+async fn simple_soft_first() {
```
SuperFluffy, line 153:
```suggestion
async fn executes_soft_first_then_updates_firm() {
```

crates/astria-conductor/tests/blackbox/soft_and_firm.rs
```diff
@@ -23,21 +20,13 @@ use crate::{
...
-/// 4. block information for rollup number 1, sequencer height 2 is reconstructed from Celestia
-///    height 1
-/// 5. the rollup's firm commitment state is updated (but without executing the block)
+/// Tests if a single block is executed and the rollup's state updated (first firm, then soft is
+/// ignored).
 ///
-/// NOTE: there is a potential race condition in this test in that the information could be first
-/// retrieved from Celestia before Sequencer and executed against the rollup. In that case step 3.
-/// would be skipped (no soft commitment update).
+/// This is to test for the unlikely scenario of receiving a firm block from Celestia before the
```
SuperFluffy, line 26:
```suggestion
/// This is to test for the scenario of receiving a firm block from Celestia before the
```

crates/astria-conductor/tests/blackbox/soft_and_firm.rs
```diff
@@ -86,6 +82,110 @@ async fn simple() {
...
+            hash: [2; 64],
+            parent: [1; 64],
+        ),
+        base_celestia_height: 1,
+        expected_calls: 0,
+    );
+
+    timeout(
+        Duration::from_millis(500),
+        update_commitment_state_soft.wait_until_satisfied(),
```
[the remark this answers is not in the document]
SuperFluffy: I just remembered: if you bind the guard to a variable, then its condition will be checked at drop. So you don't need to change anything in the macros, just do:

```rust
let _the_guard = mount_the_update_commitment_state!(...);
```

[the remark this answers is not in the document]
ethanoroshiba: Considering this test is basically just firm only, I'm wondering if it makes more sense to incorporate its functionality with `executes_firm_then_soft_at_next_height()`? We could add two filtered sequencer block responses, one with a shorter delay at the first height, then one with a longer delay at the next height. Then, we check that there are no calls to update commitment state at the original height, and leave the rest of the test as normal. What do you think?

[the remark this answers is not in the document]
SuperFluffy: Yeah, sounds reasonable.

crates/astria-conductor/tests/blackbox/helpers/macros.rs
```diff
@@ -182,7 +182,8 @@ macro_rules! mount_update_commitment_state {
         mock_name: $mock_name:expr,
         firm: ( number: $firm_number:expr, hash: $firm_hash:expr, parent: $firm_parent:expr$(,)? ),
         soft: ( number: $soft_number:expr, hash: $soft_hash:expr, parent: $soft_parent:expr$(,)? ),
-        base_celestia_height: $base_celestia_height:expr
+        base_celestia_height: $base_celestia_height:expr,
+        expected_calls: $expected_calls:expr
```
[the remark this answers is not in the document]
ethanoroshiba: I had originally not done this since I had thought the default of having no `mock_name` or `expected_calls` fields would be good enough to cover both; just implemented another default for having `mock_name` but not `expected_calls` as well as you suggested :)

crates/astria-grpc-mock/src/mounted_mock.rs
```diff
@@ -125,6 +125,12 @@ impl MountedMock {
             return MockResult::NoMatch;
         }
 
+        if let Some(delay) = self.inner.delay {
```
SuperFluffy, line 128: This is not the right place to do this: 

1. It's conceptually wrong: why does the mock only match after some time? Why does the mock not match before `duration`, but then start matching after `duration`?
2. instead, we should use a `tokio::time::sleep` future that actually responds after some time has passed
3. this cannot be done in `MountedMock::match_and_respond`, because it will stall the entire mock set. See the definition of `MockServer::handle_request` which locks the `RwLock` for the entire duration for handling the request

crates/astria-grpc-mock/src/mock.rs
```diff
@@ -73,6 +75,12 @@ impl Mock {
     pub async fn mount_as_scoped(self, server: &MockServer) -> MockGuard {
         server.register_as_scoped(self).await
     }
+
+    #[must_use = "a mock must be mounted on a server to be useful"]
+    pub fn set_delay(mut self, delay: Option<std::time::Duration>) -> Self {
```
SuperFluffy, line 80: Put this on the response constructor, not the mock.

crates/astria-grpc-mock/src/response.rs
SuperFluffy, line 1: For this file: looks like this original architecture has run its course and creates a ton of boilerplate for no gain.

Can we do something similar to wiremock and create a `ResponseTemplate` similar to theirs?

I think the main change will be moving the trait object from `Mock` to some `Response`/`ResponseTemplate` type, so this would probably end up looking something like this:

```rust
struct ResponseTemplate {
    // taken from mock
    response: Box<dyn Response>,
    delay: Option<Duration>,
}

// and in mock
struct Mock {
    response: ResponseTemplate,
}
```

crates/astria-conductor/tests/blackbox/helpers/macros.rs
```diff
@@ -122,16 +123,31 @@ macro_rules! mount_celestia_blobs {
...
             .mount_celestia_blob_get_all(
                 $celestia_height,
                 $crate::rollup_namespace(),
                 vec![blobs.rollup],
+                $delay,
             )
             .await
     }};
+    (
+        $test_env:ident,
```
SuperFluffy, line 139: I'd flip this and put the special case as the first match, and the base case (which includes the `$delay` pattern) as the last arm of the pattern match.

crates/astria-grpc-mock/src/response.rs
SuperFluffy, line 1: The changes to `ResponseTemplate` (and `MockResponse`) closely follow wiremock, but I think I would like to change from their convention because it does not provide a way to pass a delay to the dynamic response.

Please revert the changes to the `Respond` trait and have it still respond a `MockResponse`, and have `Mock` take a `ResponseTemplate` instead of a `Box<dyn Response>`.

Change `ResponseTemplate` to read:

```rust
struct ResponseTemplate {
    response: Box<dyn Respond>,
    delay: Option<Duration>,
}

impl ResponseTemplate {
    pub(crate) fn respond(&self, req: &tonic::Request<AnyMessage>) -> (MockResponse, Option<Duration>) {
        (self.response(req), self.delay)
    }
}
```

Then `dynamic_response` will give you the `ResponseTemplate` and you can still set a delay on that, bringing everything together nicely.

crates/astria-grpc-mock/src/response.rs
```diff
@@ -119,6 +138,26 @@ impl Clone for MockResponse {
...
+impl ResponseTemplate {
+    pub(crate) fn respond(
+        &self,
+        req: &tonic::Request<AnyMessage>,
+    ) -> (ResponseResult, Option<Duration>) {
+        (self.response.respond(req), self.delay)
+    }
+
+    #[must_use]
+    pub fn set_delay(mut self, delay: Option<Duration>) -> Self {
```
SuperFluffy, line 155: Nit: I think I'd change this to `fn set_delay(mut self, delay: Duration) -> Self {`

This makes the calls in conductor's mocks slightly less elegant, but IMO makes more sense from a user's point of view.

crates/astria-conductor/tests/blackbox/soft_and_firm.rs
```diff
@@ -87,20 +124,135 @@ async fn simple() {
...
+        ),
+        soft: (
+            number: 2,
+            hash: [2; 64],
+            parent: [1; 64],
+        ),
+        base_celestia_height: 1,
+    );
+
+    let _stale_update_soft_commitment_state = mount_update_commitment_state!(
```
SuperFluffy, line 253: This would be very surprising to a reader: make a comment above this to explain why this is bound like that and not `wait_until_satisfied` (answer: so that the mock condition is tested on drop; this should actually be mounted on the server instead of returning a guard, but is a workaround for how the helpers for the conductor blackbox tests are implemented).

SuperFluffy (approved):
Looks great. I left a few minor nits but otherwise think this is ready to be merged.

Can you create a followup issue to test delayed responses in the mock crate?

Fraser999 (approved):
I haven't reviewed the code changes in depth, but running the new/updated tests in a loop indicates the issue is fixed now, so approving on the basis.

PR #1482, opened 2024-09-10
feat(conductor)!: implement chain ID checks

### LINKED ISSUE #986: validate communicating with correct sequencer chain

ethanoroshiba:
## Summary
Implemented chain ID validation for both the sequencer and Celestia readers in Conductor.

## Background
This is to safeguard against connecting to an incorrect network.

This change was previously blocking on a switch to solely using gRPC for communication with sequencer, but after offline discussion it was decided to implement the chain ID check anyways using the HTTP client.

## Changes
- Added sequencer and Celestia chain ID environment variables.
- Added chain ID checks to the initialization of both the sequencer and Celestia readers.
- Changed conductor `run_until_stopped()` to return with an error so that it will not fail silently.

## Testing
Added 2 blackbox tests to ensure the conductor exits with the proper error in the event of both a Celestia and sequencer chain ID mismatch.

## Breaking Changelist
- Added `ASTRIA_CONDUCTOR_SEQUENCER_CHAIN_ID` and `ASTRIA_CONDUCTOR_CELESTIA_CHAIN_ID` environment variables, without which conductor will not function.

## Related Issues
closes #986


aajimal (approved):

noot (approved):

SuperFluffy (changes requested):
Blocking this merge because this change exposes private types publicly

crates/astria-conductor/src/celestia/builder.rs
```diff
@@ -23,6 +23,7 @@ pub(crate) struct Builder {
     pub(crate) executor: executor::Handle,
     pub(crate) sequencer_cometbft_client: SequencerClient,
     pub(crate) sequencer_requests_per_second: u32,
+    pub(crate) celestia_chain_id: String,
```
SuperFluffy, line 26:
```suggestion
    pub(crate) expected_celestia_chain_id: String,
```

crates/astria-conductor/tests/blackbox/soft_only.rs
```diff
@@ -1,20 +1,39 @@
 use std::time::Duration;
 
-use astria_conductor::config::CommitLevel;
+use astria_conductor::{
+    conductor::SequencerChainIdError,
```
SuperFluffy, line 4: no

crates/astria-conductor/tests/blackbox/firm_only.rs
```diff
@@ -1,14 +1,43 @@
 use std::time::Duration;
 
-use astria_conductor::config::CommitLevel;
+use astria_conductor::{
+    conductor::CelestiaChainIdError,
```
SuperFluffy, line 4: no

crates/astria-conductor/src/config.rs
```diff
@@ -61,6 +61,12 @@ pub struct Config {
     /// The number of requests per second that will be sent to Sequencer.
     pub sequencer_requests_per_second: u32,
 
+    /// The chain ID of the sequencer network the conductor should be communiacting with.
+    pub sequencer_chain_id: String,
```
SuperFluffy, line 65: bikeshedding: i wonder if calling this `expected_sequencer_chain_id` would be more descriptive.

(and same for the one below)

crates/astria-conductor/src/sequencer/builder.rs
```diff
@@ -10,6 +10,7 @@ pub(crate) struct Builder {
     pub(crate) sequencer_grpc_client: SequencerGrpcClient,
     pub(crate) sequencer_cometbft_client: sequencer_client::HttpClient,
     pub(crate) sequencer_block_time: Duration,
+    pub(crate) sequencer_chain_id: String,
```
SuperFluffy, line 13: Since this is a builder, this would be nicer:

```suggestion
    pub(crate) expected_sequencer_chain_id: String,
```

crates/astria-conductor/src/sequencer/mod.rs
```diff
@@ -55,6 +57,11 @@ mod reporting;
 pub(crate) use builder::Builder;
 pub(crate) use client::SequencerGrpcClient;
 
+#[derive(Debug, thiserror::Error)]
+pub enum ValidateChainIdError {
```
SuperFluffy, line 61: Private.

Also in line with the rest of the crate I would not type this out.

crates/astria-conductor/tests/blackbox/helpers/macros.rs
```diff
@@ -15,15 +15,15 @@ macro_rules! block {
 
 #[macro_export]
 macro_rules! celestia_network_head {
-    (height: $height:expr) => {
+    (height: $height:expr,chain_id: $chain_id:expr $(,)?) => {
```
SuperFluffy, line 18: Reduce noise by providing a pattern where `chain_id` is left unset, i.e. something like this:

```rust
macro_rules! celestia_network_head {
    (height: $height:expr) => {
        celestia_network_head!(height: $height, chain_id: crate::CELESTIA_CHAIN_ID,);
    }
    (height: $height:expr, chain_id: $expr) => {
        todo!("the actual definition");
    }
}
```

crates/astria-conductor/tests/blackbox/soft_only.rs
```diff
@@ -333,3 +360,97 @@ async fn requests_from_later_genesis_height() {
...
+    match res {
+        Ok(Ok(())) => panic!("conductor should have exited with an error, no error received"),
+        Ok(Err(e)) => {
+            let mut source = e.source();
+            while source.is_some() {
+                let err = source.unwrap();
+                if let Some(SequencerChainIdError::MismatchedSequencerChainId {
+                    expected,
+                    actual,
+                }) = err.downcast_ref::<SequencerChainIdError>()
```
SuperFluffy, line 442: Don't test implementation details in blackbox tests.

It's fine to assert that it exits after receiving a bad chain ID.

crates/astria-conductor/tests/blackbox/firm_only.rs
```diff
@@ -377,3 +406,109 @@ async fn fetch_from_later_celestia_height() {
...
+    .await;
+
+    let res = conductor.await;
+    match res {
+        Ok(Ok(())) => panic!("conductor should have exited with an error, no error received"),
+        Ok(Err(e)) => {
+            let mut source = e.source();
+            while source.is_some() {
+                let err = source.unwrap();
+                if let Some(CelestiaChainIdError::MismatchedCelestiaChainId {
```
SuperFluffy, line 499: As above, don't assert on implementation details in blackbox tests. Just check behaviour.

SuperFluffy (changes requested):
Don't expose implementation details in the public API of service crates.

crates/astria-conductor/tests/blackbox/firm_only.rs
ethanoroshiba: With checking the error source text itself, I've managed to work it such that the tests still check the error result without exposing any underlying types. It feels like the tests should probably ensure that the correct thing is causing the error, as opposed to succeeding for any kind of early exit, but I understand this is fragile so I can change if needed.

crates/astria-conductor/src/celestia/mod.rs
```diff
@@ -170,6 +175,16 @@ impl Reader {
     async fn initialize(
```
SuperFluffy, line 175: This `initialize` function also fetches the sequencer chain ID a few lines down: after all, the celestia reader does two things:

1. read blobs from celestia
2. verify the sequencer block information extracted from the blobs against the connected sequencer.

This function only checks that the celestia chain ID matches, but not if the sequencer chain ID against the blocks will be verified matches!

Please update this method to:

1. also check the sequencer chain ID,
2. and to perform all the 3 checks (test for celestia chain ID, waiting for the executor, test for sequencer chain ID) concurrently using `try_join!`

crates/astria-conductor/src/celestia/mod.rs
```diff
@@ -187,6 +202,31 @@ impl Reader {
     }
 }
 
+#[instrument(skip_all, err)]
+async fn get_celestia_chain_id(celestia_client: &CelestiaClient) -> eyre::Result<String> {
```
SuperFluffy, line 206: Just return the (celestia_tendermint specific) `chain::Id` similar to what `get_sequencer_chain_id` further down is doing.

crates/astria-conductor/src/celestia/mod.rs
```diff
@@ -610,7 +650,9 @@ async fn enqueue_block(
 }
 
 #[instrument(skip_all, err)]
-async fn get_sequencer_chain_id(client: SequencerClient) -> eyre::Result<tendermint::chain::Id> {
+pub(crate) async fn get_sequencer_chain_id(
```
SuperFluffy, line 653: Please don't make this crate public here. I'd rather have this method copied to the `sequencer` module.

crates/astria-conductor/src/sequencer/mod.rs
```diff
@@ -99,6 +103,18 @@ impl Reader {
 
     #[instrument(skip_all, err)]
     async fn initialize(&mut self) -> eyre::Result<executor::Handle<StateIsInit>> {
+        let actual_sequencer_chain_id =
+            get_sequencer_chain_id(self.sequencer_cometbft_client.clone())
+                .await
+                .wrap_err("failed to get chain ID from Sequencer")?
+                .to_string();
```
SuperFluffy, line 110: Why allocate? It provides a way to get a str slice which you can check against: https://docs.rs/tendermint/latest/tendermint/chain/id/struct.Id.html#method.as_str

crates/astria-conductor/src/celestia/mod.rs
```diff
@@ -169,24 +177,76 @@ impl Reader {
...
         &mut self,
-    ) -> eyre::Result<(executor::Handle<StateIsInit>, tendermint::chain::Id)> {
+    ) -> eyre::Result<((), executor::Handle<StateIsInit>, tendermint::chain::Id)> {
+        let validate_celestia_chain_id = async {
+            let actual_celestia_chain_id = get_celestia_chain_id(&self.celestia_client)
+                .await
+                .wrap_err("failed to fetch Celestia chain ID")?;
+            let expected_celestia_chain_id = &self.expected_celestia_chain_id;
+            ensure!(
+                self.expected_celestia_chain_id == actual_celestia_chain_id.to_string(),
```
SuperFluffy, line 187: Why allocate? It provides a way to get a str slice which you can check against: https://docs.rs/tendermint/latest/tendermint/chain/id/struct.Id.html#method.as_str

(celestia is actually using a different tendermint library, but that one, too, should provide `as_str`)

crates/astria-conductor/tests/blackbox/firm_only.rs
```diff
@@ -377,3 +405,111 @@ async fn fetch_from_later_celestia_height() {
...
+                        "expected Celestia chain id `{CELESTIA_CHAIN_ID}` does not match actual: \
+                         `{bad_chain_id}`"
+                    )
+                    .as_str(),
+                ) {
+                    return;
+                }
+                source = err.source();
+            }
+            panic!("conductor exited with incorrect error: {e}")
```
SuperFluffy, line 511: You probably want `{e:?}` here.

crates/astria-conductor/tests/blackbox/helpers/mod.rs
```diff
@@ -134,7 +136,7 @@ impl Drop for TestConductor {
                 {
                     Ok(Ok(_)) => None,
                     Ok(Err(conductor_err)) => Some(format!(
-                        "conductor shut down with an error:\n{conductor_err:?}"
+                        "conductor failed during shutdown:\n{conductor_err:?}"
```
SuperFluffy, line 139: I am not sure this change clarifies anything...

crates/astria-conductor/tests/blackbox/firm_only.rs
```diff
@@ -377,3 +405,111 @@ async fn fetch_from_later_celestia_height() {
...
+
+#[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+async fn exits_on_celestia_chain_id_mismatch() {
+    use astria_grpc_mock::{
+        matcher,
+        response as GrpcResponse,
+        Mock as GrpcMock,
+    };
+
+    // We have to create our own test conductor and perform mounts manually because `TestConductor`
```
SuperFluffy, line 417: Instead of doing all that lifting here, could you have made use of https://doc.rust-lang.org/std/mem/struct.ManuallyDrop.html to inhibit the custom `Drop` from being called?

ethanoroshiba: In attempting to do so, I'm not sure this method will work either. Calling `await` on the conductor contained by `ManuallyDrop<TestConductor>` errs because await takes ownership of the conductor. `ManuallyDrop` does have a method `take()` which would allow for this, but it is an unsafe function and as far as I know these can't be used in tests

SuperFluffy: OK, let's leave this for now.

crates/astria-conductor/tests/blackbox/soft_only.rs
```diff
@@ -333,3 +359,96 @@ async fn requests_from_later_genesis_height() {
...
+                        "expected chain id `{SEQUENCER_CHAIN_ID}` does not match actual: \
+                         `{bad_chain_id}`"
+                    )
+                    .as_str(),
+                ) {
+                    return;
+                }
+                source = err.source();
+            }
+            panic!("conductor did not exit with incorrect error: {e}")
```
SuperFluffy, line 450:
```suggestion
            panic!("expected exit due to chain ID mismatch, but got a different error: {e:?}")

SuperFluffy (approved):
I think this is ready to get merged. :) Thank you for your work!

To bring this over the line I  did two things:

+ flatten the return value of the `conductor::Handle` future by giving some extra context;
+ replaced the undocumented `eyre::ErrReport` alias by `eyre::Report`.

Issue #1416, opened 2024-08-28
sequencer: enforce only certain actions can be bundled


Lilyjjo:
For #1412's implementation, after adding transaction categories of #1414, a next step would be using the transaction categories to enforce that only certain actions could be together in bundles. 

This would be checked in `CheckTx` before adding actions to the mempool as part of the routine static checks for transaction validity.

┆Issue Number: ENG-753

Issue #1412, opened 2024-08-27
block ordering and tx inclusion action rules


joroshiba:
Sudo actions have some interesting side effects to the rest of the block, and as such should be posted at the bottom of the block with no non sudo actions occuring after them. We may even want a hierarchical order in which groupings can occur, and rules about coupling:

I propose the following transaction groupings, all actions in the same groupings have a property of if they can be bundled. If they can, they can only be bundled with same grouping txs. They are in order of how they should appear in a block.

1. General
    - These are general purpose txs which have no odd implications on later txs
    - Bundlable? ✅
    - Can charge fees
    - Actions
      - `TransferAction`
      - `SequenceAction`
      - `BridgeLockAction`
      - `BridgeUnlockAction`
      - `IbcRelay`
      - `Ics20Withdrawal`
      - `ValidatorUpdate`
2. Bridge Control
    - Bundlable? ❌
    - These actions change fundamentals on the signing account. They can't be bundled cleanly because same signer can't do another action after it. Additionally, can have implications on txs in general validity based on ordering. Putting after makes clean break.
    - Charge fees
    - Actions
      - `InitBridgeAccountAction`
      - `BridgeSudoChangeAction`
3. Sudo Actions
    - Bundlable? ✅
    - Most of these have implications on fees charged and valid, affect all txs which have fees and must be after any tx which can charge fees. IbcRelayer affects the validity of `IbcRelay` actions as such should be towards end of block.
    - No fees, submittable only by sudo actor
    - Actions
      - `FeeAssetChangeAction`
      - `FeeChangeAction`
      - `IbcRelayerChangeAction`
4. Sudo Control
    - Bundlable? ❌
    - This updates the sudo account, impossible to have more than one in a valid bundle. The sudo account is also the payee of fees, so it needs to be after all fee paying items, it would affect the validity of any sudo actions. I think it's best to include after any such actions occur.
    - `SudoAddressChangeAction`

I think these might be best viewed as a sort of "action category" where a transaction can only have actions that are all the same category, and categories have two properties (priority, and if it can be bundled). Thus the category of the first action would the that for the whole transaction, and stateless checks can easily be added over this. 

┆Issue Number: ENG-750

noot:
regarding #2 (bridge control):
> These actions change fundamentals on the signing account. They can't be bundled cleanly because same signer can't do another action after it. Additionally, can have implications on txs in general validity based on ordering. Putting after makes clean break.

[...]

PR #1486, opened 2024-09-11
feat(sequencer)!: introduce transaction categories (non-type version)

### LINKED ISSUE #1414: sequencer: add notion of transaction categories

- Sudo Actions 
- Bridge Control
- Sudo Control

┆Issue Number: ENG-751

Lilyjjo:
## Summary
Adds the notion of `PriorityCategories` for `Actions`. Enforces that, for `UnsignedTransactions` to be valid, that their contained `Actions` are all of the same `PriorityCategory` and are limited to a single action if the `PriorityCategory` is not bundleable. 

## Background

## Testing
TODO

## Metrics
TODO

## Breaking Changelist
TODO, assuming that this is breaking 

## Related Issues
Initial steps for #1412

closes #1414, #1416

### DIFF
--- a/crates/astria-core/src/protocol/transaction/v1alpha1/action.rs
+++ b/crates/astria-core/src/protocol/transaction/v1alpha1/action.rs
@@ -45,6 +45,50 @@ pub enum Action {
     FeeChange(FeeChangeAction),
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub enum PriorityCategory {
+    BundlableGeneral,
+    General,
+    BundleableSudo,
+    Sudo,
+}
+
+pub trait Priority {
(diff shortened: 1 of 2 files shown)


crates/astria-core/src/protocol/transaction/v1alpha1/action.rs
```diff
@@ -45,6 +45,50 @@ pub enum Action {
...
+    Sudo,
+}
+
+pub trait Priority {
+    fn priority(&self) -> PriorityCategory;
+}
+
+pub trait Bundleable {
+    fn is_bundleable(&self) -> bool;
+}
```
joroshiba, line 62: I'm realizing now we probably don't even need the traits this can just be an impl. No need for the trait if we don't need multiple types implementing it.

PR #1487, opened 2024-09-11
chore(sequencer): change app/test_utils.rs addresses to versions with known private keys

Lilyjjo:
## Summary
Updated public addresses in `app/test_util.rs` to addresses with known private keys.

## Background
Needed for PR #1408. We want to separate out the 'breaking change' into a different PR since it is unrelated.


SuperFluffy (approved):
I have added a section on breaking changes to explain why this is not a breaking change.

PR #1488, opened 2024-09-11

noot:
## Summary
patch penumbra and our hermes fork temporarily so that height is passed from hermes's requests to the ibc grpc server.

## Background
there was an IBC proof validation failure error on dusk-10. the computed app hash was one height *after* the expected app hash, which could be caused by the grpc queries always using the latest state snapshot, instead of at the height intended by hermes. however the grpc requests don't contain the height (:/) so the hack is to put the height in the grpc request metadata, which is then used by the grpc query server to use the correct snapshot.

the actual fix is to get https://github.com/cosmos/ibc-go/pull/5685 merged and put the heights in the actual grpc requests.

## Changes
- patch penumbra dep to https://github.com/noot/penumbra/pull/3 temporarily
- hermes commit in chart is https://github.com/astriaorg/hermes/pull/12

## Testing
ibc smoke test

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8852,3 +8852,28 @@ dependencies = [
  "cc",
  "pkg-config",
 ]
+
+[[patch.unused]]
+name = "cnidarium"
+version = "0.80.4"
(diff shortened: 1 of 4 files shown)


noot:
closing in favour of #1491 for now

PR #1489, opened 2024-09-11
Feature: Add `graph-node` to charts.

sambukowski:
Added a `graph-node` helm chart to the charts.


charts/deploy.just
```diff
@@ -53,6 +53,11 @@ deploy-metrics-server:
 deploy-ingress-controller:
   kubectl apply -f https://raw.githubusercontent.com/kubernetes/ingress-nginx/main/deploy/static/provider/kind/deploy.yaml
 
+deploy-graph-node namespace=defaultNamespace: (deploy-chart "graph-node" namespace)
+delete-graph-node namespace=defaultNamespace: (delete-chart "graph-node" namespace)
+delete-pvc namespace=defaultNamespace:
```
quasystaty1, line 58: one wrong command and can delete an entire cluster PVC, should remove or delete only graph pvcs?

charts/graph-node/templates/deployment.yaml
```diff
@@ -0,0 +1,120 @@
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: graph-node
```
quasystaty1, line 5: should add namespace override for all templates, similar to other charts?

charts/graph-node/templates/services.yaml
```diff
@@ -0,0 +1,50 @@
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: graph-node
+spec:
+  selector:
+    app: graph-node
+  ports:
```
quasystaty1, line 9: nit: configurable ports

charts/graph-node/values.yaml
```diff
@@ -0,0 +1,46 @@
+serviceMonitor:
```
quasystaty1, line 1: Can add a global values and template them down the chart
```suggestion
global:
  namespaceOverride: ""
  replicaCount: 1

serviceMonitor:
```

quasystaty1 (approved):
LGTM

charts/graph-node/templates/deployment.yaml
```diff
@@ -0,0 +1,123 @@
...
+        - name: postgres_pass
+          value: {{ .Values.postgres.password }}
+        - name: postgres_db
+          value: {{ .Values.postgres.database }}
+        - name: ipfs
+          value: ipfs:5001
+        - name: ethereum
+          value: {{ .Values.environment.ethereumNetwork }}:{{ .Values.environment.ethereumRPC }}
+        - name: GRAPH_LOG
+          value: info
```
joroshiba, line 40: Prefer generally to pass env vars via config map to seperate configuration from deployment.

charts/graph-node/templates/deployment.yaml
```diff
@@ -0,0 +1,123 @@
...
+        ports:
+        - containerPort: {{ .Values.ipfs.ports.api | default 5001 }}
+        volumeMounts:
+        - name: ipfs-storage
+          mountPath: /data/ipfs
+      volumes:
+      - name: ipfs-storage
+        persistentVolumeClaim:
+          claimName: ipfs-pvc
+
```
joroshiba, line 70: nit: remove empty line

charts/graph-node/templates/deployment.yaml
```diff
@@ -0,0 +1,123 @@
...
+        - name: ipfs
+          value: ipfs:5001
+        - name: ethereum
+          value: {{ .Values.environment.ethereumNetwork }}:{{ .Values.environment.ethereumRPC }}
+        - name: GRAPH_LOG
+          value: info
+
+---
+apiVersion: apps/v1
+kind: StatefulSet
```
joroshiba, line 44: nit: statefulsets should be in statefulsets.yaml instead of deployments.yaml

charts/graph-node/templates/persistent-volume-claims.yaml
```diff
@@ -0,0 +1,25 @@
...
+metadata:
+  name: ipfs-pvc
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: {{ .Values.ipfs.storage.size }}
+  storageClassName: {{ .Values.persistence.storageClass }}
+
```
joroshiba, line 13: nit: unnecessary line

charts/graph-node/templates/services.yaml
```diff
@@ -0,0 +1,53 @@
...
+kind: Service
+metadata:
+  name: graph-node
+  namespace: {{ .Values.global.namespaceOverride | default .Release.Namespace }}
+spec:
+  selector:
+    app: graph-node
+  ports:
+    - name: http
+      port: {{ .Values.graphNode.ports.http | default 8000 }}
```
joroshiba, line 12: We shouldn't need to specify defaults here if we provide defaults in the value file

joroshiba (commented):
High level nit, should be enforced with a linter, I can work on that. Prefer to have lists have an indent for clarity, although both are valid yaml I think it's a bit easier to read
```
entry:
  - item
```
instead of 
```
entry:
- item
```

PR #1491, opened 2024-09-11

noot:
same as #1488 but based off sequencer tag 0.16.0.


noot:
closing for #1691

Issue #1436, opened 2024-08-30
Replace cnidarium `StateDelta`, `StateRead` and `StateWrite` types


Fraser999:
Replace these types with our own as described in [Block State Improvements in the Sequencer](https://astria.atlassian.net/wiki/spaces/Astria/pages/36864002/Block+State+Improvements+in+the+Sequencer).

joroshiba:
This issue is stale because it has been open 45 days with no activity. Remove stale label or this issue
be closed in 7 days.

PR #1492, opened 2024-09-11
chore(sequencer)!: exclusively use Borsh encoding for stored data (ENG-768)

### LINKED ISSUE #1434: Use single serialization format for all block state data
Format to use will be decided once #1433 is completed and discussed.

Every type of data to be encoded should have a separate type in a new `storage` module, and potentially we should have an enum of all these types, with an instance of the enum being what is actually encoded and stored to disk (although this is not agreed upon at the moment).

┆Issue Number: ENG-768

Fraser999:
## Summary
This PR primarily changes the encoding format of all data being written to storage to Borsh.

## Background
We currently have a variety of different encoding formats, and this can be confusing, sub-optimal (in terms of storage space consumed and serialization/deserialization performance) and potentially problematic as e.g. JSON-encoding leaves a lot of flexibility in how the actual serialized data can look.

This PR is part one of three which aim to improve the performance and quality of the storage component.  As such, the APIs of the various `StateReadExt` and `StateWriteExt` extension traits were updated slightly in preparation for the upcoming changes.  In broad terms, for getters this meant having ref parameters rather than value ones (even for copyable types like `[u8; 32]` this is significantly more performant), and for "putters", parameters which are used for DB keys are refs, while the DB value parameters become values, since in the next PR these values will be added to a cache.

## Changes
- Added a new `storage` module.  This will ultimately contain our own equivalents of the `cnidarium` types, but for now consists only of a collection of submodules for types currently being written to storage.  There is a top-level enum `StoredValue` which becomes the only type being written to storage by our own code.
- To accommodate for converting between the storage types and the corresponding domain types defined in `astria-core` and `astria-merkle`, some of these have been provided with constructors named `unchecked_from_parts`.  This allows the type to be constructed from a trusted source like our own DB, skipping any validation steps which might otherwise be done.
- Updated all `StateReadExt` and `StateWriteExt` traits to use the new `StoredValue`, which internally uses Borsh encoding.
- Updated the APIs of all these extension traits as described above.  This change resulted in a slew of similar updates to avoid needless copying `[u8; 32]` and `[u8; 20]` arrays, and that has unfortunately made the PR larger than expected.
- A few core types which currently derive or implement `serde` traits have had that removed, since it only existed to support JSON-encoding the type for storing.  In one case it was for JSON-encoding to a debug log line, but I replaced that with a `Display` impl.

## Testing
- Existing unit tests of the `StateReadExt` and `StateWriteExt` traits already had almost full coverage.  Where coverage was missing, I added round trip tests.

## Breaking Changelist
- The on-disk format of stored data has changed.

## Related Issues
Closes #1434.


SuperFluffy (commented):
High level comment on the approach taken in this design:

At a high level, sequencer is defined by a set of components, each of which provides ways to change and read their respective state. So far this also meant that each component had full control over how the objects written to its state are converted to/from the domain types.

The `storage` and `stored_value` modules invert this logic by defining and passing in the on-disk types into the components.

I am ambivalent about this change since on the one hand it potentially allows avoiding code duplication. But on the other hand it breaks with the separation between the components we had so far.

Fraser999:
> So far this also meant that each component had full control over how the objects written to its state are converted to/from the domain types.

I see that as one of the main problems with the existing approach.  It's because we've allowed this flexibility that different components use different serialization for the same domain types.  In a lot of cases, there is only the domain type - the component would serialize directly from the domain type into bytes.  I'm very keen to remove that flexibility so that all components _have_ to use `Borsh` and we have a complete collection of storage types which map to/from domain types, both serving different purposes (domain types supporting business logic only, storage types supporting being held in a DB only).

> The storage and stored_value modules invert this logic by defining and passing in the on-disk types into the components.

This is only temporary until #1436 is completed.  When that work is completed, the components will be dealing with the domain types only.

> But on the other hand it breaks with the separation between the components we had so far.

Well, I don't think there's complete separation currently.  For example, `accounts::StateReadExt` is used in a few other components.  But this isn't really changed with the new approach IMO.  Ultimately the components will continue to access/modify state relating to other components via their state extension traits.

SuperFluffy (commented):
Reviewed the changes in `astria-core` for now (most of which I think we should not have in this PR). Going to review the rest of sequencer now.

crates/astria-sequencer/src/storage/stored_value.rs
SuperFluffy, line 1: This module is so tiny, can we move this into `storage`?

crates/astria-sequencer/src/storage/stored_value.rs
```diff
@@ -0,0 +1,55 @@
...
+    pub(crate) fn serialize(&self) -> Result<Vec<u8>> {
+        borsh::to_vec(&self).wrap_err("failed to serialize stored value")
+    }
+
+    pub(crate) fn deserialize(bytes: &[u8]) -> Result<Self> {
+        borsh::from_slice(bytes).wrap_err("failed to deserialize stored value")
+    }
+}
+
+impl<'a> Display for StoredValue<'a> {
```
SuperFluffy, line 40: What's this display impl used for? For some form of tracing?

crates/astria-sequencer/src/ibc/storage/values.rs
SuperFluffy, line 1: high level comment that applies to all the types in this crate:

1. should there be a direct step from the lowest level value to the highest level `StoredValue`? In my mind the crate keeps its `storage::Value` as the only type that's converted to `StoredValue` so that the only conversions visible from outside the crate are, in this case, `ibc::storage::Value <> crate::storage::StoredValue`
2. Why are all storage values in here crate-public? Should other components/modules see their definition? In my mind inter-component communication happens through the domain types, and the on-disk format doesn't cross the component boundary.

crates/astria-core/src/primitive/v1/asset/denom.rs
```diff
@@ -282,6 +288,40 @@ impl TracePrefixed {
...
+    #[must_use]
+    pub fn base_denom(&self) -> &str {
+        &self.base_denom
+    }
+
+    /// This should only be used where the inputs have been provided by a trusted entity, e.g. read
+    /// from our own state store.
+    #[doc(hidden)]
+    #[must_use]
+    pub fn unchecked_from_parts<I: IntoIterator<Item = (String, String)>>(
```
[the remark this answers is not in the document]
Fraser999: > I understand that this is done to provide a faster way to go into and out of these types, but exposing this publicly makes me unhappy.

Me too.

> I am not sure how to work around this other than providing borsh de/ser traits in astria-core, which goes against the plan to keep that wire-types confined to astria-sequencer.

I guess the question is whether we see the storage types as part of our public API or not.  If we don't expect anything outside of sequencer to ever access the DB then it makes sense to me to keep all the storage types inside `astria-sequencer`.  If not, then I think they do indeed all belong somewhere like `astria-core`.

We could always feature-gate these "unchecked" constructors and leave the storage types inside sequencer to make things a little safer here I guess?

crates/astria-sequencer/src/ibc/ics20_withdrawal.rs
```diff
@@ -105,14 +107,14 @@ async fn create_ibc_packet_from_withdrawal<S: StateRead>(
...
 /// 3. `action.bridge_address` is unset, but `from` is a bridge account.
-async fn establish_withdrawal_target<S: StateRead>(
-    action: &action::Ics20Withdrawal,
+async fn establish_withdrawal_target<'a, S: StateRead>(
+    action: &'a action::Ics20Withdrawal,
     state: &S,
-    from: [u8; 20],
-) -> Result<[u8; 20]> {
+    from: &'a [u8; 20],
+) -> Result<&'a [u8; 20]> {
```
SuperFluffy, line 114: The lifetime `'a` might cause confusion on refactors.

crates/astria-sequencer/src/assets/state_ext.rs
```diff
@@ -192,30 +196,37 @@ impl<T: ?Sized + StateRead> StateReadExt for T {}
 #[async_trait]
 pub(crate) trait StateWriteExt: StateWrite {
     #[instrument(skip_all)]
-    fn put_native_asset(&mut self, asset: &asset::TracePrefixed) {
-        self.put_raw(NATIVE_ASSET_KEY.to_string(), asset.to_string().into_bytes());
+    fn put_native_asset(&mut self, asset: asset::TracePrefixed) -> Result<()> {
```
SuperFluffy, line 199: Intuitively it makes sense for a `put_*` method to take an object by value. On the other hand I notice that where before you could pass the asset by value, you know explicitly `clone` it, but then in the function body use a reference `&asset`.

Given that many other changes in this PR are performance focused, should we revert this diff?

crates/astria-sequencer/src/assets/state_ext.rs
```diff
@@ -192,30 +196,37 @@ impl<T: ?Sized + StateRead> StateReadExt for T {}
...
+        self.put_raw(NATIVE_ASSET_KEY.to_string(), bytes);
+        Ok(())
     }
 
     #[instrument(skip_all)]
-    fn put_ibc_asset(&mut self, asset: &asset::TracePrefixed) -> Result<()> {
-        let bytes = borsh::to_vec(&DenominationTrace(asset.to_string()))
-            .wrap_err("failed to serialize asset")?;
-        self.put_raw(asset_storage_key(asset), bytes);
+    fn put_ibc_asset(&mut self, asset: asset::TracePrefixed) -> Result<()> {
```
SuperFluffy, line 208: Intuitively it makes sense for a `put_*` method to take an object by value. On the other hand I notice that where before you could pass the asset by value, you know explicitly `clone` it, but then in the function body use a reference `&asset`.

Given that many other changes in this PR are performance focused, should we revert this diff?

crates/astria-sequencer/src/authority/action.rs
SuperFluffy, line 1: Comment on all the errors on the `put_*` methods: I'd prefer to keep the context around because it's easier to require this for all fallible errors calls rather than explain (and remember) where it's necessary and where it's not.

crates/astria-sequencer/src/authority/mod.rs
```diff
@@ -1,72 +1,63 @@
...
+use crate::accounts::AddressBytes;
 
-impl From<[u8; ADDRESS_LEN]> for ValidatorSetKey {
-    fn from(value: [u8; ADDRESS_LEN]) -> Self {
-        Self(value)
-    }
-}
+/// A map of public keys to validator updates.
+#[derive(Debug, PartialEq, Eq)]
+#[cfg_attr(test, derive(Clone))]
```
SuperFluffy, line 25: Why not just always derive clone?

crates/astria-sequencer/src/authority/storage/values.rs
```diff
@@ -0,0 +1,149 @@
...
+    ValidatorSet(ValidatorSet<'a>),
+}
+
+impl<'a> Display for Value<'a> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        match self {
+            Value::AddressBytes(address) => {
+                write!(f, "address {}", base64(address.0.as_ref()))
+            }
+            Value::ValidatorSet(_validator_set) => write!(f, "validator set"),
```
SuperFluffy, line 39: Some `Display` impls are relatively verbose, while this one is very sparse. Seems inconsistent? Let's make this apparent by using doing `validator set {...}`?

crates/astria-sequencer/src/app/storage/values/storage_version.rs
```diff
@@ -0,0 +1,51 @@
...
+use astria_eyre::eyre::bail;
+use borsh::{
+    BorshDeserialize,
+    BorshSerialize,
+};
+
+use super::Value;
+
+#[derive(Debug, BorshSerialize, BorshDeserialize)]
+pub(crate) struct StorageVersion(u64);
```
SuperFluffy, line 16: We should have a domain type for this too

crates/astria-sequencer/src/app/storage/values/revision_number.rs
```diff
@@ -0,0 +1,51 @@
...
+use astria_eyre::eyre::bail;
+use borsh::{
+    BorshDeserialize,
+    BorshSerialize,
+};
+
+use super::Value;
+
+#[derive(Debug, BorshSerialize, BorshDeserialize)]
+pub(crate) struct RevisionNumber(u64);
```
SuperFluffy, line 16: We should have a domain type for this if it requires an on-disk type.

crates/astria-sequencer/src/bridge/component.rs
SuperFluffy, line 1: Please just add context rather than have the comment do the heavy lifting.

crates/astria-sequencer/src/bridge/init_bridge_account_action.rs
```diff
@@ -93,21 +92,28 @@ impl ActionHandler for InitBridgeAccountAction {
...
+        state.put_bridge_account_rollup_id(&from, self.rollup_id)?;
         state
-            .put_bridge_account_ibc_asset(from, &self.asset)
+            .put_bridge_account_ibc_asset(&from, &self.asset)
             .wrap_err("failed to put asset ID")?;
-        state.put_bridge_account_sudo_address(from, self.sudo_address.map_or(from, Address::bytes));
+        // No need to add context as this method already reports sufficient context on error.
+        state.put_bridge_account_sudo_address(
+            &from,
+            self.sudo_address.map_or(from, |address| *address.bytes()),
```
SuperFluffy, line 103: This looked more elegant before. :) We could just add `as_bytes` to the `AddressBytes` trait and keep `bytes` as before.

crates/astria-sequencer/src/accounts/mod.rs
```diff
@@ -39,34 +37,19 @@ impl AddressBytes for Address {
...
 impl AddressBytes for [u8; ADDRESS_LEN] {
-    fn address_bytes(&self) -> [u8; ADDRESS_LEN] {
-        *self
-    }
-}
-
-impl AddressBytes for SignedTransaction {
-    fn address_bytes(&self) -> [u8; ADDRESS_LEN] {
-        self.address_bytes()
+    fn address_bytes(&self) -> &[u8; ADDRESS_LEN] {
```
SuperFluffy, line 40: There are a few instances where making this `[u8; ADDRESS_LEN]` seemed more elegant. We could just add another trait method `as_bytes` in addition to `bytes`?

crates/astria-core/src/primitive/v1/asset/denom.rs
```diff
@@ -507,8 +547,8 @@ impl IbcPrefixed {
     }
 
     #[must_use]
-    pub fn get(&self) -> [u8; 32] {
-        self.id
+    pub fn get(&self) -> &[u8; 32] {
```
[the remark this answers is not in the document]
Fraser999: Updated as discussed offline in [7fa7de3](https://github.com/astriaorg/astria/pull/1492/commits/7fa7de380896b42b810ce3e2c81ef5a6d07d450f).

crates/astria-core/src/primitive/v1/asset/denom.rs
```diff
@@ -134,6 +134,12 @@ impl<'a> From<&'a Denom> for IbcPrefixed {
     }
 }
 
+impl<'a> From<&'a IbcPrefixed> for IbcPrefixed {
```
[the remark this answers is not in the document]
Fraser999: It's used in several place of the new state ext APIs which favour taking items like this by ref where possible.

crates/astria-core/src/primitive/v1/mod.rs
```diff
@@ -116,18 +116,18 @@ impl RollupId {
...
     /// use astria_core::primitive::v1::RollupId;
     /// let bytes = [42u8; 32];
     /// let rollup_id = RollupId::new(bytes);
-    /// assert_eq!(bytes, rollup_id.get());
+    /// assert_eq!(bytes, *rollup_id.get());
     /// ```
     #[must_use]
-    pub const fn get(self) -> [u8; 32] {
-        self.inner
+    pub const fn get(&self) -> &[u8; 32] {
```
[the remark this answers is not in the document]
Fraser999: Updated as discussed offline in [7fa7de3](https://github.com/astriaorg/astria/pull/1492/commits/7fa7de380896b42b810ce3e2c81ef5a6d07d450f).

crates/astria-core/src/primitive/v1/mod.rs
```diff
@@ -684,8 +666,8 @@ impl TransactionId {
 
     /// Returns the 32-byte transaction hash.
     #[must_use]
-    pub fn get(self) -> [u8; TRANSACTION_ID_LEN] {
-        self.inner
+    pub fn get(&self) -> &[u8; TRANSACTION_ID_LEN] {
```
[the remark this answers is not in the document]
Fraser999: Updated as discussed offline in [7fa7de3](https://github.com/astriaorg/astria/pull/1492/commits/7fa7de380896b42b810ce3e2c81ef5a6d07d450f).

crates/astria-core/src/sequencerblock/v1alpha1/block.rs
```diff
@@ -92,8 +96,8 @@ pub struct RollupTransactions {
 impl RollupTransactions {
     /// Returns the [`RollupId`] identifying the rollup these transactions belong to.
     #[must_use]
-    pub fn rollup_id(&self) -> RollupId {
-        self.rollup_id
+    pub fn rollup_id(&self) -> &RollupId {
```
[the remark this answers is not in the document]
Fraser999: Will leave as is, as agreed offline.

crates/astria-core/src/sequencerblock/v1alpha1/block.rs
```diff
@@ -371,13 +392,13 @@ impl SequencerBlockHeader {
     }
 
     #[must_use]
-    pub fn rollup_transactions_root(&self) -> [u8; 32] {
-        self.rollup_transactions_root
+    pub fn rollup_transactions_root(&self) -> &[u8; 32] {
```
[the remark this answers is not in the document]
Fraser999: Will leave as is, as agreed offline.

crates/astria-core/src/sequencerblock/v1alpha1/block.rs
```diff
@@ -371,13 +392,13 @@ impl SequencerBlockHeader {
...
+        &self.rollup_transactions_root
     }
 
     #[must_use]
-    pub fn data_hash(&self) -> [u8; 32] {
-        self.data_hash
+    pub fn data_hash(&self) -> &[u8; 32] {
```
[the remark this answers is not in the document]
Fraser999: Will leave as is, as agreed offline.

crates/astria-core/src/sequencerblock/v1alpha1/block.rs
```diff
@@ -576,8 +620,8 @@ impl SequencerBlock {
     ///
     /// This is done by hashing the `CometBFT` header stored in this block.
     #[must_use]
-    pub fn block_hash(&self) -> [u8; 32] {
-        self.block_hash
+    pub fn block_hash(&self) -> &[u8; 32] {
+        &self.block_hash
```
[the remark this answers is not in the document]
Fraser999: Will leave as is, as agreed offline.

crates/astria-core/src/sequencerblock/v1alpha1/celestia.rs
```diff
@@ -183,8 +183,8 @@ impl SubmittedRollupData {
     }
 
     #[must_use]
-    pub fn sequencer_block_hash(&self) -> [u8; 32] {
-        self.sequencer_block_hash
+    pub fn sequencer_block_hash(&self) -> &[u8; 32] {
```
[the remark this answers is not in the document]
Fraser999: Will leave as is, as agreed offline.

crates/astria-core/src/sequencerblock/v1alpha1/celestia.rs
```diff
@@ -543,7 +543,7 @@ impl SubmittedMetadata {
     /// Returns the Merkle Tree Hash constructed from the rollup transactions of the original
     /// [`SequencerBlock`] this blob was derived from.
     #[must_use]
-    pub fn rollup_transactions_root(&self) -> [u8; 32] {
+    pub fn rollup_transactions_root(&self) -> &[u8; 32] {
```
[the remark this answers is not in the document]
Fraser999: Will leave as is, as agreed offline.

crates/astria-sequencer/src/accounts/action.rs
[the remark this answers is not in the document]
Fraser999: They predominantly are, specifically around taking keys by ref rather than value.  Will leave as is, as agreed offline.

crates/astria-sequencer/src/accounts/mod.rs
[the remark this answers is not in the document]
Fraser999: They predominantly are, specifically around taking keys by ref rather than value.  Will leave as is, as agreed offline.

crates/astria-sequencer/src/storage/stored_value.rs
Fraser999: Will leave as is, as agreed offline, since further items will be added to the storage module in upcoming PRs.

crates/astria-sequencer/src/storage/stored_value.rs
Fraser999: Yes, currently it's only used by the `TryFrom<StoredValue>` impls in the components' storage modules to provide info in the `bail!` message as to what unexpected storage enum variant was retrieved rather than the expected one.

I tried to keep the output info pretty concise.  A couple of example error logs look like
```
accounts stored value type mismatch: expected fee, found address stored value address prefix abc
accounts stored value type mismatch: expected fee, found assets stored value trace-prefixed denom a/long/path/to/denom
```

We probably don't need a lot of info there.  I could always remove this `Display` impl and all those in the components' storage modules too if you prefer?  Or I'm up for changing this how you see fit - even the `Debug` impl isn't generally too bad:

```
accounts stored value type mismatch: expected fee, found Address(AddressPrefix(AddressPrefix("abc")))
accounts stored value type mismatch: expected fee, found Assets(TracePrefixedDenom(TracePrefixedDenom { trace: [("a", "long"), ("path", "to")], base_denom: "denom" }))
```
If we go that route, we'd probably want to hand-roll `Debug` for some of the bigger structs like block header to avoid this becoming too large.  OTOH - we shouldn't ever see one of the type-mismatch errors in production.

crates/astria-sequencer/src/sequence/storage/values.rs
```diff
@@ -0,0 +1,56 @@
...
+    }
+}
+
+impl From<Fee> for u128 {
+    fn from(fee: Fee) -> Self {
+        fee.0
+    }
+}
+
+impl<'a> From<Fee> for crate::storage::StoredValue<'a> {
```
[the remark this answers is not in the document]
Fraser999: With the visibility changes we discussed offline, meaning these low-level types become private to the relevant component's module, I think it's not harmful to support these `From` impls as is.  I'm not sure I see any benefit to putting in the extra step, and it would be quite a lot more boilerplate.

crates/astria-sequencer/src/sequence/storage/values.rs
```diff
@@ -0,0 +1,56 @@
...
+            Value::Fee(fee) => write!(f, "fee {}", fee.0),
+        }
+    }
+}
+
+#[derive(Debug, BorshSerialize, BorshDeserialize)]
+pub(crate) struct Fee(u128);
+
+impl From<u128> for Fee {
+    fn from(fee: u128) -> Self {
```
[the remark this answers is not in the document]
Fraser999: #1596

crates/astria-sequencer/src/assets/state_ext.rs
Fraser999: The idea is that this PR was updating the state ext APIs ready for the further storage changes.  In the next PR, I'll be introducing the in-mem cache, so we'll need the actual value for insertion there before serialising it if required.

crates/astria-sequencer/src/assets/state_ext.rs
Fraser999: As per previous reply.

crates/astria-sequencer/src/app/storage/values/storage_version.rs
Fraser999: Ticketed at #1599.

crates/astria-sequencer/src/app/storage/values/revision_number.rs
Fraser999: Ticketed at #1599.

crates/astria-core/src/crypto.rs
```diff
@@ -167,8 +167,8 @@ impl VerificationKey {
     ///
     /// The address is the first 20 bytes of the sha256 hash of the verification key.
     #[must_use]
-    pub fn address_bytes(&self) -> [u8; ADDRESS_LEN] {
-        *self.address_bytes.get_or_init(|| {
+    pub fn address_bytes(&self) -> &[u8; ADDRESS_LEN] {
```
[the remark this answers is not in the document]
Fraser999: Fixed in [132bddc](https://github.com/astriaorg/astria/pull/1492/commits/132bddc7e5a700a5289c61dee5188628a929795b) as discussed offline.

crates/astria-core/src/primitive/v1/mod.rs
```diff
@@ -522,8 +497,8 @@ impl<TFormat> Address<TFormat> {
     }
 
     #[must_use]
-    pub fn bytes(self) -> [u8; ADDRESS_LEN] {
-        self.bytes
+    pub fn bytes(&self) -> &[u8; ADDRESS_LEN] {
```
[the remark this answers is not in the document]
Fraser999: Fixed in [132bddc](https://github.com/astriaorg/astria/pull/1492/commits/132bddc7e5a700a5289c61dee5188628a929795b) as discussed offline.

crates/astria-sequencer/src/bridge/init_bridge_account_action.rs
Fraser999: Fixed in [132bddc](https://github.com/astriaorg/astria/pull/1492/commits/132bddc7e5a700a5289c61dee5188628a929795b).

crates/astria-sequencer/src/accounts/mod.rs
Fraser999: Fixed in [132bddc](https://github.com/astriaorg/astria/pull/1492/commits/132bddc7e5a700a5289c61dee5188628a929795b).

crates/astria-core/src/primitive/v1/asset/denom.rs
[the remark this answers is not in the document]
Fraser999: Fixed in [d16b193](https://github.com/astriaorg/astria/pull/1492/commits/d16b193cc17ce74b03add742b6c9fe68a6be0fb7).

crates/astria-core/src/primitive/v1/mod.rs
```diff
@@ -474,31 +474,6 @@ pub struct Address<T = Bech32m> {
     format: PhantomData<T>,
 }
 
-// The serde impls need to be manually implemented for Address because they
-// only work for Address<Bech32m> which cannot be expressed using serde
-// attributes.
-#[cfg(feature = "serde")]
```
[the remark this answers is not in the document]
Fraser999: Reverted in [3c47298](https://github.com/astriaorg/astria/pull/1492/commits/3c47298b2c9dd33d1a47592208a0e8cef2033d3b) and ticketed at #1603.

crates/astria-core/src/primitive/v1/mod.rs
```diff
@@ -664,11 +651,6 @@ where
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, PartialOrd, Ord)]
-#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
```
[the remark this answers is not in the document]
Fraser999: Reverted in [3c47298](https://github.com/astriaorg/astria/pull/1492/commits/3c47298b2c9dd33d1a47592208a0e8cef2033d3b) and ticketed at #1603.

crates/astria-core/src/sequencerblock/v1alpha1/block.rs
```diff
@@ -1377,6 +1447,21 @@ impl Deposit {
     }
 }
 
+impl Display for Deposit {
```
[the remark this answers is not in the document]
Fraser999: Reverted in [3c47298](https://github.com/astriaorg/astria/pull/1492/commits/3c47298b2c9dd33d1a47592208a0e8cef2033d3b) and ticketed at #1603.

crates/astria-merkle/src/audit.rs
```diff
@@ -407,7 +407,6 @@ impl UncheckedProof {
 /// The proof is the concatenation of all sibling hashes required to reconstruct
 /// the Merkle tree from a leaf. This is also called the audit path.
 #[derive(Clone, Debug, PartialEq, Eq)]
-#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
```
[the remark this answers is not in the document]
Fraser999: Reverted in [3c47298](https://github.com/astriaorg/astria/pull/1492/commits/3c47298b2c9dd33d1a47592208a0e8cef2033d3b) and ticketed at #1603.

crates/astria-sequencer/src/transaction/mod.rs
```diff
@@ -185,26 +187,27 @@ impl ActionHandler for SignedTransaction {
             .wrap_err("failed to check balance for total fees and transfers")?;
 
         if state
-            .get_bridge_account_rollup_id(self)
+            .get_bridge_account_rollup_id(&self)
             .await
             .wrap_err("failed to check account rollup id")?
             .is_some()
         {
+            // No need to add context as this method already reports sufficient context on error.
```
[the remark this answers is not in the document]
Fraser999: Done in [fe21bcb4](https://github.com/astriaorg/astria/pull/1492/commits/fe21bcb4bcec36bfcc1a18e3f0fdc6cafe448fb2).

crates/astria-sequencer/src/sequence/component.rs
```diff
@@ -20,11 +20,11 @@ impl Component for SequenceComponent {
 
     #[instrument(name = "SequenceComponent::init_chain", skip_all)]
     async fn init_chain<S: StateWriteExt>(mut state: S, app_state: &Self::AppState) -> Result<()> {
-        state.put_sequence_action_base_fee(app_state.fees().sequence_base_fee);
+        // No need to add context as these `put` methods already report sufficient context on error.
+        state.put_sequence_action_base_fee(app_state.fees().sequence_base_fee)?;
```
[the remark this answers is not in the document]
Fraser999: Done in [fe21bcb4](https://github.com/astriaorg/astria/pull/1492/commits/fe21bcb4bcec36bfcc1a18e3f0fdc6cafe448fb2).

crates/astria-sequencer/src/authority/action.rs
Fraser999: Done in [fe21bcb4](https://github.com/astriaorg/astria/pull/1492/commits/fe21bcb4bcec36bfcc1a18e3f0fdc6cafe448fb2).

crates/astria-sequencer/src/bridge/component.rs
Fraser999: Done in [fe21bcb4](https://github.com/astriaorg/astria/pull/1492/commits/fe21bcb4bcec36bfcc1a18e3f0fdc6cafe448fb2).

crates/astria-sequencer/src/ibc/ics20_withdrawal.rs
Fraser999: It seems reasonably innocuous to me.  I'd prefer to leave as is to avoid copying the bytes needlessly.

crates/astria-sequencer/src/sequence/mod.rs
```diff
@@ -1,6 +1,7 @@
 pub(crate) mod action;
 pub(crate) mod component;
 mod state_ext;
+pub(crate) mod storage;
```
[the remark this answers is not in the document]
Fraser999: I updated the visibility of the types inside this module in [26e3683](https://github.com/astriaorg/astria/pull/1492/commits/26e36839ad833bd65b420e6f080b14b20cb4acee).  I kept the `storage` module `pub(crate)` so that consumers of the `Value` need to do `use crate::sequence::storage::Value` rather than `crate::sequence::Value`, since that's a bit vague.

`Value` is the only type exported at `pub(crate)` visibility from the `storage` module for all components.  Happy to change this if you want (e.g. I could make the `storage` module private, and export the `Value` using an alias).

crates/astria-sequencer/src/ibc/storage/values.rs
```diff
@@ -0,0 +1,139 @@
...
+use borsh::{
+    BorshDeserialize,
+    BorshSerialize,
+};
+use telemetry::display::base64;
+
+use crate::accounts::AddressBytes as DomainAddressBytes;
+
+#[derive(Debug, BorshSerialize, BorshDeserialize)]
+pub(crate) enum Value<'a> {
```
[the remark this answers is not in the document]
Fraser999: As discussed offline, updated in [26e3683](https://github.com/astriaorg/astria/pull/1492/commits/26e36839ad833bd65b420e6f080b14b20cb4acee).

crates/astria-sequencer/src/ibc/storage/values.rs
Fraser999: I added a comment regarding point 1 above in reply to another comment.  For point 2, as discussed offline, updated in [26e3683](https://github.com/astriaorg/astria/pull/1492/commits/26e36839ad833bd65b420e6f080b14b20cb4acee).

crates/astria-sequencer/src/authority/storage/values.rs
Fraser999: Fixed in [26e3683](https://github.com/astriaorg/astria/pull/1492/commits/26e36839ad833bd65b420e6f080b14b20cb4acee).

crates/astria-core/Cargo.toml
```diff
@@ -55,6 +55,10 @@ server = ["dep:tonic"]
 test-utils = []
 base64-serde = ["dep:base64-serde"]
 brotli = ["dep:brotli"]
+# When enabled, this adds constructors for some types that skip the normal constructor validity
+# checks. It supports the case where the inputs are already deemed valid, e.g. having read them from
+# local storage.
+unchecked-constructors = []
```
SuperFluffy, line 61: I like this solution. Really forces you to opt into this.

I think in addition to this flag, we should make a note on all feature-gated methods that we don't provide stability guarantees for them.

crates/astria-core/src/primitive/v1/asset/denom.rs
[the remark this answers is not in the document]
SuperFluffy: Ambivalent on this (I prefer to have single entrypoints to do one thing, `Clone` and `Copy` in this case), but let's see how this plays out. :)

crates/astria-core/src/primitive/v1/asset/denom.rs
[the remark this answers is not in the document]
SuperFluffy: I like the approach. We should also make note that while these constructors are in the public API, we don't consider them stable or supported. Use at your own risk, etc.

crates/astria-sequencer/src/storage/stored_value.rs
SuperFluffy: OK, with that context I think `Debug` is totally ok here. As you note, something really terrible would have happened if we see this, so turning on `Debug` is probably ok (and including it in some error).

Because of that I would even avoid handrolling and just go with the full derived `Debug` - after all, who knows in how far it would be helpful. Just toss the entire kitchensink at the user. :)

crates/astria-sequencer/src/sequence/storage/values.rs
```diff
@@ -0,0 +1,56 @@
...
+
+#[derive(Debug, BorshSerialize, BorshDeserialize)]
+pub(crate) enum Value {
+    Fee(Fee),
+}
+
+impl Display for Value {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        match self {
+            Value::Fee(fee) => write!(f, "fee {}", fee.0),
```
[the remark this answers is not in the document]
SuperFluffy: Resolving this because of the discussion in another thread (summary: go with debug?).

crates/astria-sequencer/src/sequence/storage/values.rs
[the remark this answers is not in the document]
SuperFluffy: Yeah, in this case I am fine with it.

crates/astria-sequencer/src/sequence/mod.rs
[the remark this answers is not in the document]
SuperFluffy: No, this makes sense and I would have probably done the same, thank you for context!

crates/astria-sequencer/src/ibc/storage/values.rs
```diff
@@ -0,0 +1,139 @@
...
+    Balance(Balance),
+    AddressBytes(AddressBytes<'a>),
+    Fee(Fee),
+}
+
+impl<'a> Display for Value<'a> {
```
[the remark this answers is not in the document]
SuperFluffy: Resolving because discussion is happening here: https://github.com/astriaorg/astria/pull/1492#pullrequestreview-2337024905

crates/astria-sequencer/src/ibc/component.rs
```diff
@@ -45,7 +45,9 @@ impl Component for IbcComponent {
             .wrap_err("failed to set IBC sudo key")?;
 
         for address in app_state.ibc_relayer_addresses() {
-            state.put_ibc_relayer_address(address);
+            state
+                .put_ibc_relayer_address(address)
+                .wrap_err("failed to put IBC relayer address")?;
```
SuperFluffy, line 50:
```suggestion
                .wrap_err("failed to write IBC relayer address to state")?;
```

crates/astria-sequencer/src/grpc/storage/values/sequencer_block_header.rs
```diff
@@ -0,0 +1,182 @@
...
+                "grpc stored value type mismatch: expected sequencer block header, found {value}"
+            );
+        };
+        Ok(block_header)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use std::str::FromStr;
```
SuperFluffy, line 162: Just `parse` in the test instead of importing?

crates/astria-sequencer/src/grpc/storage/values/rollup_ids.rs
```diff
@@ -0,0 +1,82 @@
...
+        DomainRollupId::new(rollup_id.0.into_owned())
+    }
+}
+
+#[derive(Debug, BorshSerialize, BorshDeserialize)]
+pub(in crate::grpc) struct RollupIds<'a>(Vec<RollupId<'a>>);
+
+impl<'a> Display for RollupIds<'a> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        write!(f, "[{}]", self.0.iter().join(", "))
```
SuperFluffy, line 50: Can we avoid the allocation here and just:
```suggestion
        f.write_str("[")?;
        let mut ids = self.0.iter();
        if let Some(id) = ids.next();
            id.fmt(f)?;
        }
        for id in ids {
            f.write_str(", ")?;
            id.fmt(f)?;
        }
        f.write_str("]")?;
        Ok(())
    }
```
    
            
        write!(f, "[{}]", self.0.iter().join(", "))

crates/astria-sequencer/src/grpc/state_ext.rs
```diff
@@ -272,124 +107,230 @@ pub(crate) trait StateReadExt: StateRead {
...
+        header,
+        rollup_transactions: Default::default(),
+        rollup_transactions_proof,
+        rollup_ids_proof,
+    };
 
-        let key = rollup_ids_proof_by_hash_key(&block_hash);
-        self.nonverifiable_put_raw(key, rollup_ids_proof.into_raw().encode_to_vec());
+    for rollup_id in rollup_ids {
+        let rollup_txs = state.get_rollup_data(hash, &rollup_id).await?;
```
SuperFluffy, line 237: 1. please provide error context
2. this executes async futures in a loop. We should replace this by a stream. Am fine doing this in a followup, but we should track it.

crates/astria-sequencer/src/grpc/state_ext.rs
```diff
@@ -272,124 +107,230 @@ pub(crate) trait StateReadExt: StateRead {
...
+        .await
+        .wrap_err("failed to get rollup transactions proof by block hash")?;
+    let rollup_ids_proof = state
+        .get_rollup_ids_proof_by_block_hash(hash)
+        .await
+        .wrap_err("failed to get rollup ids proof by block hash")?;
+
+    #[expect(
+        clippy::default_trait_access,
+        reason = "want to avoid explicitly importing `index_map` crate to sequencer crate"
```
SuperFluffy, line 226: Builder to the rescue. :)

crates/astria-sequencer/src/grpc/state_ext.rs
SuperFluffy, line 1: Comment on this file: all `put_*` methods that take `state: &mut S` could probably get away with taking just `mut state: S`.

SuperFluffy (commented):
Comments:

1. the errors for failing to serialize from/to the on-disk type should use `Debug` (very rare, very bad if it happens: it's fine to be extra noisy).
2. on-disk types with a `Display` impl should should have a unit test so that their display impl matches that of the domain type (to avoid confusion) non-exhaustive list of types where this applies: `RollupId`, `BlockHash`.
3. we should provide an `OndiskType` trait that encapsulates the `From`/`TryFrom` impls we have everywhere right now.
4. we really really should split the tests into separate modules. This is not the first PR where I wish I could just ignore the tests when going through the code. :-/

crates/astria-sequencer/src/grpc/sequencer.rs
```diff
@@ -117,15 +118,22 @@ impl SequencerService for SequencerServer {
...
-                    "failed to get sequencer block proofs from storage: {e}"
+                    "failed to get rollup transactions proof from storage: {e}"
                 ))
             })?;
 
+        let rollup_ids_proof = snapshot
+            .get_rollup_ids_proof_by_block_hash(&block_hash)
+            .await
+            .map_err(|e| {
+                Status::internal(format!("failed to get rollup ids proof from storage: {e}"))
```
SuperFluffy, line 134: This will likely underreport the error. Ideally we log this, but let's change this to `{e:#}` for the time being?

crates/astria-sequencer/src/grpc/sequencer.rs
```diff
@@ -117,15 +118,22 @@ impl SequencerService for SequencerServer {
...
 
-        let (rollup_transactions_proof, rollup_ids_proof) = snapshot
-            .get_block_proofs_by_block_hash(&block_hash)
+        let rollup_transactions_proof = snapshot
+            .get_rollup_transactions_proof_by_block_hash(&block_hash)
             .await
             .map_err(|e| {
                 Status::internal(format!(
-                    "failed to get sequencer block proofs from storage: {e}"
+                    "failed to get rollup transactions proof from storage: {e}"
```
SuperFluffy, line 126: Same as below: it's technically out of scope for this PR, but I haven't noticed this earlier. Let's report the error as `{e:#}`?

crates/astria-sequencer/src/bridge/storage/values/transaction_id.rs
```diff
@@ -0,0 +1,63 @@
...
+    Value,
+    ValueImpl,
+};
+
+#[derive(Debug, BorshSerialize, BorshDeserialize)]
+pub(in crate::bridge) struct TransactionId<'a>(Cow<'a, [u8; TRANSACTION_ID_LEN]>);
+
+impl<'a> Display for TransactionId<'a> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        hex(self.0.as_slice()).fmt(f)
```
SuperFluffy, line 31: We should assert that this matches the formatting of `TransactionId`.

crates/astria-sequencer/src/bridge/storage/values/rollup_id.rs
```diff
@@ -0,0 +1,60 @@
...
+    Value,
+    ValueImpl,
+};
+
+#[derive(Debug, BorshSerialize, BorshDeserialize)]
+pub(in crate::bridge) struct RollupId<'a>(Cow<'a, [u8; 32]>);
+
+impl<'a> Display for RollupId<'a> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        base64(self.0.as_slice()).fmt(f)
```
SuperFluffy, line 28: Should assert this matches the formatting of the domain type `RollupId`

crates/astria-sequencer/src/bridge/storage/values/address_bytes.rs
```diff
@@ -0,0 +1,61 @@
...
+    ValueImpl,
+};
+use crate::accounts::AddressBytes as DomainAddressBytes;
+
+#[derive(Debug, BorshSerialize, BorshDeserialize)]
+pub(in crate::bridge) struct AddressBytes<'a>(Cow<'a, [u8; ADDRESS_LEN]>);
+
+impl<'a> Display for AddressBytes<'a> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        base64(self.0.as_slice()).fmt(f)
```
SuperFluffy, line 29: Should assert this matches the display format of the domain type.

crates/astria-sequencer/src/app/storage/values/storage_version.rs
```diff
@@ -0,0 +1,56 @@
...
+    BorshSerialize,
+};
+
+use super::{
+    Value,
+    ValueImpl,
+};
+
+#[derive(Debug, BorshSerialize, BorshDeserialize)]
+pub(in crate::app) struct StorageVersion(u64);
```
SuperFluffy, line 19: We should have a domain type for this one.

crates/astria-sequencer/src/app/storage/values/revision_number.rs
```diff
@@ -0,0 +1,56 @@
...
+    BorshSerialize,
+};
+
+use super::{
+    Value,
+    ValueImpl,
+};
+
+#[derive(Debug, BorshSerialize, BorshDeserialize)]
+pub(in crate::app) struct RevisionNumber(u64);
```
SuperFluffy, line 19: Also should have a domain type for this one.

crates/astria-sequencer/src/app/storage/values/block_timestamp.rs
```diff
@@ -0,0 +1,94 @@
...
+    BorshSerialize,
+};
+
+use super::{
+    Value,
+    ValueImpl,
+};
+
+#[derive(Debug)]
+pub(in crate::app) struct BlockTimestamp(tendermint::time::Time);
```
SuperFluffy, line 23: General note: I wish we didn't use the tendermint time but instead our own type.

crates/astria-sequencer/src/app/storage/values/block_height.rs
```diff
@@ -0,0 +1,55 @@
...
+    BorshSerialize,
+};
+
+use super::{
+    Value,
+    ValueImpl,
+};
+
+#[derive(Debug, BorshSerialize, BorshDeserialize)]
+pub(in crate::app) struct BlockHeight(u64);
```
SuperFluffy, line 19: Should probably have a domain type for this.

crates/astria-sequencer/src/accounts/storage/values.rs
```diff
@@ -0,0 +1,128 @@
...
+            ValueImpl::Nonce(nonce) => write!(f, "nonce {}", nonce.0),
+            ValueImpl::Fee(fee) => write!(f, "fee {}", fee.0),
+        }
+    }
+}
+
+#[derive(Debug, BorshSerialize, BorshDeserialize)]
+pub(in crate::accounts) struct Balance(u128);
+
+impl From<u128> for Balance {
```
SuperFluffy, line 36: This probably warrants a domain type.

crates/astria-sequencer/src/accounts/storage/values.rs
```diff
@@ -0,0 +1,128 @@
...
+            bail!("accounts stored value type mismatch: expected balance, found {value}");
+        };
+        Ok(balance)
+    }
+}
+
+#[derive(Debug, BorshSerialize, BorshDeserialize)]
+pub(in crate::accounts) struct Nonce(u32);
+
+impl From<u32> for Nonce {
```
SuperFluffy, line 69: And this should also have a domain type.

crates/astria-sequencer/src/accounts/state_ext.rs
```diff
@@ -245,46 +235,51 @@ impl<T: StateRead + ?Sized> StateReadExt for T {}
 #[async_trait]
 pub(crate) trait StateWriteExt: StateWrite {
     #[instrument(skip_all, fields(address = %address.display_address(), %asset, balance), err)]
-    fn put_account_balance<TAddress, TAsset>(
+    fn put_account_balance<'a, TAddress, TAsset>(
         &mut self,
-        address: TAddress,
-        asset: TAsset,
+        address: &TAddress,
+        asset: &'a TAsset,
```
SuperFluffy, line 241: Gonna ask this here because I have noticed this throughout the PR: what required changing this from `TAsset` and `TAddress` to take their borrows, respectively? If I am not mistaken these methods allowed being called with `&asset` and `&address` already.

crates/astria-sequencer/Cargo.toml
```diff
@@ -45,6 +50,7 @@ divan = { workspace = true, optional = true }
 futures = { workspace = true }
 hex = { workspace = true, features = ["serde"] }
 ibc-types = { workspace = true, features = ["with_serde"] }
+itertools = { workspace = true }
```
SuperFluffy, line 53: The instance where this was used allocates. I'd like to not allocate. :)

crates/astria-merkle/src/audit.rs
```diff
@@ -621,4 +621,21 @@ impl Proof {
     pub fn verify(&self, leaf: &[u8], root_hash: [u8; 32]) -> bool {
         self.audit().with_leaf(leaf).with_root(root_hash).perform()
     }
+
+    /// This should only be used where `parts` has been provided by a trusted entity, e.g. read from
+    /// our own state store.
+    #[doc(hidden)]
+    #[must_use]
+    pub fn unchecked_from_parts(parts: UncheckedProof) -> Self {
```
SuperFluffy, line 629: Can you feature gate this one as well?

SuperFluffy (approved):
Thank you for the heroic task of going through all of sequencer and fixing its state write/read. I fat fingered my previous review but the comments were intended for this approval.

I don't see any blockers. From the previous list my main ask is to get remove all custom display impls in favor of just the derived `Debug` - we discussed this in the relevant comment chain, but as you noted the `Display` impls are intended for providing errors on failed state-decoding.

If that happens we are in so deep waters, that I think spewing a Debug log all over the screen is perfectly acceptable.

(side note that didn't make it into the list of the other comment:

1. We should extend snapshot tests to all keys - but that's likely covered in your open followup PR.

)

crates/astria-sequencer/src/storage/stored_value.rs
Fraser999: Done in [44170ce](https://github.com/astriaorg/astria/pull/1492/commits/44170ceefc479c5f7839a388b322ac2412962df2).

crates/astria-sequencer/Cargo.toml
Fraser999: Removed in [44170ce](https://github.com/astriaorg/astria/pull/1492/commits/44170ceefc479c5f7839a388b322ac2412962df2).

crates/astria-sequencer/src/grpc/sequencer.rs
Fraser999: I think there's no difference in these two forms now we've replaced `anyhow` with our own `eyre` wrapper.

joroshiba (approved):

crates/astria-sequencer/src/grpc/sequencer.rs
Fraser999: As per reply above.

crates/astria-sequencer/src/bridge/storage/values/transaction_id.rs
Fraser999: As discussed offline, we can skip that since we're using Debug formatting.

crates/astria-sequencer/src/bridge/storage/values/rollup_id.rs
Fraser999: As discussed offline, we can skip that since we're using Debug formatting.

crates/astria-sequencer/src/bridge/storage/values/address_bytes.rs
Fraser999: As discussed offline, we can skip that since we're using Debug formatting.

crates/astria-sequencer/src/app/storage/values/revision_number.rs
Fraser999: Ticketed at #1599.

crates/astria-sequencer/src/app/storage/values/block_height.rs
Fraser999: Ticketed at #1604.

crates/astria-sequencer/src/accounts/storage/values.rs
Fraser999: Ticketed at #1596.

crates/astria-sequencer/src/accounts/storage/values.rs
Fraser999: Ticketed at #1605.

crates/astria-sequencer/src/accounts/state_ext.rs
Fraser999: They did, but they also allowed for passing by value.  I wanted to disallow the ability to pass by value in these cases, since it would be easy to accidentally copy the data.

crates/astria-core/Cargo.toml
Fraser999: I tend to assume `doc(hidden)` implies this, but I guess better to be explicit.

Updated in [ccc3f9e](https://github.com/astriaorg/astria/pull/1492/commits/ccc3f9e62661d8c6f0d73a52a119ed18aa89cc60).

crates/astria-merkle/src/audit.rs
Fraser999: Done in [ccc3f9e](https://github.com/astriaorg/astria/pull/1492/commits/ccc3f9e62661d8c6f0d73a52a119ed18aa89cc60).

crates/astria-sequencer/src/grpc/state_ext.rs
Fraser999: A good constructor would be even better! One which took an iterator over `RollupTransactions` so we don't need to care how it's held internally :)

crates/astria-sequencer/src/grpc/state_ext.rs
Fraser999: I'll leave as is for now, since by the end of the storage epic we should be able to just take `state: &S` instead.

crates/astria-sequencer/src/grpc/storage/values/rollup_ids.rs
Fraser999: This code is now deleted (using Debug instead).

crates/astria-sequencer/src/sequence/component.rs
```diff
@@ -1,7 +1,10 @@
 use std::sync::Arc;
 
 use astria_core::protocol::genesis::v1alpha1::GenesisAppState;
-use astria_eyre::eyre::Result;
+use astria_eyre::eyre::{
+    Result,
+    WrapErr,
```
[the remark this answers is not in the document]
Fraser999: Done in [dd3bfa6](https://github.com/astriaorg/astria/pull/1492/commits/dd3bfa68494d3d14555a09c0dfb5cbe6228b6f7d).

crates/astria-sequencer/src/ibc/component.rs
Fraser999: Done in [dd3bfa6](https://github.com/astriaorg/astria/pull/1492/commits/dd3bfa68494d3d14555a09c0dfb5cbe6228b6f7d).

crates/astria-sequencer/src/grpc/storage/values/sequencer_block_header.rs
Fraser999: Done in [dd3bfa6](https://github.com/astriaorg/astria/pull/1492/commits/dd3bfa68494d3d14555a09c0dfb5cbe6228b6f7d).

crates/astria-sequencer/src/grpc/state_ext.rs
Fraser999: Point 1 done in [dd3bfa6](https://github.com/astriaorg/astria/pull/1492/commits/dd3bfa68494d3d14555a09c0dfb5cbe6228b6f7d).

Point 2 tracked at #1606.

crates/astria-sequencer/src/accounts/state_ext.rs
SuperFluffy: OK, cool. I figured that would be it.

crates/astria-sequencer/src/grpc/sequencer.rs
SuperFluffy: Yeah, you are right.

Issue #1439, opened 2024-09-03
Failed ICS20 withdrawals use bridge address instead of rollup address


SuperFluffy:
Refunds of failed ICS20 withdrawals use the bridge address instead of destination chain address.

The fix is to use `Ics20WithdrawalFromRollup.rollup_return_address` from here:
https://github.com/astriaorg/astria/blob/c3489732628454814d6cee59d458af430560d5a3/proto/protocolapis/astria/protocol/memos/v1alpha1/types.proto#L42

Below shows the problematic lines of code.

1. `execute_rollup_withdrawal_refund` constructs a `Deposit` using a `destination_address` here: https://github.com/astriaorg/astria/blob/c3489732628454814d6cee59d458af430560d5a3/crates/astria-sequencer/src/ibc/ics20_transfer.rs#L529
2. the packet sender is then parsed as the bridge address AND the destination address:
    1. here the `recipient` var is set to `packet.sender`: https://github.com/astriaorg/astria/blob/c3489732628454814d6cee59d458af430560d5a3/crates/astria-sequencer/src/ibc/ics20_transfer.rs#L383-L384
    2. here `packet.sender` is parsed as a bridge account: https://github.com/astriaorg/astria/blob/c3489732628454814d6cee59d458af430560d5a3/crates/astria-sequencer/src/ibc/ics20_transfer.rs#L411
3. here we are passing in `recipient` as the `destination_address` and `bridge_address` as the `bridge_address`: https://github.com/astriaorg/astria/blob/c3489732628454814d6cee59d458af430560d5a3/crates/astria-sequencer/src/ibc/ics20_transfer.rs#L415-L420
4. but they are the same value!

The comment on `Deposit.destination_chain_address` reads:
https://github.com/astriaorg/astria/blob/c3489732628454814d6cee59d458af430560d5a3/proto/sequencerblockapis/astria/sequencerblock/v1alpha1/block.proto#L86-L88

┆Issue Number: ENG-771

PR #1495, opened 2024-09-12
fix, refactor(sequencer): refactor ics20 logic

SuperFluffy:
## Summary
Fixes incorrect addresses in ics20 refunds, and fixes/adds source vs sink zone detection of received/refunded asets.

## Background
This PR started with the goal of using `Ics20WithdrawalFromRollup.rollup_return_address` when emitting ics20 deposit events instead of the rollup's bridge address (which for the rollup was quite meaningless).

However, because the original logic was overly convoluted, this patch evolved into a refactor which subsequently revealed 2 more bugs.

## Changes
- Refactor ics20 logic: split up `execute_ics20_transfer` into two functions `receive_tokens` and `refund_tokens`
- Fix deposit events being emitted with the packet sender/bridge address as the receiving address on the rollup: instead use `Ics20WithdrawalFromRollup.rollup_return_address` which is the address originally supplied and understood by the rollup.
- Fix bridge lock events of source tokens being emitted with an extra (port, channel) pair added (this resulting token is likely completely meaningless and not understood by the rollup): instead strip the leading (port, channel) pair to get the original token on sequencer.
- Fix refunds to a rollup of failed ics20 transfers not checking if sequencer is their source or sink zone: instead perform the check and decrease the ibc escrow account is necessary.

## Testing

Renamed and expanded tests to also check for sink vs source zone assets being received, the correct asset being emitted in bridge lock events, and the correct rollup return address being emitted in refunds. Specifically these tests were added or renamed:

- `receive_source_zone_asset_on_sequencer_account`
- `receive_sink_zone_asset_on_sequencer_account`
- `receive_source_zone_asset_on_bridge_account_and_emit_to_rollup`
- `receive_sink_zone_asset_on_bridge_account_and_emit_to_rollup`
- `transfer_to_bridge_is_rejected_due_to_invalid_memo`
- `transfer_to_bridge_account_is_rejected_due_to_not_permitted_token`
- `refund_sequencer_account_with_source_zone_asset`
- `refund_sequencer_account_with_sink_zone_asset`
- `refund_rollup_with_sink_zone_asset`
- `refund_rollup_with_source_zone_asset`
- `refund_rollup_with_source_zone_asset_compat_prefix`

## Related Issues
Closes https://github.com/astriaorg/astria/issues/1439
Closes https://github.com/astriaorg/astria/issues/1496
Closes https://github.com/astriaorg/astria/issues/1514

### DIFF
--- a/crates/astria-core/src/primitive/v1/asset/denom.rs
+++ b/crates/astria-core/src/primitive/v1/asset/denom.rs
@@ -207,82 +207,42 @@ impl TracePrefixed {
         self.trace.is_empty()
     }
 
-    /// Checks if the trace prefixed denom starts with `s`.
+    /// Checks if the trace prefixed denom has `port` in left-most position.
     ///
     /// # Examples
     ///
     /// ```
     /// use astria_core::primitive::v1::asset::denom::TracePrefixed;
     /// let denom = "four/segments/of/a/denom".parse::<TracePrefixed>().unwrap();
-    ///
-    /// // Empty string is always true:
-    /// assert!(denom.starts_with_str(""));
-    /// // Single slash is always false:
-    /// assert!(!denom.starts_with_str("/"));
-    /// // Emptry strings are false:
-    /// assert!(!denom.starts_with_str(" "));
-    ///
-    /// // In general, whitespace is not trimmed and leads to false
-    /// assert!(!denom.starts_with_str("four/segments /"));
-    ///
-    /// // Trailing slashes don't change the result if they are part of the trace prefix:
-    /// assert!(denom.starts_with_str("four/segments"));
-    /// assert!(denom.starts_with_str("four/segments/"));
-    ///
-    /// // Trailing slashes on the full trace prefix denom however return false:
-    /// assert!(!denom.starts_with_str("four/segments/of/a/denom/"));
-    ///
-    /// // Providing only a port is true
-    /// assert!(denom.starts_with_str("four"));
-    /// // Providing a full port/channel pair followed by just a port is also true
-    /// assert!(denom.starts_with_str("four/segments/of"));
-    ///
-    /// // Half of a port or channel is false
-    /// assert!(!denom.starts_with_str("four/segm"));
-    ///
-    /// // The full trace prefixed denom is true:
-    /// assert!(denom.starts_with_str("four/segments/of/a/denom"));
+    /// assert!(denom.has_leading_port("four"));
+    /// assert!(!denom.has_leading_port("segments"));
+    /// assert!(!denom.has_leading_port("of"));
+    /// assert!(!denom.has_leading_port("a"));
+    /// assert!(!denom.has_leading_port("denom"));
+    /// assert!(!denom.has_leading_port(""));
     /// ```
     #[must_use]
-    pub fn starts_with_str(&self, s: &str) -> bool {
-        if s.is_empty() {
-            return true;
-        }
-        let mut had_trailing_slash = false;
-        let s = s
-            .strip_suffix('/')
-            .inspect(|_| had_trailing_slash = true)
-            .unwrap_or(s);
-        if s.is_empty() {
-            return false;
-        }
-        let mut parts = s.split('/');
-        for segment in self.trace.iter() {
-            // first iteration: we know that s is not empty after stripping the /
-            // so that this is not wrongly returning true.
-            let Some(port) = parts.next() else {
-                return true;
-            };
-            if segment.port() != port {
-                return false;
-            }
-            let Some(channel) = parts.next() else {
-                return true;
-            };
-            if segment.channel() != channel {
-                return false;
-            }
-        }
-        let Some(base_denom) = parts.next() else {
-            return true;
-        };
-        if base_denom != self.base_denom {
-            return false;
-        }
-        if had_trailing_slash {
-            return false;
-        }
-        parts.next().is_none()
+    pub fn has_leading_port<T: AsRef<str>>(&self, port: T) -> bool {
+        self.trace.leading_port() == Some(port.as_ref())
+    }
+
+    /// Checks if the trace prefixed denom has `channel` in left-most position.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use astria_core::primitive::v1::asset::denom::TracePrefixed;
+    /// let denom = "four/segments/of/a/denom".parse::<TracePrefixed>().unwrap();
+    /// assert!(!denom.has_leading_channel("four"));
+    /// assert!(denom.has_leading_channel("segments"));
+    /// assert!(!denom.has_leading_channel("of"));
+    /// assert!(!denom.has_leading_channel("a"));
+    /// assert!(!denom.has_leading_channel("denom"));
+    /// assert!(!denom.has_leading_channel(""));
+    /// ```
+    #[must_use]
+    pub fn has_leading_channel<T: AsRef<str>>(&self, channel: T) -> bool {
+        self.trace.leading_channel() == Some(channel.as_ref())
     }
 
     #[must_use]
@@ -290,7 +250,7 @@ impl TracePrefixed {
         self.trace.last_channel()
     }
 
-    pub fn pop_trace_segment(&mut self) -> Option<PortAndChannel> {
+    pub fn pop_leading_port_and_channel(&mut self) -> Option<PortAndChannel> {
         self.trace.pop()
     }
 
@@ -311,6 +271,14 @@ impl TraceSegments {
         }
     }
 
+    fn leading_port(&self) -> Option<&str> {
+        self.inner.front().map(|segment| &*segment.port)
+    }
+
+    fn leading_channel(&self) -> Option<&str> {
+        self.inner.front().map(|segment| &*segment.channel)
+    }
+
     fn push(&mut self, seg: PortAndChannel) {
         self.inner.push_back(seg);
     }
@@ -325,10 +293,6 @@ impl TraceSegments {
 
     fn is_empty(&self) -> bool {
         self.inner.is_empty()
-    }
-
-    fn iter(&self) -> impl Iterator<Item = &PortAndChannel> {
-        self.inner.iter()
     }
 }
 
@@ -717,37 +681,14 @@ mod tests {
     #[test]
     fn pop_path() {
         let mut denom = "a/long/path/to/denom".parse::<TracePrefixed>().unwrap();
-        let port_and_channel = denom.pop_trace_segment().unwrap();
+        let port_and_channel = denom.pop_leading_port_and_channel().unwrap();
         assert_eq!("a", port_and_channel.port());
         assert_eq!("long", port_and_channel.channel());
 
-        let port_and_channel = denom.pop_trace_segment().unwrap();
+        let port_and_channel = denom.pop_leading_port_and_channel().unwrap();
         assert_eq!("path", port_and_channel.port());
         assert_eq!("to", port_and_channel.channel());
 
-        assert_eq!(None, denom.pop_trace_segment());
-    }
-
-    #[test]
-    fn start_prefixes() {
-        let denom = "four/segments/of/a/denom".parse::<TracePrefixed>().unwrap();
-
-        assert!(denom.starts_with_str(""));
-        assert!(!denom.starts_with_str("/"));
-        assert!(!denom.starts_with_str(" "));
-
-        assert!(!denom.starts_with_str("four/segments /"));
-
-        assert!(denom.starts_with_str("four/segments"));
-        assert!(denom.starts_with_str("four/segments/"));
-
-        assert!(!denom.starts_with_str("four/segments/of/a/denom/"));
-
-        assert!(denom.starts_with_str("four"));
-        assert!(denom.starts_with_str("four/segments/of"));
-
-        assert!(!denom.starts_with_str("four/segm"));
-
-        assert!(denom.starts_with_str("four/segments/of/a/denom"));
+        assert_eq!(None, denom.pop_leading_port_and_channel());
     }
 }
--- a/crates/astria-core/src/sequencerblock/v1alpha1/block.rs
+++ b/crates/astria-core/src/sequencerblock/v1alpha1/block.rs
@@ -1291,85 +1291,23 @@ impl FilteredSequencerBlockError {
 )]
 pub struct Deposit {
     // the address on the sequencer to which the funds were sent to.
-    bridge_address: Address,
+    pub bridge_address: Address,
     // the rollup ID registered to the `bridge_address`
-    rollup_id: RollupId,
+    pub rollup_id: RollupId,
     // the amount that was transferred to `bridge_address`
-    amount: u128,
+    pub amount: u128,
     // the IBC ICS20 denom of the asset that was transferred
-    asset: asset::Denom,
+    pub asset: asset::Denom,
     // the address on the destination chain (rollup) which to send the bridged funds to
-    destination_chain_address: String,
+    pub destination_chain_address: String,
     // the transaction ID of the source action for the deposit, consisting
     // of the transaction hash.
-    source_transaction_id: TransactionId,
+    pub source_transaction_id: TransactionId,
     // index of the deposit's source action within its transaction
-    source_action_index: u64,
-}
-
-impl From<Deposit> for crate::generated::sequencerblock::v1alpha1::Deposit {
-    fn from(deposit: Deposit) -> Self {
-        deposit.into_raw()
-    }
+    pub source_action_index: u64,
 }
 
 impl Deposit {
-    #[must_use]
-    pub fn new(
-        bridge_address: Address,
-        rollup_id: RollupId,
-        amount: u128,
-        asset: asset::Denom,
-        destination_chain_address: String,
-        source_transaction_id: TransactionId,
-        source_action_index: u64,
-    ) -> Self {
-        Self {
-            bridge_address,
-            rollup_id,
-            amount,
-            asset,
-            destination_chain_address,
-            source_transaction_id,
-            source_action_index,
-        }
-    }
-
-    #[must_use]
-    pub fn bridge_address(&self) -> &Address {
-        &self.bridge_address
-    }
-
-    #[must_use]
-    pub fn rollup_id(&self) -> &RollupId {
-        &self.rollup_id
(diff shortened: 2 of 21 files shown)


crates/astria-core/src/primitive/v1/asset/denom.rs
```diff
@@ -207,90 +207,50 @@ impl TracePrefixed {
...
     /// let denom = "four/segments/of/a/denom".parse::<TracePrefixed>().unwrap();
+    /// assert!(denom.has_leading_port("four"));
+    /// assert!(!denom.has_leading_port("segments"));
+    /// assert!(!denom.has_leading_port("of"));
+    /// assert!(!denom.has_leading_port("a"));
+    /// assert!(!denom.has_leading_port("denom"));
+    /// assert!(!denom.has_leading_port(""));
+    /// ```
+    #[must_use]
+    pub fn has_leading_port<T: AsRef<str>>(&self, port: T) -> bool {
```
SuperFluffy, line 225: Added `has_leading_port` and `has_leading_channel` to have a stricter prefix check instead of `starts_with_str`.

crates/astria-core/src/primitive/v1/asset/denom.rs
```diff
@@ -207,90 +207,50 @@ impl TracePrefixed {
...
-            return true;
-        };
-        if base_denom != self.base_denom {
-            return false;
-        }
-        if had_trailing_slash {
-            return false;
-        }
-        parts.next().is_none()
+    pub fn has_leading_channel<T: AsRef<str>>(&self, channel: T) -> bool {
```
SuperFluffy, line 244: Added `has_leading_port` and `has_leading_channel` to have a stricter prefix check instead of `starts_with_str`.

crates/astria-core/src/sequencerblock/v1alpha1/block.rs
```diff
@@ -1291,86 +1291,23 @@ impl FilteredSequencerBlockError {
 )]
 pub struct Deposit {
```
SuperFluffy, line 1292: This is just a stupid container without any invariants.

Made all fields public and removed the getters.

crates/astria-core/src/sequencerblock/v1alpha1/block.rs
```diff
@@ -1291,86 +1291,23 @@ impl FilteredSequencerBlockError {
...
-impl From<Deposit> for crate::generated::sequencerblock::v1alpha1::Deposit {
-    fn from(deposit: Deposit) -> Self {
-        deposit.into_raw()
-    }
+    pub source_action_index: u64,
 }
 
 impl Deposit {
-    #[must_use]
-    pub fn new(
```
SuperFluffy, line 1318: Terrible evergrowing constructor. Since there are no invariants to be upheld I removed it in favor of directly constructing `Deposit` from public fields.

crates/astria-core/src/sequencerblock/v1alpha1/block.rs
```diff
@@ -1439,6 +1376,12 @@ impl Deposit {
     }
 }
 
+impl From<Deposit> for crate::generated::sequencerblock::v1alpha1::Deposit {
```
SuperFluffy, line 1379: Not new, just moved down from above; this is an artifact of the rafctor, but the trait impl fits better down here.

crates/astria-sequencer/src/api_state_ext.rs
SuperFluffy, line 1: contains only changes due to `Deposit::new` being replaced by direction declaration `Deposit {}`.

crates/astria-sequencer/src/app/tests_block_fees.rs
SuperFluffy, line 1: contains only changes due to `Deposit::new` being replaced by direction declaration `Deposit {}`.

crates/astria-sequencer/src/app/tests_breaking_changes.rs
SuperFluffy, line 1: contains only changes due to `Deposit::new` being replaced by direction declaration `Deposit {}`.

crates/astria-sequencer/src/app/tests_execute_transaction.rs
SuperFluffy, line 1: contains only changes due to `Deposit::new` being replaced by direction declaration `Deposit {}`.

crates/astria-sequencer/src/bridge/bridge_lock_action.rs
SuperFluffy, line 1: contains only changes due to `Deposit::new` being replaced by direction declaration `Deposit {}`.

crates/astria-sequencer/src/bridge/state_ext.rs
SuperFluffy, line 1: contains only changes due to `Deposit::new` being replaced by direction declaration `Deposit {}`.

crates/astria-sequencer/src/ibc/state_ext.rs
```diff
@@ -127,10 +127,36 @@ pub(crate) trait StateReadExt: StateRead {
     }
 }
 
-impl<T: StateRead> StateReadExt for T {}
+impl<T: StateRead + ?Sized> StateReadExt for T {}
 
 #[async_trait]
 pub(crate) trait StateWriteExt: StateWrite {
+    #[instrument(skip_all, fields(%channel))]
+    async fn decrease_ibc_channel_balance<TAsset>(
```
SuperFluffy, line 135: New utility to atomically read-update-write the balance on the escrow account.

crates/astria-sequencer/src/transaction/checks.rs
SuperFluffy, line 1: contains only changes due to `Deposit::new` being replaced by direction declaration `Deposit {}`.

crates/astria-sequencer/src/utils.rs
SuperFluffy, line 1: contains only changes due to deposit losing its getters

crates/astria-sequencer/src/ibc/ics20_transfer.rs
```diff
@@ -288,290 +353,280 @@ impl AppHandlerExecute for Ics20Transfer {
...
-        let ack: TokenTransferAcknowledgement = serde_json::from_slice(
-            msg.acknowledgement.as_slice(),
-        )
-        .expect("valid acknowledgement, should have been checked in acknowledge_packet_check");
-        if ack.is_successful() {
-            return Ok(());
+        let ack: TokenTransferAcknowledgement =
+            serde_json::from_slice(msg.acknowledgement.as_slice())
+                .context("failed to deserialize token transfer acknowledgement")?;
+        if !ack.is_successful() {
```
SuperFluffy, line 369: This just flips the logic: instead of returning early on success, success is now the fall-through case while failure triggers a refund.

crates/astria-sequencer/src/ibc/ics20_transfer.rs
SuperFluffy, line 1: This file should probably not be reviewed using its diff, but instead read from source.

As a high level overview:

`execute_ics20_transfer` (that branched on an argument `is_refund`) is split up into `receive_tokens` and `refund_tokens`.

Both functions were cleaned up and should be much easier to read.

crates/astria-sequencer/src/ibc/ics20_transfer.rs
```diff
@@ -288,290 +353,280 @@ impl AppHandlerExecute for Ics20Transfer {
...
+        );
+    }
 
-    // prefix the denomination with the destination port and channel if not a refund
-    let trace_with_dest =
-        prepend_denom_if_not_refund(&denom_trace, dest_port, dest_channel, is_refund);
-
-    // check if this is a transfer to a bridge account and
-    // execute relevant state changes if it is
-    execute_ics20_transfer_bridge_lock(
```
SuperFluffy, line 445: I believe this function was incorrect: its argument `trace_with_dest = dest_port/dest_channel/source_port/source_channel/asset` was always constructed irrespective of whether sequencer was a source or a sink zone, but the construction is only correct if sequencer is a sink!

If sequencer is a source, then `dest_port/dest_channel` must *not* be prepended, and instead `source_port/soure_channel` removed, leaving just `asset`.

crates/astria-sequencer-utils/src/blob_parser.rs
SuperFluffy, line 1: contains only changes due to `Deposit::new` being replaced by direction declaration `Deposit {}`.

crates/astria-sequencer/src/ibc/ics20_transfer.rs
```diff
@@ -288,290 +353,268 @@ impl AppHandlerExecute for Ics20Transfer {
...
+                packet_data.sender
+            )
+        })?;
+
+    let asset = parse_asset(&state, &packet_data.denom)
+        .await
+        .with_context(|| format!("failed parsing packet.asset `{}`", packet_data.denom))?;
+
+    if let Ok(memo) = serde_json::from_str::<Ics20WithdrawalFromRollup>(&packet_data.memo) {
+        info!(
```
noot, line 512: info seems high for this, maybe debug instead?

crates/astria-sequencer/src/ibc/ics20_transfer.rs
```diff
@@ -288,290 +353,268 @@ impl AppHandlerExecute for Ics20Transfer {
...
+                memo.rollup_return_address,
+                memo_name = astria_core::generated::protocol::memos::v1alpha1::Ics20WithdrawalFromRollup::full_name(),
+                "was able to parse memo; refunding to rollup",
+            );
+        refund_tokens_to_rollup(&mut state, receiver, asset, amount, memo)
+            .await
+            .context("failed to refund the rollup")?;
+    } else {
+        info!(
```
noot, line 521: same here

crates/astria-sequencer/src/ibc/ics20_transfer.rs
```diff
@@ -288,290 +353,268 @@ impl AppHandlerExecute for Ics20Transfer {
...
+            recipient = packet_data.sender,
+            "memo in fungible token packet was unknown; taking packet sender as recipient address \
+             on sequencer",
+        );
+        let recipient = packet_data.sender.parse().with_context(|| {
```
noot, line 526: can you use the parsed `receiver` from above?

crates/astria-sequencer/src/ibc/ics20_transfer.rs
```diff
@@ -288,290 +353,268 @@ impl AppHandlerExecute for Ics20Transfer {
...
+    amount: u128,
+    source_port: &PortId,
+    source_channel: &ChannelId,
+) -> Result<()> {
+    if is_refund_source_zone(&asset, source_port, source_channel) {
         state
-            .increase_balance(recipient, &denom_trace, packet_amount)
+            .decrease_ibc_channel_balance(source_channel, &asset, amount)
+            .await
+            .context("failed to withdraw refund amount from ibc20 channel")?;
```
noot, line 559:
```suggestion
            .context("failed to withdraw refund amount from ics20 channel")?;
```

crates/astria-sequencer/src/ibc/ics20_transfer.rs
```diff
@@ -288,290 +353,268 @@ impl AppHandlerExecute for Ics20Transfer {
...
+        state
+            .increase_balance(recipient, &asset, amount)
             .await
             .context("failed to update user account balance in execute_ics20_transfer")?;
     } else {
         // register denomination in global ID -> denom map if it's not already there
+        // XXX(janis): is this correct? Shouldn't an asset that is being refunded always
+        // be present in the state if one attempted to transfer it out?
```
noot, line 567: yes, you can remove this, if the asset was withdrawn it must have existed in the state already (as it's either native or was transferred in)

crates/astria-sequencer/src/ibc/ics20_transfer.rs
```diff
@@ -288,290 +353,268 @@ impl AppHandlerExecute for Ics20Transfer {
...
-/// bech32 address. If the sender is an ibc-compat address, then it will be converted to its
-/// base-prefixed version.
-///
-/// The emitted deposit as seen by the rollup will *never* be the ibc-compat prefixed version.
 // TODO: Add `err` with https://github.com/astriaorg/astria/issues/1386 being done
 #[instrument(skip_all)]
-async fn execute_withdrawal_refund_to_rollup<S: StateWrite>(
-    state: &mut S,
-    packet_data: FungibleTokenPacketData,
+async fn refund_tokens_to_rollup<S: StateWrite>(
```
noot, line 592: also check `is_refund_source_zone` in this function?

crates/astria-sequencer/src/ibc/ics20_transfer.rs
```diff
@@ -753,343 +797,388 @@ mod test {
...
+        )
+        .parse::<Denom>()
+        .unwrap();
+        state_tx.put_bridge_account_rollup_id(bridge_address, &rollup_id);
+        state_tx
+            .put_bridge_account_ibc_asset(bridge_address, &remote_asset_on_sequencer)
+            .unwrap();
+        state_tx
+            .put_ibc_channel_balance(&packet().chan_on_b, &remote_asset_on_sequencer, amount)
+            .unwrap();
```
noot, line 975: can remove this

crates/astria-sequencer/src/ibc/ics20_transfer.rs
```diff
@@ -753,343 +797,388 @@ mod test {
...
+            .get_ibc_channel_balance(&packet().chan_on_b, &nria())
             .await
             .expect("ics20 transfer to user account from escrow account should succeed");
-        assert_eq!(balance, 0);
+        assert_eq!(escrow_balance, 0);
     }
 
     #[tokio::test]
-    async fn execute_ics20_transfer_to_user_account_is_source_refund() {
+    async fn refund_user_account_with_source_zone_asset() {
```
noot, line 1136: also add a test for `refund_user_account_with_sink_zone_asset`?

noot (commented):
overall, this is a great change, a lot more readable than before. just a few smaller comments before approving!

crates/astria-sequencer/src/ibc/ics20_transfer.rs
SuperFluffy: I have removed these logs and instead instrumented the handlers to provide that information. 522f79bf

crates/astria-sequencer/src/ibc/ics20_transfer.rs
SuperFluffy: I have removed these logs and instead instrumented the handlers to provide that information. 522f79bf

crates/astria-sequencer/src/ibc/ics20_transfer.rs
SuperFluffy: Yes, thank you for noticing! I overlooked that bit. Done in 522f79bf

crates/astria-sequencer/src/ibc/ics20_transfer.rs
SuperFluffy: Nice! Good that my intuition was correct. Done in  69a86d02

crates/astria-sequencer/src/ibc/ics20_transfer.rs
SuperFluffy: Can you double check? I was trying to be faithful to the original implementation, which did not do that check for these rollup-refunds.

Here (on current main) you can see that the refund to the rollup is short-circuiting on the memo parse:
https://github.com/astriaorg/astria/blob/1be156e4818f3234fb0a3f03257f33ca2627d95f/crates/astria-sequencer/src/ibc/ics20_transfer.rs#L423-L432

And here the final denom that will be put in the deposit is constructed:
https://github.com/astriaorg/astria/blob/1be156e4818f3234fb0a3f03257f33ca2627d95f/crates/astria-sequencer/src/ibc/ics20_transfer.rs#L570-L580

crates/astria-sequencer/src/ibc/ics20_transfer.rs
SuperFluffy: Removed in  5231ec7f

crates/astria-sequencer/src/ibc/ics20_transfer.rs
SuperFluffy: Done in [5231ec7](https://github.com/astriaorg/astria/pull/1495/commits/5231ec7fb31d3a3a45b6cd56189aae9ad2465d99)

SuperFluffy:
Rebased on top of recent main.

In addition I instrumented all code that's called during ics20 execution, injecting address information, assets, amounts, etc into the generated spans. Function calls using eyre Reports as errors also generates an event in error cases.

crates/astria-sequencer/src/ibc/ics20_transfer.rs
noot: I think this is correct now - refunding a source asset should result in unescrowing funds from the channel account as well, not just re-minting.

crates/astria-sequencer/src/ibc/ics20_transfer.rs
```diff
@@ -308,298 +373,261 @@ impl AppHandlerExecute for Ics20Transfer {
...
+        state
+            .increase_balance(recipient, &asset, amount)
+            .await
+            .context("failed to increase user account balance")?;
+    } else {
+        state
+            .increase_balance(recipient, asset, amount)
+            .await
+            .wrap_err("failed to update user account balance in execute_ics20_transfer")?;
+    }
```
noot, line 607:
```suggestion
    if is_refund_source_zone(&asset, source_port, source_channel) {
        state
            .decrease_ibc_channel_balance(source_channel, &asset, amount)
            .await
            .context("failed to withdraw refund amount from escrow account")?;
    } 
    
   state
          .increase_balance(recipient, asset, amount)
          .await
          .wrap_err("failed to update user account balance when refunding")?;
 
```

noot (approved):
looks good to me :D

crates/astria-sequencer/src/ibc/ics20_transfer.rs
SuperFluffy: Nice!

SuperFluffy: I did the same thing in `receive_tokens`! 🥳

PR #1502, opened 2024-09-13
feat(sequencer)!: introduce transaction categories (type version)

### LINKED ISSUE #1414: sequencer: add notion of transaction categories

Lilyjjo:
## Summary
Introduces transaction 'types' that limit which types of Actions can be included in the same transaction.  

## Background

## Testing
TODO

## Metrics
TODO

## Breaking Changelist
TODO, assuming that this is breaking 

## Related Issues
Initial steps for #1412

closes #1414, #1416

### DIFF
--- a/crates/astria-bridge-contracts/src/lib.rs
+++ b/crates/astria-bridge-contracts/src/lib.rs
@@ -14,10 +14,7 @@ use astria_core::{
     },
     protocol::{
         memos,
-        transaction::v1alpha1::{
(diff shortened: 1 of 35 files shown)


Lilyjjo:
Note: trying out non-type version again as this is a very large PR with questionable gain

PR #1504, opened 2024-09-13
ci(cli): astria-cli image for smoke tests

### LINKED ISSUE #1475: test(cli): support docker image
We should have a docker image for the astria-cli, and update our smoke test to run within the image to ensure that local installed binaries don't affect smoke test passing.

This will also allow us to provide instructions for people to use the cli without running a full local install

┆Issue Number: ENG-805

joroshiba:
## Summary
Adds a image build for the cli tool, and updates the justfiles to utilize it.

## Background
There are two workflows which this fixes: 1) Users running the repo locally and running smoke tests could encounter issues by not having the cli installed or not having it updated. Now we just need to have docker. Additionally, we have been needing to build the `just-cli` for each of our smoke test runs. Instead all can now rely on the docker image.

## Changes
- Starts creating a new CLI image
- Updates the smoke tests to utilize CLI image in place of installing the cli
- Update docker build scripts to make it easier to use a local image locally

## Testing
CI/CD on this PR

## Related Issues

closes #1475

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -157,7 +157,9 @@ jobs:
           just deploy smoke-test $TAG
       - name: Run Smoke test
         timeout-minutes: 3
-        run: just run-smoke-test
+        run: |
+          TAG=sha-$(git rev-parse --short HEAD)
+          just run-smoke-test $TAG
 
   smoke-cli:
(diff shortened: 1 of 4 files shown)


SuperFluffy (approved):
Yeah, this is nice and elegant. I was expecting this to be a lot more work.

I wonder if we could get rid of the `install-cli` recipe entirely now. I'd rather have people use the docker image.

quasystaty1 (approved):

steezeburger (approved):

PR #1506, opened 2024-09-13
fix(sequencer): provide context in checktx response log

### LINKED ISSUE #1464: Mempool checkTx responses lack error information
Because sequencer uses `error.to_string()`, checkTx responses only contain the highest error level but completely lack its source chain, making it impossible to understand why an action was rejected (see code linked below).

The fix is to use anyhow's alternative error formatting through `format!("{error:#}")`.

https://github.com/astriaorg/astria/blob/f46d7f52ce1849b651b97dcf595406e62a2ed72f/crates/astria-sequencer/src/service/mempool.rs#L146-L151

┆Issue Number: ENG-799

itamarreif:
## Summary
We were using `e.to_string` to provide error messages in `CheckTx` responses. This results in dropping the context that the error objects carry around, only providing the top level error. This uses `anyhow`'s `format!("{e:#}")` to format the error with its context.

## Background
For example, the following snippet:
```rust
use anyhow::anyhow;
use astria_eyre::anyhow_to_eyre;

fn main() {
    let _ = astria_eyre::install();

    let anyhow_err = anyhow!("anyhow 0").context("anyhow 1");
    let combined = anyhow_to_eyre(anyhow_err)
        .wrap_err("astria eyre 0")
        .wrap_err("astria eyre 1");
    println!("err context: {combined:#}");
}
```

Will produce the following output:
```
err context: {"0": "astria eyre 1", "1": "astria eyre 0", "2": "anyhow 1", "3": "anyhow 0"}
```

## Changes
- changed all instances of `e.to_string()` to `format("{e:#}") in the sequencer's mempool service

## Related Issues

closes #1464

### DIFF
--- a/crates/astria-sequencer/src/service/mempool.rs
+++ b/crates/astria-sequencer/src/service/mempool.rs
@@ -147,7 +147,7 @@ async fn handle_check_tx<S: accounts::StateReadExt + address::StateReadExt + 'st
         Err(e) => {
             return response::CheckTx {
                 code: Code::Err(AbciErrorCode::INVALID_PARAMETER.value()),
-                log: e.to_string(),
+                log: format!("{e:#}"),
                 info: "failed decoding bytes as a protobuf SignedTransaction".into(),
                 ..response::CheckTx::default()
             };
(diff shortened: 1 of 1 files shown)


SuperFluffy:
Can you check if this still works with the recent move from anyhow to eyre?

itamarreif:
> Can you check if this still works with the recent move from anyhow to eyre?
Checked and provided the code snippet + output.

SuperFluffy (approved):

PR #1507, opened 2024-09-16
chore(sequencer)!: change `Deposit` byte length calculation

### LINKED ISSUE #1503: Investigate usage of fixed length integers for `Deposit` fields
Currently, our implementation of `Deposit` can vary in byte length based on the size of `source_action_index` due to Protobuf's [variable-width integers](https://protobuf.dev/programming-guides/encoding/#varints). This, in turn, affects our fees in a way that it shouldn't. We should use fixed length integers for this, and also investigate if our primitive type `Uint128` has similar issues which may cause the byte length to change based on deposit amount.

# Relevant Code
https://github.com/astriaorg/astria/blob/ac7222e3a8c230ab187ce7387efccd7f39120795/proto/sequencerblockapis/astria/sequencerblock/v1alpha1/block.proto#L76-L93

┆Issue Number: ENG-815

ethanoroshiba:
## Summary
Changed `Deposit` byte length calculation to use constant base value plus the lengths of`asset` and `destination_chain_address`.

## Background
The `get_deposit_byte_length()` function uses `prost::message::encoded_len()` to calculate the on-wire size of raw deposit events for fee calculation. Due to protobuf using variable length integers, however, this would mean that fees for `BridgeLock` actions would increase with larger deposit amounts or higher action indices, which should not be the case. This PR aims to address this issue and provide more fair fee calculation.

Offline discussion resulted in the decision to calculate a "fair" base length and then add the lengths of `asset` and `destination_chain_address`

## Changes
- Calculate `get_deposit_byte_length()` based on a constant base length plus the lengths of `asset` and `destination_chain_address`. The base length is determined by a unit test with "reasonable" values for all fields except for `asset` and `destination_chain_address`.

## Testing

- Added unit test to ensure calculated byte length doesn't change for any other changes than to `asset` and `destination_chain_address`.

## Breaking Changelist
- Updating the byte length calculation changes fees for `BridgeLock` actions, breaking the app hash.

## Related Issues
closes #1503


noot (approved):
codewise this looks fine, arguably values that encode to a smaller length should be cheaper, as the fee should be proportional to what's written to DA. however this will return a larger value rather than smaller, which seems okay

SuperFluffy (changes requested):
Blocking this because the bridge address construction cannot assume a fixed prefix.

SuperFluffy:
Suggestions for unblocking this while we redesign our fee handling.

1. Make a "base case" of a deposit with `destination_chain_address` and `asset` empty and an arbitrary prefix set (take the prefix as the max number of bytes permitted in a bech32m encoding, which would be 83 ASCII chars)
2. define an constant from the `Message::encoded_len` of this base case scenario
3. define the fee as the constant from 2. and add the lengths of `asset` and `destination_chain_address` to that

Support 2. with a unit test, and provide a few unit tests to support the implementation's claim that the calculated fee/byte length be never smaller than the actual proto len.

ethanoroshiba:
Should we also change `get_deposit_byte_length()` to be called something more reflective of its true functionality?

crates/astria-sequencer/src/bridge/bridge_lock_action.rs
```diff
@@ -145,22 +149,17 @@ impl ActionHandler for BridgeLockAction {
...
     let variable_length = deposit
         .asset()
         .to_string()
         .len()
         .checked_add(deposit.destination_chain_address().len())
         .expect("deposit byte length overflowed usize") as u128;
-    base_deposit_byte_length
+    DEPOSIT_BASE_BYTE_LENGTH
         .checked_add(variable_length)
         .expect("deposit byte length overflowed u128")
```
SuperFluffy, line 164: This is bad so deep in the stack. Let's just have this return an `Option<u128>`

crates/astria-sequencer/src/bridge/bridge_lock_action.rs
```diff
@@ -319,45 +312,63 @@ mod tests {
...
+    /// are empty strings, whose length will be added to the base cost at the time of
+    /// calculation.
+    ///
+    /// This test determines 165 bytes for an average deposit with empty `asset` and
+    /// `destination_chain_address`, which is divided by 10 to get our base byte length of 16. This
+    /// is to allow for more flexibility in overall fees (we have more flexibility multiplying by a
+    /// lower number, and if we want fees to be higher we can just raise the multiplier).
     #[test]
     #[ignore]
     fn get_base_deposit_byte_length() {
```
SuperFluffy, line 357: this dest should not be `ignore` and it should test that `DEPOSIT_BASE_BYTE_LENGTH` is the result.

crates/astria-sequencer/src/bridge/bridge_lock_action.rs
```diff
@@ -319,45 +312,63 @@ mod tests {
...
+        let deposit = Deposit::new(
+            astria_address(&[1; 20]),
+            RollupId::from_unhashed_bytes(b"test_rollup_id"),
+            u128::MAX,
+            test_asset(),
+            "someaddress".to_string(),
+            TransactionId::new([0; 32]),
+            u64::MAX,
+        );
+        assert_json_snapshot!(get_deposit_byte_len(&deposit));
```
SuperFluffy, line 343: To be honest I am not a fan of making this a snapshot test. We must provide a test for calculating `DEPOSIT_BASE_BYTE_LENGTH` in code.

I guess we can also add a snapshot test just for `DEPOSIT_BASE_BYTE_LENGTH` alone even though the use of that is questionable.

crates/astria-sequencer/src/bridge/bridge_lock_action.rs
```diff
@@ -145,22 +149,17 @@ impl ActionHandler for BridgeLockAction {
...
 }
 
-/// Returns a modified byte length of the deposit event. Length is calculated with maximum possible
-/// constants for all fields except `asset` and `destination_chain_address`, ergo it may not be
-/// representative of on-wire length. This should always return a byte length greater than or equal
-/// to the actual size of the serialized `Deposit`, but never less.
+/// Returns a modified byte length of the deposit event. Length is calculated with reasonable values
+/// for all fields except `asset` and `destination_chain_address`, ergo it may not be representative
+/// of on-wire length.
 pub(crate) fn get_deposit_byte_len(deposit: &Deposit) -> u128 {
```
SuperFluffy, line 155: This should be renamed to give a more "business logic" name. For example `fn calculate_deposit_fee` or something like that.

SuperFluffy (commented):
> Should we also change `get_deposit_byte_length()` to be called something more reflective of its true functionality?

Yeah, we should make give this a more meaningful name in its context. It's extremely rescriptive of what it's doing right now but doesn't give extra meaning.

crates/astria-sequencer/src/bridge/bridge_lock_action.rs
```diff
@@ -145,19 +154,31 @@ impl ActionHandler for BridgeLockAction {
...
+/// Returns a modified byte length of the deposit event. Length is calculated with reasonable values
+/// for all fields except `asset` and `destination_chain_address`, ergo it may not be representative
+/// of on-wire length.
+pub(crate) fn calculate_base_deposit_fee(deposit: &Deposit) -> Option<u128> {
+    let variable_length = deposit
+        .asset
+        .to_string()
+        .len()
+        .checked_add(deposit.destination_chain_address.len())
+        .expect("deposit byte length overflowed usize") as u128;
```
SuperFluffy, line 166: Options allow for the `?` operator. Please use that here.

This whole expression could then become something like:

```rust
let variable_len = deposit.asset.display_len().checked_add(deposit.destination_chain_address.len())?;
DEPOSIT_BASE_FEE.checked_add(variable_len)
```

Alternatively everything chained:

```rust
deposit.asset
    .display_len()
    .checked_add(deposit.destination_chain_address.len())
    .and_then(|var_len| DEPOSIT_BASE_FEE.checked_add(var_len))
```

crates/astria-sequencer/src/bridge/bridge_lock_action.rs
```diff
@@ -145,19 +154,31 @@ impl ActionHandler for BridgeLockAction {
...
-    use prost::Message as _;
-    let raw = deposit.clone().into_raw();
-    raw.encoded_len() as u128
+/// Returns a modified byte length of the deposit event. Length is calculated with reasonable values
+/// for all fields except `asset` and `destination_chain_address`, ergo it may not be representative
+/// of on-wire length.
+pub(crate) fn calculate_base_deposit_fee(deposit: &Deposit) -> Option<u128> {
+    let variable_length = deposit
+        .asset
+        .to_string()
```
SuperFluffy, line 163: Can you extend `Deposit` to provide a function like `display_len()` that calculates this using its formatted representation? Also provide tests in `astria-core` to test that.

We can avoid this allocation.

SuperFluffy (changes requested):

crates/astria-sequencer/src/bridge/bridge_lock_action.rs
ethanoroshiba: Done in c46fa8a, let me know if this implementation works for you!

crates/astria-core/src/primitive/v1/asset/denom.rs
```diff
@@ -691,4 +709,18 @@ mod tests {
 
         assert_eq!(None, denom.pop_leading_port_and_channel());
     }
+
+    #[test]
+    fn display_len_ok() {
```
SuperFluffy, line 714: Yeah, thats what I had in mind. But in addition provide more exhaustive tests for some other scenarios. Use `#[track_caller]` to define the actual assertion and call it repeatedly from a test, similar to what's done here: https://github.com/astriaorg/astria/blob/f11f9000a28a4a2959ee488bfad73b751a4194c9/crates/astria-core/src/primitive/v1/mod.rs#L771-L796

Here is some background reading on this: https://matklad.github.io/2021/05/31/how-to-test.html#Test-Driven-Design-Ossification

crates/astria-core/src/primitive/v1/asset/denom.rs
```diff
@@ -691,4 +709,18 @@ mod tests {
...
+        let trace_denom = Denom::from("a/long/path/to/denom".parse::<TracePrefixed>().unwrap());
```
SuperFluffy, line 715: Parse directly into a `Denom`. This is unnecessarily noisy.

crates/astria-core/src/primitive/v1/asset/denom.rs
```diff
@@ -691,4 +709,18 @@ mod tests {
...
+        assert_eq!(
+            trace_denom.to_string().len(),
```
SuperFluffy, line 717: Use the length of the input string (expressing it with the `track_caller` check I suggested above will make that neater).

crates/astria-core/src/primitive/v1/asset/denom.rs
```diff
@@ -75,6 +75,24 @@ impl Denom {
         };
         trace
     }
+
+    /// Calculates the length of the display formatted [Denom] without allocating a String.
+    #[must_use]
+    pub fn display_len(&self) -> Option<usize> {
```
SuperFluffy, line 81: In general we use checked addition, but in this case returning `usize` is more than acceptable.

On any currently available machine you will not blow through 2^64-1 bytes of RAM.

crates/astria-core/src/primitive/v1/asset/denom.rs
```diff
@@ -75,6 +75,24 @@ impl Denom {
...
+        match self {
+            Denom::TracePrefixed(trace) => {
+                let mut len: usize = 0;
+                for segment in &trace.trace.inner {
+                    len = len
+                        .checked_add(segment.port.len())
```
SuperFluffy, line 87: As mentioned above: this is unnecessary.

crates/astria-core/src/primitive/v1/asset/denom.rs
```diff
@@ -75,6 +75,24 @@ impl Denom {
         };
         trace
     }
+
+    /// Calculates the length of the display formatted [Denom] without allocating a String.
+    #[must_use]
+    pub fn display_len(&self) -> Option<usize> {
+        match self {
+            Denom::TracePrefixed(trace) => {
```
SuperFluffy, line 83: Define these on the trace-prefixed and ibc-prefixed types. Then delegate to them in this enum.

crates/astria-sequencer/src/bridge/bridge_lock_action.rs
```diff
@@ -99,7 +103,10 @@ impl ActionHandler for BridgeLockAction {
             .await
             .wrap_err("failed to get byte cost multiplier")?;
         let fee = byte_cost_multiplier
-            .saturating_mul(get_deposit_byte_len(&deposit))
+            .saturating_mul(
+                calculate_base_deposit_fee(&deposit)
```
SuperFluffy, line 107: Because we are doing saturating arithmetic here I would just keep doing that. Instead of returning early just replace it with with u128::MAX

crates/astria-sequencer/src/bridge/bridge_lock_action.rs
```diff
@@ -126,7 +133,10 @@ impl ActionHandler for BridgeLockAction {
             .get_bridge_lock_byte_cost_multiplier()
             .await
             .wrap_err("failed to get byte cost multiplier")?;
-        let fee = byte_cost_multiplier.saturating_mul(get_deposit_byte_len(&deposit));
+        let fee = byte_cost_multiplier.saturating_mul(
+            calculate_base_deposit_fee(&deposit)
```
SuperFluffy, line 137: As above, replace with `u128::MAX` in this context.

crates/astria-sequencer/src/bridge/bridge_lock_action.rs
```diff
@@ -226,18 +246,118 @@ mod tests {
...
+    /// on "reasonable" values for all fields except `asset` and `destination_chain_address`. These
+    /// are empty strings, whose length will be added to the base cost at the time of
+    /// calculation.
+    ///
+    /// This test determines 165 bytes for an average deposit with empty `asset` and
+    /// `destination_chain_address`, which is divided by 10 to get our base byte length of 16. This
+    /// is to allow for more flexibility in overall fees (we have more flexibility multiplying by a
+    /// lower number, and if we want fees to be higher we can just raise the multiplier).
+    #[test]
+    fn get_base_deposit_fee() {
```
SuperFluffy, line 345:
```suggestion
    fn calculate_base_deposit_fee() {
```

crates/astria-sequencer/src/bridge/bridge_lock_action.rs
```diff
@@ -226,18 +246,118 @@ mod tests {
...
+            .unwrap()
+                * 2;
         state
             .put_account_balance(from_address, &asset, 100 + expected_deposit_fee)
             .unwrap();
         bridge_lock.check_and_execute(&mut state).await.unwrap();
     }
+
+    #[test]
+    fn test_calculate_base_deposit_fee() {
```
SuperFluffy, line 267: You are in a test so don't call the test a test:
```suggestion
    fn calculated_fee_matches_expected_value() {
```

I'd also refactor the test to use `#[track_caller]`. See my comment for testing `display_len` in astria_core.

crates/astria-sequencer/src/bridge/bridge_lock_action.rs
```diff
@@ -226,18 +246,118 @@ mod tests {
...
+            bridge_address: astria_address(&[1; 20]),
+            rollup_id: RollupId::from_unhashed_bytes(b"test_rollup_id"),
+            amount: 0,
+            asset: "test_asset".parse().unwrap(),
+            destination_chain_address: "someaddress".to_string(),
+            source_transaction_id: TransactionId::new([0; 32]),
+            source_action_index: 0,
+        };
+        let calculated_len_2 = calculate_base_deposit_fee(&deposit).unwrap();
+        assert!(calculated_len_2 >= calculated_len_1);
```
SuperFluffy, line 303: Try to avoid `assert!`. If this fails the reader will not know what exactly failed. It's better to use `assert_eq!`. IMO that's also fine here because all of the values are deterministic.

crates/astria-sequencer/src/bridge/bridge_lock_action.rs
```diff
@@ -34,6 +34,10 @@ use crate::{
     utils::create_deposit_event,
 };
 
+// The base byte length of a deposit, as determined by the unit test
```
SuperFluffy, line 37: Make this a doc comment, i.e. start with `///`

crates/astria-sequencer/src/bridge/bridge_lock_action.rs
```diff
@@ -34,6 +34,10 @@ use crate::{
...
+// `get_base_deposit_byte_length()` below.
```
SuperFluffy, line 38:
```suggestion
/// `tests::get_base_deposit_byte_length()` below.

File: crates/astria-sequencer/src/bridge/bridge_lock_action.rs (hunk 13 above)
Comment (ethanoroshiba) [reply crates/astria-sequencer/src/bridge/bridge_lock_action.rs]: Replaced with `unwrap_or(u128::MAX)` in 3faab50. Let me know if this works!

File: crates/astria-core/src/primitive/v1/asset/denom.rs (hunk 12 above)
Comment (ethanoroshiba) [reply crates/astria-core/src/primitive/v1/asset/denom.rs]: I did so with `TracePrefixed`, but clippy threw a (IMO pretty reasonable) warning about having an unused self when applying to IBC prefixed, so I kept the fixed value in `Denom::display_len()`. Lmk if this works for you!

File: crates/astria-core/src/primitive/v1/asset/denom.rs (hunk 7 above)
Comment (ethanoroshiba) [reply crates/astria-core/src/primitive/v1/asset/denom.rs]: Thanks for the feedback and the link for background reading! I did this for both the core tests as well as the bridge lock tests, and attempted to be pretty exhaustive, but let me know if there are other cases that should be tested as well :)

File: crates/astria-sequencer/src/bridge/bridge_lock_action.rs [hunk 20]
```diff
@@ -145,19 +150,29 @@ impl ActionHandler for BridgeLockAction {
...
-    raw.encoded_len() as u128
+/// Returns a modified byte length of the deposit event. Length is calculated with reasonable values
+/// for all fields except `asset` and `destination_chain_address`, ergo it may not be representative
+/// of on-wire length.
+pub(crate) fn calculate_base_deposit_fee(deposit: &Deposit) -> Option<u128> {
+    deposit
+        .asset
+        .display_len()
+        .checked_add(deposit.destination_chain_address.len())
+        .and_then(|var_len| DEPOSIT_BASE_FEE.checked_add(var_len as u128))
```
SuperFluffy, line 161: It would be incredibly odd if we couldn't convert a `usize` to a `u128`. But for the sake of consistency, can you `u128::try_from(u128).expect("converting a usize to a u128 should work on any currently existing machine")`? This is for consistency.

crates/astria-sequencer/src/bridge/bridge_lock_action.rs
```diff
@@ -226,18 +241,100 @@ mod tests {
...
                 rollup_id,
                 amount: 100,
                 asset: asset.clone(),
                 destination_chain_address: "someaddress".to_string(),
                 source_transaction_id: transaction_id,
                 source_action_index: 0,
-            }) * 2;
+            })
+            .unwrap()
+                * 2;
```
SuperFluffy, line 254: That's one ugly `rustfmt`

crates/astria-sequencer/src/bridge/bridge_lock_action.rs
```diff
@@ -226,18 +241,100 @@ mod tests {
...
+            destination_chain_address: absurd_string,
+            ..default_deposit()
+        });
+    }
+
+    #[track_caller]
+    #[allow(clippy::arithmetic_side_effects)] // allow: test will never overflow u128
+    fn assert_correct_base_deposit_fee(deposit: &Deposit) {
+        let calculated_len = calculate_base_deposit_fee(deposit).unwrap();
+        let expected_len = DEPOSIT_BASE_FEE
```
SuperFluffy, line 294: This just copies what's inside `fn calculate_base_deposit_fee`. I think you should replace this with a fixed number instead, because all of this is deterministic.

crates/astria-sequencer/src/bridge/bridge_lock_action.rs
```diff
@@ -226,18 +241,100 @@ mod tests {
...
+            amount: Some(1000.into()),
+            asset: String::new(),
+            destination_chain_address: String::new(),
+            source_transaction_id: Some(TransactionId::new([0; TRANSACTION_ID_LEN]).to_raw()),
+            source_action_index: 0,
+        };
+        assert_eq!(DEPOSIT_BASE_FEE, raw_deposit.encoded_len() as u128 / 10);
+    }
+
+    fn default_deposit() -> Deposit {
```
SuperFluffy, line 329: It's not really default, maybe:
```suggestion
    fn reference_deposit() -> Deposit {
```

crates/astria-core/src/primitive/v1/asset/denom.rs
```diff
@@ -75,6 +75,15 @@ impl Denom {
         };
         trace
     }
+
+    /// Calculates the length of the display formatted [Denom] without allocating a String.
+    #[must_use]
+    pub fn display_len(&self) -> usize {
+        match self {
+            Denom::TracePrefixed(trace) => trace.display_len(),
+            Denom::IbcPrefixed(_) => 68, // "ibc/" + 64 hex characters
```
SuperFluffy, line 84: impl this fixed number on the ibc type as well and delegate there, for consistency.

SuperFluffy (approved):
This looks great, especially the tests are very neat. Thank you for those!

I just left a few minor nits, but otherwise I see this as being ready for merging.

crates/astria-core/src/primitive/v1/asset/denom.rs
```diff
@@ -257,6 +266,18 @@ impl TracePrefixed {
     pub fn push_trace_segment(&mut self, segment: PortAndChannel) {
         self.trace.push(segment);
     }
+
+    /// Calculates the length of the display formatted [`TracePrefixed`] without allocating a
+    /// String.
+    #[must_use]
+    #[allow(clippy::arithmetic_side_effects)] // allow: length should never overflow usize::MAX
```
SuperFluffy, line 273: We are already on top of `1.81.0` - let's start using `#[expect(<LINT>, reason = "<reason>")]`.

PR #1508, opened 2024-09-16
chore(core, sequencer): add deposit builder

### LINKED ISSUE #1448: Clean up deposit constructor
Currently, `Deposit::new()` as many arguments and is unwieldy. We should prefer `try_from_raw` where possible and look into a better way of constructing the object, potentially a builder.

https://github.com/astriaorg/astria/pull/1410#discussion_r1743729542

https://github.com/astriaorg/astria/blob/291156235de56134f74e393459ab32ed11634451/crates/astria-core/src/sequencerblock/v1alpha1/block.rs#L1317-L1335

┆Issue Number: ENG-787

ethanoroshiba:
## Summary
Added `DepositBuilder` to construct deposit events.

## Background
Constructor was previously relying on a constructor with 7 arguments, as pointed out here: https://github.com/astriaorg/astria/pull/1410#discussion_r1743729542

## Changes
- Added `DepositBuilder` to core, replaced all instances of `Deposit::new()` with `DepositBuilder::build`.

## Testing
Passing all tests

## Related Issues

closes #1448

### DIFF
--- a/crates/astria-core/src/sequencerblock/v1alpha1/block.rs
+++ b/crates/astria-core/src/sequencerblock/v1alpha1/block.rs
@@ -1315,27 +1315,6 @@ impl From<Deposit> for crate::generated::sequencerblock::v1alpha1::Deposit {
 
 impl Deposit {
     #[must_use]
-    pub fn new(
-        bridge_address: Address,
-        rollup_id: RollupId,
-        amount: u128,
-        asset: asset::Denom,
-        destination_chain_address: String,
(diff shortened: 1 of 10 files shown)


noot:
I think this overlaps with the changes in #1495

SuperFluffy:
Agree with @noot. Since the builder does not provide any extra benefit (checks for example) and since all fields in the deposit are `pub` the builder doesn't provide any benefit.

Issue #1474, opened 2024-09-10
cli: ensure support for every sequencer action type & abci query


joroshiba:
Currently `astria-cli` doesn't support all of our action types which is a hindrance to running manual and adding smoke tests. We want to support all action types, and abci query such that they are all easily tested through our domain types logic.

┆Issue Number: ENG-804

joroshiba:
This issue is stale because it has been open 45 days with no activity. Remove stale label or this issue
be closed in 7 days.

joroshiba:
This issue was closed because it was stale

PR #1509, opened 2024-09-16
feat(sequencer)!: Add IBC sudo change action

### LINKED ISSUE #1480: feat(sequencer): action to change ibc sudo account
Currently the IBC sudo account cannot be changed, there should be an option to update this key.

ethanoroshiba:
## Summary
Added an action for changing the IBC sudo address.

## Background
IBC sudo address wasn't able to be changed before.

## Changes
- Added IBC sudo change action to proto, core, and sequencer.

## Testing
Added two new unit tests to ensure the new action works and errs as expected.

## Breaking Changelist
- Added test of new action to `app_execute_transaction_with_every_action_snapshot()`, breaking app hash.

## Related Issues
closes #1480


noot (approved):

SuperFluffy (commented):
Things that were brought up offline, but where no conclusion was had:

1. should the submitter provide the old address in the payload as an extra precaution?
2. should the new address be also signed by the new sudoer, and the signature provided as part of the payload? The idea being that the submitter must prove that control isn't just given away?

crates/astria-cli/src/cli/sequencer.rs
SuperFluffy, line 1: I prefer these CLI changes be left out of this PR. Especially since they don't come with any sort of testing.

There is an active effort of redoing CLI (and working on actions): https://github.com/astriaorg/astria/issues/1474

joroshiba:
> * should the submitter provide the old address in the payload as an extra precaution?
I don't think this is necessary as address is derivable from their signature.

> * should the new address be also signed by the new sudoer, and the signature provided as part of the payload? The idea being that the submitter must prove that control isn't just given away?

This would be kinda nice but out of scope for this PR, we would need some tooling to create these requests and it should probably apply to all the account signer changes. This is essentially saying  you need approval of the person receiving the permissions to give them away.

SuperFluffy (approved):

PR #1512, opened 2024-09-17
feat(sequencer)!: transaction categories on UnsignedTransaction

### LINKED ISSUE #1414: sequencer: add notion of transaction categories

Lilyjjo:
## Summary
Adds restrictions on what type of `Actions` can be included in the same `UnsignedTransaction`. 

## Background

## Changes
- Introduced the `Actions` struct to hold valid groupings of `Actions`.
- Changed the `UnsignedTransaction` struct to have private fields.
- Changed the `UnsignedTransaction`'s `action` to be a `Actions` type instead of just a vector of `Actions`.

## Testing
TODO

## Metrics
TODO

## Breaking Changelist
TODO

## Related Issues
Initial steps for #1412

closes #1414, #1416


crates/astria-composer/src/executor/mod.rs
```diff
@@ -681,11 +681,12 @@ impl Future for SubmitFut {
...
-                    let tx = UnsignedTransaction {
-                        actions: this.bundle.clone().into_actions(),
-                        params,
-                    }
-                    .into_signed(this.signing_key);
+                    let tx = UnsignedTransaction::builder()
+                        .actions(this.bundle.clone().into_actions())
+                        .params(params)
+                        .build()
+                        .expect("failed to build transaction from actions")
```
Lilyjjo, line 688: I don't know if it is acceptable for a malformed transaction to crash this service / if this function call is behind a protected endpoint.

crates/astria-composer/src/executor/mod.rs
```diff
@@ -759,11 +760,12 @@ impl Future for SubmitFut {
...
-                        let tx = UnsignedTransaction {
-                            actions: this.bundle.clone().into_actions(),
-                            params,
-                        }
-                        .into_signed(this.signing_key);
+                        let tx = UnsignedTransaction::builder()
+                            .actions(this.bundle.clone().into_actions())
+                            .params(params)
+                            .build()
+                            .expect("failed to build transaction from actions")
```
Lilyjjo, line 767: Same here.

crates/astria-composer/src/executor/mod.rs
Lilyjjo: The code in the `bundle_factory` is restricted to just work for `SequenceActions` so this shouldn't ever fail, might be better to move the construction of the `UnsignedTransaction` to the bundle code or to have the `bundle_factory` code to keep everything as explicit `SequenceActions`

crates/astria-sequencer/src/app/mod.rs
```diff
@@ -1024,10 +1025,10 @@ impl App {
 
         // flag mempool for cleaning if we ran a fee change action
         self.recost_mempool = self.recost_mempool
-            || signed_tx
-                .actions()
-                .iter()
-                .any(|action| matches!(action, Action::FeeAssetChange(_) | Action::FeeChange(_)));
+            || matches!(signed_tx.group(), Some(ActionGroup::BundlableSudo(_)))
+                && signed_tx.actions().iter().any(|action| {
+                    matches!(action, Action::FeeAssetChange(_) | Action::FeeChange(_))
```
SuperFluffy, line 1030: It's idiomatic to implement these as inherent methods on an enum so that this would read like:

```rust
// inherent method
impl Action {
    fn is_fee_asset_change(&self) -> bool {
        matches!(action, Self::FeeAssetChange(_))
    }
    fn is_fee_change(&self) -> bool {
        matches!(action, Self::FeeChange(_))
    }
}

// the function
&& signed_tx.actions().iter().any(|act| act.is_fee_asset_change() || act.is_fee_change())
```

crates/astria-cli/src/commands/bridge/submit.rs
```diff
@@ -129,14 +129,17 @@ async fn submit_transaction(
...
+    let tx = UnsignedTransaction::builder()
+        .actions(actions)
+        .params(
+            TransactionParams::builder()
+                .nonce(nonce_res.nonce)
+                .chain_id(chain_id)
+                .build(),
+        )
+        .build()
+        .expect("failed to build transaction from actions")
```
SuperFluffy, line 141: You are in an environment that exits early on errors. I would return an error here to potentially get some error feedbackk.

crates/astria-cli/src/commands/sequencer.rs
```diff
@@ -475,14 +475,17 @@ async fn submit_transaction(
...
+    let tx = UnsignedTransaction::builder()
+        .actions(vec![action])
+        .params(
+            TransactionParams::builder()
+                .nonce(nonce_res.nonce)
+                .chain_id(chain_id)
+                .build(),
+        )
+        .build()
+        .expect("failed to build transaction from actions")
```
SuperFluffy, line 487: Same here - you can just return an error, so `wrap_err("...")?` is not more work

crates/astria-sequencer/src/proposal/commitment.rs
```diff
@@ -121,13 +121,17 @@ mod test {
...
         let signing_key = SigningKey::new(OsRng);
-        let tx = UnsignedTransaction {
-            params: TransactionParams::builder()
-                .nonce(0)
-                .chain_id("test-chain-1")
-                .build(),
-            actions: vec![sequence_action.clone().into(), transfer_action.into()],
-        };
+
+        let tx = UnsignedTransaction::builder()
```
SuperFluffy, line 125: can you make a utility for this? It looks like there is a lot of repetition for constructing these transactions here.

SuperFluffy (commented):
Looks like an elegant solution to me, and I am always a fan of locking down types.

I think this is ready to be merged, but I would like to see the errors for the wire-type -> domain-type conversion to contain more information.

Fraser999 (commented):
Initial (incomplete) review.

crates/astria-core/src/protocol/transaction/v1alpha1/action_groups.rs
```diff
@@ -0,0 +1,228 @@
...
+    fn from(val: BundlableSudo) -> ActionGroup {
+        ActionGroup::BundlableSudo(val)
+    }
+}
+
+impl From<Sudo> for ActionGroup {
+    fn from(val: Sudo) -> ActionGroup {
+        ActionGroup::Sudo(val)
+    }
+}
```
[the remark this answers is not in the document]
SuperFluffy: Note that this was my original suggestion in order to document the group membership for each action.

Doing it via the `Action` enum moves this property away from each action.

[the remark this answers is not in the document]
Fraser999: Well, the associated consts are invisible in rustdoc since they're private, but that's easily remedied I guess.  Likewise, the `BelongsToGroup` trait is private, so it's also invisible.

If we make the `BelongsToGroup` trait `pub` but remove the associated consts, the docs are still pretty unhelpful since each action will have the same generated docs, like:
```
impl BelongsToGroup for BridgeLockAction
fn belongs_to_group(&self) -> ActionGroup
```

We could maybe change the trait to:
```rust
pub trait BelongsToGroup: Sealed {
    const GROUP: ActionGroup;
}
```
so that generated docs look like:
```
impl BelongsToGroup for BridgeLockAction
const GROUP: ActionGroup = ActionGroup::BundlableGeneral
```
It would mean the trait can't be implemented on the `Action` enum, but I think that's a good thing if the whole reason for the trait is to support rustdoc documentation showing which group the given action belongs to, and that is impossible for an enum which can be one of several groups.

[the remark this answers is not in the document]
SuperFluffy: Yeah, I like that.

crates/astria-composer/src/executor/mod.rs
Lilyjjo: Went the route of putting it into a function

Lilyjjo:
Since last set of reviews:
- Rewrote action_group.rs trait setup in suggested way
- Changed `UnsignedTransaction()` from using a `Builder` to use a `new()` constructor
- Made error handling more verbose

crates/astria-composer/src/executor/bundle_factory/mod.rs
```diff
@@ -74,6 +76,12 @@ impl SizedBundle {
         }
     }
 
+    #[allow(clippy::panic)] // method is expected to never panic as an invariant of the type
```
noot, line 79: because the actions are only ever of the `SequenceAction` type?

crates/astria-core/src/protocol/test_utils.rs
```diff
@@ -104,16 +104,18 @@ impl ConfigureSequencerBlock {
...
                 actions,
-                params: TransactionParams::builder()
+                TransactionParams::builder()
                     .nonce(1)
                     .chain_id(chain_id.clone())
                     .build(),
-            };
+            )
+            .expect(
+                "failed to build unsigned transaction, should be okay since only sequence actions",
```
noot, line 115:
```suggestion
                "should be able to build unsigned transaction since only sequence actions are contained",
```

crates/astria-core/src/protocol/transaction/v1alpha1/action_group.rs
```diff
@@ -0,0 +1,491 @@
...
+        self.inner.first().map(Group::group)
+    }
+
+    pub(super) fn default() -> Self {
+        Self {
+            inner: vec![],
+        }
+    }
+
+    pub(super) fn from_list_of_actions(actions: Vec<Action>) -> Result<Self, Error> {
```
noot, line 242:
```suggestion
    pub(super) fn try_from_list_of_actions(actions: Vec<Action>) -> Result<Self, Error> {
```

since it returns a result

crates/astria-core/src/protocol/transaction/v1alpha1/action_group.rs
```diff
@@ -0,0 +1,491 @@
...
+pub struct General;
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub struct BundlableSudo;
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub struct Sudo;
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum ActionGroup {
+    BundlableGeneral(BundlableGeneral),
+    General(General),
```
noot, line 117:
```suggestion
    UnbundleableGeneral(UnbundleableGeneral),
```

I think `General` is a bit unclear, would prefer to explicitly say unbundleable (and maybe even rename BundleableGeneral to just General?)

crates/astria-core/src/protocol/transaction/v1alpha1/action_group.rs
```diff
@@ -0,0 +1,491 @@
...
+    BundlableSudo(BundlableSudo),
+    Sudo(Sudo),
```
noot, line 119:
```suggestion
    UnbundleableSudo(UnbundleableSudo),
```

same here

crates/astria-core/src/protocol/transaction/v1alpha1/action_group.rs
```diff
@@ -0,0 +1,491 @@
...
+}
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub struct BundlableGeneral;
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub struct General;
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub struct BundlableSudo;
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub struct Sudo;
```
noot, line 112: are these ever used anywhere? they seem unnecessary, can the `ActionGroup` enum variant be used?

crates/astria-sequencer/src/app/test_utils.rs
```diff
@@ -223,20 +223,21 @@ pub(crate) fn mock_tx(
...
             }
             .into(),
         ],
-    };
+        TransactionParams::builder()
+            .nonce(nonce)
+            .chain_id("test")
+            .build(),
+    )
+    .expect("Failed to build unsigned transaction");
```
noot, line 240:
```suggestion
    .expect("can build unsigned transaction");
```

crates/astria-sequencer/src/app/tests_app/mempool.rs
```diff
@@ -48,23 +48,23 @@ async fn trigger_cleaning() {
...
             .into(),
         ],
-    }
-    .into_signed(&sudo);
+        TransactionParams::builder()
+            .nonce(0)
+            .chain_id("test")
+            .build(),
+    )
+    .expect("failed to build unsigned transaction")
```
noot, line 66:
```suggestion
    .expect("can build unsigned transaction")
```

noot: same for other in this file

crates/astria-core/src/protocol/transaction/v1alpha1/action_group.rs
Lilyjjo: I agree, I'll change the names to note the unbundleable instead!

crates/astria-core/src/protocol/transaction/v1alpha1/action_group.rs
Lilyjjo: You're right, this was left over from a different implementation route. Good catch!

crates/astria-composer/src/executor/bundle_factory/mod.rs
Lilyjjo: made more descriptive

crates/astria-sequencer/src/app/mod.rs
```diff
@@ -1024,10 +1025,11 @@ impl App {
 
         // flag mempool for cleaning if we ran a fee change action
         self.recost_mempool = self.recost_mempool
-            || signed_tx
-                .actions()
-                .iter()
-                .any(|action| matches!(action, Action::FeeAssetChange(_) | Action::FeeChange(_)));
+            || matches!(signed_tx.group(), Some(ActionGroup::BundlableSudo(_)))
```
SuperFluffy: It's convetion to create inherent methods for checks like this (similar to the check below; I think I commented this before?):

```rust
impl SignedTransaction {
    // NOTE: I used `contains_*` rather than `is_` because it read weird; maybe you can come up with better naming?
    fn contains_bundleable_sudo_actions(&self) -> bool {
        matches!(self.group(), ActionGroup::BundleSudo(_))
    }
}
```

crates/astria-core/src/protocol/transaction/v1alpha1/action_group.rs
```diff
@@ -0,0 +1,491 @@
...
+            Action::BridgeUnlock(_) => BridgeUnlockAction::GROUP,
+            Action::BridgeSudoChange(_) => BridgeSudoChangeAction::GROUP,
+            Action::FeeChange(_) => FeeChangeAction::GROUP,
+            Action::FeeAssetChange(_) => FeeAssetChangeAction::GROUP,
+            Action::Ibc(_) => IbcRelay::GROUP,
+        }
+    }
+}
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
```
SuperFluffy, line 81: Note on all these types:

Do we really need to expose them publicly? Looking at the sequencer code, this is only needed in the `matches!` statement in mempool?

Instead of exposing all these groups publicly, can we just provide inherents methods like `UnsignedTransaction::contains_bundleable_sudo_actions` (or potentially better names)?

This would significantly reduce the public surface area this patch adds.

crates/astria-core/src/protocol/transaction/v1alpha1/action_group.rs
```diff
@@ -0,0 +1,491 @@
...
+pub enum ActionGroup {
+    BundlableGeneral(BundlableGeneral),
+    General(General),
+    BundlableSudo(BundlableSudo),
+    Sudo(Sudo),
+}
+
+impl fmt::Display for ActionGroup {
```
SuperFluffy, line 89: These display impls are essentially their `Debug` counterparts. Can you make them more human readable?

Just something like `bundleable general` instead of snakecase.

crates/astria-core/src/protocol/transaction/v1alpha1/action_group.rs
```diff
@@ -0,0 +1,491 @@
...
+}
+
+impl From<Sudo> for ActionGroup {
+    fn from(val: Sudo) -> ActionGroup {
+        ActionGroup::Sudo(val)
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+enum ErrorKind {
```
SuperFluffy, line 101: Noted further down on the custom `Display` implementation on public `Error` type: lean on the `error` attribute of the `thiserror` proc macro to generate that code for you. I.e. use struct variants, for example like so:

```rust
#[derive(Debug, thiserrror::Error)]
enum ErrorKind {
    #[error("actions of different types are not permitted; input contained `{group_one}` and `{group_two}`")]
    Mixed {
        group_one: ActionGroup,
        group_two: ActionGroup,
    }
}
```

crates/astria-core/src/protocol/transaction/v1alpha1/action_group.rs
```diff
@@ -0,0 +1,491 @@
...
+    #[error("input contains mixed action types")]
+    Mixed,
+    #[error("attempted to create bundle with non bundleable `ActionGroup` type")]
+    NotBundleable,
+}
+
+#[derive(Debug)]
```
SuperFluffy, line 108: For consistency with the rest of the crate:
```rust
#[error(transparent)]
pub struct Error(ErrorKind);
```
And put the context into the kinds.

crates/astria-core/src/protocol/transaction/v1alpha1/action_group.rs
```diff
@@ -0,0 +1,491 @@
...
+        let group = match actions_iter.next() {
+            Some(action) => action.group(),
+            None => {
+                // empty `actions`, so invariants met
+                return Ok(Self::default());
+            }
+        };
+
+        // assert size constraints on non-bundlable action groups
+        if matches!(group, ActionGroup::General(_) | ActionGroup::Sudo(_)) && actions.len() > 1 {
```
SuperFluffy: Create an inherent method for this: `ActionGroup::is_bundleable`

crates/astria-core/src/protocol/transaction/v1alpha1/action_group.rs
```diff
@@ -0,0 +1,491 @@
...
+            }
+        }
+
+        Ok(Self {
+            inner: actions,
+        })
+    }
+}
+
+#[cfg(test)]
```
SuperFluffy, line 217: Let's move these into a `tests.rs` crate, i.e.:

```
action_group/
  mod.rs
  tests.rs
```

crates/astria-core/src/protocol/transaction/v1alpha1/mod.rs
```diff
@@ -198,7 +203,12 @@ impl SignedTransaction {
 
     #[must_use]
     pub fn actions(&self) -> &[Action] {
-        &self.transaction.actions
+        self.transaction.actions.actions()
+    }
+
+    #[must_use]
+    pub fn group(&self) -> Option<ActionGroup> {
```
SuperFluffy, line 210: Noted elsewhere: I think we can avoid exposing this publicly for the time being by providing `contains_xxx_group_actions` or `is_xxx_group_actions`

crates/astria-sequencer/src/app/test_utils.rs
SuperFluffy: The semantics of @noot suggestion are better, but the new statement is somewhat vacuous. But since you want to create `mock_tx` it's clear from the context that you probably expect this to work as inputs to other tests:
```suggestion
    .unwrap();
```

crates/astria-sequencer/src/app/tests_app/mempool.rs
SuperFluffy: As above: this does not provide info as to why you expect this to have worked. If you think it's trivially clear why this should have worked, prefer `unwrap()`.

crates/astria-sequencer/src/app/tests_app/mempool.rs
```diff
@@ -147,24 +147,23 @@ async fn do_not_trigger_cleaning() {
...
             .into(),
         ],
-    }
-    .into_signed(&alice);
+        TransactionParams::builder()
+            .nonce(0)
+            .chain_id("test")
+            .build(),
+    )
+    .expect("failed to build unsigned transaction")
```
SuperFluffy: As above: this does not provide info as to why you expect this to have worked. If you think it's trivially clear why this should have worked, prefer `unwrap()`.

crates/astria-core/src/protocol/transaction/v1alpha1/action.rs
```diff
@@ -252,6 +260,30 @@ impl TryFrom<raw::Action> for Action {
     }
 }
 
+pub(super) trait ActionName {
```
SuperFluffy, line 263: Please remove this. All of these should be done via the `Protobuf` trait implemented at the root of the crate - I intend to rename it to `DomainType`.

crates/astria-core/src/protocol/transaction/v1alpha1/action_group.rs
Lilyjjo: We might need access to them in a future PR for sorting transactions in the block building process, I can make them private for now

crates/astria-core/src/protocol/transaction/v1alpha1/action.rs
SuperFluffy: Discussed offline: this should flow through `Protobuf::full_name` but is out of scope for this PR. @Lilyjjo will create a PR to follow up on this.

crates/astria-sequencer/src/app/snapshots/astria_sequencer__app__tests_breaking_changes__app_execute_transaction_with_every_action_snapshot.snap
```diff
@@ -1,39 +1,38 @@
...
-  84,
-  121,
-  150,
-  21,
-  207,
-  248,
-  173,
-  132,
-  77,
+   96,
```
Lilyjjo, line 6:
```suggestion
  96,
```

crates/astria-core/src/protocol/transaction/v1alpha1/action.rs
Lilyjjo: Tracked in a note in the code and here: #1567

crates/astria-core/src/protocol/transaction/v1alpha1/action_group/tests.rs
```diff
@@ -0,0 +1,244 @@
...
+            new_address: address,
+        }),
+    ];
+
+    assert_eq!(
+        Actions::try_from_list_of_actions(actions)
+            .unwrap_err()
+            .to_string(),
+        "attempted to create bundle with non bundleable `ActionGroup` type: unbundleable sudo"
+    );
```
Fraser999, line 162:
```suggestion
    let error_kind = Actions::try_from_list_of_actions(actions).unwrap_err().0;
    assert!(
        matches!(error_kind, ErrorKind::NotBundleable { .. }),
        "expected ErrorKind::NotBundleable, got {error_kind:?}"
    );
```
Similar suggestion for following two tests also.

crates/astria-core/src/protocol/transaction/v1alpha1/mod.rs
```diff
@@ -381,35 +417,50 @@ enum UnsignedTransactionErrorKind {
...
     nonce: u32,
     chain_id: TChainId,
+    actions: Vec<Action>,
 }
 
-impl TransactionParamsBuilder {
-    fn new() -> Self {
+impl UnsignedTransactionBuilder {
+    #[must_use]
+    #[allow(clippy::new_without_default)]
```
Fraser999, line 432: We should be able to just derive `Default` on the builder rather than doing this `allow` I think.

crates/astria-core/src/protocol/transaction/v1alpha1/mod.rs
```diff
@@ -381,35 +417,50 @@ enum UnsignedTransactionErrorKind {
         raw::UnsignedTransaction::type_url()
     )]
     DecodeAny(#[source] prost::DecodeError),
+    #[error("`actions` field does not make a valid `ActionGroup`")]
+    ActionGroup(#[source] action_group::Error),
 }
 
-pub struct TransactionParamsBuilder<TChainId = std::borrow::Cow<'static, str>> {
+pub struct UnsignedTransactionBuilder<TChainId = std::borrow::Cow<'static, str>> {
```
Fraser999, line 424: Having a generic parameter seems a little excessive to me.  I think it'd be fine to just have `chain_id: String` and have the setter as `pub fn chain_id<T: Into<String>>(self, chain_id: T)`

crates/astria-sequencer/src/app/tests_breaking_changes.rs
```diff
@@ -307,3 +351,175 @@ async fn app_execute_transaction_with_every_action_snapshot() {
...
+                    new_address: bob_address,
+                }
+                .into(),
+            ])
+            .nonce(2)
+            .chain_id("test")
+            .try_build()
+            .unwrap_err()
+            .to_string(),
+        "attempted to create bundle with non bundleable `ActionGroup` type: unbundleable sudo"
```
Fraser999, line 469: As before, I think these tests should match on the `ErrorKind` variant rather than on the `Display` output of the error.

Fraser999 (commented):
Other than the aspect discussed offline, just nitpicks from me.

crates/astria-sequencer/src/app/tests_breaking_changes.rs
Lilyjjo: I made the whole `ActionGroup` enum public in the process to achieve this.

crates/astria-core/src/protocol/transaction/v1alpha1/action_group/mod.rs
```diff
@@ -0,0 +1,201 @@
...
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+#[error(transparent)]
+pub struct Error(ErrorKind);
+
+impl Error {
+    #[must_use]
+    pub fn kind(&self) -> &ErrorKind {
```
SuperFluffy, line 118: private?

crates/astria-core/src/protocol/transaction/v1alpha1/action_group/mod.rs
```diff
@@ -0,0 +1,201 @@
...
+
+    fn not_bundleable(group: ActionGroup) -> Self {
+        Self(ErrorKind::NotBundleable {
+            group,
+        })
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum ErrorKind {
```
SuperFluffy, line 142: private? Do you need to check why bundling failed and take action based on it?

crates/astria-core/src/protocol/transaction/v1alpha1/mod.rs
```diff
@@ -381,59 +417,69 @@ enum UnsignedTransactionErrorKind {
         raw::UnsignedTransaction::type_url()
     )]
     DecodeAny(#[source] prost::DecodeError),
+    #[error("`actions` field does not make a valid `ActionGroup`")]
```
SuperFluffy, line 420: `ActionGroup` is not a concept outside of the crate:
```suggestion
    #[error("`actions` field does not form a valid group of actions")]
```

crates/astria-core/src/protocol/transaction/v1alpha1/action_group/mod.rs
Lilyjjo: I didn't know how to access the error for a cleaner testing style that Fraser recommended without making public: https://github.com/astriaorg/astria/blob/3e79a1c7794fe1fd78dcc25ae72de8b8f8b399c7/crates/astria-sequencer/src/app/tests_breaking_changes.rs#L525

crates/astria-core/src/protocol/transaction/v1alpha1/action_group/mod.rs
Lilyjjo: Responded on the comment above

crates/astria-core/src/protocol/transaction/v1alpha1/action_group/mod.rs
SuperFluffy: I think these tests are misplaced.

If I see it correctly, you want to check if the bundling logic works correctly, right? The correct place to test this is where the bundling is implemented, i.e. here in `astria-core`. `astria-sequencer` as a downstream consumer should be able to rely on the implementation being correct.

From a quick glance I could not see if there is any logic that branches on the error kind and subsequently takes a different path.

crates/astria-sequencer/src/app/tests_breaking_changes.rs
```diff
@@ -307,3 +352,178 @@ async fn app_execute_transaction_with_every_action_snapshot() {
...
+
+// Note: this test ensures that all actions are in their expected
+// bundleable vs non-bundleable category. Tests all actions except
+// IbcRelay.
+//
+// If any PR changes the bundleable status of an action, the PR must
+// be marked as breaking.
+#[allow(clippy::too_many_lines)]
+#[tokio::test]
+async fn app_transaction_bundle_categories() {
```
SuperFluffy, line 364: I commented elsewhere, the tests here test implementation details of `astria-core` and are redundant.

It is implicitly understood that changes in behavior of `astria-core` are immediate breaking for `astria-sequencer` because they are a stepping stone.

Lilyjjo: Ok, will remove.

crates/astria-core/src/protocol/transaction/v1alpha1/action_group/mod.rs
SuperFluffy: There really isn't, but that's why we have a review process and rely on authors to note breaking changes.

But it is implicitly understood that changing behavior of fundamental types in `astria-core` cause immediate breakage in `astria-sequencer`.

Copying tests and publicly exposing implementation deatils for what's effectively a unit test in other parts of the code base just adds extra churn without giving you extra safety.

crates/astria-composer/src/executor/bundle_factory/mod.rs
```diff
@@ -112,6 +130,7 @@ impl SizedBundle {
     }
 
     /// Consume self and return the underlying buffer of actions.
+    #[cfg(test)]
```
SuperFluffy, line 133: It's inadvisable to test code paths that only exists for tests - you are likely to check wrong behavior.

I think you can remove `into_actions` and, in the tests, just use `to_actions` instead.

crates/astria-composer/src/executor/bundle_factory/tests.rs
```diff
@@ -333,4 +333,26 @@ mod bundle_factory {
         assert_eq!(bundle_factory.finished.len(), 0);
         assert!(!bundle_factory.is_full());
     }
+
+    #[test]
+    fn unsigned_transaction_construction() {
```
SuperFluffy, line 338:
```suggestion
    fn transaction_construction_does_not_panic() {
```

crates/astria-core/src/protocol/transaction/v1alpha1/action.rs
```diff
@@ -169,6 +169,14 @@ impl Action {
         };
         Some(transfer_action)
     }
+
+    pub fn is_fee_asset_change(&self) -> bool {
```
SuperFluffy, line 173: Let's make a tracking issue to add unit tests for these.

Lilyjjo: created and commented in code: https://github.com/astriaorg/astria/issues/1593

crates/astria-composer/src/executor/bundle_factory/mod.rs
```diff
@@ -74,6 +75,23 @@ impl SizedBundle {
         }
     }
 
+    /// Constructs an [`UnsignedTransaction`] from the actions contained in the bundle and `params`.
+    // Method is expected to never panic because only `SequenceActions` are added to the bundle,
+    // which should produce a valid variant of the `ActionGroup` type.
+    #[allow(clippy::panic)]
```
SuperFluffy, line 81: Heads-up: with https://github.com/astriaorg/astria/pull/1561 being merged we switch everything to `expect` and require a `reason`:
```suggestion
#[expect(clippy::panic, reason = "method is expected to never panic because only `SequenceActions` are added to the bundle, which should produce a valid variant of the `ActionGroup` type; this is checked by `tests::transaction_construction_should_not_panic")]
```

crates/astria-core/src/protocol/transaction/v1alpha1/action_group/mod.rs
```diff
@@ -0,0 +1,196 @@
...
+        }
+    }
+}
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub(super) enum ActionGroup {
+    General,
+    UnbundleableGeneral,
+    Sudo,
+    UnbundleableSudo,
```
SuperFluffy, line 88: Bikeshed:

@noot left some comments on these where `Bundleable*` was removed and `Unbundleable*` introduced.

My preference would have been to go all out and use `General -> BundleableGeneral` and `Sudo -> BundleableSudo` (and keep `Unbundleable*`) so that this is always spelled out explicitly.

But I am going to leave the final call on this one to you.

crates/astria-core/src/protocol/transaction/v1alpha1/action_group/mod.rs
```diff
@@ -0,0 +1,196 @@
...
+
+    pub(super) fn is_sudo(self) -> bool {
+        matches!(self, ActionGroup::Sudo)
+    }
+}
+
+impl fmt::Display for ActionGroup {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            ActionGroup::General => write!(f, "general"),
```
SuperFluffy, line 104: Just leaving a note that should you go with my bikeshedding to rename prefix with `Bundleable*`, to also change the display name.

SuperFluffy (approved):
Thank you for going going through the many review comments, online and offline!

crates/astria-core/src/protocol/transaction/v1alpha1/action_group/mod.rs
Lilyjjo: Changed to BundleableXXX. Agree it's easier to understand

PR #1515, opened 2024-09-18
fix(sequencer)!: rewrite check_tx to be more efficient and fix regression

### LINKED ISSUE #1481: sequencer/mempool: CheckTX fails all rechecks
During the mempool rewrite CheckTX's functionality was accidentally changed to fail all `CheckTxKind::Recheck` calls. Previously attempting to re-insert a transaction into the mempool would not cause an error. 

This is a UX regression as CometBFT is used to display to users the possible statuses of their transactions. We should allow the `CheckTxKind::Recheck` calls to succeed until we have better ways to report to users the status of their transactions. Nice to have also is basic unit testing for CheckTX.

┆Issue Number: ENG-811

Lilyjjo:
## Summary
Previously `handle_check_tx()`'s would always run all stateless checks and attempt insertion. This PR modified the function to only run the checks on transactions that have not been removed or are not contained inside of the app side mempool.

This PR additionally adds a HashSet of transaction hashes that are contained in the app's Mempool to the state of the app Mempool. This is used to enable seeing if a transaction is contained in the Mempool without having to decode it into a `SignedTransaction`. This will be kept in sync with unit tests which check if the length of the mempool (which is now calculated with the HashSet) matches the expected state of mutating the underlying containers.

We considered using just a toggle on `CheckTX::New` and `CheckTX::Recheck` instead of the HashSet, but there is the possibility of CometBFT persisting transactions on restart which could make use of toggling on the app mempool knowing the transaction instead. TBD, but it's also useful to be able to query the mempool if it contains a transaction. 

This PR also restores the original behavior of `CheckTx` to not kick out transaction needlessly on rechecks, this regression was added when upgrading the mempool in #1323. Pre #1323, re-inserting the same transaction into the mempool would not cause an error. Post #1323, the re-insertion would cause an error. We now check to make sure that the transaction is not inside of the mempool before attempting to re-insert it. 

## Background
A lot of the functionality from the original `handle_check_tx()` has moved into the app side mempool itself, including the balance and nonce checks. All of the remaining checks will not change on `CheckTx::Recheck` and only need to be ran when considering a transaction for insertion into the appside mempool. 

## Testing
Ran locally and added unit tests to ensure future nonces can be added and rechecks do not remove transactions needlessly. 

## Metrics
- Deleted `check_tx_removed_stale_nonce`
- Added `check_tx_duration_seconds_check_tracked`, `mempool_tx_logic_error`
- Changed `check_tx_duration_seconds_check_nonce` to `check_tx_duration_seconds_fetch_nonce`

## Breaking Changes
Added new ABCI error codes for transaction insertion errors that would be useful to surface to the end user: `ALREADY_PRESENT`, `NONCE_TAKEN`, `ACCOUNT_SIZE_LIMIT`.

## Related Issues
closes #1481


crates/astria-sequencer/src/mempool/mod.rs
```diff
@@ -184,11 +180,18 @@ impl Mempool {
                 // Release the lock asap.
                 drop(pending);
                 // try to add to parked queue
-                parked.add(
-                    timemarked_tx,
+                match parked.add(
+                    timemarked_tx.clone(),
```
noot, line 184: i think you can do `let id = timemarked_tx.id()` before this, then use that later on in the function and remove the `clone()`

crates/astria-sequencer/src/service/mempool/mod.rs
```diff
@@ -120,50 +127,181 @@ async fn handle_check_tx<S: accounts::StateReadExt + address::StateReadExt + 'st
...
+    // perform stateless checks
+    let signed_tx = match stateless_checks(tx, &state, metrics).await {
+        StatelessReturn::Response(rsp) => return rsp,
+        StatelessReturn::Tx(signed_tx) => signed_tx,
+    };
+
+    // attempt to insert the transaction into the mempool
+    if let Some(rsp) = insert_into_mempool(mempool, &state, signed_tx, metrics).await {
+        return rsp;
+    }
```
noot, line 155: this function returning an `Option` that's only set if there's an error response is somewhat confusing, would prefer for this function to return a `Result<(), response::CheckTx>` so the error variant is the error response

noot: same for `check_removed_comet_bft` and `check_tracked`

crates/astria-sequencer/src/service/mempool/mod.rs
```diff
@@ -120,50 +127,181 @@ async fn handle_check_tx<S: accounts::StateReadExt + address::StateReadExt + 'st
...
+/// Otherwise, it returns the [`SignedTransaction`] to be inserted into the mempool.
+enum StatelessReturn {
+    Response(response::CheckTx),
+    Tx(SignedTransaction),
+}
+
+/// Performs stateless checks on the transaction.
+///
+/// Returns a [`response::CheckTx`] if the transaction fails any of the checks.
+/// Otherwise, it returns the [`SignedTransaction`] to be inserted into the mempool.
```
noot, line 260: you can use a `Result<SignedTransaction, response::CheckTx>` as this function's return

noot:
I think this PR looks good, just a few smaller comments!

 > This PR additionally adds a HashSet of transaction hashes that are contained in the app's Mempool to the state of the app Mempool. This is used in the scenario where the sequencer restarts-- CometBFT persists transactions but our app mempool does not. The cache is used to allow transactions to re-enter the mempool on restarts.

how does this work? with cometbft recheck?

crates/astria-sequencer/src/service/mempool/mod.rs
Lilyjjo: Changed it, feels much better that way

Lilyjjo:
> > This PR additionally adds a HashSet of transaction hashes that are contained in the app's Mempool to the state of the app Mempool. This is used in the scenario where the sequencer restarts-- CometBFT persists transactions but our app mempool does not. The cache is used to allow transactions to re-enter the mempool on restarts.
> 
> how does this work? with cometbft recheck?

Hmmm. I think I misheard why we wanted to allow rechecks to add to the appside mempool. I checked out the CometBFT code and didn't see it persisting the transactions anywhere that was accessible, (I wonder if I missed it). In the scenario where CometBFT doesn't persist, then the allowing recheck would only be useful if CometBFT doesn't crash if the sequencer app crashses. I'll do some more research

crates/astria-sequencer/src/service/mempool/mod.rs
```diff
@@ -120,50 +127,165 @@ async fn handle_check_tx<S: accounts::StateReadExt + address::StateReadExt + 'st
...
+/// Checks if the transaction is already in the mempool.
+async fn is_tracked(tx_hash: [u8; 32], mempool: &AppMempool, metrics: &Metrics) -> bool {
+    let start_tracked_check = Instant::now();
+
+    let result = mempool.tracked(tx_hash).await;
+
+    let finished_check_tracked = Instant::now();
+    metrics.record_check_tx_duration_seconds_check_tracked(
+        finished_check_tracked.saturating_duration_since(start_tracked_check),
+    );
```
Fraser999, line 172:
```suggestion
    metrics.record_check_tx_duration_seconds_check_tracked(start_tracked_check.elapsed());
```

crates/astria-sequencer/src/service/mempool/mod.rs
```diff
@@ -120,50 +127,165 @@ async fn handle_check_tx<S: accounts::StateReadExt + address::StateReadExt + 'st
...
+
+    result
+}
+
+/// Checks if the transaction has been removed from the appside mempool.
+///
+/// Returns an Err([`response::CheckTx`]) with an error code and message if the transaction has been
```
Fraser999, line 179:
```suggestion
/// Returns an `Err(response::CheckTx)` with an error code and message if the transaction has been
```

crates/astria-sequencer/src/service/mempool/mod.rs
```diff
@@ -120,50 +127,165 @@ async fn handle_check_tx<S: accounts::StateReadExt + address::StateReadExt + 'st
...
+                    ..response::CheckTx::default()
+                });
+            }
+        }
+    };
+
+    let finished_removal_check = Instant::now();
+    metrics.record_check_tx_duration_seconds_check_removed(
+        finished_removal_check.saturating_duration_since(start_removal_check),
+    );
```
Fraser999, line 236:
```suggestion
    metrics.record_check_tx_duration_seconds_check_removed(start_removal_check.elapsed());
```

crates/astria-sequencer/src/service/mempool/mod.rs
```diff
@@ -174,133 +296,93 @@ async fn handle_check_tx<S: accounts::StateReadExt + address::StateReadExt + 'st
...
             ..response::CheckTx::default()
-        };
+        });
     }
 
     let finished_check_chain_id = Instant::now();
     metrics.record_check_tx_duration_seconds_check_chain_id(
-        finished_check_chain_id.saturating_duration_since(finished_check_nonce),
+        finished_check_chain_id.saturating_duration_since(finished_check_stateless),
     );
```
Fraser999, line 324: This could be
```rust
    metrics.record_check_tx_duration_seconds_check_chain_id(finished_check_stateless.elapsed());
```

crates/astria-sequencer/src/mempool/mod.rs
```diff
@@ -220,6 +223,10 @@ impl Mempool {
                         );
                     }
                 }
+
```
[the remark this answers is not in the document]
Lilyjjo: good catch. added similar logic to the error cases in maintenance too. i'll add a metric for these synchronization errors and ask Ido to setup an alert for it

crates/astria-sequencer/src/mempool/mod.rs
```diff
@@ -133,31 +135,54 @@ pub(crate) struct Mempool {
...
+    /// Adds a transaction to the mempool's tracked transactions.
+    /// Will increment logic error metrics and log error if transaction is already present.
+    fn add_to_contained_txs(&self, tx_hash: [u8; 32], contained_txs: &mut HashSet<[u8; 32]>) {
+        if !contained_txs.insert(tx_hash) {
+            self.metrics.increment_mempool_tx_logic_error();
+            error!(
+                "attempted to re-add transaction {tx_hash:?} to mempool's tracked container, is \
+                 logic error"
+            );
```
Fraser999, line 172:
```suggestion
            error!(
                tx_hash = %telemetry::display::hex(&tx_hash),
                "attempted to re-add transaction to mempool's tracked container, is logic error"
            );
```

crates/astria-sequencer/src/mempool/mod.rs
```diff
@@ -133,31 +135,54 @@ pub(crate) struct Mempool {
...
+
+    /// Removes a transaction from the mempool's tracked transactions.
+    /// Will increment logic error metrics and log error if transaction is not present.
+    fn remove_from_contained_txs(&self, tx_hash: [u8; 32], contained_txs: &mut HashSet<[u8; 32]>) {
+        if !contained_txs.remove(&tx_hash) {
+            self.metrics.increment_mempool_tx_logic_error();
+            error!(
+                "attempted to remove transaction non present {tx_hash:?} from mempool's tracked \
+                 container, is logic error"
+            );
```
Fraser999, line 184:
```suggestion
            error!(
                tx_hash = %telemetry::display::hex(&tx_hash),
                "attempted to remove transaction absent from mempool's tracked container, is logic \
                error"
            );
```

crates/astria-sequencer/src/mempool/mod.rs
```diff
@@ -211,15 +243,24 @@ impl Mempool {
...
                     if let Err(error) =
                         pending.add(ttx, current_account_nonce, &current_account_balances)
                     {
+                        // remove from tracked
+                        let mut contained_txs = self.contained_txs.write().await;
+                        self.remove_from_contained_txs(timemarked_tx.id(), &mut contained_txs);
                         error!(
                             current_account_nonce,
-                            "failed to promote transaction during insertion: {error:#}"
+                            "failed to promote transaction {tx_id:?} during insertion: {error:#}"
```
Fraser999, line 255: This should be more like:
```rust
                        error!(
                            current_account_nonce,
                            tx_hash = %telemetry::display::hex(&tx_id),
                            %error,
                            "failed to promote transaction during insertion"
                        );
```

crates/astria-sequencer/src/mempool/mod.rs
```diff
@@ -356,18 +406,29 @@ impl Mempool {
...
                     if let Err(error) = pending.add(tx, current_nonce, &current_balances) {
+                        // remove from tracked
+                        let mut contained_txs = self.contained_txs.write().await;
+                        self.remove_from_contained_txs(tx_id, &mut contained_txs);
+                        // this shouldn't happen
+                        self.metrics.increment_mempool_tx_logic_error();
                         error!(
                             current_nonce,
-                            "failed to promote transaction during maintenance: {error:#}"
+                            "failed to promote transaction {tx_id:?} during maintenance: {error:#}"
```
Fraser999, line 418: Similar change here would be good.

Fraser999 (approved):
Approved with the proviso that the new metric name is made more general so it can be reused across the full `astria-sequencer` codebase.

SuperFluffy (commented):
I will follow @Fraser999 approval, but I have made some comments here and there.

I want to especially call attention to the bit which is acquiring a write-lock in a loop: can you leave a comment there why that's acceptable?

SuperFluffy (changes requested):
I'm blocking this so I can have a second pass over this.

crates/astria-sequencer/src/mempool/mod.rs
```diff
@@ -211,15 +245,28 @@ impl Mempool {
                 );
                 // promote the transactions
                 for ttx in to_promote {
+                    let tx_id = ttx.id();
                     if let Err(error) =
                         pending.add(ttx, current_account_nonce, &current_account_balances)
                     {
+                        // remove from tracked
```
SuperFluffy, line 252: If code doing something as simple as this needs to be commented, then its either not self evident (leave a comment tagged with `XXX` or `NOTE`, see below), or adjust the called methods to make it evident.

SuperFluffy (commented):
I have gone over the PR and left some comments. Reading the PR description this patch attempts to tackle 3 things at the same time:

1. avoid rerunning expensive checks that have already been performed once. I believe this new functionality is covered by the changes in `crate::service::mempool`. It is enabled by the functions `check_removed_comet_bft` and `is_tracked`. The `is_tracked` makes use of the new `contained_txs` hashset. Question: is this an optimization to avoid using `pending` and `parked` mempool fields? Clearly the mempool needed to track the transactions contained therein before - so what's the reason to have an extra index of transactions? It would be nice to see that explained in the patch description. It would furthermore be useful to understand how this new index and the previous caches will be kept in sync.

2. fix a regression for `CheckTxKind::Recheck`. Grepping this PR I only find `recheck` mention in tests. Looking at the `fn handle_check_tx` function the `request::CheckTx` is immediately deconstructed and the `CheckTxKind::Recheck` never actively considered. It hence looks to me as if we don't discriminate between checks of new transactions and rechecked of transactions existing in the mempool. Given that this is a regression called out in both the patch title and text, it would be nice to have this explained.

3. introduce a way to maintain state on sequencer restarts. This part seems to be completely orthogonal to the goals 1. and 2. All related code should be removed from this patch and moved into a new PR. Having said that, it seems that the index of checked transactions (the hashset mentioned in the PR description) was introduced to enable the main goal of the PR (namely skipping unnecessary checks summarized under 1.). Enabling restart logic thus seems to be a useful side effect? Either way, it is confusing to the reviewer in how far this should be made part of the patch, and why it wasn't moved to a new patch. The patch description should be amended accordingly and better explained.

-----

The semantics of adding or removing transactions from the mempool-tracked txs is weird. Looking at the code that makes use of `add_from_contained_txs` and `remove_from_contained_txs` it looks like you would be better served by creating a custom guard like this (I have intentionally left out the metrics stuff to keep the example minimal; add them as necessary):

```rust
struct TxLock<'a> {
    mempool: &'a Mempool,
    txs: RwLockWriteGuard<'a, HashSet<[u8; 32]>>,
}

impl<'a> TxLock<'a> {
    fn add(&mut self, id: [u8; 32]) {
        self.tx.insert(id);
    }
    fn remove(&mut self, id: [u8; 32]) {
        self.tx.remove(id);
    }
}

impl Mempool {
    async fn lock_txs(&self) -> TxLock<'_> {
        TxLock {
            mempool: self,
            txs: mempool.contained_txs.write().await,
        }
    }
}
```

crates/astria-sequencer/src/service/mempool/mod.rs
```diff
@@ -119,50 +128,159 @@ async fn handle_check_tx<S: accounts::StateReadExt + address::StateReadExt + 'st
...
+
+/// Checks if the transaction has been removed from the appside mempool.
+///
+/// Returns an `Err(response::CheckTx)` with an error code and message if the transaction has been
+/// removed from the appside mempool.
+async fn check_removed_comet_bft(
+    tx_hash: [u8; 32],
+    mempool: &AppMempool,
+    metrics: &Metrics,
+) -> Result<(), response::CheckTx> {
```
SuperFluffy, line 183: You could just return a `RemovalReason` here and instead implement a new trait `IntoCheckTxResponse` on it so that you don't need to do the conversion inline.

crates/astria-sequencer/src/service/mempool/mod.rs
```diff
@@ -173,133 +291,91 @@ async fn handle_check_tx<S: accounts::StateReadExt + address::StateReadExt + 'st
...
+///
+/// Returns a `Err(response::CheckTx)` with an error code and message if the transaction fails
+/// insertion into the mempool.
+#[expect(clippy::too_many_lines, reason = "error handling is long")]
+async fn insert_into_mempool<S: accounts::StateReadExt + address::StateReadExt + 'static>(
+    mempool: &AppMempool,
+    state: &S,
+    signed_tx: SignedTransaction,
+    metrics: &'static Metrics,
+) -> Result<(), response::CheckTx> {
```
SuperFluffy, line 334: I have noted it for `RemovalReason`, but it also applies here: all the error-to-response conversions could be made a lot neater by adding a trait `IntoCheckTxResponse` and moving the conversions there instead of having them inlined in the function body.

crates/astria-sequencer/src/mempool/mod.rs
```diff
@@ -133,31 +135,56 @@ pub(crate) struct Mempool {
...
-            async { self.pending.read().await.len() },
-            async { self.parked.read().await.len() }
-        );
-        pending_len.saturating_add(parked_len)
+        self.contained_txs.read().await.len()
+    }
+
+    /// Adds a transaction to the mempool's tracked transactions.
+    /// Will increment logic error metrics and log error if transaction is already present.
+    fn add_to_contained_txs(&self, tx_hash: [u8; 32], contained_txs: &mut HashSet<[u8; 32]>) {
```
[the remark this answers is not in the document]
SuperFluffy: Yeah, I provided an alternative in my overall review. Should have called this out here.

I understand why you did this, but there are clearer ways to express this.

crates/astria-sequencer/src/service/mempool/mod.rs
```diff
@@ -18,6 +21,7 @@ use astria_core::{
     },
 };
 use astria_eyre::eyre::WrapErr as _;
+use bytes::Bytes;
```
SuperFluffy, line 24: Is this import used anywhere here?

crates/astria-sequencer/src/service/mempool/mod.rs
```diff
@@ -173,133 +291,91 @@ async fn handle_check_tx<S: accounts::StateReadExt + address::StateReadExt + 'st
...
+async fn insert_into_mempool<S: accounts::StateReadExt + address::StateReadExt + 'static>(
+    mempool: &AppMempool,
+    state: &S,
+    signed_tx: SignedTransaction,
+    metrics: &'static Metrics,
+) -> Result<(), response::CheckTx> {
+    let start_convert_address = Instant::now();
 
-    // tx is valid, push to mempool with current state
+    // generate address for the signed transaction
```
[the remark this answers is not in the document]
Lilyjjo: Tracked: https://github.com/astriaorg/astria/issues/1620

crates/astria-sequencer/src/service/mempool/mod.rs
Lilyjjo: Yes, it's used in the stateless_checks() function

crates/astria-sequencer/src/service/mempool/mod.rs
Lilyjjo: Nice. Wrote the trait and defined if for `RemovalReason` and `InsertionError`

crates/astria-sequencer/src/service/mempool/mod.rs
```diff
@@ -173,133 +291,91 @@ async fn handle_check_tx<S: accounts::StateReadExt + address::StateReadExt + 'st
...
-        finished_check_removed.saturating_duration_since(finished_check_chain_id),
-    );
+    Ok(signed_tx)
+}
+
+/// Attempts to insert the transaction into the mempool.
+///
+/// Returns a `Err(response::CheckTx)` with an error code and message if the transaction fails
+/// insertion into the mempool.
+#[expect(clippy::too_many_lines, reason = "error handling is long")]
```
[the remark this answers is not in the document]
Lilyjjo: Fixed it.

crates/astria-sequencer/src/service/mempool/mod.rs
SuperFluffy: Yeah, neat.

crates/astria-sequencer/src/transaction/checks.rs
```diff
@@ -25,32 +25,11 @@ use tracing::instrument;
...
 use crate::{
     accounts::StateReadExt as _,
-    address::StateReadExt as _,
     app::StateReadExt as _,
     bridge::StateReadExt as _,
     ibc::StateReadExt as _,
 };
 
-#[instrument(skip_all)]
-pub(crate) async fn check_nonce_mempool<S: StateRead>(
```
SuperFluffy, line 35: I'd like for all of these to migrate over to the mempool at some point.

SuperFluffy (approved):
This looks neat, thank you for your work.

I have made some notes on how you can now use your newly defined `ContainedTxLock` to avoid accidentally holding on to the lock for too long (which is an actual benefit of the design!).

Please also have a look at how composer (and bridge-withdrawer) might be affected by the nonce-related ABCI codes. Maybe create issues for followup tasks.

Happy to merge this after updating the PR description with respect to my last review.

crates/astria-sequencer/src/transaction/checks.rs
Lilyjjo: created an issue https://github.com/astriaorg/astria/issues/1632

crates/astria-core/src/protocol/abci.rs
```diff
@@ -52,6 +55,13 @@ impl AbciErrorCode {
             }
             Self::LOWER_NONCE_INVALIDATED => "lower nonce was invalidated in mempool".into(),
             Self::BAD_REQUEST => "the request payload was malformed".into(),
+            Self::ALREADY_PRESENT => "the transaction is already present in the mempool".into(),
+            Self::NONCE_TAKEN => "there is already a transaction with the same nonce for the \
```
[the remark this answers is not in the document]
Lilyjjo: Tracked issue: https://github.com/astriaorg/astria/issues/1633

PR #1519, opened 2024-09-18
feat(proto): add bundle and optimistic block apis

itamarreif:
## Summary
This adds the protobuf definitions for optimistic block stream and the bundle stream.

## Background
The Auctioneer will auction off a bundle slot which will be placed at the top of the rollup block deterministically during execution. The APIs in this PR are used to drive the bundle auction (detailed in [this document](https://www.notion.so/astria-org/20240919-Status-Update-1066bd31a90c80b68bb2ea9b898645d0)). At a high level:

The Auctioneer receives blocks optimistically from the sequencer in order to maximize the auction duration.

While the auction is running, the Auctioneer receives orders from the rollup node that have been validated against the optimistic block.

After a block is committed, the Auctioneer will submit the highest paying bundle for inclusion by the sequencer.

### Optimistic Block Stream
1. The Sequencer will stream blocks to the Auctioneer optimistically, i.e. before they are finalized (ProcessProposal).
2. The Sequencer will also stream block commitments to the Auctioneer, which mark blocks as final thanks to CometBFT's single slot finality. 

### Bundle Stream
1. The Auctioneer will stream optimistic blocks to the rollup node for execution. The rollup will execute the block and return the resulting header info.
2. The rollup node will stream bundles (indexed by parent block hash) to the Auctioneer.

## Changes
- Add the bundle service to the composer apis
- Add the optimistic block service to the sequencer block apis

## Breaking Changelist

## Related Issues
closes https://github.com/astriaorg/astria/issues/1553

### DIFF
--- a/proto/executionapis/astria/bundle/v1alpha1/bundle.proto
+++ b/proto/executionapis/astria/bundle/v1alpha1/bundle.proto
@@ -1,0 +1,67 @@
+syntax = "proto3";
+
+package astria.bundle.v1alpha1;
+
+import "astria/execution/v1alpha2/execution.proto";
+import "astria/sequencerblock/v1alpha1/block.proto";
+import "google/protobuf/timestamp.proto";
+
+// The "BaseBlock" is the information needed to simulate bundles on top of
+// a Sequencer block which may not have been committed yet.
+message BaseBlock {
+  // This is the block hash for the proposed block.
+  bytes sequencer_block_hash = 1;
+  // List of transactions to include in the new block.
+  repeated astria.sequencerblock.v1alpha1.RollupData transactions = 2;
+  // Timestamp to be used for new block.
+  google.protobuf.Timestamp timestamp = 3;
+}
+
+message StreamExecuteOptimisticBlockRequest {
+  BaseBlock block = 1;
+}
+
+message StreamExecuteOptimisticBlockResponse {
+  // Metadata identifying the block resulting from executing a block. Includes number, hash,
+  // parent hash and timestamp.
+  astria.execution.v1alpha2.Block block = 1;
+  // The base_sequencer_block_hash is the hash from the base sequencer block this block
+  // is based on. This is used to associate an optimistic execution result with the hash
+  // received once a sequencer block is committed.
+  bytes base_sequencer_block_hash = 2;
+}
+
+message StreamBundlesRequest {}
+
+// Information for the bundle submitter to know how to submit the bundle.
+// The fee and base_sequencer_block_hash are not necessarily strictly necessary
+// it allows for the case where the server doesn't always send the highest fee
+// bundles after the previous but could just stream any confirmed bundles.
+message Bundle {
+  // The fee that can be expected to be received for submitting this bundle.
+  // This allows the bundle producer to stream any confirmed bundles they would be ok
+  // with submitting. Used to avoid race conditions in received bundle packets. Could
+  // also be used by a bundle submitter to allow multiple entities to submit bundles.
+  uint64 fee = 1;
+  // The byte list of transactions to be included.
+  repeated bytes transactions = 2;
+  // The base_sequencer_block_hash is the hash from the base block this bundle
+  // is based on. This is used to verify that the bundle is based on the correct
+  // Sequencer block.
+  bytes base_sequencer_block_hash = 3;
+  // The hash of previous rollup block, on top of which the bundle will be executed as ToB.
+  bytes prev_rollup_block_hash = 4;
+}
+
+service OptimisticExecutionService {
+  // Stream blocks from the Auctioneer to Geth for optimistic execution. Geth will stream back
+  // metadata from the executed blocks.
+  rpc StreamExecuteOptimisticBlock(stream StreamExecuteOptimisticBlockRequest) returns (stream StreamExecuteOptimisticBlockResponse);
+}
+
+service BundleService {
+  // A bundle submitter requests bundles given a new optimistic Sequencer block,
+  // and receives a stream of potential bundles for submission, until either a timeout
+  // or the connection is closed by the client.
+  rpc StreamBundles(StreamBundlesRequest) returns (stream Bundle);
+}
--- a/proto/executionapis/astria/bundle/v1alpha1/bundle.proto
+++ b/proto/executionapis/astria/bundle/v1alpha1/bundle.proto
@@ -17,11 +17,11 @@ message BaseBlock {
   google.protobuf.Timestamp timestamp = 3;
 }
 
-message StreamExecuteOptimisticBlockRequest {
-  BaseBlock block = 1;
+message ExecuteOptimisticBlockStreamRequest {
+  BaseBlock base_block = 1;
 }
 
-message StreamExecuteOptimisticBlockResponse {
+message ExecuteOptimisticBlockStreamResponse {
   // Metadata identifying the block resulting from executing a block. Includes number, hash,
   // parent hash and timestamp.
   astria.execution.v1alpha2.Block block = 1;
@@ -31,7 +31,7 @@ message StreamExecuteOptimisticBlockResponse {
   bytes base_sequencer_block_hash = 2;
 }
 
-message StreamBundlesRequest {}
+message GetBundleStreamRequest {}
 
 // Information for the bundle submitter to know how to submit the bundle.
 // The fee and base_sequencer_block_hash are not necessarily strictly necessary
@@ -53,15 +53,19 @@ message Bundle {
   bytes prev_rollup_block_hash = 4;
 }
 
+message GetBundleStreamResponse {
+  Bundle bundle = 1;
+}
+
 service OptimisticExecutionService {
   // Stream blocks from the Auctioneer to Geth for optimistic execution. Geth will stream back
   // metadata from the executed blocks.
-  rpc StreamExecuteOptimisticBlock(stream StreamExecuteOptimisticBlockRequest) returns (stream StreamExecuteOptimisticBlockResponse);
+  rpc ExecuteOptimisticBlockStream(stream ExecuteOptimisticBlockStreamRequest) returns (stream ExecuteOptimisticBlockStreamResponse);
 }
 
 service BundleService {
   // A bundle submitter requests bundles given a new optimistic Sequencer block,
   // and receives a stream of potential bundles for submission, until either a timeout
   // or the connection is closed by the client.
-  rpc StreamBundles(StreamBundlesRequest) returns (stream Bundle);
+  rpc GetBundleStream(GetBundleStreamRequest) returns (stream GetBundleStreamResponse);
 }
--- a/proto/executionapis/astria/bundle/v1alpha1/bundle.proto
+++ b/proto/executionapis/astria/bundle/v1alpha1/bundle.proto
@@ -2,35 +2,6 @@ syntax = "proto3";
 
 package astria.bundle.v1alpha1;
 
-import "astria/execution/v1alpha2/execution.proto";
-import "astria/sequencerblock/v1alpha1/block.proto";
-import "google/protobuf/timestamp.proto";
-
-// The "BaseBlock" is the information needed to simulate bundles on top of
-// a Sequencer block which may not have been committed yet.
-message BaseBlock {
-  // This is the block hash for the proposed block.
-  bytes sequencer_block_hash = 1;
-  // List of transactions to include in the new block.
-  repeated astria.sequencerblock.v1alpha1.RollupData transactions = 2;
-  // Timestamp to be used for new block.
-  google.protobuf.Timestamp timestamp = 3;
-}
-
-message ExecuteOptimisticBlockStreamRequest {
-  BaseBlock base_block = 1;
-}
-
-message ExecuteOptimisticBlockStreamResponse {
-  // Metadata identifying the block resulting from executing a block. Includes number, hash,
-  // parent hash and timestamp.
-  astria.execution.v1alpha2.Block block = 1;
-  // The base_sequencer_block_hash is the hash from the base sequencer block this block
-  // is based on. This is used to associate an optimistic execution result with the hash
-  // received once a sequencer block is committed.
-  bytes base_sequencer_block_hash = 2;
-}
-
 message GetBundleStreamRequest {}
 
 // Information for the bundle submitter to know how to submit the bundle.
@@ -55,12 +26,6 @@ message Bundle {
 
 message GetBundleStreamResponse {
   Bundle bundle = 1;
-}
-
-service OptimisticExecutionService {
-  // Stream blocks from the Auctioneer to Geth for optimistic execution. Geth will stream back
-  // metadata from the executed blocks.
-  rpc ExecuteOptimisticBlockStream(stream ExecuteOptimisticBlockStreamRequest) returns (stream ExecuteOptimisticBlockStreamResponse);
 }
 
 service BundleService {
--- a/proto/sequencerblockapis/astria/sequencerblock/v1alpha1/optimistic_block.proto
+++ b/proto/sequencerblockapis/astria/sequencerblock/v1alpha1/optimistic_block.proto
@@ -1,0 +1,35 @@
+syntax = "proto3";
+
+package astria.sequencerblock.v1alpha1;
+
+import "astria/primitive/v1/types.proto";
+import "astria/sequencerblock/v1alpha1/block.proto";
+
+message StreamBlockCommitmentsRequest {}
+
+// Identifying metadata for blocks that have been successfully committed in the Sequencer.
+message SequencerBlockCommit {
+  // Height of the sequencer block that was committed.
+  uint64 height = 1;
+  // Hash of the sequencer block that was committed.
+  bytes block_hash = 2;
+}
+
+message StreamOptimisticBlockRequest {
+  // The rollup id for which the Sequencer block is being streamed.
+  astria.primitive.v1.RollupId rollup_id = 1;
+}
+
+message StreamOptimisticBlockResponse {
+  // The optimistic Sequencer block that is being streamed, filtered for the provided rollup id.
+  astria.sequencerblock.v1alpha1.FilteredSequencerBlock block = 1;
+}
+
+// The Sequencer will serve this to the aucitoneer
+service OptimisticBlockService {
+  // The Sequencer will stream the optimistic Sequencer block (filtered for the provided
+  // rollup id) to the Auctioneer.
+  rpc StreamOptimisticBlock(StreamOptimisticBlockRequest) returns (stream StreamOptimisticBlockResponse);
+  // The Sequencer will stream the block commits to the Auctioneer.
+  rpc StreamBlockCommitments(StreamBlockCommitmentsRequest) returns (stream SequencerBlockCommit);
+}
--- a/proto/sequencerblockapis/astria/sequencerblock/v1alpha1/optimistic_block.proto
+++ b/proto/sequencerblockapis/astria/sequencerblock/v1alpha1/optimistic_block.proto
@@ -5,7 +5,7 @@ package astria.sequencerblock.v1alpha1;
 import "astria/primitive/v1/types.proto";
 import "astria/sequencerblock/v1alpha1/block.proto";
 
-message StreamBlockCommitmentsRequest {}
+message GetBlockCommitmentStreamRequest {}
 
 // Identifying metadata for blocks that have been successfully committed in the Sequencer.
 message SequencerBlockCommit {
@@ -15,12 +15,16 @@ message SequencerBlockCommit {
   bytes block_hash = 2;
 }
 
-message StreamOptimisticBlockRequest {
+message GetBlockCommitmentStreamResponse {
+  SequencerBlockCommit commitment = 1;
+}
+
+message GetOptimisticBlockStreamRequest {
   // The rollup id for which the Sequencer block is being streamed.
   astria.primitive.v1.RollupId rollup_id = 1;
 }
 
-message StreamOptimisticBlockResponse {
+message GetOptimisticBlockStreamResponse {
   // The optimistic Sequencer block that is being streamed, filtered for the provided rollup id.
   astria.sequencerblock.v1alpha1.FilteredSequencerBlock block = 1;
 }
@@ -29,7 +33,7 @@ message StreamOptimisticBlockResponse {
 service OptimisticBlockService {
   // The Sequencer will stream the optimistic Sequencer block (filtered for the provided
   // rollup id) to the Auctioneer.
-  rpc StreamOptimisticBlock(StreamOptimisticBlockRequest) returns (stream StreamOptimisticBlockResponse);
+  rpc GetOptimisticBlockStream(GetOptimisticBlockStreamRequest) returns (stream GetOptimisticBlockStreamResponse);
   // The Sequencer will stream the block commits to the Auctioneer.
-  rpc StreamBlockCommitments(StreamBlockCommitmentsRequest) returns (stream SequencerBlockCommit);
+  rpc GetBlockCommitmentStream(GetBlockCommitmentStreamRequest) returns (stream GetBlockCommitmentStreamResponse);
 }
--- a/crates/astria-core/src/generated/astria.bundle.v1alpha1.rs
+++ b/crates/astria-core/src/generated/astria.bundle.v1alpha1.rs
@@ -1,3 +1,373 @@
+#[allow(clippy::derive_partial_eq_without_eq)]
+#[derive(Clone, PartialEq, ::prost::Message)]
+pub struct GetBundleStreamRequest {}
+impl ::prost::Name for GetBundleStreamRequest {
+    const NAME: &'static str = "GetBundleStreamRequest";
+    const PACKAGE: &'static str = "astria.bundle.v1alpha1";
+    fn full_name() -> ::prost::alloc::string::String {
+        ::prost::alloc::format!("astria.bundle.v1alpha1.{}", Self::NAME)
+    }
+}
+/// Information for the bundle submitter to know how to submit the bundle.
+/// The fee and base_sequencer_block_hash are not necessarily strictly necessary
+/// it allows for the case where the server doesn't always send the highest fee
+/// bundles after the previous but could just stream any confirmed bundles.
+#[allow(clippy::derive_partial_eq_without_eq)]
+#[derive(Clone, PartialEq, ::prost::Message)]
+pub struct Bundle {
+    /// The fee that can be expected to be received for submitting this bundle.
+    /// This allows the bundle producer to stream any confirmed bundles they would be ok
+    /// with submitting. Used to avoid race conditions in received bundle packets. Could
+    /// also be used by a bundle submitter to allow multiple entities to submit bundles.
+    #[prost(uint64, tag = "1")]
+    pub fee: u64,
+    /// The byte list of transactions to be included.
+    #[prost(bytes = "bytes", repeated, tag = "2")]
+    pub transactions: ::prost::alloc::vec::Vec<::prost::bytes::Bytes>,
+    /// The base_sequencer_block_hash is the hash from the base block this bundle
+    /// is based on. This is used to verify that the bundle is based on the correct
+    /// Sequencer block.
+    #[prost(bytes = "bytes", tag = "3")]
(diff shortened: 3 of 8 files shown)


proto/composerapis/astria/auctioneer/v1alpha1/optimistic_block.proto
joroshiba, line 1: I think we should include this in sequencerblockapis instead of composer apis

proto/composerapis/astria/auctioneer/v1alpha1/optimistic_block.proto
```diff
@@ -0,0 +1,39 @@
...
+    uint64 height = 1;
+    // Hash of the sequencer block that was committed.
+    bytes block_hash = 2;
+}
+
+// The Sequencer will serve this to the aucitoneer
+service OptimisticBlockService {
+    // The Sequencer will stream the optimistic Sequencer block (filtered for the provided
+    // rollup id) to the Auctioneer.
+    rpc StreamOptimisticBlock(astria.primitive.v1.RollupId) returns (stream astria.sequencerblock.v1alpha1.FilteredSequencerBlock);
```
joroshiba, line 36: nit: lets make a `message StreamOptimisticBlockRequest` which contains the rollup id instead of having it be the actual request object. I know this is what I did in my gist, but having the message wrapper provides flexibility for adding future parameters.

Similar on the response

proto/sequencerblockapis/astria/sequencerblock/v1alpha1/optimistic_block.proto
```diff
@@ -0,0 +1,49 @@
...
+import "astria/sequencerblock/v1alpha1/block.proto";
+import "astria/primitive/v1/types.proto";
+
+// The "BaseBlock" is the information needed to simulate bundles on top of
+// a Sequencer block which may not have been committed yet.
+message BaseBlock {
+    // This is the block hash for the proposed block.
+    bytes sequencer_block_hash = 1;
+    // The hash of previous rollup block, which new block will be created on top of.
+    bytes prev_rollup_block_hash = 3;
```
bharath-123, line 17: Can we assume that this will be the latest soft block of the rollup? Because I believe we would have to query the soft block from auctioneer via the `GetCommitmentState` method which might add to latency.

proto/composerapis/astria/auctioneer/v1alpha1/bundle.proto
```diff
@@ -0,0 +1,39 @@
...
+    // also be used by a bundle submitter to allow multiple entities to submit bundles.
+    uint64 fee = 1;
+    // The byte list of transactions to be included.
+    repeated bytes transactions = 2;
+    // The base_sequencer_block_hash is the hash from the base block this bundle
+    // is based on. This is used to verify that the bundle is based on the correct
+    // Sequencer block.
+    bytes base_sequencer_block_hash = 3;
+    // The hash of previous rollup block, on top of which the bundle will be executed as ToB.
+    bytes prev_rollup_block_hash = 4;
```
bharath-123, line 27: I am not sure what Auctioneer would do with this information

proto/composerapis/astria/auctioneer/v1alpha1/bundle.proto
```diff
@@ -0,0 +1,39 @@
...
+// Information for the bundle submitter to know how to submit the bundle.
+// The fee and base_sequencer_block_hash are not necessarily strictly necessary
+// it allows for the case where the server doesn't always send the highest fee
+// bundles after the previous but could just stream any confirmed bundles.
+message Bundle {
+    // The fee that can be expected to be received for submitting this bundle.
+    // This allows the bundle producer to stream any confirmed bundles they would be ok
+    // with submitting. Used to avoid race conditions in received bundle packets. Could
+    // also be used by a bundle submitter to allow multiple entities to submit bundles.
+    uint64 fee = 1;
```
noot, line 19: how is this verified? also what token is this in (rollup or sequencer)? i'm guessing rollup?

proto/composerapis/astria/auctioneer/v1alpha1/bundle.proto
```diff
@@ -0,0 +1,39 @@
...
+    bytes base_sequencer_block_hash = 3;
+    // The hash of previous rollup block, on top of which the bundle will be executed as ToB.
+    bytes prev_rollup_block_hash = 4;
+}
+
+// Geth will serve this to the Auctioneer
+service BundleService {
+    // Stream blocks from the Auctioneer to Geth for optimistic execution. Geth will stream back
+    // metadata from the executed blocks.
+    rpc ExecuteOptimisticBlocks(stream BaseBlock) returns (stream astria.execution.v1alpha2.Block);
```
noot, line 34:
```suggestion
    rpc ExecuteOptimisticBlock(stream BaseBlock) returns (stream astria.execution.v1alpha2.Block);
```

however i'm not sure why this is needed when we already have an `ExecuteBlock` method - they seem to be taking the same inputs and doing the same thing (just executing a block and returning it)?

proto/composerapis/astria/auctioneer/v1alpha1/bundle.proto
```diff
@@ -0,0 +1,39 @@
...
+    // This allows the bundle producer to stream any confirmed bundles they would be ok
+    // with submitting. Used to avoid race conditions in received bundle packets. Could
+    // also be used by a bundle submitter to allow multiple entities to submit bundles.
+    uint64 fee = 1;
+    // The byte list of transactions to be included.
+    repeated bytes transactions = 2;
+    // The base_sequencer_block_hash is the hash from the base block this bundle
+    // is based on. This is used to verify that the bundle is based on the correct
+    // Sequencer block.
+    bytes base_sequencer_block_hash = 3;
```
noot, line 25: who's verifying it?

proto/sequencerblockapis/astria/sequencerblock/v1alpha1/optimistic_block.proto
```diff
@@ -0,0 +1,49 @@
...
+import "google/protobuf/empty.proto";
+import "astria/execution/v1alpha2/execution.proto";
+import "astria/sequencerblock/v1alpha1/block.proto";
+import "astria/primitive/v1/types.proto";
+
+// The "BaseBlock" is the information needed to simulate bundles on top of
+// a Sequencer block which may not have been committed yet.
+message BaseBlock {
+    // This is the block hash for the proposed block.
+    bytes sequencer_block_hash = 1;
```
noot, line 15: why is this needed? this type is only passed into geth, is it just putting it in the `Bundle`s?

proto/sequencerblockapis/astria/sequencerblock/v1alpha1/optimistic_block.proto
```diff
@@ -0,0 +1,49 @@
...
+    // The hash of previous rollup block, which new block will be created on top of.
+    bytes prev_rollup_block_hash = 3;
+    // List of transactions to include in the new block.
+    repeated astria.sequencerblock.v1alpha1.RollupData transactions = 4;
+    // Timestamp to be used for new block.
+    google.protobuf.Timestamp timestamp = 5;
```
noot, line 21:
```suggestion
    bytes prev_rollup_block_hash = 2;
    // List of transactions to include in the new block.
    repeated astria.sequencerblock.v1alpha1.RollupData transactions = 3;
    // Timestamp to be used for new block.
    google.protobuf.Timestamp timestamp = 4;
```

proto/sequencerblockapis/astria/sequencerblock/v1alpha1/optimistic_block.proto
```diff
@@ -0,0 +1,49 @@
...
+}
+
+// Identifying metadata for blocks that have been successfully committed in the Sequencer.
+message BlockCommit {
```
noot, line 25:
```suggestion
message SequencerBlockCommit {
```

proto/sequencerblockapis/astria/sequencerblock/v1alpha1/optimistic_block.proto
```diff
@@ -0,0 +1,49 @@
...
+    // Height of the sequencer block that was committed.
+    uint64 height = 1;
+    // Hash of the sequencer block that was committed.
+    bytes block_hash = 2;
+}
+
+message StreamOptimisticBlockRequest {
+    // The rollup id for which the Sequencer block is being streamed.
+    astria.primitive.v1.RollupId rollup_id = 1;
```
noot, line 34: should this potentially be an array of rollup ids (for potential multi-rollup bundling)?

proto/composerapis/astria/auctioneer/v1alpha1/bundle.proto
bharath-123: this would be in rollup token and since we are running the auctioneer, we should be able to assume that this information passed between geth and the composer can be trusted, similar to how we can assume the information between a geth node and a conductor can be trusted.

bharath-123: the fee will mainly be used to maintain the highest paying transaction seen so far in the auctioneer.

proto/composerapis/astria/auctioneer/v1alpha1/bundle.proto
bharath-123: Its not just doing the same thing as conductor. When geth receives blocks from `ExecuteOptimisticBlock`, it does the following:
1. Execute the block and update the optimistically synced fork
2. Update the legacy mempool to use the optimistic block to perform stateful checks on
3. Clear the mempool of transactions to allow searchers to send transactions valid on a new block.

(2) and (3) will be done with events

It is also a good idea to separate it out from ExecuteBlock for the following reasons:
1. We are using a stream here instead of a unary RPC for efficiency 
2. ExecuteBlock holds a lock when executing blocks. it might be better to separate out the calls and let only conductor only interact with ExecuteBlock to avoid any unnecessary contentions. 
3. Conductor is already an audited path, changes in conductor would involve us to re-audit it.

proto/sequencerblockapis/astria/sequencerblock/v1alpha1/optimistic_block.proto
bharath-123: In my mind, auctioneer is very specific only to flame so i suppose having a single rollup id should be fine.

proto/composerapis/astria/auctioneer/v1alpha1/bundle.proto
bharath-123: another way would be for auctioneer to de-serialize the EVM transaction which would be more trustless but would be higher latency.

proto/composerapis/astria/auctioneer/v1alpha1/bundle.proto
bharath-123: @itamarreif i believe here, auctioneer could use the rollup block hash instead of sequencer block hash? cause if this is the case, rollup would have to maintain the sequencer_block_hash?
I believe auctioneer would anyways receive via the `ExecuteOptimisticBlock`

proto/composerapis/astria/auctioneer/v1alpha1/bundle.proto
bharath-123: Auctioneer would require it to build the bundle to send to the sequencer. When conductor is applying this bundle, we need to make sure that we are applying on the correct parent block.

proto/composerapis/astria/auctioneer/v1alpha1/bundle.proto
noot: > this would be in rollup token and since we are running the auctioneer, we should be able to assume that this information passed between geth and the composer can be trusted, similar to how we can assume the information between a geth node and a conductor can be trusted.

you mean passed between geth and auctioneer? so the fee is parsed by geth from the txs and verified in geth, then set to auctioneer essentially?

proto/composerapis/astria/auctioneer/v1alpha1/bundle.proto
noot: >  the optimistically synced fork

does an auctioneer-geth have two separate forks then? an optimistic fork and a fork from conductor blocks? i'm not sure this is functionally different from the existing head-soft-firm block structure in geth, optimistic can be head, soft and firm as the same as now

does `ExecuteOptimisticBlock` atomically update the block to be head as well then? if so, it'll need a lock, as `SetCommitmentState` also updates the head of chain

proto/sequencerblockapis/astria/sequencerblock/v1alpha1/optimistic_block.proto
noot: yeah i guess this needs a bigger product discussion, like do we ever see this being used for not-flame or cross-rollup in general

proto/composerapis/astria/auctioneer/v1alpha1/bundle.proto
itamarreif: yeah.

transactions are submitted to geth via the mempool rpc. when the geth mempool promotes transactions to pending they are checked for nonce and sufficient balance for the fee. geth provides a subscription for pending transactions (similar to what we do in composer) where promoted txs are sent to.
the geth-side bundle service will reap transactions from the pending transactions subscription (this is a `Subscription` object internal to geth that the websocket subscription also uses), and wrap them in this `Bundle` struct which it will then stream to the auctioneer.

because we plan to clear the mempool on geth after every optimistic block execution, the idea is that bundles received by the auctioneer via this stream after the optimistic execution result will have been checked against the optimistic state by geth's mempool's promotion to pending.

the auctioneer then has three streams, one for optimistic block execution results, one for block commitments and one for bundles. more on this in this part of the doc: https://www.notion.so/astria-org/20240919-Status-Update-1066bd31a90c80b68bb2ea9b898645d0?pvs=4#1066bd31a90c80f193e3d6cb72a2f6e9

proto/composerapis/astria/auctioneer/v1alpha1/bundle.proto
itamarreif: I'm unsure this is needed here. the bundles will be mapped to an auction based on the rollup parent hash. the rollup ends up reading these bundle objects in the read path, and it currently is not aware of any sequencer-level hashes.

It might make sense to return the sequencer block hash alongside the optimistic execution result tho. i.e.
```protobuf
message OptimisticExecutionResult {
    bytes sequencer_block_hash = 1;
    astria.execution.v1alpha2.Block execution_result = 2;
}

service BundleService {
    rpc ExecuteOptimisticBlocks(stream BaseBlock) returns (stream OptimisticExecutionResult);
}
```

proto/sequencerblockapis/astria/sequencerblock/v1alpha1/optimistic_block.proto
itamarreif: I think this should probably be returned from the optimistic execution result stream, see above comment

proto/sequencerblockapis/astria/sequencerblock/v1alpha1/optimistic_block.proto
itamarreif: I think it definitely could make sense to do this for multiple rollups but there's not much use for it right now so can be fine to keep as a single rollup id for `v1alpha1`

proto/sequencerblockapis/astria/sequencerblock/v1alpha1/optimistic_block.proto
itamarreif: we can probably remove this? what is it used for

proto/sequencerblockapis/astria/sequencerblock/v1alpha1/optimistic_block.proto
```diff
@@ -0,0 +1,50 @@
...
+package astria.sequencerblock.v1alpha1;
+
+import "astria/primitive/v1/types.proto";
+import "astria/sequencerblock/v1alpha1/block.proto";
+import "google/protobuf/empty.proto";
+import "google/protobuf/timestamp.proto";
+
+// The "BaseBlock" is the information needed to simulate bundles on top of
+// a Sequencer block which may not have been committed yet.
+message BaseBlock {
```
noot, line 12: this should be in the bundle.proto file

proto/composerapis/astria/auctioneer/v1alpha1/bundle.proto
```diff
@@ -0,0 +1,50 @@
...
+message StreamExecuteOptimisticBlockRequest {
+  astria.sequencerblock.v1alpha1.BaseBlock base_block = 1;
+}
+
+message StreamExecuteOptimisticBlockResponse {
+  astria.execution.v1alpha2.Block block = 1;
+}
+
+message StreamBundlesRequest {
+  google.protobuf.Empty empty = 1;
```
joroshiba, line 38: I don't think with the message you are required to have empty take up a message field can just use an empty message

proto/composerapis/astria/auctioneer/v1alpha1/bundle.proto
```diff
@@ -0,0 +1,50 @@
...
+  bytes base_sequencer_block_hash = 3;
+  // The hash of previous rollup block, on top of which the bundle will be executed as ToB.
+  bytes prev_rollup_block_hash = 4;
+}
+
+message StreamExecuteOptimisticBlockRequest {
+  astria.sequencerblock.v1alpha1.BaseBlock base_block = 1;
+}
+
+message StreamExecuteOptimisticBlockResponse {
```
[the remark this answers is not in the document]
joroshiba: This would be a nice to have feature I think

proto/sequencerblockapis/astria/sequencerblock/v1alpha1/optimistic_block.proto
joroshiba: Agreed on should be in the bundle.proto, open question of should we use the `ExecuteBlockRequest` shape from execution API?

proto/composerapis/astria/auctioneer/v1alpha1/bundle.proto
joroshiba, line 1: I think this should be a part of the excutionapis, as it's a service implemented on an execution client.

SequencerBlock apis are implemented on sequencer, execution on execution client, composer on composer. Composer (well auctioneer?) is the client not the server in this instance.

proto/composerapis/astria/auctioneer/v1alpha1/bundle.proto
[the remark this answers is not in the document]
itamarreif: added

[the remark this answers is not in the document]
itamarreif: added

itamarreif:
> Agreed on should be in the bundle.proto, open question of should we use the ExecuteBlockRequest shape from execution API?

the bundle api's `BaseBlock` is the same as the execution api's `ExecuteBlockRequest` except for the fact that it contains the sequencer block hash rather than the previous rollup hash.

the reason for having the sequencer block hash is that the BaseBlock is derived from a FilteredSequencerBlock  (thus has the corresponding sequencer hash), and we want to link the execution result to a sequencer block commitment that will received later (so we provide the sequencer hash to geth's bundle service).

the reasons we don't have the previous rollup block hash in the auctioneer are:
1. getting it will require calling GetCommitmentState  which will incur additional latency
2. we dont care about the previous rollup block hash in the auctioneer, just about the block hash that results from the optimistic block execution (i.e. the rollup block hash sent in response)

proto/composerapis/astria/auctioneer/v1alpha1/bundle.proto
```diff
@@ -0,0 +1,50 @@
...
+// Geth will serve this to the Auctioneer
+service BundleService {
+  // Stream blocks from the Auctioneer to Geth for optimistic execution. Geth will stream back
+  // metadata from the executed blocks.
+  rpc StreamExecuteOptimisticBlock(stream StreamExecuteOptimisticBlockRequest) returns (stream StreamExecuteOptimisticBlockResponse);
+  // A bundle submitter requests bundles given a new optimistic Sequencer block,
+  // and receives a stream of potential bundles for submission, until either a timeout
+  // or the connection is closed by the client.
+  rpc StreamBundles(StreamBundlesRequest) returns (stream Bundle);
+}
```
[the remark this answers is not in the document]
itamarreif: Separating the two would make sense to me. Thoughts on this?"
```protobuf
service OptimisticExecutionService {
  rpc StreamExecuteBlock(stream StreamExecuteBlockRequest) returns (stream StreamExecuteBlockResponse);
}

service BundleService {
  rpc StreamBundles(StreamBundlesRequest) returns stream (Bundle);
}
```

or should the endpoint be `StreamExecuteOptimisticBlock`?

proto/sequencerblockapis/astria/sequencerblock/v1alpha1/optimistic_block.proto
itamarreif: > Agreed on should be in the bundle.proto, open question of should we use the ExecuteBlockRequest shape from execution API?

The bundle api's `BaseBlock` is the same as the execution api's `ExecuteBlockRequest` except for the fact that it contains the sequencer block hash rather than the previous rollup hash.

The reason for having the sequencer block hash is that the BaseBlock is derived from a FilteredSequencerBlock  (thus has the corresponding sequencer hash), and we want to link the execution result to a sequencer block commitment that will received later (so we provide the sequencer hash to geth's bundle service).

The reasons we don't have the previous rollup block hash in the auctioneer are:
1. Getting it will require calling GetCommitmentState  which will incur additional latency
2. We dont actually care about the previous rollup block hash in the auctioneer, just about the block hash that results from the optimistic block execution (i.e. the rollup block hash sent in response)

proto/composerapis/astria/auctioneer/v1alpha1/bundle.proto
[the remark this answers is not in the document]
noot: I like the separation of the two you suggested

proto/composerapis/astria/auctioneer/v1alpha1/bundle.proto
bharath-123: yes so currently, we are maintaining 2 separate forks. An optimistic fork and a fork from conductor blocks. We are not updating the block to be head in geth as per the current impl. 
We maintain a separate block pointer for the optimistic block as per: https://github.com/astriaorg/flame/blob/bharath/add-mempool-clearing-event/core/blockchain.go#L242
This is updated in `ExecuteOptimisticBlock` method as per https://github.com/astriaorg/flame/blob/bharath/add-mempool-clearing-event/grpc/execution/server.go#L302 (I haven't hooked it up with the grpc streams interface yet, so this name will change)
Essentially, we do the following:
1. We insert the optimistic block into the chain without setting the head
2. We set the optimistic block pointer to this block header to reference the state in the future

bharath-123: my thinking is that, by this method we don't need to hold the lock held in `ExecuteBlock` and `UpdateStateCommitment` because the pieces of data being operated on are different and do not interfere.

proto/composerapis/astria/auctioneer/v1alpha1/bundle.proto
bharath-123: we can make geth send this, if this helps with auctioneer

proto/sequencerblockapis/astria/sequencerblock/v1alpha1/optimistic_block.proto
bharath-123: yeah agreed with itamar. We don't want to call `GetCommitmentState` in auctioneer as it ll add to latency. I am a little unsure about requiring `sequencer_block_hash` in the geth node. I think auctioneer could maintain the session by itself

itamarreif: The auctioneer will be managing this state regardless in order to drive the auction.

Without adding this into the requests and responses we rely on both in-order arrival of messages over the stream and the no delays/drops that could fuck things up. Feels easier to just have this as part of the request/response, but there's no need to keep track of it inside of geth since the request should be in scope when u make the response?

proto/sequencerblockapis/astria/sequencerblock/v1alpha1/optimistic_block.proto
```diff
@@ -0,0 +1,35 @@
...
+  astria.sequencerblock.v1alpha1.FilteredSequencerBlock block = 1;
+}
+
+// The Sequencer will serve this to the aucitoneer
+service OptimisticBlockService {
+  // The Sequencer will stream the optimistic Sequencer block (filtered for the provided
+  // rollup id) to the Auctioneer.
+  rpc StreamOptimisticBlock(StreamOptimisticBlockRequest) returns (stream StreamOptimisticBlockResponse);
+  // The Sequencer will stream the block commits to the Auctioneer.
+  rpc StreamBlockCommitments(StreamBlockCommitmentsRequest) returns (stream SequencerBlockCommit);
```
joroshiba, line 34: nit: maybe `GetBlockCommitmentStream`?

proto/sequencerblockapis/astria/sequencerblock/v1alpha1/optimistic_block.proto
```diff
@@ -0,0 +1,35 @@
...
+message StreamOptimisticBlockResponse {
+  // The optimistic Sequencer block that is being streamed, filtered for the provided rollup id.
+  astria.sequencerblock.v1alpha1.FilteredSequencerBlock block = 1;
+}
+
+// The Sequencer will serve this to the aucitoneer
+service OptimisticBlockService {
+  // The Sequencer will stream the optimistic Sequencer block (filtered for the provided
+  // rollup id) to the Auctioneer.
+  rpc StreamOptimisticBlock(StreamOptimisticBlockRequest) returns (stream StreamOptimisticBlockResponse);
```
joroshiba, line 32: nit: maybe `GetOptimisticBlockStream`?

proto/executionapis/astria/bundle/v1alpha1/bundle.proto
```diff
@@ -0,0 +1,67 @@
...
+  // Stream blocks from the Auctioneer to Geth for optimistic execution. Geth will stream back
+  // metadata from the executed blocks.
+  rpc StreamExecuteOptimisticBlock(stream StreamExecuteOptimisticBlockRequest) returns (stream StreamExecuteOptimisticBlockResponse);
+}
+
+service BundleService {
+  // A bundle submitter requests bundles given a new optimistic Sequencer block,
+  // and receives a stream of potential bundles for submission, until either a timeout
+  // or the connection is closed by the client.
+  rpc StreamBundles(StreamBundlesRequest) returns (stream Bundle);
```
joroshiba, line 66: `GetBundleStream`

proto/executionapis/astria/bundle/v1alpha1/bundle.proto
```diff
@@ -0,0 +1,67 @@
...
+  // Sequencer block.
+  bytes base_sequencer_block_hash = 3;
+  // The hash of previous rollup block, on top of which the bundle will be executed as ToB.
+  bytes prev_rollup_block_hash = 4;
+}
+
+service OptimisticExecutionService {
+  // Stream blocks from the Auctioneer to Geth for optimistic execution. Geth will stream back
+  // metadata from the executed blocks.
+  rpc StreamExecuteOptimisticBlock(stream StreamExecuteOptimisticBlockRequest) returns (stream StreamExecuteOptimisticBlockResponse);
```
joroshiba, line 59: `ExecuteOptimisticBlockStream`

joroshiba (approved):
some nits on naming but otherwise looks good to me.

proto/executionapis/astria/bundle/v1alpha1/bundle.proto
```diff
@@ -0,0 +1,67 @@
...
+  bytes base_sequencer_block_hash = 2;
+}
+
+message StreamBundlesRequest {}
+
+// Information for the bundle submitter to know how to submit the bundle.
+// The fee and base_sequencer_block_hash are not necessarily strictly necessary
+// it allows for the case where the server doesn't always send the highest fee
+// bundles after the previous but could just stream any confirmed bundles.
+message Bundle {
```
bharath-123, line 40: I think we should name this `StreamBundleResponse` for parity?

Issue #1522, opened 2024-09-19
Update CI to Rust 1.81


SuperFluffy:
Rust [1.81 was released](https://blog.rust-lang.org/2024/09/05/Rust-1.81.0.html) and comes with a load of niceties that we would like to use. Among them of immediate interest are:

1. lint reasons (https://github.com/astriaorg/astria/issues/1521)
2. fixes `clippy::blocks_in_conditions` lint triggering on proc macros like `tracing::instrument` (which https://github.com/astriaorg/astria/pull/1368 needed to work around)
3. lazy lock and lazy cell (to replace once_cell) in 1.80 (https://github.com/astriaorg/astria/issues/1520)
4. checked cfg names (also 1.80 which flags a few issues in our code and should be immediately available).

Note that this is orthgonal to setting the MSRV in individual Astria crates.

┆Issue Number: ENG-818

Issue #1521, opened 2024-09-19
Replace all allow attributes by expect and require lint reasons


SuperFluffy:
Rust 1.81 stabilized lint reasons and the `expect` attribute. We should go through the codebase and replace all uses of `allow` by `expect`, and also require the lints `clippy::allow_attributes` and `clippy::allow_attributes_without_reason`.

┆Issue Number: ENG-819

Issue #1520, opened 2024-09-19
Replace once_cell by sync primitives added in 1.80


SuperFluffy:
Rust 1.80 stabilized `LazyLock` which allows us to remove `once_cell` as a dependency. We should update all code to make use of it after bumping our msrv.

┆Issue Number: ENG-817

PR #1523, opened 2024-09-19
chore(ci): bump rust toolchain to 1.81

SuperFluffy:
## Summary
Updates the rust toolchain used for all rust releated github jobs to and the containerfile to 1.81.0. 

## Background
Rust [1.81 was released](https://blog.rust-lang.org/2024/09/05/Rust-1.81.0.html) and comes with a load of niceties that we would like to use. Among them of immediate interest are:

1. lint reasons (https://github.com/astriaorg/astria/issues/1521)
2. fixes `clippy::blocks_in_conditions` lint triggering on proc macros like `tracing::instrument` (which https://github.com/astriaorg/astria/pull/1368 needed to work around)
3. lazy lock and lazy cell (to replace once_cell) in 1.80 (https://github.com/astriaorg/astria/issues/1520)
4. checked cfg names (also 1.80 which flags a few issues in our code and should be immediately available).

Note that this is orthgonal to setting the MSRV in individual Astria crates.

## Changes
- Add a `rust-toolchain.toml` with an entry `channel = "1.81.0"` to document the default toolchain used for the repository.
- Updates all github workflows to use the `1.81.0` toolchain (except for rustfmt and dylint, which need nightlies, respectively).

## Related Issues
Closes https://github.com/astriaorg/astria/issues/1522


.github/workflows/reusable-build.yml
```diff
@@ -13,6 +13,8 @@ on:
 env:
   REGISTRY: ghcr.io
   FULL_REF: ${{ inputs.tag && format('refs/tags/{0}', inputs.tag) || github.ref }}
+  # This must match the entry in rustup-toolchain.toml at the repository root
```
Fraser999, line 16: Looks like no such file was committed, but it should be named `rust-toolchain.toml` I think, not `rustup-toolchain.toml`.

.github/workflows/reusable-build.yml
```diff
@@ -13,6 +13,8 @@ on:
...
+  RUSTUP_TOOLCHAIN: "1.81.0"
```
Fraser999, line 17: I guess we could name this `RUST_TOOLCHAIN` to be consistent with the filename?

Also, could we read the appropriate value from the `rust-toolchain.toml` file here rather than having to keep these in sync?

.github/workflows/test.yml
```diff
@@ -2,6 +2,9 @@ name: Test
 env:
   CI: true
   RUSTFLAGS: "-D warnings -D unreachable-pub --cfg tokio_unstable"
+
+  # This must match the entry in rustup-toolchain.toml at the repository root
+  RUSTUP_TOOLCHAIN: "1.81.0"
```
Fraser999, line 7: As per two comments above.

lint/tracing_debug_field/rust-toolchain
Fraser999, line 1: It might be nice to provide the explicit `.toml` suffix on this file, particularly if we add a new `rust-toolchain.toml` in the project root.

.github/workflows/reusable-build.yml
SuperFluffy: Yeah, that was confusing. Renamed this to `rust-toolchain.toml` and committed a rustup-toolchain.toml (as per our offline discussion).

.github/workflows/reusable-build.yml
SuperFluffy: `RUSTUP_TOOLCHAIN` is the actual env var that's read by rust: https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file

And I guess we could extract this, but I didn't want to move away from dtolnay/rust-toolchain (at least for now), and parsing toml files in workflows can only be done through obscure github actions, which I didn't feel comfortable introducing.

Followup?

.github/workflows/test.yml
SuperFluffy: Answered there.

lint/tracing_debug_field/rust-toolchain
SuperFluffy: This file is autogenerated by `cargo dylint`.

Fraser999 (approved):

joroshiba (approved):

PR #1525, opened 2024-09-19
chore(sequencer)!: put blocks and deposits to non-verified storage (ENG-812)

### LINKED ISSUE #1493: Check verifiable storage usage
We recently noticed that the `SequencerBlock` was being stored in the verifiable store, but it can be held in the non-verifiable store since it is a product of already having achieved consensus on the constituent parts.

Ensure all other items being put to verifiable storage need to be stored there, and change them to non-verifiable if not.

┆Issue Number: ENG-812

Fraser999:
## Summary
This changes where sequencer blocks and the `TracePrefixed` native asset are stored, and also persists `Deposit`s to non-verifiable storage.

## Background
Sequencer blocks are currently written to verified storage, which is unnecessary as all of their constituent data is already available there.

We want to avoid the cost of storing full blocks in verified storage, but still have the ability to provide a third party with a full block if requested.  To further assist with this, deposits will also be written to non-verified storage.

It also seemed appropriate to store the `TracePrefixed` native asset in verified storage, since this is something which all validators should reach consensus over.

## Changes
- `SequencerBlock`s are now stored in non-verifiable storage.
- `Deposit`s are now stored in non-verifiable storage.
- The native asset is now stored in verifiable storage.

## Testing
Existing unit tests have been updated as required to assert these changes.

## Breaking Changelist
- The set of data being written to verified storage has changed, which is a breaking change in terms of on-disk data and generation of state root hashes.  Otherwise no APIs have changed.

## Related Issues
Closes #1493.


crates/astria-sequencer/src/app/tests_app/mod.rs
```diff
@@ -361,10 +358,6 @@ async fn app_create_sequencer_block_with_sequenced_data_and_deposits() {
         .unwrap();
     app.commit(storage).await;
 
-    // ensure deposits are cleared at the end of the block
```
noot, line 364: ensure the deposit nonce is cleared still?

Fraser999: Done in [639449c](https://github.com/astriaorg/astria/pull/1525/commits/639449cb5f751222f20548f4ee3d1acc18698814).

I've made `get_deposit_nonce` private to `bridge::state_ext` since it's not needed anywhere outside of there, so for this test case I just created a new test-only asserting function.

crates/astria-sequencer/src/bridge/state_ext.rs
```diff
@@ -265,60 +266,26 @@ pub(crate) trait StateReadExt: StateRead + address::StateReadExt {
...
-
-        let Nonce(nonce) =
-            Nonce(u32::from_be_bytes(bytes.try_into().expect(
-                "all deposit nonces stored should be 4 bytes; this is a bug",
-            )));
-        Ok(nonce)
-    }
-
     #[instrument(skip_all)]
     async fn get_deposit_rollup_ids(&self) -> Result<HashSet<RollupId>> {
```
SuperFluffy, line 270: I'd like to replace this by a stream.

crates/astria-sequencer/src/bridge/state_ext.rs
```diff
@@ -265,60 +266,26 @@ pub(crate) trait StateReadExt: StateRead + address::StateReadExt {
...
-        while let Some(Ok((_, value))) = stream.next().await {
-            let raw = RawDeposit::decode(value.as_ref()).wrap_err("invalid deposit bytes")?;
-            let deposit = Deposit::try_from_raw(raw).wrap_err("invalid deposit raw proto")?;
-            deposits.push(deposit);
-        }
-        Ok(deposits)
-    }
-
     #[instrument(skip_all)]
     async fn get_block_deposits(&self) -> Result<HashMap<RollupId, Vec<Deposit>>> {
```
SuperFluffy, line 290: Same here. This should be ideally a stream that's consumed into a datastructure at the call site.

Probaby best done in a followup.

crates/astria-sequencer/src/bridge/state_ext.rs
```diff
@@ -509,26 +465,14 @@ pub(crate) trait StateWriteExt: StateWrite {
...
-        while let Some(Ok((key, _))) = stream.next().await {
-            self.nonverifiable_delete(key);
-        }
-    }
-
-    #[instrument(skip_all)]
-    async fn clear_block_deposits(&mut self) -> Result<()> {
+    async fn clear_block_deposit_nonces(&mut self) -> Result<()> {
         let deposit_rollup_ids = self
             .get_deposit_rollup_ids()
```
SuperFluffy, line 471: Re my comment above: this would allow skipping this allocation entirely

crates/astria-sequencer/src/bridge/state_ext.rs
noot: i will work on this

noot (approved):

crates/astria-sequencer/src/bridge/state_ext.rs
Fraser999: Happy to do this in a followup.  Ticket #1549.

crates/astria-sequencer/src/bridge/state_ext.rs
Fraser999: As above.

crates/astria-sequencer/src/api_state_ext.rs
SuperFluffy, line 1: @noot Can I ask you to double check these keys? I think I recall (but might be mistaken) that you wanted plaintext/human readable keys.

I don't think we ever formalized this, but I would like you to give your explicit approval here.

crates/astria-sequencer/src/bridge/state_ext.rs
SuperFluffy: Yeah, dropped the ball on this one, already created a ticket for this here: https://github.com/astriaorg/astria/issues/1544

crates/astria-sequencer/src/utils.rs
```diff
@@ -13,17 +13,6 @@ use tendermint::abci::{
     EventAttributeIndexExt as _,
 };
 
-pub(crate) struct Hex<'a>(pub(crate) &'a [u8]);
```
SuperFluffy, line 16: This was put in to hex-components of state keys be human readable.

noot: I think this is fine to remove, as we've never actually needed the storage keys to be human-readable

crates/astria-sequencer/src/bridge/state_ext.rs
Fraser999: I missed @noot's comment when I added mine.  Even happier to leave this to her :)

crates/astria-sequencer/src/utils.rs
Fraser999: We were using the `hex` crate for decoding, so I didn't see a problem in using it for the encoding too.  A little less code, and also a bit more performant than our own.

I should probably have called this out in the PR description though - sorry!

crates/astria-sequencer/src/bridge/state_ext.rs
```diff
@@ -1182,7 +1016,7 @@ mod test {
 
         // uninitialized delete ok
         state
-            .clear_block_deposits()
+            .clear_block_deposit_nonces()
             .await
             .expect("calls to clear block deposit should succeed");
```
ethanoroshiba, line 1021:
```suggestion
            .expect("calls to clear block deposit nonces should succeed");
```

crates/astria-sequencer/src/bridge/state_ext.rs
```diff
@@ -1216,56 +1050,61 @@ mod test {
...
-                .get_deposit_events(&rollup_id_1)
+            get_deposit_nonce(&state, &rollup_id_1)
                 .await
                 .expect("deposit should return empty when none exists"),
-            vec![],
-            "deposits were cleared and should return empty vector"
+            0u32,
+            "nonce should have been deleted also"
         );
+        // check that deposits were not deleted
```
ethanoroshiba, line 1083: I think it may make sense to adjust the name of the test to reflect clearing only nonces and specifically not deposits here:
https://github.com/astriaorg/astria/blob/639449cb5f751222f20548f4ee3d1acc18698814/crates/astria-sequencer/src/bridge/state_ext.rs#L1025

crates/astria-sequencer/src/bridge/state_ext.rs
```diff
@@ -1216,56 +1050,61 @@ mod test {
...
         assert_eq!(
-            state
-                .get_deposit_nonce(&rollup_id)
+            super::get_deposit_events(&state, &rollup_id)
                 .await
                 .expect("deposit should return empty when none exists"),
-            0u32,
-            "nonce should have been deleted also"
+            vec![deposit],
+            "deposits were cleared and should return empty vector"
```
ethanoroshiba, line 1089: Is this accurate? I may be mistaken, but it seems like there specifically should be a deposit in the vector returned

crates/astria-sequencer/src/bridge/state_ext.rs
```diff
@@ -1216,56 +1050,61 @@ mod test {
...
         assert_eq!(
-            state
-                .get_deposit_nonce(&rollup_id_1)
+            super::get_deposit_events(&state, &rollup_id_1)
                 .await
                 .expect("deposit should return empty when none exists"),
-            0u32,
-            "nonce should have been deleted also"
+            vec![deposit_1],
+            "deposits were cleared and should return empty vector"
```
ethanoroshiba, line 1096: As above, just wanted to note the other instance

crates/astria-sequencer/src/bridge/state_ext.rs
Fraser999: Fixed in [803b3b4](https://github.com/astriaorg/astria/pull/1525/commits/803b3b4e413de94fbc471e177a3e2cfff96a52f9).

crates/astria-sequencer/src/bridge/state_ext.rs
Fraser999: Fixed in [803b3b4](https://github.com/astriaorg/astria/pull/1525/commits/803b3b4e413de94fbc471e177a3e2cfff96a52f9).

crates/astria-sequencer/src/bridge/state_ext.rs
Fraser999: Fixed in [803b3b4](https://github.com/astriaorg/astria/pull/1525/commits/803b3b4e413de94fbc471e177a3e2cfff96a52f9).

crates/astria-sequencer/src/utils.rs
SuperFluffy: The hex crate always allocates though. So for case `Hex(your_bytes).to_string()` I would prefer to use `hex` directly. But using `hex::encode` when constructing state keys means you always do an extra String allocation before constructing your final string.

I'd rather leave this here and optimize it.

Fraser999: As discussed elsewhere, I'll revert this and we can handle it in a follow-up PR.  Ticketed at #1554.

noot (commented):
changes based on our call:
- store deposits for each block in a different storage key, separated by block number.
- these can be put in the nonverifiable store instead of the ephemeral object store for simplifying the end of block logic

noot (approved):
looks good to me!

PR #1526, opened 2024-09-19
feat(sequencer): rework fee handling

### LINKED ISSUE #1145: sequencer: cleanup/refactor of action fee checks

### LINKED ISSUE #1369: Rework fee reporting

### LINKED ISSUE #1382: Add test to ensure correct block fees for `Ics20Withdrawal`
Currently, the sequencer app tests for correct block fees on every fee-including action other than `Ics20Withdrawal`. This is because the action's `check_and_execute` requires having the IBC channel and ports set, but Penumbra does make the 'channel' module public ~~and its `StateWriteExt` trait is not publicy re-exported~~ (it is now reexported, should be able to implement after #1389).

https://github.com/astriaorg/astria/blob/fbc99f08915897b1237db585163173bd42986306/crates/astria-sequencer/src/app/tests_block_fees.rs#L335

┆Issue Number: ENG-732

ethanoroshiba:
## Summary
Added new `fees` component to hold all fee handling, which implements `check_and_pay_fees()` for each action.

## Background
This is meant to streamline and standardize fee handling in a way that keeps the same functionality, but makes it more difficult to introduce bugs in fee handling in the future. Offline discussion suggested changing the return type of `execute_transaction()` to be our own type as opposed to `Vec<Event>`, but I think this would be best done in a followup, since deposit ABCI events are also recorded via `apply()`.

## Changes
- Added new component `fees` with trait `FeeHandler` and struct `Fee`.
- Created new function `check_execute_and_pay_fees()`, which first calls `check_and_execute()` and then calls `FeeHandler::check_and_pay_fees()`. Changed all actions to use this new function in place of `check_and_execute()`
- Moved all actions' fee handling from `check_and_execute()` to `check_and_pay_fees()`.
- Changed block fee read/writes on the state delta to use the ephemeral cache with `Vec<Fee>`. This allows for `end_block()` to still pay out all fees at the end of the block, and create fee ABCI events at that time. Moved these to `fees::state_ext.rs`.

## Testing
All previous tests passing.

## Breaking Changelist
- Changed block fees storage key, as it does not need to incorporate the asset name into the storage key anymore. This breaks the `app:state_ext::storage_keys_are_unchanged` snapshot

## Related Issues
closes #1369
closes #1145


crates/astria-sequencer/src/app/action_handler.rs
```diff
@@ -1,14 +1,9 @@
 use cnidarium::StateWrite;
 
-/// This trait is a verbatim copy of `cnidarium_component::ActionHandler`.
```
ethanoroshiba, line 3: Deleted because this didn't seem like a super accurate representation of `ActionHandler` anymore, but can revert if others disagree.

crates/astria-sequencer/src/bridge/bridge_sudo_change_action.rs
```diff
@@ -76,27 +72,54 @@ impl ActionHandler for BridgeSudoChangeAction {
...
+            .add_fee_to_block_fees(
+                self.fee_asset.clone(),
+                fee,
+                tx_context.transaction_id,
+                tx_context.source_action_index,
+            )
             .wrap_err("failed to add to block fees")?;
         state
-            .decrease_balance(self.bridge_address, &self.fee_asset, fee)
+            .decrease_balance(from, &self.fee_asset, fee)
```
ethanoroshiba, line 119: Should bridge sudo change action be deducting the fee from the signer as opposed to the bridge address?

SuperFluffy (commented):
I have two notes:
1. I don't see an advantage in changing `AppHandler` itself. The same thing can be done by creating a function and it keeps the `AppHandler` clean.

```rust
fn check_execute_and_pay_fees<T: AppHandler + FeeHandler, S: StateWrite>(action: &T, mut state: S) -> Result<()> {
    action.check_and_execute(state)?;
    action.pay_fees(state)?;
}
```

2. I think it would be nice to make a `fees` component and move all state accesses there.

crates/astria-sequencer/src/assets/state_ext.rs
```diff
@@ -212,54 +171,44 @@ pub(crate) trait StateWriteExt: StateWrite {
...
     }
 
     #[instrument(skip_all)]
-    async fn clear_block_fees(&mut self) {
-        let mut stream =
-            std::pin::pin!(self.nonverifiable_prefix_raw(BLOCK_FEES_PREFIX.as_bytes()));
-        while let Some(Ok((key, _))) = stream.next().await {
-            self.nonverifiable_delete(key);
-        }
+    fn clear_block_fees(&mut self) {
```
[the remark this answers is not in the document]
ethanoroshiba: Removed in d69191a

PR #1548, opened 2024-09-23
chore(sequencer): Move all storage keys to a single parent module (ENG-813)

### LINKED ISSUE #1494: Move all sequencer storage (DB) keys to a single location
It is difficult to be sure that a given storage key is unique since they are defined in multiple different locations in the sequencer codebase.

Move all these keys (const values and functions generating DB keys) to a single module inside the new `storage` module.

┆Issue Number: ENG-813

Fraser999:
## Summary
This groups all keys used for DB entries to a single module, with two submodules, one for verifiable and the other for non-verifiable storage.

## Background
DB keys must be unique, and it's easier to assess this when reviewing, updating or adding new keys if the keys are held in a single location.

## Changes
- Consts and functions for generating DB keys were moved to `storage::verifiable_keys` or `storage::nonverifiable_keys` as appropriate.  The const or generated value of every DB key should be unchanged by this PR. 

## Testing
- Existing snapshot tests were moved.  New snapshot tests were added for all functions used to generate DB keys.  The snapshot values in new `.snap` files which correspond to moved snapshot tests are unchanged.
- New tests added to ensure prefix keys are indeed prefixes of the relevant keys.

## Related Issues
Closes #1494.


Fraser999:
Closed in favour of #1616.

PR #1561, opened 2024-09-25
chore(all): Migrate all instances of `#[allow]` to `#[expect]`

### LINKED ISSUE #1521: Replace all allow attributes by expect and require lint reasons

ethanoroshiba:
## Summary
Implemented usage of `#[expect]` instead of `#[allow]` following bump to Rust 1.81

## Background
With the recent bump to Rust 1.81 (#1523), the usage of `#[expect]` is now stable. This is preferable because it warns in case the lint is unfulfilled, keeping us from having superfluous allow attributes.

## Changes
- Changed `test.yml` to include `clippy::allow_attributes` and `clippy::allow_attributes_with_no_reason` lints
- Updated all crates to use Rust 1.81.0
- Migrated all instances of `#[allow]` to `#[expect]`, with the exception of our generated modules.
- Deleted all allow attributes which were no longer useful

## Testing
Passing all tests

## Related Issues
closes #1521


crates/astria-bridge-withdrawer/tests/blackbox/helpers/test_bridge_withdrawer.rs
```diff
@@ -237,7 +237,7 @@ impl TestBridgeWithdrawer {
     }
 }
 
-#[allow(clippy::module_name_repetitions)]
+#[expect(clippy::module_name_repetitions, reason = "naming is for clarity here")]
```
SuperFluffy, line 240: I think we should actually fix this by renaming to `Config` so that this type can be identified by `test_bridge_withdrawer::Config`

crates/astria-bridge-withdrawer/tests/blackbox/helpers/mod.rs
```diff
@@ -1,5 +1,7 @@
-// These are tests; failing with panics is ok.
-#![allow(clippy::missing_panics_doc)]
+#![expect(
```
SuperFluffy, line 1: Can you move this expect to `blackbox/main.rs`?

crates/astria-bridge-withdrawer/tests/blackbox/helpers/mock_sequencer.rs
```diff
@@ -38,7 +38,10 @@ use tonic::{
 
 const GET_PENDING_NONCE_GRPC_NAME: &str = "get_pending_nonce";
 
-#[allow(clippy::module_name_repetitions)]
+#[expect(
```
SuperFluffy, line 41: Rename to `mock_sequencer::Server` to fix the expect.

crates/astria-conductor/tests/blackbox/main.rs
```diff
@@ -1,6 +1,8 @@
-// allow: clippy lints that are not ok in production code but acceptable or wanted in tests
 pub mod firm_only;
-#[allow(clippy::missing_panics_doc)]
+#[expect(
```
SuperFluffy, line 2: IMO change this to a module/crate wide `#![expect(...)]`

crates/astria-core/src/primitive/v1/asset/denom.rs
```diff
@@ -634,17 +634,14 @@ mod tests {
             TooManySegments,
         };
         #[track_caller]
-        // allow: silly lint
-        #[allow(clippy::needless_pass_by_value)]
+        #[expect(clippy::needless_pass_by_value, reason = "silly lint")]
```
SuperFluffy, line 637: I wrote this, it's not good.
```suggestion
        #[expect(
            clippy::needless_pass_by_value,
            reason = "asserting on owned variants is less noisy then passing them by reference"
        )]

File: crates/astria-core/src/protocol/genesis/v1alpha1.rs [hunk 6]
```diff
@@ -131,6 +131,11 @@ impl Protobuf for GenesisAppState {
     type Error = GenesisAppStateError;
     type Raw = raw::GenesisAppState;
 
+    #[expect(
+        clippy::allow_attributes,
+        clippy::allow_attributes_without_reason,
+        reason = "false positive"
```
SuperFluffy, line 137: Can you spell why it's a false positive?

Also, can you make a tracking issue for us to fix this once its fixed upstream?

crates/astria-core/src/protocol/transaction/v1alpha1/action.rs
SuperFluffy, line 1: Eh, I really want to rename all of these so they lose their `Action` suffix.

crates/astria-core/src/sequencerblock/v1alpha1/block.rs
```diff
@@ -547,7 +546,10 @@ pub struct SequencerBlockParts {
 /// `SequencerBlock` is constructed from a tendermint/cometbft block by
 /// converting its opaque `data` bytes into sequencer specific types.
 #[derive(Clone, Debug, PartialEq)]
-#[allow(clippy::module_name_repetitions)]
+#[expect(
+    clippy::module_name_repetitions,
+    reason = "we want consistent and specific naming"
```
SuperFluffy, line 551: reason is matching the protobuf spec

crates/astria-core/src/sequencerblock/v1alpha1/block.rs
```diff
@@ -967,7 +968,10 @@ pub struct FilteredSequencerBlockParts {
 }
 
 #[derive(Debug, Clone, PartialEq)]
-#[allow(clippy::module_name_repetitions)]
+#[expect(
+    clippy::module_name_repetitions,
+    reason = "we want consistent and specific naming"
```
SuperFluffy, line 973: reason -> protobuf spec

crates/astria-merkle/src/lib.rs
```diff
@@ -218,7 +218,7 @@ impl<'a> LeafBuilder<'a> {
     /// Writes `bytes` into the builder, appending to the leaf.
     ///
     /// See [`Tree::build_leaf`] for example usage.
-    #[allow(clippy::missing_panics_doc)] // invariant of the system
+    #[expect(clippy::missing_panics_doc, reason = "invariant of the system")]
```
SuperFluffy, line 221:
```suggestion
    #[expect(clippy::missing_panics_doc, reason = "invariant of the system: the hasher must be set")]
```

crates/astria-sequencer-relayer/src/relayer/celestia_client/celestia_keys.rs
```diff
@@ -124,8 +124,10 @@ mod tests {
     #[test]
     fn should_fail_to_construct_from_missing_file() {
         let error = CelestiaKeys::from_path("missing").unwrap_err();
-        // allow: `assert!(matches!(..))` provides poor feedback on failure.
-        #[allow(clippy::manual_assert)]
+        #[expect(
+            clippy::manual_assert,
+            reason = "`assert!(matches!(..))` provides poor feedback on failure"
+        )]
```
SuperFluffy, line 130: Can you make a followup issue for these? It'd be nice to have some kind of `assert_error_variant` utility function in these tests.

crates/astria-sequencer-relayer/src/relayer/celestia_client/tests.rs
```diff
@@ -55,8 +57,10 @@ fn new_msg_pay_for_blobs_should_fail_for_large_blob() {
         commitment: Commitment([0; 32]),
     };
     let error = new_msg_pay_for_blobs(&[blob], Bech32Address("a".to_string())).unwrap_err();
-    // allow: `assert!(matches!(..))` provides poor feedback on failure.
-    #[allow(clippy::manual_assert)]
+    #[expect(
+        clippy::manual_assert,
+        reason = "`assert!(matches!(..))` provides poor feedback on failure"
```
SuperFluffy, line 62: Also here and below: we should really have a `assert_error_variant` or something like that.

crates/astria-sequencer/src/app/test_utils.rs
SuperFluffy, line 1: Comment on all the `expect` lines that go above a `cfg_attr` attribute: I think you should be able to put them under the attribute, something like this:

```rust
#[cfg_attr(
    feature = "bla",
    expect(dead_code, reason = "benchmarks use only some parts of the test code").,
)]
```

crates/astria-sequencer/src/bridge/bridge_lock_action.rs
```diff
@@ -289,7 +289,10 @@ mod tests {
     }
 
     #[track_caller]
-    #[allow(clippy::arithmetic_side_effects)] // allow: test will never overflow u128
+    #[expect(
+        clippy::arithmetic_side_effects,
+        reason = "test will never overflow u128"
```
SuperFluffy, line 294:
```rust
        reason = "adding length of strings will never overflow u128 on currently existing machines"
```

crates/astria-sequencer/src/bridge/query.rs
```diff
@@ -41,9 +41,9 @@ fn error_query_response(
     }
 }
 
-// allow / FIXME: there is a lot of code duplication due to `error_query_response`.
+// FIXME: there is a lot of code duplication due to `error_query_response`.
 // this could be significantly shortened.
-#[allow(clippy::too_many_lines)]
+#[expect(clippy::too_many_lines, reason = "should be refactored")]
```
SuperFluffy, line 46: follow up issue?

crates/astria-sequencer/src/mempool/benchmarks.rs
```diff
@@ -285,8 +285,11 @@ fn run_maintenance<T: MempoolSize>(bencher: divan::Bencher) {
         .unwrap();
     // Set the new nonce so that the entire `REMOVAL_CACHE_SIZE` entries in the
     // `comet_bft_removal_cache` are filled (assuming this test case has enough txs).
-    // allow: this is test-only code, using small values, and where the result is not critical.
-    #[allow(clippy::arithmetic_side_effects, clippy::cast_possible_truncation)]
+    #[expect(
+        clippy::arithmetic_side_effects,
+        clippy::cast_possible_truncation,
+        reason = "this is test-only code, using small values, and where the result is not critical"
+    )]
```
SuperFluffy, line 292: OK, this expect is so wordy, I think it'd be easier to just change all of these operations to use checked arithmetic and `unwrap`...

crates/astria-sequencer/src/mempool/benchmarks.rs
```diff
@@ -325,8 +328,11 @@ fn run_maintenance_tx_recosting<T: MempoolSize>(bencher: divan::Bencher) {
...
```
SuperFluffy, line 335: Same here .. the `allow` is so wordy, it'd be easier to just use checked arithmetic.

crates/astria-sequencer/src/mempool/transactions_container.rs
```diff
@@ -827,8 +827,10 @@ mod test {
 
     // From https://doc.rust-lang.org/std/cmp/trait.PartialOrd.html
     #[test]
-    // allow: we want explicit assertions here to match the documented expected behavior.
-    #[allow(clippy::nonminimal_bool)]
+    #[expect(
+        clippy::nonminimal_bool,
+        reason = "we want explicit assertions here to match the documented expected behavior"
```
SuperFluffy, line 832: We should have a followup to fix these. The explicit assertions could be done in a way that makes it easier for the reader.

crates/astria-sequencer/src/test_utils.rs
```diff
@@ -15,6 +15,11 @@ pub(crate) fn astria_address(bytes: &[u8]) -> Address {
         .unwrap()
 }
 
+#[expect(
```
SuperFluffy, line 18: Put the expect under the  `cfg_attr`?

crates/astria-sequencer/src/transaction/mod.rs
```diff
@@ -154,10 +154,10 @@ impl ActionHandler for SignedTransaction {
         Ok(())
     }
 
-    // allowed / FIXME: because most lines come from delegating (and error wrapping) to the
+    // FIXME: because most lines come from delegating (and error wrapping) to the
     // individual actions. This could be tidied up by implementing `ActionHandler for Action`
     // and letting it delegate.
-    #[allow(clippy::too_many_lines)]
+    #[expect(clippy::too_many_lines, reason = "should be refactored")]
```
SuperFluffy, line 160: issue to track?

SuperFluffy (approved):
Looks great, thank you for going through all of those.

I have left a number of comments where I think we should follow up with issues, or where the `allow` and/or `expect` are so wordy, that we should actually just get rid of them in favor of a simpler fix.

Also, there are a few instances where you had `expect` on top of `cfg_attr` attributes, which I think could be removed, and instead replacing the value of the `cfg_attr` by `expect`.

crates/astria-core/src/protocol/genesis/v1alpha1.rs
ethanoroshiba: https://github.com/astriaorg/astria/issues/1580

crates/astria-merkle/benches/benchmark.rs
```diff
@@ -7,8 +7,14 @@ use divan::{
...
-// allow: unused warning if `bench_include_allocs` feature is not enabled.
-#[allow(dead_code)]
+#[expect(
+    clippy::allow_attributes,
+    reason = "this allow is conditional and hence necessary"
+)]
+#[allow(
+    dead_code,
+    reason = "unused warning if `bench_include_allocs` feature is not enabled"
+)]
```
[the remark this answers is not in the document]
ethanoroshiba: Replaced with `#cfg(feature = "bench_include_allocs")`

crates/astria-sequencer-relayer/src/relayer/celestia_client/celestia_keys.rs
ethanoroshiba: https://github.com/astriaorg/astria/issues/1581

crates/astria-sequencer/src/bridge/query.rs
ethanoroshiba: https://github.com/astriaorg/astria/issues/1582

crates/astria-sequencer/src/mempool/transactions_container.rs
ethanoroshiba: https://github.com/astriaorg/astria/issues/1583

quasystaty1 (approved):

Issue #1322, opened 2024-07-31
feat: Trusted Builder MVP


bharath-123:
This issue tracks the progress and the work required to build the trusted builder MVP

```
### Design work
- [x] Write a design doc: https://www.notion.so/astria-org/Trusted-Block-Builder-MVP-8eba7cd7cbd34e4e90a4adf726879028?pvs=4
```

```
### Astria-Geth changes
- [x] Support returning the included transactions as per the new protobuf changes for ExecuteBlock
- [x] Support simulateOnly mode for ExecuteBlock where we only create a block but do not insert it to the chain
- [x] Support validating BuilderBundlePacket
- [ ] Support validating BuilderBundlePacket signatures
- [x] Support moving transactions deserialized from BuilderBundlePacket to the top of block
```

```
### Composer changes
- [x] Support calling GetCommitmentState
- [x] Support calling ExecuteBlock on the final bundle to be submitted to simulate it
- [x] Support creating the BuilderBundlePacket
- [] Support signing of BuilderBundlePacket
- [x] Support sending the BuilderBundlePacket as a sequence action to sequencer
- [ ] Compute the max bytes of seq actions we can send based on the BuilderBundlePacket fields
- [x] Update blackbox tests of composer to call the execution api.
```

```
### Conductor Changes
- [x] Modify conductor to ensure that it is compatible with the new ExecutionAPI protobuf changes
- [x] Fix the unit tests in conductor post the protobuf changes.
```

```
### Protobuf changes
- [x] Modify ExecuteBlock protobuf to return the included transactions
- [x] Modify ExecuteBlock protobuf to accept a simulateOnly flag which will allow us to only create a block but not add it to the chain
- [x] Create the BuilderBundlePacket protobufs
- [x] Ensure that the Execution API protobuf changes and BuilderBundlePacket protobuf changes are pushed to buf.build
```

```
### Testing
- [x] Test if Astria Geth and Conductor compatible with the Execution API changes by deploying them locally
- [x] Launch the changes in a local cluster and send txs with spamooor
```

┆Issue Number: ENG-671

joroshiba:
This issue is stale because it has been open 45 days with no activity. Remove stale label or this issue
be closed in 7 days.

joroshiba:
This issue was closed because it was stale

PR #1562, opened 2024-09-25
refactor(sequencer): generate `SequencerBlock` after transaction execution in proposal phase

noot:
## Summary
refactor the sequencer app to generate and store the resulting `SequencerBlock` after transaction execution even in the proposal phase. 

## Background
previously, we were only generating the `SequencerBlock` in `finalize_block`, however with the upcoming builder APIs (#1519) we require the (proposed) `SequencerBlock` to be available after execution in the proposal phase. 

## Changes
- create a `post_execute_transactions` method and move the after-execution logic that generates the `SequencerBlock` from `finalize_block` to there.
- call this method after transaction execution in `process_proposal`. 
- if txs were executed in `prepare_proposal`, `post_execute_transactions` is still called in `process_proposal`, as the block hash is not available in `prepare_proposal`.

## Testing
existing unit tests pass, app logic was not changed, just moved around

## Related Issues
 related to #1322

### DIFF
--- a/crates/astria-sequencer/src/app/mod.rs
+++ b/crates/astria-sequencer/src/app/mod.rs
@@ -293,6 +293,7 @@ impl App {
         self.state = Arc::new(StateDelta::new(storage.latest_snapshot()));
 
         // clear the cache of transaction execution results
+        self.execution_results = None;
         self.finalize_block = None;
         self.executed_proposal_hash = Hash::default();
     }
@@ -377,9 +378,10 @@ impl App {
                 self.executed_proposal_hash = process_proposal.hash;
 
                 // if we're the proposer, we should have the execution results from
-                // `prepare_proposal`. run the post-tx-execution hook to set
-                // `self.finalize_block`. we can't run this in `prepare_proposal` as
-                // we don't know the block hash there.
+                // `prepare_proposal`. run the post-tx-execution hook to generate the
+                // `SequencerBlock` and to set `self.finalize_block`.
+                //
+                // we can't run this in `prepare_proposal` as we don't know the block hash there.
                 let Some(tx_results) = self.execution_results.take() else {
                     bail!("execution results must be present after executing transactions")
                 };
@@ -797,6 +799,10 @@ impl App {
         Ok(())
     }
 
+    /// updates the app state after transaction execution, and generates the resulting
+    /// `SequencerBlock`.
+    ///
+    /// this must be called after a block's transactions are executed.
     #[instrument(name = "App::post_execute_transactions", skip_all)]
     async fn post_execute_transactions(
         &mut self,
--- a/crates/astria-sequencer/src/app/mod.rs
+++ b/crates/astria-sequencer/src/app/mod.rs
@@ -68,6 +68,7 @@ use tendermint::{
     AppHash,
     Hash,
 };
+use tokio::sync::watch;
 use tracing::{
     debug,
     info,
@@ -162,9 +163,13 @@ pub(crate) struct App {
     // to the mempool to recost all transactions.
     recost_mempool: bool,
 
-    // cache of results of executing of transactions in `prepare_proposal` or `process_proposal`.
-    // cleared at the end of each block.
-    execution_results: Option<Vec<tendermint::abci::types::ExecTxResult>>,
+    // cache of results of executing of transactions in `prepare_proposal`.
+    // cleared in `process_proposal` if we're the proposer.
+    execution_results: Option<Vec<ExecTxResult>>,
+
+    // cache of results of executing of transactions in `process_proposal`.
+    // cleared at the end of the block.
+    finalize_block: Option<abci::response::FinalizeBlock>,
 
     // the current `StagedWriteBatch` which contains the rocksdb write batch
     // of the current block being executed, created from the state delta,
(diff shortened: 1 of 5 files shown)


crates/astria-sequencer/src/app/mod.rs
```diff
@@ -162,9 +162,13 @@ pub(crate) struct App {
...
-    // cache of results of executing of transactions in `prepare_proposal` or `process_proposal`.
-    // cleared at the end of each block.
-    execution_results: Option<Vec<tendermint::abci::types::ExecTxResult>>,
+    // cache of results of executing of transactions in `prepare_proposal`.
+    // cleared in `process_proposal` if we're the proposer.
+    execution_results: Option<Vec<ExecTxResult>>,
+
+    // cache of results of executing of transactions in `process_proposal`.
+    // cleared at the end of the block.
+    finalize_block: Option<abci::response::FinalizeBlock>,
```
joroshiba, line 171: Should we utilize the ephemeral storage within cnidarium vs adding to state on the app? I know @SuperFluffy has been trying to decrease our global state bloat.

noot: this is functionally the same - it's an object stored in memory and cleared at the start of each block. the ephemeral store is also only in memory and is cleared at the end of each block

SuperFluffy: I agree that we should employ the ephemeral store more rather than adding fields to app.

There is a tendency to add more and more fields to `App` that need to be actively set/unset, which is not clean (and I believe with which we had issues in the past).

For a reader it's also better to have a single source of truth of effects that happen during block building, rather than finding some in the ephemeral store and some in the "global" app state.

crates/astria-sequencer/src/app/mod.rs
```diff
@@ -458,6 +482,16 @@ impl App {
         );
 
         self.executed_proposal_hash = process_proposal.hash;
+        self.post_execute_transactions(
```
SuperFluffy, line 485: Why not pass this the entire `process_proposal`?

crates/astria-sequencer/src/app/tests_app/mempool.rs
SuperFluffy, line 1: This test is so complex, I don't know how to evaluate this change to it.

crates/astria-sequencer/src/app/mod.rs
```diff
@@ -758,18 +791,24 @@ impl App {
...
+    /// `SequencerBlock`.
     ///
-    /// This is called by cometbft after the block has already been
-    /// committed by the network's consensus.
-    #[instrument(name = "App::finalize_block", skip_all)]
-    pub(crate) async fn finalize_block(
+    /// this must be called after a block's transactions are executed.
+    #[instrument(name = "App::post_execute_transactions", skip_all)]
+    async fn post_execute_transactions(
         &mut self,
```
SuperFluffy, line 800: Hard to follow all the side effects in this function.

crates/astria-sequencer/src/app/mod.rs
```diff
@@ -781,16 +820,73 @@ impl App {
...
+            .wrap_err("failed to write sequencer block to state")?;
+
+        // events that occur after end_block are ignored here;
+        // there should be none anyways.
+        let _ = self.apply(state_tx);
+
+        // use a dummy app hash here - the actual app hash will be filled out in finalize_block.
+        let dummy_app_hash = AppHash::default();
+
+        let finalize_block = abci::response::FinalizeBlock {
```
SuperFluffy, line 867: It's surprising that this method changes the internal state of `self.finalize_block`, but the `finalize_block` takes a `finalize_block` argument and then mutates the `finalize_block` field that some other part of the stack changed.

SuperFluffy (commented):
I have a hard time following the changes, which might be an artifact of how app execution is structured.

I am also surprised that making sequencer blocks available during an earlier phase leads to changes in mempool tests.

crates/astria-sequencer/src/app/mod.rs
noot: this is also called in `finalize_block` potentially, so it might not have `process_proposal` available

crates/astria-sequencer/src/app/tests_app/mempool.rs
noot: yeah :/ i didn't write the test, just modified it/added some asserts for sanity when updating it

crates/astria-sequencer/src/app/mod.rs
noot: i could change this to be some intermediate type that doesn't have the app hash yet, which is turned into `finalize_block` response only in `finalize_block`. would that be easier to follow?

noot:
> I am also surprised that making sequencer blocks available during an earlier phase leads to changes in mempool tests.

the changes are because `process_proposal` must now also be called if `prepare_proposal` is called (this is the normal ABCI flow btw). that's really the only change to the tests

crates/astria-sequencer/src/app/mod.rs
noot: this function is basically everything that was in `finalize_block` after txs were executed (minus mempool cleaning, as that needs to be done only when the block is actually finalized)

crates/astria-sequencer/src/app/mod.rs
noot: moved the fields to the ephemeral object store here: https://github.com/astriaorg/astria/pull/1562/commits/591fd94c89ae27b26c5b067f423dd2a657f87d41

this is nice because the fields are automatically cleared when `update_state_for_new_round` is called, however having to get a `StateDelta` for writing to it is bit more cumbersome in parts (putting the `execution_results` in `prepare_proposal`)

Fraser999 (approved):
There's one mention of `self.execution_results` left behind in the doc comments for `execute_transactions_prepare_proposal` which would be good to have updated.

I also think we should look to abstract the usage of the ephemeral store behind our own API provided by the extension traits in `app/state_ext.rs` (similar to how cached block deposits are handled in the `bridge` extension traits).  We'd also then keep the keys for these in `app::state_ext` along with the others.  But I can handle changing to that as part of the ongoing storage updates.

Neither of these are blocking though, so approving as is.

crates/astria-sequencer/src/app/mod.rs
```diff
@@ -629,15 +662,16 @@ impl App {
         self.metrics
             .set_transactions_in_mempool_total(self.mempool.len().await);
 
-        self.execution_results = Some(execution_results);
+        let mut state_tx = Arc::try_begin_transaction(&mut self.state)
```
SuperFluffy, line 665: This looks complex. Can you add comment above explaining why it's necessary? Seems like a hack around cnidarium not providing us with something we need.

crates/astria-sequencer/src/app/mod.rs
SuperFluffy: I suppose it's fine. We should probably sit down and refactor of all of this. It looks like we have a bit of code duplication going on.

PR #1564, opened 2024-09-25
fix(charts): hermes monitoring and prometheus rules 

quasystaty1:
## Summary
Fixes hermes related monitoring service with Prometheus scraping and adds prometheus rules template for alerts.
## Background
hermes monitoring service is misconfigured and alerts are yet to be implemented.
## Changes
adds `serviceMonitor` and `prometheusRule` templates enabling data scraping and alerts with prometheus.
also fixes current service porting and default image and labels.
## Testing
against dusk-10, locally

## Metrics
enables `hermes-relayer` metrics scraping

### DIFF
--- a/charts/hermes/files/config.toml
+++ b/charts/hermes/files/config.toml
@@ -8,7 +8,7 @@ enabled = {{ .Values.rest.enabled }}
 
 # Specify the IPv4/6 host over which the built-in HTTP server will serve the RESTful
 # API requests. Default: 127.0.0.1
-host = '127.0.0.1'
+host = '0.0.0.0'
 
 # Specify the port over which the built-in HTTP server will serve the restful API
 # requests. Default: 3000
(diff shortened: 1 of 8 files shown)


charts/hermes/files/config.toml
```diff
@@ -8,7 +8,7 @@ enabled = {{ .Values.rest.enabled }}
 
 # Specify the IPv4/6 host over which the built-in HTTP server will serve the RESTful
 # API requests. Default: 127.0.0.1
-host = '127.0.0.1'
+host = '0.0.0.0'
```
quasystaty1, line 11: Allows docker to expose metrics and api

charts/hermes/values.yaml
```diff
@@ -3,7 +3,7 @@ global:
   replicaCount: 1
   logLevel: debug
 
-image: ghcr.io/penumbra-zone/hermes:main
+image: ghcr.io/astriaorg/hermes:sha-450f848
```
quasystaty1, line 6: uses the updated hermes version which should be tagged

WafflesVonMaple (approved):

Issue #1546, opened 2024-09-23
Don't distinguish between command types and their handlers


SuperFluffy:
Astria-cli distinguishes between the [`Cli`](https://github.com/astriaorg/astria/blob/f11f9000a28a4a2959ee488bfad73b751a4194c9/crates/astria-cli/src/cli/mod.rs#L18) container, which provides the entry point to the command line utility, and their various handlers that take the parsed CLI options as an argument.

This separation is confusing and arguably not idiomatic.

astria-cli should be refactored such that each subcommand derives the `clap::Args`, but also provides an entrypoint `fn run(self)`, i.e.:

```rust
#[derive(Args, Debug)]
struct TheSubcommand {
    // options, flags, etc
}

impl TheSubcommand {
    fn run(self) -> Result<T, E> {
        // use the parsed arguments
    }
}
```

┆Issue Number: ENG-834

PR #1568, opened 2024-09-26
refactor(cli): merge argument parsing and command execution

SuperFluffy:
## Summary
Refactors the CLI to give it a cleaner structure.

## Background
The CLI was split into a `cli` module that was concerned with parsing arguments, and a `commands` which consumed the argument data structures.

The patch clears this up by implementing inherent `<Args>::run` methods on all leaf argument types, so that the implementation follows the same structure for all its effects.

Note that this match does restructure the public facing command line (except for a few lines of documentation). It merely restructures the implementation.

## Changes
- Refactors each type deriving `clap::Args` to have an inherent method `run`.
- Flattens the crate structure to not be unnecessarily nested.

## Testing
These will be tested in follow PRs. Smoke tests using a subset of the astria CLI commands should still work.

## Related Issues
Closes https://github.com/astriaorg/astria/issues/1546


sambukowski (approved):

crates/astria-cli/src/sequencer/block_height.rs
```diff
@@ -0,0 +1,63 @@
...
+        default_value = crate::DEFAULT_SEQUENCER_RPC
+    )]
+    sequencer_url: String,
+    /// The chain id of the sequencing chain being used
+    #[arg(
+        long = "sequencer.chain-id",
+        env = "ROLLUP_SEQUENCER_CHAIN_ID",
+        default_value = crate::DEFAULT_SEQUENCER_CHAIN_ID
+    )]
+    sequencer_chain_id: String,
```
sambukowski, line 45: I don't think this is used here?

sambukowski (approved):
LGTM! Only thing is the nit of the extra chain id.

crates/astria-cli/src/sequencer/block_height.rs
SuperFluffy: No, it's not. But fixing this is out of scope of this PR. This bit is copied verbatim from the previous version.

joroshiba (approved):

PR #1574, opened 2024-09-26
feat(sequencer): make ABCI response for account balances deterministic

### LINKED ISSUE #1451: sequencer: `balance_request()` query not guaranteed to return balances in deterministic order
The query `balance_request()`'s dependent function `get_account_balances()` is not guaranteed to return keys in a deterministic order. The [underlying stream](https://github.com/astriaorg/astria/blob/e9d5f0f0fc2f2449292aeeb07a1f2f7e7227a637/crates/astria-sequencer/src/accounts/state_ext.rs#L76) isn't guaranteed to be sorted.

It is bad for an external API to be non-deterministic and should be fixed.

┆Issue Number: ENG-789

ethanoroshiba:
## Summary
Ensures that a response to an ABCI request for account balances is deterministic.

## Background
Previously was not sorted, and the `AssetBalance` stream is non-deterministic, which isn't great for an external API.

## Changes
- Created deterministic sorting function for sorting the `AssetBalance`s firstly in descending order by balance, then in ascending order by name if the balances are equal.

## Testing
Added unit test to ensure the sorting function works as expected.

## Related Issues
closes #1451


crates/astria-sequencer/src/accounts/query.rs
```diff
@@ -215,3 +220,84 @@ async fn preprocess_request(
...
+
+    match lhs.balance.cmp(&rhs.balance) {
+        // Denoms should never have the same display name
+        Ordering::Equal => lhs.denom.to_string().cmp(&rhs.denom.to_string()),
+        Ordering::Less => Ordering::Greater,
+        Ordering::Greater => Ordering::Less,
+    }
+}
+
+mod test {
```
SuperFluffy, line 238:
```suggestion
#[cfg(test)]
mod tests
```

crates/astria-sequencer/src/accounts/query.rs
```diff
@@ -215,3 +220,84 @@ async fn preprocess_request(
...
+
+/// Custom deterministic sorting function for `AssetBalance` that sorts by balance in descending
+/// order and then by denom in ascending order. This function should never return `Ordering::Equal`,
+/// as calling `sort_unstable_by` on equal values has a non-deterministic result.
+fn compare_asset_balances(lhs: &AssetBalance, rhs: &AssetBalance) -> std::cmp::Ordering {
+    use std::cmp::Ordering;
+
+    match lhs.balance.cmp(&rhs.balance) {
+        // Denoms should never have the same display name
+        Ordering::Equal => lhs.denom.to_string().cmp(&rhs.denom.to_string()),
```
SuperFluffy, line 232: Allocations in a hot loop are nasty (even in edge cases). Can you instead hand-roll a `PartialEq` impl on `Denom`? I think if you `PartialEq for TracePrefixed` and `PartialEq for IbcPrefixed`, and then make the decision to order a denom first by `ibc`, then by `trace-prefixed` (I think that's natural? I let you decide), this should be relatively straight forward.

crates/astria-sequencer/src/accounts/query.rs
```diff
@@ -100,6 +100,11 @@ pub(crate) async fn balance_request(
             };
         }
     };
+
+    // Unstable sort does not allocate auxillary memory and is typically faster. Custom sorting
+    // function is deterministic and only allocates in the case that two values are equal.
```
SuperFluffy, line 105: Commented elsewhere: I would very much like to implement `PartialOrd` and `Ord` on `Denom` and its variants instead of doing the allocation here.

SuperFluffy:
Note on the PR title:
1. I removed `!` because I don't think it's breaking. If you had gone from `ordered -> unordered` I would have agreed (since external users might have relied on the response being ordered). But going from `unordered -> ordered` is more structure so that's fine.
2. I changed it from `chore` to `feat` because I was mildly leaning towards this being more of a fix (but maybe chore or fix would have been also good).
3. I edited the PR title to be more descriptive for users who don't have intimate knowledge of the particular module you are changing. In general I try to avoid mentioning specifics item names (types, functions) in favor of being more descriptive.

crates/astria-sequencer/src/accounts/query.rs
```diff
@@ -88,7 +88,7 @@ pub(crate) async fn balance_request(
         Err(err_rsp) => return err_rsp,
```
SuperFluffy, line 88: At the top of this function (i.e. `fn balance_request`), please add a doc comment that explains that it will return the balances sorted by denom and amount.

crates/astria-sequencer/src/accounts/query.rs
ethanoroshiba: Done in fb7110f. Also continued down then chain, implementing `PartialOrd` and `Ord` for `TraceSegments` and `PortAndChannel`

crates/astria-sequencer/src/accounts/query.rs
SuperFluffy: Reopened the issue: the convention is to call this `mod tests`, not `test` as per the suggestion.

crates/astria-core/src/primitive/v1/asset/denom.rs
SuperFluffy, line 1: For the `Ord` impl here, can you provide exhaustive tests that demonstrate and document the order of the denoms?

Naively, I would expect a lexical ordering among trace-prefixed and ibc-prefixed denoms, so it would be great to have tests that also demonstrate that ordering.

I think I should also walk back my suggestion to order ibc-prefixed, trace-prefixed later: a naive implementation that keeps ibc denominations in string form (without parsing) would just use lexical ordering of the strings. I think the parsed variant here should also reflect that.

In addition to unit tests for `IbcPrefixed` and `TracePrefixed`, do a more complex example:

1. create a "string list" of ibc denomination strings (mixing both trace-prefixed and ibc-prefixed).
2. parse this list as a `Denom list`.
3. sort the string list (this will give lexical order).
4. sort the denom list by its `Ord` implementation and then use `to_string` to convert all elements to strings again
5. assert that results of steps 3 and 4 are the same.

crates/astria-sequencer/src/accounts/query.rs
```diff
@@ -215,3 +223,84 @@ async fn preprocess_request(
     };
     Ok((address, snapshot, height))
 }
+
+/// Custom deterministic sorting function for `AssetBalance` that sorts by balance in descending
+/// order and then by denom in ascending order. This function should never return `Ordering::Equal`,
+/// as calling `sort_unstable_by` on equal values has a non-deterministic result.
+fn compare_asset_balances(lhs: &AssetBalance, rhs: &AssetBalance) -> std::cmp::Ordering {
```
SuperFluffy, line 230: Rethinking this: these should just be sorted by lexical order of the denominations.

You cannot compare the magnitude of different quantities, so we shouldn't order by them.

crates/astria-sequencer/src/accounts/query.rs
ethanoroshiba: Sorry about that, should be fixed in 535c1c4

crates/astria-core/src/primitive/v1/asset/denom.rs
```diff
@@ -147,6 +147,29 @@ impl FromStr for Denom {
...
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for Denom {
+    /// If the denoms are the same type, returns their comparison. Otherwise, returns IBC prefixed
+    /// denoms as less than trace prefixed denoms.
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        match self {
```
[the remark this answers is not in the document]
ethanoroshiba: Was unable to import `Self` due to Rust's restrictions, but implemented the tuples in 535c1c4.

crates/astria-sequencer/src/accounts/query.rs
```diff
@@ -100,6 +102,13 @@ pub(crate) async fn balance_request(
             };
         }
     };
+
+    // Unstable sort does not allocate auxillary memory and is typically faster. This utilizes a
```
SuperFluffy, line 106: IMO this comment is not required.

crates/astria-core/src/protocol/account/v1alpha1/mod.rs
```diff
@@ -55,6 +55,18 @@ impl AssetBalance {
     }
 }
 
+impl PartialOrd for AssetBalance {
```
SuperFluffy, line 58: I think this `PartialOrd` is not correct because it completely ignores the balance itself.

Worse, it's actually at odds with the `PartialEq` and `Eq` derivation, because:


```rust
let asset_one = AssetBalance {
    denom: "mydenom".parse().unwrap(),
    balance: 42,
};
let asset_two = AssetBalance {
    denom: "mydenom".parse().unwrap(),
    balance: 42,
};
assert_eq!(asset_one.cmp(&asset_two), Ordering::Equal);
assert!(asset_one.neq(&asset_two));
```

crates/astria-sequencer/src/accounts/query.rs
```diff
@@ -100,6 +102,13 @@ pub(crate) async fn balance_request(
...
+    // custom `Ord` implementation which sorts by `Denom` in alphabetical order. The comparison
+    // should never return `Ordering::Equal` since `Denom` is a unique identifier, hence the
+    // unstable sort will be deterministic.
+    balances.sort_unstable();
```
SuperFluffy, line 110: Instead of using the problematic `Ord` on `AssetBalance`, use https://doc.rust-lang.org/std/primitive.slice.html#method.sort_unstable_by_key:
```rust
balances.sort_unstable_by_key(asset_balance.denom());
```

crates/astria-core/src/primitive/v1/asset/denom.rs
```diff
@@ -742,4 +862,76 @@ mod tests {
         let denom = denom_str.parse::<Denom>().unwrap();
         assert_eq!(denom_str.len(), denom.display_len());
     }
+
+    #[test]
+    fn ibc_prefixed_ord_matches_lexical_sort() {
```
SuperFluffy, line 867: I'd like to have these be more atomic and not sort a vec (it's pretty difficult to read).

Instead do something like:

```rust
#[track_caller]
fn assert_denom_order<T: FromStr>(left: &str, right: &str]) {
    let left_denom: T = left.parse().unwrap();
    let right_denom: T = right.parse().unwrap();
    assert_eq!(left_denom.cmp(&right_denom), left.cmp(right));
}

#[test]
fn denoms_are_lexically_ordered() {
    assert_denom_order::<Denom>("ibc/etc", "trace/etc");
    todo!("and all the rest");
}
```

crates/astria-sequencer/src/accounts/query.rs
ethanoroshiba: This method requires cloning the denom since it doesn't implement `Copy`. Is there a different way you'd like to accomplish this without extra allocation? Aside from that, my best suggestion would be to just fix the `Ord` impl.

SuperFluffy: Why do you need to clone? There is a blanket `impl Ord for &A where A: Ord`. So implement `AssetBalance::denom(&self) -> &Denom`, so you can call `balances.sort_unstable_by_key(AssetBalance::denom)`.

ethanoroshiba: Implemented changes we discussed offline in 0af3bcf.

crates/astria-core/src/primitive/v1/asset/denom.rs
ethanoroshiba: Did something similar in 0e5eeb5, let me know if it works for you!

crates/astria-core/src/protocol/account/v1alpha1/mod.rs
ethanoroshiba: Removed in 0af3bcf.

crates/astria-sequencer/src/accounts/query.rs
SuperFluffy, line 106: Remove this comment.

SuperFluffy (approved):
This looks great, thanks! The PartialOrd/Ord impls on `Denom` will be useful for a small PR I am writing right now.

Please remove the comment above `balances.sort_unstable_by` (for someone with cursory knowledge of the Rust std the use of `sort_unstable` is not unexpected), but otherwise this looks good.

crates/astria-core/src/protocol/account/v1alpha1/mod.rs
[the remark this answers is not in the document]
ethanoroshiba: I've removed the impls for `Ord` and `PartialOrd` from `AssetBalance`, but aside from that all of them are necessary. `Denom::cmp` calls either `IbcPrefixed::cmp` or `TracePrefixed::cmp`, and trace calls `TraceSegments::cmp`, etc.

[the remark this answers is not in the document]
SuperFluffy: Yeah, those make sense. I just thought of `AssetBalance` :)

PR #1578, opened 2024-09-27
chore: rename all modules named `test` to `tests`

Fraser999:
## Summary
A trivial rename of some modules.

## Background
It's canonical to name unit test modules as `tests`.  We currently have a mixture of `test` and `tests` throughout the codebase.

## Changes
- Renamed all `test` modules to `tests`.  No actual tests were changed.

### DIFF
--- a/crates/astria-cli/src/commands/sequencer.rs
+++ b/crates/astria-cli/src/commands/sequencer.rs
@@ -501,7 +501,7 @@ async fn submit_transaction(
 }
 
 #[cfg(test)]
-mod test {
+mod tests {
     use super::*;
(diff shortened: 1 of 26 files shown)


SuperFluffy (approved):
These are great.

In a future patch we should also break out the tests into free-standing files (at least for oversized modules).

And also merge https://github.com/astriaorg/astria/blob/33dae4256f7f6d1f936f9e0de166695be1bf416d/crates/astria-sequencer/src/app/tests_block_fees.rs and its sibling test modules into a folder structure under a common `tests/` module.

PR #1579, opened 2024-09-27
chore(grpc-mock): restructure and add mock grpc tests

### LINKED ISSUE #902: Test grpc mock combinators
`astria-grpc-mock-test` should test more complex matchers. It especially does not test `.and` combinators to chain several matchers.

### LINKED ISSUE #1559: test grpc delayed responses
#1472 adds functionality to `astria-grpc-mock` to delay responses by a certain amount of time after a matching request has been made. This should be tested in `astria-grpc-mock-test` together with changes from #902

ethanoroshiba:
## Summary
Divided GRPC mock tests by component and added a few more.

## Background
Some parts of the GRPC mock still weren't tested, and the crate also needed a test for the new delay capabilities.

## Changes
- Divided tests up by `mock`, `response`, and `matcher`.
- Added field to `HealthCheckRequest` so that we could more accurately test `matcher::partialpbjson`.
- Added tests for the `and` mock combinator, matchers not matching incorrect requests, `up_to_n_times`, `DynamicResponse`, and delay.

## Testing
All tests passing.

## Related Issues
closes #902 
closes #1559


crates/astria-grpc-mock-test/tests/health/test_response.rs
```diff
@@ -0,0 +1,137 @@
...
+#[tokio::test]
+async fn response_delay_works_as_expected() {
+    let server = start_mock_server().await;
+    let mut client = HealthClient::connect(format!("http://{}", server.local_addr))
+        .await
+        .unwrap();
+    let mock = Mock::for_rpc_given("check", matcher::message_type::<HealthCheckRequest>())
+        .respond_with(
+            response::default_response::<HealthCheckResponse>()
+                .set_delay(Duration::from_millis(250)),
```
Fraser999, line 114: I'd maybe make this duration a `const` to be used here and below.

crates/astria-grpc-mock-test/tests/health/test_response.rs
```diff
@@ -0,0 +1,137 @@
...
+    });
+
+    timeout(
+        Duration::from_millis(250),
+        join(mock_guard.wait_until_satisfied(), rsp_fut),
+    )
+    .await
+    .unwrap_err();
+
+    let rsp = client
```
Fraser999, line 129: I'd suggest doing `let start = Instant::now();` just before this, and after the `.await.unwrap()` below, you can assert that `start.elapsed() >= const delay`.

crates/astria-grpc-mock-test/proto/health.proto
SuperFluffy, line 1: I'd prefer to not change this file because it's taken from a well known service.

crates/astria-grpc-mock-test/tests/health/test_response.rs
```diff
@@ -0,0 +1,137 @@
...
+        .unwrap()
+        .to_digit(10)
+        .unwrap();
+    HealthCheckResponse {
+        status: i32::try_from(status_return).unwrap(),
+    }
+}
+
+#[tokio::test]
+async fn response_delay_works_as_expected() {
```
SuperFluffy, line 106: The test on this file should not be on the mock (as it's now), but on the client receiving its request with a delay (I think this is also how it's implemented in the mock).

I think this delay can be tested simple:

1. mount a mock with any number of responses and a delay of 250ms.
2. run 2 requests against the mock, first with a timeout 200ms, second with 300ms timeout. Observe that the 200ms failed, but the 300ms passes.

crates/astria-grpc-mock-test/proto/health.proto
```diff
@@ -26,7 +26,8 @@ option java_outer_classname = "HealthProto";
 option java_package = "io.grpc.health.v1";
 
 message HealthCheckRequest {
-  string service = 1;
+  string name = 1;
+  string service = 2;
```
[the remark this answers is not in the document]
ethanoroshiba: Thanks for the catch! I'll keep this in mind moving forward. I've reverted these changes per Janis' request as well

crates/astria-grpc-mock-test/tests/health/test_response.rs
ethanoroshiba: After implementing Janis' suggestion below, I'm not sure this is necessary anymore, but happy to put it in if you think it is!

crates/astria-grpc-mock-test/tests/health/test_response.rs
ethanoroshiba: Same comment as above, but happy to put it in still if needed

crates/astria-grpc-mock-test/proto/health.proto
ethanoroshiba: Reverted in 4b15aa0, but needed a type with more fields to adequately test `message_partial_pbjson` so I added a new struct `MockMessage` to `test_utils`. Let me know if this works for you!

crates/astria-grpc-mock-test/tests/health/test_response.rs
ethanoroshiba: Done in 4b15aa0.

crates/astria-grpc-mock-test/tests/health/test_response.rs
Fraser999: I think it's still worthwhile, since we can specify e.g.
```rust
const DELAY: Duration = Duration::from_millis(250);
const FIFTY_MILLIS: Duration = Duration::from_millis(50);
```
then below use, `DELAY` to set the delay and `DELAY - FIFTY_MILLIS` for the error case and `DELAY + FIFTY_MILLIS` for the success one.

I think that's slightly easier to understand and update if needed.

crates/astria-grpc-mock-test/tests/health/test_response.rs
```diff
@@ -105,33 +100,33 @@ fn dynamic_responder(request: &HealthCheckRequest) -> HealthCheckResponse {
...
-    let rsp = client
-        .check(HealthCheckRequest {
-            name: "helloworld".to_string(),
-            service: "helloworld".to_string(),
-        })
+    timeout(Duration::from_millis(200), rsp_fut_expect_err)
         .await
-        .unwrap();
-    assert_eq!(&HealthCheckResponse::default(), rsp.get_ref());
+        .unwrap_err(); // should be error
```
Fraser999, line 125: I'd remove the comment as it's just noise IMO (`.unwrap_err()` is effectively saying the same thing as the comment)

Same point for the comment after `unwrap()` below.

crates/astria-grpc-mock-test/tests/health/test_response.rs
ethanoroshiba: Done in 7cebad8

SuperFluffy (commented):
This is looking good. I would not do the manual the manual mock message - just because we might end up in a situation where we bump either of prost, tonic, or pbjson, subtly breaking the derived code.

ethanoroshiba:
@SuperFluffy in RE to the comment ^^above (since I can't reply directly), I'm not sure the echo server would be sufficient as is. To adequately test the `partial_pbjson` matcher, we need a message with at least two fields, which the echo request/response does not have. Were you meaning to copy it over and add another field?

joroshiba:
closed in 7 days.

Fraser999 (approved):

crates/astria-grpc-mock-test/tests/health/main.rs
```diff
@@ -3,270 +3,7 @@
...
-        .respond_with(response::default_response::<HealthCheckRequest>());
-
-    let guard = server.mocked.register_as_scoped(mock).await;
-    let check = client.check(HealthCheckRequest {
-        service: "helloworld".to_string(),
-    });
-
-    let _ = join!(guard.wait_until_satisfied(), check);
-}
+mod test_matcher;
```
SuperFluffy, line 6: This looks good, but please avoid prepending `test_` for every submodule.

SuperFluffy:
> @SuperFluffy in RE to the comment ^^above (since I can't reply directly), I'm not sure the echo server would be sufficient as is. To adequately test the `partial_pbjson` matcher, we need a message with at least two fields, which the echo request/response does not have. Were you meaning to copy it over and add another field?

I didn't realize I never answered. Sorry about that.

What I meant to say was that I rather you rename the health service to something else, or introduce a new service over a "well known service".

crates/astria-grpc-mock-test/tests/health/main.rs
ethanoroshiba: Done in 8a4b360.

joroshiba:
closed in 7 days.

joroshiba:

PR #1588, opened 2024-09-27
ci: code freeze through github actions

joroshiba:
## Summary
Adds functionality to block PRs into sections of code under code freeze within the CI workflow

## Background
We are moving towards mainnet & undergoing audits. During past audits we have not had a firm mechanism to ensure we are not making changes to code under audit. This provides a stopper, but can be overriden still.

## Changes
- New CI workflow that checks for changes to code frozen pathways, fails and adds errors to those files as well as building an error summary. Using the `override-freeze` tag disables the check and changes reports to warnings.

## Testing
Ran against this PR with multiple changes, tested both adding and removing the label to override.

### DIFF
--- a/.github/code-freeze-filter.yaml
+++ b/.github/code-freeze-filter.yaml
@@ -4,11 +4,15 @@
 
 # Please provide a reasoning for each component that is frozen.
 
-# Conductor is frozen for zellic audit.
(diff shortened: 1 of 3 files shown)


WafflesVonMaple (approved):
Very nice idea

PR #1592, opened 2024-09-30
refactor(sequencer): use builder pattern for transaction container tests

Lilyjjo:
## Summary
Switched the method of constructing transactions from a constructor to a builder pattern. 

## Background
We're about to add more relevant transaction fields (ActionGroup type, valid block number), and this change will make it possible to add these extra fields without having to change all of the other tests. 

## Testing
The changes are contained to the unit tests to `crates/astria-sequencer/mempool/transaction_container.rs`. All unit tests still pass.

### DIFF
--- a/crates/astria-sequencer/src/mempool/transactions_container.rs
+++ b/crates/astria-sequencer/src/mempool/transactions_container.rs
@@ -779,7 +779,7 @@ impl<const MAX_TX_COUNT: usize> TransactionsContainer<ParkedTransactionsForAccou
 }
 
 #[cfg(test)]
-mod tests {
+mod test {
     use astria_core::crypto::SigningKey;
 
     use super::*;
@@ -787,6 +787,7 @@ mod tests {
         denom_0,
         denom_1,
         denom_3,
+        get_alice_signing_key,
         mock_balances,
         mock_state_getter,
         mock_state_put_account_nonce,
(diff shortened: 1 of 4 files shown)


crates/astria-sequencer/src/mempool/transactions_container.rs
```diff
@@ -798,17 +799,72 @@ mod tests {
...
+        }
+
+        fn signer(self, signer: SigningKey) -> Self {
+            Self {
+                signer,
+                ..self
+            }
+        }
+
+        fn denom_0(self, denom_0_cost: u128) -> Self {
```
SuperFluffy, line 848: I had a hard time grokking these. IMO stay closer to the original `cost` of the time marked transaction. I would be able to immediately grep the code base.

You can add these for more elegance:
```rust
fn cost_map(self, cost: HashMap<IbcPrefixed, u128>) -> Self {
    Self {
        cost: cost,
        ..self
    }
}

fn set_cost(self, key: IbcPrefixed, value: u128) -> Self {
    self.cost[key] = value;
    self
}
```
We use this pattern throughout the codebase, so you could also do that:
```rust
fn set_cost<T: Into<IbcPrefixed>>(self, key: T, value: u128) -> Self {
    self.cost[key.into()] = value;
    self
}
```

SuperFluffy (approved):
I am fine with the changes (minus the conventions on the naming, see my comment).
 

Because this is the first time I am reading these tests - my problem with the tests is not necessarily how you construct mocked transactions (although this builder is certainly an improvement), but that I don't really understand what's going on. For example: 

1. variable names like `ttx_s0_0` are context free and I have no idea why it's called that.
2. there are `signer_0` and `signer_1` that get cloned everywhere - why not just use `alice` or `barbara` (probably not the right names, but you use the alice key in the builder; just also use `alice()` instead of `signing_key_1.clone()`.
3. `denom_0`, `denom_1`, `denom_2` also made no sense to me inially. I needed to read the code to understand what's being tested. IMO stay closer to your actual impl and just provide ways to set the actual objects in the map (concerns around safety and panicking don't matter in tests, so we can make those very readable). For the assets, similar to the keys before, I would also provide constants like `nria() -> Denom` or `usdc() -> Denom` or `thefunnyasset() -> Denom` for which you can make use of the fact that `impl From<Denom> for IbcPrefixed`, and then for the mock you do `my_mock.set_cost(usdc(), 12345)`

PR #1608, opened 2024-10-01
feat(sequencer)!: add validator update with name action

### LINKED ISSUE #1590: feat(sequencer): updates of validator names
Currently for sudo driven validator updates, we simply reuse the validator update from cometbft. We should wrap this information inside our own custom action and add the ability to set a name.

In order to do this we will also need to add storage of the validator names within the app state machine. Currently these are only set in genesis, and there is no information app side. We should have this information as well as an ABCI query for it. 

Because `InitChain` ([cometbft docs](https://github.com/cometbft/cometbft/blob/main/spec/abci/abci++_methods.md#initchain)) does not pass this information over to the app side, this could create some oddities of behaivior. It may make sense to pass all validator information through the app state & ignore the validator information stored in genesis to avoid oddities with unset names.

┆Issue Number: ENG-883

ethanoroshiba:
## Summary
Added `ValidatorUpdateWithName` action which wraps a `ValidatorUpdate` and has an additional field for the validator name to be stored app side.

## Background
Validator names could previously only be set in the CometBFT genesis, this is meant to give us a way to set validator names app-side.

## Changes
- Added `ValidatorUpdateWithName` action to proto, core, and sequencer.
- Implemented app-side storage for the validator names corresponding to their address.
- Added ABCI query for validator name.

## Testing
- Added unit test for `ValidatorUpdateWithName` action.
- Added 3 unit tests for `validator_name_request` ABCI query.
- Added `ValidatorUpdateWithName` to `app_execute_transaction_with_every_action_snapshot`

## Breaking Changelist
- Backwards compatible, but added new action.
- Added new action to snapshot test with every action, breaking this snapshot test.

## Related Issues
closes #1590


crates/astria-core/src/protocol/transaction/v1alpha1/action.rs
```diff
@@ -33,6 +33,7 @@ pub enum Action {
     Sequence(SequenceAction),
     Transfer(TransferAction),
     ValidatorUpdate(ValidatorUpdate),
+    ValidatorUpdateWithName(ValidatorUpdateWithName),
```
ethanoroshiba, line 36: Open to calling this action something else if people would like. Maybe something like `NamedValidatorUpdate`

crates/astria-sequencer/src/authority/query.rs
```diff
@@ -0,0 +1,272 @@
...
+        Err(err) => {
+            return error_query_response(
+                Some(err),
+                AbciErrorCode::INTERNAL_ERROR,
+                "failed to get validator names",
+            );
+        }
+    };
+
+    match validator_names.get(&hex::encode(address.address_bytes())) {
```
Fraser999, line 59: Can we avoid hex-encoding the addresses in this collection?

PR #1609, opened 2024-10-01
feat(sequencer)!: make empty transactions invalid 

### LINKED ISSUE #1607: core, sequencer: do not allow transactions with no actions
It was decided that transactions with no actions are invalid and should not be included in blocks or the mempool. Currently these transactions cost nothing and would be a DOS vector. In the future if we want to allow for nonce replacement, allowing other transactions like a transfer of zero balance to serve as a replacement type.

┆Issue Number: ENG-897

Lilyjjo:
## Summary
This PR makes empty transactions (i.e. no actions) invalid. 

## Background
Transactions with no actions cost nothing and could be used to clog the mempool and blocks. In the future if we want a dummy transaction to aid in removing actions via nonce replacement, we could do a noop action like a zero balance transfer. 

## Changes
Made empty transactions invalid during the `UnsignedTransaction` construction process. 

## Testing
Unit test was added.

## Breaking Changelist
Empty transactions are no longer valid. 

## Related Issues

closes #1607

### DIFF
--- a/crates/astria-core/src/protocol/transaction/v1alpha1/action_group/mod.rs
+++ b/crates/astria-core/src/protocol/transaction/v1alpha1/action_group/mod.rs
@@ -153,7 +153,7 @@ enum ErrorKind {
     },
     #[error("attempted to create bundle with non bundleable `ActionGroup` type: {group}")]
     NotBundleable { group: ActionGroup },
-    #[error("attempted to create ActionGroup with empty list of actions")]
+    #[error("actions cannot be empty")]
     Empty,
 }
 
--- a/crates/astria-core/src/protocol/transaction/v1alpha1/action_group/mod.rs
+++ b/crates/astria-core/src/protocol/transaction/v1alpha1/action_group/mod.rs
@@ -134,6 +134,10 @@ impl Error {
             group,
         })
     }
+
+    fn empty() -> Self {
+        Self(ErrorKind::Empty)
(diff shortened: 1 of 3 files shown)


Lilyjjo:
@SuperFluffy I disagree with that the use case of the empty transaction is useful. Typically signing a transaction means you're going to pay for it, and I don't see that we should allow nonce-replacement of a higher fee paying transaction with an empty one

crates/astria-core/src/protocol/transaction/v1alpha1/action_group/mod.rs
```diff
@@ -149,10 +153,13 @@ enum ErrorKind {
     },
     #[error("attempted to create bundle with non bundleable `ActionGroup` type: {group}")]
     NotBundleable { group: ActionGroup },
+    #[error("attempted to create ActionGroup with empty list of actions")]
```
SuperFluffy, line 156: A reader would be confused what `ActionGroup` the error is talking about because fundamentally this is about the actions contained in a transaction not being empty:

```suggestion
#[error("actions cannot be empty")]
```

noot (approved):
looks good to me. i agree that if we implement nonce replacement, we need a way to specify fees (potentially per action, so eg a zero value transfer with a higher fee than the tx being replaced). disallowing empty txs in the meantime is good!

PR #1617, opened 2024-10-03
Feature(cli): Add smoke tests for cli commands

### LINKED ISSUE #1570: Test every CLI command
Our CLI commands are mostly untested - except for those subcommands that are used in smoke tests.

Every subcommand should be tested against a running instance of sequencer.

┆Issue Number: ENG-868

sambukowski:
Add smoke tests for all existing sequencer commands for the cli.
Tests added:
- `account create`
- `address bech32m`
- `account balance`
- `balance get`
- `account nonce`
- `blockheight get`
- `sudo ibc-relayer add`
- `sudo ibc-relayer remove`
- `sudo fee-asset add`
- `sudo fee-asset remove`
- `sudo sudo-address-change`
- `sudo validator-update`
- `transfer`

┆Issue Number: ENG-868

closes #1570

### DIFF
--- a/charts/cli-test.just
+++ b/charts/cli-test.just
@@ -12,7 +12,6 @@ privKey1 := "2bd806c97f0e00af1a1fc3328fa763a9269723c8db8fac4f93af71db186d6e90"
 privKey2 := "dfa7108e38ab71f89f356c72afc38600d5758f11a8c337164713e4471411d2e0"
 ibcAddAddress := "astria13ahqz4pjqfmynk9ylrqv4fwe4957x2p0h5782u"
 pubKey := "88787e29db8d5247c6adfac9909b56e6b2705c3120b2e3885e8ec8aa416a10f1"
-# transferAddress := "astria13ahqz4pjqfmynk9ylrqv4fwe4957x2p0h5782u"
 newPower := "100"
 originalPower := "10"
 sequencerRpc := "http://rpc.sequencer.localdev.me"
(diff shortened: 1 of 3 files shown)


charts/cli-test.just
```diff
@@ -0,0 +1,193 @@
...
+  echo "Output:\n$output"
+  just cli-test validate-output "$output" "$expected_format"
+
+# Test the `bech32m` command
+bech32m tag=defaultTag:
+  #!/usr/bin/env bash
+  set -e 
+
+  # make a new account to get the address bytes
+  output=$(just cli-test command {{ tag }} sequencer account create)
```
quasystaty1, line 56: Maybe should be a default address in bytes configured instead of creating an account? create account might change to print a bech32m prefixed address instead of bytes

charts/cli-test.just
```diff
@@ -0,0 +1,193 @@
...
+deploy:
+  @just deploy cluster
+  @just deploy ingress-controller
+  @just wait-for-ingress-controller
+  @just deploy astria-local
+
+command tag *ARGS:
+  docker run --rm --network host "ghcr.io/astriaorg/astria-cli{{ if tag != '' { replace(':#', '#', tag) } else { '' } }}" {{ ARGS }}
+
+fundedAccount := "astria1rsxyjrcm255ds9euthjx6yc3vrjt9sxrm9cfgm"
```
quasystaty1, line 16: nit: `sudoAccount`, but can be treated as a `fundedAccount`

quasystaty1 (commented):
smoke tests LGTM.

quasystaty1 (approved):

SuperFluffy:
We can't write the smoke tests this way. They just verify if the form of the output is of a certain shape, but not if the commands were actually executed (which is the whole point behind smoke tests).

SuperFluffy (changes requested):
We can't write the smoke tests this way. They just verify if the form of the output is of a certain shape, but not if the commands were actually executed (which is the whole point behind smoke tests).

sambukowski:
This will be reassessed in the future.

Issue #1417, opened 2024-08-28
sequencer: build blocks with transaction category ordering 


Lilyjjo:
For implementing #1412, after finishing #1414 and #1416, the sequencer's block building process can be changed to order transactions served by the mempool by their transaction category.

┆Issue Number: ENG-754

Lilyjjo:
Implemented in #1618

PR #1618, opened 2024-10-03
feat(sequencer)!: enforce block ordering by transaction group 

### LINKED ISSUE #1412: block ordering and tx inclusion action rules

Lilyjjo:
## Summary
We want to order blocks by the transaction categories introduced in #1512 to finish the goals of #1412.

## Background
Certain transactions have the ability to cause other transactions to become invalid. For example, `FeeChangeActions` can cause transactions that were valid to become invalid. For the sake of UX we want to order the potentially-invalidating actions after others so we don't needlessly execute transactions that were just invalidated. More background can be found in #1412. 

## Changes
- The mempool now orders the transactions it feeds to `prepare_proposal` by `ActionGroup`.
- `prepare_proposal` will now skip any transactions in it's block construction process that are out of group order.
- `process_proposal` will now reject any blocks that contain out-of-order transactions.

Note: the mempool does not check for out of order actions inside of the pending queue. This was decided because the amount of times that actions besides `BundleableGeneral` are expected to be ran is low. Instead we let `prepare_proposal` gracefully ignore out-of-order actions in a way that allows them to be included in future blocks.

For example, let's say an account sends these two transactions into the mempool:
- `tx_0: {nonce: 0, action_group: UnbundleableGeneral}`
- `tx_1: {nonce: 1, action_group: BundleableGeneral}`

The mempool will feed these transaction in the order of `{tx_1, tx_0}` to `prepare_proposal`, which is correct on a group ordering level but incorrect on a nonce ordering level. `prepare_proposal` will handle this by skipping `tx_1` and only including `tx_0` in the block. The mempool will re-feed `tx_1` on the next round to `prepare_proposal` to be included. 

## Testing
Unit tests and ran locally. 

## Breaking Changelist
Block that would've passed before with incorrect transaction ordering will now fail.

## Related Issues
closes #1412 #1417


crates/astria-sequencer/src/app/mod.rs
```diff
@@ -713,10 +740,21 @@ impl App {
                     transaction_hash = %telemetry::display::base64(&tx_hash),
                     block_size_constraints = %json(&block_size_constraints),
                     tx_data_bytes = tx_sequence_data_bytes,
-                    "excluding transaction: max block sequenced data limit reached"
+                    "transaction error: max block sequenced data limit passed"
                 );
-                excluded_tx_count = excluded_tx_count.saturating_add(1);
-                continue;
+                return Err(eyre!("max block sequenced data limit passed"));
```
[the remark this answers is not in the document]
Lilyjjo: Added the unit tests checking for the other error cases!

Fraser999 (approved):

crates/astria-composer/src/executor/bundle_factory/mod.rs
```diff
@@ -78,7 +78,7 @@ impl SizedBundle {
     /// Constructs an [`UnsignedTransaction`] from the actions contained in the bundle and `params`.
     /// # Panics
     /// Method is expected to never panic because only `SequenceActions` are added to the bundle,
-    /// which should produce a valid variant of the `ActionGroup` type.
+    /// which should produce a valid variant of the `Action::Group` type.
```
SuperFluffy, line 81:
```suggestion
/// which should produce a valid variant of the [`action::Group`] type.

File: crates/astria-composer/src/executor/bundle_factory/mod.rs [hunk 3]
```diff
@@ -91,8 +91,8 @@ impl SizedBundle {
             .try_build()
             .expect(
                 "method is expected to never panic because only `SequenceActions` are added to \
-                 the bundle, which should produce a valid variant of the `ActionGroup` type; this \
-                 is checked by `tests::transaction_construction_should_not_panic",
+                 the bundle, which should produce a valid variant of the `Action::Group` type; \
```
SuperFluffy, line 94:
```suggestion
the bundle, which should produce a valid variant of the `action::Group` type; \
```

crates/astria-core/src/protocol/transaction/v1alpha1/mod.rs
```diff
@@ -432,11 +442,11 @@ impl UnsignedTransactionBuilder {
     /// Constructs a [`UnsignedTransaction`] from the configured builder.
     ///
     /// # Errors
-    /// Returns an error if the actions do not make a valid `ActionGroup`.
+    /// Returns an error if the actions do not make a valid `Action::Group`.
```
SuperFluffy, line 445:
```suggestion
/// Returns an error if the actions do not make a valid [`action::Group`].
```

crates/astria-sequencer/src/mempool/transactions_container.rs
```diff
@@ -129,28 +134,35 @@ impl fmt::Display for TimemarkedTransaction {
 struct TransactionPriority {
     nonce_diff: u32,
     time_first_seen: Instant,
+    group: Group,
 }
 
 impl PartialEq for TransactionPriority {
```
SuperFluffy, line 140: Isn't this the same as deriving `PartialEq`?

SuperFluffy (commented):
Looks fine to me, going to follow @Fraser999's approval.

I just left some minor comments.

crates/astria-sequencer/src/mempool/transactions_container.rs
Lilyjjo: Can you say more about what you mean by this? I'm a little confused

SuperFluffy: I meant that your manual implementation looks the same as deriving `PartialEq`.

PR #1621, opened 2024-10-03
chore(cli): return Bech32m Prefixed Address

### LINKED ISSUE #1619: chore(cli): bech32m prefixed create account
create account command returns hex address, better UX to return a bech32m prefixed address.

┆Issue Number: ENG-902

quasystaty1:
## Summary
Create account command was returning a hexadecimal address, leading to a less user-friendly experience.

## Changes
- Now returns a Bech32m prefixed address

## Testing
Locally executing the command

## Related Issues

closes #1619

### DIFF
--- a/crates/astria-cli/src/sequencer/account.rs
+++ b/crates/astria-cli/src/sequencer/account.rs
@@ -41,8 +41,9 @@ enum SubCommand {
 
 #[derive(Debug, clap::Args)]
 struct Create {
-    #[command(flatten)]
-    inner: ArgsCreate,
+    /// The address prefix
+    #[arg(long, default_value = "astria")]
+    prefix: String,
 }
 
 impl Create {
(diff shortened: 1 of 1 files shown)


sambukowski (approved):

crates/astria-cli/src/sequencer/account.rs
```diff
@@ -40,19 +40,24 @@ enum SubCommand {
 }
 
 #[derive(Debug, clap::Args)]
-struct Create;
+struct Create {
+    #[command(flatten)]
```
SuperFluffy, line 44: Flattening is not necessary for this type.

It was done for the other arg-types in this module because they use the same arguments (to avoid code duplication).

Remove `ArgsCrate` and fold them into this type.

noot (approved):

PR #1630, opened 2024-10-03
chore(proto, core)!: remove action suffix from all action types

### LINKED ISSUE #1629: Remove Address suffix from protobuf and domain types
All our actions (both on-wire protobuf and domain types) carry a `Action` suffix (except for `ValidatorUpdate`, which is externally defined, and `Ics20Withdrawal`, to be close to the penumbra code we share).

The `Action` suffix is redundant because there is no way to confuse the types for items that are not actions, we don't have a consistency among all actions, and we mostly follow Rust naming conventions, which rejects item names that repeats module names (which is the case for us, as we define the domain types in a module called `actions`).

We should remove the suffix.

SuperFluffy:
## Summary
This patch removes the `*Action` suffix from all protobuf on-wire and domain types that represent sequencer actions.

## Background
The `*Action` suffix is redundant because at no point in the last 12 months was there a risk of confusing action types with non-action types that have a similar name (if there are any even).

Furthermore, we follow Rust naming conventions for types. Since we defined domain types in a module called `action`, clippy's `module_name_repetition` warning triggered on the names, requiring `allow` attributes to quiet the warnings and to stay in sync with the on-wire protobuf types.

## Changes
- Remove the `Action` suffix from all action protobuf and domain types
- Regenerate rust protobuf decoding/encoding code
- Rename `enum FeeChange` to `enum FeeChangeKind` because the former `FeeChangeAction` now clashes with the enum.

## Testing
Only type names were changed. The code still compiles and all tests run. No business logic was changed.

## Breaking Changelist
- This is not a network breaking change because domain type names (or protobuf names) do not manifest in business logic outside of tracing and logging.
- This would be a breaking change for public consumers of our protobuf API.

## Related Issues
Closes #1629


joroshiba (approved):
Protobuf changes LGTM

proto/protocolapis/astria/protocol/transaction/v1alpha1/transaction.proto
```diff
@@ -0,0 +1,35 @@
+syntax = "proto3";
+
+package astria.protocol.transaction.v1alpha1;
+
+import "astria/primitive/v1/types.proto";
+import "astria/protocol/transaction/v1alpha1/actions.proto";
```
joroshiba, line 6: I don't believe you need to import these since they are in the same package.

SuperFluffy: Unfortunately I do. Proof:
```protobuf
import "astria/primitive/v1/types.proto";
// import "astria/protocol/transaction/v1alpha1/actions.proto";
import "astria_vendored/penumbra/core/component/ibc/v1/ibc.proto";
import "astria_vendored/tendermint/abci/types.proto";
import "google/protobuf/any.proto";
```

```sh
proto/protocolapis/astria/protocol/transaction/v1alpha1/transaction.proto:25:12:field astria.protocol.transaction.v1alpha1.UnsignedTransaction.actions: unknown type Action
```

proto/protocolapis/astria/protocol/transactions/v1alpha1/action.proto
```diff
@@ -244,7 +218,7 @@ message BridgeSudoChangeAction {
   string fee_asset = 4;
 }
 
-message FeeChangeAction {
+message FeeChange {
```
joroshiba, line 221: Thinking on the semantics of the fee change given the need for the rust type change from this. Would it make more sense to have this message be something like:

```proto
enum FeeVariable {
  UNKNOWN_FEE_VARIABLE = 0;
  TRANSFER_BASE_FEE = 1;
  ...
}

message FeeChange {
  FeeVariable fee_variable = 1;
  astria.primitive.v1.Uint128 value = 2;
}
```

Then the rust fee variables could be related to the enum and the `FeeChangeKind` might be able to go away? Probably not for this PR though.

joroshiba (approved):
Approved with note on yet another potential fee related change as follow up.

proto/protocolapis/astria/protocol/transactions/v1alpha1/action.proto
SuperFluffy: Discussed offline: generally agree. This action is not nice to read and its semantics are off. But will push that to a different PR.

PR #1636, opened 2024-10-04
fix(cli): ensure checkTx passes before waiting for inclusion

### LINKED ISSUE #1635: fix(cli): handle response on checkTx
`submit_transaction `function should return an error when a getting a non 0 code from the sequencer client upon transaction submission. Currently it waits for [inclusion first](https://github.com/astriaorg/astria/blob/8e10279e0f8c4c2729c77ca9d3f8f0d840f289b3/crates/astria-cli/src/utils.rs#L61).

┆Issue Number: ENG-911

quasystaty1:
## Summary
Ensuring checkTx passes before waiting for inclusion
## Background
When using `submit_transaction_sync` method, cli should exit on `checkTx` error code. Currently, cli first waits for inclusion which will never happen as the transaction failed `checkTx`

## Related Issues

closes #1635

### DIFF
--- a/crates/astria-cli/src/utils.rs
+++ b/crates/astria-cli/src/utils.rs
@@ -58,9 +58,9 @@ pub(crate) async fn submit_transaction(
         .await
         .wrap_err("failed to submit transaction")?;
 
-    let tx_response = sequencer_client.wait_for_tx_inclusion(res.hash).await;
-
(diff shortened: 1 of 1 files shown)


sambukowski (approved):

SuperFluffy (approved):
Change makes sense, thank you.

In the future we should provide flags to perform fire-and-forget submission or wait for inclusion.

PR #1638, opened 2024-10-07
feat(sequencer)!: add limit to total amount of transactions in parked 

Lilyjjo:
## Summary
Adds a limit to the total amount of transactions in the 'parked' mempool structure. We made it configurable as it should change if the CometBFT consensus mempool parameters are modified. 

Sane values would limit the max to less than half of the space in the CometBFT mempool. For example, the CometBFT mempool has the following parameters:
```
  # Maximum number of transactions in the mempool
  size: 2000
  # Maximum size of a single transaction in the mempool
  maxTxBytes: 250000
  # Maximum size of all transactions in the mempool
  maxTxsBytes: 100000000
  cacheSize: 10000
```
It would make sense to set the max pending size to a value like 200 transactions. The max size the 200 transactions could take is  50000000 bytes (200 txs * 250000 max bytes per tx), which is half of the available space in the mempool.

## Background
We want to ensure that there is always space in the mempool for pending transactions so the chain can make progress. 

## Changes
- Added a limit to the number of transactions able to be placed in the parked transaction container structure. 

## Testing
Unit tests and ran locally. 

## Metrics
- Added `TRANSACTIONS_IN_MEMPOOL_PARKED` which gets triggered after successful adds to the parked queue.

## Breaking Changelist
- A new config variable was added to the sequencer's config file: `ASTRIA_SEQUENCER_MEMPOOL_PARKED_MAX_TX_COUNT`
- A new ABCI error code was added: `PARKED_FULL`.


Fraser999 (approved):

quasystaty1 (approved):
Charts LGTM

PR #1639, opened 2024-10-07
feat(cli): Extend Tracing

### LINKED ISSUE #1545: Extend tracing to all of astria-cli
So far only parts of astria-cli use tracing. We want to extend this to the entire crate and all subcommands.

┆Issue Number: ENG-833

sambukowski:
## Summary

## Related Issues
Eng-833

closes #1545


SuperFluffy (changes requested):
The changes as they are don't work: there is no longer anything written to STDOUT since everything gets written to STDERR. That's a problem because one can no longer split both debug information from standard output.

sambukowski:
Not going to go this direction for the cli.

PR #1643, opened 2024-10-08
fix(composer)!: update to work with appside mempool

### LINKED ISSUE #1633: composer: handle new mempool ABCI error codes on transaction submission
PR https://github.com/astriaorg/astria/pull/1515 added three new `ABCI` error codes that affects the composer's nonce resubmission logic. Instead of returning `INVALID_NONCE` on all error codes, we now also return `NONCE_TAKEN`, `ALREADY_PRESENT`, and `ACCOUNT_SIZE_LIMIT`. The composer should handle these other transaction submission error codes as well. 

Error code descriptors:
`ALREADY_PRESENT`: the exact transaction is already inside of the mempool and doesn't need to be re-sent.
`ACCOUNT_SIZE_LIMIT`: the parked queue for the account is full, this would indicate that the account associated with the composer isn't having sent transactions qualify for the pending/execution queue. This could be due to a nonce gap or the account lacking funds to cover transaction costs.
`NONCE_TAKEN`: there is already a transaction for this account with the used nonce inside of the mempool. We don't allow for transaction replacement right now, so a higher nonce should probably be used.

┆Issue Number: ENG-909

Lilyjjo:
## Summary
Updates composer to work with the new appside mempool. New ABCI error codes were added and now nonces can be stacked, so the submission logic should use the GRPC `pending_nonce()` instead of the RPC `get_latest_nonce()`.

## Background
New ABCI error codes were added in PR https://github.com/astriaorg/astria/pull/1515 which need to be handled in the composer's submission logic. 

New Error codes:
- `NONCE_TAKEN` -> try resubmitting with new top-of-pending. If this is happening because the composer is out of funds, that is okay as the composer has it's own fund monitoring logic
- `PARKED_FULL` -> normal error case ok
- `ACCOUNT_SIZE_LIMIT` -> normal error case ok 
- `ALREADY_PRESENT` -> normal error case ok, shouldn't be returned 

## Background
We changed the internals of the mempool which resulted in the users of the mempool to have changes as well.

## Testing
Smoke tests

## Related Issues
closes #1633


itamarreif (changes requested):
Please also add a black box test for this error. 

https://github.com/astriaorg/astria/blob/main/crates/astria-composer/tests/blackbox/geth_collector.rs#L86

SuperFluffy (changes requested):
The abci error codes are only one part in addressing https://github.com/astriaorg/astria/issues/1633.

This patch also needs to change how composer fetches the next nonce to use (since right it makes use of cometbft's `get_nonce`, which is not aware of the mempool).

Lilyjjo:
note: not ready for review yet

charts/composer/templates/configmap.yaml
```diff
@@ -8,7 +8,8 @@ data:
   ASTRIA_COMPOSER_API_LISTEN_ADDR: "0.0.0.0:{{ .Values.ports.healthApi }}"
   ASTRIA_COMPOSER_GRPC_ADDR: "0.0.0.0:{{ .Values.ports.grpc }}"
   ASTRIA_COMPOSER_SEQUENCER_CHAIN_ID: "{{ tpl .Values.config.sequencerChainId . }}"
-  ASTRIA_COMPOSER_SEQUENCER_URL: "{{ tpl .Values.config.sequencerRpc . }}"
+  ASTRIA_COMPOSER_SEQUENCER_HTTP_URL: "{{ tpl .Values.config.sequencerRpc . }}"
```
itamarreif, line 11: we call this `COMETBFT_ENDPOINT` in other places

crates/astria-composer/src/executor/mod.rs
```diff
@@ -493,14 +506,22 @@ async fn get_latest_nonce(
             },
         );
     let res = tryhard::retry_fn(|| {
-        let client = client.clone();
-        let span = info_span!(parent: span.clone(), "attempt get nonce");
+        let mut client = client.clone();
```
itamarreif, line 509: why is this mut

crates/astria-composer/src/executor/mod.rs
```diff
@@ -493,14 +506,22 @@ async fn get_latest_nonce(
...
+                })
+                .await
+                .map(|rsp| rsp.into_inner().inner)
+        }
+        .instrument(span)
     })
     .with_config(retry_config)
     .await
-    .wrap_err("failed getting latest nonce from sequencer after 1024 attempts");
+    .wrap_err("failed getting pending nonce from sequencing after 1024 attempts");
```
itamarreif, line 524: "from sequencer" instead of "from sequencing"?

crates/astria-composer/tests/blackbox/geth_collector.rs
```diff
@@ -127,18 +99,57 @@ async fn invalid_nonce_causes_resubmission_under_different_nonce() {
 
     // wait for 1 sequencer block time to make sure the bundle is preempted
     tokio::time::timeout(
-        Duration::from_millis(test_composer.cfg.block_time_ms),
+        Duration::from_millis(test_composer.cfg.block_time_ms + 1000),
```
itamarreif, line 102: should probably remove the `+ 1000` here?

charts/composer/templates/configmap.yaml
Lilyjjo: updated the name in the code as well

crates/astria-composer/src/executor/mod.rs
Lilyjjo: not sure, but the generate code that it's from needs it mut

crates/astria-composer/tests/blackbox/helper/mock_grpc_sequencer.rs
```diff
@@ -0,0 +1,120 @@
...
+            })
+        };
+        Self {
+            _server: server,
+            mock_server,
+            local_addr,
+        }
+    }
+
+    pub(crate) async fn mount_pending_nonce_response(
```
itamarreif, line 73: this should probably return a mock guard

crates/astria-composer/tests/blackbox/geth_collector.rs
```diff
@@ -103,22 +80,17 @@ async fn invalid_nonce_causes_resubmission_under_different_nonce() {
...
-    )
-    .await;
-
     let expected_rollup_ids = vec![RollupId::from_unhashed_bytes("test1")];
     // Expect nonce 1 again so that the resubmitted tx is accepted
     let valid_nonce_guard =
         mount_broadcast_tx_sync_mock(&test_composer.sequencer, expected_rollup_ids, vec![1]).await;
 
+    // Mount a response of 1 to a nonce query
+    test_composer
```
itamarreif, line 89: should get a mock guard from the mount helper and then wait for it to be satisfied

itamarreif (commented):
the pending nonce mock should use a mockguard and `wait_until_satisfied` as the mount is used several times in the test (e.g. for initializing nonce on startup and then on refetch after the invalid/taken nonce errors). besides that, just a couple nits on the error message and config variable names.

charts/composer/values.yaml
```diff
@@ -19,7 +19,8 @@ config:
   maxSubmitInterval: 2000
   sequencerAddressPrefix: astria
   sequencerNativeAssetBaseDenomination: "nria"
-  sequencerRpc: ""
+  cometbftEndpoint: "http://node0-sequencer-rpc-service.astria-dev-cluster.svc.cluster.local:26657"
```
quasystaty1, line 22: Should template from `evm-stack` chart, set value to `""` in `composer` chart, use global values of the `evm-stack` to override, similar to other values.

charts/composer/templates/configmap.yaml
```diff
@@ -8,7 +8,8 @@ data:
   ASTRIA_COMPOSER_API_LISTEN_ADDR: "0.0.0.0:{{ .Values.ports.healthApi }}"
   ASTRIA_COMPOSER_GRPC_ADDR: "0.0.0.0:{{ .Values.ports.grpc }}"
   ASTRIA_COMPOSER_SEQUENCER_CHAIN_ID: "{{ tpl .Values.config.sequencerChainId . }}"
-  ASTRIA_COMPOSER_SEQUENCER_URL: "{{ tpl .Values.config.sequencerRpc . }}"
+  ASTRIA_COMPOSER_COMETBFT_ENDPOINT: "{{ tpl .Values.config.cometbftEndpoint . }}"
```
quasystaty1, line 11: Should move under `global.dev` statements, marking a breaking change.

quasystaty1 (changes requested):

SuperFluffy (commented):
I'd like to have a few (user facing) names changed, but otherwise this looks great.

itamarreif (commented):
Went over the deleted tests and made an issue for adding the coverage back as black box tests: https://github.com/astriaorg/astria/issues/1652

SuperFluffy (approved):
Thanks for the fix!

The tests in the `executor` module are in a weird state and should be part of the blackbox tests instead. As they are testing relatively low impact functionality as per @itamarreif's notes, I am in favor of removing them here and following up in a future patch.

charts/composer/templates/configmap.yaml
```diff
@@ -30,6 +29,8 @@ data:
   OTEL_EXPORTER_OTLP_TRACE_HEADERS: "{{ tpl .Values.otel.traceHeaders . }}"
   OTEL_SERVICE_NAME: "{{ tpl .Values.otel.serviceName . }}"
   {{- if not .Values.global.dev }}
+  ASTRIA_COMPOSER_SEQUENCER_ABCI_ENDPOINT: "{{ tpl .Values.config.sequencerRpc . }}"
```
quasystaty1, line 32: when `.Values.global.dev` sets to true it marks a breaking change to configs. Therefore, new env variables should be added to `else` bracket while values being removed should be added to `{{- if not .Values.global.dev }} ` bracket.

quasystaty1 (changes requested):

charts/composer/templates/configmap.yaml
```diff
@@ -31,6 +30,8 @@ data:
   OTEL_SERVICE_NAME: "{{ tpl .Values.otel.serviceName . }}"
   {{- if not .Values.global.dev }}
```
quasystaty1, line 31: nit: Should move removed values to mark breaking changes (see previous [comment](https://github.com/astriaorg/astria/pull/1643/files#r1797634184))
```suggestion
  {{- if not .Values.global.dev }}
  ASTRIA_COMPOSER_SEQUENCER_URL: "{{ tpl .Values.config.sequencerRpc . }}"
```

itamarreif (approved):

PR #1646, opened 2024-10-08
fix(sequencer): fix app hash in horcrux sentries

joroshiba:
## Summary
Changes the comparison when deciding if we need to re-execute a proposal to use the validator address and the timestamp as identifiers, to enable horcrux sentry nodes to not have app hash clashes. 

## Background
When running horcrux all sentry nodes act as the same validator and all nodes create a new proposal, but the signer only signs one of them. This creates issues since all sentry nodes except the signed node will not reexecute the proposal since it was created by their validator address even though the proposals are different. Timestamps are used to add additional fingerprinting to this without adding high computation cost.

## Changes
- Utilize timestamp as part of identifying whether to clear and execute proposal in prepare proposal

## Testing
A validator reported this issue when running testnet, and similar changes were tested and validated.

### DIFF
--- a/crates/astria-sequencer/src/app/mod.rs
+++ b/crates/astria-sequencer/src/app/mod.rs
@@ -187,8 +187,8 @@ pub(crate) struct App {
     mempool: Mempool,
 
     // TODO(https://github.com/astriaorg/astria/issues/1660): The executed_proposal_fingerprint and
-    // executed_proposal_hash fields should be stored in the ephemeral storage instead of on the app
-    // struct, to avoid any issues with forgetting to reset them.
+    // executed_proposal_hash fields should be stored in the ephemeral storage instead of on the
+    // app struct, to avoid any issues with forgetting to reset them.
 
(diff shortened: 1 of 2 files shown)


crates/astria-sequencer/src/app/mod.rs
```diff
@@ -136,6 +136,38 @@ const POST_TRANSACTION_EXECUTION_RESULT_KEY: &str = "post_transaction_execution_
...
+            timestamp: proposal.time,
+        }
+    }
+
+    fn from_process_proposal(proposal: &abci::request::ProcessProposal) -> Self {
+        Self {
+            validator_address: proposal.proposer_address,
+            timestamp: proposal.time,
+        }
+    }
```
Fraser999, line 168: These could both just be `From` impls, but not a big deal.

Fraser999 (approved):

Fraser999 (approved):

PR #1647, opened 2024-10-08
feat(sequencer): rework all fees

ethanoroshiba:
## Summary

## Background

## Changes

## Testing
How are these changes tested?

## Metrics

## Breaking Changelist

## Related Issues

closes


Fraser999 (approved):
Approval from me, but please wait for approval from @SuperFluffy too.

Fraser999 (approved):

crates/astria-sequencer/src/bridge/storage/keys.rs
```diff
@@ -96,9 +91,6 @@ mod tests {
 
     #[test]
     fn keys_should_not_change() {
-        insta::assert_snapshot!(INIT_BRIDGE_ACCOUNT_BASE_FEE);
```
SuperFluffy, line 99: It's hard to track what exactly changed because all the tests get renumbered.

Shouldn't do it in this PR, but give unique names to all the snapshot asserts in a future PR.

crates/astria-sequencer/src/fees/mod.rs
```diff
@@ -0,0 +1,390 @@
...
+#[instrument(skip_all, err(level = Level::WARN))]
+async fn check_and_pay_fees<S: StateWrite, T: FeeHandler + Protobuf>(
+    act: &T,
+    base: u128,
+    multiplier: u128,
+    mut state: S,
+    fee_asset: &asset::Denom,
+) -> eyre::Result<()> {
+    let total_fees = base
+        .checked_add(act.variable_component().saturating_mul(multiplier))
```
SuperFluffy, line 331: If we do saturing arithmetic for the multiplication, I think we can also do saturating arithmetic here.

crates/astria-sequencer/src/fees/mod.rs
```diff
@@ -0,0 +1,390 @@
...
+    let source_action_index = transaction_context.source_action_index;
+
+    ensure!(
+        state
+            .get_account_balance(&from, fee_asset)
+            .await
+            .wrap_err("failed to get account balance")?
+            >= total_fees,
+        "insufficient funds for transfer and fee payment",
+    );
```
SuperFluffy, line 347: This check is unnecessary and can be removed. `decrease_balance` below returns an error if there are not enough funds.

I suggest to remove this line, and move `state.decrease_balance` under the `state.is_allowed_fee_asset` check.

crates/astria-sequencer/src/fees/state_ext.rs
```diff
@@ -0,0 +1,807 @@
...
+    fn get_block_fees(&self) -> Result<Vec<Fee>> {
+        let mut block_fees = self.object_get(keys::BLOCK);
+        match block_fees {
+            Some(_) => {}
+            None => {
+                block_fees = Some(vec![]);
+            }
+        }
+        Ok(block_fees.expect("block fees should not be `None` after populating"))
+    }
```
SuperFluffy, line 61:
```suggestion
    fn get_block_fees(&self) -> Vec<Fee> {
        self.object_get(keys::BLOCK)
              .unwrap_or_default()
    }
```

crates/astria-sequencer/src/fees/storage/keys.rs
```diff
@@ -0,0 +1,73 @@
...
+                writeln!(formatter, "{BRIDGE_UNLOCK}")?;
+                writeln!(formatter, "{BRIDGE_SUDO_CHANGE}")?;
+                writeln!(formatter, "{IBC_RELAY}")?;
+                writeln!(formatter, "{VALIDATOR_UPDATE}")?;
+                writeln!(formatter, "{FEE_ASSET_CHANGE}")?;
+                writeln!(formatter, "{FEE_CHANGE}")?;
+                writeln!(formatter, "{IBC_RELAYER_CHANGE}")?;
+                writeln!(formatter, "{SUDO_ADDRESS_CHANGE}")?;
+                writeln!(formatter, "{IBC_SUDO_CHANGE}")?;
+                Ok(())
```
SuperFluffy, line 50: I think I'd personally prefer 14 snapshot files to be honest, because it makes reviewing changes much easier.

Either way, these should be sorted alphabetically.

crates/astria-sequencer/src/fees/storage/snapshots/astria_sequencer__fees__storage__keys__tests__keys_should_not_change.snap
```diff
@@ -0,0 +1,19 @@
+---
+source: crates/astria-sequencer/src/fees/storage/keys.rs
+assertion_line: 48
+expression: Helper
```
SuperFluffy, line 4: This `Helper` gives us one snapshot file, but at the cost of no longer having a pinpoint to which exact line failed.

If we want to have snapshot files, then we should buy in completely and actually have 14 files and the exact expressions that failed.

Alternativeliy, we can store the expected results inline:
```rust
insta::assert_snapshot!(the_expression, @"the_result");
```
For example:
```rust:
insta::assert_snapshot!(IBC_SUDO_CHANGE, @"fees/ibc_sudo_change");
```

SuperFluffy (commented):
Great work, I am very happy with these changes!

I left a few minor comments here and there, some of which would be great to have addressed before merging (especially error message on which failed was invalid when validating the fees genesis object in `astria-core`).

The big thing before giving my last approval is `FEE_ASSET_PREFIX` and the `fee_asset` constructor in the `assets` module: now that we have a `fees` component, these feel misplaced. I noted it there, but in my mind the `assets` module/component should only be concerned with which assets the sequencer chain knows (and how ibc-prefixed assets map back to trace-prefixed ones).

I think it feels right that it's the `fees` component's prerogative to keep track over and decide which fee asset is permitted or not.

proto/protocolapis/astria/protocol/transactions/v1alpha1/action.proto
```diff
@@ -244,10 +246,5 @@ message IbcSudoChange {
 // Response to a transaction fee ABCI query.
 message TransactionFeeResponse {
```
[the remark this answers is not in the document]
ethanoroshiba: Now in `fees/v1alpha1/types.proto`

proto/protocolapis/astria/protocol/genesis/v1alpha1/types.proto
```diff
@@ -40,12 +41,22 @@ message IbcParameters {
...
 
-message Fees {
-  astria.primitive.v1.Uint128 transfer_base_fee = 1;
-  astria.primitive.v1.Uint128 sequence_base_fee = 2;
-  astria.primitive.v1.Uint128 sequence_byte_cost_multiplier = 3;
-  astria.primitive.v1.Uint128 init_bridge_account_base_fee = 4;
-  astria.primitive.v1.Uint128 bridge_lock_byte_cost_multiplier = 5;
-  astria.primitive.v1.Uint128 bridge_sudo_change_fee = 6;
-  astria.primitive.v1.Uint128 ics20_withdrawal_base_fee = 7;
+message GenesisFees {
```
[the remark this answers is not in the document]
ethanoroshiba: Personally, I feel that it makes the most sense to keep `GenesisFees` within the genesis proto, just since these fees should be strictly associated with the genesis state, but I'd be happy to move if you feel strongly about it.

proto/protocolapis/astria/protocol/fees/v1alpha1/types.proto
```diff
@@ -0,0 +1,80 @@
...
+package astria.protocol.fees.v1alpha1;
+
+import "astria/primitive/v1/types.proto";
+
+message TransactionFee {
+  string asset = 1;
+  astria.primitive.v1.Uint128 fee = 2;
+}
+
+message TransferFeeComponents {
```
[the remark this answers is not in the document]
ethanoroshiba: Resolved offline

crates/astria-sequencer/src/fees/mod.rs
```diff
@@ -61,10 +60,9 @@ const DEPOSIT_BASE_FEE: u128 = 16;
 
 #[async_trait::async_trait]
 pub(crate) trait FeeHandler {
-    async fn check_and_pay_fees<S: StateWrite>(
-        &self,
-        mut state: S,
-    ) -> astria_eyre::eyre::Result<()>;
+    async fn handle_fees_if_present<S: StateWrite>(&self, state: S) -> eyre::Result<()>;
+
+    fn computed_cost_base_component(&self) -> u128;
```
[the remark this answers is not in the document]
ethanoroshiba: Resolved offline.

crates/astria-sequencer/src/fees/mod.rs
```diff
@@ -87,374 +85,357 @@ impl Fee {
...
+        )
+        .await
+    }
 
-        state.record(deposit_abci_event);
-        Ok(())
+    #[instrument(skip_all)]
+    fn computed_cost_base_component(&self) -> u128 {
+        base_deposit_fee(&self.asset, &self.destination_chain_address)
+            .expect("deposit fee calculation should not fail")
```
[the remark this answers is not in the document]
ethanoroshiba: Resolved offline.

crates/astria-sequencer/src/fees/mod.rs
```diff
@@ -87,374 +85,357 @@ impl Fee {
...
+    Ok(())
 }
 
 /// Returns a modified byte length of the deposit event. Length is calculated with reasonable values
 /// for all fields except `asset` and `destination_chain_address`, ergo it may not be representative
 /// of on-wire length.
-pub(crate) fn calculate_base_deposit_fee(deposit: &Deposit) -> Option<u128> {
-    deposit
-        .asset
+#[cfg(test)]
```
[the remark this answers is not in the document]
ethanoroshiba: Resolved offline.

crates/astria-core/src/protocol/genesis/v1alpha1.rs
```diff
@@ -630,15 +750,20 @@ impl FeesError {
...
+    }
 }
 
 #[derive(Debug, thiserror::Error)]
 #[error("failed ensuring invariants of {}", Fees::full_name())]
 enum FeesErrorKind {
     #[error("field was not set: `{name}`")]
     FieldNotSet { name: &'static str },
+    #[error("conversion of fee components failed")]
+    FeeComponentsConversion(#[source] FeeComponentError),
```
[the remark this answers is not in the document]
ethanoroshiba: The field which failed is contained within the source `FeeComponentError`'s inner `FeeComponentErrorKind`. So, it should still be printed out when printing the source chain. I think it may be better to leave it this way, since this more accurately reflects the level at which the validation failure occurred.

crates/astria-sequencer/src/accounts/storage/snapshots/astria_sequencer__accounts__storage__keys__tests__keys_should_not_change-2.snap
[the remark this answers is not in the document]
ethanoroshiba: Followup issue, perhaps?

crates/astria-sequencer/src/authority/action.rs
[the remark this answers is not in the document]
ethanoroshiba: Are you referring to moving the `ActionHandler` impls for the fee-related actions to the `fees` component, or moving all `ActionHandler` impls to a central location?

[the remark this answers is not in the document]
SuperFluffy: Yes to both. :) In the current structure it feels odd that we change things related to fees in `authority`. It should be in `fees`. I think this change can be made now.

But the `ActionHandler` trait in general should be in a central location, and all its impls should also be in a central location (but that's a change that should be done separate from this PR).

crates/astria-sequencer/src/accounts/storage/snapshots/astria_sequencer__accounts__storage__keys__tests__keys_should_not_change-2.snap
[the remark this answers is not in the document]
SuperFluffy: Yeah, sounds good (we should do this for all the tests).

crates/astria-sequencer/src/test_utils.rs
```diff
@@ -3,6 +3,8 @@ use astria_core::primitive::v1::{
     Address,
     Bech32,
 };
+#[cfg(test)]
```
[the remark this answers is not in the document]
ethanoroshiba: `test_utils` is also enabled for the `benchmark` feature, in which case this import (and the associated function) are not used, so the `cfg(test)` attribute is necessary I believe. This is part of what I was aiming to fix in #1613

crates/astria-sequencer/src/authority/action.rs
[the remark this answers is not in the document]
ethanoroshiba: Sounds good, I've moved both `FeeAssetChange` and `FeeChange` into the fees component

crates/astria-sequencer/src/accounts/storage/snapshots/astria_sequencer__accounts__storage__keys__tests__keys_should_not_change-2.snap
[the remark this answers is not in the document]
ethanoroshiba: https://github.com/astriaorg/astria/issues/1656

crates/astria-sequencer/src/bridge/storage/keys.rs
ethanoroshiba: Tracking issue: https://github.com/astriaorg/astria/issues/1656

crates/astria-sequencer/src/fees/storage/keys.rs
SuperFluffy: It will require several hops figuring it out. Don't view this from the point of view of the author but from the point of view of the reader who needs to match lines in the git diff to the lines in the rust test.

It's a bad experience.

crates/astria-core/src/protocol/fees/v1alpha1.rs
```diff
@@ -10,6 +10,7 @@ use crate::{
 #[error("failed validating on-wire type `{on_wire}` as domain type")]
 pub struct FeeComponentError {
     on_wire: String,
+    #[source]
```
SuperFluffy, line 13: nit: you can just rename `inner` to `source` and then you don't need to tag it with `#[source]`

crates/astria-sequencer/src/fees/storage/keys.rs
```diff
@@ -12,7 +23,31 @@ pub(in crate::fees) const FEE_CHANGE: &str = "fees/fee_change";
 pub(in crate::fees) const IBC_RELAYER_CHANGE: &str = "fees/ibc_relayer_change";
 pub(in crate::fees) const SUDO_ADDRESS_CHANGE: &str = "fees/sudo_address_change";
 pub(in crate::fees) const IBC_SUDO_CHANGE: &str = "fees/ibc_sudo_change";
-pub(in crate::fees) const BLOCK: &str = "fees/block_fees";
+pub(in crate::fees) const BLOCK: &str = "fees/block"; // NOTE: `BLOCK` is only used in the ephemeral store.
+pub(in crate::fees) const FEE_ASSET_PREFIX: &str = "fees/fee_asset/";
```
SuperFluffy, line 27: Let's rename this to a more explicit (no need to duplicate `fee` everywhere):
```suggestion
pub(in crate::fees) const ALLOWED_ASSET_PREFIX: &str = "fees/allowed_asset/";
```

crates/astria-sequencer/src/fees/storage/keys.rs
```diff
@@ -12,7 +23,31 @@ pub(in crate::fees) const FEE_CHANGE: &str = "fees/fee_change";
...
+
+pub(in crate::fees) fn fee_asset<'a, TAsset>(asset: &'a TAsset) -> String
```
SuperFluffy, line 29: Assuming this is only used for allowed fee assets (Which I think it is)
```suggestion
pub(in crate::fees) fn allowed_asset<'a, TAsset>(asset: &'a TAsset) -> String
```

crates/astria-sequencer/src/fees/state_ext.rs
```diff
@@ -475,6 +503,26 @@ pub(crate) trait StateWriteExt: StateWrite {
...
+
+    #[instrument(skip_all)]
+    fn put_allowed_fee_asset<'a, TAsset>(&mut self, asset: &'a TAsset) -> Result<()>
+    where
+        &'a TAsset: Into<Cow<'a, asset::IbcPrefixed>>,
+    {
+        let bytes = StoredValue::Unit
+            .serialize()
+            .context("failed to serialize unit for allowed fee asset")?;
+        self.nonverifiable_put_raw(keys::fee_asset(asset).into_bytes(), bytes);
```
SuperFluffy, line 523: Discussed offline: this should be placed into the verifiable storage, not non-verifiable. (changing this will trigger changes to all gets/reads).

crates/astria-sequencer/src/fees/storage/keys.rs
```diff
@@ -71,28 +66,21 @@ mod tests {
...
-                writeln!(formatter, "{INIT_BRIDGE_ACCOUNT}")?;
-                writeln!(formatter, "{SEQUENCE}")?;
-                writeln!(formatter, "{SUDO_ADDRESS_CHANGE}")?;
-                writeln!(formatter, "{TRANSFER}")?;
-                writeln!(formatter, "{VALIDATOR_UPDATE}")?;
-                Ok(())
-            }
-        }
-        insta::assert_snapshot!(Helper);
+        assert_snapshot!("bridge_lock_fees_key", BRIDGE_LOCK);
```
SuperFluffy, line 69: all of these are so nice (looking at the names of the snapshots in your next commit)

SuperFluffy (commented):
I have made a few last minute observations.

crates/astria-sequencer/src/fees/storage/keys.rs
ethanoroshiba: 😁

crates/astria-sequencer/src/fees/storage/keys.rs
```diff
@@ -18,34 +18,34 @@ pub(in crate::fees) const BRIDGE_UNLOCK: &str = "fees/bridge_unlock";
 pub(in crate::fees) const BRIDGE_SUDO_CHANGE: &str = "fees/bridge_sudo_change";
 pub(in crate::fees) const IBC_RELAY: &str = "fees/ibc_relay";
 pub(in crate::fees) const VALIDATOR_UPDATE: &str = "fees/validator_update";
-pub(in crate::fees) const FEE_ASSET_CHANGE: &str = "fees/fee_asset_change";
+pub(in crate::fees) const FEE_ASSET_CHANGE: &str = "fees/allowed_asset_change";
```
SuperFluffy, line 21: This one should probably remain:
```suggestion
pub(in crate::fees) const FEE_ASSET_CHANGE: &str = "fees/fee_asset_change";
```

SuperFluffy (approved):
Fantastic, great work!

Before merging, please update the PR text to reflect that allowed assets are now also placed in verifiable storage.

charts/sequencer/files/cometbft/config/genesis.json
```diff
@@ -3,13 +3,62 @@
...
+        "multiplier": {{ include "sequencer.toUint128Proto" .Values.genesis.fees.sudoAddressChange.multiplier }}
+      },
+      "transfer": {
+        "base": {{ include "sequencer.toUint128Proto" .Values.genesis.fees.transfer.base }},
+        "multiplier": {{ include "sequencer.toUint128Proto" .Values.genesis.fees.transfer.multiplier }}
+      },
+      "validator_update": {
+        "base": {{ include "sequencer.toUint128Proto" .Values.genesis.fees.validatorUpdate.base }},
+        "multiplier": {{ include "sequencer.toUint128Proto" .Values.genesis.fees.validatorUpdate.multiplier }}
+      }
```
joroshiba, line 61: These changes should all be wrapped in a check for if in `.Values.global.dev` mode, which indicates we are running on latest main vs latest release.

In general we have done this format, which I know seems kinda backwards:

```
    <anything that should be deleted after upgrade>
    {{- else }}
    <anything that should be standard after upgrade only included in dev mode right now>
    {{- end }}
```

charts/sequencer/files/cometbft/config/genesis.json
```diff
@@ -3,13 +3,72 @@
...
-      "bridge_sudo_change_fee": {{ include "sequencer.toUint128Proto" .Values.genesis.fees.bridgeSudoChangeFee }},
-      "ics20_withdrawal_base_fee": {{ include "sequencer.toUint128Proto" .Values.genesis.fees.ics20WithdrawalBaseFee }}
+      {{- if not .Values.global.dev }}
+        "transfer_base_fee": {{ include "sequencer.toUint128Proto" .Values.genesis.fees.transferBaseFee }},
+        "sequence_base_fee": {{ include "sequencer.toUint128Proto" .Values.genesis.fees.sequenceBaseFee }},
+        "sequence_byte_cost_multiplier": {{ include "sequencer.toUint128Proto" .Values.genesis.fees.sequenceByteCostMultiplier }},
+        "init_bridge_account_base_fee": {{ include "sequencer.toUint128Proto" .Values.genesis.fees.initBridgeAccountBaseFee }},
+        "bridge_lock_byte_cost_multiplier": {{ include "sequencer.toUint128Proto" .Values.genesis.fees.bridgeLockByteCostMultiplier }},
+        "bridge_sudo_change_fee": {{ include "sequencer.toUint128Proto" .Values.genesis.fees.bridgeSudoChangeFee }},
+        "ics20_withdrawal_base_fee": {{ include "sequencer.toUint128Proto" .Values.genesis.fees.ics20WithdrawalBaseFee }}
```
joroshiba, line 13:
```suggestion
        "transfer_base_fee": {{ include "sequencer.toUint128Proto" .Values.genesis.fees.transfer.base }},
        "sequence_base_fee": {{ include "sequencer.toUint128Proto" .Values.genesis.fees.sequence.base }},
        "sequence_byte_cost_multiplier": {{ include "sequencer.toUint128Proto" .Values.genesis.fees.sequence.multiplier }},
        "init_bridge_account_base_fee": {{ include "sequencer.toUint128Proto" .Values.genesis.fees.initBridgeAccount.base }},
        "bridge_lock_byte_cost_multiplier": {{ include "sequencer.toUint128Proto" .Values.genesis.fees.bridgeLock.multiplier }},
        "bridge_sudo_change_fee": {{ include "sequencer.toUint128Proto" .Values.genesis.fees.bridgeSudoChange.base }},
        "ics20_withdrawal_base_fee": {{ include "sequencer.toUint128Proto" .Values.genesis.fees.ics20Withdrawal.base }}
```

crates/astria-sequencer/src/bridge/storage/keys.rs
joroshiba: Related..... I don't believe this is removing the snapshots which are no longer used. Because the numbers changed and we just have fewer.

ethanoroshiba: I deleted and regenerated all the snapshots in `fees/storage/snapshots`, so hopefully they should all be up to date?

crates/astria-sequencer/src/fees/component.rs
```diff
@@ -0,0 +1,119 @@
...
+
+        let sudo_address_change_fees = app_state.fees().sudo_address_change;
+        state
+            .put_sudo_address_change_fees(sudo_address_change_fees)
+            .wrap_err("failed to store sudo address change fee components")?;
+
+        let ibc_sudo_change_fees = app_state.fees().ibc_sudo_change;
+        state
+            .put_ibc_sudo_change_fees(ibc_sudo_change_fees)
+            .wrap_err("failed to store ibc sudo change fee components")?;
```
joroshiba, line 99: Should it fail if we have not set a single asset? It seems like the only action which needs to be required to be set in genesis (and thus enabled) is the FeeChange itself? If we can't set the fee then we can't do it for any other actions either?

crates/astria-core/src/protocol/genesis/v1alpha1.rs
```diff
@@ -542,80 +559,183 @@ impl From<raw::IbcParameters> for IBCParameters {
...
+    pub bridge_lock: BridgeLockFeeComponents,
+    pub bridge_unlock: BridgeUnlockFeeComponents,
+    pub bridge_sudo_change: BridgeSudoChangeFeeComponents,
+    pub ibc_relay: IbcRelayFeeComponents,
+    pub validator_update: ValidatorUpdateFeeComponents,
+    pub fee_asset_change: FeeAssetChangeFeeComponents,
+    pub fee_change: FeeChangeFeeComponents,
+    pub ibc_relayer_change: IbcRelayerChangeFeeComponents,
+    pub sudo_address_change: SudoAddressChangeFeeComponents,
+    pub ibc_sudo_change: IbcSudoChangeFeeComponents,
```
joroshiba, line 576: Should most of these be `Option`? It seems the only which stricly must be set is `fee_change`?

joroshiba (approved):
Approval for infra, couple notes on the genesis that I'm not blocking on.

crates/astria-core/src/protocol/genesis/v1alpha1.rs
ethanoroshiba: Tracking issue: https://github.com/astriaorg/astria/issues/1662

PR #1651, opened 2024-10-11
chore!: use URL-safe alphabet for base64 encoding (ENG-913)

### LINKED ISSUE #1649: Provide core-utils crate containing base64 helpers
We want to use a URL-safe alphabet for all cases where we base64-encode values.

To that end, we should provide a new crate named `astria-core-utils` which will provide opinionated helpers to support base64 encoding and decoding, i.e. the helpers will exclusively use a [URL-safe alphabet and include padding](https://docs.rs/base64/latest/base64/engine/general_purpose/constant.URL_SAFE.html) internally and offer no other options to the caller.

This work should also include replacing all existing instances of base64-encoding/-decoding with the new helpers, and should remove the `base64` crate from all manifests other than the one for the new core-utils crate.

┆Issue Number: ENG-913

Fraser999:
## Summary
This changes all base64-encoding to use the URL-safe alphabet.

## Background
We want our base64-encoded data to be URL-safe, and this was not previously the case.

## Changes
- Provided a new crate `astria-core-utils` with a single public module `base64`.  This module provides a Display/Debug helper for formatting byte strings while avoiding allocations, and simple `encode` and `decode` functions.  Further functions can be provided in the future if required (e.g. encoding to a buffer).
- All direct usages of the `base64` crate and of `telemetry::display::base64` have been replaced by using the new functions/type...
- ... except in `astria-sequencer-utils`, where the input to the blob-parser will be standard base64.  Hence, in that crate only, we still read the input data as standard base64, but all outputs are made using the URL-safe encoding from `astria-core-utils`
- The `pbjson` dependency was changed to use [a fork](https://github.com/astriaorg/pbjson/tree/url-safe), which has the effect of making the generated serde impls of the protobuf types also use the URL-safe alphabet when encoding.  (`pbjson` already supports decoding from URL-safe-encoded data).

## Testing
Unit tests were added.

## Breaking Changelist
- This affects the RollupIDs in the sequencer-relayer's config file, as they need to be URL-safe encoded now.

## Related Issues
Closes #1649.


crates/astria-core-utils/src/base64/error.rs
```diff
@@ -0,0 +1,79 @@
...
+    fmt::{
+        self,
+        Display,
+        Formatter,
+    },
+};
+
+/// Errors that can occur while decoding.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum Error {
```
SuperFluffy, line 12: Let's not make the variants part of the public interface for now? If this just becomes a `pub struct Error(InnerError)` we also don't need to wrap all the `base64` decode errors.

Also call this `DecodeError`?

crates/astria-core-utils/src/base64/error.rs
```diff
@@ -0,0 +1,79 @@
...
+                len,
+            } => write!(formatter, "invalid input length {len}"),
+            Self::Padding => write!(formatter, "padding absent or invalid"),
+        }
+    }
+}
+
+impl error::Error for Error {}
+
+impl From<base64::DecodeError> for Error {
```
SuperFluffy, line 62: I think this unfortunately makes `base64::DecodeError` part of our public interface.

crates/astria-core/Cargo.toml
```diff
@@ -19,8 +19,9 @@ keywords = ["astria", "grpc", "rpc", "blockchain", "execution", "protobuf"]
 bech32 = "0.11.0"
 brotli = { version = "5.0.0", optional = true }
 celestia-types = { version = "0.1.1", optional = true }
-pbjson = { version = "0.6.0", optional = true }
+pbjson = { version = "0.6.0", git = "https://github.com/astriaorg/pbjson.git", branch = "url-safe", optional = true }
```
SuperFluffy, line 22: Looks like we can't bump to `0.7.0` yet, because we also need `pbjson-types`, which itself requires `prost 0.12`. :-/

Cargo.toml
SuperFluffy, line 1: `pbjson-types` is listed as a workspace dependency here. Instead of making our fork of pbjson a direct dependency, should we make it a patch pointing to our branch?

crates/astria-core/src/primitive/v1/mod.rs
```diff
@@ -90,7 +87,7 @@ impl Protobuf for merkle::Proof {
 pub struct RollupId {
```
SuperFluffy, line 87: `RollupId` is actually a concept that we have a wiretype for. For consistency, we should fix this serde derivation and point it to `astria.primitive.v1.RollupId`.

crates/astria-core/src/primitive/v1/mod.rs
```diff
@@ -923,6 +921,18 @@ mod tests {
         let _ = address.into_raw();
     }
 
+    #[test]
+    fn pbjson_base64_should_be_url_safe() {
```
SuperFluffy, line 925: Is this test necessary at this level (given the premise that downstream users should be able to rely on the correct implementation of upstream libraries)?

crates/astria-sequencer-utils/Cargo.toml
```diff
@@ -9,7 +9,7 @@ repository = "https://github.com/astriaorg/astria"
 homepage = "https://astria.org"
 
 [dependencies]
-base64 = { workspace = true }
+base64 = "0.22.1"
```
SuperFluffy, line 12: I guess the base64 encoded blob data we feed into the blob parser command is not controlled by us?

SuperFluffy (commented):
This is great. I have left some comments on how to get our fork of pbsjon into this crate, with an eye toward `pbsjon-types` still using the upstream `pbjson`.

I also think we can avoid making the base64 error part of the public API for now.

crates/astria-core/src/primitive/v1/mod.rs
Fraser999: I think tests like these are worthwhile anywhere we don't just do a standard `derive(Serialize)`, since it'd be easy to regress otherwise.

This is even more the case now we're going to be using a less obvious way to serialize the domain `RollupId` type.

Potentially other snapshot tests would catch this, but I don't like to rely on those to test this functionality, since that too may change over time.

crates/astria-core/src/primitive/v1/mod.rs
Fraser999: After offline discussion, agreed to leave this as is due to our usage of `RollupId`s as keys in maps which are JSON-encoded using the serde trait.  This fails if using the wiretype serialization since the JSON map requires the key to be a string, and making the suggested change means a JSON-encoded `RollupId` is a JSON object rather than a string.

Fraser999:
Closing pending discussion around the scope of changes for this update.

PR #1654, opened 2024-10-12
feat(cli): implement frost_ed25519 threshold signing cli

### LINKED ISSUE #833: sequencer: investigate ed25519 threshold signing libraries for bridge accounts
as title says - bridge accounts are just an address (ie normal ed25519 pub/priv key). we decided not to implement native multisig accounts for now, so we should determine a way to have a bridge account controlled by an off-chain multisig via threshold signatures.

noot:
## Summary
implement threshold signing cli using the frost_ed25519 library.

## Background
can be used for turning sequencer accounts into multisigs without any changes to the sequencer itself.

## Changes
- implement `dkg`, `sign`, and `verify` commands
- see README in module for info on usage

## Testing
i ran it manually and verified the signature.

## Related Issues

closes #833


crates/astria-cli/src/sequencer/threshold/sign.rs
```diff
@@ -0,0 +1,274 @@
...
+struct CommitmentsWithIdentifier {
+    identifier: Identifier,
+    commitments: frost_ed25519::round1::SigningCommitments,
+}
+
+#[derive(Debug, clap::Args)]
+struct PrepareMessage {
+    /// message to be signed
+    #[arg(long)]
+    message: String,
```
joroshiba, line 113: Do we need something that is not a string for the message?

This message is generally going to be the transaction we are signing right? At some point it needs to become a protobuf serialized `TransactionBody` (formerly `UnsignedTransaction`)? 

In the simplest form we could have a message which is just pbjson via a file read? Then we might want to add a way to generate txs without signing and output the pbjson form of them, but not that's more better ux, we could generate that ourselves relatively simply?

crates/astria-cli/src/sequencer/threshold/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+4. The coordinator aggregates all the signature shares from the previous step:
+
+```sh
+cargo run -- sequencer threshold sign aggregate \
+--public-key-package-path frost.pub \
+--signing-package-path signing_package.out
+```
+
+The output is an ed25510 signature, eg:
```
joroshiba, line 90:
```suggestion
The output is an ed25519 signature, eg:
```

crates/astria-cli/src/sequencer/threshold/sign.rs
noot: updated this to take a file that contains the message bytes. I could add an option to deserialize the message as a `TransactionBody` and display that before signing?

> This message is generally going to be the transaction we are signing right?

yes, I was thinking of updating the other CLI commands to have an option to output the `TransactionBody` into a file, which is then used as the input to this command. 

I also updated `aggregate` to have an option to output the signed message as a `Transaction` (signed) that can then be submitted to the sequencer.

Fraser999 (changes requested):
Mostly requests for more feedback to the user when things go wrong, but there are a couple of points which I think are blockers:
* looks like a bug in the generation of a `SignedTransaction` as a file
* I think if `verify` fails, the process should exit with a non-zero return code

crates/astria-cli/src/sequencer/threshold/README.md
```diff
@@ -0,0 +1,126 @@
...
+ Note that the coordinator commands are interactive and require inputs
+ generated by the participants in the preceding steps.
+
+1. Each participant runs part 1:
+
+```sh
+cargo run -- sequencer threshold sign part1 \
+--secret-key-package-path frost_1.priv \
+--nonces-path nonces_1.out
+```
```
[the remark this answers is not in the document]
noot: thank you, i was wondering why that wasn't working lol

crates/astria-cli/src/sequencer/threshold/dkg.rs
```diff
@@ -0,0 +1,188 @@
...
+
+        // round 1
+        let (round1_secret_package, round1_public_package) =
+            dkg::part1(id, max_signers, min_signers, rng).wrap_err("failed to run dkg part1")?;
+        let round1_public_package_with_id = Round1PackageWithIdentifier {
+            identifier: id,
+            package: round1_public_package,
+        };
+        println!(
+            "Send our public package to all other participants: {}",
```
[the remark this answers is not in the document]
noot: updated to match the outputs to be copied in `sign.rs`

crates/astria-cli/src/sequencer/threshold/sign.rs
joroshiba, line 1: In general throughout PR we are taking strings for paths vs using something like `PathBuf` as done in the [GenesisExample](https://github.com/astriaorg/astria/blob/main/crates/astria-sequencer-utils/src/genesis_example.rs#L217). Not sure on implications but noting the inconsistency.

joroshiba (dismissed):

Fraser999 (approved):
Other than the indentation in the README being off, no other issues remaining - certainly nothing blocking, so approving assuming the README will be fixed.

crates/astria-cli/Cargo.toml
```diff
@@ -13,6 +13,12 @@ name = "astria-cli"
 
 [dependencies]
 color-eyre = "0.6"
+# v2.0.0-rc.0 - can be updated once https://github.com/ZcashFoundation/frost/issues/755 is closed
+frost-ed25519 = { git = "https://github.com/ZcashFoundation/frost.git", rev = "f95aeb22d1e7ede968f7916340c66ebb9683ae50", features = [
+] }
```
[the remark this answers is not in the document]
noot: updated!

Fraser999 (approved):

PR #1655, opened 2024-10-14
fix(sequencer)!: allow compat prefixed addresses when receiving ics20 transfers

### LINKED ISSUE #1653: ICS-20 Transfers to sequencer should support compat addresses
Currently, when performing an ICS-20 transfer to the Sequencer, the transfer only accepts bech32m address formats. The sequencer should accept compatible addresses (bech32).

┆Issue Number: ENG-915

quasystaty1:
## Summary
Support ics20 transfer from a counterparty to the astria sequencer.

## Background
Currently, when performing an ics20 transfer to the Sequencer, the transfer only accepts bech32m address formats. This is inconsistent with refunding failed ics20 from Sequencer to the counterparty, where both compat (bech32) and normal (bech32m) addresses are accepted. Sequencer should accept both.

The main use case is when receiving assets from Noble (such as USDC). While Noble only enforces bech32 when handling failed ics20 transfers, this still requires ics20 transfers from Astria Sequencer to Noble to be bech32 (and not bech32m) encoded. The bech32 compatible address is hence the sender address that will be known by the counterparty/Noble, and we must allow transfers of funds back to it.

## Changes
- sequencer now accepts ics20-transfers with compat bech32 addresses
- relevant tests includes prefixes in storage

## Testing
All test passes

## Breaking Changelist
- sequencer will now accept ics-transfer with bech32 addresses, this is breaking as previously failed transfers will now be accepted

## Related Issues
closes #1653

### DIFF
--- a/charts/ibc-test.just
+++ b/charts/ibc-test.just
@@ -70,6 +70,9 @@ run tag=defaultTag:
     exit 1
   fi
 
+  # Execute the transfer from Celstia to sequencer with compat address
+  just ibc-test _do-compat-ibc-transfer
+
   # check that celestia balance updated correctly
   for i in {1..50}
   do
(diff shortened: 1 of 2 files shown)


SuperFluffy (commented):
Thank you for the fix. This was an oversight.

I have changed the patch title to from `feat` to `fix` and adjusted the text slightly.

Before approving: note that at the moment we completely lack tests for ics20 transfers. Since you originally encountered this problem when trying to do ics20 transfers, can you confirm that these now work?

Better yet, is there a way to provide a smoke test for this? (I am happy with you doing a manual confirmation if adding smoke tests is too much of a lift).

joroshiba (approved):
Infra approval

quasystaty1:
Added a smoke test

Issue #1668, opened 2024-10-16
Remove source channel from ics20 withdrawal action


SuperFluffy:
The `source_channel` field in `astria.protocol.transactions.v1alpha1.Ics20Withdrawal` is not necessary: it must be part of the asset to be withdrawn. In fact, `astria-bridge-contracts` is parsing it out from the asset and placing it in the action.

┆Issue Number: ENG-926

SuperFluffy:
This issue is not actionable, the source channel is necessary because of the following scenarios:

1. transfer of sequencer native assets.
2. transfer of non-native assets to yet another chain (for example, tia bridged to sequencer that is being sent to osmosis).

The only case in which source channel is not needed is if a wrapped asset (an asset for which sequencer is the sink zone) is returned back to its source (so that sequencer can pop the leading port/channel pair and determine the channel from there).

PR #1669, opened 2024-10-16
chore(proto)!: remove source channel field from ics20 withdrawal action

SuperFluffy:
## Summary
Removes the field `astria.protocol.transaction.v1alpha1.Ics20Withdrawal.source_channel`.

## Background
The source channel of an asset in an ics20 transfer is part of the assets name. To take the example of sequencer, if a counterparty sends an asset to sequencer, sequencer will prepend the asset by `<dest_port>` and `<dest_channel>`. This is the port/channel pair on sequencer for the connection with the counterparty. The `source_channel` field then is actually this `<dest_channel>` that astria is prepending to the asset before storing it locally.

Looking at the `astria-bridge-contracts` library (also changed in this patch), `source_channel` is actually parsed out of the asset to be transferred and then placed into the action. This is not necessary.

Instead, sequencer will now be concerned with determining the source channel from the asset. If the asset is an ibc-prefixed asset, it will look up its full trace in its state.

This allows rollups to withdraw assets even if only the ibc prefixed denom is known.

## Changes
- Remove field `astria.protocol.transaction.v1alpha1.Ics20Withdrawal.source_channel`.
- Remove all source channel logic from astria cli, astria bridge contracts
- Have sequencer determine the source channel from the asset to be withdrawn.

## Testing
This must be tested through a smoke test.

## Breaking Changelist
The shape of an action was changed. This is a network breaking change.

## Related Issues
Closes https://github.com/astriaorg/astria/issues/1668.

### DIFF
--- a/crates/astria-bridge-contracts/src/lib.rs
+++ b/crates/astria-bridge-contracts/src/lib.rs
@@ -71,18 +71,6 @@ impl BuildError {
             field,
         })
     }
-
-    #[must_use]
-    fn ics20_asset_without_channel() -> Self {
-        Self(BuildErrorKind::Ics20AssetWithoutChannel)
-    }
-
-    #[must_use]
-    fn parse_ics20_asset_source_channel(source: ibc_types::IdentifierError) -> Self {
-        Self(BuildErrorKind::ParseIcs20AssetSourceChannel {
-            source,
-        })
-    }
 }
 
 #[derive(Debug, thiserror::Error)]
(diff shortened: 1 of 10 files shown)


noot:
we can't remove this field because we can't withdraw native tokens without it, as they aren't prefixed due to originating on the sequencer

SuperFluffy:
This PR works only when unwrapping an IBC asset. This PR however breaks the following use cases:

1. transfer of sequencer native assets.
2. transfer of non-native assets to yet another chain (for example, tia bridge to sequencer that is being sent to osmos).

PR #1671, opened 2024-10-16
feat(bridge-withdrawer): add `use_compat_address` configuration value

### LINKED ISSUE #1424: Bridge withdrawer, contracts, CLI need to understand when to demand compat addresses
More information will be filled in after https://github.com/astriaorg/astria/issues/1407 is completed.

┆Issue Number: ENG-760

noot:
## Summary
add configuration option for `use_compat_address`, which can then be set by the operator of the bridge withdrawer.

## Background
required for withdrawals from a rollup to noble, in case the withdrawal on noble times out or fails.

## Changes
- add configuration option for `use_compat_address`

## Testing
existing unit tests, would need to have a withdrawal timeout test for noble to fully test this

## Related Issues

closes #1424


crates/astria-bridge-contracts/src/lib.rs
```diff
@@ -271,6 +274,13 @@ where
                     .parse()
                     .map_err(BuildError::parse_ics20_asset_source_channel)?,
             );
+
+            // USDC from Noble requires using the bech32 compat address as the sender/return
+            // address.
+            //
+            // since we're always unwrapping the asset (sending back to the originating chain), we
+            // can check that the base denom is USDC, and if it is, we know we're sending to Noble.
+            use_compat_address = ics20_asset_to_withdraw.base_denom() == "usdc";
```
SuperFluffy, line 283: Instead of hardcoding this here, I'd prefer to add a method `fn use_compat_address(self, use_compat_address) -> Self` on the builder and let the user of this crate decide this.

SuperFluffy (changes requested):
I think `astria-bridge-contracts` is the wrong place to decide if the compat address should be used or not. For one, what if there are other denominations that require a compat address? Also, what if I want to send wrapped usdc to another chain that is not noble (and capable of dealing with bech32m)?

The contracts crate should allow specifying compatibility mode, but through its builder not via a heuristic.

Instead, let's move this check to `astria-bridge-withdrawer` - and provide a new flag for the `astria-cli bridge collect-withdrawals` subcommand.

noot:
@SuperFluffy updated the PR to add a configuration value to the bridge withdrawer

crates/astria-bridge-withdrawer/src/config.rs
```diff
@@ -26,6 +26,8 @@ pub struct Config {
     pub rollup_asset_denomination: asset::denom::TracePrefixed,
     // The bridge address corresponding to the bridged rollup asset on the sequencer.
     pub sequencer_bridge_address: String,
+    // Whether to use compat addresses for `Ics20Withdrawal`s.
+    pub use_compat_address: bool,
```
SuperFluffy, line 30: We have a one-withdrawer-per-asset (and destination) situation, right?

SuperFluffy (approved):
I think these changes are better from the point of view of the builder.

You have decided to forego all heuristics, right? So this means that one would spin up one bridge-withdrawer per asset and target, and set the compat address there?


Before merging, can you please also provide a similar config to astria-cli in the `bridge::collect` submodule?

crates/astria-bridge-withdrawer/src/config.rs
noot: yeah exactly, the withdrawer is one instance per bridge/asset, so the withdrawer for noble usdc would set this to true for example

crates/astria-cli/src/bridge/collect.rs
```diff
@@ -72,6 +72,9 @@ pub(super) struct Command {
     /// rollup's bridge configuration for that asset.
     #[arg(long)]
     bridge_address: Address,
+    /// Use compat addresses for `Ics20Withdrawal`s.
```
SuperFluffy, line 75:
```suggestion
/// Use compat addresses when withdrawing assets to chains that require bech32 addresses (like for noble USDC).
```

crates/astria-bridge-withdrawer/local.env.example
```diff
@@ -48,6 +48,9 @@ ASTRIA_BRIDGE_WITHDRAWER_ROLLUP_ASSET_DENOMINATION="nria"
 # Should match the bridge address in the geth rollup's bridge configuration for that asset.
 ASTRIA_BRIDGE_WITHDRAWER_SEQUENCER_BRIDGE_ADDRESS=""
 
+# Whether to use compat addresses for `Ics20Withdrawal`s.
+ASTRIA_BRIDGE_WITHDRAWER_USE_COMPAT_ADDRESS=false
```
quasystaty1, line 52: Should add the new env variable to `evm-bridge-withdrawer` chart configmap.

quasystaty1 (changes requested):

charts/evm-bridge-withdrawer/templates/configmaps.yaml
```diff
@@ -12,6 +12,7 @@ data:
   ASTRIA_BRIDGE_WITHDRAWER_SEQUENCER_ADDRESS_PREFIX: "{{ .Values.config.sequencerAddressPrefix }}"
   ASTRIA_BRIDGE_WITHDRAWER_SEQUENCER_GRPC_ENDPOINT: "{{ tpl .Values.config.sequencerGrpcEndpoint . }}"
   ASTRIA_BRIDGE_WITHDRAWER_SEQUENCER_BRIDGE_ADDRESS: "{{ .Values.config.sequencerBridgeAddress }}"
+  ASTRIA_BRIDGE_WITHDRAWER_USE_COMPAT_ADDRESS: "{{ .Values.config.useCompatAddress }}"
```
quasystaty1, line 15: Should put inside {{ else }} arm to mark as breaking changes.

quasystaty1 (approved):

PR #1672, opened 2024-10-16
chore(proto)!: upgrade to proto v1s throughout

joroshiba:
## Summary
Upgrades the protos throughout stack which are not currently being developed to v1. Removed generation of old rust clients, but duplicated protos.


crates/astria-core/src/protocol/transaction/v1/snapshots/astria_core__protocol__transaction__v1__tests__transaction_id.snap
```diff
@@ -3,5 +3,5 @@ source: crates/astria-core/src/protocol/transaction/v1alpha1/mod.rs
 expression: tx.id().to_raw()
 ---
 {
-  "inner": "68687a46955d1f96040e0007ead684941c708e3af37ee2400869aa485f4c7007"
```
joroshiba, line 6: Calling out that this snap changes due to the change in protobuf.any field name

buf.yaml
joroshiba, line 1: Updating the services that external parties may communicate with to be based on `package` instead of just `wire_json` for breaking checks.

quasystaty1 (approved):
Charts LGTM

Fraser999 (approved):
LGTM.  I'm assuming proto/sequencerblockapis/astria/sequencerblock/v1alpha1/optimistic_block.proto isn't ready to make it into `v1` yet?

SuperFluffy (approved):
Looks good to me to get merged (minus the optimistic block bits, with a nod to @Fraser999's comment).

Separate from all this: for the astria-core module structure I think we should get rid of all the `v1` module paths. I doubt we will ever support 2 different APIs in one crate, and the `::v1::` everywhere is noisy.

Instead, we could link "link" versions to certain APIs and backport fixes, if necessary.

But that's a discussion that outside the scope of this PR.

PR #1673, opened 2024-10-16
chore(sequencer): remove unused enable mint env

quasystaty1:
## Summary
remove unused enable mint env. We removed the mint module in https://github.com/astriaorg/astria/pull/1134, the config var was unused but left in.


joroshiba (approved):

SuperFluffy:
We should have also removed it from env: https://github.com/astriaorg/astria/blob/30ef51e9a463bf5425bb3399f8958406c03a6fad/crates/astria-sequencer/local.env.example#L9-L12

PR #1675, opened 2024-10-17
chore(charts): update back to latest after proto merge

joroshiba:
## Summary
We merged major proto updates, dependent charts have been updated in latest and releases to use those.


quasystaty1 (approved):
LGTM

PR #1680, opened 2024-10-17
chore(charts): disable locals in geth

joroshiba:
## Summary
Updates default geth deployement to disable locals pricing

## Background
Treating txs as local is really intended for a truly local instance, where the source is the person running the node but in our case treats every tx coming through RPC as local. We don't want this.


quasystaty1 (approved):
approved with small comment to pass lint checks

PR #1684, opened 2024-10-17
chore(proto): reserved only for what would be breaking change

joroshiba:
## Summary
We have been using reserved to highlight gaps, but reserved should ideally only be utilized for reserving previously used identifiers. 

## Changes
- removes reserved numbers that using were not actual breaking change
- updates workflow to allow override and to always run all breaking checks even if previous fails

## Breaking Changes
This has no real breaking changes, hence allowing override. The removal of the reserved makes it so that we don't have to do this in the future when adding new fields.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -14,7 +14,6 @@ jobs:
     runs-on: ubuntu-latest
     needs: run_checker
     if: needs.run_checker.outputs.run_lint_proto == 'true'
-    continue-on-error: true
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


ethanoroshiba (approved):
LGTM. Looks like the way the lint workflow is setup, the individual check may still fail but allow the overall status checks to pass, which I hope should be a good flag for anyone reviewing breaking proto changes in the future. Still, I think we may want to flag this to the broader engineering group as a "this label should probably almost never be used" thing.

noot (approved):

WafflesVonMaple (approved):

PR #1690, opened 2024-10-18
chore(core, sequencer): add names to all snapshot tests

### LINKED ISSUE #1656: add snapshot names to all snapshot asserts
As noted [here](https://github.com/astriaorg/astria/pull/1647#discussion_r1799439267), many snapshots throughout the codebase do not have an associated name, and are hence based on line number. If assertions are moved around, then the snapshots will break, which should not happen. We should add names to all snapshot tests.

┆Issue Number: ENG-916

ethanoroshiba:
## Summary
Added names to all snapshot tests which did not have them.

## Background
If a certain test had multiple snapshot assertions, they would be named based on their ordering. Hence any reordering, removal, or addition of the snapshots would break the test, which should not be the case. This change aims to fix this problem as well as establish a best practice of naming snapshots moving forward.

## Changes
- Added names to all snapshot tests and regenerated all snapshots not within Conductor (due to code freeze)

## Testing
Passing all tests.

## Related Issues
closes #1656


noot (approved):

crates/astria-core/src/primitive/v1/snapshots/astria_core__primitive__v1__tests__compat_address.snap
```diff
@@ -1,5 +1,5 @@
 ---
 source: crates/astria-core/src/primitive/v1/mod.rs
-expression: "&bech32_address"
+expression: "&compat_address"
```
SuperFluffy, line 3: And for this, add a bit `compat_bech32_non_m_address`

crates/astria-core/src/primitive/v1/mod.rs
```diff
@@ -845,15 +845,15 @@ mod tests {
             .prefix(ASTRIA_ADDRESS_PREFIX)
             .try_build()
             .unwrap();
-        insta::assert_json_snapshot!(&main_address.to_raw());
+        insta::assert_json_snapshot!("main_address", &main_address.to_raw());
```
SuperFluffy, line 848: Let's never name this `main_bech32m_address`

SuperFluffy (approved):
The changes make for much cleaner files

crates/astria-core/src/primitive/v1/mod.rs
ethanoroshiba: I'm assuming you meant to name it `main_bech32m_address`?

joroshiba (approved):

PR #1691, opened 2024-10-18

noot:
## Summary
patch penumbra temporarily so that height is passed from hermes's requests to the ibc grpc server.

## Background
there was an IBC proof validation failure error on dusk-10. the computed app hash was one height *after* the expected app hash, which could be caused by the grpc queries always using the latest state snapshot, instead of at the height intended by hermes. however the grpc requests don't contain the height so the hack is to put the height in the grpc request metadata, which is then used by the grpc query server to use the correct snapshot.

this should probably be upstreamed into penumbra as this is actually the recommended solution by the ibc-go team (https://github.com/cosmos/ibc-go/pull/7303)

## Changes
- patch penumbra dep to https://github.com/astriaorg/penumbra/pull/2 temporarily

## Testing
ibc smoke test


SuperFluffy:
Please create the fork under the astriaorg organization and point this patch to https://github.com/astriaorg/penumbra

noot:
@SuperFluffy done

Fraser999 (approved):

noot:
closing in favour of #1856

PR #1693, opened 2024-10-18
feat(bridge-withdrawer): metric to track total settled funds

joroshiba:
## Summary
Add a metric to track the total funds moving through the bridge withdrawer.

## Background
It may be useful to keep track of the total funds moving through for alerting purposes, building a purpose built tool for this would be annoying and we can easily track with a metric here.

## Metrics
- `sequencer_block_settled_value`

### DIFF
--- a/crates/astria-bridge-withdrawer/src/metrics.rs
+++ b/crates/astria-bridge-withdrawer/src/metrics.rs
@@ -18,6 +18,7 @@ pub struct Metrics {
     nonce_fetch_latency: Histogram,
     sequencer_submission_failure_count: Counter,
     sequencer_submission_latency: Histogram,
+    sequencer_block_settled_value: Gauge,
 }
 
 impl Metrics {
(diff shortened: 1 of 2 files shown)


crates/astria-bridge-withdrawer/src/metrics.rs
```diff
@@ -44,6 +45,21 @@ impl Metrics {
...
+
+    #[expect(
+        clippy::cast_precision_loss,
+        reason = "metric with potential loss of precision, noted with a log warning when it occurs"
+    )]
+    pub(crate) fn set_sequencer_block_settled_value(&self, value: u128) {
+        let max_u32: u128 = u32::MAX.into();
+        if value > max_u32 {
+            tracing::warn!(
+                "sequencer_block_settled_value is being set with a value that exceeds u32::MAX, \
```
SuperFluffy, line 57: Just add the value that you are casting here. Might be interesting to look at.

Also use the const name instead of writing it out in the message (`SEQUENCER_BLOCK_...`) in case we rename it. Can be an event field or just a part `{}` in the format args.

crates/astria-bridge-withdrawer/src/metrics.rs
```diff
@@ -121,6 +146,7 @@ mod tests {
         NONCE_FETCH_LATENCY,
         SEQUENCER_SUBMISSION_FAILURE_COUNT,
         SEQUENCER_SUBMISSION_LATENCY,
+        SEQUNCER_BLOCK_SETTLED_VALUE,
```
SuperFluffy, line 149: Typo: `SEQUENCER_BLOCK_SETTLED_VALUE`

SuperFluffy (approved):
Looks mostly fine but please address comments prior to merging

Fraser999 (approved):

PR #1699, opened 2024-10-21
chore(charts): celestia versions bump

quasystaty1:
## Summary
updates celestia-node and celestia-local charts to support the new versions.
## Background
Celestia mocha-4 testnet went through an upgrade, nodes should upgrade by nov 7th

## Changes
- celestia-app `v2.0.0 -> v2.3.1`
- celestia-node `v0.16.0-rc -> v0.18.3-mocha`
- adjusts celestia-node` config.toml` env variables
- adds pruner service to configs
## Testing
celestia-local tested locally running smoke-tests, celestia-node sync tested locally & against flame.

## Breaking Changelist
- celestia environment variable changes

### DIFF
--- a/charts/celestia-node/values.yaml
+++ b/charts/celestia-node/values.yaml
@@ -20,7 +20,7 @@ config:
 
 images:
   pullPolicy: IfNotPresent
-  node: ghcr.io/celestiaorg/celestia-node:v0.18.2-mocha
+  node: ghcr.io/celestiaorg/celestia-node:v0.18.3-mocha
 
 ports:
   celestia:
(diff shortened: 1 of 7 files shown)


charts/celestia-node/values.yaml
```diff
@@ -19,7 +20,7 @@ config:
 
 images:
   pullPolicy: IfNotPresent
-  node: ghcr.io/celestiaorg/celestia-node:v0.16.0-rc0
+  node: ghcr.io/celestiaorg/celestia-node:v0.18.2-mocha
```
joroshiba, line 23: Do any of these settings only work on mocha? Noting that we will have use a different image for mainnet instances.

WafflesVonMaple (approved):

PR #1703, opened 2024-10-21
feat(core, sequencer)!: make native asset optional

SuperFluffy:
## Summary
Makes setting a native asset optional.

## Background
Native assets are a remnant of sequencer that is now deeply entrenched in how it runs (whitebox) tests. While we don't want to set a native assets, in mainnet, it is too big a lift to remove them at this point.

## Changes
- Change genesis validation to permit empty native assets.
- Update account chain init to only require native assets if accounts are set.
- Remove native asset lookup from querying assets (this can be done via looking up all ibc assets.

## Testing
The absence of native assets is not tested as of this patch. All tests still function with native assets set.

A proper test would follow these steps:
1. Spin up a test net without native assets, accounts, or permitted fee assets set
2. Bridge in an IBC asset via an ics20 transfer
3. Set a permitted fee asset to the bridged asset
4. Execute transactions.

## Breaking Changelist
This is marked as breaking since a node with before this change would reject an empty native asset in its genesis while a node after this patch would initialize successfully.

However, this is mild breakage because both nodes would not form a network in the first place.


crates/astria-sequencer/src/accounts/component.rs
```diff
@@ -29,14 +29,16 @@ impl Component for AccountsComponent {
...
-            .wrap_err("failed to read native asset from state")?;
-        for account in app_state.accounts() {
-            state
-                .put_account_balance(&account.address, &native_asset, account.balance)
-                .wrap_err("failed writing account balance to state")?;
+        if !app_state.accounts().is_empty() {
+            let native_asset = state
+                .get_native_asset()
+                .await
+                .wrap_err("failed to read native asset from state")?;
```
joroshiba, line 36: Will this fail if we didn't set the native asset?

SuperFluffy: Yes: we only set the native asset during init chain. It is our only way to get the native asset into the state.

So if we don't set it, then this call will return an error.

crates/astria-sequencer/src/accounts/query.rs
```diff
@@ -55,21 +55,12 @@ async fn get_trace_prefixed_account_balances<S: StateRead>(
     let stream = state
         .account_asset_balances(address)
         .map_ok(|asset_balance| async move {
-            let native_asset = state
```
SuperFluffy, line 58: Note on this change: getting the native asset here was never necessary and unnecessarily complicates the code.

joroshiba (approved):

PR #1707, opened 2024-10-21
chore(proto): move optimistic block protos to v1

itamarreif:
## Summary
This updates the protos used for the auctioneer to be based on the new v1 protos.

## Background
We bumped our proto definition versions from `v1alpha_` to `v1`.

## Changes
- Moved the `sequencerblock/optimistic_block` file into its own subpackage in order to have `sequencerblock.optimisticblock.v1alpha1`
- Updated to use `FilteredSequencerBlock` from v1

## Breaking Changelist
- This shouldn't break anything as none of these protos are used by anything in main currently.

closes https://github.com/astriaorg/astria/issues/1767


noot (commented):
auctioneer isn't shipped yet, maybe this should stay v1alpha1 for now?

itamarreif:
> auctioneer isn't shipped yet, maybe this should stay v1alpha1 for now?

The auctioneer stuff remains v1alpha1 here, it is just updated to use the correct (v1) version of everything else.

see: https://github.com/astriaorg/astria/pull/1707/files#diff-816642b078e8e91f48aba99b3f04ffaed46960ef63a4c8d22ae068ffebd9d6f7R3

SuperFluffy:
Have to push back on this naming convention. v1 sequencer block with v1alpha1 optimistic block? We should have probably not nested the optimistic block bits under sequencerblock in this way, but provided a structure like in the protocol APIs.

I am not sure how to best tackle this, but I feel like going with `astria.optimisticblock.v1alpha1` (i.e. a sibling instead of a child) that imports `astria.sequencerblock.v1` is more natural.

Would like to pull @joroshiba into this.

proto/sequencerblockapis/astria/sequencerblock/optimisticblock/v1alpha1/optimistic_block.proto
SuperFluffy, line 1: Please call this file `service.proto`, because it only contains services.

joroshiba (commented):
LGTM, think this is good sane versioning for the proto. Agreed with Janis on naming `service.proto` as a nit.

joroshiba (approved):

SuperFluffy:
This is unfortunately lacking an entry in `CHANGELOG.md` and also imports the wrong code.

Issue #1563, opened 2024-09-25
Sequencer-side implementation for Optimistic Block


itamarreif:
In order for the Auctioneer to drive optimistic block execution in Geth, it needs to receive a stream of optimistic blocks and block commitments. As described in the [PR introducing this API](https://github.com/astriaorg/astria/pull/1519):

>The Auctioneer receives blocks optimistically from the sequencer in order to maximize the auction duration.
>
>After a block is committed, the Auctioneer will submit the highest paying bundle for inclusion by the sequencer.
>
>### Optimistic Block Stream
>1. The Sequencer will stream blocks to the Auctioneer optimistically, i.e. before they are finalized (ProcessProposal).
>2. The Sequencer will also stream block commitments to the Auctioneer, which mark blocks as final thanks to CometBFT's single slot finality.

This functionality needs to be optional, as only one sequencer node in the network needs to be serving this stream (i.e. the sequencer node run alongside the Auctioneer, but not any of the ones connected to a Composer).

```[tasklist]
- [ ] https://github.com/astriaorg/astria/pull/1519
- [ ] Implement the sequencer-side server for the optimistic block stream
- [ ] Link to PR with the auctioneer-side implementation that uses the client for this
```

┆Issue Number: ENG-847

joroshiba:
This issue is stale because it has been open 45 days with no activity. Remove stale label or this issue
be closed in 7 days.

SuperFluffy:
Subsuming this under https://github.com/astriaorg/astria/issues/1533 and closing this issue

PR #1709, opened 2024-10-22
feat: Implement Optimistic block grpc streams

bharath-123:
## Summary
This PR implements the Optimistic block grpc stream apis protos of which were defined in https://github.com/astriaorg/astria/pull/1519. These stream apis are used by auctioneer in order to optimistically simulate incoming bundles which are competing to be placed at the top-of-block for the given rollup.

## Background
We have use cases where users of a given rollup would want to simulate a bundle on top of the latest block. The issue with simulating for rollup users is that the sequencer block times are too small and it would be difficult to simulate a bundle on top of the current latest block and submit it to the sequencer in time to get the bundle included in the next subsequent block. 

To increase the time available for bundle simulation, the rollup can simulate on top of a block which has not yet been committed yet i.e a block which has not yet been finalized and is undergoing consensus. 

This PR provides two grpc streams in order to enable that:
1. `getOptimisticBlockStream`: This grpc method returns a stream given a rollup id. The returned stream produces the latest optimistic blocks which are blocks which have finished `process_proposal`. 
2. `getBlockCommitmentStream`: This grpc methods returns a stream which produces the block hash and block height of the latest committed blocks. These are blocks which have finished `finalize_block`.

The Optimistic block grpc streams are flagged behind `ASTRIA_SEQUENCER_NO_OPTIMISTIC_BLOCKS`. Ideally the grpc streams would run on a dedicated full node which is not validating.

## Changes
- Add `ASTRIA_SEQUENCER_NO_OPTIMISTIC_BLOCKS` to the sequencer environment which feature flags the optimistic block grpc streams
- Add an `EventBus` which can be subscribed to, to listen to blocks undergoing consensus in process_proposal and blocks being finalized in finalize_block.
- Add `EventReceivers` and `EventSender` types which are generic types which contains watches over which the specified data can be sent. 
- Add `OptimisticBlockFacade` which is the struct over which the optimistic block service apis are implemented. 
- Add `OptimisticBlockServiceInner` which is a struct which contains a Joinset of the stream tasks. This struct defines a long lived task which in turn spawns stream tasks on the joinset when clients try to subscribe to the optimistic block api streams.
- Move `start_grpc_server` to `crate::grpc::mod.rs` and renamed it to `start_server`

## Testing
Deployed a sequencer network locally and wrote a grpc client to read from the stream. 

## Related Issues
closes https://github.com/astriaorg/astria/issues/1563


crates/astria-sequencer/src/app/mod.rs
```diff
@@ -437,16 +477,26 @@ impl App {
...
+                        process_proposal.time,
+                        process_proposal.proposer_address,
+                        process_proposal.txs,
+                        tx_results,
+                    )
+                    .await
+                    .wrap_err("failed to run post execute transactions handler")?;
+
+                if let Some(optimistic_block_channels) = &self.optimistic_block_channels {
+                    if let Err(e) = optimistic_block_channels
```
bharath-123, line 493: we do not want to fail process_proposal if sending on the watch fails.

crates/astria-sequencer/src/app/mod.rs
```diff
@@ -892,7 +952,7 @@ impl App {
         proposer_address: account::Id,
         txs: Vec<bytes::Bytes>,
         tx_results: Vec<ExecTxResult>,
-    ) -> Result<()> {
+    ) -> Result<SequencerBlock> {
```
bharath-123, line 955: I believe this would be faster since we do not have again read the `SequencerBlock` from state when we want to send it via the optimistic block channels.

crates/astria-sequencer/src/app/mod.rs
```diff
@@ -942,7 +1002,7 @@ impl App {
         )
         .wrap_err("failed to convert block info and data to SequencerBlock")?;
         state_tx
-            .put_sequencer_block(sequencer_block)
+            .put_sequencer_block(sequencer_block.clone())
```
bharath-123, line 1005: this may or may not be a bad idea, since we are cloning a potentially big struct which might make a db read more efficient?

bharath-123:
This PR is blocked on https://github.com/astriaorg/astria/pull/1707

crates/astria-sequencer/src/grpc/optimistic.rs
```diff
@@ -0,0 +1,152 @@
...
+    pub(crate) fn new(optimistic_block_channels: OptimisticBlockChannels) -> Self {
+        Self {
+            optimistic_block_channels,
+        }
+    }
+}
+
+#[async_trait::async_trait]
+impl OptimisticBlockService for OptimisticBlockServer {
+    type GetBlockCommitmentStreamStream =
```
bharath-123, line 47: this is a weird name, I believe they just suffix the rpc method name with `Stream`. I think we can avoid the word stream in our rpc methods?

crates/astria-sequencer/src/grpc/optimistic.rs
```diff
@@ -0,0 +1,152 @@
...
+                        optimistic_block.to_filtered_block(vec![rollup_id]);
+                    let raw_filtered_optimistic_block = filtered_optimistic_block.into_raw();
+
+                    let get_optimistic_block_stream_response = GetOptimisticBlockStreamResponse {
+                        block: Some(raw_filtered_optimistic_block),
+                    };
+
+                    match tx.send(Ok(get_optimistic_block_stream_response)).await {
+                        Ok(()) => {
+                            debug!("sent optimistic block");
```
bharath-123, line 93: should we log this? this will log multiple times with each new listener of the stream.

crates/astria-core/src/sequencerblock/v1/mod.rs
```diff
@@ -1,5 +1,6 @@
 pub mod block;
 pub mod celestia;
+pub mod optimistic_block;
```
bharath-123, line 3: Final proto definitions are being worked on at https://github.com/astriaorg/astria/pull/1707

crates/astria-sequencer/src/app/mod.rs
```diff
@@ -175,6 +180,48 @@ impl From<abci::request::ProcessProposal> for ProposalFingerprint {
...
+            if let Err(e) = self.optimistic_block_sender.send(block) {
+                error!(error = %e, "failed to send optimistic block");
+            }
+        }
+    }
+
+    pub(crate) fn send_committed_block(&self, block: Option<SequencerBlockCommit>) {
+        if self.committed_block_sender().receiver_count() > 0 {
+            if let Err(e) = self.committed_block_sender.send(block) {
+                error!(error = %e, "failed to send committed block");
```
bharath-123, line 219: we should not error out `process_proposal` or `finalize_block` if sending a block on the watch fails

crates/astria-sequencer/src/sequencer.rs
```diff
@@ -159,6 +199,7 @@ fn start_grpc_server(
     storage: &cnidarium::Storage,
     mempool: Mempool,
     grpc_addr: std::net::SocketAddr,
+    optimistic_block_channels: Option<OptimisticBlockChannels>,
     shutdown_rx: oneshot::Receiver<()>,
 ) -> JoinHandle<Result<(), tonic::transport::Error>> {
```
SuperFluffy, line 204: Relevant to some observations I made earlier: if `start_grpc_server` returned a tuple `(JoinHandle, GrpcServerChannel)` (terrible names, just for the sake of the argument), then other parts of the system could use `GrpcServerChannel` to communicate with its components without needing to be aware of what's going on inside.

crates/astria-sequencer/src/grpc/optimistic.rs
```diff
@@ -0,0 +1,151 @@
...
+    },
+    Request,
+    Response,
+    Status,
+};
+use tracing::error;
+
+use crate::app::OptimisticBlockChannels;
+
+pub(crate) struct OptimisticBlockServer {
```
SuperFluffy, line 30: I think this should not be surfaced all the way to the app. Leave this private to the `gRPC` module.

Also:
```suggestion
struct OptimisticBlockServiceImpl {
```

crates/astria-sequencer/src/grpc/optimistic.rs
```diff
@@ -0,0 +1,151 @@
...
+        let (tx, rx) =
+            tokio::sync::mpsc::channel::<Result<GetOptimisticBlockStreamResponse, Status>>(128);
+
+        let mut optimistic_block_receiver = self
+            .optimistic_block_channels
+            .optimistic_block_sender()
+            .subscribe();
+
+        tokio::spawn(async move {
+            loop {
```
SuperFluffy, line 73: We will never see the events emitted in this loop because they are not part of a span, as far as I can see.

crates/astria-sequencer/src/grpc/optimistic.rs
```diff
@@ -0,0 +1,151 @@
...
+
+        let (tx, rx) =
+            tokio::sync::mpsc::channel::<Result<GetOptimisticBlockStreamResponse, Status>>(128);
+
+        let mut optimistic_block_receiver = self
+            .optimistic_block_channels
+            .optimistic_block_sender()
+            .subscribe();
+
+        tokio::spawn(async move {
```
SuperFluffy, line 72: I am not a friend of fire-and-forget tasks. I'd rather add another indrection here by spawning on a joinset/joinmap and then joining the invidual tasks there. That would require chantes to the `OptimisticBlockServerService` impl

crates/astria-sequencer/src/grpc/optimistic.rs
SuperFluffy, line 1: I am not a big fan of spawning tasks onto the runtime and then ignoring what's going on inside them (as is currently done in both the `get_optimistic_block_stream` and `get_block_commitment_stream` handlers).

I'd follow this strategy:

1. Add a new `struct OptimisticBlockServiceInner` that acts as an event loop + task spawner. It will contain an mpsc receiver to know when to spawn new tasks (it will receive a message from the facade described in 2), a joinset (or map), and the optimistic block channels. Provide a `OptimisticBlockServiceInner::run` that you launch as a tokio task and provide a loop { select!() }` over the mpsc receiver and the joinset
2. Change the current `OptimisticBlockServer` to be a facade to the impl in 1, and reimplement both `get_*_stream` handlers to send a request to `OptimisticBlockServiceImpl` that will itself spawn a task on its joinset.
3. In `start_grpc_server`, you can move the join handle to the event loop obtained in 1 into the shutdown function to await it there.

SuperFluffy (changes requested):
From a high level it looks simple enough.

There are some comments for a) productionization and b) long level maintenance that I would like to see addressed:

1. I would prefer if the `App` type has no awareness of whether or not optimistic blocks are set. Leave that to the grpc server.
2. Don't spawn tasks directly on the tokio runtime, especially if they can potentially panic. I have provided a suggestion in the `grpc::optimistic` module as to how that could be done.

crates/astria-sequencer/local.env.example
```diff
@@ -33,6 +33,10 @@ ASTRIA_SEQUENCER_METRICS_HTTP_LISTENER_ADDR="127.0.0.1:9000"
 # `ASTRIA_SEQUENCER_FORCE_STDOUT` is set to `true`.
 ASTRIA_SEQUENCER_PRETTY_PRINT=false
 
+# If true disables streaming optimistic blocks to clients.
```
SuperFluffy, line 36: I think I flagged this before and you closed the issue with a thumbsup:

```
# Disables streaming optimistic blocks to clients
```

bharath-123: my bad. i forgot to make this change. will do it.

crates/astria-sequencer/src/sequencer.rs
```diff
@@ -195,12 +236,31 @@ fn start_grpc_server(
...
+        }
+    };
 
     info!(grpc_addr = grpc_addr.to_string(), "starting grpc server");
-    tokio::task::spawn(
-        grpc_server.serve_with_shutdown(grpc_addr, shutdown_rx.unwrap_or_else(|_| ())),
-    )
+
+    tokio::task::spawn(grpc_server.serve_with_shutdown(grpc_addr, async move {
+        tokio::select! {
```
SuperFluffy, line 254: Make this `select!` `biased` on the shutdown rx.

crates/astria-sequencer/src/sequencer.rs
```diff
@@ -195,12 +236,31 @@ fn start_grpc_server(
...
+            _ = shutdown_rx => {
+                info!("grpc server shutting down");
+            },
+            _ = optimistic_block_inner_handle => {
+                info!("optimistic block inner handle task exited");
+            }
+        }
+        optimistic_streams_cancellation_token.cancel();
```
SuperFluffy, line 262: After this, wait a moment so that the service has time to shutdown (only necessary if you didn't get here because of the task optimistic stream task having shut down).

crates/astria-sequencer/src/sequencer.rs
```diff
@@ -195,12 +236,31 @@ fn start_grpc_server(
...
 
     info!(grpc_addr = grpc_addr.to_string(), "starting grpc server");
-    tokio::task::spawn(
-        grpc_server.serve_with_shutdown(grpc_addr, shutdown_rx.unwrap_or_else(|_| ())),
-    )
+
+    tokio::task::spawn(grpc_server.serve_with_shutdown(grpc_addr, async move {
+        tokio::select! {
+            _ = shutdown_rx => {
+                info!("grpc server shutting down");
```
SuperFluffy, line 256: This`info!` event would be running outside of a span. Create a span first before emitting these (applies to both the branch and the other one below).

crates/astria-sequencer/src/sequencer.rs
```diff
@@ -195,12 +236,31 @@ fn start_grpc_server(
...
+            },
+            _ = optimistic_block_inner_handle => {
+                info!("optimistic block inner handle task exited");
```
SuperFluffy, line 259: This would probably be pretty unexpected (it should never just go down). Make this into a `warn!`).

crates/astria-sequencer/src/app/mod.rs
```diff
@@ -433,16 +452,19 @@ impl App {
...
-                    process_proposal.hash,
-                    process_proposal.height,
-                    process_proposal.time,
-                    process_proposal.proposer_address,
-                    process_proposal.txs,
-                    tx_results,
-                )
-                .await
-                .wrap_err("failed to run post execute transactions handler")?;
+                let sequencer_block = self
```
SuperFluffy, line 455: Unhappy about the code duplication here and at the bottom.

Please create a follow up issue so that we can shift this code around and not have all of this code duplication.

SuperFluffy (commented):
Going in the right direction. I mentioned a couple of nice to haves and a few gotchas that we should fix before merge.

crates/astria-sequencer/src/app/mod.rs
```diff
@@ -232,6 +237,13 @@ pub(crate) struct App {
     )]
     app_hash: AppHash,
 
+    // This contains the channels to send the optimistically executed block
```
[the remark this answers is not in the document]
bharath-123: you are right, thats my bad

crates/astria-sequencer/src/app/event_bus.rs
```diff
@@ -0,0 +1,101 @@
...
+            if let Err(e) = self.sender.send(Some(event)) {
+                error!(error = %e, "failed to send event");
+            }
+            self.is_init.cancel();
+        }
+    }
+}
+
+#[derive(Clone)]
+pub(crate) struct EventBus {
```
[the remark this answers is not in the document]
bharath-123: and we can make EventSender local to the module

crates/astria-sequencer/src/grpc/mod.rs
```diff
@@ -1,8 +1,138 @@
...
+
+    info!(grpc_addr = grpc_addr.to_string(), "starting grpc server");
+
+    tokio::task::spawn(grpc_server.serve_with_shutdown(grpc_addr, async move {
+        let reason = tokio::select! {
+            biased;
+            _ = shutdown_rx => {
+                Ok("grpc server shutting down")
+            },
+            _ = optimistic_block_inner_handle => {
```
SuperFluffy, line 118: Related to my comment above:

```suggestion
            res = async { optimistic_block_inner_handle.as_mut().unwrap().await }, if optimistic_block_inner_handle.is_some() => {
```

Also use the return value. The reason why we are doing all this is to know if it crashed for whatever reason (otherwise we could have just spawned on the runtime).

crates/astria-sequencer/src/grpc/mod.rs
```diff
@@ -1,8 +1,138 @@
...
+        // (from Penumbra) Add permissive CORS headers, so pd's gRPC services are accessible
+        // from arbitrary web contexts, including from localhost.
+        .layer(cors_layer)
+        .add_service(ClientQueryServer::new(ibc.clone()))
+        .add_service(ChannelQueryServer::new(ibc.clone()))
+        .add_service(ConnectionQueryServer::new(ibc.clone()))
+        .add_service(SequencerServiceServer::new(sequencer_api))
+        .add_optional_service(optimistic_block_service_server);
+
+    let optimistic_block_inner_handle = {
```
SuperFluffy, line 102: Let's just call this `optimistic_block_task` or something easier to digest.

crates/astria-core/src/sequencerblock/v1/optimistic.rs
SuperFluffy, line 1: Note that we need to bring this in line with the auctioneer PR before we merge.

SuperFluffy (commented):
Bunch of small things. Nothing groundbreaking, but would be nice to clean up before we merge.

crates/astria-sequencer/src/app/event_bus.rs
```diff
@@ -0,0 +1,160 @@
...
+    }
+}
+
+/// `EventSender` contains the sender side of the events sent by the Sequencer App.
+/// At any given time, it sends the latest value of the event.
+struct EventSender<T> {
+    // The sender side of the watch which is used to send the latest value of the event.
+    // We use an Option here to allow for the sender to be initialized with a None value
+    // which allows the type to not have a Default value.
+    sender: Sender<Option<T>>,
```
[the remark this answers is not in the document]
bharath-123: i ll call it inner

SuperFluffy:
Manually merged this into https://github.com/astriaorg/astria/pull/1839, so this can be closed.

PR #1714, opened 2024-10-22
fix(bridge-withdrawer)!: correctly identify rollup return address in ics20 withdrawal actions

itamarreif:
## Summary
Use the full hex encoded sender of the smart contract event as the rollup return address in the ics20 withdrawal event memo.

## Background
Before this patch, the `astria.protocol.memos::v1::Ics20WithdrawalFromRollup` memo used the abbreviated form of the ethers contract event sender, making it impossible to return funds if necessary. Specifically, the field was set as `rollupReturnAddress":"0xc59a…2d94"`

After this patch, the full address is used, making it possible to identify the sender.

## Changes
- change the event conversion logic to use `AbiEncode::encode_hex()` instead of `H160::to_string()`
- update the black box test mock to also do this

## Breaking Changelist
- This is a breaking change because two differerent versions of bridge-withdrawer will produce different sequencer transactions.

## Related Issues
Original issue where this was first encountered but not fixed: #1427

### DIFF
--- a/crates/astria-bridge-contracts/src/lib.rs
+++ b/crates/astria-bridge-contracts/src/lib.rs
@@ -422,7 +422,7 @@ where
         let memo = memo_to_json(&memos::v1::Ics20WithdrawalFromRollup {
             memo: event.memo.clone(),
             rollup_block_number,
-            rollup_return_address: event.sender.to_string(),
+            rollup_return_address: event.sender.encode_hex(),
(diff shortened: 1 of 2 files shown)


joroshiba (commented):
Do we have any tests over this? I would ideally have a snapshot test that validates that json memos don't change

joroshiba (approved):
Approved based on my other passing test built on top of this.

PR #1718, opened 2024-10-22
fix(sequencer): move fee event recording to tx level

ethanoroshiba:
## Summary
Moves ABCI fee events to `FinalizeBlock.tx_results` to identify fees on a per transaction basis.

## Background
Post-fee rework, fee ABCI event construction was happening outside of transaction execution and (wrongly) included in the `FinalizeBlock.events` field of the finalize block response. This made it difficult to identify which fee events were related to which transaction.

This patch moves fee event construction back into the (atomic) transaction execution, so that the generated ABCI fee events per transaction are again part of `FinalizeBlock.tx_results`.

For example for Astrotek fees will again show up next to each transaction.

## Changes
- Moved fee event construction back into transaction execution.

## Testing
Adjusted fee event unit test to ensure fee event is returned as part of transaction execution.


joroshiba (approved):

SuperFluffy (changes requested):
I don't understand the patch notes: it notes that event construction is moved from "End Block" (a cometbft concept) to `add_fees_to_block_fees`. Can you explain what this means in terms of the execution flow?

If the event construction no longer happens as part of End Block, where does it happen now?

The patch notes also notes that this is a fix but it is not clear what is being fixed. Why is it a problem that the event is associated with the block and not the transaction? Weren't the `sourceTransactionId` and `sourceActionIndex` fields in the event intended to address this?

ethanoroshiba:
@SuperFluffy I attempted to update the patch notes to reflect the methodology behind the changes more. Let me know if this works for you. As for why I called it "fix", this is because it was pointed out as an issue that a transaction's fees were no longer being recorded with the transaction itself. That was the reason we added the source action and transaction id to the fee events though, so it may be a UX decision. I'll bring it up offline.

crates/astria-sequencer/src/fees/state_ext.rs
```diff
@@ -532,6 +541,20 @@ pub(crate) trait StateWriteExt: StateWrite {
...
 
+/// Creates `abci::Event` of kind `tx.fees` for sequencer fee reporting
+fn construct_tx_fee_event(fee: &Fee) -> Event {
+    Event::new(
+        "tx.fees",
+        [
+            ("actionName", fee.action_name.to_string()).index(),
+            ("asset", fee.asset.to_string()).index(),
+            ("feeAmount", fee.amount.to_string()).index(),
+            ("sourceTransactionId", fee.source_transaction_id.to_string()).index(),
```
joroshiba, line 552: This shouldn't be needed when correctly attached to a given tx?

crates/astria-sequencer/src/fees/state_ext.rs
```diff
@@ -532,6 +541,20 @@ pub(crate) trait StateWriteExt: StateWrite {
 
 impl<T: StateWrite> StateWriteExt for T {}
 
+/// Creates `abci::Event` of kind `tx.fees` for sequencer fee reporting
+fn construct_tx_fee_event(fee: &Fee) -> Event {
+    Event::new(
+        "tx.fees",
+        [
+            ("actionName", fee.action_name.to_string()).index(),
```
joroshiba, line 549: We could also probably omit the action name, since it can be assumed learned via the transaction index?

joroshiba:
> The patch notes also notes that this is a fix but it is not clear what is being fixed. Why is it a problem that the event is associated with the block and not the transaction?

@SuperFluffy Changing where the events are generated changes the external API for fees. You can no longer discover the fee paid by querying a transaction. You must now query the block and search for your tx hash in the results. Generally speaking this makes the block query return oversized payloads as now every query for a block returns the fees paid for every action on it. Fundamentally this a TX result not a block result.

If we wanted a block result it should be the total fees paid over the whole block.

noot (approved):

crates/astria-sequencer/src/fees/state_ext.rs
ethanoroshiba: removed in a72a7c9

crates/astria-sequencer/src/fees/state_ext.rs
ethanoroshiba: Might be a nicety to have though? Idk, just a thought

SuperFluffy: I think it's nice to have it. Makes it much easier to identify the right action.

crates/astria-sequencer/src/fees/state_ext.rs
```diff
@@ -565,6 +569,19 @@ pub(crate) trait StateWriteExt: StateWrite {
...
+            ("asset", fee.asset.to_string()).index(),
+            ("feeAmount", fee.amount.to_string()).index(),
+            ("sourceActionIndex", fee.source_action_index.to_string()).index(),
```
SuperFluffy, line 580: I actually think we should change this name to `"positionInTransaction"`. I was part of the original review and had to do a double take what `sourceActionIndex` means.

crates/astria-sequencer/src/fees/mod.rs
```diff
@@ -379,17 +370,3 @@ fn base_deposit_fee(asset: &asset::Denom, destination_chain_address: &str) -> u1
...
-
-/// Creates `abci::Event` of kind `tx.fees` for sequencer fee reporting
-pub(crate) fn construct_tx_fee_event(fee: &Fee) -> Event {
-    Event::new(
-        "tx.fees",
-        [
-            ("actionName", fee.action_name.to_string()).index(),
-            ("asset", fee.asset.to_string()).index(),
-            ("feeAmount", fee.amount.to_string()).index(),
-            ("sourceTransactionId", fee.source_transaction_id.to_string()).index(),
```
SuperFluffy, line 391: The argument for removing the transaction ID is that it will appear in the tx results so that the tx will be identified that way?

SuperFluffy: Our transaction ID is the sha256 of the transaction on-wire bytes. This is the same as

SuperFluffy (approved):

crates/astria-sequencer/src/fees/state_ext.rs
ethanoroshiba: Followup issue here: https://github.com/astriaorg/astria/issues/1735

PR #1726, opened 2024-10-23
chore(sequencer): don't generate tests using macros

### LINKED ISSUE #1715: sequencer: make fee change testing easier to debug
As mentioned [here](https://github.com/astriaorg/astria/pull/1701#discussion_r1810756793), this macro will be difficult to debug, so we should replace it.

┆Issue Number: ENG-944

ethanoroshiba:
## Summary
Replaced macros generating fee action unit tests by expanding them.

## Background
Macro-generated tests would be a headache to debug, since failures may point to lines of code which don't exist. While there is significantly more code, I think it's preferable to aide in future debugging.

## Changes
- Removed macro which generated `FeeChange` unit tests.
- Created helper functions for shortening fee change tests as much as possible.
- Created fee change tests for every possible variant of `FeeChange` (i.e. one for each action).

## Testing
Passing all tests.

## Related Issues
closes #1715


Fraser999 (changes requested):
I'm afraid I see this PR as a step in the wrong direction.

While I agree that using macros can lead to somewhat hard-to-debug cases, I think the benefits here far outweigh that minor drawback.  Debugging wouldn't be particularly hard (the generated tests are relatively simple), whereas getting all the copy-pasted code correct is much harder, for authors, maintainers and reviewers.

I haven't given the changes a proper review, but skimming has revealed a few such errors, lending weight to the case for using a macro here.  In the PR which introduced the macro, I made several such mistakes in the highly repetitive, non-macro fee code, and the macro tests caught these.  The macro tests were in fact introduced because I noticed that the existing unit test for fee change was only covering a few of the fee variants.

My strong preference would be to close this PR unmerged.  If we don't reach agreement on that, then I'll do a full review at that point.

PR #1730, opened 2024-10-23
chore(sequencer): init allowed assets in fees component

SuperFluffy:
## Summary
Moves the initialization of allowed assets from the overall app to the fees component.

## Background
https://github.com/astriaorg/astria/pull/1647 moved fee asset state handling to the new `fees` component. It was overlooked that writing allowed fee assets should also happen in its init chain logic.

## Changes
- Move initialization of app init state to the fees component.

## Testing
Nothing has changed in terms of how the state is represented. Tests that rely on init chain still work.

## Breaking Changelist
This is not a breaking change because the state after this change looks the same.

### DIFF
--- a/crates/astria-sequencer/src/accounts/component.rs
+++ b/crates/astria-sequencer/src/accounts/component.rs
@@ -2,6 +2,7 @@ use std::sync::Arc;
 
 use astria_core::protocol::genesis::v1::GenesisAppState;
 use astria_eyre::eyre::{
+    OptionExt as _,
     Result,
     WrapErr as _,
 };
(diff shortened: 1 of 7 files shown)


crates/astria-sequencer/src/accounts/component.rs
```diff
@@ -29,14 +29,16 @@ impl Component for AccountsComponent {
...
     {
-        let native_asset = state
-            .get_native_asset()
-            .await
-            .wrap_err("failed to read native asset from state")?;
-        for account in app_state.accounts() {
-            state
-                .put_account_balance(&account.address, &native_asset, account.balance)
-                .wrap_err("failed writing account balance to state")?;
+        if !app_state.accounts().is_empty() {
```
Fraser999, line 32: This isn't really needed is it?  I'd be inclined to remove it for clarity, but not a big deal.

Fraser999 (approved):

crates/astria-sequencer/src/accounts/component.rs
SuperFluffy: I need the conditional because I only need the native asset (fetched on the line below) if the accounts are set. If there are no accounts in genesis (and no native asset), then I don't want to potentially fail init chain on that call.

Is that what you meant or am I overlooking something?

Fraser999: No - you're correct.  We do need the `if` block.

Fraser999 (approved):

PR #1739, opened 2024-10-24
chore(sequencer): refactor fees

### LINKED ISSUE #1715: sequencer: make fee change testing easier to debug

ethanoroshiba:
## Summary
Added trait for fee components to allow generic usage of them.

## Background
Previously, there was a lot of boilerplate required when testing fee components, since we are usually testing them for each and every action. Additionally, this introduces a lot of room for human error, since it's easy to mismatch an action and its associated methods. This change aims to rely more heavily on trait methods and helper functions, so that generic types can be used to eliminate boilerplate and room for error.

## Changes
- Added new `FeeComponents` trait and implemented it for all fee components.
- Added new type and method to `FeeHandler` trait.
- Changed all actions in the fees request (`fees::query::get_fees_for_transaction`) to use helper functions with generic inputs.
- Refactored fee change tests to use helper function and macro, as opposed to generated tests. The aim of this is to make it more difficult to make manual mistakes in the future, while still making debugging of these tests easier.
- Refactored round trip state read/write tests.

## Testing
Passing all tests

## Related Issues
closes #1715


crates/astria-sequencer/src/fees/action.rs
```diff
@@ -137,98 +137,290 @@ impl ActionHandler for FeeAssetChange {
...
+            address_bytes: [1; 20],
+            transaction_id: TransactionId::new([0; 32]),
+            source_action_index: 0,
+        });
+        state.put_sudo_address([1; 20]).unwrap();
+
+        assert!(
+            F::fee_components(&state)
+                .await
+                .unwrap_or_else(|_| panic!("should not error fetching unstored {action_name} fees"))
```
noot, line 400:
```suggestion
                .expect("should not error fetching unstored {action_name} fees")
```

same for other `unwrap_or_else` with `panic!`

crates/astria-sequencer/src/fees/action.rs
```diff
@@ -137,98 +137,290 @@ impl ActionHandler for FeeAssetChange {
...
+        )
+        .await;
+    }
+
+    async fn test_fee_change_action<F: FeeHandler<FeeComponents = T>, T>(
+        initial_fees: T,
+        initial_fee_change: FeeChange,
+        new_fees: T,
+        new_fee_change: FeeChange,
+        action_name: &str,
```
noot, line 381: not sure using `action_name` is super useful here, as if the test fails, it'll print the name of the test which has the action name

crates/astria-sequencer/src/fees/query.rs
```diff
@@ -240,184 +257,155 @@ pub(crate) async fn get_fees_for_transaction<S: StateRead>(
...
             }
             Action::ValidatorUpdate(_) => {
-                validator_update_fees
-                    .get_or_try_init(|| async { state.get_validator_update_fees().await })
-                    .await
-                    .wrap_err("failed to get validator update fees")?
-                    .ok_or_eyre(
-                        "fees not found for `ValidatorUpdate` action, hence it is disabled",
-                    )?;
+                get_or_init_fees::<ValidatorUpdate, _, _>(
```
noot, line 351: it doesn't look like there's a point to calling `get_or_init_fees` in this action or any of the ones below that are feeless, as they don't update `fees_by_asset`, which is the point of this function

crates/astria-sequencer/src/fees/action.rs
ethanoroshiba: Done in 3a71450

crates/astria-sequencer/src/fees/action.rs
ethanoroshiba: Removed in 3a71450

crates/astria-sequencer/src/fees/query.rs
ethanoroshiba: This was done so that the check would fail if the transaction would fail during execution (due to a given action not having fees). I think it may be worth it to keep this logic since it is used in mempool as well as `check_and_execute` and may catch a failure-bound transaction before it gets executed.

ethanoroshiba:
Closing in favor of https://github.com/astriaorg/astria/pull/1794

PR #1755, opened 2024-10-25
chore(cli)!: default to dusk-11

joroshiba:
## Summary
Updates the CLI to work with a production network by default

## Background
The CLI has continued to default to dusk-10 and had an old env var name that didn't make sense for chain id without rollup code inside.

## Changes
- Default chain and RPC for dusk 11

## Breaking Changelist
- Changed the env var for sequencer chain id spec on many commands from `ROLLUP_SEQUENCER_CHAIN_ID` to `SEQUENCER_CHAIN_ID`

## Testing
N/A

## Changelogs
Changelogs updated.

### DIFF
--- a/crates/astria-cli/CHANGELOG.md
+++ b/crates/astria-cli/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Changes
+
 - Default to dusk-11 network [#1755](https://github.com/astriaorg/astria/pull/1755)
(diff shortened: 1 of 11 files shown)


Fraser999 (approved):

SuperFluffy (commented):
This PR clashes with https://github.com/astriaorg/astria/pull/1792, which seeks to remove the defaults altogether.

If we choose to distribute a binary that talks to mainnet, then that should be done via a config file that we ship along the binary, but not as a default.

SuperFluffy:
Closing in favor of https://github.com/astriaorg/astria/pull/1792

PR #1761, opened 2024-10-28
chore(sequencer): add instrumentation

### LINKED ISSUE #1321: chore: add instrumentation on all async calls

ethanoroshiba:
## Summary
Added instrumentation to most async functions which did not have it.

## Background

## Changes
- Added instrumentation to most async functions which did not have it.
- Added `err` argument to the `instrument` macro where applicable.
- Added fields to the `instrument` macro where applicable.

## Testing
Passing all tests.

## Changelogs
No updates required

## Related Issues
closes #1321


Fraser999 (changes requested):
Blocking this for now (as much a reminder to myself as anything else).

Discussed offline with @ethanoroshiba that this will be revisited to ensure we're only using error-level instrumentation where it's warranted (i.e. where it indicates a serious problem or bug in the logic) and not for non-critical cases (e.g. handling a malformed client request).

crates/astria-sequencer/src/mempool/transactions_container.rs
```diff
@@ -569,6 +573,7 @@ pub(super) trait TransactionsContainer<T: TransactionsForAccount> {
     /// Recosts transactions for an account.
     ///
     /// Logs an error if fails to recost a transaction.
+    #[instrument(skip_all, fields(address = %hex::encode(address)))]
```
Fraser999, line 576: I think the address should be base64-encoded.

crates/astria-sequencer/src/mempool/mod.rs
```diff
@@ -482,7 +485,7 @@ impl Mempool {
     /// does not take into account gapped nonces in the parked queue. For example, if the
     /// pending queue for an account has nonces [0,1] and the parked queue has [3], [1] will be
     /// returned.
-    #[instrument(skip_all)]
+    #[instrument(skip_all, fields(address = %hex::encode(address)))]
```
Fraser999, line 488: Address should be base64-encoded I think.

crates/astria-sequencer/src/sequencer.rs
```diff
@@ -57,15 +58,19 @@ impl Sequencer {
             .try_exists()
             .context("failed checking for existence of db storage file")?
         {
-            info!(
-                path = %config.db_filepath.display(),
-                "opening storage db"
-            );
+            info_span!("Sequencer::run_until_stopped").in_scope(|| {
```
Fraser999, line 61: We could just hold the span in a variable and re-use it throughout this method:
```rust
        let span = info_span!("Sequencer::run_until_stopped");
        let db_exists = config
            .db_filepath
            .try_exists()
            .wrap_err("failed checking for existence of db storage file")?;
        span.in_scope(|| {
            info!(path = %config.db_filepath.display(), "{} storage db", if db_exists {
                "opening"
            } else {
                "creating"
            });
        });
```

crates/astria-sequencer/src/sequencer.rs
```diff
@@ -119,27 +124,29 @@ impl Sequencer {
             .wrap_err("failed to parse grpc_addr address")?;
         let grpc_server_handle = start_grpc_server(&storage, mempool, grpc_addr, shutdown_rx);
 
-        info!(config.listen_addr, "starting sequencer");
+        info_span!("Sequencer::run_until_stopped")
+            .in_scope(|| info!(config.listen_addr, "starting sequencer"));
```
Fraser999, line 128:
```suggestion
        span.in_scope(|| info!(config.listen_addr, "starting sequencer"));
```

crates/astria-sequencer/src/sequencer.rs
```diff
@@ -119,27 +124,29 @@ impl Sequencer {
...
+                        .in_scope(|| error!(err = e.as_ref(), "ABCI server exited with error"));
                 }
             }
             let _ = server_exit_tx.send(());
         });
 
         select! {
             _ = signals.stop_rx.changed() => {
-                info!("shutting down sequencer");
+                info_span!("Sequencer::run_until_stopped").in_scope(|| info!("shutting down sequencer"));
```
Fraser999, line 145:
```suggestion
                span.in_scope(|| info!("shutting down sequencer"));
```

crates/astria-sequencer/src/sequencer.rs
```diff
@@ -119,27 +124,29 @@ impl Sequencer {
...
             }
 
             _ = server_exit_rx => {
-                error!("ABCI server task exited, this shouldn't happen");
+                error_span!("Sequencer::run_until_stopped").in_scope(|| error!("ABCI server task exited, this shouldn't happen"));
```
Fraser999, line 149:
```suggestion
                span.in_scope(|| error!("ABCI server task exited, this shouldn't happen"));
```

ethanoroshiba: Made requested changes in 66e1b84 :)

Fraser999 (approved):

PR #1762, opened 2024-10-29
fix(bridge-withdrawer, cli)!: use decimals from erc20 contract

joroshiba:
## Summary
Updates CLI and bridge withdrawer to read the decimals off the contract when it is an erc20, to get the accurate decimals to utilize during withdrawal.

## Background
Previously it was assumed that all erc20 contracts would have 18 decimals this is not guaranteed. #1760 is a similar fix but relies on configuring the withdrawer correctly. This version automatically reads off of the contract instead if it is an erc20.

## Changes
- Read decimals off contract if  or fallback to native prevision.

## Testing
Ran against Flame to ensure it gets decimals correct on 6 decimal contract + CI tests.

## Changelogs
Changelogs updated.

### DIFF
--- a/charts/evm-bridge-withdrawer/values.yaml
+++ b/charts/evm-bridge-withdrawer/values.yaml
@@ -27,7 +27,6 @@ config:
   minExpectedFeeAssetBalance: "1000000"
   rollupAssetDenom: ""
   evmContractAddress: "0x"
-  evmContractIsERC20: "true"
   evmRpcEndpoint: ""
   sequencerPrivateKey:
     devContent: ""
--- a/charts/evm-bridge-withdrawer/values.yaml
+++ b/charts/evm-bridge-withdrawer/values.yaml
@@ -27,6 +27,7 @@ config:
   minExpectedFeeAssetBalance: "1000000"
   rollupAssetDenom: ""
   evmContractAddress: "0x"
+  evmContractIsERC20: "true"
   evmRpcEndpoint: ""
   sequencerPrivateKey:
     devContent: ""
--- a/charts/evm-stack/Chart.lock
+++ b/charts/evm-stack/Chart.lock
@@ -13,12 +13,12 @@ dependencies:
   version: 0.1.2
 - name: evm-bridge-withdrawer
   repository: file://../evm-bridge-withdrawer
-  version: 1.0.0
+  version: 1.0.1
 - name: postgresql
   repository: https://charts.bitnami.com/bitnami
   version: 15.2.4
 - name: blockscout-stack
   repository: https://blockscout.github.io/helm-charts
   version: 1.6.2
-digest: sha256:c849edbcf6b12442d9ec8ffe2f21ceb4f50e58a65c2186340cae6df034f34eae
-generated: "2024-10-27T10:49:54.580428-07:00"
+digest: sha256:b0f9adad7d8a55f86ebd7b2e7aeb4f5b479b5b38a6c423ff013b9a5a57446f4d
+generated: "2024-10-28T18:10:18.442906-07:00"
--- a/charts/evm-stack/Chart.lock
+++ b/charts/evm-stack/Chart.lock
@@ -21,4 +21,4 @@ dependencies:
   repository: https://blockscout.github.io/helm-charts
   version: 1.6.2
 digest: sha256:c849edbcf6b12442d9ec8ffe2f21ceb4f50e58a65c2186340cae6df034f34eae
-generated: "2024-10-29T17:33:57.517611+01:00"
+generated: "2024-10-27T10:49:54.580428-07:00"
--- a/charts/evm-stack/Chart.lock
+++ b/charts/evm-stack/Chart.lock
@@ -13,12 +13,12 @@ dependencies:
   version: 0.1.2
 - name: evm-bridge-withdrawer
   repository: file://../evm-bridge-withdrawer
-  version: 1.0.1
+  version: 1.0.0
 - name: postgresql
   repository: https://charts.bitnami.com/bitnami
   version: 15.2.4
 - name: blockscout-stack
   repository: https://blockscout.github.io/helm-charts
   version: 1.6.2
-digest: sha256:b0f9adad7d8a55f86ebd7b2e7aeb4f5b479b5b38a6c423ff013b9a5a57446f4d
-generated: "2024-10-28T18:10:18.442906-07:00"
+digest: sha256:c849edbcf6b12442d9ec8ffe2f21ceb4f50e58a65c2186340cae6df034f34eae
+generated: "2024-10-29T17:33:57.517611+01:00"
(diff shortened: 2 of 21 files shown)


charts/evm-bridge-withdrawer/values.yaml
```diff
@@ -27,6 +27,7 @@ config:
   minExpectedFeeAssetBalance: "1000000"
   rollupAssetDenom: ""
   evmContractAddress: "0x"
+  evmContractIsERC20: "true"
```
joroshiba, line 30: I set the default to true here because it's a little safer. If it is not an ERC20, it will fail on startup. Otherwise it could startup and not read the decimals resulting in using the wrong decimals.

crates/astria-bridge-contracts/CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+- Add `contract_is_erc20` setting for watcher [#1762](https://github.com/astriaorg/astria/pull/1762)
```
Fraser999, line 12:
```suggestion
### Added

- Add `contract_is_erc20` setting for watcher [#1762](https://github.com/astriaorg/astria/pull/1762).
```

crates/astria-bridge-withdrawer/CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Add `ethereum_contract_is_erc20` configuration value [#1762](https://github.com/astriaorg/astria/pull/1762)
```
Fraser999, line 12:
```suggestion
### Added

- Add `ethereum_contract_is_erc20` configuration value [#1762](https://github.com/astriaorg/astria/pull/1762).
```

crates/astria-cli/CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Add `contract_is_erc20` flag on bridge collect command [#1762](https://github.com/astriaorg/astria/pull/1762)
```
Fraser999, line 12:
```suggestion
### Added

- Add `contract_is_erc20` flag on `bridge collect-withdrawals` command [#1762](https://github.com/astriaorg/astria/pull/1762).
```

crates/astria-cli/src/bridge/collect.rs
```diff
@@ -50,6 +50,9 @@ pub(super) struct Command {
     /// The eth address of the astria bridge contracts.
     #[arg(long)]
     contract_address: ethers::types::Address,
+    /// Whether the contract is an ERC20 contract.
```
Fraser999, line 53: I think the doc comment should target the user of the CLI since it's used to create the help string.  So, I'd lean more towards just "The contract is an ERC20 contract" since that's what enabling the flag represents.

crates/astria-cli/src/bridge/collect.rs
```diff
@@ -50,6 +50,9 @@ pub(super) struct Command {
...
+    #[arg(long)]
+    contract_is_erc20: bool,
```
Fraser999, line 55: Could we get away with just `erc20` rather than `contract_is_erc20` to be succinct?  I don't think the user would be any more confused if we omit the prefix, and while `contract_is_erc20` seems more suitable as a Rust variable, I think the command-line flag would be fine as `--erc20`.

Not a big deal though.

(If this does get changed, we'll need to update the CHANGELOG entry to match.)

crates/astria-bridge-contracts/CHANGELOG.md
SuperFluffy: Removed that setting. I guess the entry in `CHANGES` is now more appropriate (because I only change behavior but not the public API of the crate)?

crates/astria-bridge-withdrawer/CHANGELOG.md
SuperFluffy: The changes here were rolled back. The only changes now happen in its dependency, `astria-bridge-contracts`. Should this CHANGELOG still get an entry?

crates/astria-cli/CHANGELOG.md
SuperFluffy: The changes here were rolled back. The only changes now happen in its dependency, `astria-bridge-contracts`. Should this CHANGELOG still get an entry?

crates/astria-cli/src/bridge/collect.rs
SuperFluffy: Removed the setting (now the default behavior of the dependency with a fallback if the contract is missing that call).

crates/astria-bridge-contracts/src/lib.rs
```diff
@@ -295,7 +321,21 @@ where
...
+                contract_address,
+                provider.clone(),
+            );
+            contract_decimals = erc_20_contract
+                .decimals()
+                .call()
+                .await
+                .map_err(BuildError::no_decimals_configured_erc20)?
+                .into();
+        }
```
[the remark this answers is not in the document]
SuperFluffy: I pushed a few changes to the PR to make this the default behavior (call `decimals()`, fallback to 18 if missing).

charts/evm-stack/Chart.lock
```diff
@@ -21,4 +21,4 @@ dependencies:
   repository: https://blockscout.github.io/helm-charts
   version: 1.6.2
 digest: sha256:c849edbcf6b12442d9ec8ffe2f21ceb4f50e58a65c2186340cae6df034f34eae
-generated: "2024-10-27T10:49:54.580428-07:00"
+generated: "2024-10-29T17:33:57.517611+01:00"
```
SuperFluffy, line 24: After the reviews and before merging to main I am going to squash all commits and drop this hunk from the patch.

I needed to regenerate this lock file because I rolled back the patch versions of all the charts that were changed in a previous version of this PR. This regen also updated the timestamp.

crates/astria-cli/CHANGELOG.md
Fraser999: Probably not if it's a small opaque change to the CLI.

crates/astria-bridge-withdrawer/CHANGELOG.md
Fraser999: Probably not if it's a small opaque change to the withdrawer.

crates/astria-bridge-contracts/CHANGELOG.md
Fraser999: Yes, that makes sense to me.

charts/evm-stack/Chart.lock
SuperFluffy: I `git checkout <from-fork> -- charts/evm-stack/Chart.lock` so that the linter does not trigger on this.

noot (approved):

PR #1763, opened 2024-10-29
feat(sequencer)!: make parked mempool nonces replaceable

Lilyjjo:
## Summary
This PR enables nonce replacement for all transactions inside of the mempool's parked queue. 

It was discovered while writing this PR that nonce replacement for the bottom nonce in parked was already enabled if the new nonce was able to enter the pending queue, when this happened the replaced nonce was being cleared out during the mempool's maintenance loop. This PR enables all of parked to have nonce replacement as well as adds the new notion of 'removed due to nonce replacement' to the information propagated up to cometBFT. 

## Background
People have started to use the internal rust CLI and have had trouble with transactions getting stuck inside of the mempool's parked queue due to incorrect parameters. This PR enables nonce replacement of parked transactions for a better UX. This feature should be disabled if people start trying to spam the mempool with it.

## Changes
- Enabled nonce replacement in the mempool's parked queue. 
- Removes replaced nonces from the appside and cometBFT mempool.

## Testing
Ran locally and wrote unit tests.

## Breaking Changes
Added new ABCI error code `NONCE_REPLACEMENT`. This is to signal through cometBFT when a transaction has been removed from the mempool due to being replaced.

## Metrics
Added counter `MEMPOOL_NONCE_REPLACEMENT` to track how many times users have used nonce replacement.

## Changelogs
Changelogs updated.


crates/astria-core/src/protocol/abci.rs
```diff
@@ -64,6 +65,9 @@ impl AbciErrorCode {
                 "the account has reached the maximum number of parked transactions".into()
             }
             Self::PARKED_FULL => "the mempool is out of space for more parked transactions".into(),
+            Self::NONCE_REPLACEMENT => {
+                "the transaction was replaced by a different transaction with the same nonce".into()
```
SuperFluffy, line 69: What will the flow look like to actually see this abci error code?

If I understand the follow correctly:

1. transaction is stuck in mempool
2. new transaction is submitted to the mempool, replacing the old one
3. check-tx for the new transaction should return a success code (that is abci code 0).

But `NONCE_REPLACEMENT` is an error code for a transaction that got replaced if I understand correctly.

crates/astria-sequencer/src/mempool/transactions_container.rs
```diff
@@ -434,7 +447,7 @@ pub(super) trait TransactionsForAccount: Default {
         ttx: TimemarkedTransaction,
         current_account_nonce: u32,
         current_account_balances: &HashMap<IbcPrefixed, u128>,
-    ) -> Result<(), InsertionError> {
+    ) -> Result<Option<[u8; 32]>, InsertionError> {
```
SuperFluffy, line 450: This has a bad code smell. The behavior of the trait method can no longer be understood from by just looking at the signature, but I also need to understand the implementation of `nonce_replacement_enabled` to understand its behavior.

Now that the behavior of the two queues is so fundamentally different, they should be represented by two different types.

SuperFluffy:
For new ABCI error codes we should be able to observe the error from the outside. We should add a smoke test that performs the nonce replacement and observes the abci code.

crates/astria-core/src/protocol/abci.rs
Lilyjjo: The next task I have is for general transaction observability (#1773), we don't currently have the infra to communicate it now to the users but we should

SuperFluffy: I don't think one this is strictly related to the other. The `AbciErrorCode`s are intended to be returned in response to abci queries and put into the `code` field of the various responses (in this case `CheckTx`).

All the ABCI error codes can be observed right now (even if clunky). I don't understand when/how this code would be observed.

joroshiba:
closed in 7 days.

joroshiba:

PR #1770, opened 2024-10-30
chore(core, sequencer): move stateless checks to domain type parsing

### LINKED ISSUE #1430: refactor stateless checks to be done on types with withdrawals
PR #1391 introduces some stateless checks in `BridgeUnlockAction` and `Ics20WithdrawalAction` which are very similar. This is done because of parsing done in stateless check of memo for `Ics20WithdrawalAction`. 

As a follow up we should update the canonical domain rust types to doing this validation during parsing, ideally there is some shared logic so it can be kept insync between the two actions.

┆Issue Number: ENG-766

ethanoroshiba:
## Summary
Moved stateless checks for `Ics20Withdrawal` and `BridgeUnlock` to occur when the types are parsed from protobuf.

## Background
Made more sense to check these fields sooner in the process rather than later, and be able to consolidate some of the checks that are shared between `BridgeUnlock` and `Ics20Withdrawal`. 

## Changes
- Moved stateless checks for `Ics20Withdrawal` and `BridgeUnlock` to profobuf `try_from_raw` and `try_from_raw_ref` impls.
- Changed `Ics20Withdrawal` to be an enum whose variants represent a user withdrawal and a withdrawal from rollup.

## Testing
Passing all tests.

## Changelogs
No updates required.

## Related Issues
closes #1430


crates/astria-core/src/protocol/transaction/v1/action/mod.rs
```diff
@@ -1032,12 +1035,19 @@ impl Protobuf for Ics20Withdrawal {
...
+        let amount = u128::from(amount.ok_or(Ics20WithdrawalError::field_not_set("amount"))?);
         let return_address = Address::try_from_raw(
             &return_address.ok_or(Ics20WithdrawalError::field_not_set("return_address"))?,
         )
         .map_err(Ics20WithdrawalError::return_address)?;
 
+        if timeout_time == 0 {
+            return Err(Ics20WithdrawalError::invalid_timeout_time());
+        }
+        if amount == 0 {
```
ethanoroshiba, line 1047: Do we even need to check this? Protobuf represents 0 values as empty, and when decoding this it should result in `None` I think

crates/astria-core/src/protocol/transaction/v1/action/mod.rs
```diff
@@ -1047,8 +1057,20 @@ impl Protobuf for Ics20Withdrawal {
             .transpose()
             .map_err(Ics20WithdrawalError::invalid_bridge_address)?;
 
+        if bridge_address.is_some() {
+            let parsed_withdrawal: Ics20WithdrawalFromRollup =
+                serde_json::from_str(&memo).map_err(Ics20WithdrawalError::parse_memo)?;
```
ethanoroshiba, line 1062: Should we edit our domain type `Ics20Withdrawal` to then include the parsed withdrawal as well? Then, we don't have to parse again during `check_and_execute`

crates/astria-core/src/protocol/transaction/v1/action/mod.rs
```diff
@@ -1032,12 +1035,19 @@ impl Protobuf for Ics20Withdrawal {
             bridge_address,
             use_compat_address,
         } = proto;
-        let amount = amount.ok_or(Ics20WithdrawalError::field_not_set("amount"))?;
+        let amount = u128::from(amount.ok_or(Ics20WithdrawalError::field_not_set("amount"))?);
```
SuperFluffy, line 1038: IMO this looks more complicated than what we had before.

crates/astria-core/src/protocol/transaction/v1/action/mod.rs
SuperFluffy: I don't even think we should reject this. If somebody wants to send (and pay for) a transfer, why would we not permit it?

crates/astria-core/src/protocol/transaction/v1/action/mod.rs
```diff
@@ -1210,6 +1272,14 @@ enum Ics20WithdrawalErrorKind {
     InvalidBridgeAddress(#[source] AddressError),
     #[error("`denom` field was invalid")]
     InvalidDenom { source: asset::ParseDenomError },
+    #[error("`timeout_time` must be non-zero")]
+    InvalidTimeoutTime,
```
SuperFluffy, line 1276: Call this `ZeroTimeoutTime`

crates/astria-core/src/protocol/transaction/v1/action/mod.rs
```diff
@@ -1210,6 +1272,14 @@ enum Ics20WithdrawalErrorKind {
...
+    #[error("`amount` must be greater than zero")]
+    InvalidAmount,
```
SuperFluffy, line 1278: Call this `ZeroAmount`.

crates/astria-core/src/protocol/transaction/v1/action/mod.rs
```diff
@@ -1107,8 +1136,20 @@ impl Protobuf for Ics20Withdrawal {
             .transpose()
             .map_err(Ics20WithdrawalError::invalid_bridge_address)?;
 
+        if bridge_address.is_some() {
```
SuperFluffy, line 1139: I am pretty unhappy that we now have a JSON formatted payload in here. I wonder if we can make a breaking change to these transactions and add another variant to the protobuf that makes this a protobuf.

crates/astria-core/src/protocol/transaction/v1/action/mod.rs
SuperFluffy: Yes, absolutely. Easiest is probably to just make `Ics20Withdrawal` an enum over 2 variants (one where `bridge_address` is set and contains the parsed payload, and one where it's not set and just contains the passed through `String` memo).

You should also lean on a `serde(try_from = "")` to a) first read into some `UnverifiedIcs20WithdrawalFromRollup`, and then b) via a `TryFrom<UnverifiedIcs20WithdrawalFromRollup> Ics20WithdrawalFromRollup` come up with the final type.

crates/astria-core/Cargo.toml
```diff
@@ -37,6 +37,7 @@ penumbra-proto = { workspace = true }
 prost = { workspace = true }
 rand = { workspace = true }
 serde = { workspace = true, features = ["derive"], optional = true }
+serde_json = { workspace = true }
```
SuperFluffy, line 40: Will this change force us to always activate serde for all of `astria-core`? After all, the memo serde impls are feature gated I think.

crates/astria-core/src/protocol/transaction/v1/action/mod.rs
ethanoroshiba: Removed in 84b00ef

crates/astria-core/src/protocol/transaction/v1/action/mod.rs
ethanoroshiba: Split into enum 84b00ef. I wanted to ask further before continuing with your second suggestion: in order to do this we would need to implement a new `UnverifiedIcs20WithdrawalFromRollup` type and then the `TryFrom` trait as well for it. While this may be fine, it would be a little bit backwards from the way we currently do things. 

`Ics20WithdrawalFromRollup` is a raw protobuf type, which usually represents the "unverified" version as opposed to the verified one. We could rename it to reflect that it is unverified, and then create a new domain type for the verified version, but I'm not sure if the name change would be proto breaking. In addition to this, it seems like it would be many more lines for questionable benefit to just performing these verifications in `Ics20Withdrawal::try_from_raw`. Do you still think it would be worth it to do this?

crates/astria-core/src/protocol/transaction/v1/action/mod.rs
ethanoroshiba: Renamed in 84b00ef.

crates/astria-core/src/protocol/transaction/v1/action/mod.rs
ethanoroshiba: Removed entirely in 84b00ef.

crates/astria-core/src/protocol/transaction/v1/action/mod.rs
ethanoroshiba: Considering this would be network breaking, I'm not sure what we can really do about it now, barring a hard fork :/

We could make a new action which uses a protobuf-formatted memo instead and follow the same upgrade path as the new `ValidatorUpdateV2` action, but I'm not sure how worth it that would be.

crates/astria-core/Cargo.toml
ethanoroshiba: I tried to address this in 84b00ef. Let me know if this is what you had in mind!

crates/astria-bridge-contracts/src/lib.rs
```diff
@@ -454,32 +457,35 @@ where
...
             .map_err(GetWithdrawalActionsError::calculate_withdrawal_amount)?;
 
-        let action = Ics20Withdrawal {
+        let action = Ics20Withdrawal::FromRollup(Box::new(Ics20WithdrawalWithBridgeAddress {
             denom,
             destination_chain_address: event.destination_chain_address,
             return_address: self.bridge_address,
             amount,
-            memo,
+            ics20_withdrawal_from_rollup,
```
noot, line 478: why do we have this `memos::v1::Ics20WithdrawalFromRollup` and the json-encoded memo? we can probably omit the json-encoded one here, and leave that to the sequencer to convert the `memos::v1::Ics20WithdrawalFromRollup` to json before putting it in the IBC packet

crates/astria-core/src/protocol/transaction/v1/action/mod.rs
```diff
@@ -887,10 +882,20 @@ enum IbcSudoChangeErrorKind {
 ///
 /// It also contains a `return_address` field which may or may not be the same as the signer
 /// of the packet. The funds will be returned to the `return_address` in the case of a timeout.
-#[derive(Debug, Clone, PartialEq, Eq)]
-pub struct Ics20Withdrawal {
+#[derive(Debug, Clone, PartialEq)]
+pub enum Ics20Withdrawal {
+    NoBridgeAddress(Box<Ics20WithdrawalNoBridgeAddress>),
```
noot, line 887: would prefer naming like `FromSequencer` to match `FromRollup`

crates/astria-core/src/protocol/transaction/v1/action/mod.rs
```diff
@@ -907,67 +912,144 @@ pub struct Ics20Withdrawal {
...
-    // treated as a bridge withdrawal (ie. the bridge account's withdrawer address is checked).
-    pub bridge_address: Option<Address>,
-
+    pub bridge_address: Address,
     // whether to use a bech32-compatible format of the `.return_address` when generating
     // fungible token packets (as opposed to Astria-native bech32m addresses). This is
     // necessary for chains like noble which enforce a strict bech32 format.
     pub use_compat_address: bool,
+    // the memo included from the raw action. This is kept so that the action can be infallibly
+    // converted back into its raw format.
```
noot, line 954: why is infallible conversion needed?

SuperFluffy: All domain types (that have a wire counterpart) must be convertible to their wire format.

This is a design decision of the crate and something that would be a pain to lose. It's not yet enforced but should be through a trait impl (`Protobuf` at the moment, but we should rename that).

noot: like can we change `self.memo()` to return the json-serialized `ics20_withdrawal_from_rollup` if it's the `Self::FromRollup` variant?

ethanoroshiba: It's been a minute since I made these changes, so I may be misremembering, but I believe the infallibility is needed to satisfy the `Protobuf` trait implementation, and we need to store the original memo because re-serializing `ics20_withdrawal_from_rollup` is not an infallible operation.

crates/astria-core/src/protocol/transaction/v1/action/mod.rs
ethanoroshiba: Changed in 7a5329c :)

crates/astria-bridge-contracts/src/lib.rs
ethanoroshiba: Related to discussion below: https://github.com/astriaorg/astria/pull/1770#discussion_r1870169100

joroshiba:
closed in 7 days.

joroshiba:

PR #1772, opened 2024-10-31
fix(sequencer): remove unwrap from app utilized mempool logic 

### LINKED ISSUE #1769: Unclear unwrap inside core execution logic
`astria_sequencer::App::execute_transaction_prepare_proposal` [1] unwraps a `Result`. The `expect` message it provides does not give any indication why unwrapping at that point is acceptable. Tracing it through the stack the error that is being bubbled up comes from a call deep inside `astria_sequencer::Mempool::builder_queue` [2] and [3].

It is not clear why this `expect` here is acceptable or why it is an invariant of the system that the return value of of the method call is always `Ok`.

1: https://github.com/astriaorg/astria/blob/ca72c64a6513dff0c55aaa14c2a5d7d683ab8370/crates/astria-sequencer/src/app/mod.rs#L589 
2: https://github.com/astriaorg/astria/blob/ca72c64a6513dff0c55aaa14c2a5d7d683ab8370/crates/astria-sequencer/src/mempool/mod.rs#L298
3: https://github.com/astriaorg/astria/blob/ca72c64a6513dff0c55aaa14c2a5d7d683ab8370/crates/astria-sequencer/src/mempool/transactions_container.rs#L773

The code that introduced this was https://github.com/astriaorg/astria/pull/1323.

┆Issue Number: ENG-960

Lilyjjo:
## Summary
This PR changes the mempool's `builder_queue()` to be infallible. The code previously would return an error if the nonce fetch from the database failed. Now it handles the error case by using the pending's lowest nonce for the account as an educated guess of what nonce to use for transaction priority construction. This is okay since `prepare_proposal()`'s logic will just reject invalid nonces if the mempool's state is out of date due to a bug.

## Background
We shouldn't have non-existential issues in the mempool cause failures in the sequencer.

## Changes
- Changed mempool `builder_queue()` to be infallible with reasonable fallbacks.

## Testing
Manually changed the code path to only use the new logic and watched all tests pass except for those explicitly testing the jitter between the mempool and database (which shouldn't happen if the rest of the code works). The tests in those scenarios only fail due to unmet assertions and not because of panics.

## Changelogs
No updates required.

## Related Issues
closes #1769


crates/astria-sequencer/src/mempool/transactions_container.rs
```diff
@@ -784,10 +783,27 @@ impl PendingTransactions {
         let mut queue = Vec::with_capacity(self.len());
         // Add all transactions to the queue.
         for (address, account_txs) in &self.txs {
-            let current_account_nonce = state
-                .get_account_nonce(address)
-                .await
-                .wrap_err("failed to fetch account nonce for builder queue")?;
+            let current_account_nonce = match state.get_account_nonce(address).await {
```
noot, line 786: if this function returns an error, it's pretty severe (either we can't read from state, and the app will crash soon anyways, or we can't deserialize the storage value which would have caused an error earlier) so i think `expect()`ing here is fine

SuperFluffy: `expect`ing is not a replacement for error handling. `expect`ing should only be used for when system invariants are so badly violated that you can't recover from them at all, which is not the case here.

joroshiba:
closed in 7 days.

noot (approved):
looks good!

PR #1774, opened 2024-10-31
Apply new storage conventions to Connect oracle

Fraser999:
## Summary
This mainly brings the new oracle code in line with storage conventions formed after the initial cut of the PR.

## Background
We want a consistent approach to storage across all components.  As well as using a single serialization format (Borsh), this will allow us to iterate on the storage design in the near future.

## Changes
- Added `storage` submodules to the two sequencer modules `connect::market_map` and `connect::oracle`.  These provide the keys and values which are used to store the relevant data.
- Added snapshot tests and other unit tests for the keys.
- Added put/get tests for the extension traits, which revealed bugs in both the streaming getters of the oracle.
- Removed unneeded serde trait derives from several core types.
- Renamed `connect::marketmap` in the sequencer to `connect::market_map`

## Testing
Added several unit tests.

## Changelogs
We should update changelogs once the feature is ready to be merged to `main`.  Not done in this PR.

## Breaking Changelist
- This breaks the state snapshot tests due to changing how the oracle data is serialized in storage.


crates/astria-core/src/sequencerblock/v1/block.rs
```diff
@@ -783,21 +783,21 @@ impl SequencerBlock {
...
+            // This iterator over `data` has been advanced by 3 steps while retrieving the initial
+            // elements. The next element to be yielded will be the first of the actual rollup txs.
+            data_iter,
+            // TODO: this needs to go into the block header
+            _extended_commit_info,
+            rollup_transactions_root,
+            rollup_transactions_proof,
+            rollup_ids_root,
+            rollup_ids_proof,
+        } = take_initial_elements_from_data(data)?;
```
noot, line 797: note to self: need to allow this to not have the extended_commit_info as to be nonbreaking

noot (approved):
looks great!

PR #1781, opened 2024-11-01
fix(sequencer)!: enforce trace-prefixed assets in bridge lock actions

ethanoroshiba:
## Summary
Changed `BridgeLock` to only support trace-prefixed assets.

## Background
Previously, the bridge lock action supported both IBC and trace prefixed assets. If someone submitted a `BridgeLock` action with an `asset::IbcPrefixed` asset, the deposit would then contain the IBC-prefixed asset, and the EVM would fail since it only checks for trace prefixed assets.

## Changes
- Changed `asset` field of `BridgeLock` to type `asset::TracePrefixed`.

## Testing
Passing all tests.

## Changelogs
Changelog updated.

## Breaking Changelist
- Technically breaking, but should not be network breaking as long as no `BridgeLock` actions with IBC-prefixed assets have been submitted.

### DIFF
--- a/crates/astria-cli/src/sequencer/bridge_lock.rs
+++ b/crates/astria-cli/src/sequencer/bridge_lock.rs
@@ -10,6 +10,7 @@ use astria_core::{
 };
 use color_eyre::eyre::{
     self,
+    OptionExt as _,
     WrapErr as _,
 };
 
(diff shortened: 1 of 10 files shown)


noot (commented):
like we mentioned on the call, would be preferred to just change the emitted `Deposit` to use a trace-prefixed asset.

ethanoroshiba:
Closing in favor of this solution: https://github.com/astriaorg/astria/issues/1806

PR #1786, opened 2024-11-04
fix: index all sent packet event attributes

ethanoroshiba:
## Summary

## Background

## Changes

## Testing
How are these changes tested?

## Changelogs

## Metrics

## Breaking Changelist

## Related Issues

closes


crates/astria-sequencer/src/app/mod.rs
```diff
@@ -1325,3 +1330,15 @@ struct PostTransactionExecutionResult {
     validator_updates: Vec<tendermint::validator::Update>,
     consensus_param_updates: Option<tendermint::consensus::Params>,
 }
+
+fn index_all_send_packet_event_attributes(events: &mut [Event]) {
```
joroshiba, line 1334: Should we make this function more broad? Maybe take a list of events we want to index, or a way to index everything?

ethanoroshiba: Code would be simple to index everything, just wasn't sure how proper/useful that would be. Happy to do this though

ethanoroshiba: Done in 94416b1.

noot (approved):
the title and changelog should be updated to say all events are now indexed, not just send_packet ones. otherwise looks good!

SuperFluffy (changes requested):
Couple notes:

1. If abci events are toggled on every execution, then this should be in `execute_transactions` and the code not duplicated in every `fn execute_transactions_*` method.
2. Please expand the background section, right now it contains a placeholder.
3. These changes are observable by providing a unit test on `execute_transactions`. The test section right now is equivalent to saying "this is not tested".
4. This should further be observable by spinning up a sequencer network and checking the provided ABCI events.

crates/astria-sequencer/src/app/mod.rs
```diff
@@ -1175,7 +1175,17 @@ impl App {
...
                     .any(|act| act.is_fee_asset_change() || act.is_fee_change());
 
-        Ok(state_tx.apply().1)
+        // index all event attributes
+        let mut events = state_tx.apply().1;
+        for event in &mut events {
+            event
+                .attributes
+                .iter_mut()
+                .filter(|attr| !attr.index)
```
Fraser999, line 1184: Not sure there's any point to this `filter`?

SuperFluffy: Yeah, that filter is useless.

SuperFluffy (approved):
I have updated the patch title and description a bit (arguably, this is more chore-like or feat-like than a fix for sequencer proper).

I have also linked the relevant entry in the cometbft repository (preferable over a website that is more likely to change).


Please address @Fraser999's comment re the unnecessary filter adapter when enabled all indexing.

PR #1788, opened 2024-11-05
chore(sequencer): implement retry logic for connecting to the oracle sidecar

### LINKED ISSUE #1787: Implement blocking retry logic in sequencer startup for initial attempt to fetch oracle data
Change the initial attempt to retrieve data from the Connect oracle service so that:
* it retries as per other similar retry strategies
* it blocks until success or until a hard-coded timeout
* if timeout, execute graceful shutdown

┆Issue Number: ENG-996

Fraser999:
## Summary
This changes sequencer initialization to be blocked while attempting to connect to the oracle sidecar (if configured to use the sidecar), and to fail if repeated connection attempts fail.

## Background
This was a todo in the code.

## Changes
- Added blocking retry logic around creating a connected oracle client.

## Testing
Added a unit test to confirm the process waits for the expected duration.  We could probably do with adding gRPC mock tests for the sidecar, but I think that's outside the scope of this PR.

## Changelogs
No updates required as this is (effectively) a feature branch which will have all relevant changelogs updated as one of the final steps of completing the feature.

## Related Issues
Closes #1787


crates/astria-sequencer/src/sequencer.rs
```diff
@@ -303,3 +286,104 @@ fn spawn_signal_handler() -> SignalReceiver {
...
+    let mut oracle_client = OracleClient::new(
+        endpoint
+            .connect()
+            .await
+            .wrap_err("failed to connect to oracle sidecar")?,
+    );
+    let _ = oracle_client
+        .prices(QueryPricesRequest::default())
+        .await
+        .wrap_err("oracle sidecar responded with error to query for prices");
```
noot, line 349: i think we should move this out of the retried function and call it after we successfully connect to the client, as if the grpc call for prices fails, it indicates a more serious error that probably isn't fixable upon retry. thoughts?

Fraser999: I think this is maybe a general problem with most or even all of our retry loops like this.  We probably have some error cases which are permanent/unrecoverable, and others which are transient.  However we always treat them as transient and keep retrying.

In this specific case, I guess we could also get a transient failure of the gRPC (e.g. timeout or dropped connection)?  We should probably look to abandon the retry loop if the error indicates a permanent failure, but keep trying if not.  However, that's another can of worms, since matching on error variants isn't a lot of fun when using eyre/anyhow.

noot: yeah, maybe we can do a larger revisit of the retry functions. this is fine for now, but something to keep in mind

PR #1791, opened 2024-11-06
refactor: make extended commit info third injected tx

noot:
## Summary
make extended commit info third injected tx, and update `SequencerBlock::try_from_block_info_and_data` to handle whether the extended commit info is in the block or not.

## Background
 it makes more sense to append to the existing list of injected txs which are the two commitments at the start of the block. 

## Changes
- make extended commit info the third injected tx instead of the first
- update `SequencerBlock::try_from_block_info_and_data` to handle whether the extended commit info is in the block or not
- cleanup of related unit tests

## Testing
unit tests

## Breaking Changelist
- unbreaks `SequencerBlock::try_from_block_info_and_data`

### DIFF
--- a/crates/astria-core/src/sequencerblock/v1/block.rs
+++ b/crates/astria-core/src/sequencerblock/v1/block.rs
@@ -772,6 +772,10 @@ impl SequencerBlock {
     /// # Panics
     ///
     /// - if a rollup data merkle proof cannot be constructed.
+    #[expect(
+        clippy::too_many_arguments,
+        reason = "all arguments are required to make a block"
+    )]
     pub fn try_from_block_info_and_data(
         block_hash: [u8; 32],
         chain_id: tendermint::chain::Id,
@@ -780,13 +784,15 @@ impl SequencerBlock {
         proposer_address: account::Id,
         data: Vec<Bytes>,
         deposits: HashMap<RollupId, Vec<Deposit>>,
+        with_extended_commit_info: bool,
     ) -> Result<Self, SequencerBlockError> {
         use prost::Message as _;
 
         let InitialDataElements {
             data_root_hash,
-            // This iterator over `data` has been advanced by 3 steps while retrieving the initial
-            // elements. The next element to be yielded will be the first of the actual rollup txs.
+            // This iterator over `data` has been advanced to skip the injected transactions while
+            // retrieving the initial elements. The next element to be yielded will be
+            // the first of the actual user-submitted txs.
             data_iter,
             // TODO: this needs to go into the block header
             _extended_commit_info,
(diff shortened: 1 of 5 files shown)


crates/astria-core/src/sequencerblock/v1/block.rs
```diff
@@ -772,6 +772,10 @@ impl SequencerBlock {
     /// # Panics
     ///
     /// - if a rollup data merkle proof cannot be constructed.
+    #[expect(
```
SuperFluffy, line 775: Refactor this into a builder.

Needing all the arguments is not a good reason to ignore the lint.

crates/astria-core/src/sequencerblock/v1/block.rs
```diff
@@ -1042,6 +1045,16 @@ fn take_initial_elements_from_data(
          `data` used to construct the tree",
     );
 
+    let extended_commit_info = if with_extended_commit_info {
```
SuperFluffy, line 1048: This assigned variable is never used.

noot: it needs to be taken out of the iter either way, and next PR i will be using this value in `SequencerBlock`. should i still remove from here?

Fraser999: I think this is fine to leave as is since this is being merged to a feature branch rather than `main`.

crates/astria-sequencer/src/app/mod.rs
```diff
@@ -525,11 +525,11 @@ impl App {
...
-            Some(encoded_extended_commit_info) => {
-                std::iter::once(encoded_extended_commit_info.into())
-                    .chain(res.into_iter().chain(included_tx_bytes))
-                    .collect()
-            }
+            Some(encoded_extended_commit_info) => res
+                .into_iter()
+                .chain(std::iter::once(encoded_extended_commit_info.into()))
+                .chain(included_tx_bytes)
+                .collect(),
```
Fraser999, line 532: An `Option` can be turned into an iterator with 0 or 1 iterations via `into_iter`, so we can shorten this a little:

```rust
        // generate commitment to sequence::Actions and deposits and commitment to the rollup IDs
        // included in the block, chain on the extended commit info if `Some`, and finally chain on
        // the tx bytes.
        let txs = generate_rollup_datas_commitment(&signed_txs_included, deposits)
            .into_iter()
            .chain(encoded_extended_commit_info.map(Bytes::from).into_iter())
            .chain(included_tx_bytes)
            .collect();
```

Fraser999 (approved):
One minor non-blocking suggestion.

Issue #1790, opened 2024-11-06
Add signed auction result to bundle API


itamarreif:
The Auctioneer needs to submit the auction results to the sequencer for the results to be read by the Rollup during block execution.

The result of the auction is a [`Bundle`](https://github.com/astriaorg/astria/blob/1a69dc8ac1f763c0fd3cb919d6b92bde0f2af8b4/proto/executionapis/astria/bundle/v1alpha1/bundle.proto#L11), which is serialized into a `RollupDataSubmission` action and submitted to the sequencer-level (i.e. signed and paid for by the Auctioneer's sequencer-level account).

The Rollup will receive `RollupData`s with the sequenced data, and it will look for the Auctioneer's signed `Bundle` object for transactions to unbundle and place first in the block.

The Rollup will need to verify the Auctioneer's signature, which will have to be serialized into the `RollupDataSubmission` since the sequence does not include those in the `RollupData` events it creates from actions.

┆Issue Number: ENG-998

joroshiba:
This issue is stale because it has been open 45 days with no activity. Remove stale label or this issue
be closed in 7 days.

joroshiba:
This issue was closed because it was stale

PR #1795, opened 2024-11-07
feat(protos): add auction result to bundle protos

itamarreif:
## Summary

## Background

## Changes

## Testing
How are these changes tested?

## Changelogs

## Metrics

## Breaking Changelist

## Related Issues

closes https://github.com/astriaorg/astria/issues/1790


proto/executionapis/astria/bundle/v1alpha1/bundle.proto
```diff
@@ -24,6 +24,22 @@ message Bundle {
   bytes prev_rollup_block_hash = 4;
 }
 
+// The AuctionResult message is submitted by the Auctioneer to the rollup as a
+// `RollupDataSubmission` on the sequencer.
+// The rollup will verify the signature and public key against its configuration,
+// then unbundle the body into rollup transactions and execute them first in the
+// block.
+message AuctionResult {
```
SuperFluffy, line 32: I have to walk back my suggestion for calling this a `Result` because it would clash with a regular Rust `Result`: `Result<AuctionResult, E>`.

Let's go with `WinningBundle` instead?

SuperFluffy (changes requested):
Setting this to requesting changes to make it clear that this PR was reviewed and needs further action.

proto/executionapis/astria/bundle/v1alpha1/bundle.proto
joroshiba: I think it's probably fine to just call this a SignedBundle? That's what it is and provides a more generic option. IE this could allow it to be a bundle generated and signed by a TEE and submitted.

The fact that this signer indicates it is a Auction Result bundle is an implementation detail for the rollup?

itamarreif:
Changed the name of the envelope type to `Allocation`, following the terminology in auction theory (i.e. the auctioneer decides on an allocation).

proto/executionapis/astria/bundle/v1alpha1/bundle.proto
```diff
@@ -24,6 +24,22 @@ message Bundle {
...
+// block.
+message Allocation {
+  // The Ed25519 signature of the Auctioneer, to be verified against config by the
+  // rollup.
+  bytes signature = 1;
+  // The Ed25519 public key of the Auctioneer, to be verified against config by the
+  // rollup.
+  bytes public_key = 2;
+  // The bundle that was allocated the winning slot by the Auctioneer.
+  Bundle payload = 3;
```
SuperFluffy, line 40: I think we want to follow sequencer's `Transaction` and make this an opaque `Any` type so that we need not worry about differences in protobuf encoders.

proto/executionapis/astria/bundle/v1alpha1/bundle.proto
joroshiba: After discussion online, I'm ok with `allocation`.

Long term we will want something generic for signed data probably.

Issue #1798, opened 2024-11-11
Break out all astria-core submodules into self-standing crates


SuperFluffy:
We want to try and break out as many of astria-core submodules into free standing crates as possible.

This is a tracking issue to gather all PRs to that effect.

Goals:

1. reduce compile times for services and binaries that only need a subset of astria core types.
2. Better separation and decoupling between astria core items
3. Better maintainability

┆Issue Number: ENG-1003

joroshiba:
This issue is stale because it has been open 45 days with no activity. Remove stale label or this issue
be closed in 7 days.

joroshiba:
This issue was closed because it was stale

PR #1799, opened 2024-11-11
refactor(core): make address, crypto crates

SuperFluffy:
## Summary
Breaks out `astria_core::crypto` to crate `astria-core-crypto` and address related logic in `astria_core::primitive::v1` to `astria-core-address`

## Background

A useful sideeffect of this change is that the coupling between the different parts of astria-core is reduced (for example, by removing the rarely used utility method `SigningKey::try_address`, allowing to decouple address and crypto logic).

## Changes
- Move the module `astria_core::crypto` into new create `astria-core-crypto` and reexport under the `astria_core::crypto` alias (not breaking to consumers)
- Move address related types out of `astria_core::primitive::v1` into `astria-core-address`, reexport commonly used types from the old module a `astria-core-address` as `astria_core::primitive::v1::address` (partially breaking to consumers because publicly accessible types are moved into a new submodule)
- Remove the serde `Serialize` and `Deserialize` impls on the domain type `Address` (breaking because one needs to explicitly construct protobuf/wire type `Address`; not breaking on the wire, json format stays the same)
- Implement the `Protobuf` ("domain type") trait for `Address`, removing its inherent constructors and methods to move from/to raw protobuf types ("wire types"). Requireds to uncouple `astria-core-address` from `astria-core` (specifically, the generated protobuf types) (breaking for consumers because they now need to import the `Protobuf` trait)
- Create a thin crate `astria-core-consts` that only contains the const `ADDRESS_LENGTH` to allow decoupling of address and crypto logic.
- Remove the `SigningKey::try_address` utility method that was only ever used in tests to remove all coupling between cryptography primitives and addresses (breaking for users of this method)

## Testing
Tests were shuffled around but not removed. Snapshot tests stay in place in astria core for now (because they are specifically demanded for wire/protobuf types).

## Changelogs

## Breaking Changelist
- These changes would be breaking the public API of `astria-core`, but we don't provide any stability guarantees for it.
 
## Related Issues


SuperFluffy:
Closing in favor of #1800 and #1802

PR #1800, opened 2024-11-11
refactor(core): make crypto module into crate

SuperFluffy:
## Summary
Breaks out `astria_core::crypto` to crate `astria-core-crypto`.

## Background

A useful sideeffect of this change is that the coupling between the different parts of astria-core is reduced (for example, by removing the rarely used utility method `SigningKey::try_address`, allowing to decouple address and crypto logic).

## Changes
- Move the module `astria_core::crypto` into new create `astria-core-crypto` and reexport under the `astria_core::crypto` alias (not breaking to consumers)
- Remove the tight coupling between the crypto and address parts of the stack by removing the `SigningKey::try_address` (rarely used and only inside a few tests) method and moving `ADDRESS_LENGTH` to a thin `astria-core-consts` crate (this is a breaking change for users of `SigningKey::try_address`).

## Testing
Tests for astria core's crypto refinement types were moved to `astria-core-crypto` and left unchanged. Sequencer tests were updated to use `Address::builder` instead of `SigningKey::try_address` and otherwise also left unchanged.

## Changelogs
Changelogs updated.

## Breaking Changelist
- These changes would be breaking the public API of `astria-core` due to the removal fo `SigningKey::try_address` but we don't provide any stability guarantees for it.
 
## Related Issues


crates/astria-core-crypto/src/lib.rs
```diff
@@ -13,6 +13,7 @@ use std::{
     sync::OnceLock,
 };
 
+use astria_core_consts::ADDRESS_LENGTH;
```
SuperFluffy, line 16: This should be reexported because its part of the crate's public interface (via `SigningKey::address_bytes -> [u8; ADDRESS_LENGTH]` and `VerificationKey::address_bytes -> [u8; ADDRESS_LENGTH]`)

SuperFluffy: Done in [d22d29f](https://github.com/astriaorg/astria/pull/1800/commits/d22d29fa87208bf4c24719eb1d7a18ff1b1b84cd)

Fraser999 (approved):

PR #1801, opened 2024-11-11
refactor(sequencer): provide storage and snapshot types

### LINKED ISSUE #1435: Replace cnidarium `Storage` and `Snapshot` types
Replace these two types with our own as described in [Block State Improvements in the Sequencer](https://astria.atlassian.net/wiki/spaces/Astria/pages/36864002/Block+State+Improvements+in+the+Sequencer), but implement `cnidarium::StateRead` on the `Snapshot` (i.e. don’t provide our own `State[Read|Write]` traits yet).

Fraser999:
## Summary
This adds new storage types to sequencer and replaces bare usage of cnidarium equivalent types.

## Background
We want to become less coupled to the raw `cnidarium` APIs, and we also want to provide a cache of recently-read values.

## Changes
- Provided new `storage::Storage` type, wrapping the `cnidarium::Storage` type.
- Provided new `storage::Snapshot` type, wrapping the `cnidarium::Snapshot` type and including a cache of fetched values.
- When the sequencer main loop has finished, we now await the server handle and then call `storage.release().await;` which should probably have been the case before this PR.
- Applied consistent naming to variables of type `StateDelta<Snapshot>` and `StateDelta<StateDelta<Snapshot>>`.

## Testing
Added unit tests for the new types covering the non-trivial methods (many new methods simply call directly through to the equivalent inner method and don't need independent unit tests).

## Changelogs
Changelogs updated.

## Metrics
- Added `astria_sequencer_verifiable_cache_hit` counter
- Added `astria_sequencer_verifiable_cache_miss` counter
- Added `astria_sequencer_verifiable_cache_item_total` histogram
- Added `astria_sequencer_non_verifiable_cache_hit` counter
- Added `astria_sequencer_non_verifiable_cache_miss` counter
- Added `astria_sequencer_non_verifiable_cache_item_total` histogram

## Related Issues
Closes #1435.

### DIFF
--- a/crates/astria-sequencer/Cargo.toml
+++ b/crates/astria-sequencer/Cargo.toml
@@ -32,12 +32,14 @@ telemetry = { package = "astria-telemetry", path = "../astria-telemetry", featur
   "display",
 ] }
 
+anyhow = "1.0.89"
 borsh = { version = "1.5.1", features = ["bytes", "derive"] }
 cnidarium = { git = "https://github.com/penumbra-zone/penumbra.git", tag = "v0.80.7", features = [
   "metrics",
 ] }
 ibc-proto = { version = "0.41.0", features = ["server"] }
 matchit = "0.7.2"
+quick_cache = "0.6.9"
 tower = "0.4"
 tower-abci = "0.12.0"
 tower-actor = "0.1.0"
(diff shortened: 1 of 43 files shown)


crates/astria-sequencer/Cargo.toml
```diff
@@ -32,12 +32,14 @@ telemetry = { package = "astria-telemetry", path = "../astria-telemetry", featur
   "display",
 ] }
 
+anyhow = "1.0.89"
```
SuperFluffy, line 35: There was a big effort to not have `anyhow` be a direct dependency of sequencer. Can we avoid reintroducing it here?

Fraser999: I think we might be able to do that as part of #1436, but for now it's required since we're implementing [`cnidarium::StateRead`](https://rustdoc.penumbra.zone/main/cnidarium/trait.StateRead.html) on the new `Snapshot` type and that requires `anyhow` (`cnidarium` doesn't re-export it either).

joroshiba:
closed in 7 days.

PR #1802, opened 2024-11-12
refactor(core): move address logic to crate

SuperFluffy:
## Summary
Breaks out address related logic in `astria_core::primitive::v1` to `astria-core-address`

## Background

## Changes
- Move address related types out of `astria_core::primitive::v1` into `astria-core-address`
- Reexport all moved types under the same module (aliasing, where necessary)
- Remove the serde `Serialize` and `Deserialize` impls on the domain type `Address` (breaking because one needs to explicitly construct protobuf/wire type `Address`; not breaking on the wire, json format stays the same)
- Implement the `Protobuf` ("domain type") trait for `Address`, removing its inherent constructors and methods to move from/to raw protobuf types ("wire types"); decouples `astria-core-address` from `astria-core` (specifically, the generated protobuf types) (breaking for consumers because they now need to import the `Protobuf` trait)
- Create a thin crate `astria-core-consts` that only contains the const `ADDRESS_LENGTH` to allow decoupling of address and crypto logic (addressed in a separate patch)
- Removed `AddressBuilder::with_iter` from the public interface (only used inside the crate and not outside) (breaking for consumers of that method)

## Testing
Tests were shuffled around but not removed. Snapshot tests stay in place in astria core for now (because they are required for wire/protobuf types).

## Changelogs
Changelogs updated.

## Breaking Changelist
- If astria-core gave stability guarantees, this would be a breaking change for consumers of `astria-core` because some inherent methods were removed from its public API.
 
## Related Issues


crates/astria-core-consts/Cargo.toml
SuperFluffy, line 1: Note that this new crate is shared with https://github.com/astriaorg/astria/pull/1800

Fraser999 (approved):

PR #1804, opened 2024-11-12
fix: avoid redirect operator in justfile docker commands

Fraser999:
## Summary
This replaces `sed '...' <<< text` with `echo text | sed '...'` in the main justfile.

## Background
On Ubuntu, the default shell doesn't recognise the redirect operator `<<<`, hence the `just` commands `docker-build` and `docker-build-and-load` both fail on Ubuntu.

## Changes
- Replaced usage of `<<<` by piping `echo` output.

## Testing
Ran the commands on Ubuntu and macOS.

## Changelogs
No updates required.

### DIFF
--- a/justfile
+++ b/justfile
@@ -14,11 +14,11 @@ default_repo_name := 'ghcr.io/astriaorg'
 
 # Builds docker image for the crate. Defaults to 'local' tag.
 docker-build crate tag=default_docker_tag repo_name=default_repo_name:
-  docker buildx build --load --build-arg TARGETBINARY={{crate}} -f containerfiles/Dockerfile -t {{repo_name}}/$(sed 's/^[^-]*-//' <<< {{crate}}):{{tag}} .
+  docker buildx build --load --build-arg TARGETBINARY={{crate}} -f containerfiles/Dockerfile -t {{repo_name}}/$(echo {{crate}} | sed 's/^[^-]*-//'):{{tag}} .
 
 docker-build-and-load crate tag=default_docker_tag repo_name=default_repo_name:
   @just docker-build {{crate}} {{tag}} {{repo_name}}
-  @just load-image $(sed 's/^[^-]*-//' <<< {{crate}}) {{tag}} {{repo_name}}
+  @just load-image $(echo {{crate}} | sed 's/^[^-]*-//') {{tag}} {{repo_name}}
 
 # Installs the astria rust cli from local codebase
 install-cli:


justfile
```diff
@@ -14,11 +14,11 @@ default_repo_name := 'ghcr.io/astriaorg'
 
 # Builds docker image for the crate. Defaults to 'local' tag.
 docker-build crate tag=default_docker_tag repo_name=default_repo_name:
-  docker buildx build --load --build-arg TARGETBINARY={{crate}} -f containerfiles/Dockerfile -t {{repo_name}}/$(sed 's/^[^-]*-//' <<< {{crate}}):{{tag}} .
+  docker buildx build --load --build-arg TARGETBINARY={{crate}} -f containerfiles/Dockerfile -t {{repo_name}}/$(echo {{crate}} | sed 's/^[^-]*-//'):{{tag}} .
```
SuperFluffy, line 17: These commands are broken on my system (also before this patch) because I am neither running posix shell nor bash. Let's fix it for all users by prepending it with a `#!/usr/bin/env bash` or `#!/usr/bin/env sh`

SuperFluffy (changes requested):
This patch is needed but still broken. Let's fix it for all users.

SuperFluffy:
On a second note: we should probably just list the commands to build services so that we don't have to rely on the mapping `crates/astria/<SERVICE> -> `ghcr.io/astriaorg/<SERVICE>` always working.

We only have a handful of services so this wouldn't cause too much boilerplate.

justfile
```diff
@@ -13,12 +13,41 @@ default_docker_tag := 'local'
...
 
 # Builds docker image for the crate. Defaults to 'local' tag.
-docker-build crate tag=default_docker_tag repo_name=default_repo_name:
-  docker buildx build --load --build-arg TARGETBINARY={{crate}} -f containerfiles/Dockerfile -t {{repo_name}}/$(echo {{crate}} | sed 's/^[^-]*-//'):{{tag}} .
-
-docker-build-and-load crate tag=default_docker_tag repo_name=default_repo_name:
-  @just docker-build {{crate}} {{tag}} {{repo_name}}
-  @just load-image $(echo {{crate}} | sed 's/^[^-]*-//') {{tag}} {{repo_name}}
+# NOTE: `_crate_short_name` is invoked as dependency of this command so that failure to pass a valid
+# crate will produce a maningful error message.
```
SuperFluffy, line 17:
```suggestion
# binary will produce a meaningful error message.
```

justfile
```diff
@@ -13,12 +13,41 @@ default_docker_tag := 'local'
...
+_crate_short_name crate quiet="":
+  #!/usr/bin/env sh
+  set -eu
+  case {{crate}} in
+    astria-bridge-withdrawer) short_name=bridge-withdrawer ;;
+    astria-composer) short_name=composer ;;
+    astria-conductor) short_name=conductor ;;
+    astria-sequencer) short_name=sequencer ;;
+    astria-sequencer-relayer) short_name=sequencer-relayer ;;
+    *) echo "{{crate}} is not a supported crate" && exit 2
```
SuperFluffy, line 48: I think we should call this binary because at the level of the dockerfiles users should not have to care about Rust idioms.
```suggestion
    *) echo "{{crate}} is not a supported binary" && exit 2
```

justfile
```diff
@@ -13,12 +13,41 @@ default_docker_tag := 'local'
...
-
-docker-build-and-load crate tag=default_docker_tag repo_name=default_repo_name:
-  @just docker-build {{crate}} {{tag}} {{repo_name}}
-  @just load-image $(echo {{crate}} | sed 's/^[^-]*-//') {{tag}} {{repo_name}}
+# NOTE: `_crate_short_name` is invoked as dependency of this command so that failure to pass a valid
+# crate will produce a maningful error message.
+docker-build crate tag=default_docker_tag repo_name=default_repo_name: (_crate_short_name crate "quiet")
+  #!/usr/bin/env sh
+  set -eu
+  short_name=`just _crate_short_name {{crate}}`
```
SuperFluffy, line 21:
```suggestion
  short_name=$(just _crate_short_name {{crate}})
```
My comment re command substitution was due to my shell (fish), but setting `#!/usr/bin/env sh` fixes this (and backticks is probably more confusing nowadays than `$()`).

justfile
```diff
@@ -13,12 +13,41 @@ default_docker_tag := 'local'
...
+  set -x
+  docker buildx build --load --build-arg TARGETBINARY={{crate}} -f containerfiles/Dockerfile -t {{repo_name}}/$short_name:{{tag}} .
+
+# Builds and loads docker image for the crate. Defaults to 'local' tag.
+# NOTE: `_crate_short_name` is invoked as dependency of this command so that failure to pass a valid
+# crate will produce a maningful error message.
+docker-build-and-load crate tag=default_docker_tag repo_name=default_repo_name: (_crate_short_name crate "quiet")
+  #!/usr/bin/env sh
+  set -eu
+  short_name=`just _crate_short_name {{crate}}`
```
SuperFluffy, line 31:
```suggestion
  short_name=$(just _crate_short_name {{crate}})
```

justfile
```diff
@@ -13,12 +13,41 @@ default_docker_tag := 'local'
...
+  set -x
+  just docker-build {{crate}} {{tag}} {{repo_name}}
+  just load-image $short_name {{tag}} {{repo_name}}
+
+# Maps a crate name to the shortened name used in the docker tag.
+# If `quiet` is an empty string the shortened name will be echoed. If `quiet` is a non-empty string,
+# the only output will be in the case of an error, where the input `crate` is not a valid one.
+_crate_short_name crate quiet="":
```
SuperFluffy, line 39: Instead of prefixing with `_`, maybe an attribute makes it clearer?
```suggestion
#[private]
crate_short_name crate:
```
You will note I also removed the `quiet` argument - I am not 100% we need it? I notice it is being used as an input to both `docker-*` recipes, but you also have `set -eu` to fail early. Can we just rely on that or is there extra use with having the arg?

justfile
```diff
@@ -13,12 +13,41 @@ default_docker_tag := 'local'
...
+  #!/usr/bin/env sh
+  set -eu
+  case {{crate}} in
```
SuperFluffy, line 42: I think we are also building the CLI nowadays? Not 100% sure.

justfile
Fraser999: I'll make the change suggested here, but I don't want to change all usages of the term "crate" in this file (some of which predate this PR).  I think "crate" is ubiquitous in Rust, to the point that if a reader is seeing `cargo` being used in this file, they'll very likely understand the term "crate" in this context.

justfile
Fraser999: As above

justfile
Fraser999: We're using the leading underscore style elsewhere in this file rather than the `[private]` attribute, so I'll leave as is.

We need the quiet arg to silence the output on success.  The recipe is being invoked twice in each of the two parent recipes.

The first time it is being specified as a dependency so that it is invoked before running the parent.  This is purely to validate the specified crate is a supported one.  If that invocation fails, we get a clear error message and exit code 2.  In the success case, without the `quiet` arg, we'd get the shortened crate name printed to the console, which is confusing and needless noise.  So in that invocation, we set `quiet` to a non-empty string to suppress the output on success.

The second time it is called is in the body of the parent recipe, and in that case we require the successful output to populate our shell variable `short_name`, so we set `quiet` to an empty string.

Without the first call (when we call it as a dependency), if we only called in the body and the specified crate was invalid, we'd get no helpful error message.

justfile
SuperFluffy: Yeah, most of the usage I consider fine - just the outside facing messages I think should not be about rust (focusing on the output, the service/image, instead of the input).

justfile
SuperFluffy: Disagree: just uses that for its own DSL (the portion you linked), while here we have an inline script and command substitution is considered better practice and is strictly preferred over backticks.

justfile
SuperFluffy: Thank you for the explanation!

justfile
Fraser999: Done in [c88db3b](https://github.com/astriaorg/astria/pull/1804/commits/c88db3b0d0a5d043baaf0816dbceea307555dec8).

justfile
Fraser999: Done in [c88db3b](https://github.com/astriaorg/astria/pull/1804/commits/c88db3b0d0a5d043baaf0816dbceea307555dec8).

SuperFluffy (approved):
Please check if we need to also add the CLI to the list of built services.

justfile
Fraser999: Done in [c88db3b](https://github.com/astriaorg/astria/pull/1804/commits/c88db3b0d0a5d043baaf0816dbceea307555dec8).

quasystaty1 (approved):

quasystaty1:
There's a minor issue with the build logic for the cli Docker image unrelated to this PR. The `astria-cli `package is the only one prefixed with astria, causing the just command to tag the local image as `ghcr.io/astriaorg/cli:local instead of ghcr.io/astriaorg/astria-cli:local`. For consistency and specific use cases, this should ideally be updated, though it’s not something we use regularly

Fraser999:
Agreed with @quasystaty1 that the CLI issue will be handled in a follow-up PR.

PR #1807, opened 2024-11-12
fix(sequencer)!: ensure deposits use trace-prefixed assets

### LINKED ISSUE #1806: make all deposits trace-prefixed
Currently, `BridgeLock` actions use whichever asset was included in the action when constructing their corresponding `Deposit`. Our EVM cannot interpret IBC-prefixed assets, and we should not expect others to, either. We should make sure that any IBC-prefixed assets included in a bridge lock action are matched to their trace-prefixed counterpart to be included in the `Deposit`.

┆Issue Number: ENG-1004

ethanoroshiba:
## Summary
Ensures deposits created within bridge lock actions use only trace-prefixed assets.

## Background
Currently, deposits created within a bridge lock action include whichever asset is used in the action, which can be either IBC- or trace-prefixed. This poses a problem, since our EVM has no concept of IBC-prefixed assets, and we shouldn't expect others using our network to, either. One solution discussed was to enforce all `BridgeLock` actions to use trace prefixed assets, but this would be a consensus breaking change. This aims to remedy the bug whilst making no breaking changes.

## Changes
- Map any IBC-prefixed assets included in a `BridgeLock` action to trace-prefixed when constructing the `Deposit`.

## Testing
- Added unit test to ensure an IBC-prefixed asset included in a bridge lock action is correctly mapped to a trace-prefixed asset in the corresponding deposit.
- Tested that change would be non-breaking by syncing a full node to mainnet

## Changelogs
Changelog updated.

## Breaking Changes
Technically, this change is breaking, since any sequencer version prior to this which received a `BridgeLock` action containing an IBC-prefixed asset would emit a `Deposit` with an IBC-prefixed asset, while any sequencer version after would only emit deposits with trace-prefixed assets. However, owing to the fact that no such cases have happened yet on mainnet, this change would not be practically breaking so long as no `BridgeLock` actions containing IBC-prefixed denoms have been submitted by the time this patch is pushed out. This is actively being tested by a full-node running this PR against mainnet.

## Steps to sync this PR to Mainnet on your local machine

Ensure you have docker installed + running

### Download and build Sequencer
```bash
git clone https://github.com/astriaorg/astria.git
cd astria
git checkout ENG-1004/trace_prefixed_deposits
cargo build -p astria-sequencer --release
just docker-build astria-sequencer
```

### Helm install
```bash
helm repo add astria https://astriaorg.github.io/charts/
```
Download this file and put in the `astria` repo: https://github.com/astriaorg/networks/blob/main/astria/full-node-values.yaml

Make these edits to `full-node-values.yaml`, replacing "your_moniker" with your chosen moniker:

```yaml
namespace: "astria-node"
moniker: `<your_moniker>`
images:
     sequencer:
          tag: local
cometbft:
     secrets:
          nodeKey:
               devContent:
                    priv_key:
                         value: 1yh4XrMHn75sSW5cOhGDTVgv5BbqXlhrLduxHcE2t1osbwKQzo7xlvSK1vh5CVDvHESPYK/56uTKXM/1ifqHbw==
     privValidatorKey:
          devContent:
               address: 8C17BBDC7C350C83C550163458FC9B7A5B54A64E
               pub_key:
                    value: 4v1RdMiKkWgBBTTP26iRSLOEkAY99gMVfZijm6OCzjs=
               priv_key:
                    value: WxsUPI2QASIwK+XV547R8xzL66a5oBR2G8h53t0Gcl7i/VF0yIqRaAEFNM/bqJFIs4SQBj32AxV9mKObo4LOOw==
sequencer:
     metrics:
          enabled: false
     otel:
          enabled: false
storage:
     local: true
     entities:
          sequencerSharedStorage:
               storageClassName: <your_moniker>-sequencer-shared-storage-local
```

From the `astria` repo:
```bash
just deploy cluster
just deploy ingress-controller
just wait-for-ingress-controller
just load-image sequencer
helm install astria-full-node astria/sequencer --version 1.0.0-rc.4 \
  --namespace astria-node --create-namespace -f full-node-values.yaml
```

After this you should be able to go into k9s and see both the sequencer and CometBFT running

## Related Issues
closes #1806

### DIFF
--- a/crates/astria-sequencer/src/bridge/bridge_lock_action.rs
+++ b/crates/astria-sequencer/src/bridge/bridge_lock_action.rs
@@ -20,6 +20,7 @@ use crate::{
     },
     address::StateReadExt as _,
     app::ActionHandler,
+    assets::StateReadExt as _,
     bridge::{
         StateReadExt as _,
         StateWriteExt as _,
@@ -68,11 +69,21 @@ impl ActionHandler for BridgeLock {
             .expect("current source should be set before executing action")
             .source_action_index;
 
+        // map asset to trace prefixed asset for deposit, if it is not already
+        let deposit_asset = match self.asset.as_trace_prefixed() {
+            Some(asset) => asset.clone(),
+            None => state
+                .map_ibc_to_trace_prefixed_asset(&allowed_asset)
+                .await
+                .wrap_err("failed to map IBC asset to trace prefixed asset")?
+                .ok_or_eyre("mapping from IBC prefixed bridge asset to trace prefixed not found")?,
+        };
+
         let deposit = Deposit {
             bridge_address: self.to,
             rollup_id,
             amount: self.amount,
-            asset: self.asset.clone(),
+            asset: deposit_asset.into(),
             destination_chain_address: self.destination_chain_address.clone(),
             source_transaction_id,
             source_action_index,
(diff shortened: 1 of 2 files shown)


SuperFluffy:
Please add a section `# Breaking Changes` that clarifies why this is not a breaking change. This has all the ingredients to be breaking but turns out not to be due to specific implementation details of sequencer.

crates/astria-sequencer/src/bridge/bridge_lock_action.rs
```diff
@@ -68,11 +69,21 @@ impl ActionHandler for BridgeLock {
...
 
+        // map asset to trace prefixed asset for deposit, if it is not already
+        let deposit_asset = match self.asset.as_trace_prefixed() {
+            Some(asset) => asset.clone(),
+            None => state
+                .map_ibc_to_trace_prefixed_asset(&allowed_asset)
+                .await
+                .wrap_err("failed to map IBC asset to trace prefixed asset")?
+                .ok_or_eyre("mapping from IBC prefixed bridge asset to trace prefixed not found")?,
+        };
```
noot, line 80:
```suggestion
        let deposit_asset = if self.asset.is_ibc_prefixed() {
            state
                .map_ibc_to_trace_prefixed_asset(&allowed_asset)
                .await
                .wrap_err("failed to map IBC asset to trace prefixed asset")?
                .ok_or_eyre("mapping from IBC prefixed bridge asset to trace prefixed not found")?
        } else {
            self.asset
        };
```

don't need to close `self.asset`

noot (approved):
small suggestion, otherwise looks good + thank you for the detailed explanation of why this is non-breaking

SuperFluffy (approved):

PR #1814, opened 2024-11-18
chore(bridge-withdrawer): accumulating counter instead of guage for settled value

joroshiba:
## Summary
Replaces the guage metric `batch_total_settled_value` with the counter metric `accumulated_settled_value`

## Background
The guage metric was not useful because we were not capturing the guage at each packet time, it was most likely going to always be zero. Instead replace with a counter which we can run a rate on to see the total amount of flow out. Because this is strictly increasing we aren't subject to different outputs with different scraping intervals. 

## Changes
- remove `batch_total_settled_value` and replace with `accumulated_settled_value`

## Testing
How are these changes tested?

## Changelogs
Changelogs updated

## Metrics
- Remove `batch_total_settled_value`
- Added `accumulated_settled_value` metric

## Breaking Changelist
This is technically speaking a breaking change on the metric interface, however the existing metric never worked, so I wouldn't consider truly breaking.

### DIFF
--- a/crates/astria-bridge-withdrawer/src/metrics.rs
+++ b/crates/astria-bridge-withdrawer/src/metrics.rs
@@ -18,7 +18,7 @@ pub struct Metrics {
     nonce_fetch_latency: Histogram,
     sequencer_submission_failure_count: Counter,
     sequencer_submission_latency: Histogram,
-    batch_total_settled_value: Gauge,
+    accumulated_settled_value: Gauge,
 }
 
 impl Metrics {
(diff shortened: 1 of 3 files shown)


crates/astria-bridge-withdrawer/src/metrics.rs
```diff
@@ -97,10 +97,10 @@ impl metrics::Metrics for Metrics {
             )?
             .register()?;
 
-        let batch_total_settled_value = builder
+        let accumulated_settled_value = builder
             .new_gauge_factory(
-                BATCH_TOTAL_SETTLED_VALUE,
-                "Total value of withdrawals settled in a given sequencer block",
+                ACCUMULATED_SETTLED_VALUE,
+                "Total value of withdrawals settled over time from batches",
```
Fraser999, line 103: Would be good to include the units used here.  Also maybe worth replacing `_VALUE` in the metric name with the units too (or appending it to the name)

joroshiba:
closed in 7 days.

joroshiba:

PR #1819, opened 2024-11-18
chore(sequencer): refactor app

### LINKED ISSUE #1785: sequencer: refactor app
There is a lot of repeated code between `execute_transactions_prepare_proposal()`, `execute_transactions_process_proposal()`, and `finalize_block()`. We should refactor to use shared code where possible

┆Issue Number: ENG-995

ethanoroshiba:
## Summary
Refactored `<prepare/process>_proposal_execute_transactions` methods.

## Background
These methods previously contained a lot of shared code. This is meant to eliminate this code duplication and also make the steps that are happening in the proposals clearer and easier to understand.

## Changes
- Renamed `prepare_proposal_execute_transactions` to `prepare_proposal_tx_execution` to better represent that this is the execution step of the larger `prepare_proposal` method.
- Renamed `process_proposal_execute_transactions` to `process_proposal_tx_execution`.
- Refactored block size checks and tx execution to shared helper functions, which take a new enum `Proposal` containing information necessary for either `Prepare` or `Process`.

## Testing
Passing all tests.

## Changelogs
No updates required

## Related Issues
closes #1785


crates/astria-sequencer/src/app/mod.rs
```diff
@@ -1333,3 +1195,243 @@ struct PostTransactionExecutionResult {
...
+        .fold(0usize, |acc, seq| acc.saturating_add(seq.data.len()));
+    let tx_bytes = tx.to_raw().encode_to_vec();
+    let tx_hash_base_64 =
+        telemetry::display::base64(tx_hash.unwrap_or_else(|| Sha256::digest(&tx_bytes).into()))
+            .to_string();
+    let tx_len = tx_bytes.len();
+    let debug_msg = match proposal_info {
+        Proposal::Prepare(_) => "excluding transaction",
+        Proposal::Process(_) => "transaction error",
+    };
```
Fraser999, line 1311: I think it's generally better to only create new variables right before they're needed rather than en masse at the start of a method.  Generally it makes the code simpler to understand, since you don't need to remember how a given variable was constructed/what it represents at the point it's used.  It also means we don't waste cycles constructing items which might never get used in the case of an early return.

crates/astria-sequencer/src/app/mod.rs
```diff
@@ -1333,3 +1195,243 @@ struct PostTransactionExecutionResult {
...
+#[derive(PartialEq)]
+enum ExitContinue {
+    Exit,
+    Continue,
+}
+
+enum Proposal<'a> {
+    Prepare(PrepareProposalInformation<'a>),
+    Process(ProcessProposalInformation),
+}
```
Fraser999, line 1208: I'm not sure there's much benefit to making separate structs for each of the variants.  I guess I'd have just flattened all this into the enum:

```suggestion
enum Proposal {
    Prepare {
        validated_txs: Vec<bytes::Bytes>,
        included_signed_txs: Vec<Transaction>,
        failed_tx_count: usize,
        execution_results: Vec<ExecTxResult>,
        excluded_txs: usize,
        current_tx_group: Group,
        mempool: Mempool,
        metrics: &'static Metrics,
    },
    Process {
        execution_results: Vec<ExecTxResult>,
        current_tx_group: Group,
    },
}
```

crates/astria-sequencer/src/app/mod.rs
```diff
@@ -1333,3 +1195,243 @@ struct PostTransactionExecutionResult {
     validator_updates: Vec<tendermint::validator::Update>,
     consensus_param_updates: Option<tendermint::consensus::Params>,
 }
+
+#[derive(PartialEq)]
+enum ExitContinue {
+    Exit,
+    Continue,
+}
```
Fraser999, line 1203:
```suggestion
enum BreakOrContinue {
    Break,
    Continue,
}
```

Fraser999 (commented):
Generally looks good.  A couple of minor suggestions though.

crates/astria-sequencer/src/app/mod.rs
ethanoroshiba: Made requested changes in 078bad3 :)

Fraser999 (approved):

Issue #1823, opened 2024-11-19
proto: remove optimistic block generated proto files


bharath-123:
After https://github.com/astriaorg/astria/pull/1707 got merged, there are a couple of generated proto files which is basically dead code due to naming and versioning changes in https://github.com/astriaorg/astria/pull/1707. 

We should remove those files and also update https://github.com/astriaorg/astria/blob/main/crates/astria-core/src/generated/mod.rs#L169 to optimistic

┆Issue Number: ENG-1021

bharath-123:
This issue is being fixed in: https://github.com/astriaorg/astria/pull/1825

PR #1825, opened 2024-11-19
fix(ci, core)!: ensure committed and code generated from protobuf spec match

SuperFluffy:
## Summary
Deletes and regenerates all code in `astria-core/src/generated`. Fixes bad imports by nesting generated Astria APIs under a new `astria` module. Fixes badly named modules.

## Background
https://github.com/astriaorg/astria/pull/1236 noticed that some protobuf files were skipped during code generation because of a `tonic_build::Builder::extern_path` statement. This showed up again when reworking the protobuf compilation tool to first clear the target directory for storing the generated rust files, where now an `include!` statement in `astria_core::generated` failed because the file was not in fact generated. This was not detected because the file was still committed to the repository at an earlier point.

https://github.com/astriaorg/astria/pull/1707 then added new generated code under a wrongly named `optimistic_block` module, which should have been named `optimistic` to be in line with its protobuf package counterpart.

The present change is primarily to the `tools/protobuf-compiler` binary, which first purges the output directory (minus the handwritten `mod.rs`) before generating new code.

## Changes
- Update `tools/protobuf-compiler` to clear `astria-core/src/generated` prior to repopulating it from `proto/`
- Remove the module `sequencerblock::optimisticblock`
- Add module `sequencerblock::optimistic`
- Update all services that import Astria APIs from `astria_core::generated` to `astria_core::generated::astria`.

## Testing
This is just code organization. Import paths were updated, code still compiles and tests pass.

## Changelogs
Changelogs updated.

## Breaking Changelist
- This is a breaking change for all consumers of `astria_core` that now need to import many items from `astria_core::generated::astria` that were directly under `astria_core::generated` before.

## Override Freeze
This code touches services by changing their imports (`astria_core::generated -> astria_core::generated::astria`) but does change any implementation details.

## Related Issues
Fixes and amends https://github.com/astriaorg/astria/pull/1707.
Supersedes https://github.com/astriaorg/astria/pull/1824
Closes https://github.com/astriaorg/astria/issues/1823


tools/protobuf-compiler/CHANGELOG.md
Fraser999, line 1: I doubt if we need a changelog for this crate, since it'll probably never be published or used externally.

tools/protobuf-compiler/src/main.rs
```diff
@@ -206,7 +204,8 @@ fn get_buf_from_env() -> PathBuf {
         "linux" => "You can download it from https://github.com/bufbuild/buf/releases; if you are on Arch Linux, install it from the AUR with `rua install buf` or another helper",
         _other =>  "Check if there is a precompiled version for your OS at https://github.com/bufbuild/buf/releases"
     };
-    let error_msg = "Could not find `buf` installation and this build crate cannot proceed without
+    let error_msg = "Could not find `buf` installation and this build crate cannot proceed \
+                     without
     this knowledge. If `buf` is installed and this crate had trouble finding
     it, you can set the `BUF` environment variable with the specific path to your
     installed `buf` binary.";
```
Fraser999, line 211: Would be good to tidy up this string.

tools/protobuf-compiler/src/main.rs
```diff
@@ -217,3 +216,25 @@ fn get_buf_from_env() -> PathBuf {
         .or_else(|| which::which("buf").ok())
         .expect(&msg)
 }
+
+fn purge_out_dir(path: impl AsRef<Path>) {
+    for entry in read_dir(path)
+        .expect("should be able to read target folder for generated files")
```
Fraser999, line 222: Would be helpful to use `path.display()` instead of "target folder" or "out dir" in the `expect` calls here and below.

tools/protobuf-compiler/CHANGELOG.md
SuperFluffy: IMO it's still useful and good practice to have it.

tools/protobuf-compiler/src/main.rs
SuperFluffy: Done in [6d0092d](https://github.com/astriaorg/astria/pull/1825/commits/6d0092d8130dea8717f5f924b729ea25ef8e634c)

tools/protobuf-compiler/src/main.rs
SuperFluffy: Done in [6d0092d](https://github.com/astriaorg/astria/pull/1825/commits/6d0092d8130dea8717f5f924b729ea25ef8e634c)

crates/astria-core/src/generated/mod.rs
joroshiba, line 1: Could we move this out of the generated folder since it is a manually edited file? I recognize that this breaks our convention generally, but I think for generated files the convention should be a directory should be wholly generated files.

joroshiba (approved):
Approving with note that I think we should migrate the generated mod.rs

PR #1827, opened 2024-11-20
fix(charts): configurable components block times variables

quasystaty1:
## Summary
Configurable block time environment variables for conductor and sequencer-relayer components
## Background
Currently, `ASTRIA_CONDUCTOR_CELESTIA_BLOCK_TIME_MS` and `ASTRIA_SEQUENCER_RELAYER_BLOCK_TIME` environment  variables are hard coded in the charts, the variables determines polling frequency and thus should be easily configurable. 

## Changes
- configurable blocktime variables for conductor and sequencer relayer components.

## Testing
locally templating the chart with new default values.
## Changelogs
No updates required

### DIFF
--- a/charts/evm-rollup/values.yaml
+++ b/charts/evm-rollup/values.yaml
@@ -173,7 +173,7 @@ config:
     # rate.
     sequencerBlockTimeMs: 2000
     # The expected fastest block time possible from DA, determines polling rate.
-    celestiaBlockTimeMs: 12000
+    celestiaBlockTimeMs: 6000
     # URL path for the sequencer
     sequencerRpc: ""
     # gRPC path for the sequencer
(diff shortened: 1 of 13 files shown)


charts/evm-rollup/values.yaml
```diff
@@ -172,6 +172,8 @@ config:
     # The expected fastest block time possible from sequencer, determines polling
     # rate.
     sequencerBlockTimeMs: 2000
+    # The expected fastest block time possible from DA, determines polling rate.
+    celestiaBlockTimeMs: 12000
```
joroshiba, line 176: Probably update this default.

WafflesVonMaple (approved):

joroshiba (approved):

charts/evm-rollup/values.yaml
quasystaty1: Should update `celestia-local` blocktime as well then

Issue #1818, opened 2024-11-18
conductor: parse oracle data from `extended_commit_info` and send to rollup if it exists


noot:
once the extended commit info is inside `SequencerBlock` and written to DA, conductor can access it and send the data inside of it to rollups.

conductor does not need to do additional validation on the extended commit info, as it will only be finalized by the sequencer if valid. 

the `RollupData` enum should be updated to contain an `OracleData` variant, which can be a vec of prices such as:
```
message Price {
  CurrencyPair currency_pair = 1;
  QuotePrice price = 2;
  uint64 decimals = 3;
}
```

conductor then needs to parse the extended commit info into a `OracleData` by parsing the prices and taking the median (logic which is already implemented in sequencer) and then pass this as a `RollupData` to the rollup.

┆Issue Number: ENG-1017

PR #1832, opened 2024-11-21
feat(sequencer): store currency pair id mapping along with committed extended commit info

noot:
## Summary
the price data posted by validators is of the form currency pair ID -> price, however when this is sent to a rollup there is then no way to determine the ID -> actual currency pair without referencing the sequencer state. this is non ideal as we don't want to have to perform lookups using a sequencer node for the oracle data.

## Background
see above

## Changes
- create `ExtendedCommitInfoWithCurrencyPairMapping` type which contains `ExtendedCommitInfo` (as previous) as well as the ID->currency pair mapping for the price data inside that `ExtendedCommitInfo`
- the proposer fills this mapping using the current sequencer state (before tx execution)
- encoded `ExtendedCommitInfoWithCurrencyPairMapping` is then used as the third injected tx in the block, whereas before it was just `ExtendedCommitInfo`
- the other validators then verify this mapping using the current sequencer state (before tx execution) in `process_proposal`

## Testing
unit tests + CI tests

## Related Issues
required for #1818


Fraser999 (approved):
This looks correct to me, but I'd like to see a test for the new validation functionality provided by `validate_id_to_currency_pair_mapping`.  However, as discussed offline, this is probably best done on top of the `noot/more-tests` branch where there are a bunch of other fixes and additional tests, so I'm fine to merge this, then prioritise adding the test into that branch and opening a PR.

PR #1834, opened 2024-11-22
chore(composer): add missing blackbox tests

### LINKED ISSUE #1652: Add black box tests to composer to cover deleted internal tests
PR #1643 removes fragile tests that were internal to the `composer` crate as they were a maintenance burden. 

The tests that are removed:
1. `full_bundle` - This is covered by the unit tests in the `bundle_factory` module internal to the `composer`. Probably doesn't require adding a new black box test.
2. `bundle_triggered_by_block_timer` - This test assumes quite a bit of knowledge about the internal functions of the composer (specifically that if the bundle isn't filled then it will be submitted after the block timer).
3. `two_seq_actions_single_bundle` - A test similar to the existing single_rollup_tx_payload_integrity that has two transactions should be added to cover this.
4. `chain_id_mismatch_returns_error` - Likely requires bubbling errors up and out from the composer service.

┆Issue Number: ENG-914

ethanoroshiba:
## Summary
Added blackbox tests to cover whitebox tests which were removed in #1643.

## Background
Tests were removed in #1643 because the internal tests were very whitebox-y and fragile. They have not yet been added back as blackbox tests.

## Changes
- Added 3 new blackbox tests to cover `bundle_triggered_by_block_timer`, `two_seq_actions_single_bundle`, and `chain_id_mismatch_returns_error`.
- Removed old internal tests.

## Testing
Passing all tests

## Changelogs
No updates required.

## Related Issues
closes #1652


crates/astria-composer/tests/blackbox/executor.rs
```diff
@@ -0,0 +1,224 @@
...
+            .submit_rollup_transaction(SubmitRollupTransactionRequest {
+                rollup_id: Some(seq0.rollup_id.into_raw()),
+                data: seq0.data.clone(),
+            })
+            .await
+            .expect("rollup transactions should have been submitted successfully to grpc collector")
+    })
+    .await
+    .unwrap();
+    time::timeout(Duration::from_millis(1000), async {
```
ethanoroshiba, line 148: This sometimes times out, even when increasing the timeout time. Unsure as to if this is a problem with the test or something flaky about composer.

joroshiba:
closed in 7 days.

crates/astria-composer/tests/blackbox/executor.rs
```diff
@@ -0,0 +1,224 @@
...
+    let seq0 = RollupDataSubmission {
+        data: data.clone().into(),
+        rollup_id,
+        fee_asset: "nria".parse().unwrap(),
+    };
+
+    // make sure at least one block has passed so that the executor will submit the bundle
+    // despite it not being full
+    time::pause();
+    time::timeout(Duration::from_millis(1000), async {
```
Fraser999, line 51: I think we need to make this a bit more than the block time to guarantee the submission is actioned.  So something like
```rust
let submission_timeout = Duration::from_millis(test_composer.cfg.block_time_ms.saturating_add(100));
```

crates/astria-composer/tests/blackbox/executor.rs
```diff
@@ -0,0 +1,224 @@
...
+            .submit_rollup_transaction(SubmitRollupTransactionRequest {
+                rollup_id: Some(rollup_id.into_raw()),
+                data: data.into(),
+            })
+            .await
+            .expect("rollup transactions should have been submitted successfully to grpc collector")
+    })
+    .await
+    .unwrap();
+    time::advance(Duration::from_millis(2000)).await;
```
Fraser999, line 62:
```suggestion
    time::advance(Duration::from_millis(test_composer.cfg.block_time_ms)).await;
```

crates/astria-composer/tests/blackbox/executor.rs
Fraser999: I think it's the test.  I think the following change is what we're aiming for:
```rust
    let submission_timeout = Duration::from_millis(test_composer.cfg.block_time_ms.saturating_add(100));
    time::timeout(submission_timeout, async {
        composer_client
            .submit_rollup_transaction(SubmitRollupTransactionRequest {
                rollup_id: Some(seq0.rollup_id.into_raw()),
                data: seq0.data.clone(),
            })
            .await
            .expect("rollup transactions should have been submitted successfully to grpc collector");
        composer_client
            .submit_rollup_transaction(SubmitRollupTransactionRequest {
                rollup_id: Some(seq1.rollup_id.into_raw()),
                data: seq1.data.clone(),
            })
            .await
            .expect("rollup transactions should have been submitted successfully to grpc collector");
    })
    .await
    .unwrap();
```

crates/astria-composer/tests/blackbox/executor.rs
```diff
@@ -0,0 +1,224 @@
...
+            .submit_rollup_transaction(SubmitRollupTransactionRequest {
+                rollup_id: Some(seq1.rollup_id.into_raw()),
+                data: seq1.data.clone(),
+            })
+            .await
+            .expect("rollup transactions should have been submitted successfully to grpc collector")
+    })
+    .await
+    .unwrap();
+    time::advance(Duration::from_millis(2000)).await;
```
Fraser999, line 159:
```suggestion
    time::advance(Duration::from_millis(test_composer.cfg.block_time_ms)).await;
```

crates/astria-composer/tests/blackbox/executor.rs
ethanoroshiba: Changed in 0c8a8ba.

crates/astria-composer/tests/blackbox/executor.rs
ethanoroshiba: Thanks for the catch! Made this change in 0c8a8ba.

Fraser999 (approved):

PR #1836, opened 2024-11-25
fix(sequencer): provide unit tests and fixes for vote extensions

Fraser999:
## Summary
Increase test coverage of the `app::vote_extensions` module and fixes issues discovered while implementing these tests.

## Background
This was work spun out of previous PRs to this feature branch to allow parallelizing efforts to complete the feature.

## Changes
- Provided comprehensive unit test coverage of the `app::vote_extensions` module.
- Fixed a handful of issues uncovered.

## Testing
These are tests.

## Changelogs
No updates required - the changelogs will be updated right before the feature merges to `main`.

### DIFF
--- a/crates/astria-sequencer/src/app/vote_extension.rs
+++ b/crates/astria-sequencer/src/app/vote_extension.rs
@@ -53,6 +53,7 @@ use tendermint_proto::google::protobuf::Timestamp;
 use tonic::transport::Channel;
 use tracing::{
     debug,
+    error,
     info,
     instrument,
     warn,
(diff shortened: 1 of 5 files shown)


crates/astria-sequencer/src/app/vote_extension.rs
```diff
@@ -620,46 +626,85 @@ fn aggregate_oracle_votes(
...
+        return price_list
+            .get(midpoint)
+            .copied()
+            .expect("`midpoint` is a valid index");
     }
 
-    prices
+    let Some(lower_index) = midpoint.checked_sub(1) else {
+        // We can only get here if `price_list` is empty; just return 0.
+        return Price::new(0);
```
noot, line 649: the price list should never be empty if constructed correctly in the caller, arguably this should error/panic as silently returning 0 could be bad

Fraser999: I agree as currently called, this line should never get hit.  I'll change this to return an error, but in `apply_prices_from_vote_extensions` where we iterate over all the medians, I'll just skip any medians which errored.  That way, a bug here won't halt block production.

PR #1838, opened 2024-11-26
chore(composer): propagate errors

### LINKED ISSUE #1833: composer: propagate component startup errors
Currently, the Composer `run_until_stopped()` method has two loops which wait on executor and the collectors to be ready before continuing. These checks do a fine job at exiting early if these components fail during startup, but do not propagate the errors which caused the failure. For example, if the chain ID check fails, the following error chain is propagated, not including executor source error chain:

```
executor failed to become ready
executor failed while waiting for it to become ready
channel closed
```

We should find a way to propagate these startup errors such that they will end up in the error chain in the composer `JoinHandle`.

### Relevant Code:
https://github.com/astriaorg/astria/blob/e67fdb949de83365b9a6e6ef0bbbf10482da7f4d/crates/astria-composer/src/composer.rs#L261-L267

┆Issue Number: ENG-1026

ethanoroshiba:
## Summary
Propagate errors which occur while composer is starting up and/or running so that the will be returned by the Composer's handle.

## Background
Previously, composer would only exit with an error if the collectors' or executor's status channels were closed, and then the error message did not provide detailed information about the error that occurred. Additionally, if either of the `wait_for_*` loops failed, the composer would not shut down gracefully. This change is meant to expose the first eyre report which causes the composer to shut down, and gracefully shut down in all circumstances.

## Changes
- Started collector and executor `wait_for_ready` loops concurrently, and continue with graceful shutdown even if these fail.
- Store the first error composer encounters, and return it after graceful shutdown. If waiting for collectors or executor fails, Composer continues so that it can ascertain the underlying error from the task which caused it.

## Testing
Passing all tests

## Changelogs
Changelog updated

## Related Issues
closes #1833


crates/astria-composer/tests/blackbox/executor.rs
```diff
@@ -203,20 +204,13 @@ async fn two_rollup_data_submissions_single_bundle() {
...
-
     let bad_chain_id = "bad_id";
     let test_composer = spawn_composer(&["test1"], Some(bad_chain_id), false).await;
     let err = test_composer.composer.await.unwrap().unwrap_err();
     for cause in err.chain() {
-        if cause
-            .to_string()
-            .contains("executor failed while waiting for it to become ready")
-        {
+        if cause.to_string().contains(&format!(
```
SuperFluffy, line 211: Write the expected error message to a variable and match on that. You can then use that in the panic message further down.

```rust
let expected = format!("expected ...");
if cause.to_string().contains(&expected) {}
```

crates/astria-composer/tests/blackbox/executor.rs
```diff
@@ -203,20 +204,13 @@ async fn two_rollup_data_submissions_single_bundle() {
...
+            "expected chain ID `{TEST_CHAIN_ID}`, but received `{bad_chain_id}`"
+        )) {
             return;
         }
     }
```
SuperFluffy, line 216: For the panic below: I think we should have a formatted panic message that contains the errors we encountered and the error we expected (but didn't see):

```rust
let mut msg = String::new();
writeln!(&mut msg, "did not find expected executor error message");
writeln!(&mut msg, "expected cause:\n\t{expected}");
writeln!(&mut msg, "actual cause chain:\n\t{err:?"});
```

crates/astria-composer/src/composer.rs
```diff
@@ -259,12 +267,19 @@ impl Composer {
         let mut executor_task = tokio::spawn(executor.run_until_stopped());
 
         // wait for collectors and executor to come online
-        wait_for_collectors(&geth_collector_statuses, &mut composer_status_sender)
-            .await
-            .wrap_err("geth collectors failed to become ready")?;
-        wait_for_executor(executor_status, &mut composer_status_sender)
-            .await
-            .wrap_err("executor failed to become ready")?;
+        let composer_status_sender = Arc::new(Mutex::new(composer_status_sender));
```
SuperFluffy, line 270: This seems pretty heavy handed. I am trying to understand why this is necessary to propagate an error up?

ethanoroshiba: I believe this started because I took issue with the fact that these functions would exit early without performing a graceful shutdown if either the collectors or executor failed during initialization. It was awhile ago so I'm not sure what exactly led me to do this, but I think I had figured that while the code was being touched we could spin up these tasks concurrently, hence the need for the `Arc`. I don't believe this is necessary for the error propagation though, as you said.

ethanoroshiba: Actually idk why I did this, cloning seems to work just fine. Made that change here: 2115be7

crates/astria-composer/tests/blackbox/executor.rs
ethanoroshiba: Done in f69e37c

crates/astria-composer/tests/blackbox/executor.rs
ethanoroshiba: Done in f69e37c

Fraser999 (approved):

crates/astria-composer/tests/blackbox/executor.rs
```diff
@@ -1,4 +1,7 @@
-use std::time::Duration;
+use std::{
+    fmt::Write,
```
SuperFluffy, line 2: Unnecessary import of the `Write` trait?

crates/astria-composer/tests/blackbox/executor.rs
```diff
@@ -206,13 +209,21 @@ async fn two_rollup_data_submissions_single_bundle() {
...
     }
-    panic!("did not find expected executor error message")
+    let mut panic_msg = String::new();
+    writeln!(
+        &mut panic_msg,
+        "did not find expected executor error message"
+    )
+    .unwrap();
+    writeln!(&mut panic_msg, "expected cause:\n\t{expected_err_msg}").unwrap();
+    writeln!(&mut panic_msg, "actual cause chain:\n\t{err:?}").unwrap();
```
SuperFluffy, line 227: This `Debug` log might look funky because debug logs are usually multiple lines, so the first line will have a tab before it, but the subsequent lines will not. A nice crate to do that on every line is https://crates.io/crates/indenter.

But I think the error here is the single-line astria-eyre one? Then in this case my comment likely doesn't matter.

crates/astria-composer/tests/blackbox/executor.rs
ethanoroshiba: This is necessary for the `writeln!` macros below to be able to write to a string.

crates/astria-composer/tests/blackbox/executor.rs
ethanoroshiba: Potentially, but if for some reason the error is incorrect it could have a longer source chain. I could just remove the indentation, but I'm not sure I dislike the indenting as it is tbh. It'd look similar to how one would write paragraphs normally:

```
did not find expected executor error message
expected cause:
      some expected cause
actual cause chain:
      {0: asdfasdlskjdflksdjlalskdjflsalsdkjflasjdfalkdsfj, 1: asflkajsldkfjalksjdflakjsdfkasdjflka
sjdflja, 2: askdjfaksjdfkjasdkfhaksjdfhjksadjfh, 3: some underlying error}
```

PR #1840, opened 2024-12-02
feat(sequencer, core, conductor): pass oracle data to rollup via execution API

### LINKED ISSUE #1818: conductor: parse oracle data from `extended_commit_info` and send to rollup if it exists

noot:
## Summary
update conductor to pass oracle data to rollup via execution API.

## Background
this is the final change needed within the astria monorepo for the oracle integration, as the oracle data is now fully passed up the stack. handling the oracle data is now left up to each rollup.

## Changes
- add a new `OracleData` variant to the `RollupData` enum, which contains a `Vec<Price>`, where each `Price` contains the relevant oracle data for a currency pair for that block
- update the injected tx type `ExtendedCommitInfoWithCurrencyPairMapping` such that the id->currency pair mapping also contains the decimals for that currency pair's price; this is necessary information that I overlooked in the previous PR creating that type
- add the extended commit info + proof to `FilteredSequencerBlock`
- update `SequencerBlock`, `FilteredSequencerBlock` and `SubmittedMetadata` constructors such that the extended commit info bytes must be a valid encoded protobuf `ExtendedCommitInfoWithCurrencyPairMapping`
- move functionality for calculating the aggregated prices from the extended commit info from sequencer to core so that conductor can also use it
- update conductor's executor to calculate the aggregated price data and pass it as `RollupData::OracleData` to the rollup via the execution API

## Testing
unit tests

## Related Issues

closes  #1818


crates/astria-conductor/src/executor/mod.rs
```diff
@@ -458,7 +461,8 @@ impl Executor {
     ))]
     async fn execute_firm(&mut self, block: ReconstructedBlock) -> eyre::Result<()> {
         let celestia_height = block.celestia_height;
-        let executable_block = ExecutableBlock::from_reconstructed(block);
+        let executable_block = ExecutableBlock::from_reconstructed(block)
+            .wrap_err("failed to construct executable block")?;
```
SuperFluffy, line 465: This should not fail. Reconstructed blocks should be completely valid blocks. Nothing that reaches the execution task should be invalid.

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -697,23 +701,44 @@ struct ExecutableBlock {
 }
 
 impl ExecutableBlock {
-    fn from_reconstructed(block: ReconstructedBlock) -> Self {
+    fn from_reconstructed(block: ReconstructedBlock) -> eyre::Result<Self> {
```
SuperFluffy, line 704: This should not be fallible.

SuperFluffy (changes requested):

crates/astria-conductor/src/executor/mod.rs
noot: updated to be infallible - no oracle data is added to the txs if it can't be constructed.

crates/astria-conductor/src/executor/mod.rs
noot: same as above

Fraser999 (approved):
One tiny nitpick, otherwise LGTM.

PR #1843, opened 2024-12-03
feat(proto, conductor, cd)!: changes for Forma migration

### LINKED ISSUE #1841: Forma migration
Tasks needed to be completed for Forma migration

┆Issue Number: ENG-1031

ethanoroshiba:
## Summary
Makes the changes necessary in the monorepo to migrate Forma to Mainnet.

## Background
Migrating Forma to mainnet necessitates introducing stop and start logic into the Conductor. This makes the proto and Conductor changes necessary for that. In addition to this, this will allow for future needs to stop and start EVMs on our networks. These changes will also need to be reflected in the `GenesisInfo` sent by rollups to the conductor.

The shutdown and restart logic for the Conductor should be as follows:
- **Firm Only Mode**: Once the stop height is reached for the firm block stream, the firm block at this height is executed (and commitment state updated) before restarting the internal conductor, prompting the rollup for a new `GenesisInfo` with new start/stop heights (and potentially chain IDs).
- **Soft and Firm Mode**: Once the stop height is reached for the soft block stream, no more soft blocks are executed (including the one at the stop height). The firm block is still executed for this height, and then Conductor restarts.
- **Soft Only Mode**: Once the stop height is reached for the soft block stream, no more soft blocks are executed (including the one at the stop height). Conductor is then restarted.

```yaml
config:
        AstriaRollupName:
        AstriaForks: []
                # - < fork name >:
                #         height: ""
                #         halt: ""
                #         feeCollector: ""
                #         EIP1559Params: {}
                #         sequencer:
                #                 chainId: ""
                #                 stopHeight: ""
                #         celestia:
                #                 chainId: ""
                #                 heightVariance: ""
                #         bridgeAddresses: []
```

## Changes
- Changed `sequencer_genesis_block_height` to `sequencer_start_block_height` in `GenesisInfo` (in the execution API).
- Added `sequencer_stop_block_height`, `rollup_start_block_height`, `sequencer_chain_id`, and `celestia_chain_id`, and `halt_at_stop_height` to `GenesisInfo`.
- Changed sequencer to rollup height calculation to start from `rollup_start_block_height` in `GenesisInfo`, so that Conductor doesn't start from rollup height 1 again after restarting.
- Removed sequencer and celestia chain ID environment vars from Conductor, and changed chain ID checks to rely on obtaining these from the initialized executor state.
- Added restart and halt logic for when the stop height is reached, where Conductor will shut down at the stop height if `halt_at_stop_height` is `true`, and restart otherwise.

## Testing
- Added restart logic to `check_for_restart_ok` unit test.
- Added blackbox tests to ensure conductor correctly restarts in each mode, and continues to execute at the correct height after fetching updated genesis info and commitment state.
- Added blackbox test in `SoftAndFirm` mode to ensure the conductor shuts down at the stop height if `GenesisInfo.halt_at_stop_height` is `true`.
- Added smoke test which restarts EVM and Conductor mid-way through the regular smoke test.

## Changelogs
Changelogs updated

## Breaking Changelist
- Proto changes are not technically breaking, since fields were only added, but the old genesis info shape will be rejected since it does not contain the chain IDs to validate against the networks the conductor is connecting to (and would fail even if this was the case, due to `sequencer_stop_block_height` being read from wire format as 0). Thus, rollups will need to provide an updated `GenesisInfo` to conductor (including`astria-geth`).
- Broke domain type `GenesisInfo`

## Related Issues
closes #1841


ethanoroshiba:
Opening this for review despite failing status checks so that review isn't blocking on Geth changes. Once a Geth image can be pulled, we'll be able to run the smoke tests.

proto/executionapis/astria/execution/v1/execution.proto
```diff
@@ -14,9 +14,15 @@ message GenesisInfo {
   // The rollup_id is the unique identifier for the rollup chain.
   astria.primitive.v1.RollupId rollup_id = 1;
   // The first block height of sequencer chain to use for rollup transactions.
-  uint32 sequencer_genesis_block_height = 2;
+  uint32 sequencer_start_block_height = 2;
```
SuperFluffy, line 17: Does this warrant a breaking change?

SuperFluffy:
Note on the PR: the document https://github.com/astriaorg/astria/pull/1843 is private. Please put all relevant information into this PR without referring to an outside, private document.

proto/executionapis/astria/execution/v1/execution.proto
ethanoroshiba: I don't believe this would be a breaking change (even though the proto breaking changes lint would disagree), since the field is identified by the number itself as opposed to the name. The name change is just to more accurately reflect the new functionality, since conductor can now start from anywhere, not just genesis. I do think the proto changes at large are effectively breaking though, since the old shape will no longer work with with conductor (`sequencer_stop _height` should default to 0 in that case, in which case it will never execute any blocks).

SuperFluffy: You are right that the name is indeed a better reflection of how this field is used.

However, I would in general be more conservative when changing APIs: a breaking change (in this case, in the JSON serialization) weighs heavier than a name that's no longer a completely honest representation of the functionality.

crates/astria-conductor/src/celestia/mod.rs
```diff
@@ -178,45 +172,45 @@ impl Reader {
...
         let get_and_validate_sequencer_chain_id = async {
             let actual_sequencer_chain_id =
                 get_sequencer_chain_id(self.sequencer_cometbft_client.clone())
                     .await
                     .wrap_err("failed to get sequencer chain ID")?;
-            let expected_sequencer_chain_id = &self.expected_sequencer_chain_id;
+            let expected_sequencer_chain_id = executor.sequencer_chain_id();
             ensure!(
-                self.expected_sequencer_chain_id == actual_sequencer_chain_id.to_string(),
+                expected_sequencer_chain_id == actual_sequencer_chain_id.to_string(),
```
SuperFluffy, line 201: Don't allocate here. It must have skipped my original review, but the check for the correct celestia chain ID uses `as_str()`, and so should this.

crates/astria-conductor/src/celestia/mod.rs
```diff
@@ -178,45 +172,45 @@ impl Reader {
     async fn initialize(
         &mut self,
     ) -> eyre::Result<((), executor::Handle<StateIsInit>, tendermint::chain::Id)> {
```
SuperFluffy, line 174: I just noticed this function signature. This needs to be refactored in a followup. There is no reason to return an empty tuple out of all things.

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -67,6 +67,14 @@ pub(crate) struct StateNotInit;
 #[derive(Clone, Debug)]
 pub(crate) struct StateIsInit;
 
+#[derive(Debug, thiserror::Error)]
+pub(crate) enum StopHeightExceded {
```
SuperFluffy, line 71: This shouldn't be used in conductor nor sequencer but remain private to this module.

Also, I don't think this should be an error as this is part of normal execution. As it currently stands, this will be emitted using an error event, which is not good.

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -67,6 +67,14 @@ pub(crate) struct StateNotInit;
...
+    #[error("firm height exceeded sequencer stop height")]
+    Celestia,
```
SuperFluffy, line 73: Provide the actional height that's observed and expected

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -67,6 +67,14 @@ pub(crate) struct StateNotInit;
...
+    #[error("soft height met or exceeded sequencer stop height")]
+    Sequencer,
```
SuperFluffy, line 75: Same as above: provide more information.

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -396,6 +412,35 @@ impl Executor {
         // TODO(https://github.com/astriaorg/astria/issues/624): add retry logic before failing hard.
         let executable_block = ExecutableBlock::from_sequencer(block, self.state.rollup_id());
 
+        // Stop executing soft blocks at the sequencer stop block height (exclusive). If we are also
+        // executing firm blocks, we let execution continue since one more firm block will be
+        // executed before `execute_firm` initiates a restart. If we are in soft-only mode, we
+        // return a `StopHeightExceded::Sequencer` error to signal a restart.
+        if self.state.sequencer_stop_block_height().value() > 0
```
SuperFluffy, line 419: Put this into a separate method call, like `is_stop_height_exceeded`.

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -396,6 +412,35 @@ impl Executor {
...
+            && executable_block.height >= self.state.sequencer_stop_block_height()
+        {
+            let res = if self.mode.is_with_firm() {
+                info!(
+                    height = %executable_block.height,
+                    "received soft block whose height is greater than or equal to stop block height in {} mode. \
+                    dropping soft block and waiting for next firm block before attempting restart",
```
SuperFluffy, line 426: Better state that you will inform all soft block from now on.

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -396,6 +412,35 @@ impl Executor {
...
+        // executing firm blocks, we let execution continue since one more firm block will be
+        // executed before `execute_firm` initiates a restart. If we are in soft-only mode, we
+        // return a `StopHeightExceded::Sequencer` error to signal a restart.
+        if self.state.sequencer_stop_block_height().value() > 0
+            && executable_block.height >= self.state.sequencer_stop_block_height()
+        {
+            let res = if self.mode.is_with_firm() {
+                info!(
+                    height = %executable_block.height,
+                    "received soft block whose height is greater than or equal to stop block height in {} mode. \
```
SuperFluffy, line 425: I don't think it's necessary to talk about the mode here.

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -396,6 +412,35 @@ impl Executor {
...
+                    dropping soft block and waiting for next firm block before attempting restart",
+                    self.mode,
+                );
+                Ok(())
+            } else {
+                info!(
+                    height = %executable_block.height,
```
SuperFluffy, line 432: The info log together with the error is not necessary and provides redundant information.

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -396,6 +412,35 @@ impl Executor {
...
+                    "received soft block whose height is greater than or equal to stop block \
+                    height in soft only mode. shutting down and attempting restart",
+                );
+                Err(StopHeightExceded::Sequencer).wrap_err(
+                    "soft height met or exceeded sequencer stop height, attempting restart with \
+                     new genesis info",
+                )
+            };
+            return res;
```
SuperFluffy, line 441: Don't bind `res` and explicitly return. Just `return Ok` and `return Err` directly.

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -396,6 +412,35 @@ impl Executor {
         // TODO(https://github.com/astriaorg/astria/issues/624): add retry logic before failing hard.
```
SuperFluffy, line 412: Comment for above: the error cases are not actual errors but part of normal execution. As such, I would change the return type of `execute_soft` to encode if a restart should happen due to the max being reached.

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -457,19 +505,36 @@ impl Executor {
         err,
     ))]
     async fn execute_firm(&mut self, block: ReconstructedBlock) -> eyre::Result<()> {
```
SuperFluffy, line 507: The same arguments as for `execute_soft` also apply to `execute_firm`. Please see my comments there,

crates/astria-core/src/execution/v1/mod.rs
```diff
@@ -49,9 +49,17 @@ pub struct GenesisInfo {
     /// The rollup id which is used to identify the rollup txs.
     rollup_id: RollupId,
     /// The Sequencer block height which contains the first block of the rollup.
-    sequencer_genesis_block_height: tendermint::block::Height,
+    sequencer_start_block_height: tendermint::block::Height,
+    /// The Sequencer block height to stop at.
+    sequencer_stop_block_height: tendermint::block::Height,
```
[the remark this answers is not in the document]
ethanoroshiba: Hmm, I think it would still just restart in this case? Josh introduced some logic such that if it receives a stop height of 0, it acts as if there is no stop block height. But in the case where stop block height is not 0 and is less than start block height, the conductor will go to execute and then just restart again, so the functionality should already be tested, but I can add a specific test for this scenario just to be sure if you'd like. In either case, if the stop height is ever below the start, I think something is wrong on the rollup end

crates/astria-conductor/src/celestia/mod.rs
ethanoroshiba: Made comment with followup issue linked.

crates/astria-core/src/execution/v1/mod.rs
```diff
@@ -49,9 +49,17 @@ pub struct GenesisInfo {
...
     /// The allowed variance in the block height of celestia when looking for sequencer blocks.
     celestia_block_variance: u64,
+    /// The rollup block number to map to the sequencer start block height.
+    rollup_start_block_height: u64,
+    /// The chain ID of the sequencer network.
+    sequencer_chain_id: String,
```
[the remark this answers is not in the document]
SuperFluffy: Reopening this comment: in how far was this addressed?

crates/astria-core/src/execution/v1/mod.rs
```diff
@@ -49,9 +49,17 @@ pub struct GenesisInfo {
...
+    /// The chain ID of the celestia network.
+    celestia_chain_id: String,
```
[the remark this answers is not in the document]
SuperFluffy: As above, in how far was this addressed, i.e. what decision was taken?

crates/astria-core/src/execution/v1/mod.rs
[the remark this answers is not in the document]
ethanoroshiba: Addressed here: 2f7ed73. Domain type `GenesisInfo` now takes `tendermint` and `celestia_tendermint` chain IDs, which are validated upon converting from wire format. In conductor state, though, they are stored as strings per our offline discussion.

crates/astria-core/src/execution/v1/mod.rs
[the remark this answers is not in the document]
ethanoroshiba: See above

crates/astria-conductor/src/executor/mod.rs
SuperFluffy: This should be done in this PR. The current restart logic is with respect to an error encountered from an RPC. This new restart logic is due to regular operation.

I don't think a cancellation token is necessary here. I think it's sufficient to have the executor task actively close its channel to the celestia/sequencer tasks. Then you get those 2 tasks to shut down for free while maintaining a consistent view. Only need to remember to not kill the entire conductor if one of those goes down

proto/executionapis/astria/execution/v1/execution.proto
SuperFluffy, line 1: Should we add a field here to indicate to conductor whether or not it should stop or restart on having reached the stop height?

ethanoroshiba: I don't think it's strictly necessary, but I wonder if it may be better UX. I'll confer with @mycodecrafting to see what he thinks

crates/astria-conductor/src/executor/mod.rs
ethanoroshiba: Apologies for missing that, could've sworn I removed it. Removed here: 00b10a4

crates/astria-conductor/src/executor/mod.rs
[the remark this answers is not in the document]
ethanoroshiba: Attempted to do so via dropping in 00b10a4. Let me know if this works!

crates/astria-conductor/src/executor/mod.rs
ethanoroshiba: Took a swing at this in 00b10a4. I struggled to find a very elegant way of accomplishing the enum propagation, so let me know if this approach works for you

proto/executionapis/astria/execution/v1/execution.proto
ethanoroshiba: Introduced this field in 680d984 and implemented the logic for it in 02d04d7

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -448,28 +574,42 @@ impl Executor {
...
         block.hash = %telemetry::display::base64(&block.block_hash),
         block.height = block.sequencer_height().value(),
         err,
     ))]
-    async fn execute_firm(&mut self, block: ReconstructedBlock) -> eyre::Result<()> {
+    async fn execute_firm(
+        &mut self,
+        block: ReconstructedBlock,
+    ) -> eyre::Result<Option<StopHeightExceded>> {
+        if self.is_firm_block_height_exceded(&block) {
```
ethanoroshiba, line 589: Should we also close the firm block channel here, similar to how we close the soft block channel?

ethanoroshiba:
Closing in favor of #1928.

PR #1850, opened 2024-12-04
fix(sequencer): remove unnecessary unwrap

### LINKED ISSUE #1848: sequencer: remove unnecessary unwrap
We should be erroring here and not unwrapping: https://github.com/astriaorg/astria/blob/34fcac61e5aaf72153e00cafb64d9f9fc53299b6/crates/astria-sequencer/src/fees/mod.rs#L339

┆Issue Number: ENG-1035

ethanoroshiba:
## Summary
Removed unnecessary unwrap from fee handling.

## Background
Previously, if transaction context was missing for some reason, sequencer would panic. This is not a service stopping error and we should be handling it gracefully.

## Changes
- Removed unnecessary unwrap from fee handling.

## Testing
Passing all tests, no new ones needed.

## Changelogs
No updates required

## Related Issues
closes #1848

### DIFF
--- a/crates/astria-sequencer/src/fees/mod.rs
+++ b/crates/astria-sequencer/src/fees/mod.rs
@@ -336,7 +336,7 @@ async fn check_and_pay_fees<S: StateWrite, T: FeeHandler + Protobuf>(
     let total_fees = base.saturating_add(act.variable_component().saturating_mul(multiplier));
     let transaction_context = state
         .get_transaction_context()
-        .expect("transaction source must be present in state when executing an action");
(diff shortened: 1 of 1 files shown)


SuperFluffy (commented):
Thanks!

It's probably reasonable to do this for all context fetches. This would effectively halt block production while keeping sequencer running - but that's ok?

SuperFluffy (approved):

ethanoroshiba:
Closing for consistency. There are many more examples where `get_transaction_context()` is expected, and they should all be cleaned up in accordance with this comment:
https://github.com/astriaorg/astria/pull/1850#pullrequestreview-2480956344

PR #1856, opened 2024-12-06
fix(sequencer): bump penumbra dep to commit with ibc height fix

noot:
## Summary
bump penumbra dep to commit with ibc height fix as it was merged upstream: https://github.com/penumbra-zone/penumbra/commit/ac7abacc9bb09503d6fd6a396bc0b6850079084e

## Background
see #1691

## Changes
- bump penumbra dep to commit with ibc height fix 

## Testing
it has been tested on the current live networks already

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1872,8 +1872,8 @@ dependencies = [
 
 [[package]]
 name = "cnidarium"
-version = "0.80.7"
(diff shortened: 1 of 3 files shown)


SuperFluffy (approved):
Tracking issue to put this back onto their next release?

PR #1863, opened 2024-12-06
chore(sequencer-relayer): change blob submitter to use boxed blocks

### LINKED ISSUE #1860: chore(sequencer-relayer): avoid large err variant
Large error variants should be avoided as described here: https://rust-lang.github.io/rust-clippy/master/index.html#result_large_err

Problem point:
https://github.com/astriaorg/astria/blob/7e00378748f6e1149e91a8caed93a0e965b0b9f9/crates/astria-sequencer-relayer/src/relayer/write/mod.rs#L99

┆Issue Number: ENG-1041

ethanoroshiba:
## Summary
Changed block channel to use boxed sequencer blocks.

## Background
Running clippy with Rust 1.83.0 in #1857 triggered a lint for large error variant in the send methods for `BlobSubmitterHandle`. Large enum variants (including in Results) should be avoided because they are only as small as their largest variant: https://rust-lang.github.io/rust-clippy/master/index.html#result_large_err. Creating the channel with a boxed block tackles this problem at its source.

## Changes
- Boxed `SequencerBlock` in `BlobSubmitterHandle`, which tackles the problem of potentially large send errors at its source.

## Testing
Passing all tests

## Changelogs
No updates required.

## Breaking Changes
Overridden code freeze since this is a very small, non breaking change that shouldn't have any bearing since our previous audit.

## Related Issues
closes #1860


crates/astria-sequencer-relayer/src/relayer/write/mod.rs
```diff
@@ -96,8 +96,8 @@ impl BlobSubmitterHandle {
     pub(super) fn try_send(
         &self,
         block: SequencerBlock,
```
SuperFluffy, line 98: I mentioned it for `conductor` as well: IMO you could just make the channel be over `Box<SequencerBlock>`, which is a reasonable thing to do (so that the channel doesn't hold very large structures itself).

You would then not need to box every error variant individually.

ethanoroshiba: Done in d113119 :)

crates/astria-sequencer-relayer/src/relayer/mod.rs
```diff
@@ -276,14 +276,19 @@ impl Relayer {
     }
 
     #[instrument(skip_all, fields(%height))]
+    #[expect(
```
SuperFluffy, line 279: I mean, you could just give this an alias, `type ForwardingFut = Fuse<BoxFuture<...>>`

Fraser999 (approved):

PR #1865, opened 2024-12-06
chore(conductor): send boxed objects over channels

### LINKED ISSUE #1858: chore(conductor): avoid large enum variants
We should void having large enum variants as described here: https://rust-lang.github.io/rust-clippy/master/index.html#large_enum_variant

Problem point:
https://github.com/astriaorg/astria/blob/7e00378748f6e1149e91a8caed93a0e965b0b9f9/crates/astria-conductor/src/executor/mod.rs#L70-L90

┆Issue Number: ENG-1039

ethanoroshiba:
## Summary
Changed firm block channel to use boxed `ReconstructedBlock`.

## Background
A large error variant lint was triggered when running clippy with Rust 1.83.0 in #1857 due to send errors returning the whole block. Large enum variants should be avoided because enums are only as small as their largest variant: https://rust-lang.github.io/rust-clippy/master/index.html#large_enum_variant. Changing the channel to consist of a boxed block instead solves this problem at its source.

## Changes
- Changed firm block channel to use boxed `ReconstructedBlock` instead of it being unboxed.

## Testing
Passing all tests.

## Changelogs
No updates needed.

## Breaking Changes
Overridden code freeze since this is a very small, non breaking change that shouldn't have any bearing since our previous audit.

## Related Issues
closes #1858

### DIFF
--- a/crates/astria-conductor/src/celestia/mod.rs
+++ b/crates/astria-conductor/src/celestia/mod.rs
@@ -515,18 +515,20 @@ impl RunningReader {
         match self.executor.try_send_firm_block(block) {
             Ok(()) => self.advance_reference_celestia_height(celestia_height),
             Err(FirmTrySendError::Channel {
-                source: mpsc::error::TrySendError::Full(block),
-            }) => {
-                trace!(
-                    "executor channel is full; rescheduling block fetch until the channel opens up"
-                );
-                self.enqueued_block = enqueue_block(self.executor.clone(), block).boxed().fuse();
-            }
-
-            Err(FirmTrySendError::Channel {
-                source: mpsc::error::TrySendError::Closed(_),
-            }) => bail!("exiting because executor channel is closed"),
-
+                source,
+            }) => match *source {
+                mpsc::error::TrySendError::Full(block) => {
+                    trace!(
+                        "executor channel is full; rescheduling block fetch until the channel \
+                         opens up"
+                    );
+                    self.enqueued_block =
+                        enqueue_block(self.executor.clone(), block).boxed().fuse();
+                }
+                mpsc::error::TrySendError::Closed(_) => {
+                    bail!("exiting because executor channel is closed");
+                }
+            },
             Err(FirmTrySendError::NotSet) => bail!(
                 "exiting because executor was configured without firm commitments; this Celestia \
                  reader should have never been started"
--- a/crates/astria-conductor/src/celestia/mod.rs
+++ b/crates/astria-conductor/src/celestia/mod.rs
@@ -512,18 +512,18 @@ impl RunningReader {
     #[instrument(skip_all)]
     fn forward_block_to_executor(&mut self, block: ReconstructedBlock) -> eyre::Result<()> {
         let celestia_height = block.celestia_height;
-        match self.executor.try_send_firm_block(block) {
+        match self.executor.try_send_firm_block(Box::new(block)) {
             Ok(()) => self.advance_reference_celestia_height(celestia_height),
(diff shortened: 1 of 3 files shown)


SuperFluffy:
Interesting that this showed up now - I remember clippy flagging this at some point, and I was almost certain that I did add an exception attribute, but maybe I am hallucinating.

Instead of boxing the problematic error variant, can you change the mpsc channel to be `Box<T>` instead? You'd then fix the error variants "for free" so to say.

SuperFluffy:
Neither PR text nor PR title match what the changes.

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -146,20 +146,18 @@ impl Handle<StateIsInit> {
     #[instrument(skip_all, err)]
     pub(crate) async fn send_firm_block(
         self,
-        block: ReconstructedBlock,
+        block: Box<ReconstructedBlock>,
```
SuperFluffy, line 149: If we change the function signature you no longer have to explicilty box/unbox in various places:

```suggestion
        block: impl Into<Box<ReconstructedBlock>>,
```

this makes use of the blanket impl `From<T> for Box<T>` and `From<Box<T>> for Box<T>`.

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -146,20 +146,18 @@ impl Handle<StateIsInit> {
...
         let sender = self.firm_blocks.as_ref().ok_or(FirmSendError::NotSet)?;
-        sender.send(block).await?;
-        Ok(())
+        Ok(sender.send(block).await?)
     }
 
     pub(crate) fn try_send_firm_block(
         &self,
-        block: ReconstructedBlock,
+        block: Box<ReconstructedBlock>,
```
SuperFluffy, line 157: Similar to `fn try_send_firm_block`:

```suggestion
        block: impl Into<Box<ReconstructedBlock>>,
```

crates/astria-conductor/src/celestia/mod.rs
```diff
@@ -512,21 +512,23 @@ impl RunningReader {
...
-
+                source,
+            }) => match source {
+                mpsc::error::TrySendError::Full(block) => {
+                    trace!(
+                        "executor channel is full; rescheduling block fetch until the channel \
+                         opens up"
+                    );
+                    self.enqueued_block =
+                        enqueue_block(self.executor.clone(), *block).boxed().fuse();
```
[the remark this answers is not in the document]
ethanoroshiba: Adjusted `enqueue_block` function signature to avoid using `*` in d2338a04.

crates/astria-conductor/src/celestia/mod.rs
```diff
@@ -665,7 +667,7 @@ async fn enqueue_block(
     block: ReconstructedBlock,
 ) -> Result<u64, FirmSendError> {
     let celestia_height = block.celestia_height;
-    executor.send_firm_block(block).await?;
+    executor.send_firm_block(Box::new(block)).await?;
```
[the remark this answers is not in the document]
ethanoroshiba: Done in d2338a04.

crates/astria-conductor/src/executor/mod.rs
ethanoroshiba: Done in d2338a04.

crates/astria-conductor/src/executor/mod.rs
ethanoroshiba: Thanks for these catches and suggestions! Made the requested changes in d2338a04.

PR #1871, opened 2024-12-11
chore(ci): update rust lint workflows and `just` recipes

ethanoroshiba:
## Summary
Moved rust lints to just recipes, added all to lint rust recipe.

## Background
Discrepancies between the just recipe and GH workflow made it so that the lint command might succeed but then fail when a PR is opened. This is meant to make it easier to match the GH workflow locally and test to ensure this status check will pass.

## Changes
- Moved all rust lints to just recipes.
- Changed `_lint-rust` to call all of the above recipes.

## Testing
Manually tested

## Changelogs
No updates required

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -73,7 +73,7 @@ jobs:
           toolchain: nightly-2024-09-15
           components: rustfmt
       - name: run rustfmt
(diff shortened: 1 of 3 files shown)


joroshiba (approved):
Great keeping this in sync with single source 👍

PR #1877, opened 2024-12-13
feat(sequencer, charts)!: support uds for abci

joroshiba:
## Summary
Add support to the sequencer to support running the ABCI connection over UDS instead of TCP only, updates to chart to use the UDS connection by default.

## Background
Local testing found up to a 25x speed up using UDS instead of TCP loopback for ABCI communication. We can continue to offer the option of using TCP connections, and support the faster connection within our charts by default.

## Changes
- mild refactor of sequencer startup to move abci server start to it's own function, mirroring grpc server setup
- replace old `listen_addr` config with new `abci_listener_url` config where protocol is used to inform sequencer whether to use tcp or unix

## Testing
Synced a full node with mainnet, and tested with ABCI replay.

## Changelogs
Changelogs updated.

## Breaking Changelist
- Removed `ASTRIA_SEQUENCER_LISTEN_ADDR` config variable, replaced with `ASTRIA_SEQUENCER_ABCI_LISTENER_URL`

## Related Issues
closes https://github.com/astriaorg/astria/issues/1891


charts/sequencer/templates/statefulsets.yaml
```diff
@@ -16,6 +16,11 @@ spec:
       labels:
         app: {{ .Values.moniker }}-sequencer
     spec:
+      securityContext:
+        runAsUser: 1000
+        runAsGroup: 3000
+        fsGroup: 2000
+        supplementalGroups: [4000]
```
joroshiba, line 23: @WafflesVonMaple feel a bit out of my depth on these settings. I'm also certain there is more configuration going on here than necessary, but these setting work.

WafflesVonMaple: runAsUser: # hardcode user to non-root if not set in dockerfile
runAsGroup: # hardcode group to non root if not set in Dockerfile
runAsNonRoot: # hardcode to non-root. Redundant if dockerfile is set to USER 1000

General safety dictates runAsUser/Group are not explicit replacement for runAsNonRoot as the ID must exist in the container for it to run. 

runAsNonRoot also limits access to host resources that might mistakenly get exposed to the container 

fsGroup/GroupChangePolicy defines group which kube will change perm of all files in volumes mounted. Caution to be exercised as changing group ownership of an entire volume can cause startup delay or other issues

crates/astria-sequencer/src/sequencer.rs
```diff
@@ -210,6 +188,83 @@ fn start_grpc_server(
...
+    // Builds the server but does not start listening.
+    let server = Server::builder()
+        .consensus(consensus_service)
+        .info(info_service)
+        .mempool(mempool_service)
+        .snapshot(snapshot_service)
+        .finish()
+        .ok_or_eyre("server builder didn't return server; are all fields set?")?;
+
+    // Validate and parse the listen_url received from the config.
```
SuperFluffy, line 220: This (as well as the subsequent server start) could be done more elegantly by providing a type `AbciListenUrl` with variants `Tcp` and `Unix` and implement `FromStr` for it. Then we could also have a few simple unit tests.

crates/astria-sequencer/src/sequencer.rs
```diff
@@ -210,6 +188,83 @@ fn start_grpc_server(
...
+    let abci_url = Url::parse(listen_url).wrap_err("failed to parse listen_addr")?;
+    let validated_listen_addr = match abci_url.scheme() {
+        "unix" => match abci_url.to_file_path() {
+            Ok(_) => Ok(abci_url.path().to_string()),
+            Err(e) => Err(eyre!(
+                "failed parsing unix listen_addr file path, error: {:?}",
```
SuperFluffy, line 226: I figure `eyre::WrapErr::wrap_err` was not doable because this is a boxed error? We have solved a related issue in `astria_eyre` by wrapping the boxed error. You can then construct a proper chain:

```rust
eyre!(e).wrap_err("failed rasing unix listen_addr file path")
```

SuperFluffy (commented):
From a high level this looks fine, but I would like to see a slightly more elegant approach that is unit testable by providing an enum for the two supported server URL kinds.

SuperFluffy (approved):
I have added an enum `AbciListenUrl` with variants `Tcp` and `Uds`, together with a `FromStr` impl. I have also moved the parse to the config because that seemed to be the appropriate place for it (sequencer always runs by binding to a socket; lazy eval at a later point does not seem to make sense in this case).

Final change was from `ASTRIA_SEQUENCER_ABCI_LISTENER_URL` to `ASTRIA_SEQUENCER_ABCI_LISTEN_URL` because we are listening on that socket, not expecting a listener (CometBFT connects to the app, not the other way round).

Fraser999 (approved):

WafflesVonMaple (approved):

SuperFluffy (approved):

Issue #1446, opened 2024-09-04
sequencer: implement oracle-related actions


noot:
follow up to https://github.com/astriaorg/astria/pull/1236.

slinky contains `Msg` types which should be implemented as actions within the sequencer:

-   https://github.com/skip-mev/slinky/blob/158cde8a4b774ac4eec5c6d1a2c16de6a8c6abb5/proto/slinky/oracle/v1/tx.proto
-   https://github.com/skip-mev/slinky/blob/158cde8a4b774ac4eec5c6d1a2c16de6a8c6abb5/proto/slinky/marketmap/v1/tx.proto

┆Issue Number: ENG-786

joroshiba:
This issue is stale because it has been open 45 days with no activity. Remove stale label or this issue
be closed in 7 days.

noot:
closed by #1878 and #1882

PR #1878, opened 2024-12-13
feat(sequencer): implement oracle module actions 

noot:
## Summary
 implement oracle module actions `AddCurrencyPair` and `RemoveCurrencyPair`

see https://github.com/skip-mev/connect/blob/main/proto/connect/oracle/v2/tx.proto and https://github.com/skip-mev/connect/blob/9ea31680774e2f71e683c0a4989df5bf2d5f2302/x/oracle/keeper/msg_server.go#L28 for relevant connect code

## Background

these are required for the oracle authorities to be able to add/remove currency pairs from state; these are the pairs whose prices will be included in VEs.

## Changes
-  implement oracle module actions `AddCurrencyPair` and `RemoveCurrencyPair`
- make `CurrencyPairState.QuotePrice` optional, to reflect that this may not be set upon currency pair initialization (as no oracle price has been received for it)
- remove `get/put_num_removed_currency_pairs` from the oracle state_ext as it's not needed for our logic due to the fact that we execute txs in the proposal phase, so by the time vote extensions are issued for a block, the state has already been updated and currency pairs that are removed in that block have already been removed.

## Testing
unit tests

## Breaking Changelist
- adds new actions to the sequencer
- sequencer genesis is also updated (new fees)

## Related Issues

part of # #1446


crates/astria-sequencer/src/action_handler/impls/add_currency_pairs.rs
```diff
@@ -0,0 +1,103 @@
...
+
+            let currency_pair_state = CurrencyPairState {
+                price: None,
+                nonce: CurrencyPairNonce::new(0),
+                id: next_currency_pair_id,
+            };
+            state
+                .put_currency_pair_state(pair.clone(), currency_pair_state)
+                .wrap_err("failed to put currency pair state")?;
+            num_currency_pairs
```
Fraser999, line 83:
```suggestion
            num_currency_pairs = num_currency_pairs
                .checked_add(1)
                .ok_or_eyre("overflow when incrementing number of currency pairs")?;
```

crates/astria-sequencer/src/action_handler/impls/add_currency_pairs.rs
```diff
@@ -0,0 +1,103 @@
...
+            .put_next_currency_pair_id(CurrencyPairId::new(num_currency_pairs))
+            .wrap_err("failed to put next currency pair id")?;
+        state
+            .put_num_currency_pairs(
+                num_currency_pairs.saturating_add(
+                    self.pairs
+                        .len()
+                        .try_into()
+                        .expect("number of pairs cannot exceed u64::MAX"),
+                ),
```
Fraser999, line 98: Seems like we should just use the value of `num_currency_pairs` here, assuming we increment it as we add each pair.  That way we accommodate the case where we skipped some of `self.pairs` as duplicated.

crates/astria-sequencer/src/action_handler/impls/add_currency_pairs.rs
```diff
@@ -0,0 +1,103 @@
...
+            state
+                .put_currency_pair_state(pair.clone(), currency_pair_state)
+                .wrap_err("failed to put currency pair state")?;
+            num_currency_pairs
+                .checked_add(1)
+                .ok_or_eyre("overflow when incrementing number of currency pairs")?;
+        }
+
+        state
+            .put_next_currency_pair_id(CurrencyPairId::new(num_currency_pairs))
```
Fraser999, line 89: This would be just `next_currency_pair_id` if we increment it inside the `for` loop.

crates/astria-sequencer/src/action_handler/impls/add_currency_pairs.rs
Fraser999, line 1: I think we should have unit tests for this file, including at least one where we try to add a duplicated pair and check the values of the stored next id and number of pairs.

crates/astria-sequencer/src/action_handler/impls/remove_currency_pairs.rs
```diff
@@ -0,0 +1,70 @@
...
+        );
+
+        for pair in &self.pairs {
+            state
+                .delete_currency_pair(pair)
+                .await
+                .wrap_err("failed to delete currency pair")?;
+            num_currency_pairs
+                .checked_sub(1)
+                .ok_or_eyre("failed to decrement number of currency pairs")?;
```
Fraser999, line 61:
```suggestion
            num_currency_pairs = num_currency_pairs
                .checked_sub(1)
                .ok_or_eyre("failed to decrement number of currency pairs")?;
```

crates/astria-sequencer/src/action_handler/impls/add_currency_pairs.rs
```diff
@@ -0,0 +1,103 @@
...
+            .wrap_err("failed to get number of currency pairs")?;
+
+        for pair in &self.pairs {
+            if state
+                .get_currency_pair_state(pair)
+                .await
+                .wrap_err("failed to get currency pair state")?
+                .is_some()
+            {
+                debug!("currency pair {} already exists, skipping", pair);
```
Fraser999, line 71: This might be worth making `warn!` since it possibly indicates a user error?

crates/astria-sequencer/src/action_handler/impls/remove_currency_pairs.rs
```diff
@@ -0,0 +1,70 @@
...
+        );
+
+        let num_currency_pairs = state
+            .get_num_currency_pairs()
+            .await
+            .wrap_err("failed to get number of currency pairs")?;
+        ensure!(
+            num_currency_pairs >= self.pairs.len() as u64,
+            "cannot remove more currency pairs than exist",
+        );
```
Fraser999, line 52: Not sure if this is worthwhile?  We'd error out in the `for` loop below at the `checked_sub`.  Not a big deal though - maybe the more explicit error message here is better.

crates/astria-sequencer/src/action_handler/impls/remove_currency_pairs.rs
```diff
@@ -0,0 +1,70 @@
...
+
+        for pair in &self.pairs {
+            state
+                .delete_currency_pair(pair)
+                .await
+                .wrap_err("failed to delete currency pair")?;
```
Fraser999, line 58: This silently ignores the case where the pair wasn't previously in storage.  In that case, we shouldn't decrement `num_currency_pairs`.

We should probably change the return type of `oracle::StateWriteExt::delete_currency_pair` to `Result<bool>`, where `true` means the pair existed.

It might be worth renaming that method from `delete_currency_pair` to `remove_currency_pair` too for consistency.

Fraser999, line 58: If we remove a pair, and then later re-add it, it will be assigned a different ID.  Is that acceptable?  (Seems like it should work, and it's certainly easier to implement here, but that behaviour seems a bit surprising.  Saying that, I don't really have a good handle on how visible these IDs are externally.)

crates/astria-sequencer/src/action_handler/impls/remove_currency_pairs.rs
Fraser999, line 1: I think we should have unit tests for this file, including at least one where we try to remove a non-existent pair and check the stored value of number of pairs.

proto/protocolapis/astria/protocol/transaction/v1/action.proto
```diff
@@ -30,6 +31,10 @@ message Action {
     FeeAssetChange fee_asset_change = 53;
     FeeChange fee_change = 55;
     IbcSudoChange ibc_sudo_change = 56;
+
+    // Oracle actions are defined on 61-70
```
Fraser999, line 35: Do you think we should start this range at 71 or even higher to leave plenty of room for other sudo actions in the future?

Fraser999 (changes requested):
There are a couple of blocking issues in the action handlers, but mostly LGTM.

crates/astria-sequencer/src/action_handler/impls/remove_currency_pairs.rs
noot: i like the explicit error message

crates/astria-sequencer/src/action_handler/impls/remove_currency_pairs.rs
noot: yeah, that seems fine. the IDs are mainly so that p2p messages can just contain the IDs rather than the whole currency pair string from what I understand, so they're mainly internal to the sequencer app.

crates/astria-core/src/protocol/transaction/v1/action/group/mod.rs
```diff
@@ -55,6 +57,9 @@ impl_belong_to_group!(
     (FeeAssetChange, Group::BundleableSudo),
     (IbcRelay, Group::BundleableGeneral),
     (IbcSudoChange, Group::UnbundleableSudo),
+    // TODO: should these have a different group?
```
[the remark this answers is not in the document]
noot: I think `BundleableGeneral` is correct, as it's an action that's ok to be bundled and isn't a sudo-action (but it is a permissioned action, just not sent by the sudo address)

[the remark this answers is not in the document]
noot: the other option is to make a new `BundleableMarketMapAdmin` group, but not sure that's really necessary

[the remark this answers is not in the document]
Fraser999: I can't think of any problem in using `BundleableGeneral` either.  Happy to not add a new group if we don't really need to :)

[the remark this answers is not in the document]
Fraser999: I guess this TODO line should be deleted pls.

Fraser999 (commented):
No blockers now - just three outstanding comments/threads left.

crates/astria-sequencer/src/action_handler/impls/add_currency_pairs.rs
noot: i don't think this necessitates a `warn` as the pair will exist after this action is done, thoughts?

proto/protocolapis/astria/protocol/transaction/v1/action.proto
noot: yeah, doesn't hurt, i will change

PR #1882, opened 2024-12-18
feat(sequencer): add market map actions

ethanoroshiba:
## Summary
Added select market map actions [here](https://github.com/skip-mev/connect/blob/main/proto/connect/marketmap/v2/tx.proto) to the sequencer.

## Background
These actions were needed for support for keeping a `MarketMap` in state and making changes to it as part of the oracle feature. They follow Skip's logic and apply it to our state machine.

#### `ChangeMarkets`
Takes a list of markets and either creates, updates, or removes them depending on its variant. Must be signed by an address included in the market map params' `market_authorities`.
- **Create:** Creates the markets in the market map. If no market map is found, one will be created. If any of the markets to create already exist, this action will err.
- **Update:** Updates the markets in the market map, matching based on `Ticker.currency_pair`). If no market map is found, or any market is missing a counterpart in the map, this action will err.
- **Remove:** Removes the markets from the market map. If a market is not found in the map, it will be ignored.

#### `UpdateParams`
Updates the market map `Params`, which contains the market authority addresses as well as an admin address. This will execute whether there are params in the state already or not. Must be signed by the sequencer network authority sudo address.

## Changes
- Added `ChangeMarkets` and `UpdateParams` to `PriceFeed` action in proto, core, and sequencer.

## Testing
- Added fairly exhaustive unit tests to each action's implementation module.
- Added app execution test for each action for more high-level testing.
- Added each action to all action snapshot.

## Breaking Changelist
- Added new actions which break app hash due to adding new information to state.
- Added `ALICE_ADDRESS` to market authorities in testing `genesis`, breaking two more snapshots.


crates/astria-core/src/protocol/transaction/v1/action/group/mod.rs
```diff
@@ -55,6 +61,12 @@ impl_belong_to_group!(
     (FeeAssetChange, Group::BundleableSudo),
     (IbcRelay, Group::BundleableGeneral),
     (IbcSudoChange, Group::UnbundleableSudo),
+    (UpsertMarkets, Group::BundleableGeneral),
+    (CreateMarkets, Group::BundleableGeneral),
+    (UpdateMarkets, Group::BundleableGeneral),
+    (UpdateParams, Group::BundleableSudo),
+    (RemoveMarketAuthorities, Group::BundleableGeneral),
+    (RemoveMarkets, Group::BundleableGeneral),
```
ethanoroshiba, line 69: Are these groups correct?

noot: I think these can all be `BundleableGeneral` for now, as they're not sent by the sudo address

crates/astria-sequencer/src/action_handler/impls/remove_markets.rs
```diff
@@ -0,0 +1,264 @@
...
+            .await
+            .wrap_err("failed to convert signer address to base prefixed address")?;
+        let market_authorities = state
+            .get_params()
+            .await?
+            .ok_or_eyre("market map params not found in state")?
+            .market_authorities;
+        if !market_authorities.contains(&from) {
+            bail!("address {from} is not a market authority");
+        }
```
noot, line 49:
```suggestion
        ensure!(
            market_authorities.contains(&from),
            "address {from} is not a market authority"
        );
```

same for other actions

crates/astria-sequencer/src/test_utils.rs
```diff
@@ -47,3 +57,33 @@ pub(crate) async fn calculate_rollup_data_submission_fee_from_state<
         )
         .expect("fee addition should not overflow")
 }
+
+pub(crate) fn example_ticker(metadata: String) -> Ticker {
```
noot, line 61:
```suggestion
pub(crate) fn example_ticker_with_metadata(metadata: String) -> Ticker {
```

crates/astria-sequencer/src/action_handler/impls/update_params.rs
```diff
@@ -0,0 +1,117 @@
...
+    async fn check_and_execute<S: cnidarium::StateWrite>(&self, mut state: S) -> eyre::Result<()> {
+        let from = &state
+            .get_transaction_context()
+            .expect("transaction source must be present in state when executing an action")
+            .address_bytes();
+        let sudo_address = state
+            .get_sudo_address()
+            .await
+            .wrap_err("failed to get sudo address from state")?;
+        ensure!(sudo_address == *from, "signer is not the sudo key");
```
noot, line 30: i think this should be the market map admin, actually

also, i think the name of this action is too unclear/generic, we should rename it to `UpdateMarketMapParams`

noot (commented):
looks good, only a few minor comments!

crates/astria-sequencer/src/action_handler/impls/update_params.rs
noot: interesting, seems okay to leave this then!

proto/protocolapis/astria/protocol/transaction/v1/action.proto
```diff
@@ -30,6 +32,14 @@ message Action {
     FeeAssetChange fee_asset_change = 53;
     FeeChange fee_change = 55;
     IbcSudoChange ibc_sudo_change = 56;
+
+    // MarketMap actions are defined on 71-80
+    UpsertMarkets upsert_markets = 71;
+    CreateMarkets create_markets = 72;
+    UpdateMarkets update_markets = 73;
```
joroshiba, line 39: I don't think we need an upsert here?

Our other similar commands use a `Change` command with an enum of add or remove vs independent actions for each method.

proto/protocolapis/astria/protocol/transaction/v1/action.proto
```diff
@@ -228,9 +238,51 @@ message FeeChange {
...
+// Removes market map authorities.
+message RemoveMarketAuthorities {
+  // RemoveAddresses is the list of addresses to remove.
+  repeated string remove_addresses = 1;
+}
+
+// Removes markets from the market map.
+message RemoveMarkets {
+  // Markets is the list of markets to remove.
+  repeated string markets = 2;
```
joroshiba, line 287: What is the string here? Create and update both contain a market with a ticker specified? Should we be identifying by ticker?

proto/protocolapis/astria/protocol/transaction/v1/action.proto
```diff
@@ -228,9 +238,51 @@ message FeeChange {
...
+message UpdateMarketMapParams {
+  // Params defines the new parameters for the x/marketmap module.
+  connect.marketmap.v2.Params params = 1;
+}
+
+// Removes market map authorities.
+message RemoveMarketAuthorities {
+  // RemoveAddresses is the list of addresses to remove.
+  repeated string remove_addresses = 1;
+}
```
joroshiba, line 282: Is there a reason this is distinct from updating the params? It looks like the params also contains this list, so instead of remove, could just update to the list you do want?

joroshiba (commented):
Reviewed only the protos so far.

proto/protocolapis/astria/protocol/transaction/v1/action.proto
ethanoroshiba: Identified by ticker in 4bd1f74.

proto/protocolapis/astria/protocol/transaction/v1/action.proto
ethanoroshiba: Consolidated actions and removed upsert in 4bd1f74.

proto/protocolapis/astria/protocol/transaction/v1/action.proto
ethanoroshiba: Removed this action in 4c61ef8

crates/astria-sequencer/src/action_handler/impls/price_feed.rs
```diff
@@ -40,28 +56,19 @@ impl ActionHandler for PriceFeed {
     }
 
     async fn check_and_execute<S: StateWrite>(&self, state: S) -> Result<()> {
-        // TODO: should we use the market map admin here, or a different admin?
```
ethanoroshiba, line 43: Noting that this needed to be moved into the enum variants' `check_and_execute` functions, since `ChangeMarkets` validates against `Params.market_authorities` and `UpdateMarketMapParams` against `sudo`.

crates/astria-sequencer/src/action_handler/impls/price_feed.rs
```diff
@@ -144,11 +155,129 @@ async fn check_and_execute_currency_pairs_removal<S: StateWrite>(
...
+        market_authorities.contains(&from),
+        "address {from} is not a market authority"
+    );
+
+    // create a new market map if one does not already exist
+    let mut market_map = state
+        .get_market_map()
+        .await
+        .wrap_err("failed to get market map")?
+        .unwrap_or(MarketMap {
```
noot, line 187: this can be an error - a market map should always exist in state (even if empty) if this action is enabled

noot (approved):
one minor comment, otherwise looks good!

crates/astria-sequencer/src/action_handler/impls/price_feed.rs
ethanoroshiba: Updated in 24839d3.

PR #1883, opened 2024-12-18
feat(charts): update geth charts for oracle genesis values

noot:
## Summary
update geth charts for oracle genesis values.

## Background
geth charts needed to be updated for the oracle integration; see https://github.com/astriaorg/astria-geth/pull/62

## Changes
- add `astriaOracleCallerAddress` and `astriaOracleContractAddress` config values
- add oracle contract bytecode to genesis alloc at `astriaOracleContractAddress`
- update dev tags for geth and sequencer

## Testing
ran it end-to-end with the usual dev cluster instructions and queried the oracle contract for price updates via the instructions here: https://github.com/astriaorg/astria-oracle-contracts/tree/main?tab=readme-ov-file#query-contract


charts/evm-rollup/values.yaml
```diff
@@ -159,7 +171,8 @@ config:
         condition: .Values.metrics.enabled
       - name: txpool.nolocals
         value: "true"
-
+      - name: verbosity
+        value: "4"  # 3=info, 4=debug
```
quasystaty1, line 175: Thank you!

quasystaty1 (approved):
LGTM, just one idea to move the images used into the value file as this is a temporarily change.

PR #1884, opened 2024-12-19
feat(core)!: use newtype sequencer block hashes

SuperFluffy:
## Summary
Adds a newtype `astria_core::sequencerblock::v1::block::Hash` to represent Sequencer block hash (an array of 32 bytes).

The standard formatting of the newtype's `std::format:Display` implementation uses lower case hex, while its alternative selector uses standard base64. Note that the latter was introduced to provide a way to format the hash using the same conventions as pbjson for the protobuf wire types.

```rust
use astria_core::sequencerblock::v1::block;
let block_hash = block::Hash::new([42; 32]);
assert_eq!(
    "2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a",
    format!("{block_hash}"),
);
assert_eq!(
    "KioqKioqKioqKioqKioqKioqKioqKioqKioqKioqKio=",
    format!("{block_hash:#}"),
};
```

## Background
During work on the Astria Auctioneer it was discovered that representing Sequencer Block Hashes with a distinct type is desirable for two reasons:

1. to distinguish Sequencer block hashes from other objects represented by an array of 32 bytes (`[u8; 32]`) that show up in the Astria ecosystem, for example Rollup block hashes.
2. to have a unified `std::format::Display` implementation for all Astria block hashes using a hexadecimal representation.

## Changes
- Introduce the `astria_core::sequencerblock::v1::block::Hash` newtype for `[u8; 32]` objects representing Astria block hashes, and replace all domain types with it.

## Testing
Added doc tests for the `Hash` constructor and normal and alternative display impl.

## Changelogs
Changelogs updated.

## Breaking Changelist
- This is a breaking change in `astria-core` but not in any of the services (because the return value of public getters was changed).
- This is not network breaking.

### DIFF
--- a/crates/astria-conductor/src/celestia/reporting.rs
+++ b/crates/astria-conductor/src/celestia/reporting.rs
@@ -4,7 +4,6 @@ use serde::ser::{
     SerializeSeq,
     SerializeStruct,
 };
-use telemetry::display::base64;
 
 use super::{
     ReconstructedBlock,
@@ -53,10 +52,23 @@ impl<'a> Serialize for ReportReconstructedBlock<'a> {
         ];
         let mut state = serializer.serialize_struct("ReconstructedBlockInfo", FIELDS.len())?;
         state.serialize_field(FIELDS[0], &self.0.celestia_height)?;
-        // TODO: use the block hash's Display impl for this
-        state.serialize_field(FIELDS[1], &base64(&*self.0.block_hash))?;
+        state.serialize_field(FIELDS[1], &SerializeDisplay(&self.0.block_hash))?;
         state.serialize_field(FIELDS[2], &self.0.transactions.len())?;
         state.serialize_field(FIELDS[3], &self.0.celestia_height)?;
         state.end()
+    }
+}
+
+struct SerializeDisplay<'a, T>(&'a T);
+
+impl<'a, T> Serialize for SerializeDisplay<'a, T>
+where
+    T: std::fmt::Display,
+{
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        serializer.collect_str(self.0)
     }
 }
--- a/crates/astria-conductor/src/celestia/reporting.rs
+++ b/crates/astria-conductor/src/celestia/reporting.rs
@@ -53,7 +53,8 @@ impl<'a> Serialize for ReportReconstructedBlock<'a> {
         ];
         let mut state = serializer.serialize_struct("ReconstructedBlockInfo", FIELDS.len())?;
         state.serialize_field(FIELDS[0], &self.0.celestia_height)?;
-        state.serialize_field(FIELDS[1], &base64(&self.0.block_hash))?;
+        // TODO: use the block hash's Display impl for this
+        state.serialize_field(FIELDS[1], &base64(&*self.0.block_hash))?;
         state.serialize_field(FIELDS[2], &self.0.transactions.len())?;
         state.serialize_field(FIELDS[3], &self.0.celestia_height)?;
         state.end()
--- a/crates/astria-core/src/sequencerblock/v1/block.rs
+++ b/crates/astria-core/src/sequencerblock/v1/block.rs
@@ -9,8 +9,8 @@ use bytes::Bytes;
 use indexmap::IndexMap;
 use sha2::Sha256;
 use tendermint::{
-    account,
     Time,
+    account,
 };
 
 use super::{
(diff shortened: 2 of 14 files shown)


crates/astria-core/src/sequencerblock/v1/block.rs
```diff
@@ -580,13 +582,107 @@ enum SequencerBlockHeaderErrorKind {
...
+///
+/// There are two main purposes of this type:
+///
+/// 1. avoid confusion with other hashes of the form `[u8; 32]` common in Astria, like rollup
+///    (ethereum) 32 byte hashes.
+/// 2. to provide a hex formatted display impl, which is the convention for block hashes.
+///
+/// Note that hex based `Display` impl of [`Hash`] does not follow the pbjson
+/// convention to display protobuf `bytes` using base64 encoding. To get the
+/// display formatting faithful to pbjson convetion use the alternative formatting selector,
```
ethanoroshiba, line 604:
```suggestion
/// display formatting faithful to pbjson convention use the alternative formatting selector,
```

crates/astria-conductor/src/celestia/reporting.rs
```diff
@@ -53,7 +53,8 @@ impl<'a> Serialize for ReportReconstructedBlock<'a> {
         ];
         let mut state = serializer.serialize_struct("ReconstructedBlockInfo", FIELDS.len())?;
         state.serialize_field(FIELDS[0], &self.0.celestia_height)?;
-        state.serialize_field(FIELDS[1], &base64(&self.0.block_hash))?;
+        // TODO: use the block hash's Display impl for this
+        state.serialize_field(FIELDS[1], &base64(&*self.0.block_hash))?;
```
ethanoroshiba, line 57: Is there a reason this can't be done now?
```suggestion
        state.serialize_field(FIELDS[1], &self.0.block_hash.to_string())?;
```

ethanoroshiba (approved):
Couple nits but otherwise LGTM. I don't foresee these changes being breaking since the storage types remain the same, but it may be worthwhile to double check this by starting a sync with `main`, stopping midway and building with this PR, then continuing with the sync.

crates/astria-conductor/src/celestia/reporting.rs
SuperFluffy: Ha, no, thanks for pointing that out!

crates/astria-core/src/sequencerblock/v1/block.rs
SuperFluffy: thanks for spotting this.  01fdbba2

crates/astria-conductor/src/celestia/reporting.rs
SuperFluffy: Addressed by  01fdbba2 using an ad-hoc `SerializeDisplay` newtype so that we can use [`Serializer::collect_str`](https://docs.rs/serde/latest/serde/ser/trait.Serializer.html#method.collect_str) and avoid the allocation.

crates/astria-core/src/sequencerblock/v1/block.rs
```diff
@@ -580,13 +582,109 @@ enum SequencerBlockHeaderErrorKind {
...
+///     "KioqKioqKioqKioqKioqKioqKioqKioqKioqKioqKio=",
+///     format!("{block_hash:#}"),
+/// );
+/// ```
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct Hash([u8; 32]);
+
+impl Hash {
+    #[must_use]
+    pub fn new(inner: [u8; 32]) -> Self {
```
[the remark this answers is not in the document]
SuperFluffy: Made all methods const in [0877766](https://github.com/astriaorg/astria/pull/1884/commits/0877766ee00e4bc71882f6570d3a95ce39493a8b)

crates/astria-core/src/sequencerblock/v1/block.rs
```diff
@@ -580,13 +582,109 @@ enum SequencerBlockHeaderErrorKind {
...
+        Self(inner)
+    }
+
+    #[must_use]
+    pub fn get(&self) -> [u8; 32] {
+        self.0
+    }
```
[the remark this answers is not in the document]
SuperFluffy: Removed the Deref impl in [0877766](https://github.com/astriaorg/astria/pull/1884/commits/0877766ee00e4bc71882f6570d3a95ce39493a8b) and followed your mentioned convetion.

crates/astria-core/src/sequencerblock/v1/block.rs
```diff
@@ -580,13 +582,109 @@ enum SequencerBlockHeaderErrorKind {
...
+}
+
+impl AsRef<[u8]> for Hash {
+    fn as_ref(&self) -> &[u8] {
+        &self.0
+    }
+}
+
+impl Deref for Hash {
```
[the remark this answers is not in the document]
SuperFluffy: I agree and removed it in [0877766](https://github.com/astriaorg/astria/pull/1884/commits/0877766ee00e4bc71882f6570d3a95ce39493a8b). I also removed the `AsRef<[u8]>` impl [3d168ae](https://github.com/astriaorg/astria/pull/1884/commits/3d168ae5503654d00af8b65240b401c11f0c3eaf) so that we don't accidentally overlooked uses of base64 and hex.

SuperFluffy:
@Fraser999 addressed all your comments. I have made a note in `crates/astria-sequencer/src/grpc/storage/values/block_hash.rs` to flag potential confusion as relates to block hashes appearing in observability vs on-disk and on-wire.

Fraser999 (approved):
LGTM - nice changeset!

PR #1885, opened 2024-12-23
chore(conductor): don't return empty tuple

### LINKED ISSUE #1879: conductor: don't return empty tuple
This function should be refactored to not return an empty tuple:
https://github.com/astriaorg/astria/blob/1f2b9677f158c0028b0faed83bf484d05079cda7/crates/astria-conductor/src/celestia/mod.rs#L180

┆Issue Number: ENG-1051

ethanoroshiba:
## Summary
Removed empty tuple member in return type.

## Background
We shouldn't be returning a tuple with an empty member.

## Changes
- Removed empty member from tuple return.

## Testing
Passing all tests.

## Changelogs
No updates required.

## Breaking Changes
Noting that this requires overriding the Conductor code freeze, but due to the very small nature of this change I believe this will be okay since it should not have any bearing on auditing of the service.

## Related Issues
closes #1879

### DIFF
--- a/crates/astria-conductor/src/celestia/mod.rs
+++ b/crates/astria-conductor/src/celestia/mod.rs
@@ -155,7 +155,7 @@ pub(crate) struct Reader {
 
 impl Reader {
     pub(crate) async fn run_until_stopped(mut self) -> eyre::Result<()> {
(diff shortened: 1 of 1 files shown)


SuperFluffy (approved):
Thanks, this is better!

PR #1898, opened 2025-01-08
refactor(sequencer): split run method to eliminate long-lived spans

### LINKED ISSUE #1893: Provide docs for `Sequencer::run_until_stopped`
https://github.com/astriaorg/astria/pull/1761 quieted a clippy warning using an incorrect reason: differently from what is claimed, the method is public.

We should instead quiet the warning by providing docs on what `Sequencer::run_until_stopped` will do.

https://github.com/astriaorg/astria/blob/ffbd008fb4c2d170db75a794e32c04c37ca533a8/crates/astria-sequencer/src/sequencer.rs#L50

┆Issue Number: ENG-1059

### LINKED ISSUE #1895: sequencer: split `run_until_stopped` into two methods
We should split `sequencer::run_until_stopped` into two parts, one for setting everything up, and the other for awaiting the stop signals/shutting down.

This is spurred firstly by the need to improve instrumentation: currently, the info span used lasts the entirety of the life of the method, and hence the sequencer. We should avoid these long lasting spans while continuing to provide detailed logging. In addition to this, splitting the method into two parts will follow more closely the convention throughout the rest of the codebase of having `new` and `run` methods.

https://github.com/astriaorg/astria/blob/ffbd008fb4c2d170db75a794e32c04c37ca533a8/crates/astria-sequencer/src/sequencer.rs#L51

┆Issue Number: ENG-1060

ethanoroshiba:
## Summary
Split `run_until_stopped()` into two methods and called them both in `spawn()`.

## Background
There are two reasons for making this change, firstly being to eliminate the long-lived span in `run_until_stopped()`. Secondly, this is for parity with the rest of the codebase, which separates the creation and/or setup of the components from the running process.

## Changes
- Moved initialization logic into `initialize()`, which is instrumented.
- Called both `initialize()` and `run_until_stopped()` in new method `spawn()`.
- Moved signal handler initialization into `spawn()` and poll for signal receipt during initialization.
- Added doc comment for the public `spawn()` method.

## Testing
Passing all current tests, no additional testing needed.

## Changelogs
Changelogs updated

## Breaking Changelist
- Breaks public API, as the public-facing run function is now `spawn()` instead of `run_until_stopped()`

## Related Issues
closes #1895
closes #1893


crates/astria-sequencer/src/sequencer.rs
```diff
@@ -46,35 +48,72 @@ use crate::{
...
+    ///
+    /// # Errors
+    /// Returns an error in the following cases:
+    /// - Database file does not exist, or cannot be loaded into storage
+    /// - The app fails to initialize
+    /// - Info service fails to initialize
+    /// - The server builder fails to return a server
+    /// - The gRPC address cannot be parsed
+    /// - The gRPC server fails to exit properly
+    pub async fn spawn(config: Config, metrics: &'static Metrics) -> Result<()> {
```
ethanoroshiba, line 75: This technically breaks the public API, is this acceptable or should I aim to keep the original function name?

Cargo.lock
Fraser999, line 1: I wouldn't expect the lock file to need updated as part of this changeset.  Committed by mistake?

crates/astria-sequencer/CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Consolidate action handling to single module [#1759](https://github.com/astriaorg/astria/pull/1759).
 - Ensure all deposit assets are trace prefixed [#1807](https://github.com/astriaorg/astria/pull/1807).
 - Update `idna` dependency to resolve cargo audit warning [#1869](https://github.com/astriaorg/astria/pull/1869).
+- Change public facing run function from `run_until_stopped` to `spawn` [#1898](https://github.com/astriaorg/astria/pull/1898).
```
Fraser999, line 19: I'd be inclined to omit this from the changelog for the sake of keeping it concise.  It's basically an invisible change as far as users/operators are concerned.

crates/astria-sequencer/src/sequencer.rs
```diff
@@ -46,15 +48,72 @@ use crate::{
...
+    async fn initialize(
+        config: Config,
+        metrics: &'static Metrics,
+    ) -> Result<(RunningGRPCServer, RunningABCIServer)> {
         cnidarium::register_metrics();
         register_histogram_global("cnidarium_get_raw_duration_seconds");
         register_histogram_global("cnidarium_nonverifiable_get_raw_duration_seconds");
-        let span = info_span!("Sequencer::run_until_stopped");
-
-        let mut signals = spawn_signal_handler();
```
Fraser999, line 57: I know I recently asked for variables to be intialized as late as possible, just before usage, but this is a case where we want the signal handler to be initialized as early in the flow as possible, so external signals can be used to terminate the process even during the remainder of the app's initialization.  I'd prefer if we reverted this change.  It should probably even be done before registering the metrics.

crates/astria-sequencer/src/sequencer.rs
Fraser999: I think it's fine.  We don't really treat the lib like a publicly-consumable lib imo.

crates/astria-sequencer/src/sequencer.rs
```diff
@@ -46,15 +48,72 @@ use crate::{
...
+
+    async fn run_until_stopped(
+        abci_server: RunningABCIServer,
+        grpc_server: RunningGRPCServer,
+    ) -> Result<()> {
+        let mut signals = spawn_signal_handler();
+
+        select! {
+            _ = signals.stop_rx.changed() => {
+                info_span!("sequencer").in_scope(|| info!("shutting down sequencer"));
```
Fraser999, line 88:
```suggestion
                info_span!("run_until_stopped").in_scope(|| info!("shutting down sequencer"));
```

crates/astria-sequencer/src/sequencer.rs
```diff
@@ -46,15 +48,72 @@ use crate::{
...
+            }
+
+            _ = abci_server.shutdown_rx => {
+                info_span!("sequencer").in_scope(|| error!("ABCI server task exited, this shouldn't happen"));
```
Fraser999, line 92:
```suggestion
                info_span!("run_until_stopped").in_scope(|| error!("ABCI server task exited, this shouldn't happen"));
```

Cargo.lock
ethanoroshiba: Strange one, should be fixed in 294c0c5.

crates/astria-sequencer/CHANGELOG.md
ethanoroshiba: Removed in 294c0c5.

crates/astria-sequencer/src/sequencer.rs
ethanoroshiba: Surfacing my offline comment:

> I was under the impression that it was remaining similar in functionality, since the signal receiver wasn't polled until the run loop previously anyways

I've moved the signal handler spawn to before anything else and poll it during initialization now as well (294c0c5). Let me know if this works!

Fraser999: Yeah - you're correct - I misread the original version.

I think what you've done is better than the original though, so let's keep it :)

Fraser999 (approved):

PR #1900, opened 2025-01-08
chore(core, sequencer): Create price feed enum action and nest relevant actions inside

### LINKED ISSUE #1896: Nest new currency pair actions into new wrapping enum
The two individual actions `AddCurrencyPairs` and `RemoveCurrencyPairs` should be replaced with the following:

```rust
pub enum PriceFeed {
    Oracle(CurrencyPairChange),
}

pub enum CurrencyPairChange {
    Addition(Vec<CurrencyPair>),
    Removal(Vec<CurrencyPair>),
}
```

┆Issue Number: ENG-1061

Fraser999:
## Summary
This unflattens the two closely-related price feed actions into a new enum.

## Background
We want to be consistent with other existing actions where we group related actions under a parent enum.  This will help avoid the main action enum becoming too cluttered over time.

## Changes
- Provided `PriceFeed` action enum, in anticipation of another variant (`MarketMap`) being added soon to support actions affecting the market map.
- Provided `CurrencyPairsChange` enum which subsumes the recently-added `AddCurrencyPairs` and `RemoveCurrencyPairs` action structs.
- Changed how action groups are defined (but not the actual definitions themselves).  This is in anticipation of the new `PriceFeed` action belonging to more than one group when the market map actions are included.
- No business logic should have changed, likewise for test logic.  This is a pure refactor.

## Testing
No new tests required, existing tests refactored as required.

## Changelogs
No updates required.

## Related Issues
Closes #1896.


noot (approved):

proto/protocolapis/astria/protocol/transaction/v1/action.proto
```diff
@@ -233,19 +232,33 @@ message FeeChange {
...
+}
+
+message CurrencyPairsToAdd {
   repeated connect.types.v2.CurrencyPair pairs = 1;
 }
 
-message RemoveCurrencyPairs {
+message CurrencyPairsToRemove {
   repeated connect.types.v2.CurrencyPair pairs = 1;
 }
```
noot, line 264: nit but since these two are the same type, can we use just one type `CurrencyPairs` and determine addition or removal based on enum variant?

Fraser999: Good point.  Done in [1180e60](https://github.com/astriaorg/astria/pull/1900/commits/1180e60eafa5da1188e7e71d5b47c009acbbe585).

PR #1905, opened 2025-01-09
fix(sequencer)!: use bridge address to determine asset in bridge unlock cost estimation instead of signer

### LINKED ISSUE #1904: bug(sequencer): fix bridge unlock cost calculation
Here, the cost calculation for `BridgeUnlock` incorrectly attempts to retrieve the bridge account asset from the signer of the transaction. This is fine if the signer is the bridge address, but if it is instead the bridge withdrawer address, the cost calculation will erroneously fail.
https://github.com/astriaorg/astria/blob/a19b82243962a89fbdee149cac00e0c3a3d9fefc/crates/astria-sequencer/src/transaction/checks.rs#L103

ethanoroshiba:
## Summary
Changed bridge unlock cost calculation to use bridge address instead of signer address.

## Background
Previously, the `get_total_transaction_cost` attempted to retrieve the bridge account asset by the tx signer instead of by the bridge address. This works fine if the sender is a bridge account, but if it is the authorized bridge withdrawer, it will fail.

## Changes
- Change `BridgeUnlock` cost calculation to retrieve asset based on the action's bridge address instead of the transaction signer.

## Testing
Added unit test to ensure function works correctly when submitting a `BridgeUnlock` from the bridge withdrawer address.

Synced to mainnet from genesis to ensure this would not be a network breaking change.

## Changelogs
Changelog updates.

## Breaking Changelist
- This change is breaking since previous `BridgeUnlock` submitted from the withdrawer address would fail. This has been synced to mainnet successfully from genesis as of January 13, 2025 at 2:32pm CST.

## Related Issues
closes #1904

### DIFF
--- a/crates/astria-sequencer/CHANGELOG.md
+++ b/crates/astria-sequencer/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Ensure all deposit assets are trace prefixed [#1807](https://github.com/astriaorg/astria/pull/1807).
 - Update `idna` dependency to resolve cargo audit warning [#1869](https://github.com/astriaorg/astria/pull/1869).
 - Remove events reporting on state storage creation [#1892](https://github.com/astriaorg/astria/pull/1892).
+- Fix bridge unlock cost calculation [#1905](https://github.com/astriaorg/astria/pull/1905).
 
 ## [1.0.0] - 2024-10-25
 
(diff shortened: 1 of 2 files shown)


crates/astria-sequencer/CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Ensure all deposit assets are trace prefixed [#1807](https://github.com/astriaorg/astria/pull/1807).
 - Update `idna` dependency to resolve cargo audit warning [#1869](https://github.com/astriaorg/astria/pull/1869).
 - Remove events reporting on state storage creation [#1892](https://github.com/astriaorg/astria/pull/1892).
+- Fix bridge unlock cost calculation [#1905](https://github.com/astriaorg/astria/pull/1905).
```
SuperFluffy, line 19: Please add a more descriptive summary of this change (both here and in the PR title).

ethanoroshiba: Done in c6cbed6.

SuperFluffy (approved):
This reads like the right change to make, but the naming of the containing function, `get_total_transaction_cost`, is highly confusing to me. This likely played a role in this bug being introduced in the first place.

I remember coming across this function earlier and being caught by this.

I think we should remove this function entirely in favor of two functions, `calculate_fees_for_transaction` (already exists) and `calculate_funds_moved_by_transaction` (essentially move the loop over the actions to it). The check for enough funds would then take the result of both methods.

I think that would make the current situation much clearer.

noot (approved):

PR #1906, opened 2025-01-10
feat(charts): Snapshots for geth nodes

mycodecrafting:
## Summary
Allows restoring a geth node from a snapshot, and creating snapshots on a schedule with the option to upload them to remote storage via rclone.

## Background
Spinning up a geth node from a snapshot is more ideal than syncing from genesis. We won't always have a PVC snapshot to restore from. This allows any geth node to migrate to Astria infra, and is needed to migrate Forma testnet.

## Changes
- Added a restore from snapshot
  - snapshots can be loaded from any compatible rclone source (including local)
  - rclone config allow snapshots to be loaded from private sources
  - optional sha256 checksum to verify the snapshot after download
- Added scheduled snapshot creation
  - Cron schedule
  - Retention count for # of snapshots to keep on local  disk
  - Optional upload of snapshot to a remote destination
- rclone config management via secrets

## Testing
Tested on local k8s dev cluster, with and without rclone w/ an AWS S3 config


mycodecrafting:
Updated to use `rclone` instead of the aws cli. (fyi @joroshiba)

This is ready for review @WafflesVonMaple

Issue #1809, opened 2024-11-12
rename and refactor `begin_block` and `end_block`


ethanoroshiba:
Both the `Component` trait and `App` have methods called `begin_block` and `end_block`. In addition to this, the corresponding, deprecated ABCI structs are needlessly passed through as arguments to these. These methods are no longer present in ABCI as of CometBFT v0.38, so we should rename them and refactor such that we aren't passing objects which have nothing done with them.

joroshiba:
This issue is stale because it has been open 45 days with no activity. Remove stale label or this issue
be closed in 7 days.

PR #1917, opened 2025-01-15
feat(core, sequencer)!: provide upgrade functionality for sequencer

### LINKED ISSUE #1874: Proof of concept for Sequencer upgrade
Create PoC and update https://astria.atlassian.net/wiki/spaces/Astria/pages/101318658/Sequencer+Upgrades as per recent discussion.

Summary of main changes to design doc:
- config env var not optional - for new networks, provide empty upgrade file
- have single `Upgrade` hold multiple `Change`s
- validate all applied upgrade `Change`s on startup (check hash of each in storage)
- borsh encode each `Change` to get hash
- in `handle_info_request`, provide details of staged upgrade(s) in the `data` field
- add `app_version` to upgrade. Set it during finalize block via `consensus_param_updates`, and in `handle_info_request`

┆Issue Number: ENG-1045

Fraser999:
## Summary
This introduces the functionality to support upgrades of the sequencer network along with the initial bulk of the first such upgrade.

## Background
We need the ability to upgrade the sequencer network.  In the short-term, we will use this upgrade ability to support introducing the oracle price feed data to the network and to provide additional sequencer actions (support for validator set updates).

The new functionality adds support for "Soft Coordinated Upgrades" and "Coordinated Upgrades" as per the definitions in the [CometBFT Taxonomy of Upgrades](https://github.com/cometbft/cometbft/wiki/Upgrades-taxonomy#cometbft-taxonomy-of-upgrades).

Somewhat separate to the _ability_ to upgrade, the PR also implements `Upgrade1` (name to be bikeshedded): a soft coordinated upgrade.  This currently provides the changes required to switch on the Connect Oracle price feed.  There is skeletal support for validator update actions, but that will be completed as part of follow-up work.

Implementing the upgrade involved changing some of the details of the Connect Oracle, since when it was written there wasn't a clear vision of how the upgrade would be implemented.

## Changes
- Provided an `upgrades/v1` module in core.  The top-level object is the `Upgrades` struct which will be parsed from a JSON file on sequencer startup.  It holds an ordered collection of `Upgrade` enums, where this is an enum over all the different possible upgrades (currently just one).  A trait was considered here, but an enum is more appropriate as we will always have a closed set of upgrades and the code is significantly simpler in this form.  Each upgrade will be comprised of one or more "changes" - a self-contained part of the upgrade.  There is a `Change` trait provided as it was simple to make this object-safe, which is useful to keep the code in sequencer clean, since we can iterate over any upgrade's changes without having to match on every upgrade variant.
- A `ChangeInfo` struct is provided which comprises the basic details of a given upgrade change.  It is put to storage when an upgrade executes, the collection of all such stored changes are checked on sequencer startup to ensure the provided upgrades.json file contains all the correct historical upgrades.  It is also used in the `GetUpgradesInfoResponse` of the new gRPC endpoint at `/v1/sequencer/upgrades` to provide a summary of the contents of the parsed upgrades.json.  This will be useful for establishing which validators have correctly staged pending upgrades in the future, assuming the gRPC is open.
- Significant modifications were made in `astria_core::sequencerblock::v1::block` to support changing how the raw txs of the CometBFT requests/responses are created.  This generally revolves around the provision of the new `DataItem` enum (which makes explicit the type of each encoded element of `txs` other than rollup txs) and the related `ParsedDataItems` struct which provides the logic for parsing and validating the encoded `txs`.  The `DataItem` enum will only be used after `Upgrade1` has been applied so that backwards compatibility is maintained. The order of items in the collection has not changed, other than there can now be an element at index 2 which is the encoded upgrade change hashes.  Note that the Merkle tree used for creating the block proofs doesn't use encoded `DataItem`s as its input, rather it maintains the current inputs for existing data types (the two commitments and the rollup txs).  NOTE: for the two new variants, one (the upgrades change hashes) uses Borsh-encoding to ensure deterministic hash input, and the other (extended commit info) retains a copy of the proposer's raw bytes to avoid re-encoding in a potentially non-deterministic way.  It might be worth considering a single approach here.
- The bulk of the sequencer changes are in `App` to support applying an upgrade if due, and in the new module `upgrades` which deals with storing upgrade info, verifying historical upgrades have been applied, and shutting down the sequencer in the case of a Coordinated Upgrade.
- The consensus params passed by CometBFT in `init_chain` are now being put to storage, and again during `Upgrade1` since they are modified during that upgrade.  Since we haven't been storing these historically, we also now have a CometBFT client which fetches these during `Upgrade1` in order to update and store them.
- The app version is updated as part of executing an upgrade.  This means it is provided to CometBFT as part of the updated consensus params in the `FinalizeBlock` response, and is also held in the sequencer's own storage.  This is the version reported in response to a gRPC `Info` request.

## Testing
- Reasonably extensive unit tests of the nontrivial changes to core have been added (sequencerblock/v1/tests.rs had previously been accidentally excluded from the build, so this was reinstated as sequencerblock/v1/block/tests.rs, updated to compile and the new tests added there).
- A lengthy unit test has been added for the sequencer `App`.  It performs the CometBFT calls which would be expected as the network crosses an upgrade point.  It covers the cases of a proposer, a non-proposing validator, and a syncing validator/full-node.
- Ran an upgrade locally (i.e. started a sequencer with vote extensions disabled and an upgrade file including `Upgrade1` and observed the node execute the upgrade and begin to produce extended commit info)
- Ran the smoke test locally.

## Changelogs
Changelogs will be updated once the feature branch is completed.

## Breaking Changelist
- The feature branch already introduced extended vote info to the raw txs, and this PR changes how those are encoded as well as providing optional upgrade change hashes in the raw txs.
- The app hash snapshot tests have been updated as upgrade change hashes are now stored, as well as becoming part of the block.
- Added `ASTRIA_SEQUENCER_UPGRADES_FILEPATH` config env var.
- Added `ASTRIA_SEQUENCER_COMETBFT_RPC_ADDR` config env var.

## Related Issues
Closes #1874.


crates/astria-core/src/protocol/genesis/v1.rs
```diff
@@ -46,12 +47,7 @@ use crate::{
     Protobuf,
 };
 
-#[derive(Clone, Debug)]
-#[cfg_attr(
-    feature = "serde",
-    derive(serde::Deserialize, serde::Serialize),
-    serde(try_from = "raw::ConnectGenesis", into = "raw::ConnectGenesis")
-)]
+#[derive(Clone, Debug, BorshSerialize)]
```
SuperFluffy, line 50: I Am surprised to see borsh as part of astria-core in general. So far my view of astria-core was for it to be a) a collection of domain types, and b) the conversion from/to on-wire types (protobuf). Borsh on the other hand was intended to be the on-disk format and an implementation detail of sequencer.

Fraser999: Yes - that's up for debate.

It's used to encode the data comprising the upgrade changes in order to get a hash of each change.  This needs to be deterministic so protobuf encoding isn't suitable.

Perhaps best to discuss this elsewhere and I can update here with the conclusion.

crates/astria-sequencer/src/app/mod.rs
```diff
@@ -978,35 +1125,23 @@ impl App {
...
         // cometbft expects a result for every tx in the block, so we need to return a
-        // tx result for the commitments, even though they're not actually user txs.
+        // tx result for the commitments and other injected data items, even though they're not
+        // actually user txs.
         //
         // the tx_results passed to this function only contain results for every user
-        // transaction, not the commitment, so its length is len(txs) - 3.
-        let mut finalize_block_tx_results: Vec<ExecTxResult> = Vec::with_capacity(txs.len());
+        // transaction, not the injected ones.
+        let non_rollup_tx_count = parsed_data_items.non_rollup_transaction_count();
```
noot, line 1140: rename to `non_injected_tx_count`? the naming here is a bit confusing, as `Transfer` is not a rollup tx for example

crates/astria-core/src/sequencerblock/v1/block/mod.rs
```diff
@@ -1165,122 +1162,424 @@ impl ExtendedCommitInfoError {
...
+    pub upgrade_change_hashes_with_proof: Option<UpgradeChangeHashesWithProof>,
+    /// The optional next item in `data` (third or fourth depending on whether an upgrade exists in
+    /// `data`): the encoded extended commit info, along with the proof that this entry exists in
+    /// the data Merkle tree. If extended commit info is enabled at this block height, this entry
+    /// must exist. If not, there must be no entry in `data` for this, as opposed to an empty or
+    /// `None` entry.
+    pub extended_commit_info_with_proof: Option<ExtendedCommitInfoWithProof>,
+    /// All remaining entries of `data`, parsed into rollup txs, in the same order as they appear
+    /// in `data`.
+    pub rollup_transactions: Vec<Arc<Transaction>>,
```
noot, line 1366:
```suggestion
    /// All remaining entries of `data`, parsed into txs, in the same order as they appear
    /// in `data`.
    pub transactions: Vec<Arc<Transaction>>,
```
naming is confusing, as not all user-submitted txs are for rollups

crates/astria-sequencer/src/app/mod.rs
```diff
@@ -1184,25 +1284,48 @@ impl App {
...
                 "post_transaction_execution_result must be present, as txs were already executed \
                  just now or during the proposal phase",
             );
 
+        self.update_consensus_params_if_upgrade_due(
+            finalize_block.height,
+            &mut consensus_param_updates,
+        )
+        .await
+        .wrap_err("failed to update consensus params")?;
```
noot, line 1305: i think having the upgrades handled in two separate (and somewhat specific) functions may get confusing in the future. what do you think about having an `UpgradeComponent` with explicit `begin_block` and `end_block` hooks, and the logic for `execute_upgrade_if_due` would go in the `begin_block` hook and the logic for `update_consensus_params_if_upgrade_due` would go in the `end_block` hook. i think that would be clearer/more explicit as to when certain changes should happen.

crates/astria-sequencer/src/sequencer.rs
```diff
@@ -100,13 +110,35 @@ impl Sequencer {
...
+            .wrap_err("failed to establish if sequencer should shut down for upgrade")?
+        {
+            info!(
+                upgrade_activation_height,
+                latest_app_hash = %hex_encoded_app_hash,
+                latest_block_time = %block_time,
+                "shutting down for upgrade"
+            );
+            return Ok(());
+        }
```
noot, line 141: how does the node restart after shutting down for an upgrade? the code here would always just shut the node down if started up at that height, so would it be to perform some update then change `should_shut_down` to false, then start the node?

proto/sequencerblockapis/astria/sequencerblock/v1/block.proto
```diff
@@ -151,6 +151,24 @@ message RollupData {
   }
 }
 
+message UpgradeChangeHashesWithProof {
+  // The SHA256 digests of all upgrade changes applied during this block, if an upgrade was
+  // activated at this height.
+  repeated bytes upgrade_change_hashes = 1;
```
noot, line 157: why have multiple hashes instead of just one?

proto/sequencerblockapis/astria/sequencerblock/v1/block.proto
```diff
@@ -151,6 +151,24 @@ message RollupData {
...
+  // specifically the third item in the data field.
+  astria.primitive.v1.Proof proof = 2;
+}
+
+message ExtendedCommitInfoWithProof {
+  // The extended commit info for the block, if vote extensions were enabled at this height.
+  bytes extended_commit_info = 1;
+  // The proof that the extended commit info is included in the cometbft block data (if it
+  // exists). If no upgrade change hashes exist in the block, this will be the third item in the
+  // cometbft block data field, otherwise it will be fourth.
```
noot, line 168: move the "If no upgrade change hashes exist" part to above `Message ExtendedCommitInfoWithProof`?

proto/sequencerblockapis/astria/sequencerblock/v1/block.proto
```diff
@@ -151,6 +151,24 @@ message RollupData {
   }
 }
 
+message UpgradeChangeHashesWithProof {
+  // The SHA256 digests of all upgrade changes applied during this block, if an upgrade was
+  // activated at this height.
+  repeated bytes upgrade_change_hashes = 1;
+  // The proof that the change hashes are included in the cometbft block data (if it exists),
+  // specifically the third item in the data field.
+  astria.primitive.v1.Proof proof = 2;
```
noot, line 160: why do we need an upgrade merkle proof? the reason for including a proof for the other fields is that it's needed by conductor (using only DA) to verify the data as it's relevant to rollups, but I don't think upgrades on the sequencer application are necessarily relevant to rollups/conductor?

proto/upgrades/astria/upgrades/v1/upgrades.proto
```diff
@@ -0,0 +1,49 @@
...
+  BaseUpgradeInfo base_info = 1;
+  ConnectOracleChange connect_oracle_change = 2;
+  ValidatorUpdateActionChange validator_update_action_change = 3;
+}
+
+// Brief details of a given upgrade change.
+//
+// This is NOT part of the upgrades file schema, rather it is used by the sequencer gRPC service in
+// the `GetUpgradesInfo` response.
+message ChangeInfo {
```
noot, line 40: maybe move to `service.proto` then?

crates/astria-core/src/sequencerblock/v1/block/mod.rs
```diff
@@ -620,38 +631,49 @@ enum SequencerBlockHeaderErrorKind {
...
     /// # Errors
     ///
-    /// - if the first transaction in the data is not the 32-byte rollup transactions root.
-    /// - if the second transaction in the data is not the 32-byte rollup IDs root.
-    /// - if `with_extended_commit_info` is set to `true` and the third transaction in the data is
-    ///   not the extended commit info.
-    /// - if any transaction in the data cannot be decoded.
-    /// - if the rollup transactions root does not match the reconstructed root.
-    /// - if the rollup IDs root does not match the reconstructed root.
+    /// Returns an error if `uses_data_item_enum` is false and `with_extended_commit_info` is true,
```
noot, line 658: i don't see `uses_data_item_enum` in this file anywhere other than this comment?

crates/astria-core/src/sequencerblock/v1/block/mod.rs
```diff
@@ -1165,122 +1162,424 @@ impl ExtendedCommitInfoError {
...
-    })
+                let extended_commit_info =
+                    ExtendedCommitInfoWithCurrencyPairMapping::try_from_raw(raw_info)
+                        .map_err(SequencerBlockError::invalid_extended_commit_info)?;
+
+                Ok((index, extended_commit_info, encoded_extended_commit_info))
+            })
+            .transpose()?;
+
+        // All further items must be rollup txs.
```
noot, line 1516:
```suggestion
        // All further items must be user submitted txs.
```

crates/astria-core/src/sequencerblock/v1/block/mod.rs
```diff
@@ -1165,122 +1162,424 @@ impl ExtendedCommitInfoError {
...
+                .expect("must be upgrade, since it's upgrade in peek");
+            // Concatenate the raw bytes of all the change hashes to use as the input for the
+            // Merkle tree leaf for this item.
+            let concatenated_hashes = concatenate_upgrade_change_hashes(&upgrade_change_hashes);
+            merkle_tree_leaves.push(Sha256::digest(&concatenated_hashes));
+            Ok((index, upgrade_change_hashes))
+        })
+        .transpose()?;
+
+        // If `with_extended_commit_info` is true, the next item must be an extended commit info.
```
noot, line 1493: do we still need `with_extended_commit_info`? if we're shipping oracle and `DataItem` in upgrade1, can't we assume if this function is called, there always will be extended commit info?

noot: i guess the only block where this isn't the case is the block at which the vote extensions are enabled, since there won't be extended commit info until the next block.

crates/astria-core/src/sequencerblock/v1/block/mod.rs
```diff
@@ -1165,122 +1162,424 @@ impl ExtendedCommitInfoError {
...
+    fn variant_name(&self) -> &'static str {
+        match self {
+            DataItem::RollupTransactionsRoot(_) => Self::ROLLUP_TXS_ROOT_VARIANT_NAME,
+            DataItem::RollupIdsRoot(_) => Self::ROLLUP_IDS_ROOT_VARIANT_NAME,
+            DataItem::UpgradeChangeHashes(_) => Self::UPGRADE_HASHES_VARIANT_NAME,
+            DataItem::ExtendedCommitInfo(_) => Self::EXTENDED_COMMIT_INFO_VARIANT_NAME,
+        }
+    }
+
+    /// Returns the Borsh-encoded bytes of `self`.
```
noot, line 1219: once again, why borsh here? very surprising

crates/astria-core/src/sequencerblock/v1/block/mod.rs
```diff
@@ -1165,122 +1162,424 @@ impl ExtendedCommitInfoError {
...
+        index: usize,
+        expected: &'static str,
+        actual: &'static str,
+    },
+}
+
+/// The parsed elements of the `data` field of a `SequencerBlock`, along with values derived from
+/// these.
+#[derive(Debug)]
+pub struct ParsedDataItems {
```
noot, line 1339: nit but how about naming this `ExpandedBlockData` or something like that? i think "parsed" is a bit confusing since there's items that are generated not parsed, and `DataItems` makes me think of only the types that are wrapped in `DataItem` and not the user txs

noot (commented):
no major problems with the implementation, but have some questions/suggestions relating to encoding, the upgrade hashes with proof type, and general naming. I think this should be good after those changes 👍

crates/astria-sequencer/src/app/mod.rs
Fraser999: Good point.  I see you mentioned this in a few places.  I'll aim to use "user submitted tx" where I was using "rollup tx" and "injected tx" for the others.

crates/astria-sequencer/src/app/mod.rs
Fraser999: I like the general idea, but the two main drawbacks I can see so far are:
- there's an open ticket #1809 which is relevant.  I'm hoping that removes the `Component` trait altogether as it's needlessly restrictive
- `execute_upgrade_if_due` which would be done in `UpgradeComponent::begin_block` currently returns `Result<Option<Vec<ChangeHash>>>`.  Since `begin_block` restricts the return type to `Result<()>`, I'd need to work around that by e.g. dumping the result into the ephemeral store, which I really dislike.

What do you think about providing an `UpgradeHandler` which is more akin to `ProposalHandler`?

crates/astria-sequencer/src/sequencer.rs
Fraser999: Yes, it's for a coordinated upgrade, where the current binary halts, and continues to halt on restart (i.e. getting stuck in a crashloop) until an operator replaces the binary with a new one.  The new binary will need to have hard-coded logic inside `fn should_shut_down()` to return `ContinueRunning` for that specific upgrade.

proto/sequencerblockapis/astria/sequencerblock/v1/block.proto
Fraser999: It makes it more fine-grained, which potentially makes it simpler for external consumers to discover whether a particular upgrade change has been applied or not.  It might never be used, but I think it would be a bit painful to retrofit this functionality, whereas it's pretty trivial to add it now.

proto/sequencerblockapis/astria/sequencerblock/v1/block.proto
Fraser999: >but I don't think upgrades on the sequencer application are necessarily relevant to rollups/conductor

I agree for this first upgrade, but I'm not sure that that will always be the case?  I don't think we have a detailed plan yet for how to handle a breaking change which _would_ affect the conductor (e.g. we change how txs are encoded).

Also, I think it's worthwhile since any external consumer could validate the upgrade via the DA, not necessarily just the conductor.

Saying that, this could be added later relatively easily I guess (not for historical upgrades of course), and including it does cost, in terms of computation and storage fees.  So maybe it's worth discussing further offline?

crates/astria-core/src/sequencerblock/v1/block/mod.rs
```diff
@@ -1165,122 +1162,424 @@ impl ExtendedCommitInfoError {
...
+            rollup_ids_root,
+            rollup_ids_proof,
+            upgrade_change_hashes_with_proof: None,
+            extended_commit_info_with_proof: None,
+            rollup_transactions,
+        })
+    }
+
+    /// Constructs a new `ParsedDataItems` from `data`, expecting the entries to be in the
+    /// post-upgrade1 form of Borsh-encoded `DataItem`s.
```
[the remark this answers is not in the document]
Fraser999: We need deterministic encoding to guarantee identical hashes across validators.

crates/astria-core/src/sequencerblock/v1/block/mod.rs
Fraser999: This comment needs a full overhaul - I'll fix it up thanks.

crates/astria-core/src/sequencerblock/v1/block/mod.rs
Fraser999: There's that case, yes, but also we still support a new network which doesn't enable the price feed and extended commit info.

crates/astria-core/src/sequencerblock/v1/block/mod.rs
Fraser999: Agree it's surprising, but as before, we need deterministic encoding.

noot: i don't understand why `DataItem` encoding needs to be deterministic while user txs can be protobuf - the encoding of all transactions is canonicalized in the block data itself. this is putting the `DataItem`s as borsh-encodded into the block itself right?

noot: > We need deterministic encoding to guarantee identical hashes across validators.

this is only for upgrade hashes within the DataItem though right?

crates/astria-sequencer/src/app/mod.rs
noot: good points, I like the idea of an `UpgradeHandler`

crates/astria-sequencer/src/app/mod.rs
Fraser999: Done in [4a045d3](https://github.com/astriaorg/astria/pull/1917/commits/4a045d36c177676c3af883aed37da3cce61f30b6).

crates/astria-core/src/sequencerblock/v1/block/mod.rs
Fraser999: Renamed to `user_submitted_transactions` in [4a045d3](https://github.com/astriaorg/astria/pull/1917/commits/4a045d36c177676c3af883aed37da3cce61f30b6).

proto/sequencerblockapis/astria/sequencerblock/v1/block.proto
Fraser999: Done in [4a045d3](https://github.com/astriaorg/astria/pull/1917/commits/4a045d36c177676c3af883aed37da3cce61f30b6).

proto/upgrades/astria/upgrades/v1/upgrades.proto
Fraser999: Good point.  Done in [4a045d3](https://github.com/astriaorg/astria/pull/1917/commits/4a045d36c177676c3af883aed37da3cce61f30b6).

crates/astria-core/src/sequencerblock/v1/block/mod.rs
Fraser999: Done in [4a045d3](https://github.com/astriaorg/astria/pull/1917/commits/4a045d36c177676c3af883aed37da3cce61f30b6).

crates/astria-core/src/sequencerblock/v1/block/mod.rs
Fraser999: Done in [4a045d3](https://github.com/astriaorg/astria/pull/1917/commits/4a045d36c177676c3af883aed37da3cce61f30b6).

crates/astria-core/src/sequencerblock/v1/block/mod.rs
Fraser999: > this is only for upgrade hashes within the DataItem though right?

True.  In this case I was thinking we'd want all nodes to generate identical encoded output so that the byte lengths were guaranteed identical, so we can reach consensus around block size limits.  Maybe we can get that using protobuf-encoding, but since these encoded data items aren't part of our API (or at least I don't consider them part of our API - maybe I'm wrong?) I thought it was better to just use a schema with guaranteed outputs, and avoid polluting our protobuf API with a new `DataItem` message type which is really just an internal implementation detail.

crates/astria-sequencer/src/app/mod.rs
Fraser999: Done in [d7c16e0](https://github.com/astriaorg/astria/pull/1917/commits/d7c16e0c2011231ff639d5762d40eb2009151f5d).

crates/astria-core/src/sequencerblock/v1/block/mod.rs
noot: > we'd want all nodes to generate identical encoded output 
there's only one node that needs to generate this output though, which is the proposer, the other nodes would just decode and check the data. 

>  avoid polluting our protobuf API with a new DataItem message type which is really just an internal implementation detail. 
I somewhat agree, but the fact that the DataItems end up in a block makes them "public"/non-internal in a way. I think it's fine to add these as proto definitions, then all transactions whether injected or user-submitted can be proto encoded

crates/astria-sequencer/src/app/mod.rs
```diff
@@ -1301,12 +1297,15 @@ impl App {
...
-            finalize_block.height,
-            &mut consensus_param_updates,
-        )
-        .await
-        .wrap_err("failed to update consensus params")?;
+        self.upgrades_handler
+            .finalize_block(
+                &mut self.state,
+                finalize_block.height,
+                &mut consensus_param_updates,
```
noot, line 1304: `consensus_param_updates` will always be `None` here, as `PostTransactionExecutionResult` comes from `end_block` which always returns `None` for `consensus_param_updates` (which makes sense as we don't have any action types that update consensus params). change `upgrades_handler.finalize_block` to return the updated consensus params and remove the `&mut consensus_param_updates` arg. potentially change `upgrades_handler.finalize_block` to `upgrades_handler.end_block` and put the `consensus_param_updates` in `PostTransactionExecutionResult`, or just remove it from `PostTransactionExecutionResult`.

proto/sequencerblockapis/astria/sequencerblock/v1/block.proto
joroshiba: I think this falls into YAGNI, as long as the general upgrade mechanism exists we could add this functionality if we have a need in future.

crates/astria-sequencer/src/app/mod.rs
Fraser999: Happy to do this.  I had assumed due to the `PostTransactionExecutionResult` already holding optional consensus param updates, that there were plans to allow actions to update these in the future.  If that's not the case, I'll make the changes you suggest but also move the `put_consensus_params` call down a level from `App::finalize_block` to `UpgradesHandler::end_block`, since if that's the only place where the params are updated, then it belongs in that method imo.

proto/sequencerblockapis/astria/sequencerblock/v1/block.proto
Fraser999: Removed in [d21fef9](https://github.com/astriaorg/astria/pull/1917/commits/d21fef98a61ef646becae6b8fd547d78ea836f75).

crates/astria-core/src/sequencerblock/v1/block/mod.rs
Fraser999: Changed to protobuf-encoding in [ad1a8ff](https://github.com/astriaorg/astria/pull/1917/commits/ad1a8ffb511268112b522fc61d045a3ad4c11a3e).

crates/astria-core/src/sequencerblock/v1/block/mod.rs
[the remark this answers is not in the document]
Fraser999: Changed to protobuf-encoding in [ad1a8ff](https://github.com/astriaorg/astria/pull/1917/commits/ad1a8ffb511268112b522fc61d045a3ad4c11a3e).

noot (approved):
no more outstanding comments from me, this looks good to go provided we do some testing. great work!

joroshiba (approved):
approved on behalf of infra for charts.

Just a nit on wanting a comment for the proto that I think should be actioned.

crates/astria-sequencer/src/grpc/sequencer.rs
```diff
@@ -237,6 +254,27 @@ impl SequencerService for SequencerServer {
...
+    ) -> Result<Response<GetUpgradesInfoResponse>, Status> {
+        let snapshot = self.storage.latest_snapshot();
+        let current_block_height = snapshot.get_block_height().await.map_err(|e| {
+            Status::internal(format!("failed to get block height from storage: {e:#}"))
+        })?;
+        let mut response = GetUpgradesInfoResponse::default();
+        for change in self.upgrades.iter().flat_map(Upgrade::changes) {
+            let change_info = change.info().to_raw();
+            if change_info.activation_height <= current_block_height {
+                response.applied.push(change_info);
```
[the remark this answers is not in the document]
Fraser999: > Is there any issue with using only the file here?

Not really.  The main reason I wanted to avoid using the raw file is that its schema becomes a de facto part of our API if users start parsing the string.  It's also fairly verbose, and doesn't reflect what's been written to verifiable storage.  

> Are there any other edge cases?

Not that I can think of.

> If after startup other nodes do not activate an upgrade which this node has as activating at that height will this node fail on finalize block/commit?

Yes.  Consensus must be reached on a scheduled upgrade at its activation point.  Assuming we get a supermajority reaching consensus on the state changes resulting from executing that upgrade, any sequencer not agreeing for whatever reason (upgrade not staged, or upgrade details vary from peers') will fail during finalize block as its state root hash will be different.

I'll add a comment here as suggested.

crates/astria-core/src/protocol/mod.rs
```diff
@@ -22,14 +24,14 @@ pub mod test_utils;
 ///
 /// TODO: This can all be done in-place once <https://github.com/rust-lang/rust/issues/80552> is stabilized.
```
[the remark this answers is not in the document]
Fraser999: I'll ticket that for a follow-up PR.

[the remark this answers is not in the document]
Fraser999: It's still not stabilized apparently (as of 1.85.1).

PR #1925, opened 2025-01-22
chore(composer): improve blackbox test efficiency

### LINKED ISSUE #1920: Improve composer test efficiency
@Fraser999 pointed out that this line has the potential to introduce delay even if initialization is already complete, due to sleeping before the `if` statement:
https://github.com/astriaorg/astria/blob/8879aafebec56fcdefb563a543778b1a25fd35c8/crates/astria-composer/tests/blackbox/helper/mod.rs#L202
The proposed solution is to change to `tokio::task::yield_now().await;`, which greatly improves the speed of many tests, but causes some others to fail, which will need to be reevaluated.

┆Issue Number: ENG-1074

ethanoroshiba:
## Summary
Improves composer testing efficiency by manually ticking the executor's block timer to preempt bundle submission instead of waiting.

## Background
Previously, `loop_until_composer_is_ready()` had a minimum of a 1 second delay, which many of the blackbox tests became reliant on to trigger bundle submission. This change removes the delay in looping until composer is ready, opting to manually tick the block timer via `tokio::time`.

## Changes
- Removed delay in `loop_until_composer_is_ready()` in favor of `tokio::task::yield_now()`.
- Added `TestComposer` method `tick_block_timer`, which sleeps for 10ms to allow for tx submission from the rollup before ticking the block timer (lowest latency I found to be consistently successful), then advances time by `block_time_ms` to trigger bundle submission.
- Made all testing timeouts 100ms and added calls to `tick_block_timer`

## Testing
All tests passing consistently, blackbox tests now run consistently within ~2s locally.

## Changelogs
No updates required.

## Related Issues
closes #1920

### DIFF
--- a/crates/astria-composer/tests/blackbox/executor.rs
+++ b/crates/astria-composer/tests/blackbox/executor.rs
@@ -45,12 +45,7 @@ async fn bundle_triggered_by_block_timer() {
         fee_asset: "nria".parse().unwrap(),
     };
 
-    // make sure at least one block has passed so that the executor will submit the bundle
-    // despite it not being full
-    time::pause();
-    let submission_timeout =
-        Duration::from_millis(test_composer.cfg.block_time_ms.saturating_add(100));
-    time::timeout(submission_timeout, async {
+    time::timeout(Duration::from_millis(100), async {
         composer_client
             .submit_rollup_transaction(SubmitRollupTransactionRequest {
                 rollup_id: Some(rollup_id.into_raw()),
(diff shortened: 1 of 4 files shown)


SuperFluffy:
I like the idea of moving away form hard sleep, but I am ambivalent on the explicit use of tokio `time::pause`, `time::advance`, and `time::resume`, because it comes with its own set of problems (like skipping over timers, triggering a timer multiple times).

It also only works with the `current_thread` runtime flavor, which we have already moved away from in conductor, sequencer-relayer because it doesn't play well with checking shutdown logic (we use a `Drop` impl for the test environments which blocks the thread from continuing on).

We should think about a alternative approach to improve test quality composer away from tokio's time utils.

joroshiba:
closed in 7 days.

joroshiba:

PR #1928, opened 2025-01-24
feat(conductor)!: re-fetch genesis or stop at set height

### LINKED ISSUE #1841: Forma migration

SuperFluffy:
## Summary
Makes the changes necessary in the monorepo to migrate Forma to Mainnet.

## Background
Migrating Forma to Astria's main net requires introducing stop and start logic into the Conductor. A migration using these changes could look like this:

1. Running against a fresh rollup node, Conductor starts and fetches the initial `astria.execution.v2.GenesisInfo` from the rollup using the execution API, which contains `rollup_start_number = 1`, and `rollup_stop_number = 1000`.
2. Conductor executes (syncs) the rollup node from rollup block numbers 1 to 999 (1000 exclusive)
3. Upon executing rollup number 999, conductor resets its inner state and fetches a new `astria.execution.v2.GenesisInfo` from the rollup with new information from which it will continue executing.

For cases where `rollup_stop_number` is unset (or equivalently set to 0), conductor will never reset or halt as there is no stop condition.

For cases where `halt_at_rollup_stop_number` is set in the genesis info, conductor will exit without refetching a new genesis info.

For conductors running in firm-only or in soft-and-firm modes only the firm-executed block triggers a restart or halt (in soft-and-firm mode, upon soft-executing at the rollup stop number no blocks will be executed past it).

For conductors running in soft-only mode the soft-executed block triggers the restart or halt.

```yaml
config:
        AstriaRollupName:
        AstriaForks: []
                # - < fork name >:
                #         height: ""
                #         halt: ""
                #         feeCollector: ""
                #         EIP1559Params: {}
                #         sequencer:
                #                 chainId: ""
                #         celestia:
                #                 chainId: ""
                #                 heightVariance: ""
                #         bridgeAddresses: []
```

### Added new `astria.execution.v2.GenesisInfo` protobuf spec
Changes from v1:
- Renamed field `sequencer_genesis_height` to `sequencer_start_height` with the same tag
- Added fields `rollup_start_number`, `rollup_stop_number`, `sequencer_chain_id`, `celestia_chain_id`, `halt_at_rollup_stop_number`.

### In `astria-core`

- Added `astria_core::execution::v2::GenesisInfo`. Changes from v1:
  - Removed method `sequencer_genesis_block_height`
  - Added methods `sequencer_start_height`, `rollup_start_block_number`, `rollup_stop_block_number`, `sequencer_chain_id`, and `celestia_chain_id`, and `halt_at_rollup_stop_number`

### In `astria-conductor`
- Changed sequencer-to-rollup-height calculation to start from `rollup_start_block_number` in `GenesisInfo`, so that Conductor doesn't start from rollup height 1 again after restarting.
- Removed `ASTRIA_CONDUCTOR_SEQUENCER_CHAIN_ID` and `ASTRIA_CONDUCTOR_CELESTIA_CHAIN_ID` env vars from Conductor (chain ID checks will now be performed against the values retrieved from the genesis info)
- Added restart and halt logic for when the stop block number is reached:  Conductor will shut down at the stop height if `halt_at_rollup_stop_number` is `true`, and restart otherwise.

### In charts
- Add just command for deploying the EVM rollup restart test, which functions the same as our regular smoke test with a restart at sequencer height 20.

## Testing
- Added miscellaneous unit tests to ensure heights are correctly calculated and whether to restart is successfully determined.
- Added blackbox tests to ensure conductor correctly restarts in each mode, and continues to execute at the correct height after fetching updated genesis info and commitment state.
- Added blackbox test in `SoftAndFirm` mode to ensure the conductor shuts down at the stop height if `GenesisInfo.halt_at_rollup_stop_number` is `true`.
- Added EVM rollup restart test which restarts EVM and Conductor at sequencer height 20 during the regular smoke test.

## Changelogs
Changelogs updated.

## Breaking Changelist
- Proto changes are not technically breaking, since fields were only added, but the old genesis info shape will be rejected since it does not contain the chain IDs to validate against the networks the conductor is connecting to. Thus, rollups will need to provide an updated `GenesisInfo` to conductor (including`astria-geth`).
- Breaks domain type `GenesisInfo`.
- Breaks Conductor config.

## Related Issues
closes #1841


crates/astria-grpc-mock/src/mock.rs
```diff
@@ -153,13 +153,15 @@ pub struct MockBuilder {
 impl MockBuilder {
     /// Adds an additional `Matcher` to the [`MockBuilder`]'s `matchers` set. There is no maximum
     /// number of matchers that can be pushed. See [`MockBuilder`] docs for example usage.
+    #[must_use]
```
SuperFluffy, line 156: The `#[must_use]` started being flagged by clippy after macking the `mock` module `pub`. Same as above: break this out into a separate PR before merge.

proto/executionapis/astria/execution/v1/execution.proto
```diff
@@ -14,9 +14,17 @@ message GenesisInfo {
   // The rollup_id is the unique identifier for the rollup chain.
   astria.primitive.v1.RollupId rollup_id = 1;
   // The first block height of sequencer chain to use for rollup transactions.
-  uint32 sequencer_genesis_block_height = 2;
+  uint32 sequencer_start_block_height = 2;
+  // The last block height of sequencer chain to use for rollup transactions.
+  uint32 sequencer_stop_block_height = 3;
```
SuperFluffy, line 19:
```suggestion
uint32 sequencer_stop_height = 3;
```

proto/executionapis/astria/execution/v1/execution.proto
```diff
@@ -14,9 +14,17 @@ message GenesisInfo {
...
   // The allowed variance in celestia for sequencer blocks to have been posted.
   uint64 celestia_block_variance = 4;
+  // The rollup block number to map to the sequencer start block height.
+  uint64 rollup_start_block_height = 5;
+  string sequencer_chain_id = 6;
+  string celestia_chain_id = 7;
+  // True if the conductor should halt at the stop height instead of attempting restart.
+  bool halt_at_stop_height = 8;
```
SuperFluffy, line 27: It feels somewhat odd having the rollup dictate the behavior of conductor here - but how else to do it? Seems to make sense.

crates/astria-conductor/tests/blackbox/helpers/mod.rs
```diff
@@ -379,7 +391,7 @@ impl TestConductor {
         &self,
         mock_name: Option<&str>,
         commitment_state: CommitmentState,
-        expected_calls: u64,
```
SuperFluffy, line 382: For soft-and-firm runs this is necessary because the soft commitment need not run (in those cases where the firm commitment was forwarded first).

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -60,12 +61,13 @@ pub(crate) use builder::Builder;
 
 mod client;
 mod state;
-#[cfg(test)]
-mod tests;
-
+// #[cfg(test)]
```
SuperFluffy, line 64: This needs to be brought back.

crates/astria-conductor/src/executor/state.rs
```diff
@@ -204,44 +225,102 @@ forward_impls!(
...
         })
     }
 
+    /// Returns if the tracked firm state of the rollup has reached the sequencer stop height.
+    ///
+    /// The sequencer stop height being reached is defined as:
+    ///
+    /// ```text
+    /// sequencer_height_of_rollup :=
+    ///    sequencer_start_height + (firm_rollup_number - rollup_start_height) + 1
```
ethanoroshiba, line 272:
```suggestion
    ///    sequencer_start_height + (firm_rollup_number - rollup_start_height)
```

crates/astria-conductor/src/executor/state.rs
```diff
@@ -204,44 +225,102 @@ forward_impls!(
...
+        };
+
+        let sequencer_height_of_rollup =
+            map_firm_to_sequencer_height(&self.genesis_info, &self.commitment_state).expect(
+                "state must only be set through State::try_from_genesis_info_and_commitment_state \
+                 and/or StateSender::try_update_commitment_state, which ensures that the number \
+                 can always be mapped to a sequencer height",
+            );
+
+        sequencer_height_of_rollup.value().saturating_add(1) >= sequencer_stop_height.get()
```
ethanoroshiba, line 289:
```suggestion
        sequencer_height_of_rollup.value() >= sequencer_stop_height.get()
```

crates/astria-conductor/src/executor/state.rs
```diff
@@ -204,44 +225,102 @@ forward_impls!(
...
+    }
+
+    /// Returns if the tracked soft state of the rollup has reached the sequencer stop height.
+    ///
+    /// The sequencer stop height being reached is defined as:
+    ///
+    /// ```text
+    /// sequencer_height_of_rollup :=
+    ///    sequencer_start_height + (soft_rollup_number - rollup_start_height) + 1
```
ethanoroshiba, line 298:
```suggestion
    ///    sequencer_start_height + (soft_rollup_number - rollup_start_height)
```

crates/astria-conductor/src/executor/state.rs
```diff
@@ -204,44 +225,102 @@ forward_impls!(
...
+    ///
+    /// has_firm_number_been_reached :=
+    ///     (sequencer_height_of_rollup + 1) >= sequencer_stop_height
```
ethanoroshiba, line 301:
```suggestion
    ///     sequencer_height_of_rollup >= sequencer_stop_height
```

crates/astria-conductor/src/executor/state.rs
```diff
@@ -204,44 +225,102 @@ forward_impls!(
...
+
+    /// Returns if the tracked soft state of the rollup has reached the sequencer stop height.
+    ///
+    /// The sequencer stop height being reached is defined as:
+    ///
+    /// ```text
+    /// sequencer_height_of_rollup :=
+    ///    sequencer_start_height + (soft_rollup_number - rollup_start_height) + 1
+    ///
+    /// has_firm_number_been_reached :=
```
ethanoroshiba, line 300:
```suggestion
    /// has_soft_number_been_reached :=
```

crates/astria-conductor/src/executor/state.rs
```diff
@@ -204,44 +225,102 @@ forward_impls!(
...
+            return false;
+        };
+
+        let sequencer_height_of_rollup =
+            map_soft_to_sequencer_height(&self.genesis_info, &self.commitment_state).expect(
+                "state must only be updated through StateSender::try_update_commitment_state, \
+                 which ensures that the number can always be mapped to a sequencer height",
+            );
+
+        sequencer_height_of_rollup.value().saturating_add(1) >= sequencer_stop_height.get()
```
ethanoroshiba, line 314:
```suggestion
        sequencer_height_of_rollup.value() >= sequencer_stop_height.get()
```

crates/astria-conductor/src/executor/state.rs
```diff
@@ -276,54 +355,92 @@ impl State {
...
+        map_soft_to_sequencer_height(&self.genesis_info, &self.commitment_state)
+            .map(SequencerHeight::increment)
+    }
 }
 
 /// Maps a rollup height to a sequencer height.
 ///
-/// Returns `None` if `sequencer_genesis_height + rollup_number` overflows
-/// `u32::MAX`.
+/// Returns `None` if `sequencer_start_height + (rollup_number - rollup_start_height)`
```
ethanoroshiba, line 413:
```suggestion
/// Returns error if `sequencer_start_height + (rollup_number - rollup_start_height)`
```

crates/astria-conductor/src/sequencer/block_stream.rs
```diff
@@ -35,6 +36,7 @@ struct Heights {
     rollup_expects: u64,
     greatest_requested_height: Option<u64>,
     latest_observed_sequencer_height: Option<u64>,
+    last_height: Option<NonZeroU64>,
```
ethanoroshiba, line 39: I think it may be better to name this `stop_height` for consistency with the rest of the changes.

crates/astria-conductor/src/sequencer/block_stream.rs
```diff
@@ -49,7 +51,11 @@ impl Heights {
         let not_too_far_ahead =
             potential_height < (self.rollup_expects.saturating_add(self.max_ahead));
         let height_exists_on_sequencer = potential_height <= latest_observed_sequencer_height;
-        if not_too_far_ahead && height_exists_on_sequencer {
+        let last_height_reached = self
+            .last_height
+            .map_or(false, |last_height| potential_height >= last_height.into());
```
ethanoroshiba, line 56:
```suggestion
            .map_or(false, |last_height| potential_height > last_height.into());
```

crates/astria-conductor/src/sequencer/mod.rs
```diff
@@ -152,6 +150,7 @@ impl RunningReader {
         } = reader;
 
         let next_expected_height = rollup_state.next_expected_soft_sequencer_height();
+        let sequencer_end_height = rollup_state.sequencer_stop_block_height();
```
ethanoroshiba, line 153: Should we name this `sequencer_stop_height` for consistency?

crates/astria-conductor/src/sequencer/mod.rs
```diff
@@ -287,6 +289,17 @@ impl RunningReader {
             .set_next_expected_height_if_greater(next_height);
         self.block_cache.drop_obsolete(next_height);
     }
+
+    /// The stop height is reached if a) the next height to be forwarded would be equal
+    /// or greater than the stop height, and b) there is no block currently in flight.
```
ethanoroshiba, line 294:
```suggestion
    /// The stop height is reached if a) the next height to be forwarded would be greater
    /// than the stop height, and b) there is no block currently in flight.
```

crates/astria-conductor/src/sequencer/mod.rs
```diff
@@ -287,6 +289,17 @@ impl RunningReader {
...
+    fn has_reached_stop_height(&self) -> bool {
+        self.rollup_state
+            .sequencer_stop_block_height()
+            .map_or(false, |stop_height| {
+                self.block_cache.next_height_to_pop() >= stop_height.get()
```
ethanoroshiba, line 299:
```suggestion
                self.block_cache.next_height_to_pop() > stop_height.get()
```

crates/astria-conductor/tests/blackbox/helpers/mod.rs
ethanoroshiba: While I agree that this is more flexible, the idea behind the decision to have a specific number of calls was such that we ensure that the Conductor is properly executing both soft and firm blocks when in `SoftAndFirm` mode. This is why the delay capability was added, to ensure that the firm block is not received first.

If we allow for 0 calls to the soft commitment mount, then we are essentially foregoing testing that `SoftAndFirm` is still executing soft blocks properly, since firm commitments will still be properly committed even if something is going wrong with our soft commitments.

crates/astria-core/src/execution/v1/mod.rs
```diff
@@ -61,14 +85,39 @@ impl GenesisInfo {
...
 
     #[must_use]
-    pub fn sequencer_genesis_block_height(&self) -> tendermint::block::Height {
-        self.sequencer_genesis_block_height
+    pub fn sequencer_start_block_height(&self) -> u64 {
+        self.sequencer_start_block_height.into()
+    }
+
+    #[must_use]
+    pub fn sequencer_stop_block_height(&self) -> Option<NonZeroU64> {
```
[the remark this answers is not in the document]
ethanoroshiba: Nice one, I think this provides a lot more clarity around the stop height.

crates/astria-grpc-mock/src/lib.rs
```diff
@@ -10,7 +10,7 @@
 use std::any::Any;
 
 pub mod matcher;
-mod mock;
+pub mod mock;
```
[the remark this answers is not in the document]
ethanoroshiba: Relating to the comment above, I think this ability is unnecessary and perhaps harmful to our testing, so I don't think there should be any need for changes to the gRPC mock.

proto/executionapis/astria/execution/v1/execution.proto
ethanoroshiba: I think ideologically the conductor is meant to act as a sidecar to the rollup, so this makes sense to me. After all, this functionality is meant to accommodate a rollup's need for forks, as opposed to the conductor's.

ethanoroshiba (changes requested):

proto/executionapis/astria/execution/v1/execution.proto
SuperFluffy: Also, should we rather define a `rollup_stop_height` (because this data is provided from the rollup to conductor)?

crates/astria-conductor/src/celestia/mod.rs
```diff
@@ -449,6 +447,17 @@ impl RunningReader {
         }
     }
 
+    /// The stop height is reached if a) the next height to be forwarded would be equal
+    /// or greater than the stop height, and b) there is no block currently in flight.
+    fn has_reached_stop_height(&self) -> bool {
+        self.rollup_state
+            .sequencer_stop_block_height()
+            .map_or(false, |stop_height| {
+                self.block_cache.next_height_to_pop() >= stop_height.get()
```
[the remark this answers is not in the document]
ethanoroshiba: Resolved in 321b163.

crates/astria-conductor/src/celestia/mod.rs
```diff
@@ -449,6 +447,17 @@ impl RunningReader {
         }
     }
 
+    /// The stop height is reached if a) the next height to be forwarded would be equal
+    /// or greater than the stop height, and b) there is no block currently in flight.
```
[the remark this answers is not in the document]
ethanoroshiba: 321b163.

crates/astria-conductor/src/executor/state.rs
```diff
@@ -204,44 +225,102 @@ forward_impls!(
...
+    /// Returns if the tracked firm state of the rollup has reached the sequencer stop height.
+    ///
+    /// The sequencer stop height being reached is defined as:
+    ///
+    /// ```text
+    /// sequencer_height_of_rollup :=
+    ///    sequencer_start_height + (firm_rollup_number - rollup_start_height) + 1
+    ///
+    /// has_firm_number_been_reached :=
+    ///     (sequencer_height_of_rollup + 1) >= sequencer_stop_height
```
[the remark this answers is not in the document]
ethanoroshiba: 321b163

crates/astria-conductor/src/executor/state.rs
ethanoroshiba: 321b163

crates/astria-conductor/src/executor/state.rs
ethanoroshiba: 321b163

crates/astria-conductor/src/executor/state.rs
ethanoroshiba: 321b163

crates/astria-conductor/src/executor/state.rs
ethanoroshiba: 321b163

crates/astria-conductor/src/sequencer/block_stream.rs
ethanoroshiba: 321b163

crates/astria-conductor/src/sequencer/mod.rs
ethanoroshiba: 321b163

crates/astria-conductor/src/sequencer/mod.rs
ethanoroshiba: 321b163

crates/astria-conductor/tests/blackbox/helpers/mod.rs
ethanoroshiba: After offline discussion, the decision was made to test the restart cases for receiving the soft block first and firm block first individually, committed in 06c5621. This allows for us to test that the soft commitment is successfully made when the soft number reaches the stop height first, and that it is successfully skipped when the firm number reaches the stop height first.

crates/astria-grpc-mock/src/lib.rs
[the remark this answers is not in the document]
ethanoroshiba: Quoting reply to the related comment here:
> After offline discussion, the decision was made to test the restart cases for receiving the soft block first and firm block first individually, committed in [06c5621](https://github.com/astriaorg/astria/commit/06c56212190e396ce01d65ca7f3dfdb67a288afa). This allows for us to test that the soft commitment is successfully made when the soft number reaches the stop height first, and that it is successfully skipped when the firm number reaches the stop height first.

crates/astria-conductor/src/conductor/inner.rs
```diff
@@ -208,17 +219,313 @@ fn check_for_restart(err: &eyre::Report) -> bool {
...
+        let genesis_info =
+            astria_core::execution::v1::GenesisInfo::try_from_raw(genesis_info).unwrap();
+        let commitment_state =
+            astria_core::execution::v1::CommitmentState::try_from_raw(commitment_state).unwrap();
+        State::try_from_genesis_info_and_commitment_state(genesis_info, commitment_state).unwrap()
+    }
 
     #[test]
-    fn check_for_restart_ok() {
+    fn should_restart_despire_error() {
```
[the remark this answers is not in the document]
ethanoroshiba: 6d45368.

crates/astria-conductor/src/executor/state.rs
ethanoroshiba: 6d45368.

proto/executionapis/astria/execution/v1/execution.proto
joroshiba: I see there being two potential states and workflows here:

1. The rollup genesis defines a halt and has no follow up configuration. The chain should just halt. The rollup can halt itself, but continue to receive committment updates, just refuse execution of new blocks.

    The conundrum is, how does conductor handle that? How does it know without being told essentially "finalize this fork before moving on"? 

3. There is a follow up fork configured. If given that information the rollup is technically ready to proceed, there is no real need to halt. However, it creates complications in conductor if there is a change in sequencer chain id. Similar problem as above, how do we make sure we finish validation of firm blocks? Probably going to need an update to configuration of node for rpc urls if any chain id changes or anything. 

Maybe we change the name of this to `finalize_at_end` or something like that? I believe in terms of actual behaivior this is all we are dictating. It's a way to ensure that conductor doesn't restart or anything because execution failed at a halted state, and to ensure that we get firm commitments for all entries in the set before moving forward. Make soft only mode a bit odd i guess?

I dunno, seems awkward but "fine" for a sidecar at least.

proto/executionapis/astria/execution/v1/execution.proto
```diff
@@ -14,9 +14,17 @@ message GenesisInfo {
   // The rollup_id is the unique identifier for the rollup chain.
   astria.primitive.v1.RollupId rollup_id = 1;
   // The first block height of sequencer chain to use for rollup transactions.
-  uint32 sequencer_genesis_block_height = 2;
+  uint32 sequencer_start_block_height = 2;
```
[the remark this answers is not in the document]
joroshiba: Yeah agreed we can remove the block

proto/executionapis/astria/execution/v1/execution.proto
joroshiba: So IIUC we currently have:

- `sequencer_start_block_height` - akin to the previous `sequencer_genesis_block_height` but is no longer assumed to be the `0` block height and is instead the block used for rollup block at   `rollup_start_block_number`
- `sequencer_stop_block_height` - the height this fork stops at from sequencer view point. 
- `rollup_start_block_number` - used to ensure executing fork on top of correct previous rollup block

Taking into consideration various other naming proposals, my thinking is these names make sense:
- `sequencer_start_height`
- `rollup_start_block_number`
- `rollup_stop_block_number`

The math for which sequencer height to switch forks at/restart loop is `sequencer_start_height + (rollup_stop_block_number - rollup_start_block_number)`.

Within the defined fork, the math for mapping a sequencer height to a rollup block number is then `current_sequencer_height - sequencer_start_height + rollup_start_block_number = current rollup block number`?

Concrete math wise, if we have this situation:
- `sequencer_start_height`: 101
- `rollup_start_block_number`: 1
-  `rollup_stop_block_number`: 1000

Then the sequencer stop height is 1100. And the rollup block corresponding to sequencer block 500, we would be at rollup block number 400

My thinking here is that height vs block number is nice for clarity of sequencer is height, rollup has block numbers. From rollup defining the fork, it's a bit of a PITA to know the sequencer start height do math for the stop height, much cleaner to have a nice rollup block height in terms of doing math.

proto/executionapis/astria/execution/v1/execution.proto
```diff
@@ -14,9 +14,17 @@ message GenesisInfo {
...
+  // The last block height of sequencer chain to use for rollup transactions.
+  uint32 sequencer_stop_block_height = 3;
   // The allowed variance in celestia for sequencer blocks to have been posted.
   uint64 celestia_block_variance = 4;
+  // The rollup block number to map to the sequencer start block height.
+  uint64 rollup_start_block_height = 5;
```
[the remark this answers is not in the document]
joroshiba: Generally like the idea of `block_number` for rollup and `height` for sequencer as it makes it more clear which you are talking about.

proto/executionapis/astria/execution/v1/execution.proto
ethanoroshiba: Addressed in a83990b.

proto/executionapis/astria/execution/v1/execution.proto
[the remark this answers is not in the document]
ethanoroshiba: Addressed in a83990b.

crates/astria-conductor/src/conductor/inner.rs
```diff
@@ -208,17 +219,313 @@ fn check_for_restart(err: &eyre::Report) -> bool {
...
+        }
+    }
+
+    fn make_rollup_state(genesis_info: GenesisInfo, commitment_state: CommitmentState) -> State {
+        let genesis_info =
+            astria_core::execution::v1::GenesisInfo::try_from_raw(genesis_info).unwrap();
+        let commitment_state =
+            astria_core::execution::v1::CommitmentState::try_from_raw(commitment_state).unwrap();
+        State::try_from_genesis_info_and_commitment_state(genesis_info, commitment_state).unwrap()
+    }
```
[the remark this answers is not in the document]
ethanoroshiba: c5e59d8.

crates/astria-conductor/src/executor/mod.rs
ethanoroshiba: c5e59d8

proto/executionapis/astria/execution/v1/execution.proto
mycodecrafting: The intention here is that halt means indefinitely stop processing new blocks (ie requires manual intervention). If there is a fork defined to halt and a followup fork after also defined, the rollup will never reach the followup fork. You have to manually intervene, remove the halt flag on the fork, re-init, and restart the rollup. The geth PR is coded to refuse execution of new blocks on a halted fork.

My assumption is that "finalize this fork before moving on" is true at every fork. i.e, at every fork, whether halted or not, soft and firm should come into sync before moving on. I believe this is how @ethanoroshiba coded it but I haven't looked too closely.

The `halt_at_stop_height` was intended to just be a way for conductor to gracefully shut itself down when you know that it cannot process anything more. Without it then I believe conductor will get stuck in a restarting loop or a crash loop in k8s when it tries to execute a new block and gets an execution error from geth.

i.e, the intention is this field does not add any functionality but just adds a nicety for ops. I'm not strongly tied to it if it makes sense to remove

crates/astria-conductor/src/sequencer/mod.rs
SuperFluffy: Done in [1176560](https://github.com/astriaorg/astria/pull/1928/commits/1176560fe6f8d2bf51179a24d401be6bc30a1f1a)

proto/executionapis/astria/execution/v1/execution.proto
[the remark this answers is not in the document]
SuperFluffy: Removed

crates/astria-grpc-mock/src/lib.rs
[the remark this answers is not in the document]
ethanoroshiba: gRPC changes made in #1945.

crates/astria-grpc-mock/src/mock.rs
ethanoroshiba: Looks like this is taken care of in #1945 via re-export.

.github/workflows/docker-build.yml
```diff
@@ -165,6 +165,37 @@ jobs:
...
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Setup Smoke Test Environment
+        timeout-minutes: 10
+        run: |
+          TAG=sha-$(git rev-parse --short HEAD)
+          just deploy cluster
+          kubectl create secret generic regcred --from-file=.dockerconfigjson=$HOME/.docker/config.json --type=kubernetes.io/dockerconfigjson
+          echo -e "\n\nDeploying with astria images tagged $TAG"
+          just --set defaultValues "dev/values/rollup/evm-restart-test.yaml" deploy smoke-test $TAG
```
SuperFluffy, line 193: @ethanoroshiba It's not easy to see that this is what defines of the restart test. Can you make this into a self-standing just command?

(I was digging through the code and didn't see this)

Also please mention this smoke test in the PR message.

ethanoroshiba: Done in 8ade517.

charts/evm-rollup/values.yaml
```diff
@@ -11,68 +11,76 @@ images:
     repo: ghcr.io/astriaorg/astria-geth
     pullPolicy: IfNotPresent
     tag: 1.0.0
-    devTag: latest
+    devTag: pr-59
```
aajimal, line 14: Nit: perhaps a specific sha version would be more stable

aajimal (approved):

crates/astria-conductor/src/executor/state.rs
```diff
@@ -280,9 +285,14 @@ impl State {
...
         genesis_info: GenesisInfo,
         commitment_state: CommitmentState,
+        commit_level: crate::config::CommitLevel,
     ) -> Result<Self, InvalidState> {
-        let _ = map_firm_to_sequencer_height(&genesis_info, &commitment_state)?;
-        let _ = map_soft_to_sequencer_height(&genesis_info, &commitment_state)?;
+        if commit_level.is_with_firm() {
+            let _ = map_firm_to_sequencer_height(&genesis_info, &commitment_state)?;
+        }
+        if commit_level.is_with_soft() {
```
SuperFluffy, line 293: This condition is wrong. Soft must update with firm in lock-step even if soft is not set.

joroshiba:
closed in 7 days.

joroshiba:

PR #1932, opened 2025-01-27
fix typo README.md

Hack666r:
- **Change:** Replaced "unparseable" with "unparsable" in the sentence: "Debug formatted fields are bad for observability as they are usually unparseable."


SuperFluffy:
Please don't do this.

https://en.m.wiktionary.org/wiki/unparseable

PR #1934, opened 2025-01-28
feat(sequencer)!: implement `BridgeTransfer` action

### LINKED ISSUE #1921: implement `BridgeTransfer` action
implement a `BridgeTransfer` action which unlocks from one bridge accounts and transfers (locks) directly into another bridge account.

┆Issue Number: ENG-1075

noot:
## Summary
implement `BridgeTransfer` action which transfers funds from one bridge account to another. essentially, it atomically performs a `BridgeUnlock` and a `BridgeLock`.

## Background
this is functionality we want and `BridgeUnlock`ing to another bridge account wouldn't create a `Deposit` correctly.

## Changes
- slightly refactor `BridgeUnlock` to move checks to their own function
- also slightly refactor `BridgeLock` to move execution and `Deposit` emission to their own function 
- implement `BridgeTransfer` action using existing `BridgeUnlock` checks and `BridgeLock` execution logic.

## Testing
unit tests

## Changelogs
Changelogs updated.

## Breaking Changelist

## Related Issues

closes #1921


proto/protocolapis/astria/protocol/transaction/v1/action.proto
```diff
@@ -211,6 +212,35 @@ message BridgeSudoChange {
...
+  astria.primitive.v1.Uint128 amount = 2;
+  // the asset used to pay the transaction fee
+  string fee_asset = 3;
+  // the address on the destination chain which
+  // will receive the bridged funds
+  string destination_chain_address = 4;
+
+  // The memo field can be used to provide unique identifying additional
+  // information about the bridge unlock transaction.
+  string memo = 5;
```
joroshiba, line 228: Do we need a memo here?

SuperFluffy:
Can we have smoke tests for this?

crates/astria-sequencer/src/action_handler/impls/bridge_lock.rs
```diff
@@ -41,75 +42,85 @@ impl ActionHandler for BridgeLock {
...
-            bridge_address: self.to,
-            rollup_id,
-            amount: self.amount,
-            asset: deposit_asset.into(),
-            destination_chain_address: self.destination_chain_address.clone(),
-            source_transaction_id,
-            source_action_index,
-        };
-        let deposit_abci_event = create_deposit_event(&deposit);
+pub(crate) async fn execute_bridge_lock<S: StateWrite>(
```
Fraser999, line 67:
```suggestion
pub(super) async fn execute_bridge_lock<S: StateWrite>(
```

crates/astria-sequencer/src/action_handler/impls/bridge_unlock.rs
```diff
@@ -107,6 +85,39 @@ impl ActionHandler for BridgeUnlock {
     }
 }
 
+pub(crate) async fn check_bridge_unlock<S: StateRead>(
```
Fraser999, line 88:
```suggestion
pub(super) async fn check_bridge_unlock<S: StateRead>(
```

proto/protocolapis/astria/protocol/transaction/v1/action.proto
```diff
@@ -211,6 +212,35 @@ message BridgeSudoChange {
...
+  // the amount to transfer
+  astria.primitive.v1.Uint128 amount = 2;
+  // the asset used to pay the transaction fee
+  string fee_asset = 3;
+  // the address on the destination chain which
+  // will receive the bridged funds
+  string destination_chain_address = 4;
+
+  // The memo field can be used to provide unique identifying additional
+  // information about the bridge unlock transaction.
```
Fraser999, line 227:
```suggestion
  // information about the bridge transfer transaction.
```

proto/protocolapis/astria/protocol/transaction/v1/action.proto
```diff
@@ -211,6 +212,35 @@ message BridgeSudoChange {
...
+  string memo = 5;
+  // the address of the bridge account to transfer from
+  astria.primitive.v1.Address bridge_address = 6;
+  // The block number on the rollup that triggered the transaction underlying
+  // this bridge unlock memo.
```
Fraser999, line 232:
```suggestion
  // this bridge transfer memo.
```

proto/protocolapis/astria/protocol/transaction/v1/action.proto
```diff
@@ -211,6 +212,35 @@ message BridgeSudoChange {
...
+  uint64 rollup_block_number = 7;
+  // An identifier of the original rollup event, such as a transaction hash which
+  // triggered a bridge unlock and is underlying event that led to this bridge
+  // unlock. This can be utilized for tracing from the bridge back to
```
Fraser999, line 236:
```suggestion
  // triggered a bridge transfer and is underlying event that led to this bridge
  // transfer. This can be utilized for tracing from the bridge back to
```

proto/protocolapis/astria/protocol/transaction/v1/action.proto
noot: not necessarily, do you see any future use for it? i left it because BridgeUnlock has it, but don't think BridgeUnlock necessarily needs it either. removed for now

Fraser999 (approved):

SuperFluffy (commented):
`astria-core` needs an entry for this change.

crates/astria-core/src/protocol/transaction/v1/action/mod.rs
```diff
@@ -1894,6 +1912,149 @@ enum BridgeSudoChangeErrorKind {
...
+
+    /// Convert from a raw, unchecked protobuf [`raw::BridgeTransferAction`].
+    ///
+    /// # Errors
+    ///
+    /// - if the `to` field is not set
+    /// - if the `to` field is invalid
+    /// - if the `amount` field is invalid
+    /// - if the `fee_asset` field is invalid
+    /// - if the `from` field is invalid
```
joroshiba, line 1969: should there also be validation that the `destination_chain_address` is set, as well as some restrictions on it's length?

Maybe this is more a sequencer concern, notice we don't have this on bridge lock but probably should? There's a limit to what a reasonable sized destination address is.

joroshiba (approved):
approving for snapshot changes + proto changes. One small nit on the validation.

Noting by default this action can't be done as it is disabled because there are no fees, correct?

crates/astria-core/src/protocol/transaction/v1/action/mod.rs
noot: added it to be unempty. we should enforce the nonempty/size constraint in the sequencer app, not just core, since users can use other tooling to create actions. and yes we should also have this on BridgeLock. i'll make an issue

PR #1940, opened 2025-01-31
fix(sequencer-relayer)!: use 'tx_status' instead of 'GetTx'

### LINKED ISSUE #1936: Relayer gets stuck in infinite loop when transaction is evicted from Celestia mempool
## Description
The relayer becomes stuck when a transaction is evicted from the Celestia mempool. Instead of handling the eviction and attempting to resubmit or recover, it enters an infinite loop trying to check the transaction status.

## Current Behavior
When a transaction is evicted:
1. The relayer continues to use `get tx` to check the block height where the transaction landed
2. Since the transaction was evicted, this endpoint never returns a result
3. The relayer enters an infinite loop with the warning:
```
WARN submit_blobs: astria_sequencer_relayer::relayer::celestia_client: waiting to confirm blob submission reason={"0": "transaction still pending"} tx_hash="2EEACE2C73108772A50EC38658C0B814CA23044F1A3B5F3FD168034E2AF097CB"
```
4. The only current solution is to manually restart the relayer

## Expected Behavior
The relayer should:
1. Detect when a transaction has been evicted
2. Handle the eviction appropriately (e.g., resubmit the transaction or log an error)
3. Continue processing new submissions

## Impact
- Relayer becomes stuck and requires manual intervention
- Firm/Soft drift on rollup that could potentially lead to a rollup halt
- Triggers monitoring alerts for finalized block sync

## Additional Context
- [Celestia core exposes a `TxStatus` endpoint](https://github.com/celestiaorg/celestia-core/blob/main/rpc/core/tx.go#L227) which could be used to gather better status updates on a transaction
  - e.g, [Comitted tx](https://celestia-rpc.polkachu.com/tx_status?hash=0x3C5905A1CC44030793C8DD8EA60773D73937829B042DE19D68FC79DA72C745F8), [Evicted tx](https://celestia-rpc.polkachu.com/tx_status?hash=0x2EEACE2C73108772A50EC38658C0B814CA23044F1A3B5F3FD168034E2AF097CB)

ethanoroshiba:
## Summary
Changes polling method for confirming Celestia commitment from `GetTx` gRPC to JSONRPC `tx_status`.

## Background
Previously, `GetTx` was polled indefinitely. On the client end of this RPC, we'd have no indication of what's happening on Celestia, so we had no way to know if a transaction was, for instance, evicted from the mempool, which would cause a halt of data submission. The [`tx_status` endpoint](https://github.com/celestiaorg/celestia-core/blob/382f0ce95893f7fefcabdbc5881e0ea62f0de3d5/rpc/core/tx.go#L226) solves this problem by providing a more detailed status of the transaction, but is only available via JSONRPC. So, this change has required setting up an HTTP client for the RPC (and not gRPC) Celestia endpoint in the Relayer.

## Changes
- Add env var for `celestia_app_http_endpoint`.
- Start `jsonrpsee` HTTP client during build of the relayer.
- Poll `tx_status` instead of `GetTx`.
- Adjusted charts accordingly

## Testing
- Added http server to test helpers and associated mounts for `tx_status`.
- Fixed a couple tests which broke as a result.
- Need to add tests to cover non "COMMITTED" statuses.

## Changelogs

## Metrics
- Should add metrics for # of evicted transactions, # failed polls

## Breaking Changelist
- Breaks relayer config.

## Related Issues
closes #1936


crates/astria-sequencer-relayer/tests/blackbox/main.rs
```diff
@@ -362,3 +393,308 @@ async fn should_exit_if_celestia_chain_id_mismatch() {
...
+            53,
+            "UNKNOWN",
+            Some(10),
+            Some(10),
+        )
+        .await;
+
+    sequencer_relayer
+        .timeout_ms(
+            11_000,
```
ethanoroshiba, line 645: This is an abhorrently long time for a test to take, but we need to test that the relayer correctly exits if the allotted time passes without the status changing from "UNKNOWN". We cannot advance time manually due to the multithreaded context, so I'm not sure of a way around this :/

crates/astria-sequencer-relayer/tests/blackbox/helpers/mock_celestia_app_server.rs
```diff
@@ -202,29 +201,47 @@ impl MockCelestiaAppServer {
...
+            .await;
+    }
+
+    pub async fn mount_tx_status_response_as_scoped(
+        &self,
+        debug_name: impl Into<String>,
+        status: String,
+        height: i64,
+        expected: Option<u64>,
+        up_to_n_times: Option<u64>,
```
Fraser999, line 222: I don't think we need this flexibility?

To keep the function signature a little simpler, I'd suggest we replace these two optional parameters with a single, non-optional one, e.g. `number_of_times: u64,` and use that to set `expect` and `up_to_n_times` on the builder.

crates/astria-sequencer-relayer/tests/blackbox/main.rs
Fraser999: We could perhaps change the actual timeout based on the `test` feature:
```rust
#[cfg(test)]
const MAX_WAIT_FOR_UNKNOWN: Duration = Duration::from_secs(2);
#[cfg(not(test))]
const MAX_WAIT_FOR_UNKNOWN: Duration = Duration::from_secs(10);
```
Alternatively, 10 seconds might be too long anyway given that Celestia block time is now 6 seconds?

crates/astria-sequencer-relayer/src/relayer/celestia_client/mod.rs
```diff
@@ -322,65 +348,106 @@ impl CelestiaClient {
...
+        hex_encoded_tx_hash: String,
+    ) -> Result<u64, ConfirmSubmissionError> {
+        // The min seconds to sleep after receiving a TxStatus response and sending the next
+        // request.
+        const POLL_INTERVAL_SECS: u64 = 1;
+        // The minimum duration between logs.
+        const LOG_INTERVAL: Duration = Duration::from_secs(5);
+        // The maximum amount of time to wait for a transaction to be committed if its status is
+        // `UNKNOWN`.
+        const MAX_WAIT_FOR_UNKNOWN: Duration = Duration::from_secs(10);
```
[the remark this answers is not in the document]
ethanoroshiba: I suppose I thought because this functionality was moved to the debug level, it could be helpful to have the full picture of each response the relayer has received. I don't have a strong opinion on this, though, so I'd be happy to add that back if you think it's better

crates/astria-sequencer-relayer/src/relayer/celestia_client/mod.rs
```diff
@@ -322,65 +348,106 @@ impl CelestiaClient {
...
-                Ok(None) => {
-                    sleep_secs = MIN_POLL_INTERVAL_SECS;
-                    log_if_due(None);
+            tokio::time::sleep(Duration::from_secs(POLL_INTERVAL_SECS)).await;
+            match self.tx_status(hex_encoded_tx_hash.clone()).await {
+                Ok(TxStatus::Unknown) => {
+                    if start.elapsed() > MAX_WAIT_FOR_UNKNOWN {
+                        self.metrics
+                            .increment_celestia_unknown_status_transaction_count();
+                        break Err(ConfirmSubmissionError::StatusUnknown {
```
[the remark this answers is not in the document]
ethanoroshiba: Changed in 4f304bb

crates/astria-sequencer-relayer/tests/blackbox/helpers/mock_celestia_app_server.rs
```diff
@@ -202,29 +201,47 @@ impl MockCelestiaAppServer {
...
+        &self,
+        debug_name: impl Into<String>,
+        status: String,
+        height: i64,
+    ) {
+        prepare_tx_status_response(status, height)
+            .with_name(debug_name)
+            .mount(&self.mock_server)
+            .await;
+    }
```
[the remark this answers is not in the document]
ethanoroshiba: Thanks for the catch, removed in 4f304bb.

crates/astria-sequencer-relayer/tests/blackbox/main.rs
ethanoroshiba: In 4f304bb I lowered the time to 6s and the minimum logging time to 3. I'm not sure how I feel about building the testing functionality into the source code, but I'll defer to you here on what you think is best, since I think I lack the context of what is best practice. I feel like generally, though, it's best to keep testing logic (especially for blackbox tests) separate

crates/astria-sequencer-relayer/tests/blackbox/helpers/mock_celestia_app_server.rs
ethanoroshiba: Good idea, made this change in 4f304bb.

crates/astria-sequencer-relayer/src/relayer/celestia_client/mod.rs
```diff
@@ -322,65 +349,123 @@ impl CelestiaClient {
...
+            // at least one Celestia block has passed and the transaction should have been
+            // submitted. We then start logging at warn level with an exponential backoff.
+            if start.elapsed() > START_WARN_DELAY {
+                warn!(
+                    reason = format!("transaction status: {status}"),
+                    tx_hash = %hex_encoded_tx_hash,
+                    elapsed_seconds = start.elapsed().as_secs_f32(),
+                    "waiting to confirm blob submission"
+                );
+                log_interval = log_interval.saturating_mul(2);
```
Fraser999, line 426: I'd either put an upper limit on this value, or maybe simpler to just log at a fixed interval, or e.g. every third `tx_status` attempt?

ethanoroshiba: Moved exponential backoff from logging to fetch in e83e712

Fraser999 (approved):

joroshiba:
closed in 7 days.

PR #1942, opened 2025-02-02
docs: fix minor typo

defitricks:
## Summary

Noticed a typo in the "What to do instead" section, point 3.
The phrase **"in to"** should be **"into"** in this context.

Fixed it to improve clarity.


joroshiba:
In order to minimize air drop farming activity on developer productivity, we do not accept first PRs that are non substantive such as spelling fixes.

PR #1943, opened 2025-02-04
feat(charts): flame chart

quasystaty1:
## Summary
Implements a new chart dedicated for the flame package and also integrates the new chart with the astria stack. The `evm-stack` now support running `astria-geth` or the flame fork with an auctioneer out of the box.
## Background
To support auctioneer deployment we need to run geth [flame fork](https://github.com/astriaorg/flame), which implements the necessary changes in [pr-30](https://github.com/astriaorg/flame/pull/30).
## Changes
- adds a new flame-rollup chart
- adds the new chart and auctioneer as dependencies of the `evm-stack`
- new commands and dev value file for local deployment

replace `just deploy-dev-rollup` with `just deploy-flame-dev-rollup` in the deployment process to run `flame` instead of `astria-geth`.

Notes: 
- handle images tags on final merge. 
- does not implement any new smoke tests or workflows.

## Testing
running a local cluster and monitoring geth logs.
`│ INFO [02-04|11:25:13.608] ExecuteOptimisticBlock completed         block_num=52 timestamp=seconds:1738668312 
`
## Changelogs
No updates required.


charts/sequencer/values.yaml
```diff
@@ -131,7 +131,7 @@ sequencer:
     otlpHeaders:
     traceHeaders:
   optimisticBlockApis:
-    enabled: false
+    enabled: true
```
bharath-123, line 134: do not enable it to true by default.

charts/flame-rollup/values.yaml
```diff
@@ -0,0 +1,313 @@
...
+    # Determines what will drive block execution, options are:
+    # - "SoftOnly" -> blocks are only pulled from the sequencer
+    # - "FirmOnly" -> blocks are only pulled from DA
+    # - "SoftAndFirm" -> blocks are pulled from both the sequencer and DA
+    executionCommitLevel: 'SoftAndFirm'
+    # The chain id of the Astria sequencer chain conductor communicates with
+    sequencerChainId: ""
+    # The expected fastest block time possible from sequencer, determines polling
+    # rate.
+    sequencerBlockTimeMs: 2000
```
bharath-123, line 178: lets just default this to 100 for flame rollups. Otherwise, we can potentially miss this while deploying since this is a critical parameter to have especially for the auctioneer flame node.

bharath-123: we need the `sequencerBlockTimeMs` to be as less as possible in order to not have any race conditions between execute block and execute optimistic block when running the flame node as auctioneer node.

charts/auctioneer/values.yaml
```diff
@@ -7,11 +7,11 @@ global:
 
 images:
   auctioneer:
-    repo: ghcr.io/astriaorg/astria-auctioneer
+    repo: ghcr.io/astriaorg/auctioneer
```
bharath-123, line 10: Note, also bump up auctioneer's cpu and memory usage.

bharath-123: increase the auctioneer cpu limits to 2 and cpu requests to 1.

charts/sequencer/templates/configmaps.yaml
```diff
@@ -64,7 +64,6 @@ data:
   ASTRIA_SEQUENCER_METRICS_HTTP_LISTENER_ADDR: "0.0.0.0:{{ .Values.ports.sequencerMetrics }}"
   ASTRIA_SEQUENCER_FORCE_STDOUT: "{{ .Values.global.useTTY }}"
   ASTRIA_SEQUENCER_PRETTY_PRINT: "{{ .Values.global.useTTY }}"
-  ASTRIA_NO_OPTIMISTIC_BLOCKS: "{{ not .Values.sequencer.optimisticBlockApis.enabled }}"
```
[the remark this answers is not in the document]
quasystaty1: It is configured twice.

charts/flame-rollup/values.yaml
quasystaty1: Understood, but this is not ideal for deployment which does not use the auctioneer, in my opinion should stay as part of the deployment configuration.

charts/sequencer/templates/configmaps.yaml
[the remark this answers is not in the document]
bharath-123: ahh cool then

charts/flame-rollup/values.yaml
bharath-123: sure, but we need to ensure that we note that this value is set to 100ms specifically for the auctioneer flame node.

quasystaty1: I'll add a note and make sure this value is visible in the default value file.

charts/auctioneer/values.yaml
bharath-123: discussed offline and the outcome is that we want values in charts to be valid for local deployment only.

charts/sequencer/values.yaml
quasystaty1: Note that the user will have to manually change this as well to enable auctioneer. I have added it to the default value file as well for visibility.

bharath-123 (approved):
i tested these changes locally running flame in auctioneer mode and non-auctioneer mode. things seem to be generally working fine.

joroshiba (approved):
Fine for now, don't love the duplication being needed, but will get the job done for right now.

PR #1948, opened 2025-02-05
feat(bridge-withdrawer): support FROST threshold signing

### LINKED ISSUE #1937: Support threshold signing in bridge withdrawer
PRD (private): https://www.notion.so/astria-org/Rough-Sketch-of-Threshold-Signing-for-Bridge-17e6bd31a90c8056a8c8fb980620ace7

┆Issue Number: ENG-1081

noot:
## Summary
support FROST ed25519 threshold signing in the bridge withdrawer. 

## Background
improves security of the withdrawer over having one private key used; instead, we can employ an m-of-n threshold signature scheme to sign withdrawals.

## Changes
- create `FrostParticipantService` gRPC proto definition, which is to be implemented by a binary which contains a FROST secret key partial that participates in the threshold signing process
- update withdrawer config/setup to support single signer (previous behaviour) or threshold signing via the `Signer` trait
- create `FrostSigner` which has gRPC clients for the signing participants
- `FrostSigner.sign()` performs the signing process by calling each participant for their commitment (part 1) and signature share (part 2) and finally aggregates the shares to create a valid signature.
- `FrostSigner.sign()` will fail if it does not receive responses from at least the minimum number of signers required. the min number of signers is determined during key generation, which is done completely separately.

## Testing
blackbox tests with mock signer servers have been added.

also, this was tested with a 2/2 threshold signer node setup. withdrawal transactions are successfully signed, submitted, and included on the sequencer. see https://www.notion.so/astria-org/bridge-threshold-withdrawer-testing-1936bd31a90c803ab33ccc1221f545a1?pvs=4

use the following testing keys which are for 2/2 threshold and represent astria address `astria1w2lxx9p02u7u934ljl060wannqm40g3y089lmh`

put the following keys into their own files.

pubkey package:

```bash
{
  "header": {
    "version": 0,
    "ciphersuite": "FROST-ED25519-SHA512-v1"
  },
  "verifying_shares": {
    "0100000000000000000000000000000000000000000000000000000000000000": "bd209bd0c47806fee2db29bde0b705f6a191602f0307e8d959d74f189ae92e9e",
    "0200000000000000000000000000000000000000000000000000000000000000": "0a677e1c6ad0c1906bf6ac53810b29efcb9ca9e7ab50fe9fca3ae49e613b922f"
  },
  "verifying_key": "ecbe5e2fbaaf23f14ca43f34d71ad733c57afcadeb5782809af48a3da17b9b69"
```

secret key 1:

```bash
{
  "header": {
    "version": 0,
    "ciphersuite": "FROST-ED25519-SHA512-v1"
  },
  "identifier": "0100000000000000000000000000000000000000000000000000000000000000",
  "signing_share": "4b45de055567d7a2027bfc9a2463de4aaa6f2df995180e638dc938013e9bf70e",
  "verifying_share": "bd209bd0c47806fee2db29bde0b705f6a191602f0307e8d959d74f189ae92e9e",
  "verifying_key": "ecbe5e2fbaaf23f14ca43f34d71ad733c57afcadeb5782809af48a3da17b9b69",
  "min_signers": 2
}
```

secret key 2:

```bash
{
  "header": {
    "version": 0,
    "ciphersuite": "FROST-ED25519-SHA512-v1"
  },
  "identifier": "0200000000000000000000000000000000000000000000000000000000000000",
  "signing_share": "29a52cd248df9a7bb2346fb4ed53318f0bff320985c8e49062c68e4fd0012d0f",
  "verifying_share": "0a677e1c6ad0c1906bf6ac53810b29efcb9ca9e7ab50fe9fca3ae49e613b922f",
  "verifying_key": "ecbe5e2fbaaf23f14ca43f34d71ad733c57afcadeb5782809af48a3da17b9b69",
  "min_signers": 2
}
```

checkout `noot/frost-participant` (see https://github.com/astriaorg/astria/pull/1960)

build astria-cli, astria-bridge-withdrawer, astria-bridge-signer and astria-sequencer.

run anvil (using this as fake rollup evm node for testing)

run astria-sequencer + cometbft as normal

transfer funds from funded address to withdrawer address

```bash
./target/debug/astria-cli sequencer transfer --amount 100000000 \
--private-key 2bd806c97f0e00af1a1fc3328fa763a9269723c8db8fac4f93af71db186d6e90 \
--sequencer-url http://localhost:26657 \
--sequencer.chain-id astria astria1w2lxx9p02u7u934ljl060wannqm40g3y089lmh
```

initialize bridge account on astria, with  `astria1w2lxx9p02u7u934ljl060wannqm40g3y089lmh` being the withdrawer address

```bash
./target/debug/astria-cli sequencer init-bridge-account \
--private-key 2bd806c97f0e00af1a1fc3328fa763a9269723c8db8fac4f93af71db186d6e90 \
--withdrawer-address astria1w2lxx9p02u7u934ljl060wannqm40g3y089lmh \
--sequencer-url http://localhost:26657 \
--sequencer.chain-id astria \
--rollup-name test
```

result:

```bash
sending tx from address: astria1rsxyjrcm255ds9euthjx6yc3vrjt9sxrm9cfgm
InitBridgeAccount completed!
Included in block: 1252
Rollup name: test
Rollup ID: n4bQgYhMfWWaL-qgxVrQFaO_TxsrC4Is0V1sFbDwCgg=
```

deploy contracts to anvil using `astria-bridge-contracts` 

- set `.env` and `source .env`

```bash
PRIVATE_KEY=0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80
RPC_URL="http://localhost:8545"
BASE_CHAIN_ASSET_PRECISION=9
BASE_CHAIN_BRIDGE_ADDRESS="astria1rsxyjrcm255ds9euthjx6yc3vrjt9sxrm9cfgm"
BASE_CHAIN_ASSET_DENOMINATION="nria"
```

- `forge script script/AstriaWithdrawer.s.sol:AstriaWithdrawerScript    --rpc-url $RPC_URL --broadcast --sig "deploy()" -vvvv`
- contract deployed at `0x5FbDB2315678afecb367f032d93F642f64180aa3` , update `.env` again

run two astria-bridge-signers, one with grpc port `9001` and first secret key package, second with grpc port `9002` and second secret key package.

start the bridge withdrawer, configured for 2/2 threshold (make sure to update pubkey package path):

```bash
ASTRIA_BRIDGE_WITHDRAWER_FROST_THRESHOLD_SIGNING_ENABLED=true
ASTRIA_BRIDGE_WITHDRAWER_FROST_MIN_SIGNERS=2
ASTRIA_BRIDGE_WITHDRAWER_FROST_PUBLIC_KEY_PACKAGE_PATH="pub_bridge.key"
ASTRIA_BRIDGE_WITHDRAWER_FROST_PARTICIPANT_ENDPOINTS="http://127.0.0.1:9001,http://127.0.0.1:9002"
ASTRIA_BRIDGE_WITHDRAWER_SEQUENCER_BRIDGE_ADDRESS="astria1rsxyjrcm255ds9euthjx6yc3vrjt9sxrm9cfgm"
ASTRIA_BRIDGE_WITHDRAWER_ETHEREUM_CONTRACT_ADDRESS="0x5FbDB2315678afecb367f032d93F642f64180aa3"

```

finally, send a withdrawal on the rollup:

```bash
forge script script/AstriaWithdrawer.s.sol:AstriaWithdrawerScript \
   --rpc-url $RPC_URL --broadcast --sig "withdrawToSequencer()" -vvvv
```

in the withdrawer, should see:

```bash
2025-02-10T17:33:57.417370Z  INFO process_batch: astria_bridge_withdrawer::bridge_withdrawer::submitter: withdraw batch successfully executed. sequencer.block=41 sequencer.tx_hash=69E3997A2A67CD4716D1051D9962219FBCEF9F3F7557C276D9F4D414D30166F2 rollup.height=4 batch.value=1000
```

## Changelogs
Changelogs updated.

## Related Issues

closes  #1937


crates/astria-bridge-withdrawer/local.env.example
```diff
@@ -31,10 +31,26 @@ ASTRIA_BRIDGE_WITHDRAWER_SEQUENCER_COMETBFT_ENDPOINT="http://127.0.0.1:26657"
 # Chain ID of the sequencer chain which transactions are submitted to.
 ASTRIA_BRIDGE_WITHDRAWER_SEQUENCER_CHAIN_ID="astria"
 
+# Set to true to enable frost threshold signing.
+ASTRIA_BRIDGE_WITHDRAWER_FROST_THRESHOLD_SIGNING_ENABLED=false
```
SuperFluffy, line 35: Convention is:

```suggestion
ASTRIA_BRIDGE_WITHDRAWER_NO_FROST_THRESHOLD_SIGNING=true
```
(also adjust all comments to follow this convention)

crates/astria-bridge-withdrawer/local.env.example
```diff
@@ -31,10 +31,26 @@ ASTRIA_BRIDGE_WITHDRAWER_SEQUENCER_COMETBFT_ENDPOINT="http://127.0.0.1:26657"
...
+# Only used if `frost_threshold_signing_enabled` is true.
+ASTRIA_BRIDGE_WITHDRAWER_FROST_MIN_SIGNERS=0
+
+# The path to the json-encoded frost public key package.
+# Only used if `frost_threshold_signing_enabled` is true.
+ASTRIA_BRIDGE_WITHDRAWER_FROST_PUBLIC_KEY_PACKAGE_PATH=""
+
+# The frost participant gRPC endpoints.
+# Only used if `frost_threshold_signing_enabled` is true.
+ASTRIA_BRIDGE_WITHDRAWER_FROST_PARTICIPANT_ENDPOINTS=[]
```
SuperFluffy, line 52: Avoid using `[]` in env vars because they interact with shell tests (started with `[`).

Maybe a comma-separated can be used to parse the individual URLs out? E.g. `https://url.one,https://url.two` - this relies on `,` being a reserved keyword for the domain part of URLs.

crates/astria-bridge-withdrawer/local.env.example
```diff
@@ -31,10 +31,26 @@ ASTRIA_BRIDGE_WITHDRAWER_SEQUENCER_COMETBFT_ENDPOINT="http://127.0.0.1:26657"
...
+# Only used if `frost_threshold_signing_enabled` is false.
 ASTRIA_BRIDGE_WITHDRAWER_SEQUENCER_KEY_PATH=/path/to/priv_sequencer_key.json
 
+# The minimum number of frost participants required to sign a transaction.
+# Only used if `frost_threshold_signing_enabled` is true.
+ASTRIA_BRIDGE_WITHDRAWER_FROST_MIN_SIGNERS=0
+
+# The path to the json-encoded frost public key package.
+# Only used if `frost_threshold_signing_enabled` is true.
+ASTRIA_BRIDGE_WITHDRAWER_FROST_PUBLIC_KEY_PACKAGE_PATH=""
```
SuperFluffy, line 48: Could we combine this with `ASTRIA_BRIDGE_WITHDRAWER_SEQUENCER_KEY_PATH`?

The naming is a bit off, but we could a) deprecate this config and b) provide an alias for the time being.

crates/astria-bridge-withdrawer/src/bridge_withdrawer/submitter/frost_signer.rs
SuperFluffy, line 1: For organization purposes - create a new module `signer` with submodules `frost` and `key` (maybe there's a better word than `key`...).

crates/astria-bridge-withdrawer/src/bridge_withdrawer/submitter/frost_signer.rs
```diff
@@ -0,0 +1,320 @@
...
+        Ok(transaction)
+    }
+
+    fn address(&self) -> &Address {
+        &self.address
+    }
+}
+
+impl FrostSigner {
+    async fn frost_part_1(
```
SuperFluffy, line 231: Is there a verb we can use here than `forst_part_1`? Maybe `perform_part_1`? Or `execute_part_1`?

crates/astria-bridge-withdrawer/src/bridge_withdrawer/submitter/frost_signer.rs
```diff
@@ -0,0 +1,320 @@
...
+        &self,
+    ) -> (
+        Vec<(Identifier, axum::body::Bytes, u32)>,
```
SuperFluffy, line 234: Can you give this a structured type as the return value? Both for the elements inside the `Vec` as well as the `(Vec, BTreeMap)` overall.

crates/astria-bridge-withdrawer/src/bridge_withdrawer/submitter/frost_signer.rs
```diff
@@ -0,0 +1,320 @@
...
+        BTreeMap<Identifier, round1::SigningCommitments>,
+    ) {
+        let stream = futures::stream::FuturesUnordered::new();
```
SuperFluffy, line 237: I am noticing that `id` is the element by which you are tracking failure as well as the return value. Maybe a tokio `JoinMap` is more straight forward here?

crates/astria-bridge-withdrawer/src/bridge_withdrawer/mod.rs
```diff
@@ -71,13 +76,20 @@ impl BridgeWithdrawer {
     /// # Errors
     ///
     /// - If the provided `api_addr` string cannot be parsed as a socket address.
-    pub fn new(cfg: Config, metrics: &'static Metrics) -> eyre::Result<(Self, ShutdownHandle)> {
+    pub async fn new(
```
SuperFluffy, line 79: I have a strong preference for this remaining sync (not async).

`BridgeWithdrawer` already has a setup phase during which async steps are run.

crates/astria-bridge-withdrawer/src/bridge_withdrawer/mod.rs
```diff
@@ -260,6 +281,46 @@ impl BridgeWithdrawer {
...
 
+async fn make_signer(
+    frost_threshold_signing_enabled: bool,
+    frost_min_signers: usize,
+    frost_public_key_package_path: String,
+    frost_participant_endpoints: Vec<String>,
+    sequencer_key_path: String,
+    sequencer_address_prefix: String,
+) -> eyre::Result<Arc<dyn Signer>> {
+    let signer: Arc<dyn Signer> = if frost_threshold_signing_enabled {
```
SuperFluffy, line 292: Noted elsewhere - I think making this an enum (static dispatch) rather than a trait (dynamic dispatch) might end up being cleaner and easier to understand.

crates/astria-bridge-withdrawer/src/bridge_withdrawer/mod.rs
```diff
@@ -260,6 +281,46 @@ impl BridgeWithdrawer {
...
+        let public_key_package_str = std::fs::read_to_string(frost_public_key_package_path)
+            .wrap_err("failed to read frost public key package")?;
+        let public_key_package =
+            serde_json::from_str::<frost_ed25519::keys::PublicKeyPackage>(&public_key_package_str)
+                .wrap_err("failed to deserialize public key package")?;
+
+        let participant_clients =
+            initialize_frost_participant_clients(frost_participant_endpoints, &public_key_package)
```
SuperFluffy, line 300: This should be done in the init/startup phase of bridge withdrawer.

proto/signerapis/astria/signer/v1/frost.proto
```diff
@@ -0,0 +1,44 @@
...
+    uint32 request_identifier = 3;
+}
+
+message Part2Response {
+    bytes signature_share = 1;
+}
+
+service FrostParticipantService {
+    rpc GetVerifyingShare(GetVerifyingShareRequest) returns (GetVerifyingShareResponse) {}
+    rpc Part1(Part1Request) returns (Part1Response) {}
```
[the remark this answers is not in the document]
noot: renamed

crates/astria-bridge-withdrawer/src/bridge_withdrawer/submitter/mod.rs
```diff
@@ -232,6 +238,39 @@ fn report_exit(reason: eyre::Result<&str>) {
     }
 }
 
+#[instrument(name = "sign_tx", skip_all, err)]
+async fn sign_tx(signer: Arc<dyn Signer>, body: TransactionBody) -> eyre::Result<Transaction> {
```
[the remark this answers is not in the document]
noot: no, will remove

crates/astria-bridge-withdrawer/src/bridge_withdrawer/mod.rs
```diff
@@ -260,6 +281,46 @@ impl BridgeWithdrawer {
     }
 }
 
+async fn make_signer(
```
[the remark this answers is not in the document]
noot: by transparent builder, do you mean the struct fields are public?

crates/astria-bridge-withdrawer/local.env.example
noot: don't think we should combine them, `ASTRIA_BRIDGE_WITHDRAWER_SEQUENCER_KEY_PATH` contains a private key while `ASTRIA_BRIDGE_WITHDRAWER_FROST_PUBLIC_KEY_PACKAGE_PATH` is the public key package (no secrets), so they're pretty different

crates/astria-bridge-withdrawer/src/bridge_withdrawer/submitter/frost_signer.rs
noot: `JoinMap` looks good, but i don't think i can `collect()` on it as it doesn't implement `Stream`? is there another way to collect the items as they're done, or do i just have to do `while map.join_next()`?

crates/astria-bridge-withdrawer/src/bridge_withdrawer/mod.rs
```diff
@@ -45,6 +50,7 @@ use self::{
 };
 use crate::{
     api,
+    bridge_withdrawer::submitter::make_signer,
```
SuperFluffy, line 53: If the Signer` is part of `Submitter` just provide all the values in `submitter::Builder` and let it delegate from there. Don't bubble it all the way if not needed.

crates/astria-bridge-withdrawer/src/main.rs
```diff
@@ -41,20 +62,62 @@ async fn main() -> ExitCode {
...
+        };
+
+        let frost_participant_endpoints = cfg
+            .frost_participant_endpoints
+            .split(',')
+            .map(str::to_string)
+            .collect();
+        let participant_clients = match initialize_frost_participant_clients(
+            frost_participant_endpoints,
+            &public_key_package,
```
SuperFluffy, line 100: I see what you did here. :D I should have specified my original comment better:

The parsing should happen inside the `signer::frost` module, ideally with a `FromStr for ParticipantEndpoints` (and with a simple unit test).

The initialization of the clients should happen when the run-loop is entered. I have originally suggested the `Startup` type, but that's wrong. It should be at the beginning of `Submitter::run` here (link to current main: https://github.com/astriaorg/astria/blob/a12efeb0e4000d8ac2adc4e70ced4954cfbbb94c/crates/astria-bridge-withdrawer/src/bridge_withdrawer/submitter/mod.rs#L79-L91

The idea is that if a shutdown signal is received we can abort initialization without having to wait 30 seconds.

crates/astria-bridge-withdrawer/src/main.rs
```diff
@@ -83,3 +146,44 @@ async fn main() -> ExitCode {
...
+        .wrap_err("failed to deserialize public key package")
+}
+
+async fn initialize_frost_participant_clients(
+    endpoints: Vec<String>,
+    public_key_package: &PublicKeyPackage,
+) -> eyre::Result<HashMap<Identifier, FrostParticipantServiceClient<tonic::transport::Channel>>> {
+    let mut participant_clients = HashMap::new();
+    for endpoint in endpoints {
+        let mut client = FrostParticipantServiceClient::connect(endpoint)
```
SuperFluffy, line 165: Note that initializing these can happen lazily/non-async by ussing `connect_lazy`.

crates/astria-bridge-withdrawer/tests/blackbox/helpers/mock_bridge_signer.rs
```diff
@@ -0,0 +1,147 @@
...
+        RoundTwoRequest,
+        RoundTwoResponse,
+    },
+};
+use astria_eyre::eyre::{
+    self,
+    WrapErr as _,
+};
+use frost_ed25519::round1;
+use rand::rngs::OsRng;
```
SuperFluffy, line 26: For tests avoid a completely random seed - you can replace this by `rand_chacha` and for example instantiate like ` let rng = rand_chacha::ChaChaRng::seed_from_u64(seed);` via the trait method `rand::SeedableRng::seed_from_u64`. We use it elsewhere in the codebase already.

crates/astria-bridge-withdrawer/src/bridge_withdrawer/submitter/signer/frost_signer.rs
```diff
@@ -0,0 +1,312 @@
...
+pub(crate) struct FrostSigner {
+    min_signers: usize,
+    public_key_package: PublicKeyPackage,
+    address: Address,
+    participant_clients:
+        HashMap<Identifier, FrostParticipantServiceClient<tonic::transport::Channel>>,
+}
+
+impl FrostSigner {
+    pub(crate) async fn sign(&self, tx: TransactionBody) -> eyre::Result<Transaction> {
```
SuperFluffy, line 138: I think `pub(super)` suffices here.

crates/astria-bridge-withdrawer/src/bridge_withdrawer/submitter/signer/frost_signer.rs
```diff
@@ -0,0 +1,312 @@
...
+        // part 1: gather commitments from participants
+        let round_1_results = self.execute_round_1().await;
+        ensure!(
+            round_1_results.responses.len() >= self.min_signers,
+            "not enough part 1 responses received; want at least {}, got {}",
```
SuperFluffy, line 143:
```suggestion
"not enough part 1 responses received; want at least `{}`, got `{}`",
```

crates/astria-bridge-withdrawer/src/bridge_withdrawer/submitter/signer/frost_signer.rs
```diff
@@ -0,0 +1,312 @@
...
+        );
+
+        // part 2: gather signature shares from participants
+        let tx_bytes = tx.to_raw().encode_to_vec();
+        let sig_shares = self
+            .execute_round_2(round_1_results.responses, tx_bytes.clone())
+            .await;
+        ensure!(
+            sig_shares.len() >= self.min_signers,
+            "not enough part 2 signature shares received; want at least {}, got {}",
```
SuperFluffy, line 155:
```suggestion
"not enough part 2 signature shares received; want at least `{}`, got `{}`",

File: crates/astria-bridge-withdrawer/src/bridge_withdrawer/submitter/signer/frost_signer.rs [hunk 20]
```diff
@@ -0,0 +1,312 @@
...
+impl FrostSigner {
+    async fn execute_round_1(&self) -> Round1Results {
+        let stream = futures::stream::FuturesUnordered::new();
+        for (id, client) in &self.participant_clients {
+            let mut client = client.clone();
+            stream.push(async move {
+                let resp = client
+                    .execute_round_one(RoundOneRequest {})
+                    .await
+                    .wrap_err(format!(
```
SuperFluffy, line 215: This should be `wrap_err_with(|| format!(...))` so that the string isn't constructed even if you don't have an error.

crates/astria-bridge-withdrawer/src/bridge_withdrawer/submitter/signer/frost_signer.rs
```diff
@@ -0,0 +1,312 @@
...
+                        "failed to get part 1 response for participant with id {id:?}"
+                    ))?;
+                Ok((id, resp.into_inner()))
+            });
+        }
+        let results: Vec<eyre::Result<_>> = stream.collect::<Vec<_>>().await;
```
SuperFluffy, line 221: I think collecting the results into a vec first is redundant. You could iterate over the stream directly, i.e. `while let Some(item) = your_stream.next().await`.

crates/astria-bridge-withdrawer/src/bridge_withdrawer/submitter/signer/frost_signer.rs
```diff
@@ -0,0 +1,312 @@
...
+        let mut signing_package_commitments: BTreeMap<Identifier, round1::SigningCommitments> =
+            BTreeMap::new();
+
+        let responses = results
+            .into_iter()
+            .filter_map(|res| match res {
+                Ok((id, part1)) => {
+                    let signing_commitment =
+                        round1::SigningCommitments::deserialize(&part1.commitment).ok()?;
```
SuperFluffy, line 230: This deserialization step could be moved into the async block so that

crates/astria-bridge-withdrawer/src/bridge_withdrawer/submitter/signer/frost_signer.rs
```diff
@@ -0,0 +1,312 @@
...
+
+struct Round1Response {
+    id: Identifier,
+    commitment: axum::body::Bytes,
+    request_identifier: u32,
+}
+
+impl FrostSigner {
+    async fn execute_round_1(&self) -> Round1Results {
+        let stream = futures::stream::FuturesUnordered::new();
```
SuperFluffy, line 208: Answering here: the `JoinMap` indeed seems not necessary; `FuturesUnordered` should be fine.

crates/astria-bridge-withdrawer/src/bridge_withdrawer/submitter/signer/frost_signer.rs
```diff
@@ -0,0 +1,312 @@
...
+                    let signing_commitment =
+                        round1::SigningCommitments::deserialize(&part1.commitment).ok()?;
+                    signing_package_commitments.insert(*id, signing_commitment);
+                    Some(Round1Response {
+                        id: *id,
+                        commitment: part1.commitment,
+                        request_identifier: part1.request_identifier,
+                    })
+                }
+                Err(_) => None,
```
SuperFluffy, line 238: You are silently dropping the error. Should this warn instead?

crates/astria-bridge-withdrawer/src/bridge_withdrawer/submitter/signer/frost_signer.rs
```diff
@@ -0,0 +1,312 @@
...
+                        commitments: request_commitments,
+                    })
+                    .await
+                    .wrap_err(format!(
+                        "failed to get part 2 response for participant with id {id:?}"
+                    ))?;
+                Ok((id, resp.into_inner()))
+            });
+        }
+        let results: Vec<eyre::Result<_>> = stream.collect::<Vec<_>>().await;
```
SuperFluffy, line 296: Same argument as before: I don't think this extra step is useful.

crates/astria-bridge-withdrawer/src/bridge_withdrawer/submitter/signer/frost_signer.rs
```diff
@@ -0,0 +1,312 @@
...
+        let sig_shares: BTreeMap<Identifier, frost_ed25519::round2::SignatureShare> = results
+            .into_iter()
+            .filter_map(|res| match res {
+                Ok((id, part2)) => {
+                    let sig_share =
+                        frost_ed25519::round2::SignatureShare::deserialize(&part2.signature_share)
+                            .ok()?;
+                    Some((id, sig_share))
+                }
+                Err(_) => None,
```
SuperFluffy, line 306: Silently dropping the error.

crates/astria-bridge-withdrawer/src/bridge_withdrawer/submitter/signer/frost_signer.rs
```diff
@@ -0,0 +1,312 @@
...
+        for Round1Response {
+            id,
+            request_identifier,
+            ..
+        } in responses
+        {
+            let mut client = self
+                .participant_clients
+                .get(&id)
+                .expect(
```
SuperFluffy, line 275: I mean - yeah, that doesn't seem wrong if you know exactly what round-1 does, but I think we should add a `wrap_err` and slap a big `BUG` in the message.

crates/astria-bridge-withdrawer/src/bridge_withdrawer/submitter/signer/frost_signer.rs
```diff
@@ -0,0 +1,312 @@
...
+                Ok((id, resp.into_inner()))
+            });
+        }
+        let results: Vec<eyre::Result<_>> = stream.collect::<Vec<_>>().await;
+        let sig_shares: BTreeMap<Identifier, frost_ed25519::round2::SignatureShare> = results
+            .into_iter()
+            .filter_map(|res| match res {
+                Ok((id, part2)) => {
+                    let sig_share =
+                        frost_ed25519::round2::SignatureShare::deserialize(&part2.signature_share)
```
SuperFluffy, line 302: Move this into the async block?

crates/astria-bridge-withdrawer/src/bridge_withdrawer/submitter/signer/frost_signer.rs
```diff
@@ -0,0 +1,312 @@
...
+                .participant_clients
+                .get(&id)
+                .expect(
+                    "participant client must exist in mapping, as we received a commitment from \
+                     them in part 1, meaning we already have their client",
+                )
+                .clone();
+            let request_commitments = request_commitments.clone();
+            let tx_bytes = tx_bytes.clone();
+            stream.push(async move {
```
SuperFluffy, line 282: I think this async block could be put into a free standing function + instrumentation (and add a `err(Display)` so that you emit an error if somerthing goes wrong so you don't drop it quietly).

crates/astria-bridge-withdrawer/src/bridge_withdrawer/submitter/signer/frost_signer.rs
```diff
@@ -0,0 +1,312 @@
...
+                let resp = client
+                    .execute_round_two(RoundTwoRequest {
+                        request_identifier,
+                        message: tx_bytes.into(),
+                        commitments: request_commitments,
+                    })
+                    .await
+                    .wrap_err(format!(
```
SuperFluffy, line 290: `wrap_err_with`

crates/astria-bridge-withdrawer/src/bridge_withdrawer/submitter/signer/frost_signer.rs
```diff
@@ -0,0 +1,312 @@
...
+    commitment: axum::body::Bytes,
+    request_identifier: u32,
+}
+
+impl FrostSigner {
+    async fn execute_round_1(&self) -> Round1Results {
+        let stream = futures::stream::FuturesUnordered::new();
+        for (id, client) in &self.participant_clients {
+            let mut client = client.clone();
+            stream.push(async move {
```
SuperFluffy, line 211: Put this async block into a freestanding function (+ instrumentation and `err(Display`) to emit an error if something goes wrong).

quasystaty1 (commented):
Generally infra LGTM, Is there a straightforward way to add a smoke test for this? I believe we would have to first include a Helm chart for the new signer binary in #1960

SuperFluffy (approved):

SuperFluffy:
Before merge, please copy the information for testing the 2/2 setup from notion into the PR text (this doc you linked https://www.notion.so/astria-org/bridge-threshold-withdrawer-testing-1936bd31a90c803ab33ccc1221f545a1 ), so that it's not hidden behind an auth wall.

quasystaty1 (changes requested):
Need to format and mount the new file expected in `ASTRIA_BRIDGE_WITHDRAWER_FROST_PUBLIC_KEY_PACKAGE_PATH`. Otherwise, small nits.

charts/evm-bridge-withdrawer/values.yaml
```diff
@@ -28,6 +28,10 @@ config:
   rollupAssetDenom: ""
   evmContractAddress: "0x"
   evmRpcEndpoint: ""
+  noFrostThresholdSigning: "true"
```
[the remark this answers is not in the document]
noot: conflicts with convention unfortunately: https://github.com/astriaorg/astria/pull/1948#discussion_r1975222272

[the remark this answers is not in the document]
quasystaty1: The suggested change is in the chart and not in actual crate.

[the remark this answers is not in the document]
SuperFluffy: Well,  *technically* the convention is for the env vars, not for the charts, so @quasystaty1's suggestion would work (and I think it's what is being done elsewhere in the charts).

quasystaty1 (approved):
Infra approval. infra requested changes will be part of #1960 , running in threshold mode is not applicable at the moment.

PR #1953, opened 2025-02-06
feat(auctioneer): add metrics

### LINKED ISSUE #1915: add metrics to auctioneer
Following https://github.com/astriaorg/astria/pull/1839 metrics need to be added to auctioneer. Low hanging fruit / high impact metrics might be:

+ bids received for a given auction
+ bids ignored (for example, because the timer was up)
+ successful auction winner submissions
+ failed auction winner submissions

┆Issue Number: ENG-1071

SuperFluffy:
## Summary
Extend Auctioneer by a first set of metrics.

## Background
Astria services must be observable, metrics are low hanging fruit.

## Changes
- Added metrics (see metrics section below for a list).

## Testing
This needs to be observed on a telemetry platform.

## Changelogs
Changelogs updated.

## Metrics

- astria_auctioneer_executed_blocks_received: counter
- astria_auctioneer_optimistic_blocks_received: coubter
- astria_auctioneer_auctions_cancelled: coubter
- astria_auctioneer_auctions_submitted: counter
- astria_auctioneer_auction_bids_processed: gauge (labels "processed" and "dropped")
- astria_auctioneer_auction_bids_received: counter

## Related Issues
closes #1915

### DIFF
--- a/crates/astria-auctioneer/src/auctioneer/mod.rs
+++ b/crates/astria-auctioneer/src/auctioneer/mod.rs
@@ -52,6 +52,7 @@ pub(super) struct Auctioneer {
     block_commitments: BlockCommitmentStream,
     bids: BidStream,
     cancelled_auctions: FuturesUnordered<auction::Auction>,
+    metrics: &'static crate::Metrics,
     executed_blocks: ExecuteOptimisticBlockStream,
     running_auction: Option<auction::Auction>,
     optimistic_blocks: OptimisticBlockStream,
@@ -61,7 +62,11 @@ pub(super) struct Auctioneer {
 
 impl Auctioneer {
     /// Creates an [`Auctioneer`] service from a [`Config`].
-    pub(super) fn new(config: Config, shutdown_token: CancellationToken) -> eyre::Result<Self> {
+    pub(super) fn new(
+        config: Config,
+        metrics: &'static crate::Metrics,
+        shutdown_token: CancellationToken,
+    ) -> eyre::Result<Self> {
         let Config {
             sequencer_grpc_endpoint,
             sequencer_abci_endpoint,
@@ -100,6 +105,7 @@ impl Auctioneer {
             rollup_id,
             cancellation_token: shutdown_token.child_token(),
             last_successful_nonce: None,
+            metrics,
         };
 
         Ok(Self {
@@ -108,6 +114,7 @@ impl Auctioneer {
             bids: rollup_channel.open_bid_stream(),
             cancelled_auctions: FuturesUnordered::new(),
             executed_blocks: rollup_channel.open_execute_optimistic_block_stream(),
+            metrics,
             optimistic_blocks: sequencer_channel.open_get_optimistic_block_stream(rollup_id),
             rollup_id,
             running_auction: None,
@@ -187,9 +194,9 @@ impl Auctioneer {
             nonce_used, ..
         }) = &res
         {
+            self.metrics.increment_auctions_submitted_count();
             self.auction_factory.set_last_successful_nonce(*nonce_used);
         }
-
         let _ = self.running_auction.take();
         res
     }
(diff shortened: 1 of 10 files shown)


crates/astria-auctioneer/src/metrics.rs
```diff
@@ -1,25 +1,153 @@
...
     }
 }
 
 metric_names!(const METRICS_NAMES:
+    BLOCK_COMMITMENTS_RECEIVED,
+    EXECUTED_BLOCKS_RECEIVED,
+    OPTIMISTIC_BLOCKS_RECEIVED,
+    AUCTIONS_CANCELLED,
+    AUCTIONS_SUBMITTED,
+    AUCTION_BIDS_PROCESSED,
```
SuperFluffy, line 151: I am not happy with this name - I want to have a gauge with 2 labels: bids admitted and bids dropped. "processed" seems confusing. Are there better ideas?

crates/astria-auctioneer/src/auctioneer/mod.rs
```diff
@@ -272,6 +283,9 @@ impl Auctioneer {
             "block_hash",
             field::display(executed_block.sequencer_block_hash()),
         );
+
+        self.metrics.increment_executed_blocks_received_counter();
```
bharath-123, line 287: this can act as a proxy for the metric "number of auctions conducted" since an auction starts when the auctioneer receives an executed block.

crates/astria-auctioneer/src/metrics.rs
bharath-123: lets make the bids received and dropped a histogram.

crates/astria-auctioneer/src/metrics.rs
```diff
@@ -1,25 +1,153 @@
...
+        Counter,
+        Gauge,
+        IntoF64,
         RegisteringBuilder,
     },
 };
 
-pub struct Metrics {}
+const AUCTION_BIDS_PROCESSED_LABEL: &str = "auction_bids_processed";
+const AUCTION_BIDS_PROCESSED_ADMITTED: &str = "admitted";
```
bharath-123, line 13: i feel like received is a better word than admitted. there is no entry barrier for the bids to gain admission, received sounds like a better word to me at least.

bharath-123:
Some metrics to add and feedback:
1. A histogram of the winning bid amounts in the auction. It will allow us to understand how much people are bidding
2. update bids_admitted and bids_dropped to be a histogram so that we can visualize how many bids we get per auction. it is useful to assess demand for our TOB.

crates/astria-auctioneer/src/metrics.rs
SuperFluffy: The distinction is important actually (received is processed + dropped), but I did change the nomenclature in 27af6c11 (now the histogram is just "auction_bids" with labels ["processed", "dropped"]).

crates/astria-auctioneer/src/metrics.rs
SuperFluffy: See my other comment in https://github.com/astriaorg/astria/pull/1953#discussion_r1946490223, copied here:

> The distinction is important actually (received is processed + dropped), but I did change the nomenclature in 27af6c11 (now the histogram is just "auction_bids" with labels ["processed", "dropped"]).

crates/astria-auctioneer/src/metrics.rs
```diff
@@ -1,25 +1,220 @@
...
+
+    pub(crate) fn record_auction_bid_delay_since_start(&self, val: impl IntoF64) {
+        self.auction_bid_delay_since_start.record(val);
+    }
+
+    pub(crate) fn record_auction_winning_bid_histogram(&self, val: impl IntoF64) {
+        self.auction_winning_bid_histogram.record(val);
+    }
+
+    pub(crate) fn record_auction_winner_submission_error_latency(&self, val: impl IntoF64) {
```
bharath-123, line 81: we also might want the number of times submission has error'ed out. We could use this for an alert to indicate when auctioneer has a high submission error rate which is something actionable?

SuperFluffy: Isn't that covered by the same histogram with the "error" label (see the method just below this one)?

bharath-123: yes, the histogram metric will return a total count too. so that should cover it!

PR #1954, opened 2025-02-06
fix(ci): tag images with full commit sha

aajimal:
## Summary
Adds an additional tag to docker images corresponding to the full commit hash.  

## Background
Manually running github actions to build a docker image was not setting the full sha hash, which argocd uses to pull images. This should always set at least the full hash to the merge commit of a PR. 

## Changes
Updated the docker metadata step to include an additional tag

## Testing
Manually with PR commit and triggering GH action. 

## Changelogs
 "No updates required."


warhod (approved):
LGTM

PR #1955, opened 2025-02-06
fix(sequencer-cli): poll TX status to avoid looping indefinitely

### LINKED ISSUE #1949: bug(sequencer-client): client loops forever if transaction execution fails in sequencer

ethanoroshiba:
## Summary
Added a transaction status ABCI query to the sequencer and changed the Sequencer Client to poll that instead of CometBFT's `tx` RPC.

## Background
The Sequencer Client's method `wait_for_tx_inclusion` previously polled CometBFT's `tx` endpoint, which would return an error no matter whether the transaction was pending, failed, or never submitted in the first place. Because of this lack of distinction, the method would poll indefinitely if a transaction failed for some reason. This is similar to the issue mentioned in #1936.

## Changes
- Add transaction status ABCI query to the sequencer.
- Added `get_transaction_status` method to mempool, returning either `Pending`, `Parked`, `RemovalCache`, or `Unknown`.
- Added protos and native types for `TransactionStatus` and `TransactionStatusResponse`.
- Renamed `wait_for_tx_inclusion` to `confirm_tx_inclusion`, and changed the logic to continue looping after receiving a `Pending` or `Parked` response. If `Unknown`, it will poll until `MAX_WAIT_TIME_UNKNOWN` has elapsed, and will then return an error. If it receives `RemovalCache`, it will try CometBFT's `tx` RPC for a short time before erroring if there is no response.
- Changed reliant bits in the Bridge Withdrawer and Sequencer CLI to use the updated method.

## Testing
- Added unit tests for mempool's `get_transaction_status` method.
- Added unit tests for Sequencer transaction status ABCI query.
- Added unit tests for Sequencer Client's `confirm_tx_inclusion` method.

## Changelogs

## Breaking Changelist
- Not breaking, but worth noting this adds to sequencer's public API.

## Related Issues
closes #1949

### DIFF
--- a/crates/astria-core/src/protocol/transaction/v1/mod.rs
+++ b/crates/astria-core/src/protocol/transaction/v1/mod.rs
@@ -566,6 +566,111 @@ impl TransactionParams {
     }
 }
 
+#[derive(Debug, thiserror::Error)]
+#[error(transparent)]
+pub struct TransactionStatusError(TransactionStatusErrorKind);
+
+impl TransactionStatusError {
+    #[must_use]
+    pub fn invalid_status(val: i32) -> Self {
+        Self(TransactionStatusErrorKind::InvalidStatus(val))
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+enum TransactionStatusErrorKind {
+    #[error("transaction status {0}i32 is invalid (valid values 1-4)")]
+    InvalidStatus(i32),
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum TransactionStatus {
+    Parked,
+    Pending,
+    RemovalCache,
+    Unknown,
+}
+
+impl TryFrom<i32> for TransactionStatus {
+    type Error = TransactionStatusError;
+
+    fn try_from(value: i32) -> Result<Self, Self::Error> {
+        match value {
+            1 => Ok(Self::Parked),
+            2 => Ok(Self::Pending),
+            3 => Ok(Self::RemovalCache),
+            4 => Ok(Self::Unknown),
+            _ => Err(TransactionStatusError::invalid_status(value)),
+        }
+    }
+}
+
+impl From<&TransactionStatus> for i32 {
+    fn from(status: &TransactionStatus) -> i32 {
+        match status {
+            TransactionStatus::Parked => 1,
+            TransactionStatus::Pending => 2,
+            TransactionStatus::RemovalCache => 3,
+            TransactionStatus::Unknown => 4,
+        }
+    }
+}
+
+impl std::fmt::Display for TransactionStatus {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            TransactionStatus::Parked => write!(f, "PARKED"),
+            TransactionStatus::Pending => write!(f, "PENDING"),
+            TransactionStatus::RemovalCache => write!(f, "REMOVAL_CACHE"),
+            TransactionStatus::Unknown => write!(f, "UNKNOWN"),
+        }
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+#[error(transparent)]
+pub struct TransactionStatusResponseError(TransactionStatusResponseErrorKind);
+
+impl TransactionStatusResponseError {
+    #[must_use]
+    pub fn invalid_status(source: TransactionStatusError) -> Self {
+        Self(TransactionStatusResponseErrorKind::InvalidStatus(source))
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+enum TransactionStatusResponseErrorKind {
+    #[error("transaction status was invalid")]
+    InvalidStatus(TransactionStatusError),
+}
+
+#[derive(Debug)]
+pub struct TransactionStatusResponse {
+    pub status: TransactionStatus,
+}
+
+impl Protobuf for TransactionStatusResponse {
+    type Error = TransactionStatusResponseError;
+    type Raw = raw::TransactionStatusResponse;
+
+    fn try_from_raw_ref(raw: &Self::Raw) -> Result<Self, Self::Error> {
+        let raw::TransactionStatusResponse {
+            status,
+        } = raw;
+        let status = TransactionStatus::try_from(*status)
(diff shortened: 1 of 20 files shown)


proto/protocolapis/astria/protocol/transaction/v1/types.proto
```diff
@@ -0,0 +1,16 @@
...
+  TransactionStatus status = 1;
+}
+
+enum TransactionStatus {
+  TRANSACTION_STATUS_UNSPECIFIED = 0;
+  TRANSACTION_STATUS_PARKED = 1;
+  TRANSACTION_STATUS_PENDING = 2;
+  TRANSACTION_STATUS_REMOVAL_CACHE = 3;
+  TRANSACTION_STATUS_UNKNOWN = 4;
+}
```
ethanoroshiba, line 16: I'm not sure whether this is useful. It's represented as an i32 in Rust anyways, and the domain `TransactionStatus` conversion is done directly from i32. Could be nice for the API, though.

crates/astria-core/src/protocol/transaction/v1/mod.rs
```diff
@@ -566,6 +566,100 @@ impl TransactionParams {
...
+    #[error("transaction status {0}i32 is invalid (valid values 1-4)")]
+    InvalidStatus(i32),
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum TransactionStatus {
+    Parked,
+    Pending,
+    RemovalCache,
+    Unknown,
```
noot, line 591: does this mean the tx wasn't found? if so, would prefer `NOT_FOUND` as it's more specific

crates/astria-sequencer-client/src/extension_trait.rs
```diff
@@ -705,16 +755,71 @@ pub trait SequencerClientExt: Client {
...
+                TransactionStatus::RemovalCache => break info,
+            };
+            sleep_millis = (sleep_millis.saturating_mul(2)).min(MAX_POLL_INTERVAL_MILLIS);
+        };
+
+        // Note: using fixed backoff here. We expect the transaction to be confirmed quickly, this
+        // is just to account for short delays or network issues. Logic for num retries is as
+        // follows: retries = MAX_WAIT_TIME_MILLIS_AFTER_REMOVAL / fixed_backoff(100ms)
+        let retry_config =
+            tryhard::RetryFutureConfig::new(MAX_WAIT_TIME_MILLIS_AFTER_REMOVAL / 100)
```
noot, line 810: replace `(MAX_WAIT_TIME_MILLIS_AFTER_REMOVAL / 100)` with a constant `MAX_RETRIES`?

crates/astria-sequencer/src/mempool/mod.rs
```diff
@@ -52,6 +59,23 @@ pub(crate) enum RemovalReason {
...
 }
 
+impl std::fmt::Display for RemovalReason {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            RemovalReason::Expired => write!(f, "transaction expired after 240 seconds"),
+            RemovalReason::NonceStale => write!(f, "transaction nonce is lower than current nonce"),
+            RemovalReason::LowerNonceInvalidated => write!(
+                f,
+                "previous transaction was not executed, to this transaction's nonce has become \
```
noot, line 69:
```suggestion
                "previous transaction was not executed, so this transaction's nonce has become \
```

crates/astria-core/src/protocol/transaction/v1/mod.rs
ethanoroshiba: Yes, I've changed the name in cdb152e

crates/astria-sequencer-client/src/extension_trait.rs
ethanoroshiba: Done in cdb152e.

crates/astria-sequencer/src/mempool/mod.rs
ethanoroshiba: Done in cdb152e. Thanks for reviewing!

crates/astria-sequencer/src/mempool/query.rs
```diff
@@ -0,0 +1,364 @@
...
+    else {
+        return Err(response::Query {
+            code: Code::Err(AbciErrorCode::INVALID_PARAMETER.value()),
+            info: "missing required parameter `tx_id`".into(),
+            ..response::Query::default()
+        });
+    };
+
+    // Support both "0x" prefixed and unprefixed transaction hashes
+    let tx_id = tx_id.trim_start_matches("0x");
```
Fraser999, line 98: Nice.  But it might be even nicer if we implemented `FromStr` for `TransactionId` so it could be used in other places?  Then here we'd just be calling `parse()`.

crates/astria-sequencer/src/mempool/query.rs
```diff
@@ -0,0 +1,364 @@
...
+        Err(err) => {
+            return response::Query {
+                code: Code::Err(AbciErrorCode::INTERNAL_ERROR.value()),
+                info: "failed converting `u64` block height to type `tendermint::block::Height`"
+                    .into(),
+                log: format!("{err:?}"),
+                ..response::Query::default()
+            };
+        }
+    };
```
Fraser999, line 61: I wonder if we need to provide this or if we can just use height 0 or 1 (the default).

Fraser999 (changes requested):
As discussed offline, I think we need to make the locking in the mempool more coarse-grained in order to allow the new query to give reliable results.  Basically, the four collections comprising the mempool need to be updated atomically.

crates/astria-sequencer/src/mempool/query.rs
ethanoroshiba: Good catch! Done in 70ba764

crates/astria-sequencer/src/mempool/mod.rs
Fraser999: Still not handled by the looks of it :D

ethanoroshiba:
After offline discussion, the decision was made to close this in favor of a timeout approach in #2048. This is because we firstly want to move towards gRPC instead of ABCI queries, and secondly since this approach will pretty much only return correct transaction statuses when querying a validator node.

PR #1957, opened 2025-02-10
fix(charts): geth flags

quasystaty1:
## Summary
changes to how flags condition is evaluated and used.
## Background
Our current evm-rollup charts has a bug, the condition which helps configuring geth flags is not being evaluated and always acts as `true`. This PR changes the condition templating. 
## Changes
- conditions are now being templated based on the key value. Also makes sure this value is `true` if the condition exists.

## Testing
- locally running smoke test against a cluster 
- using the helm command for templating: ` helm template my-chart ./charts/evm-stack -f ./dev/values/rollup/dev.yaml`
## Changelogs
No updates required.

### DIFF
--- a/charts/evm-rollup/templates/statefulsets.yaml
+++ b/charts/evm-rollup/templates/statefulsets.yaml
@@ -67,11 +67,8 @@ spec:
           command: [ "geth" ]
           args:
             {{- range $arg := .Values.config.geth.flags }}
-            {{- if hasKey $arg "condition" }}
-            {{- if eq (tpl $arg.condition $) "true" }}
-            - --{{ $arg.name }}{{ if $arg.value }}={{ tpl $arg.value $ }}{{ end }}
-            {{- end }}
-            {{- else }}
+            {{- $noCondition := not (hasKey $arg "condition") }}
+            {{- if or ($noCondition) (eq (tpl $arg.condition $) "true") }}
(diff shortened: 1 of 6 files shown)


charts/evm-rollup/templates/statefulsets.yaml
```diff
@@ -67,7 +67,11 @@ spec:
           command: [ "geth" ]
           args:
             {{- range $arg := .Values.config.geth.flags }}
-            {{- if $arg.condition | default true }}
+            {{- if hasKey $arg "condition" }}
+            {{- if eq (tpl $arg.condition $) "true" }}
+            - --{{ $arg.name }}{{ if $arg.value }}={{ tpl $arg.value $ }}{{ end }}
+            {{- end }}
+            {{- else }}
```
joroshiba, line 74: We can probably keep this duplication to a minimum, can do same in the exec script?

If there is no condition set or the condition evaluates to true

```suggestion
            {{- $noCondition := not (hasKey $arg "condition") }}
            {{- if or ($noCondition) (eq (tpl $arg.condition $) "true") }}
```

quasystaty1: Updated both statefulset and exec script with the new format, tested with and without metrics enabled using:
` helm template my-chart ./charts/evm-stack -f ./dev/values/rollup/dev.yaml`

PR #1966, opened 2025-02-12
chore(conductor): use build patterns instead of macros for mounting Mocks

### LINKED ISSUE #1927: replace blackbox testing macros with build patterns
The blackbox testing macros for Conductor have become unwieldy, especially as we've added more optional fields and tried to scale them. Additionally, they're difficult to learn and to work with when creating new blackbox tests or changing existing ones. We should aim to replace these with some sort of build pattern instead. These changes should be done gradually, moving any functionality that can be out of the macros and into `TestConductor` before switching all of the tests over.

┆Issue Number: ENG-1078

ethanoroshiba:
## Summary
Mostly (1 exception) switched from using macros to mount Mocks to using build patterns.

## Background
Scaling the need for differing patterns based on mock use cases was becoming unwieldy and hard to read/understand. For example, if we had the need for a mock to use a delay in one test, and expect a certain number of calls in another, then use both of these functionalities in another, we would need to create 3 new patterns to match on to provide each of these functionalities. Opting for a build pattern allows us to abstract these methods onto builders for open-ended scaling potential and cleaner code.

## Changes
- Created `GrpcMockBuilder` and `JsonRpcMockBuilder`, which are returned by all `mock_*` methods (except `mount_celestia_blob_get_all` on `TestConductor` and contain all necessary information for mounting the mock. These also implement typical methods used with mocks, such as `expect()`, `up_to_n_times()`, etc. The reason behind not returning a `Mock` itself is that these structs can contain information about which server they are to be mounted to, so that when writing tests the developer doesn't need to specify this (similar to how the macros functioned). Additionally, responding with a delay requires putting the delay on the responder before the `Mock` is constructed, so we mustn't construct the `Mock` immediately.
- Changed all `mount_*` methods to `mock_*` to accurately reflect what they are returning.
- Changed all tests to utilize these new methods.

## Testing
Passing all tests

## Changelogs
No updates required.

## Related Issues
closes #1927

### DIFF
--- a/crates/astria-conductor/tests/blackbox/firm_only.rs
+++ b/crates/astria-conductor/tests/blackbox/firm_only.rs
@@ -35,52 +35,47 @@ use crate::{
         MockGrpc,
         CELESTIA_BEARER_TOKEN,
         CELESTIA_CHAIN_ID,
-        SEQUENCER_CHAIN_ID,
     },
     mount_celestia_blobs,
-    mount_celestia_header_network_head,
-    mount_executed_block,
-    mount_get_commitment_state,
-    mount_get_genesis_info,
-    mount_sequencer_commit,
-    mount_sequencer_genesis,
-    mount_sequencer_validator_set,
-    mount_update_commitment_state,
 };
 
 #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
 async fn simple() {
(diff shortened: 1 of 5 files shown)


ethanoroshiba:
After offline discussion, we've decided to move towards a) using some sort of build pattern for the tests themselves, as opposed to just for the mounts, and b) making more gradual changes.

PR #1972, opened 2025-02-14
fix(charts): Update auctioneer image tags and flame devtag

bharath-123:
## Summary
This PR updates the image tag of the auctioneer in the auctioneer chart to use the latest sha post merge https://github.com/orgs/astriaorg/packages/container/auctioneer/355800545?tag=sha-2db84ed. 
We also update the devtag of the flame image to use the `latest` image.

## Background
We were using images built from a PR for auctioneer before merging it to main. Now we can start using the latest sha tags built from the auctioneer code in main branch.

## Changes
- Update the auctioneer image tag
- Update the flame devtag

## Testing
Deployed locally and tested it

closes

### DIFF
--- a/charts/auctioneer/Chart.yaml
+++ b/charts/auctioneer/Chart.yaml
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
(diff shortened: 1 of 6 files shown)


quasystaty1 (changes requested):
The PR is outdated. I think best approach here is to first create a release for flame and auctioneer and only then update these values.

joroshiba:
Closing as we cut releases and such already.

PR #1973, opened 2025-02-14
chore(conductor): remove unnecessary information from test mounts

ethanoroshiba:
## Summary
Removes unnecessary block hashes from test mounts by maintaining current block hashes in `TestConductor`.

## Background
Conductor tests have gotten unruly with the number and specificity of mounts needed to make them function properly. This is meant to make the tests more readable, simpler, and easier to write.

## Changes
- Added `TestRollupState` to `TestConductor`, which keeps track of the current block hash initializers (the repeated number in our test hashes) as well as the current firm block number (in case soft is updated but not firm).
- Added logic to all mounts which use block hashes such that they need not be supplied with them, but instead access and increment them automatically.

## Testing
All tests are passing.

## Changelogs
No updates requried.

## Related Issues
part of #1927

### DIFF
--- a/crates/astria-conductor/tests/blackbox/helpers/macros.rs
+++ b/crates/astria-conductor/tests/blackbox/helpers/macros.rs
@@ -170,44 +170,29 @@ macro_rules! mount_celestia_header_network_head {
 #[macro_export]
 macro_rules! mount_get_commitment_state {
     (
-        $test_env:ident,
-        firm: ( number: $firm_number:expr, hash: $firm_hash:expr, parent: $firm_parent:expr$(,)? ),
-        soft: ( number: $soft_number:expr, hash: $soft_hash:expr, parent: $soft_parent:expr$(,)? ),
-        base_celestia_height: $base_celestia_height:expr
-        $(,)?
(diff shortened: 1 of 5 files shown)


crates/astria-conductor/tests/blackbox/helpers/macros.rs
```diff
@@ -170,90 +170,98 @@ macro_rules! mount_celestia_header_network_head {
...
+                false,
+                $base_celestia_height,
+                $expected_calls,
+        )
+        .await
+    };
+}
+
+#[macro_export]
+macro_rules! mount_firm_update_commitment_state {
```
ethanoroshiba, line 236: I split this up for readability in tests, but I'm not sure if this makes it too opaque to a reader without context. Happy to change this back.

joroshiba:
closed in 7 days.

joroshiba:
Conflicts on stale PR, is it still relevant and need updated or now obsolete?

ethanoroshiba:
@joroshiba would still be nice to have, but will take a bit of adjustment with the v2 changes. Will try to get conflicts addressed ASAP

Issue #1971, opened 2025-02-14
feat(auctioneer): Add a smoke test for auctioneer


bharath-123:
It would be great to have a smoke test for the auctioneer binary which can help us detect any bugs early in future development processes. A smoke test would involve spinning up a sequencer network with optimistic block apis enabled, an auctioneer node and a flame node. 

We need to submit a transaction to the flame node and ensure that the transaction finalizes. We would need to send the transaction to the auctioneer flame node in time so that the transaction is considered for the auction by the auctioneer node and is submitted to the sequencer i.e we need a reliable way of sending transactions to the flame node since if a transaction doesn't finalize, we cannot distinguish it from a failure in the system or whether the transaction has been submitted too late into a block.

We can take inspiration from https://github.com/astriaorg/auctioneer-searcher-simulation to reliably submit a transaction.

joroshiba:
This issue is stale because it has been open 45 days with no activity. Remove stale label or this issue
be closed in 7 days.

joroshiba:
This issue was closed because it was stale

PR #1975, opened 2025-02-16
feat(auctioneer): POC of an auctioneer smoke test

bharath-123:
## Summary
We add a smoke test to test the auctioneer setup e2e. 

## Background
The setup for auctioneer is complex as there are multiple components and moving parts involved. Having a smoke test can help us guarantee that things don't break early on while developing it further. 
To be able to write a reliable smoke test for auctioneer, we need to have a way to reliably guarantee inclusion. Since if we submit a bid too late into a block, it can miss making it to the next sequencer block. Note that this is not a bug in the system.
In the current design of auctioneer, if the bid doesn't make it into the sequencer block or if there is a bug in the read/write paths of auctioneer then in both cases, we only observe an unfinalized transaction. We cannot distinguish between a bid submitted too late or a bug in the pipeline. 
In order to achieve this, we use the code in https://github.com/astriaorg/auctioneer-searcher-simulation/tree/bharath/smoke-test-strip-down to run a smoke test. On a high level, we listen to the optimistic block and when we receive an optimistic block, we submit the bid. We submit the bid as early as possible to ensure reliably inclusion. 
In the smoke test, we submit 5 bids and the smoke test only passes when 3/5 bids successfully make it in a block.

## Changes
- Add auctioneer-test.just file with commands to deploy and run the auctioneer smoke test
- Add dev/values/rollup/auctioneer-test.yaml to have flame configs specific to auctioneer. 
- Add a new sequencer public key to the sequencer genesis to have a funded key for auctioneer. 

## Testing
Locally deploying and running the smoke test.

## Related Issues
closes https://github.com/astriaorg/astria/issues/1971

### DIFF
--- a/charts/auctioneer-test.just
+++ b/charts/auctioneer-test.just
@@ -1,0 +1,124 @@
+_default:
+  @just --list auctioneer-test
+
+defaultTag := ""
+defaultNamespace := "astria-dev-cluster"
+evm_destination_address := "0x507056D8174aC4fb40Ec51578d18F853dFe000B2"
+smoke_test_image := "bharath123456/auctioneer-smoke-test:latest"
+sequencer_grpc_url := "grpc.sequencer.localdev.me"
+eth_rpc_url := "http://executor.astria.localdev.me"
+searcher_private_key := "5478c1df1db4867e4d29ab1cf6c6ed0ae0b2c3fbe7fc6e00edd0891d507d43f2"
+rollup_name := "astria"
+amount_to_send := "10000"
+
+delete:
+  -just delete celestia-local
+  -just delete sequencers
+  -just delete rollup
+
+@deploy tag=defaultTag:
+  echo "Deploying ingress controller..." && just deploy-ingress-controller > /dev/null
+  just wait-for-ingress-controller > /dev/null
+  echo "Deploying local celestia instance..."
+  helm install celestia-local-chart ./celestia-local --namespace {{defaultNamespace}} --set fast=true --create-namespace  > /dev/null
+  helm dependency update ./sequencer > /dev/null
+  helm dependency update ./evm-stack > /dev/null
+
+  echo "Setting up 3 astria sequencer nodes"
+  echo "Setting up the first astria sequencers..." && helm install \
+    -n astria-validator-node0 node0-sequencer-chart ./sequencer \
+    -f ../dev/values/validators/all.yml \
+    -f ../dev/values/validators/node0.yml \
+    --set sequencer.optimisticBlockApis.enabled=true \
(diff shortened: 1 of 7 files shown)


charts/auctioneer-test.just
```diff
@@ -0,0 +1,124 @@
+_default:
+  @just --list auctioneer-test
+
+defaultTag := ""
+defaultNamespace := "astria-dev-cluster"
+evm_destination_address := "0x507056D8174aC4fb40Ec51578d18F853dFe000B2"
+smoke_test_image := "bharath123456/auctioneer-smoke-test:latest"
```
bharath-123, line 7: I pushed the image to my personal dockerhub account for the purpose of a POC. it should be trivial to move it to an ghcr/astria account.

charts/auctioneer-test.just
```diff
@@ -0,0 +1,124 @@
...
+    {{ if tag != '' { replace('--set images.sequencer.devTag=# --set sequencer-relayer.images.sequencerRelayer.devTag=#', '#', tag) } else { '' } }} \
+    --create-namespace > /dev/null
+
+  echo "Setting up the third astria sequencer..." && helm install \
+    -n astria-validator-node2 node2-sequencer-chart ./sequencer \
+    -f ../dev/values/validators/all.yml \
+    -f ../dev/values/validators/node2.yml \
+    --set sequencer.optimisticBlockApis.enabled=true \
+    {{ if tag != '' { replace('--set images.sequencer.devTag=# --set sequencer-relayer.images.sequencerRelayer.devTag=#', '#', tag) } else { '' } }} \
+    --create-namespace > /dev/null
```
bharath-123, line 50: We could probably  do this in a more concise way.

charts/flame-rollup/values.yaml
```diff
@@ -11,7 +11,7 @@ images:
     repo: ghcr.io/astriaorg/flame
     pullPolicy: IfNotPresent
     tag: sha-457e1f9
-    devTag: sha-457e1f9
+    devTag: latest
```
bharath-123, line 14: We need to merge https://github.com/astriaorg/astria/pull/1972 for these changes to go into main.

charts/auctioneer/values.yaml
```diff
@@ -9,9 +9,8 @@ images:
   auctioneer:
     repo: ghcr.io/astriaorg/auctioneer
     pullPolicy: IfNotPresent
-    # TODO - update to latest tag
-    tag: "pr-1839"
-    devTag: "pr-1839"
+    tag: sha-2db84ed
```
bharath-123, line 12: We need to merge https://github.com/astriaorg/astria/pull/1972 for these changes to go into main.

joroshiba:
closed in 7 days.

joroshiba:

PR #1978, opened 2025-02-18
chore(charts, cometbft): upgrade cometbft version to v0.38.17 and hermes usage

### LINKED ISSUE #1980: chore(cometbft): upgrade cometbft to v0.38.17
cometbft [v0.38.17](https://github.com/cometbft/cometbft/blob/v0.38.17/CHANGELOG.md#v03817) includes two security fixes:
* [blocksync] Ban peer if it reports height lower than what was previously reported ([ASA-2025-001](https://github.com/cometbft/cometbft/security/advisories/GHSA-22qq-3xwm-r5x4))
* [types] Check that Part.Index equals Part.Proof.Index ([ASA-2025-001](https://github.com/cometbft/cometbft/security/advisories/GHSA-r3r4-g7hq-pq4f))

Should test and update all our nodes to use the security patch.

quasystaty1:
## Summary
upgrades cometbft version from `v0.38.8 -> v0.38.17 `
## Background
cometbft [v0.38.17](https://github.com/cometbft/cometbft/blob/v0.38.17/CHANGELOG.md#v03817) includes two security fixes. 
## Changes
- bumps cometbft version to v0.38.17
- hermes now uses serde patched image ( but in the `block_result` response)
- hermes now uses pull instead of push method for events.

## Testing
locally running smoke tests against a local cluster.

closes #1980

### DIFF
--- a/dev/values/hermes/local.yaml
+++ b/dev/values/hermes/local.yaml
@@ -57,9 +57,8 @@ chains:
         filename: celestia.json
         key: token
     eventSource:
-      mode: push
-      url: ws://celestia-app-service.astria-dev-cluster.svc.cluster.local:26657/websocket
-      batchDelay: 200ms
+      mode: pull
+      interval: 1s
     gasDenom: "utia"
     gasPrice: 0.0026
     gasMultiplier: 1.2
--- a/charts/hermes/Chart.yaml
+++ b/charts/hermes/Chart.yaml
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.5.2
+version: 0.5.3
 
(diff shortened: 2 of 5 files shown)


dev/values/hermes/local.yaml
```diff
@@ -57,9 +57,8 @@ chains:
         filename: celestia.json
         key: token
     eventSource:
-      mode: push
-      url: ws://celestia-app-service.astria-dev-cluster.svc.cluster.local:26657/websocket
-      batchDelay: 200ms
+      mode: pull
+      interval: 1s
```
joroshiba, line 61: is this change just made to be more similar to production or does the push methodology not work with newest version?

I think its totally reasonable to say we only use pull methodology, but not sure on why we need the change here.

joroshiba (approved):

dev/values/hermes/local.yaml
quasystaty1: Yes exactly. I tested with push methodology as well.

PR #1985, opened 2025-02-19
chore(sequencer): add snapshots for storage values to enforce non-state-breaking changes

### LINKED ISSUE #1982: Provide snapshot tests for storage values in sequencer
We need to ensure new variants of the many `<component>::storage::ValueImpl` enums are only ever appended to the existing ones.  Providing snapshot tests for every variant will help ensure this (as well as helping ensure existing variants don't change).

We should also provide a snapshot test for each variant of the top-level `StoredValue` enum too, for the same reasons.

ethanoroshiba:
## Summary
Added snapshot tests for all storage values to ensure that future changes do not break state.

## Background
The addition of the `BridgeTransfer` action revealed how inserting a value into the storage `ValueImpl` is a breaking change, since deserialization would then fail. New values, such as `BridgeTransferFees` need to be appended instead of inserted. Adding snapshots for each storage variant's discriminant ensures that there is an extra level of protection against making a state breaking change.

## Changes
- Added snapshots for the borsch encoding of every `ValueImpl` variant and every `StoredValue` variant.

## Testing
Manually reviewed every discriminant.

## Changelogs
No updates required.

## Related Issues
closes #1982


Fraser999 (commented):
I think we should go the whole way and just test the borsh-encoded form of each Value variant.  This will help ensure that not only the discriminant remains unchanged, but also the underlying wrapped type.  I'd output the resulting `Vec<u8>` as hex though to keep the snapshot files slightly cleaner.

I'd also be inclined to follow the approach taken for the keys too and just bundle each components' tests into a single one, e.g.

```rust
#[test]
fn values_should_not_change() {
    assert_snapshot!("balance", borsh_then_hex(&ValueImpl::Balance(Balance(0))));
    assert_snapshot!("nonce", borsh_then_hex(&ValueImpl::Nonce(Nonce(0))));
    assert_snapshot!(
        "stored_value_accounts",
        borsh_then_hex(&StoredValue::Accounts(Value(ValueImpl::Balance(Balance(0)))))
    );
}
```
where `borsh_then_hex` is a simple function added to `test_utils` like:
```rust
pub(crate) fn borsh_then_hex<T: borsh::BorshSerialize>(item: &T) -> String {
    hex::encode(&borsh::to_vec(item).unwrap())
}
```
I'm less bothered about grouping the tests though - feel free to leave them separate.

crates/astria-sequencer/src/storage/stored_value.rs
```diff
@@ -30,3 +30,19 @@ impl StoredValue<'_> {
...
+
+#[cfg(test)]
+mod tests {
+    use insta::assert_snapshot;
+
+    use super::*;
+    use crate::test_utils::borsh_then_hex;
+
+    #[test]
+    fn stored_value_unit_variant_unchanged() {
```
Fraser999, line 42: Nice job remembering to test this case :)

Fraser999 (approved):
Thanks for adding all these!

joroshiba (approved):
LGTM

PR #1990, opened 2025-02-20
feat(charts)!: sequencer defaults for networks

joroshiba:
## Summary
Moves settings which should not change for various sequencer networks into top level global field. When deploying to different networks, this makes it such that values files will only need to unique configuration for a given node set as opposed to many values being unable to be changed. Makes it more difficult to misconfigure a node setting. 

## Background

## Changes
- Add global setting of `network` to sequencer & sequencer-relayer charts.
- Configure the genesis & block timing parameters to defaults w/o override options for core networks
- Configure default images for individual networks, with optional overrides in "dev" mode. 

## Testing
CI/CD testing, and syncing of a mainnet node. 

## Changelogs
No updates required.

## Breaking Changelist
The default here is still custom to ensure backwards compatibility.

### DIFF
--- a/charts/deploy.just
+++ b/charts/deploy.just
@@ -201,7 +201,7 @@ deploy-smoke-test tag=defaultTag:
     -n astria-validator-single single-sequencer-chart ./charts/sequencer \
     -f dev/values/validators/all.yml \
     -f dev/values/validators/single.yml \
-    {{ if tag != '' { replace('--set images.sequencer.devTag=# --set sequencer-relayer.images.sequencerRelayer.devTag=#', '#', tag) } else { '' } }} \
+    {{ if tag != '' { replace('--set images.sequencer.tag=# --set sequencer-relayer.images.sequencerRelayer.tag=#', '#', tag) } else { '' } }} \
     --create-namespace > /dev/null
   @just wait-for-sequencer > /dev/null
   @echo "Starting EVM rollup..." && helm install -n astria-dev-cluster astria-chain-chart ./charts/evm-stack -f dev/values/rollup/dev.yaml \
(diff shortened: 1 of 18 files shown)


WafflesVonMaple (approved):

jbowen93:
I read this and I’m unclear what excatly is happening. 

This seems to be creating a map of network version to some config values such that you just say network: dawn-1 and that sets certain values or you can use network: custom to set things manually. 

I’m not sure what values are actually being set by eg. network: dawn-1

This also moves the genesis files here, from where? 

Was this tested locally? Presumably with the mainnet file in /dev?

jbowen93:
Ah it's the template
```
{{- define "sequencer.imageTag" -}}
{{- if or (eq .Values.global.network "custom") (eq .Values.global.dev true) }}{{ .Values.images.sequencer.tag }}
{{- else if eq .Values.global.network "mainnet" }}1.0.0
{{- else if eq .Values.global.network "dawn-1" }}1.0.0
{{- else if eq .Values.global.network "dusk-11" }}1.0.0
{{- end }}
{{- end }}
```

which sets the values based on the new network value
```
  # Supported options are `custom`, `mainnet`, `dawn-1`, or `dusk-11`.
  # If you are using a custom network, you should provide the genesis
  # configuration via values.
  #
  # When set to a specific network, there are default images and genesis files
  # that are used.
  network: "custom"
  ```

joroshiba:
> Was this tested locally? Presumably with the mainnet file in /dev?

Yes it is also being used right now in the mainnet cluster I've been keeping synced to head with latest merged code.

joroshiba:
> I’m not sure what values are actually being set by eg. network: dawn-1
It's more about what values are being ignored. I've added some comments throughout the values file to make it a little bit more clear, but the overrides for images are set in helpers, the genesis files are maintained seperately. 

> This also moves the genesis files here, from where?
The `custom.genesis.json` is identical to what existed before, the file used is now just dependent on your network setting. 

Was this tested locally? Presumably with the mainnet file in /dev?
> Yes, it is also running on the sequencer node in our k8s cluster we have synced to latest on mainnet right now. That node is configured with `dev: true` and `network: mainnet` so it uses the latest image and binary features, while running mainnet configuration otherwise.

PR #1991, opened 2025-02-20
fix(core): use i128 for price instead of u128

noot:
## Summary
it's theoretically possible for the price of an asset to become negative, although rare. in the case where a price became negative, the oracle would be unable to decode the price from the sidecar.

## Background
resolves issue 3.2 "Price can become negative" in the zellic audit of the sequencer-side oracle implementation.

## Changes
- implement `Int128` proto type, similar to our existing primitive `Uint128`
- change `Price` in core to be `i128` internally, instead of `u128`

## Testing
unit tests

### DIFF
--- a/crates/astria-core/src/connect/types.rs
+++ b/crates/astria-core/src/connect/types.rs
@@ -448,5 +448,11 @@ pub mod v2 {
             "ETH /USD".parse::<CurrencyPair>().unwrap_err();
             "ETH/ USD".parse::<CurrencyPair>().unwrap_err();
         }
+
+        #[test]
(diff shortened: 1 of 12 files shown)


SuperFluffy (approved):
Some comments on testing, but looks fine to me.

Note on the PR title: this should probably be `fix(proto, core)`

noot:
> Also is there some interaction anywhere where you have to convert u128 to i128 or vice versa? Or will the price always "stand by itself"?

@SuperFluffy no, there shouldn't be any case where prices are turned into a u128, the `Price` type will always stand alone.

PR #1998, opened 2025-02-24
tests(sequencer): add unit tests for `bridge_account_last_tx_hash` and `transaction_fee` ABCI queries

### LINKED ISSUE #1741: Add unit tests for queries missing tests
The following PRs introduced queries, none of which provided unit tests for the query handling. They should be added.

https://github.com/astriaorg/astria/pull/1158
https://github.com/astriaorg/astria/pull/1189
https://github.com/astriaorg/astria/pull/1196

┆Issue Number: ENG-951

noot:
## Summary
add unit tests for `bridge_account_last_tx_hash` and `transaction_fee` ABCI queries

## Background
unit tests were missing for these functions.

## Changes
- add unit tests for `bridge_account_last_tx_hash` and `transaction_fee` ABCI queries
- question: the `transaction_fee` query does not error if one of the `fee_assets` is not an authorized fee asset in state. is this something we want to check? is there a case where someone may want to calculate the fee for an action but not use a currently authorized `fee_asset`?

## Testing
no code changes, just tests added.

## Related Issues

closes #1741


crates/astria-sequencer/src/fees/query.rs
```diff
@@ -261,6 +261,7 @@ pub(crate) async fn transaction_fee_request(
         };
         fees.push((trace_denom.into(), value));
     }
+    fees.sort_by(|a: &(Denom, u128), b: &(Denom, u128)| a.0.cmp(&b.0));
```
Fraser999, line 264: It looks like this change (and the addition of `PartialEq` and `Eq` to `TransactionFeeResponse`) was added to the production code purely to support testing using `assert_eq`.  If so, I'd prefer if the production code was left as is and the tests changed to support the vecs having differing orders.

noot: removed this line and changed the tests to sort instead.

Fraser999 (approved):

PR #1999, opened 2025-02-25
feat(conductor): include sequencer block hash

joroshiba:
## Summary
Adds a `sequencer_block_hash` to Execution API `Block` and `ExecuteBlock` shapes such that an execution can include it in their block space.

This is an addition but is backwards compatible. Old rollups can run the new binary, rollups that don't use the sequencer hash can utilize the old binary. Rollup forks which rely on the sequencer hash must utilize the new binary. 

## Background
We are upgrading Flame to the `Cancun` upgrade which includes `BeaconRoot` in the block hash. This is a natural time to add the sequencer block, as it is a similar concept. 

## Changes
- Add new hash field to execution v1 protos and `astria-core`
- Added the field to in development execution v2 protos
- Update conductor to include the new data in `ExecuteBlock` requests
- Added information to the execute API spec.

## Testing
CI/CD testing has been updated to run the smoke tests with an updated version of geth that utilizes the sequencer and includes a beacon root contract in genesis.

I have done manual testing of the following:
- old geth w/ new conductor unchanged
- new geth w/ cancun upgrade disabled

## Changelogs
Changelogs updated

## Breaking Changelist
This is not a breaking change as it is a protobuf addition.

### DIFF
--- a/proto/executionapis/astria/execution/v2/execute_block_request.proto
+++ b/proto/executionapis/astria/execution/v2/execute_block_request.proto
@@ -20,7 +20,11 @@ message ExecuteBlockRequest {
   repeated astria.sequencerblock.v1.RollupData transactions = 3;
   // Timestamp to be used for new block.
   google.protobuf.Timestamp timestamp = 4;
-  // The block hash of the sequencer which this block is derived from.
+  // The hash of the sequencer block from which the transactions and timestamp
+  // are derived.
+  //
+  // Must be a 32 byte base16 encoded string. It may be prefixed with `0x`.
+  //
   // Utilizing this field is optional for the execution node.
   string sequencer_block_hash = 5;
 }
--- a/proto/executionapis/astria/execution/v2/execute_block_request.proto
+++ b/proto/executionapis/astria/execution/v2/execute_block_request.proto
@@ -20,4 +20,7 @@ message ExecuteBlockRequest {
   repeated astria.sequencerblock.v1.RollupData transactions = 3;
   // Timestamp to be used for new block.
   google.protobuf.Timestamp timestamp = 4;
+  // The block hash of the sequencer which this block is derived from.
+  // Utilizing this field is optional for the execution node.
+  string sequencer_block_hash = 5;
 }
--- a/proto/executionapis/astria/execution/v2/executed_block_metadata.proto
+++ b/proto/executionapis/astria/execution/v2/executed_block_metadata.proto
@@ -17,4 +17,7 @@ message ExecutedBlockMetadata {
   // Timestamp of the block, taken from the sequencer block that this rollup block
   // was constructed from.
   google.protobuf.Timestamp timestamp = 4;
+  // The block hash of the sequencer block which this execution block is derived
+  // from. Utilizing this field is optional.
+  string sequencer_block_hash = 5;
 }
--- a/proto/executionapis/astria/execution/v2/executed_block_metadata.proto
+++ b/proto/executionapis/astria/execution/v2/executed_block_metadata.proto
@@ -17,7 +17,11 @@ message ExecutedBlockMetadata {
   // Timestamp of the block, taken from the sequencer block that this rollup block
   // was constructed from.
   google.protobuf.Timestamp timestamp = 4;
-  // The block hash of the sequencer block which this execution block is derived
-  // from. Utilizing this field is optional.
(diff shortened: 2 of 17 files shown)


proto/executionapis/astria/execution/v2/executed_block_metadata.proto
```diff
@@ -17,4 +17,7 @@ message ExecutedBlockMetadata {
   // Timestamp of the block, taken from the sequencer block that this rollup block
   // was constructed from.
   google.protobuf.Timestamp timestamp = 4;
+  // The block hash of the sequencer block which this execution block is derived
+  // from. Utilizing this field is optional.
```
Fraser999, line 21: Nit: we have five variations of this comment 
* The hash of the sequencer block that this block is derived from.
* The block hash of sequencer block this is derived from.
* The hash of the sequencer block the transactions come from.
* The block hash of the sequencer which this block is derived from.
* The block hash of the sequencer block which this execution block is derived from.

I wouldn't have mentioned it except for the fact that "The block hash of the sequencer which this block is derived from." is probably worth changing.

Fraser999 (approved):
One small, non-blocking nitpick only.

I also wonder if it's worth documenting how the long hex string added to dev.yaml is generated?  If it's obvious to others though, no need.

quasystaty1 (approved):
Infra approval. Few notes:

* changes should be added to flame dev file as well for local deployments.
* note that deployments will now activate Cancun by default, this is ok with me but wanted to flag here.
* reminder to switch geth devTag back to latest once `pr-75` is merged in [astria-geth ](https://github.com/astriaorg/astria-geth)repo

SuperFluffy (approved):
Some nits mainly.

proto/executionapis/astria/execution/v2/executed_block_metadata.proto
joroshiba: It must be optional in order to sync Flame or Forma nodes with the new conductor software.

Cannot change old blocks.

SuperFluffy: This is for execution API v2 though

proto/executionapis/astria/execution/v2/execute_block_request.proto
```diff
@@ -20,4 +20,7 @@ message ExecuteBlockRequest {
   repeated astria.sequencerblock.v1.RollupData transactions = 3;
   // Timestamp to be used for new block.
   google.protobuf.Timestamp timestamp = 4;
+  // The block hash of the sequencer which this block is derived from.
+  // Utilizing this field is optional for the execution node.
```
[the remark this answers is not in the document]
joroshiba: It cannot be a requirement, otherwise you cannot sync all existing blocks of current mainnet dependent execution clients.

In v2 we COULD make it something specified whether or not it will be used in the execution session though.

proto/executionapis/astria/execution/v2/executed_block_metadata.proto
SuperFluffy: Ah, you mean upgraded flame but old blocks. Ok, that makes sense.

Can we make this required after some block number? Or too much effort?

proto/executionapis/astria/execution/v2/execute_block_request.proto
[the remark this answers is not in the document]
SuperFluffy: Yeah, I see. Noticed in your other comment.

Should we add this to the session? Or is that silly?

joroshiba:
@quasystaty1 re note here:

> note that deployments will now activate Cancun by default, this is ok with me but wanted to flag here.

Just wanted to note that this was done such that it only affects deployments using the dev configurations, so local deployments and smoke tests. When dependent changes are merged I will rollback the dev file changes and roll into the top level charts.

joroshiba:
> Don't we have changelogs for the Protos?

@SuperFluffy not currently. Can look into changing separately but not in scope here.

proto/executionapis/astria/execution/v2/executed_block_metadata.proto
joroshiba: Off line synced, decided yagni.

This is optional from the rollup perspective, it's optional to use it. Conductor will always include it in execution requests, but the execution node doesn't need to use it.

PR #2000, opened 2025-02-25
chore(ci): add release workflow for auctioneer

bharath-123:
## Summary
Adds a github workflow to build an auctioneer release

## Background
To launch auctioneer on mainnet, we need to be able to make releases of it. This workflow allows us to build an auctioneer release with a gh action.

## Changes
- Add auctioneer release workflow to `.github/workflow/release.yaml`


aajimal (approved):
👍

PR #2001, opened 2025-02-26
feat(sequencer)!: implement `RecoverClient` action based on sequencer-v1.0.0

noot:
## Summary
implement `RecoverClient` action which replaces an expired or otherwise inactive IBC client.

this PR is based off the sequencer-v1.0.0 release tag.

this is based off `MsgRecoverClient` implemented in ibc-go v8: https://github.com/cosmos/ibc-go/blob/a71577cc1bf242f8c99360777106b1321503b17f/proto/ibc/core/client/v1/tx.proto#L126

## Background
there is no way to otherwise replace an expired IBC client without creating a new one and creating a governance proposal on the counterparty chain to change the client there, which is a lot more hassle than just being able to replace an expired client on one chain.

## Changes
- implement `RecoverClient` action 
- the implementation is based off the following ibc-go implementation: 
- https://github.com/cosmos/ibc-go/blob/a71577cc1bf242f8c99360777106b1321503b17f/modules/core/02-client/keeper/client.go#L121
- https://github.com/cosmos/ibc-go/blob/a71577cc1bf242f8c99360777106b1321503b17f/modules/light-clients/07-tendermint/light_client_module.go#L179
- https://github.com/cosmos/ibc-go/blob/a71577cc1bf242f8c99360777106b1321503b17f/modules/light-clients/07-tendermint/proposal_handle.go#L29
- i had to export some types from penumbra (ConsensusStateWriteExt, ClientState) which is why it's on a different commit (commit is based off v0.80.7)

## Testing
not yet, will be done on devnet

## Changelogs

## Breaking Changelist
- a new action is added, which is a breaking change.


crates/astria-sequencer/src/ibc/recover_client.rs
```diff
@@ -0,0 +1,124 @@
...
+                .await
+                .wrap_err("failed to get sudo address")?,
+            "only the sudo address can recover a client",
+        );
+
+        let timestamp = state
+            .get_block_timestamp()
+            .await
+            .wrap_err("failed to get timestamp")?;
+        let subject_client_status = state
```
joroshiba, line 51: similar nits as in proto on naming here just too many things starting with `sub` makes it a slower read?

crates/astria-sequencer/src/ibc/recover_client.rs
```diff
@@ -0,0 +1,124 @@
...
+            .get_client_status(&self.subject_client_id, timestamp)
+            .await;
+
```
joroshiba, line 54: nit: maybe new line not needed between the fetch/verification of object

crates/astria-sequencer/src/ibc/recover_client.rs
```diff
@@ -0,0 +1,124 @@
...
+        ensure!(
+            subject_client_status != ClientStatus::Active,
+            "cannot recover an active client",
+        );
+
+        let substitute_client_status = state
+            .get_client_status(&self.substitute_client_id, timestamp)
+            .await;
+
```
joroshiba, line 63: nit: same as above

crates/astria-sequencer/src/ibc/recover_client.rs
```diff
@@ -0,0 +1,124 @@
...
+            )
+            .await
+            .map_err(anyhow_to_eyre)
+            .wrap_err(
+                "failed to put
+        verified consensus state",
+            )?;
+
+        subject_client_state.latest_height = substitute_client_state.latest_height;
+        subject_client_state.chain_id = substitute_client_state.chain_id;
```
joroshiba, line 119: Should we actually have an ensure block on this? To make it so that recovery can never change the chain being acted on?

I suppose there may be a usecase here were there was a hardfork in chain and client expired but assets are considered the same? 

Chain id changing hardforks are rare, nearly all the time we probably want to replace with a client of the same chain id and not doing so would be a safety failure? I'd be ok just not supporting that usecase at this time, and having to evaluate what to do if it does come up?

noot: the spec allows the chain ID to be changed - https://ibc.cosmos.network/architecture/adr-026-ibc-client-recovery-mechanisms/ although i'm not sure exactly why. this might be for the case you said where the counterparty forks and changes their chain ID, but we didn't do a client upgrade in time, which would invalidate the client.

proto/protocolapis/astria/protocol/transaction/v1/action.proto
```diff
@@ -228,9 +229,18 @@ message FeeChange {
...
   }
 }
 
 message IbcSudoChange {
   astria.primitive.v1.Address new_address = 1;
 }
+
+message RecoverClient {
+  // the client identifier for the client to be updated 
+  string subject_client_id = 1;
```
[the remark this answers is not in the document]
noot: don't really like just `client_id` as it's too ambiguous imo and will be more confusing reading the code. could do `client_id_to_replace` and `substitute_client_id`, or `subject_client_id` and `replacement_client_id`?

noot:
closing in favour of #2008

PR #1962, opened 2025-02-11
feat(proto): create v2 execution API

ethanoroshiba:
## Summary
Creates v2 execution API to support upcoming v2 changes to Conductor and Geth.

## Background
These changes are meant to support the upcoming v2 upgrades to [Conductor](https://github.com/astriaorg/astria/pull/1928) and [astria-geth](https://github.com/astriaorg/astria-geth/pull/59), which will facilitate the ability to have rollup forks, driven by the need to migrate Forma to Mainnet.

## Changes from `v1`
- Removed `GenesisInfo` and associated RPC in favor of `SequencerInfo`, which contains the following changes from its predecessor:
     - Changed `sequencer_genesis_block_height` to `sequencer_first_block_height`.
     - Added `rollup_first_block_number` and `rollup_stop_block_number`, for determining which rollup height to start execution at along with the ability to shutdown/restart at a given height.
     - Added `sequencer_chain_id` and `celestia_chain_id`.
     - Added `halt_at_rollup_stop_block_number`, which indicates to the Conductor to halt execution at the stop number instead of restart and refetching the sequencer info.
- Added `commitment_type` field to `GetSequencerInfoRequest`, so that the rollup can respond with the correct sequencer info for the rollup block number with the given commitment level.

## Testing
No testing needed.

## Changelogs
Protos have no changelog

PR #2006, opened 2025-02-27
feat(conductor)!: implement v2 execution API

### LINKED ISSUE #1994: Conductor changes for Forma migration
Conductor needs to be updated to support the new v2 Execution API:

- Rely on `NewExecutionSession` upon startup or restart instead of `GetGenesisInfo` and `GetCommitmentState`.
- Start and stop at `ExecutionSessionParameters.rollup_start_block_number` and `ExecutionSessionParameters.rollup_end_block_number`.

### LINKED ISSUE #2003: Move optimistic API v1 execution API dependencies to v2
Now that we have a v2 execution API, we should move our optimistic API dependencies on it over

ethanoroshiba:
## Summary
Updates Conductor to use v2 Execution API, including use of new "execution sessions".

## Background
These changes are necessary to support migrating Forma to mainnet.

Migrating Forma to Astria's main net requires introducing stop and start logic into the Conductor. A migration using these changes could look like this:

1. Running against a fresh rollup node, Conductor starts and initiates a new execution session via `astria.execution.v2.CreateExecutionSession`, which contains `rollup_start_block_number = 1`, and `rollup_end_block_number = 1000`.
2. Conductor executes (syncs) the rollup node from rollup block numbers 1 to 1000 (inclusive)
3. Upon executing rollup number 1000, conductor resets its inner state and initiates a new execution session via `astria.execution.v2.CreateExecutionSession`, with new information from which it will continue executing.

For cases where `rollup_end_block_number` is unset (or equivalently set to 0), conductor will never restart as there is no stop condition.

For conductors running in firm-only or in soft-and-firm modes, only the firm-executed block triggers a restart (in soft-and-firm mode, upon soft-executing at the rollup stop number no blocks will be executed past it).

For conductors running in soft-only mode the soft-executed block triggers the restart.

If the rollup chain is halted, it should respond with `tonic::Code::FailedPrecondition` to `CreateExecutionSession`, which will cause the Conductor to shut down upon attempting a restart.

```yaml
config:
        AstriaRollupName:
        AstriaForks: []
                # - < fork name >:
                #         height: ""
                #         halt: ""
                #         feeCollector: ""
                #         EIP1559Params: {}
                #         sequencer:
                #                 chainId: ""
                #         celestia:
                #                 chainId: ""
                #                 searchHeightMaxLookAhead: ""
                #         bridgeAddresses: []
                #         astriaOracleCallerAddress: ""
                #         astriaOracleContractAddress: ""
```

### In `astria-core`
- Replaced v1 execution domain types with v2 execution domain types:
    - `ExecutionSession`
    - `ExecutionSessionParameters`
    - `ExecutedBlockMetadata`
    - `CommitmentState` (name unchanged, but fields are)
- For more information on API changes, see #1962 

### In `astria-auctioneer`
- Updated execution types to v2

### In `astria-conductor`
- Changed sequencer-to-rollup-height calculation to start from `rollup_start_block_number` in `ExecutionSessionParameters`, so that Conductor doesn't start from rollup height 1 again after restarting.
- Removed `ASTRIA_CONDUCTOR_SEQUENCER_CHAIN_ID` and `ASTRIA_CONDUCTOR_CELESTIA_CHAIN_ID` env vars from Conductor (chain ID checks will now be performed against the values retrieved from the genesis info)
- Added restart logic for when the stop block number is reached:  Conductor will restart at the stop height and initiate a new execution session
- Removed multiplication of Celestia variance.

### In `charts`
- Add just command for deploying the EVM rollup restart test, which functions the same as our regular smoke test with a restart at sequencer height 20.

## Testing
- Added miscellaneous unit tests to ensure heights are correctly calculated and whether to restart is successfully determined.
- Added blackbox tests to ensure conductor correctly restarts in each mode, and continues to execute at the correct height after initiating a new execution session.
- Added EVM rollup restart test which restarts EVM and Conductor at sequencer height 20 during the regular smoke test.

## Changelogs
Changelogs updated

## Breaking Changelist
- Domain v1 execution types have been almost entirely replaced
- Conductor config is not backwards compatible, i.e. old conductor will work with new config but not vice versa.
- Conductor now required v2 execution API

## Related Issues
closes #1994 
closes #2003


crates/astria-core/src/execution/v2/mod.rs
```diff
@@ -0,0 +1,699 @@
...
+        Self(ExecutionSessionErrorKind::MissingExecutionSessionParameters)
+    }
+
+    fn missing_commitment_state() -> Self {
+        Self(ExecutionSessionErrorKind::MissingCommitmentState)
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+enum ExecutionSessionErrorKind {
```
SuperFluffy, line 37: Can you name the fields that were bad in all cases? I.e. "field `.commitment_state` was not set".

IMO make all the variants here struct-like (those that wrap an error that is), and do something like:

```rust
InvalidExecutionSessionParameters { source: ExecutionSessionParametersError }
```

I don't think you are using the `#[from]` anywhere.

crates/astria-core/src/execution/v2/mod.rs
```diff
@@ -0,0 +1,699 @@
...
+    session_id: String,
+    /// The configuration for the execution session.
+    execution_session_parameters: ExecutionSessionParameters,
+    /// The commitment state for executing client to start from.
+    commitment_state: CommitmentState,
+}
+
+impl ExecutionSession {
+    #[must_use]
+    pub fn session_id(&self) -> &String {
```
SuperFluffy, line 67:
```suggestion
    pub fn session_id(&self) -> &str {
```

crates/astria-core/src/execution/v2/mod.rs
```diff
@@ -0,0 +1,699 @@
...
+    type Raw = raw::ExecutionSession;
+
+    fn try_from_raw_ref(raw: &Self::Raw) -> Result<Self, Self::Error> {
+        let raw::ExecutionSession {
+            session_id,
+            execution_session_parameters,
+            commitment_state,
+        } = raw;
+        let execution_session_parameters = execution_session_parameters
+            .as_ref()
```
SuperFluffy, line 93: why `.as_ref()`? It's already a `&execution_session_parameters`

crates/astria-core/src/execution/v2/mod.rs
```diff
@@ -0,0 +1,699 @@
...
+            .ok_or_else(Self::Error::missing_execution_session_parameters)?;
+        let execution_session_parameters =
+            ExecutionSessionParameters::try_from_raw_ref(execution_session_parameters)
+                .map_err(Self::Error::execution_session_parameters)?;
+        let commitment_state = commitment_state
+            .as_ref()
```
SuperFluffy, line 99: I don't think you need the `as_ref`

crates/astria-core/src/execution/v2/mod.rs
```diff
@@ -0,0 +1,699 @@
...
+            execution_session_parameters: Some(execution_session_parameters),
+            commitment_state: Some(commitment_state),
+        }
+    }
+}
+
+// An error when transforming a [`raw::ExecutionSessionParameters`] into a
+// [`ExecutionSessionParameters`].
+#[derive(Debug, thiserror::Error)]
+#[error(transparent)]
```
SuperFluffy, line 129: As mentioned for the other error above, let's give this a proper message, i.e. `"failed to validate astria.execution.v2.ExecutionSessionParameters"`

crates/astria-core/src/execution/v2/mod.rs
```diff
@@ -0,0 +1,699 @@
...
+    }
+
+    fn invalid_sequencer_start_block_height(inner: tendermint::Error) -> Self {
+        Self(ExecutionSessionParametersErrorKind::InvalidSequencerStartBlockHeight(inner))
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+enum ExecutionSessionParametersErrorKind {
+    #[error("`rollup_id` field contained an invalid rollup ID")]
```
SuperFluffy, line 148: Make it explicit that this is a field:
```suggestion
    #[error("field `.rollup_id` field contained an invalid rollup ID")]
```

Similary for the other variants below.

crates/astria-core/src/execution/v2/mod.rs
```diff
@@ -0,0 +1,699 @@
...
+    IncorrectRollupIdLength(IncorrectRollupIdLength),
+    #[error("`rollup_id` was not set")]
+    NoRollupId,
+    #[error(
+        "`tendermint::block::Height` could not be constructed from `sequencer_start_block_height`"
```
SuperFluffy, line 153: Don't make this type part of the error message. It's sufficient to state for example "field `.sequencer_start_block_height` was invalid".

crates/astria-core/src/execution/v2/mod.rs
SuperFluffy, line 1: Have a look through all errors in this file:

1. give `transparent` errors an actual error message like "failed validating <Protobuf full name>".
2. Make the wrapped error kinds `source: <The Error Kind>`
3. In every error message that corresponds to a specific field being validated, name the field like "field `.foobar` was invalid".

Also try to avoid using rust types in error messages. Readers should not need to worry about that.

SuperFluffy (commented):
Quick initial set of observations. Going to continue with the rest now.

crates/astria-core/src/execution/mod.rs
```diff
@@ -1 +1 @@
-pub mod v1;
```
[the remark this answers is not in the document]
ethanoroshiba: Agreed, but I think if we make a decision on this we should probably keep it consistent throughout `core`, so I think maybe this would be best put as a followup? Additionally, the current setup also matches the PB path 1:1, but I'm not sure if that's important enough to maintain this setup or if it's unimportant.

crates/astria-core/src/execution/v2/mod.rs
ethanoroshiba: `.as_ref()` makes it `Option<&ExecutionSessionParameters>` as opposed to `&Option<ExecutionSessionParameters>`, since `ok_or_else()` takes ownership of the option

crates/astria-conductor/src/conductor/inner.rs
```diff
@@ -87,49 +88,59 @@ impl Inner {
...
+            }
+
+            match exit_status {
+                Ok(None) => Err(eyre!(
+                    "executor exited with a success value but without rollup status even though \
+                     it was not explicitly cancelled; this shouldn't happen"
+                )),
+                Ok(Some(status)) => should_restart_or_shutdown(&self.config, &status),
+                Err(error) => {
+                    error!(%error, "executor failed; checking error chain if conductor should be restarted");
```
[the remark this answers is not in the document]
ethanoroshiba: I'm in favor of keeping this as an error message, personally. Despite restarting in the `PermissionDenied` case, we did indeed still get an error back from the rollup. We are no longer relying on this for a planned, but only restarts in the case of a true error. I'm not super strongly opinionated on this, though, so if you are I'm happy to change

crates/astria-conductor/src/conductor/inner.rs
```diff
@@ -162,17 +173,198 @@ fn check_for_restart(err: &eyre::Report) -> bool {
     false
 }
 
+fn should_restart_or_shutdown(
```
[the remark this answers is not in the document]
ethanoroshiba: We set to `RestartOrShutdown::Shutdown` if the cancellation token is cancelled

crates/astria-core/src/execution/v2/mod.rs
```diff
@@ -0,0 +1,699 @@
...
+    generated::astria::execution::v2 as raw,
+    primitive::v1::{
+        IncorrectRollupIdLength,
+        RollupId,
+    },
+    Protobuf,
+};
+
+#[derive(Debug, thiserror::Error)]
+#[error(transparent)]
```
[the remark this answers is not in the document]
ethanoroshiba: Gave all transparent errors messages (in this module) in 80238e7.

crates/astria-core/src/execution/v2/mod.rs
ethanoroshiba: Done in 80238e7. Only ones I was not able to make struct-like were those with `&'static str`.

crates/astria-core/src/execution/v2/mod.rs
ethanoroshiba: 80238e7

crates/astria-core/src/execution/v2/mod.rs
ethanoroshiba: Done in 80238e7

crates/astria-core/src/execution/v2/mod.rs
ethanoroshiba: 80238e7

crates/astria-core/src/execution/v2/mod.rs
ethanoroshiba: 80238e7

crates/astria-core/src/execution/v2/mod.rs
ethanoroshiba: Tried to address all of these in 80238e7. Let me know if that covers it!

crates/astria-conductor/src/conductor/inner.rs
```diff
@@ -87,49 +88,59 @@ impl Inner {
     pub(super) async fn run_until_stopped(mut self) -> eyre::Result<RestartOrShutdown> {
         info_span!("Conductor::run_until_stopped").in_scope(|| info!("conductor is running"));
 
-        let exit_reason = select! {
+        let exit_status = select! {
             biased;
 
             () = self.shutdown_token.cancelled() => {
-                Ok(ShutdownSignalReceived)
+                Ok(None)
```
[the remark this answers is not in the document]
ethanoroshiba: Changed to enum `ExitReason` with variants `ShutdownSignalReceived` and `ChannelsClosed(Box<State>)` in 80238e7

crates/astria-conductor/src/conductor/inner.rs
```diff
@@ -149,7 +160,7 @@ impl Inner {
 }
 
 #[instrument(skip_all)]
-fn check_for_restart(err: &eyre::Report) -> bool {
+fn should_restart_despite_error(err: &eyre::Report) -> bool {
```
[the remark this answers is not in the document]
ethanoroshiba: Added events in 80238e7.

crates/astria-conductor/src/conductor/inner.rs
```diff
@@ -162,17 +173,198 @@ fn check_for_restart(err: &eyre::Report) -> bool {
     false
 }
 
+fn should_restart_or_shutdown(
+    config: &Config,
+    status: &crate::executor::State,
+) -> eyre::Result<RestartOrShutdown> {
+    let Some(rollup_stop_block_number) = status.rollup_end_block_number() else {
+        return Err(eyre!(
```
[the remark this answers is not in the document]
ethanoroshiba: 80238e7

crates/astria-conductor/src/conductor/inner.rs
```diff
@@ -162,17 +173,198 @@ fn check_for_restart(err: &eyre::Report) -> bool {
...
+            "executor exited with a success value even though it was not configured to run with a \
+             stop height and even though it received no shutdown signal; this should not happen"
+        ));
+    };
+
+    match config.execution_commit_level {
+        crate::config::CommitLevel::FirmOnly | crate::config::CommitLevel::SoftAndFirm => {
```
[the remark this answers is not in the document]
ethanoroshiba: 80238e7

crates/astria-conductor/src/executor/client.rs
```diff
@@ -59,15 +60,18 @@ impl Client {
         })
     }
 
-    /// Calls RPC astria.execution.v1.GetBlock
+    /// Calls RPC astria.execution.v2.GetExecutedBlockMetadata
     #[instrument(skip_all, fields(block_number, uri = %self.uri), err)]
-    pub(crate) async fn get_block_with_retry(&mut self, block_number: u32) -> eyre::Result<Block> {
-        let raw_block = tryhard::retry_fn(|| {
+    pub(crate) async fn get_block_with_retry(
```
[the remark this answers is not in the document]
ethanoroshiba: 80238e7

crates/astria-conductor/src/executor/state.rs
```diff
@@ -74,44 +82,80 @@ impl StateReceiver {
         self.inner.changed().await?;
         Ok(self.next_expected_soft_sequencer_height())
     }
+
+    pub(crate) fn sequencer_stop_height(&self) -> eyre::Result<Option<NonZeroU64>> {
```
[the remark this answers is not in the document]
ethanoroshiba: Moved check to `try_from_execution_session()` in 80238e7.

crates/astria-conductor/src/executor/state.rs
```diff
@@ -74,44 +82,80 @@ impl StateReceiver {
...
+    }
 }
 
 pub(super) struct StateSender {
     inner: watch::Sender<State>,
 }
 
-fn can_map_firm_to_sequencer_height(
-    genesis_info: &GenesisInfo,
+impl std::fmt::Display for StateSender {
```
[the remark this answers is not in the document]
ethanoroshiba: Used here: https://github.com/astriaorg/astria/blob/80238e703ec55292a3f347b91c6a2cddd9e5e09a/crates/astria-conductor/src/executor/mod.rs#L200
Happy to remove if you prefer

crates/astria-conductor/src/executor/state.rs
```diff
@@ -74,44 +82,80 @@ impl StateReceiver {
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let s = serde_json::to_string(&*self.inner.borrow()).unwrap();
```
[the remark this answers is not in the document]
ethanoroshiba: Unsure why this is here. Handled error in 80238e7.

crates/astria-conductor/src/executor/state.rs
```diff
@@ -195,207 +225,244 @@ macro_rules! forward_impls {
...
+    execution_session_parameters: ExecutionSessionParameters,
     commitment_state: CommitmentState,
-    genesis_info: GenesisInfo,
 }
 
 impl State {
-    pub(super) fn try_from_genesis_info_and_commitment_state(
-        genesis_info: GenesisInfo,
-        commitment_state: CommitmentState,
+    pub(crate) fn try_from_execution_session(
```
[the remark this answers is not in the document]
ethanoroshiba: Should be covered by moving `state` to the top level as you suggested offline. Made this change in 80238e7.

crates/astria-conductor/src/executor/state.rs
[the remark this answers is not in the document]
ethanoroshiba: Discussed offline, moved `state` to top level in 80238e7.

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -117,8 +134,14 @@ impl Executor {
         let reader_cancellation_token = self.shutdown.child_token();
 
         let (firm_blocks_tx, firm_blocks_rx) = tokio::sync::mpsc::channel(16);
-        let (soft_blocks_tx, soft_blocks_rx) =
-            tokio::sync::mpsc::channel(state.calculate_max_spread());
+        let (soft_blocks_tx, soft_blocks_rx) = tokio::sync::mpsc::channel(
+            state
+                .celestia_search_height_max_look_ahead()
+                .try_into()
+                .expect(
```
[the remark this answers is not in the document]
ethanoroshiba: 4f66336

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -365,23 +365,29 @@ impl Initialized {
...
 
-        let genesis_height = self.state.sequencer_genesis_block_height();
-        let block_height = executable_block.height;
-        let Some(block_number) =
-            state::map_sequencer_height_to_rollup_height(genesis_height, block_height)
-        else {
+        let sequencer_start_block_height = self.state.sequencer_start_block_height();
+        let rollup_start_block_number = self.state.rollup_start_block_number();
+        let current_block_height = executable_block.height;
+        let Some(block_number) = state::map_sequencer_height_to_rollup_height(
```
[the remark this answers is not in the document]
ethanoroshiba: Made into a fallible function in 4f66336. Because this function doesn't actually use the state itself (as opposed to mapping the other way), I think it may be best left as a free-standing function.

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -418,22 +424,28 @@ impl Initialized {
             "expected block at sequencer height {expected_height}, but got {block_height}",
         );
 
-        let genesis_height = self.state.sequencer_genesis_block_height();
-        let Some(block_number) =
-            state::map_sequencer_height_to_rollup_height(genesis_height, block_height)
-        else {
+        let sequencer_start_block_height = self.state.sequencer_start_block_height();
+        let rollup_start_block_number = self.state.rollup_start_block_number();
+        let Some(block_number) = state::map_sequencer_height_to_rollup_height(
```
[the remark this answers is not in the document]
ethanoroshiba: 4f66336

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -503,24 +516,30 @@ impl Initialized {
...
             .await
             .wrap_err("failed to run execute_block RPC")?;
 
         self.metrics
             .record_transactions_per_executed_block(n_transactions);
 
         info!(
-            executed_block.hash = %telemetry::display::base64(&executed_block.hash()),
-            executed_block.number = executed_block.number(),
+            executed_block_metadata.hash = %telemetry::display::base64(&executed_block_metadata.hash()),
```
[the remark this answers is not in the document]
ethanoroshiba: 4f66336

crates/astria-core/src/execution/v2/mod.rs
SuperFluffy: Yeah, this looks better. Just added one comment on the `FirmExceedsSoft` variant and `error(transparent)` (see there

crates/astria-conductor/CHANGELOG.md
```diff
@@ -15,6 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Update `idna` dependency to resolve cargo audit warning [#1869](https://github.com/astriaorg/astria/pull/1869).
 - Remove panic source on shutdown [#1919](https://github.com/astriaorg/astria/pull/1919).
+- Bump to v2 Execution API, using execution sessions and removing chain id env
```
SuperFluffy, line 18: "Bump" kinda puts it lightly because this is a fairly significant revamp:

```suggestion
Upgrade to `astria.execution.v2` APIs, implement execution sessions, remove env vars for setting chain IDs
```

crates/astria-conductor/src/block_cache.rs
```diff
@@ -74,6 +74,10 @@ impl<T> BlockCache<T> {
             cache: self,
         }
     }
+
```
SuperFluffy, line 77: Add a doc comment on what this does/what it means.

crates/astria-conductor/src/conductor/inner.rs
```diff
@@ -44,14 +48,28 @@ impl std::fmt::Display for RestartOrShutdown {
     }
 }
 
-struct ShutdownSignalReceived;
+pub(crate) enum ExitReason {
```
SuperFluffy, line 51: Different tasks can have different reasons for exits. This combines all of them into one. I don't like that this is state-wide type.

Further, the way that the celestia/sequencer readers are designed now are as subtasks of executor so their reason for going down should never be revealed anyway.

crates/astria-conductor/src/conductor/inner.rs
```diff
@@ -44,14 +48,28 @@ impl std::fmt::Display for RestartOrShutdown {
...
+    ShutdownSignalReceived,
+    ChannelsClosed(Box<State>),
```
SuperFluffy, line 53: This is an implementation detail. It's not clear why an exit reason `channels is closed` carries a "state" value.

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -117,8 +133,15 @@ impl Executor {
         let reader_cancellation_token = self.shutdown.child_token();
 
         let (firm_blocks_tx, firm_blocks_rx) = tokio::sync::mpsc::channel(16);
-        let (soft_blocks_tx, soft_blocks_rx) =
-            tokio::sync::mpsc::channel(state.calculate_max_spread());
+        let (soft_blocks_tx, soft_blocks_rx) = tokio::sync::mpsc::channel(
+            state
+                .celestia_search_height_max_look_ahead()
```
SuperFluffy, line 138: This panics if the max lookahead is set to 0.

I'd prefer if we kept `calculate_max_spread` or something similar and guaranteed that it return a value `>1`

SuperFluffy (commented):
Leaving some comments. Not completely finished with the review yet, but addressing the comments (especially `ExitReason`) should take a little.

On the breaking changes:

+ forward compatibility is not broken for config - new conductor just ignores the two env vars if still supplied
+ just mention that conductor now requires `astria.execution.v2`

crates/astria-auctioneer/src/block/mod.rs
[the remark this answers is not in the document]
ethanoroshiba: I had not added this in the changelog since it is still under the `unreleased` category. Should this be changed?

crates/astria-core/src/execution/mod.rs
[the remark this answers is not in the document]
ethanoroshiba: Added in 7f0619c

crates/astria-core/src/generated/mod.rs
```diff
@@ -55,15 +55,6 @@ pub mod astria {
 
     #[path = ""]
     pub mod execution {
-        pub mod v1 {
-            include!("astria.execution.v1.rs");
```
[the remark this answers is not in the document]
ethanoroshiba: Reinstated the include statement in 7f0619c

crates/astria-conductor/CHANGELOG.md
ethanoroshiba: 7f0619c

crates/astria-conductor/src/block_cache.rs
ethanoroshiba: 7f0619c

crates/astria-conductor/src/conductor/inner.rs
ethanoroshiba: I've separated these out into `executor::ExitStatus` and `ConductorExitReason` in 7f0619c, that way these enums more accurately represent what they are and the components they are associated with. Let me know if this works for you

crates/astria-conductor/src/executor/mod.rs
ethanoroshiba: `calculate_max_spread` did not guarantee that the value was greater than 0 previously, but instead just performed the "magic multiplication" by 6 that we wanted to get rid of. I've added an `ensure` invocation above this to handle the error more gracefully in 7f0619c. Let me know if this works for you!

crates/astria-conductor/src/executor/client.rs
```diff
@@ -59,15 +60,18 @@ impl Client {
         })
     }
 
-    /// Calls RPC astria.execution.v1.GetBlock
+    /// Calls RPC astria.execution.v2.GetExecutedBlockMetadata
     #[instrument(skip_all, fields(block_number, uri = %self.uri), err)]
-    pub(crate) async fn get_block_with_retry(&mut self, block_number: u32) -> eyre::Result<Block> {
-        let raw_block = tryhard::retry_fn(|| {
+    pub(crate) async fn get_executed_block_metadata(
```
ethanoroshiba, line 65: with retry

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -83,17 +97,32 @@ pub(crate) struct Executor {
...
+            ReaderKind::Firm => "firm celestia reader",
+            ReaderKind::Soft => "soft sequencer reader",
+        };
+        f.write_str(msg)
+    }
+}
 
-    pub(crate) async fn run_until_stopped(self) -> eyre::Result<()> {
+impl Executor {
+    pub(crate) async fn run_until_stopped(self) -> eyre::Result<ExitStatus> {
```
SuperFluffy, line 117: Return `Option<State>` instead - `None` if a shutdown signal was received during init or if getching the state failed, `Some(state)` otherwise.

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -585,57 +622,111 @@ impl Initialized {
...
+                             and closing channels"
+                        );
+                        self.reader_cancellation_token.cancel();
+                        self.firm_blocks.close();
+                        self.soft_blocks.close();
+                        Ok(())
+                    }
+
+                    (true, false) => match res {
+                        Ok(Ok(())) => Err(eyre!("task exited with sucess value")),
```
SuperFluffy, line 648: success

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -585,57 +622,111 @@ impl Initialized {
...
+                        Ok(Err(err)) => Err(err).wrap_err("task exited with error"),
+                        Err(err) => Err(err).wrap_err("task panicked"),
+                    }
+                    .wrap_err_with(|| {
+                        format!("task `{task}` exited unexpectedly before stop height was reached")
+                    }),
+
+                    // fall-through case, no stop height was configured
+                    (false, _) => match res {
+                        Ok(Ok(())) => Err(eyre!("task exited with sucess value")),
```
SuperFluffy, line 658: success

crates/astria-conductor/src/conductor/inner.rs
```diff
@@ -108,28 +127,34 @@ impl Inner {
...
     /// Waits 25 seconds for all tasks to shut down before aborting them. 25 seconds
     /// because kubernetes issues SIGKILL 30 seconds after SIGTERM, giving 5 seconds
     /// to abort the remaining tasks.
     #[instrument(skip_all, err, ret(Display))]
     async fn restart_or_shutdown(
         mut self,
-        exit_reason: eyre::Result<ShutdownSignalReceived>,
+        exit_reason: eyre::Result<ConductorExitReason>,
     ) -> eyre::Result<RestartOrShutdown> {
-        self.shutdown_token.cancel();
```
SuperFluffy, line 121: perform the cancellation here, but have `conductor::Inner` track the "public" shutdown token and a "private" child token, which gets also passed to the tasks. Then you can always cancel the child token but still check if the public token was cancelled.

crates/astria-conductor/src/conductor/inner.rs
```diff
@@ -149,30 +174,218 @@ impl Inner {
 }
 
 #[instrument(skip_all)]
-fn check_for_restart(err: &eyre::Report) -> bool {
+fn should_restart_despite_error(err: &eyre::Report) -> bool {
     let mut current = Some(err.as_ref() as &dyn std::error::Error);
     while let Some(err) = current {
         if let Some(status) = err.downcast_ref::<tonic::Status>() {
             if status.code() == tonic::Code::PermissionDenied {
+                info!("received `PermissionDenied` status, attempting new execution session");
```
ethanoroshiba, line 182:
```suggestion
                warn!("received `PermissionDenied` status, attempting new execution session");
```

crates/astria-conductor/src/conductor/inner.rs
```diff
@@ -108,28 +127,34 @@ impl Inner {
...
-                    Err(error)
+        let restart_or_shutdown = 'decide_restart: {
+            if self.shutdown_token.is_cancelled() {
+                break 'decide_restart Ok(RestartOrShutdown::Shutdown);
+            }
+
+            match exit_reason {
+                Ok(exit_reason) => should_restart_or_shutdown(&self.config, &exit_reason),
+                Err(error) => {
+                    error!(%error, "executor failed; checking error chain if conductor should attempt new execution session");
```
SuperFluffy, line 148: move into `should_restart_despite_error` and probably make this into `warn!`

crates/astria-conductor/src/conductor/inner.rs
ethanoroshiba, line 182: Log error always, info if permission denied, error otherwise

crates/astria-core/CHANGELOG.md
```diff
@@ -42,3 +42,4 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Remove inherent methods `Address::try_from_raw` and `Address::to_raw`
   [#1802](https://github.com/astriaorg/astria/pull/1802).
 - Remove `AddressBuilder::with_iter` from public interface [#1802](https://github.com/astriaorg/astria/pull/1802).
+- Removed `execution::v1` in favor of v2 execution API [#2006](https://github.com/astriaorg/astria/pull/2006).
```
SuperFluffy, line 45: Move this into `changed` above

SuperFluffy (approved):
Thank you so much for slogging this out. Please address the last comments.

crates/astria-conductor/src/executor/client.rs
ethanoroshiba: 50b3a45

crates/astria-conductor/src/executor/mod.rs
ethanoroshiba: 50b3a45

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -116,9 +145,22 @@ impl Executor {
 
         let reader_cancellation_token = self.shutdown.child_token();
 
+        // Creating a channel with a buffer size of 0 will panic, so we perform the check here.
+        ensure!(
+            state.celestia_search_height_max_look_ahead() > 0,
```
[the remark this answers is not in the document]
ethanoroshiba: Discussed offline: `NonZeroUsize` stored in state when mode is with firm, otherwise use 1024 as 2x the sequencer rate limit in soft only mode. 50b3a45

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -503,24 +520,30 @@ impl Initialized {
         } = block;
 
         let n_transactions = transactions.len();
-        let sequencer_block_hash = hash.as_bytes().to_vec().into();
+        let sequencer_block_hash = hash.to_string();
```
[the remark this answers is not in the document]
ethanoroshiba: 50b3a45

crates/astria-conductor/src/executor/mod.rs
ethanoroshiba: 50b3a45

crates/astria-conductor/src/conductor/inner.rs
ethanoroshiba: 50b3a45

crates/astria-core/CHANGELOG.md
ethanoroshiba: 50b3a45

crates/astria-conductor/src/executor/client.rs
```diff
@@ -138,6 +139,8 @@ impl Client {
             .collect::<Result<_, _>>()
             .wrap_err("failed to decode tx bytes as RollupData")?;
 
+        let sequencer_block_hash = sequencer_block_hash.to_string();
```
SuperFluffy, line 142: Very nit: I'd remove the binding and just do `sequencer_block_hash: sequencer_block_hash.to_string(),` in when you construct the `ExecuteBlockRequest` further down.

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -114,15 +102,15 @@ impl std::fmt::Display for ReaderKind {
 }
 
 impl Executor {
-    pub(crate) async fn run_until_stopped(self) -> eyre::Result<ExitStatus> {
+    pub(crate) async fn run_until_stopped(self) -> eyre::Result<Option<State>> {
```
SuperFluffy, line 105: Make a doc comment like this:
```suggestion
/// Run until a stop signal is received or until the stop height is reached.
///
/// The stop height is defined by the execution session that the executor task fetches from its rollup node. If the execution session contains no stop height, executor will run until a stop signal is reached or its execution session gets invalidated by the rollup.
///
/// Returns [`State`] when exiting gracefully (stop signal, stop height reached). `None` is returned if a shutdown signal is received during initialization before a rollup state could be established.
pub(crate) async fn run_until_stopped_or_stop_height_reached(self) -> eyre::Result<Option<State>> {
```

(note the change in the name; it's no longer just `until_stopped`, but we have an actual regular exit case)

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -245,6 +220,36 @@ impl Executor {
     }
 }
 
+type BlockChannels = (
```
SuperFluffy, line 223: Really not a fan of single use aliases. They don't provide any additional benefit here.

If you want more structure, just make this a struct:

```rust
struct Channels {
    firm_sender: (),
    firm_receiver: (),
    soft_sender: (),
    soft_receiver: (),
}
```

crates/astria-conductor/src/state.rs
```diff
@@ -284,9 +288,28 @@ impl State {
...
             })?;
         };
+        let non_zero_celestia_search_height_max_look_ahead = if commit_level.is_with_firm() {
+            NonZeroUsize::new(
+                usize::try_from(
+                    execution_session_parameters.celestia_search_height_max_look_ahead(),
+                )
+                .map_err(|_| InvalidState {
+                    map_purpose: "celestia search height max look ahead",
+                    issue: "celestia search height max look ahead overflows usize",
```
SuperFluffy, line 298: add that this should only happen on 32 bit machines.

crates/astria-conductor/src/state.rs
```diff
@@ -284,9 +288,28 @@ impl State {
                 rollup_number: rollup_end_block_number.get(),
             })?;
         };
+        let non_zero_celestia_search_height_max_look_ahead = if commit_level.is_with_firm() {
+            NonZeroUsize::new(
+                usize::try_from(
```
SuperFluffy, line 293: This should actually reject values of `0`. In protobuf, you cannot distinguish between the default of `0` and an actual value of `0`. So when you use it in `executor` and check for this being set, getting the error "value must be set" will be incredibly confusing if the user explicitly set it to 0.

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -699,11 +704,11 @@ impl Initialized {
...
 
     #[instrument(skip_all)]
-    async fn shutdown(mut self, reason: &eyre::Result<ExitStatus>) {
+    async fn shutdown(mut self, reason: &eyre::Result<Option<State>>) {
         let message = "shutting down";
         match &reason {
-            Ok(reason) => {
-                info!(%reason, message);
+            Ok(_) => {
+                info!(reason = "executor exited with no error", message);
```
SuperFluffy, line 711: that's confusing; better avoid giving any reason than this (I understand this came from the redesign of `ExitStatus`; just remove this field if it no longer serves a purpose).

ethanoroshiba: 54d805f

crates/astria-conductor/src/state.rs
ethanoroshiba: Handles each commit level separately in 54d805f.

crates/astria-conductor/src/state.rs
ethanoroshiba: 54d805f

crates/astria-conductor/src/executor/mod.rs
ethanoroshiba: 54d805f

crates/astria-conductor/src/executor/mod.rs
ethanoroshiba: 54d805f

crates/astria-conductor/src/executor/client.rs
ethanoroshiba: 54d805f

crates/astria-conductor/src/executor/mod.rs
ethanoroshiba: [50b3a45](https://github.com/astriaorg/astria/pull/2006/commits/50b3a45e25e073a09527ff1b8908a32277a20c7d)

crates/astria-conductor/src/executor/mod.rs
ethanoroshiba: [50b3a45](https://github.com/astriaorg/astria/pull/2006/commits/50b3a45e25e073a09527ff1b8908a32277a20c7d)

crates/astria-conductor/src/conductor/inner.rs
ethanoroshiba: Opted for `Option<State>` in [50b3a45](https://github.com/astriaorg/astria/pull/2006/commits/50b3a45e25e073a09527ff1b8908a32277a20c7d)

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -220,34 +233,54 @@ impl Executor {
...
+        }
+        CommitLevel::SoftAndFirm => {
+            let channel_size = usize::try_from(state.celestia_search_height_max_look_ahead())
+                .wrap_err(
+                    "celestia search height max look ahead overflows usize, is this a 32 bit \
+                     machine?",
+                )?;
+            ensure!(
+                channel_size > 0,
+                "celestia search height max look ahead must be greater than 0"
```
SuperFluffy, line 268:
```suggestion
"celestia-search-height-max-look-ahead must be greater than 0 if running in soft-and-firm mode; note that unset values in the execution session have a default of 0"
```

aajimal (approved):

Issue #2005, opened 2025-02-27
feat: implement IBC client recovery


joroshiba:
Add a sudo action to support client recovery in sequencer node.

PR #2008, opened 2025-02-28
feat(sequencer)!: implement `RecoverClient` action 

noot:
## Summary
implement `RecoverClient` action which replaces an expired or otherwise inactive IBC client.

this is based off `MsgRecoverClient` implemented in ibc-go v8: https://github.com/cosmos/ibc-go/blob/a71577cc1bf242f8c99360777106b1321503b17f/proto/ibc/core/client/v1/tx.proto#L126

## Background
there is no way to otherwise replace an expired IBC client without creating a new one and creating a governance proposal on the counterparty chain to change the client there, which is a lot more hassle than just being able to replace an expired client on one chain.

## Changes
- implement `RecoverClient` action 
- the implementation is based off the following ibc-go implementation: 
- https://github.com/cosmos/ibc-go/blob/a71577cc1bf242f8c99360777106b1321503b17f/modules/core/02-client/keeper/client.go#L121
- https://github.com/cosmos/ibc-go/blob/a71577cc1bf242f8c99360777106b1321503b17f/modules/light-clients/07-tendermint/light_client_module.go#L179
- https://github.com/cosmos/ibc-go/blob/a71577cc1bf242f8c99360777106b1321503b17f/modules/light-clients/07-tendermint/proposal_handle.go#L29
- i had to export some types from penumbra (ConsensusStateWriteExt, ClientState) which is why it's on a different commit (commit is based off v0.80.7)

## Testing
not yet, will be done on devnet

## Changelogs
changelogs updated

## Breaking Changelist
- a new action is added, which is a breaking change.

## Related Issues
closes https://github.com/astriaorg/astria/issues/2005


crates/astria-sequencer/src/ibc/mod.rs
```diff
@@ -2,6 +2,7 @@ pub(crate) mod component;
 pub(crate) mod host_interface;
 pub(crate) mod ics20_transfer;
 mod state_ext;
+
```
joroshiba, line 5: random newline

crates/astria-sequencer/src/fees/action.rs
joroshiba, line 1: this seems ancillary? Perhaps something went bad in merge?

crates/astria-sequencer/app-genesis-state.json
joroshiba, line 1: Merge artifact, I believe we might have moved towards autogenerating the genesis or something like that?

joroshiba (commented):
Don't think this encompasses feedback from other PR, and I think there may be some merge issues.

crates/astria-sequencer/src/fees/action.rs
noot: deleted

crates/astria-sequencer/app-genesis-state.json
noot: deleted

proto/protocolapis/astria/protocol/transaction/v1/action.proto
```diff
@@ -254,9 +255,18 @@ message FeeChange {
...
 message IbcSudoChange {
   astria.primitive.v1.Address new_address = 1;
 }
+
+message RecoverClient {
+  // the client identifier for the client to be updated
+  string subject_client_id = 1;
+  // the substitute client identifier for the client which will replace the subject
+  // client
+  string substitute_client_id = 2;
```
joroshiba, line 271: nit: I think subject & substitute are kinda confusing. Preference for `client_id` and `replacement_client_id`.

noot: don't really like just `client_id` as it's too ambiguous imo and will be more confusing reading the code. could do `client_id_to_replace` and `substitute_client_id`, or `subject_client_id` and `replacement_client_id`?

crates/astria-sequencer/src/action_handler/impls/recover_client.rs
```diff
@@ -0,0 +1,122 @@
...
+            )
+            .await
+            .map_err(anyhow_to_eyre)
+            .wrap_err(
+                "failed to put
+        verified consensus state",
+            )?;
+
+        subject_client_state.latest_height = substitute_client_state.latest_height;
+        subject_client_state.chain_id = substitute_client_state.chain_id;
```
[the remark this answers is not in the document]
noot: the spec allows the chain ID to be changed - https://ibc.cosmos.network/architecture/adr-026-ibc-client-recovery-mechanisms/ although i'm not sure exactly why. this might be for the case where the counterparty forks and changes their chain ID, but we didn't do a client upgrade in time, which would invalidate the client.

crates/astria-sequencer/src/fees/query.rs
Fraser999, line 1: We should add an entry for this new fee to the `AllFeeComponents` struct.

I think that will also mean we'll need to add it to the test helpers `expected_fees` and `write_all_the_fees` in service/info/mod.rs.

crates/astria-sequencer/src/action_handler/impls/recover_ibc_client.rs
```diff
@@ -0,0 +1,122 @@
...
+            .await
+            .wrap_err("failed to get timestamp")?;
+        let subject_client_status = state
+            .get_client_status(&self.subject_client_id, timestamp)
+            .await;
+
+        ensure!(
+            subject_client_status != ClientStatus::Active,
+            "cannot recover an active client",
+        );
```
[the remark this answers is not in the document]
noot: the only statuses are active, frozen, expired, and unknown, where unknown is only returned if there's an error calculating the status: https://github.com/penumbra-zone/penumbra/blob/bf46dfa1cc3b9fd59fd2c2f1edc1747a69d7aa45/crates/core/component/ibc/src/component/client.rs#L270 

i can make this stricter to error if status is unknown, as that seems to be an error case. however, if a client status is unknown, that may be a case where we want to replace it as something has gone very wrong.

crates/astria-sequencer/src/action_handler/impls/recover_ibc_client.rs
```diff
@@ -0,0 +1,122 @@
...
+    transaction::StateReadExt as _,
+};
+
+#[async_trait::async_trait]
+impl ActionHandler for action::RecoverIbcClient {
+    async fn check_stateless(&self) -> Result<()> {
+        Ok(())
+    }
+
+    async fn check_and_execute<S: StateWrite>(&self, mut state: S) -> Result<()> {
```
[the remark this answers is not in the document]
noot: the ADR allows trusting period to be changed:

> In addition, TrustingPeriod was initially not allowed to be updated by a client upgrade proposal. However, due to the number of situations experienced in production where the TrustingPeriod of a client should be allowed to be updated because of ie: initial misconfiguration for a canonical channel, governance should be allowed to update this client parameter.

https://ibc.cosmos.network/architecture/adr-026-ibc-client-recovery-mechanisms/

will add the other checks

crates/astria-sequencer/src/action_handler/impls/recover_ibc_client.rs
```diff
@@ -0,0 +1,167 @@
...
+
+        client_state.latest_height = replacement_client_state.latest_height;
+        client_state.trusting_period = replacement_client_state.trusting_period;
+        client_state.chain_id = replacement_client_state.chain_id;
+        state.put_client(&self.client_id, client_state);
+        Ok(())
+    }
+}
+
+fn ensure_required_client_state_fields_match(
```
joroshiba, line 122: nit: probably add a comment above which notes the spec which defines what checks are needed?

proto/protocolapis/astria/protocol/transaction/v1/action.proto
```diff
@@ -255,18 +255,28 @@ message FeeChange {
...
+// `client_id` must not be active, and `replacement_client_id` must be
+// active.
+//
+// This is based on https://buf.build/cosmos/ibc/docs/main:ibc.core.client.v1#ibc.core.client.v1.MsgRecoverClient
+// Although the field names are different, the first two fields are
+// still wire-compatible.
+message RecoverIbcClient {
   // the client identifier for the client to be updated
-  string subject_client_id = 1;
+  string client_id = 1;
```
joroshiba, line 278: Thanks for satisfying my naming neuroticism :)

joroshiba (approved):
API changes approved, think you will still need @Fraser999 approval for sequencer/core changes.

Note: I believe you need to commit the generated files.

crates/astria-sequencer/src/action_handler/impls/recover_client.rs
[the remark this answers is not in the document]
joroshiba: Resolving for sticking to spec

crates/astria-sequencer/src/action_handler/impls/recover_ibc_client.rs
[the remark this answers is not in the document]
Fraser999: Hmm - I see your point.

There's a fifth variant `Unauthorized`, but it can't be in that state here at least going by the current penumbra code.

This is probably fine to leave as is, but maybe worth leaving a code comment as to why we're slightly deviating from the spec?

Fraser999 (approved):
LGTM - thanks for updating.  Only (non-blocking) remaining point is a request to add a comment.

PR #2012, opened 2025-03-06
fix(sequencer): mempool pending nonce calcuations

### LINKED ISSUE #2011: Modify OptimisticBlockService API to respond with nonce for Auctioneer account
TODO: More info

https://buf.build/astria/sequencerblock-apis/docs/main:astria.sequencerblock.optimistic.v1alpha1#astria.sequencerblock.optimistic.v1alpha1.OptimisticBlockService

joroshiba:
## Summary
Updates when the mempool is rebalanced in order to solve for race condition in which call falls through to state before it has been written, by rebalancing after commit and updates the mempool methods to return the pending account nonce instead of the highest nonce in the pending mempool. 

## Background
Nonce is defined as the number of transactions which have been executed on the account. The pending nonce should then be the number of transactions executed + the number of transactions sitting in pending. Our API was not doing this correctly. Pending nonce on the mempool was being defined as the transaction with the highest nonce as opposed to what the account nonce will be after all pending transactions.

Additionally, our mempool was being rebalanced before state was committed, this can result in a race condition where you request the pending nonce after rebalancing has occured but before committed state has occured resulting in fetching not the pending nonce but the current nonce. 

## Changes
- Move mempool maintenance to after `commit` from finalized block
- Mempool defines pending nonce as pending account nonce instead of highest nonce in queue

## Testing
CI/CD tests, which have been updated.

## Changelogs
Changelogs updated.

## Related Issues
closes #2011

### DIFF
--- a/crates/astria-sequencer/CHANGELOG.md
+++ b/crates/astria-sequencer/CHANGELOG.md
@@ -9,9 +9,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Ensure getPendingNonce gRPC returns the correct nonce [#2012](https://github.com/astriaorg/astria/pull/2012).
+
 ### Added
 
 - Implement `astria.sequencerblock.optimistic.v1alpha1.OptimisticBlockService` [#1839](https://github.com/astriaorg/astria/pull/1839).
+- Add ASTRIA_SEQUENCER_ABCI_LISTEN_URL config variable [#1877](https://github.com/astriaorg/astria/pull/1877)
 
 ### Changed
 
@@ -30,10 +35,6 @@ of signer [#1905](https://github.com/astriaorg/astria/pull/1905).
 ### Removed
 
 - Remove ASTRIA_SEQUENCER_LISTEN_ADDR config variable [#1877](https://github.com/astriaorg/astria/pull/1877)
-
-### Added
-
-- Add ASTRIA_SEQUENCER_ABCI_LISTEN_URL config variable [#1877](https://github.com/astriaorg/astria/pull/1877)
 
 ## [1.0.0] - 2024-10-25
 
--- a/crates/astria-sequencer/src/mempool/mod.rs
+++ b/crates/astria-sequencer/src/mempool/mod.rs
@@ -417,16 +417,16 @@ impl Mempool {
 
             if demotion_txs.is_empty() {
                 // nothing to demote, check for transactions to promote
-                let highest_pending_nonce = pending
+                let pending_nonce = pending
                     .pending_nonce(address)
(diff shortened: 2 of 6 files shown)


crates/astria-sequencer/src/mempool/mod.rs
```diff
@@ -417,16 +417,16 @@ impl Mempool {
 
             if demotion_txs.is_empty() {
                 // nothing to demote, check for transactions to promote
-                let highest_pending_nonce = pending
+                let pending_nonce = pending
                     .pending_nonce(address)
-                    .map_or(current_nonce, |nonce| nonce.saturating_add(1));
+                    .map_or(current_nonce, |nonce| nonce);
```
joroshiba, line 422: This was a misnomer previously,  and the addition done papering over that our pending nonce math was wrong.

Moved the addition into the mempool so the math is correct throughout codebase.

crates/astria-sequencer/src/mempool/mod.rs
```diff
@@ -417,16 +417,16 @@ impl Mempool {
...
 
                 let remaining_balances =
                     pending.subtract_contained_costs(address, current_balances.clone());
-                let promtion_txs =
-                    parked.find_promotables(address, highest_pending_nonce, &remaining_balances);
+                let promotion_txs =
```
joroshiba, line 426: Just a typo I found.

crates/astria-sequencer/src/mempool/transactions_container.rs
```diff
@@ -208,8 +208,15 @@ pub(super) struct PendingTransactionsForAccount {
 }
 
 impl PendingTransactionsForAccount {
-    fn highest_nonce(&self) -> Option<u32> {
-        self.txs.last_key_value().map(|(nonce, _)| *nonce)
+    fn pending_account_nonce(&self) -> Option<u32> {
```
joroshiba, line 211: renamed to make it clearer

crates/astria-sequencer/CHANGELOG.md
```diff
@@ -9,9 +9,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 ## [Unreleased]
 
+### Fixed
+
+- Ensure getPendingNonce gRPC returns the correct nonce [#2012](https://github.com/astriaorg/astria/pull/2012).
+
 ### Added
 
 - Implement `astria.sequencerblock.optimistic.v1alpha1.OptimisticBlockService` [#1839](https://github.com/astriaorg/astria/pull/1839).
+- Add ASTRIA_SEQUENCER_ABCI_LISTEN_URL config variable [#1877](https://github.com/astriaorg/astria/pull/1877)
```
joroshiba, line 19: noticed we had two added sections in changelog, consolidated.

Fraser999 (approved):

PR #2015, opened 2025-03-06
chore(sequencer): refactor address base prefix checks

### LINKED ISSUE #2014: abstract action base prefix checks
We currently perform all address base prefix checks manually in each `check_and_execute` implementation. We should abstract these such that they are performed for every action automatically via something like trait impls for all actions. This will make it more difficult to forget to implement these checks, since the trait will be required, and will also allow for a more streamlined unit-testing process of these checks.

ethanoroshiba:
## Summary
Pulls address base prefix checks for each action into a separate trait.

## Background
It's easy to miss checks which we should be performing when writing a new action's `check_and_execute` trait method. This change pulls the address base prefix checks for each action into a separate trait so that they can be automatically performed for every action upon execution as well as tested in a more streamlined way.

## Changes
- Removed base prefix checks and tests from action `ActionHandler` impls.
- Added new module `address_check` to `address`, with traits `Addresses` and `AddressCheck: Addresses`. Added impls of these for all existing actions.
- Added call to `AddressCheck::check_addresses()` in `action_handler::transaction::check_execute_and_pay_fees()`.

## Testing
- Added unit tests for the base prefix checks for all actions which contain addresses
- Added app unit test to ensure that `check_addresses()` is called under `execute_transaction()`.

## Changelogs
No updates required.

## Related Issues
closes #2014


noot (approved):
looks good!

ethanoroshiba:
Closing as this will be covered in #2142

PR #2016, opened 2025-03-06
chore(bridge-withdrawer): change batch value metric to a histogram

joroshiba:
## Summary
Replaces the `batch_total_settled_value` gauge with `batch_settled_value` histogram.

## Background
The existing metric `batch_total_settled_value` has not been useful, because gauges aren't very good for fast data points information. Histograms are generally better for this data as we can grab aggregate percentage data on what is happening per batch, as well as a total sum for measuring increase over time. 

## Changes
- Replace the `batch_total_settled_value` gauge with `batch_settled_value` histogram
- Added label for metric based on the denom

## Testing
How are these changes tested?

## Changelogs
Changelogs Updated.

## Metrics
- `batch_settled_value` histogram added.

## Breaking Changes
While adding and removing a metric could be considered a breaking change, we are the only consumers here and stopped using the metric because it pretty simply did not work.

### DIFF
--- a/crates/astria-bridge-withdrawer/CHANGELOG.md
+++ b/crates/astria-bridge-withdrawer/CHANGELOG.md
@@ -9,10 +9,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Histogram etric `batch_settled_value` now a histogram [#2016](https://github.com/astriaorg/astria/pull/2016).
+
 ### Changed
 
 - Update `idna` dependency to resolve cargo audit warning [#1869](https://github.com/astriaorg/astria/pull/1869).
-- Metric `batch_total_settled_value` now a histogram
+
+### Removed
+
+- Metric `batch_total_settled_value` removed in favor of `batch_settled_value` [#2016](https://github.com/astriaorg/astria/pull/2016).
 
 ## [1.0.1] - 2024-11-01
 
--- a/crates/astria-bridge-withdrawer/CHANGELOG.md
+++ b/crates/astria-bridge-withdrawer/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Histogram etric `batch_settled_value` now a histogram [#2016](https://github.com/astriaorg/astria/pull/2016).
+- Histogram metric `batch_settled_value` [#2016](https://github.com/astriaorg/astria/pull/2016).
 
 ### Changed
 
@@ -19,7 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Removed
 
-- Metric `batch_total_settled_value` removed in favor of `batch_settled_value` [#2016](https://github.com/astriaorg/astria/pull/2016).
+- Gauage metric `batch_total_settled_value`  [#2016](https://github.com/astriaorg/astria/pull/2016).
 
 ## [1.0.1] - 2024-11-01
 
--- a/crates/astria-bridge-withdrawer/CHANGELOG.md
+++ b/crates/astria-bridge-withdrawer/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Update `idna` dependency to resolve cargo audit warning [#1869](https://github.com/astriaorg/astria/pull/1869).
+- Metric `batch_total_settled_value` now a histogram
 
 ## [1.0.1] - 2024-11-01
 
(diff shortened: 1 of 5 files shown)


WafflesVonMaple (approved):

crates/astria-bridge-withdrawer/CHANGELOG.md
```diff
@@ -9,10 +9,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Histogram etric `batch_settled_value` now a histogram [#2016](https://github.com/astriaorg/astria/pull/2016).
```
Fraser999, line 14:
```suggestion
- Metric `batch_settled_value` now a histogram [#2016](https://github.com/astriaorg/astria/pull/2016).
```

crates/astria-bridge-withdrawer/CHANGELOG.md
```diff
@@ -9,10 +9,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
 ### Changed
 
 - Update `idna` dependency to resolve cargo audit warning [#1869](https://github.com/astriaorg/astria/pull/1869).
 
+### Removed
+
+- Metric `batch_total_settled_value` removed in favor of `batch_settled_value` [#2016](https://github.com/astriaorg/astria/pull/2016).
```
SuperFluffy: Instead of adding & removed, should we put this under "changed"?

crates/astria-bridge-withdrawer/tests/blackbox/helpers/test_bridge_withdrawer.rs
```diff
@@ -74,25 +73,6 @@ pub(crate) const DEFAULT_IBC_DENOM: &str = "transfer/channel-0/utia";
 pub(crate) const SEQUENCER_CHAIN_ID: &str = "test-sequencer";
 const ASTRIA_ADDRESS_PREFIX: &str = "astria";
 
-static TELEMETRY: LazyLock<()> = LazyLock::new(|| {
```
SuperFluffy, line 77: Why is this removed? The change seems to have nothing to do with this change?

crates/astria-bridge-withdrawer/tests/blackbox/helpers/test_bridge_withdrawer.rs
```diff
@@ -287,12 +266,28 @@ impl TestBridgeWithdrawerConfig {
             metrics_http_listener_addr: String::new(),
             pretty_print: true,
         };
+        if std::env::var_os("TEST_LOG").is_some() {
```
SuperFluffy, line 269: This is bad, we cannot do that.

The reason why we had a lazylock is that the tracing subscriber sits at a global static mwmory location.

The present change can cause tests to fail/panic.

SuperFluffy (changes requested):
The changes to how the global tracing subscriber is set in tests can break tests - likely not in ci because we use nextest but in local tests these are run with just `caego test`.

This is a defect in how the "config" is passed to the telemetry/metrics init function which must be fixed.

crates/astria-bridge-withdrawer/tests/blackbox/helpers/test_bridge_withdrawer.rs
joroshiba: Can't pass the config into the function here.

I can strip out the new functionality of having the label off the config to roll this back.

crates/astria-bridge-withdrawer/tests/blackbox/helpers/test_bridge_withdrawer.rs
```diff
@@ -75,20 +75,40 @@ pub(crate) const SEQUENCER_CHAIN_ID: &str = "test-sequencer";
 const ASTRIA_ADDRESS_PREFIX: &str = "astria";
 
 static TELEMETRY: LazyLock<()> = LazyLock::new(|| {
+    let config = Config {
```
SuperFluffy, line 78: Can you move this and the config padded to the "test bridge withdrawer" to a free standing `fn config() -> Config` so they share the same definition?

SuperFluffy (approved):
That works for me. We should address the way the metrics are populated in the medium term

Fraser999 (approved):
Agree with @SuperFluffy's point about deduplicating the construction of the `Config` for tests.  Otherwise LGTM.  Optimistic approval pending that update.

crates/astria-bridge-withdrawer/CHANGELOG.md
joroshiba: In the past I did that and we decided to do an add and remove.

crates/astria-bridge-withdrawer/tests/blackbox/helpers/test_bridge_withdrawer.rs
joroshiba: They do not share the same definition, the other config relies on items instantiated during the test and can't be a static item.

joroshiba:
closed in 7 days.

joroshiba:

PR #1923, opened 2025-01-22
chore(core, sequencer): rename connect oracle to price feed

Fraser999:
## Summary
The term "connect" in the context of the price feed oracle has been replaced with "price feed".

Note that this PR is based on top of #1917, so is effectively blocked by it.

## Background
Using the product's name "connect" was confusing.  There was a discussion around replacing this with "oracle", but given that we may provide other oracles in the future, it was decided to rename this to "price feed" and nest the instances in a new module "oracles".

## Changes
- Renamed modules, types, variables and comments.  There are no changes to functionality.

## Testing
No additional tests required.

## Changelogs
No updates required - the changelogs will be updated once the feature is complete.

## Breaking Changelist
The feature branch includes breaking changes already, this PR only changes a few of these - no further breakages.

## Related Issues
Closes #1881.

PR #2030, opened 2025-03-12
(sequencer): provide system test for sequencer upgrade

### LINKED ISSUE #2004: Provide system tests for upgrade logic
We want to implement tests as summarised in [the Sequencer Upgrades design doc](https://www.notion.so/astria-org/Sequencer-Upgrades-1896bd31a90c803e859ac5e41c18105d#1896bd31a90c80539590f2033ac68b8e).

Fraser999:
## Summary
This introduces a new system test framework, with a relatively comprehensive test implemented for the first planned sequencer upgrade.  The test covers points 2 and 3 of [the Testing Plan Summary](https://www.notion.so/astria-org/Sequencer-Upgrades-1896bd31a90c803e859ac5e41c18105d?pvs=97#1896bd31a90c80539590f2033ac68b8e).

Note that this PR is based on top of #1923, which itself is based on top of #1917, so is effectively blocked by them.

## Background
Currently, smoke-tests are provided as inline bash in the repo's Just files.  Testing the new sequencer upgrade functionality using the same technique is undesirable, since more complex testing is required, and ideally it should be easily extensible to accommodate future upgrades.

The new upgrade test is written in Python and provided in a new top-level folder `/system-tests`.  There is a README.md provided.  It should be noted that the first set of instructions for running the test assume that the upgrades functionality has been merged to `main`, meaning that running the test with `latest` images should succeed.  This is not the case right now, so to test locally, the second set of instructions should be followed.

This test is not added to CI as part of this PR.  It is probably worth making that change, but in a follow-up PR.

## Changes
- Provided new system tests folder with an upgrade test at `/system-tests/sequencer_upgrade_test.py`. This runs checks which should be applicable to all future upgrades, and should need minimal changes to accommodate these.
- Provided `/system-tests/sequencer_upgrade1_checks.py` which are checks specific to `Upgrade1`.  The checks are invoked from the main test.  It is intended that each future upgrade will have a new file with checks specific to it added to this folder, and similarly being invoked from the main test, hence reducing churn in the main test's code.
- Provided helper objects for controlling sequencer nodes, evm rollup, and making astria-cli calls.  These should be able to be used to replace our existing smoke tests eventually.
- Provided a stripped-down copy of the `sequencerblock/v1/service.proto` in `/system-tests/helpers/proto/service.proto`.  This is an unfortunate result of the limitations of working with protoc for Python.  If there is a way to avoid duplicating this proto code, I'd gladly remove this file.
- Fixed an issue in the production logic which the test exposed. Specific to Upgrade1, we need to fetch the current consensus params from the connected CometBFT instance in order to update them (enabling the vote extensions). This works for sequencers which have been updated and provided with the correct upgrades.json before the chain reaches the activation height - they successfully execute the upgrade and run on.  However, if an operator doesn't install the new sequencer binary and upgrades.json before the upgrade activates, that sequencer goes into a crashloop at the activation height, since it disagrees with its peers about the block hash.  Upon then installing the new binary and upgrades.json, the sequencer falls into a chicken-and-egg scenario: its CometBFT starts by sending a FinalizeBlock for the last block it was dealing with - the one in which the upgrade should be applied. The sequencer now correctly tries to apply the upgrade, but in doing so needs to fetch the consensus params from CometBFT.  However, CometBFT hasn't started its server yet as it's still awaiting the response from its FinalizeBlock request, hence the processes are deadlocked.  To fix this, I provided a fallback set of hard-coded consensus params (taken from Mainnet and Testnet values) to be used if fetching from CometBFT fails.  Although hard-coded values isn't ideal, this seems to be a reasonable trade-off in this case, given the complexity of alternatives (e.g. fetching from a peer's CometBFT) and the low impact of this failing (if the hard-coded values don't match the rest of the network's, the sequencer will continue to fail to finalize that block, and the operator will need to re-sync from a snapshot or genesis).

## Testing
These are tests.

## Changelogs
No updates required.

## Related Issues
Closes #2004.


joroshiba (approved):
Approved follow up items:
- merge more of the smoke style tests down into the system tests
- consider removing the genesis options from chart for pricefeeds
- improvements to the proto usage in system tests

PR #2034, opened 2025-03-14
fix(charts): broken helper in relayer

joroshiba:
## Summary
Fixes helper in the chart which didn't have matching value types


warhod (approved):
LGTM

PR #2043, opened 2025-03-17
chore(composer): fix flaky test

### LINKED ISSUE #2041: composer: `two_rollup_data_submissions_single_bundle` test is flaky
The test `composer::executor::two_rollup_data_submissions_single_bundle` is flaky for me locally, with the following output:

```text
thread 'executor::two_rollup_data_submissions_single_bundle' panicked at crates/astria-composer/tests/blackbox/executor.rs:166:6:
called `Result::unwrap()` on an `Err` value: Elapsed(())
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    executor::two_rollup_data_submissions_single_bundle
```

ethanoroshiba:
## Summary
Fixed flaky test `composer::executor::two_rollup_data_submissions_single_bundle`.

## Background
This test previously failed semi-frequently for me locally. The issue first stemmed from the submission timeout period elapsing, but after increasing the time to span multiple blocks, it appeared that the submissions were ending up in different transactions. This seemed that it was likely due to the block timer ticking in between when the two submissions were made. As such, spinning them up concurrently appears to fix the problem.

## Changes
- Change rollup transaction submission to be concurrent as opposed to in series to avoid the block timer ticking in between submissions.

## Testing
No failures after testing >100 times.

## Changelogs
No updates required.

## Related Issues
closes #2041

### DIFF
--- a/crates/astria-composer/tests/blackbox/executor.rs
+++ b/crates/astria-composer/tests/blackbox/executor.rs
@@ -14,6 +14,7 @@ use astria_core::{
     },
     protocol::transaction::v1::action::RollupDataSubmission,
 };
+use futures::future::join;
 use tokio::time;
 
 use crate::helper::{
@@ -113,7 +114,7 @@ async fn bundle_triggered_by_block_timer() {
 #[tokio::test]
 async fn two_rollup_data_submissions_single_bundle() {
     let test_composer = spawn_composer(&["test1"], None, true).await;
-    let mut composer_client = GrpcCollectorServiceClient::connect(format!(
+    let composer_client = GrpcCollectorServiceClient::connect(format!(
         "http://{}",
         test_composer.grpc_collector_addr
     ))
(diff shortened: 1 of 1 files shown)


crates/astria-composer/tests/blackbox/executor.rs
```diff
@@ -137,33 +138,42 @@ async fn two_rollup_data_submissions_single_bundle() {
         fee_asset: "nria".parse().unwrap(),
     };
 
+    let submit_fut_1 = {
```
joroshiba, line 141: I think it would be good to have a comment here on why we need to have the two separate futures.

joroshiba (approved):
Approved, but can we get a comment in here that explains why the futures are needed?

crates/astria-composer/tests/blackbox/executor.rs
ethanoroshiba: Done in 4e6fec6

PR #2048, opened 2025-03-20
chore(sequencer-client): add timeout to 'wait_for_tx_inclusion'

### LINKED ISSUE #1949: bug(sequencer-client): client loops forever if transaction execution fails in sequencer

ethanoroshiba:
## Summary
Added a timeout of 240s to the `wait_for_tx_inclusion` method of `SequencerClient`.

## Background
The `SequencerClient` would previously poll the `tx` RPC indefinitely, so if the transaction failed for whatever reason (or even if it failed to submit), any component that calls the method `wait_for_tx_inclusion` would be halted.

## Changes
- Added timeout to `wait_for_tx_inclusion` of 240s (maximum time allowed in mempool), returning `tendermint_rpc::Error::Timeout` if it doesn't succeed within that time.

## Testing
Added unit test to ensure that the timeout returns an error correctly.

## Changelogs
No updates required

## Related Issues
closes #1949

### DIFF
--- a/crates/astria-sequencer-client/src/extension_trait.rs
+++ b/crates/astria-sequencer-client/src/extension_trait.rs
@@ -674,7 +674,7 @@ pub trait SequencerClientExt: Client {
     async fn wait_for_tx_inclusion(
         &self,
         tx_hash: tendermint::hash::Hash,
-    ) -> tendermint_rpc::endpoint::tx::Response {
+    ) -> Result<tendermint_rpc::endpoint::tx::Response, Error> {
         use std::time::Duration;
 
         use tokio::time::Instant;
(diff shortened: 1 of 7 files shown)


crates/astria-sequencer-client/src/extension_trait.rs
```diff
@@ -687,11 +687,18 @@ pub trait SequencerClientExt: Client {
         const START_LOGGING_DELAY: Duration = Duration::from_millis(2000);
         // The minimum duration between logging errors.
         const LOG_ERROR_INTERVAL: Duration = Duration::from_millis(2000);
+        // The maximum amount of time to wait for a transaction to be included.
+        #[cfg(not(test))]
+        const MAX_WAIT_TIME: Duration = Duration::from_secs(240);
```
ethanoroshiba, line 692: Happy to change this number, this is just how long it takes transactions to be cleared from mempool currently

SuperFluffy (changes requested):
This should not be part of the extension trait. The future should be timed out at the call site.

ethanoroshiba:
Closing to move timeouts to callers (will split into separate PRs for each)

PR #2049, opened 2025-03-20
chore(bridge-withdrawer, sequencer-client, cli): move retry logic for `wait_for_tx_invlusion` to calling components

### LINKED ISSUE #1949: bug(sequencer-client): client loops forever if transaction execution fails in sequencer

ethanoroshiba:
## Summary
Moves retry and logging handling of the `tx` RPC and its calling `SequencerClientExt::get_tx()` to its calling components.

## Background
Having the retry and logging logic in the `SequencerClientExt` trait didn't match the precedent set by the rest of the trait, which provides more direct access to the sequencer RPCs as opposed to implementing its own retry logic. This should be left up to the caller of the method.

Additionally, a timeout needed to be added to this functionality, since a failed transaction for any reason would result in the called components (bridge withdrawer and CLI) halting.

## Changes
- Renamed `SequencerClientExt::wait_for_tx_inclusion()` to `get_tx()`.
- Removed retry and logging logic from `SequencerClientExt::get_tx()` and placed it in helper functions in both `astria-bridge-withdrawer` and `astria-sequencer-cli`
- Added timeouts to `get_tx()` calls.

## Testing
- Updated `astria_sequencer_client::tests::http::wait_for_tx_inclusion` accordingly.

## Changelogs
No updates required

## Metrics

## Breaking Changelist

## Related Issues
closes #1949

### DIFF
--- a/crates/astria-bridge-withdrawer/src/bridge_withdrawer/submitter/mod.rs
+++ b/crates/astria-bridge-withdrawer/src/bridge_withdrawer/submitter/mod.rs
@@ -39,6 +39,10 @@ use state::State;
 use tokio::{
     select,
     sync::mpsc,
+    time::{
+        self,
+        Instant,
+    },
 };
 use tokio_util::sync::CancellationToken;
 use tonic::transport::Channel;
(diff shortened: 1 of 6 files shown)


Fraser999 (approved):

joroshiba:
closed in 7 days.

PR #2051, opened 2025-03-22
fix(bridge-withdrawer): inverted comments for `no_frost_threshold_signing` in Astria Bridge Withdrawer

shak58:
## Summary
There is discrepancy between the code and the corresponding documentation of the `no_frost_threshold_signing` variable or ASTRIA_BRIDGE_WITHDRAWER_NO_FROST_THRESHOLD_SIGNING in the Astria Bridge Withdrawer. 
Currently, comments in `local.env.example` and `config.rs` imply that setting `no_frost_threshold_signing` to `true` *enables* FROST threshold signing, while the code implies `true` actually *disables* it.

## Background
As demonstrated in [astria-bridge-withdrawer/src/bridge_withdrawer/submitter/signer/mod.rs](https://github.com/astriaorg/astria/blob/main/crates/astria-bridge-withdrawer/src/bridge_withdrawer/submitter/signer/mod.rs):

```rs
impl Builder {
    pub(super) fn try_build(self) -> eyre::Result<Signer> {
        let Self {
            no_frost_threshold_signing,
            frost_min_signers,
            frost_public_key_package_path,
            frost_participant_endpoints,
            sequencer_key_path,
            sequencer_address_prefix,
        } = self;
        let inner = if no_frost_threshold_signing {
            SignerInner::Key(Box::new(
                key::Key::builder()
                    .path(&sequencer_key_path)
                    .prefix(sequencer_address_prefix)
                    .try_build()
                    .wrap_err_with(|| {
                        format!(
                            "failed to load sequencer private key from path `{sequencer_key_path}`"
                        )
                    })?,
            ))
        } else {
            SignerInner::Frost(
                frost::Builder {
                    frost_min_signers,
                    frost_participant_endpoints,
                    frost_public_key_package_path,
                    sequencer_address_prefix,
                }
                .try_build()
                .wrap_err("failed to construct frost threshold signer")?,
            )
        };
        Ok(Signer {
            inner,
        })
    }
}
```
This is the reverse of what the inline comments say.
## Changes
- Corrected the comment in [`local.env.example`](https://github.com/astriaorg/astria/blob/main/crates/astria-bridge-withdrawer/local.env.example).
- Revised comment in [`config.rs`](https://github.com/astriaorg/astria/blob/main/crates/astria-bridge-withdrawer/src/config.rs)


crates/astria-bridge-withdrawer/local.env.example
```diff
@@ -31,7 +31,7 @@ ASTRIA_BRIDGE_WITHDRAWER_SEQUENCER_COMETBFT_ENDPOINT="http://127.0.0.1:26657"
 # Chain ID of the sequencer chain which transactions are submitted to.
 ASTRIA_BRIDGE_WITHDRAWER_SEQUENCER_CHAIN_ID="astria"
 
-# Set to true to enable frost threshold signing.
```
SuperFluffy, line 34: Let's make this:
```suggestion
# Setting this to `true` disables frost threshold signing, falling back to signing via the key at `ASTRIA_BRIDGE_WITHDRAWER_SEQUENCER_KEY_PATH`

File: crates/astria-bridge-withdrawer/src/config.rs [hunk 2]
```diff
@@ -20,7 +20,7 @@ pub struct Config {
     pub sequencer_cometbft_endpoint: String,
     // The chain id of the sequencer chain.
     pub sequencer_chain_id: String,
-    // Set to false to disable frost threshold signing.
```
SuperFluffy, line 23: Adjust to the one above.

SuperFluffy (changes requested):
That's a good catch actually. Thank you for paying attention. :-) I made a small suggestion (it's easier to think about what it means if a value is set).

SuperFluffy (approved):
Great, thanks

warhod (approved):
LGTM

PR #2053, opened 2025-03-24
chore(bridge-withdrawer, composer): encode submission payload only once

### LINKED ISSUE #2050: chore: remove `submit_transaction_sync` from sequencer client extension trait
`SequencerClientExt::submit_transaction_sync()` provides very little additional functionality to the underlying `tendermint_rpc::Client::broadcast_tx_sync()` call. We should remove this superfluous trait method and instead use `broadcast_tx_sync()` where it is called.

https://github.com/astriaorg/astria/blob/9bac0422a4c8be8a850f006f14cda7f7441b8fd7/crates/astria-sequencer-client/src/extension_trait.rs#L659-L664

ethanoroshiba:
## Summary
Changed submission logic to only encode retried transaction submissions to bytes once.

## Background
Previously, Bridge Withdrawer and Composer utilized `SequencerClientExt::submit_transaction_sync()`, which encodes the transaction to bytes within the method. Within a retry configuration, this unnecessarily performs the conversion each time. These changes eliminate these superfluous encodings in favor of retrying with the same bytes each time.

## Changes
- Changed `submit_transaction_sync()` calls to `broadcast_tx_sync()` calls.
- Moved Tx encoding to outside the retry logic.

## Testing
Passing all tests, no additional ones needed since the functionality is the same.

## Changelogs
No updates required.

## Related Issues
closes #2050

### DIFF
--- a/crates/astria-bridge-withdrawer/src/bridge_withdrawer/submitter/mod.rs
+++ b/crates/astria-bridge-withdrawer/src/bridge_withdrawer/submitter/mod.rs
@@ -25,10 +25,14 @@ use astria_eyre::eyre::{
 };
 pub(crate) use builder::Builder;
 pub(super) use builder::Handle;
+use prost::Message as _;
 use sequencer_client::{
-    tendermint_rpc::endpoint::{
-        broadcast::tx_sync,
-        tx,
+    tendermint_rpc::{
+        endpoint::{
+            broadcast::tx_sync,
+            tx,
+        },
+        Client as _,
     },
     Address,
     SequencerClientExt,
@@ -274,7 +278,7 @@ async fn submit_tx(
         .on_retry(
             |attempt,
              next_delay: Option<Duration>,
-             err: &sequencer_client::extension_trait::Error| {
+             err: &sequencer_client::tendermint_rpc::Error| {
                 metrics.increment_sequencer_submission_failure_count();
 
                 let state = Arc::clone(&state);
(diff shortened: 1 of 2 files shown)


crates/astria-sequencer-client/src/extension_trait.rs
```diff
@@ -648,21 +647,6 @@ pub trait SequencerClientExt: Client {
...
 
-    /// Submits the given transaction to the Sequencer node.
-    ///
-    /// This method blocks until the transaction is checked, but not until it's committed.
-    /// It returns the results of `CheckTx`.
-    ///
-    /// # Errors
-    ///
-    /// - If calling the tendermint RPC endpoint fails.
-    async fn submit_transaction_sync(&self, tx: Transaction) -> Result<tx_sync::Response, Error> {
```
SuperFluffy, line 659: This method is actually useful because here the caller need not know about how to transform a `Transaction` into a payload sequencer understands.

I would leave this.

SuperFluffy (changes requested):
I would push back on the primary change of this PR: it is a useful feature for a user not having to know how to transform a `Transaction -> proto Transaction -> bytes` in order to submit it to Sequencer.

With that said - what this PR does reveal is that there are some easy optimizations to be had in composer, bridge-withdrawer (and less so) auctioneer: we don't need to re-encode the bytes every time. It's sufficient to encode them once and resubmit the same bytes on certain kinds of errors (unless we need to bump the nonce, say).

crates/astria-bridge-withdrawer/src/bridge_withdrawer/submitter/mod.rs
```diff
@@ -293,11 +297,12 @@ async fn submit_tx(
...
             },
         );
+    let tx_bytes = tx.to_raw().encode_to_vec();
     let check_tx = tryhard::retry_fn(|| {
         let client = client.clone();
-        let tx = tx.clone();
+        let tx_bytes = tx_bytes.clone();
         let span = info_span!(parent: span.clone(), "attempt send");
-        async move { client.submit_transaction_sync(tx).await }.instrument(span)
+        async move { client.broadcast_tx_sync(tx_bytes).await }.instrument(span)
```
SuperFluffy, line 305: It's a shame that this requires `Vec<u8>` instead of `Bytes`, which would make the clone above much cheaper.

SuperFluffy (approved):

PR #2062, opened 2025-03-27
chore: update chart upgrade format

joroshiba:
Update to fraser PR based on my review

- Updates upgrade file format in charts to support custom vs network like genesis
- Adds a helm clean command that doesn't nuke cluster
- Moves away from using sequencer config for genesis or upgrade information
- uses release images as baseline for test
- reduces resource usage based on my testing that still works.


charts/deploy.just
```diff
@@ -44,7 +44,8 @@ deploy-cluster namespace=defaultNamespace enableHubble=hubbleEnabled:
   helm install cilium cilium/cilium --version 1.15.5 \
       -f ./dev/values/cilium.yml \
       {{ if enableHubble == "true" { '--set hubble.enabled=true --set hubble.relay.enabled=true --set hubble.ui.enabled=true --set hubble.ui.ingress.enabled=true' } else { '' } }} \
-      --namespace kube-system
+      --namespace kube-system \
+      --labels core=true
```
joroshiba, line 48: adding the label makes it possible to add a command which just deletes all the helm pieces without having to start a new cluster from scratch

charts/sequencer/files/cometbft/config/custom.genesis.json
```diff
@@ -120,12 +120,12 @@
       {{ include "sequencer.address" $value }}
       {{- end }}
     ]
-    {{- if and .Values.global.dev (not .Values.sequencer.upgrades.systemTest.enabled) }},
+    {{- if and .Values.global.dev .Values.genesis.priceFeed }},
```
joroshiba, line 123: It felt awkward the interdependence here. If you define it in the genesis fine, if you define it in an upgrade fine. It might cause issues to have both but.... don't do that?

The files should only defined this way for testing and local development anyways

charts/sequencer/files/cometbft/config/custom.genesis.json
```diff
@@ -168,7 +168,7 @@
       },
       "oracle": {
         "currencyPairGenesis": [
-          {{- range $index, $genesis := .Values.sequencer.priceFeed.currencyPairGenesis }}
+          {{- range $index, $genesis := .Values.genesis.priceFeed.currencyPairGenesis }}
```
joroshiba, line 171: it was unclear why these values were defined on the sequencer when only used in genesis or upgrade

charts/sequencer/files/upgrades/dawn-1.genesis.json
joroshiba, line 1: these files all exist for future definition, similar to how genesis works

charts/sequencer/templates/configmaps.yaml
```diff
@@ -83,9 +83,6 @@ data:
   OTEL_EXPORTER_OTLP_HEADERS: "{{ .Values.sequencer.otel.otlpHeaders }}"
   OTEL_EXPORTER_OTLP_TRACE_HEADERS: "{{ .Values.sequencer.otel.traceHeaders }}"
   OTEL_SERVICE_NAME: "{{ tpl .Values.sequencer.otel.serviceName . }}"
-  {{- if and .Values.sequencer.upgrades.systemTest.enabled .Values.sequencer.upgrades.systemTest.bootstrapping }}
```
joroshiba, line 86: Migrating to latest release eliminated need for this old value we replaced

charts/sequencer/values.yaml
```diff
@@ -139,6 +142,56 @@ genesis:
     #   address: 091E47761C58C474534F4D414AF104A6CAF90C22
     #   pubKey: lV57+rGs2vac7mvkGHP1oBFGHPJM3a+WoAzeFDCJDNU=
 
+upgrades:
```
joroshiba, line 145: define the upgrade as a top level object, like the genesis file.

dev/values/validators/all.yml
```diff
@@ -42,78 +42,26 @@ sequencer:
...
-          price: "3138872234"
-        id: "1"
-    nextId: "2"
 
 resources:
   cometbft:
     requests:
-      cpu: 1000m
-      memory: 500Mi
+      cpu: 300m
```
joroshiba, line 53: I tested these lower values on my machine with no issues.

system-tests/sequencer_upgrade_test.py
```diff
@@ -21,9 +21,9 @@
 
 # The number of sequencer validator nodes to use in the test.
 NUM_NODES = 5
-# A map of upgrade name to sequencer image to use for running BEFORE the given upgrade is executed.
+# A map of upgrade name to sequencer, relayer image to use for running BEFORE the given upgrade is executed.
 PRE_UPGRADE_IMAGE_TAGS = {
-    "upgrade1": "pr-1751"
+    "upgrade1": ("2.0.0", "1.0.1"),
```
joroshiba, line 26: Using release versions here

dev/values/validators/priceFeed.genesis.yml
joroshiba, line 1: this way you can simply set the value`--values=....` file if you want price feed information in genesis. Realistically should we get rid of this ability and instead have an upgrade activated at height 1 as an option? That would be a larger change but I'd be totally cool with that. That would moves us towards "the genesis is done" and stop adding things to it.

PR #2093, opened 2025-04-07
chore(sequencer): remove support for starting networks in post-aspen mode

### LINKED ISSUE #2073: Remove some config options from genesis sequencer chart
Recent work to support sequencer upgrades and a price feed oracle included adding optional sections to `charts/sequencer/files/cometbft/config/custom.genesis.json`.  Specifically, these are `app_state.price_feed` and `consensus_params.abci`.

The status quo is that new networks by default will have these optional sections included in their CometBFT genesis.json and will adopt the corresponding behaviour (i.e. price feed data included via vote extensions) from genesis.  They are also initialized with an empty upgrades.json file from `charts/sequencer/files/upgrades/custom.upgrades.json` by default.

Now that we have upgrade functionality, we should remove the ability to start networks in this way.  Instead, the preference is for those optional sections to be removed from `custom.genesis.json` and an upgrade file provided by default.  This probably means that we can remove the ability to produce an empty upgrades.json by removing the `upgrades.upgrade1.enabled` value. We should also change the default value of `upgrades.upgrade1.baseInfo.activationHeight` from 60 to 1 to ensure the upgrade is applied as soon as possible.

Fraser999:
## Summary
This removes support for starting a network from genesis with functionality normally provided after executing the Aspen upgrade.

## Background
Upgrades were implemented in such a way that if an empty upgrade file was provided to a sequencer starting at genesis, then the sequencer would adopt behaviour similar to if it had executed the Aspen upgrade.

Specifically, this involved the following changed behaviour:
- it would write the consensus params to verifiable storage during `init_chain`
- it would support vote extensions from block 1 onwards, i.e. providing price feed data
- it would use the new form of encoding (serialized `DataItem`s) for the rollup commitments

It was decided to revert this functionality, instead relying on all new networks starting with the same functionality provided by current `main` and immediately executing the Aspen upgrade to transition to the new behaviour.

## Changes
- Reverted support for the functionality specified above.
- Removed price feed section from the genesis app state.
- Deleted `PriceFeedGenesis` message from genesis protobuf types and flattened it into the upgrades `PriceFeedChange` message.
- Removed the `MarketMapComponent` and `OracleComponent` price feed types as these were no longer justified (previously there was logic shared between `init_chain` and executing the Aspen upgrade - the `init_chain` use case has gone).
- Updated several tests (mostly `App` tests) to have them not just run `init_chain` then `commit`, but to also run `finalize_block` then `commit` for empty blocks 1 and 2.  The Aspen upgrade is applied in block 1, and block 2 is executed so that the tests can use block 3, by which point vote extensions are included (they are not included in block 2). The relevant function is `app::test_utils::run_until_aspen_applied`.
- Updated charts and justfiles to reflect these changes.  Specifically, removed the price feed genesis values, removed ability to specify the consensus param `voteExtensionsEnableHeight` in genesis.json, and enabled Aspen upgrade at block 1 for all sequencer deploy recipes.

## Testing
No new tests added.  Existing tests updated as required to maintain their intended purposes.

## Changelogs
No updates required.

## Breaking Changelist
No new breaking changes that weren't already breaking compared to `main`.

## Related Issues
Closes #2073.


crates/astria-sequencer/src/app/mod.rs
```diff
@@ -1619,7 +1591,7 @@ impl App {
         self.upgrades_handler
             .upgrades()
             .aspen()
-            .map_or(true, |aspen| {
+            .map_or(false, |aspen| {
```
noot, line 1594: update comment on this function:
> the assumption is that this network uses `DataItem`s from genesis onwards.
should be that this network does not use `DataItem`s?

noot (approved):

crates/astria-sequencer/src/app/mod.rs
Fraser999: Fixed in [6719cb2](https://github.com/astriaorg/astria/pull/2093/commits/6719cb290cc36b58a891a3afbaa27ba0c7415696).

joroshiba (approved):
Approved for API/infra

SuperFluffy (approved):
Approving with respect to what I am seeing in astria-core. I note that I have no deeper understanding on what's going on in sequencer itself and also haven't looked at it.

PR #2085, opened 2025-04-01
feat(sequencer, core, conductor)!: merge upgrades/price-feed feature branch to main

Fraser999:
## Summary
This merges the long-running feature branch `feat/oracle` to `main`.

## Background
This is in preparation for deploying the new functionality provided in the feature branch, i.e. sequencer upgrades and price-feed data.

## Changes
- This PR is only a merge of the feature branch, comprising many individual PRs to the feature branch along with PRs merging changes from main into the feature branch to keep it in sync.

## Testing
The new features are well covered in new unit and system tests.

## Changelogs
Changelogs already updated in the feature branch.

## Metrics
New metrics:
- `astria_sequencer_extended_commit_info_bytes`
- `astria_sequencer_extend_vote_duration_seconds`
- `astria_sequencer_extend_vote_failure_count`
- `astria_sequencer_verify_vote_extension_failure_count`

## Breaking Changelist
- The addition of price feed data via vote extensions and support for sequencer upgrades are the two major breaking changes.  However, these changes will be applied once the sequencer networks upgrade at given future block heights.  Sequencers running the code in this PR will be compatible with existing networks until the upgrades are activated.

PR #2096, opened 2025-04-08
chore(conductor, core): minor updates

Fraser999:
## Summary
This addresses @SuperFluffy's review comments on #2085.

## Background
Minor updates to address review comments (branch is protected, so cannot be directly addressed via a commit to `feat/oracle`)

## Changes
- Various small updates.

## Testing
New tests added as requested.

## Changelogs
No updates required.

### DIFF
--- a/crates/astria-conductor/src/executor/mod.rs
+++ b/crates/astria-conductor/src/executor/mod.rs
@@ -660,7 +660,7 @@ impl ExecutableBlock {
         } = block;
         let timestamp = convert_tendermint_time_to_protobuf_timestamp(header.time());
         let transactions =
-            insert_price_feed_data_into_transactions_if_exists(transactions, extended_commit_info);
+            prepend_transactions_by_price_feed_if_exists(transactions, extended_commit_info);
         Self {
             hash: block_hash,
             height: header.height(),
@@ -685,7 +685,7 @@ impl ExecutableBlock {
             .unwrap_or_default();
 
         let transactions =
-            insert_price_feed_data_into_transactions_if_exists(transactions, extended_commit_info);
+            prepend_transactions_by_price_feed_if_exists(transactions, extended_commit_info);
         Self {
             hash: block_hash,
             height,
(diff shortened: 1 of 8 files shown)


crates/astria-conductor/src/executor/mod.rs
```diff
@@ -695,32 +695,41 @@ impl ExecutableBlock {
     }
 }
 
+#[instrument(skip_all)]
 fn insert_price_feed_data_into_transactions_if_exists(
```
SuperFluffy, line 699: Let's call this:
```suggestion
fn prepend_transactions_by_price_feed_if_exists
```

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -695,32 +695,41 @@ impl ExecutableBlock {
...
-        std::iter::once(rollup_data.into_raw().encode_to_vec().into())
-            .chain(transactions)
-            .collect()
-    } else {
-        transactions
-    }
+        }
+    };
+
+    // Prepend the price data to the transactions. Regardless of the order here, the rollup can
```
SuperFluffy, line 725: Let's put this entire comment into the doc comment of the containing function (see my suggestion there to rename it to `prepend_if_exists` instead of `insert_if_exists`)

crates/astria-conductor/src/executor/mod.rs
```diff
@@ -695,32 +695,41 @@ impl ExecutableBlock {
...
+    // choose whichever execution order suits best for its use case, but it's anticipated that
+    // applying the updated price feed before executing transactions would be a common use case, so
+    // the price feed is prepended for convenience.
+    let rollup_data = RollupData::PriceFeedData(Box::new(PriceFeedData::new(prices)));
+    std::iter::once(rollup_data.into_raw().encode_to_vec().into())
```
SuperFluffy, line 730: Irrespective of whether we chain or do push + rotate, can you move this to a freestanding function, i.e. smth like `fn prepend<T>(xs: Vec<T>, x: T)`?

crates/astria-sequencer/src/upgrades/upgrades_handler.rs
```diff
@@ -58,7 +58,8 @@ impl UpgradesHandler {
         upgrades_filepath: P,
         cometbft_rpc_addr: String,
     ) -> Result<Self> {
-        let upgrades = Upgrades::new(upgrades_filepath).wrap_err("failed constructing upgrades")?;
+        let upgrades =
+            Upgrades::read_from_path(upgrades_filepath).wrap_err("failed constructing upgrades")?;
```
SuperFluffy, line 62: report the path? i.e. `wrap_err_with(|| format!("failed constructing upgrades from file at `{}`", upgrades_filepath.as_ref().display())`?

although I am also fine if the assumption is that the caller of` UpgradesHandler::new` should report this.

SuperFluffy (approved):
Changes look good. thank you for the unit tests.

I still prefer push + rotate over iter::once + chain + collect. ;)

PR #2101, opened 2025-04-09
fix(sequencer): simple ibc ack failure message

joroshiba:
## Summary
After Aspen upgrade, changes to a simple ibc acknowledgement failure message which does not put sequencer error codes into consensus.

## Background
The exi

## Changes
- Changes Ics 20 transfer ack failure message to a simple string after aspen upgrade.

## Testing
CI/CD

### DIFF
--- a/crates/astria-sequencer/src/ibc/ics20_transfer.rs
+++ b/crates/astria-sequencer/src/ibc/ics20_transfer.rs
@@ -25,6 +25,10 @@ use astria_core::{
         Ics20WithdrawalFromRollup,
     },
     sequencerblock::v1::block::Deposit,
+    upgrades::v1::aspen::{
+        Aspen,
+        IbcAcknowledgementFailureChange,
+    },
 };
 use astria_eyre::{
     anyhow::{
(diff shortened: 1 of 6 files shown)


crates/astria-sequencer/src/ibc/ics20_transfer.rs
```diff
@@ -365,14 +370,25 @@ impl AppHandlerExecute for Ics20Transfer {
...
+        let use_new_error_format = state
+            .get_upgrade_change_info(&Aspen::NAME, &IbcAcknowledgementFailureChange::NAME)
+            .await
+            .map_err(|err| eyre_to_anyhow(err).context("failed to read upgrade info"))?
+            .is_some();
+
         let ack = match receive_tokens(&mut state, &msg.packet).await {
             Ok(()) => TokenTransferAcknowledgement::success(),
             Err(e) => {
                 tracing::debug!(
```
SuperFluffy, line 382: I think this should be upgraded to a `warn!` if the full error is no longer available elsewhere (in case of `use_new_error_format`)

Fraser999: Done in [6f4a27c](https://github.com/astriaorg/astria/pull/2101/commits/6f4a27ca05b8311b371458101147d6fd9b710c7c).

PR #2002, opened 2025-02-26
chore(just): organize just recipes into modules and add comments to all

ethanoroshiba:
## Summary
Leverages justfile's module system to better organize our recipes and adds descriptive and uniform comments to all.

## Background
These changes are meant to make the developer local deployment experience easier to understand and use. One primary goal is for `just --list` to provide a more helpful list of recipes. This example shows the new style of `just deploy`:

```sh
Available recipes:
    # 
    # Deploys cluster, ingress controller, and entire Astria stack, including rollup with
    # Astrotrek and sequencer faucet.
    all
    # 
    # Deploys local Sequencer and Celestia instance.
    # Usage: 
    #   $just deploy astria-local <NAMESPACE> (default: 'astria-dev-cluster')
    astria-local namespace=defaultNamespace
    # 
    # Deploys a local Astrotrek instance.
    astrotrek
    # 
    # Deploys local auctioneer.
    auctioneer
    # 
    # Deploys local EVM Bridge Withdrawer and awaits its readiness.
    bridge-withdrawer
    # 
    # Deploys a local Kubernetes cluster.
    # Usage:
    #   $just deploy cluster <NAMESPACE> <ENABLE_HUBBLE> (defaults: 'astria-dev-cluster', 'false')
    cluster namespace=defaultNamespace enableHubble=hubbleEnabled
    ...
```

The oddity of the extra line in the comment is to enforce that single line comments aren't placed to the right of the recipe, which causes formatting problems.

Another goal of these changes is to make the recipes themselves easier to navigate by splitting into submodules and following clear styling throughout (which aids greatly in identifying recipes if one doesn't have `just` syntax highlighting).

The final aim of these changes is to make the command line outputs easier to understand by hiding a lot of extraneous information, such as outputs and just recipe calls (automatically listed usually).

## Changes
- Added `just` directory to charts, containing modules `clean`, `delete`, `deploy`, `evm`, `init`, and `run`.
- Added `just` directory to dev, containing modules `argo`, `helm`, and `kubetail`.
- Added detailed comments to every recipe, using `[doc]` to support multiline comments.
- Labeled all key sections and individual recipes with consistent formatting, with a minimum of 2 lines between recipes so that it's easy to distinguish without syntax highlighting.
- Suppressed command printing in most commands using `@` and summarized with `echo` statements.
- Moved `wait-for-*` recipe calls into `deploy` recipes so that they are automatically called when a chart is deployed.
- Adjusted all existing just calls in GH workflows accordingly.
[...]

PR #2105, opened 2025-04-10
chore(just): separate deploy and wait-for recipes

### LINKED ISSUE #2074: re-separate `just deploy` and `just wait-for-*` recipes
#2002 combined the `just deploy *` and `just wait-for-*` recipes such that calls to the former would automatically await the given component's readiness. This is a good UX improvement, but can significantly slow down the deployment process if deploying multiple components. In this instance, we should be able to deploy everything at once, and then await all of them.

We should re-separate these recipes to continue with the previous precedent. Creating `deploy-and-wait` recipes would be ideal, but with the nature of `just` imports, there's not a good way to expose `just deploy` recipes while using them in `deploy-and-wait` and keeping all of the listing looking correct.

ethanoroshiba:
## Summary
Separates awaiting of deployment readiness into new `just` module.

## Background
#2002 combined deployment of components and awaiting their readiness for better UX. This, however, can slow down the process of deploying many components greatly, since they one must be completed before moving onto the next. As such, this PR breaks the `deploy` and `wait-for` recipes back into their own parts.

## Changes
- Create new `just` module `wait-for` with all related recipes.
- Use `wait-for` recipes in various `deploy` recipes, such as smoke tests. Note that `wait-for` recipes follow the naming of their deployment counterparts exactly, such that their definitions are overridden when calling `just --list deploy` and are not displayed.

## Testing
Manually tested and passing status checks.

## Changelogs
No updates required.

## Related Issues
closes #2074


charts/just/wait-for.just
```diff
@@ -0,0 +1,80 @@
...
+
+
+# Waits for EVM Bridge Withdrawer
+##################################
+[doc("
+Waits for the EVM Bridge Withdrawer to be ready.
+")]
+bridge-withdrawer:
+  @echo "Waiting for EVM Bridge Withdrawer to be ready..."
+  @kubectl wait -n astria-dev-cluster deployment evm-bridge-withdrawer-local --for=condition=Available=True --timeout=600s > /dev/null
```
warhod, line 20: nit: perhaps you can use the default variable you defined `defaultNamespace` for this and all the commands below where you use `astria-dev-cluster`

charts/just/wait-for.just
```diff
@@ -0,0 +1,80 @@
...
+
+
+# Waits for local Celestia instance
+###################################
+[doc("
+Waits for the local Celestia instance to be ready.
+")]
+celestia-local:
+  @echo "Waiting for local Celestia instance to be ready..."
+  @kubectl rollout status --watch statefulset/celestia-local -n astria-dev-cluster --timeout=600s > /dev/null
```
warhod, line 30: not sure why it's `kubectl rollout status --watch statefulset/celestia-local` here but everywhere else is `kubectl wait ...` perhaps they're interchangeable?

charts/just/wait-for.just
```diff
@@ -0,0 +1,80 @@
...
+
+
+# Waits for Hermes
+###################
+[doc("
+Waits for Hermes to be ready.
+")]
+hermes-local:
+  @echo "Waiting for Hermes to be ready..."
+  @kubectl wait -n astria-dev-cluster deployment hermes-local-chart --for=condition=Available=True --timeout=480s
```
warhod, line 40: the timeout here is 2 minutes less (am sure there's a good reason but wanted to check in case)

charts/just/wait-for.just
```diff
@@ -0,0 +1,80 @@
...
+
+
+# Waits for ingress controller
+###############################
+[doc("
+Waits for the ingress controller to be ready.
+")]
+ingress-controller:
+  @echo "Waiting for ingress controller to be ready..."
+  @while ! kubectl wait --namespace ingress-nginx --for=condition=ready pod --selector=app.kubernetes.io/component=controller --timeout=600s; do \
```
warhod, line 50: tiny nit, you're using `-n` instead of `--namespace` everywhere else

warhod: `while !` can also be `until` for a slightly more readable code

charts/just/deploy.just
```diff
@@ -143,24 +158,21 @@ astrotrek:
...
 
 # Deploy Rollup
 ################
 [doc("
 Deploys a geth rollup chain + faucet + blockscout + ingress with defaults running
-against local network, along with a bridge withdawer. Then, waits for the rollup
-to be ready. Local Sequencer and Celestia networks must be running before deploying
-the rollup.
+against local network, along with a bridge withdawer. Local Sequencer and Celestia
+networks must be running before deploying the rollup.
```
warhod, line 175: Wondering if this can be codified as a task dependency? (or at the very least, document the commands here to run `just deploy xxx`)

ethanoroshiba: Added example of command that can be called in ac3adf6. Let me know if this works for you!

charts/just/wait-for.just
ethanoroshiba: Done in ac3adf6

charts/just/wait-for.just
ethanoroshiba: Unsure of the reason for this, it was just copied over from the previously private recipes in `deploy`. I don't see any reason it shouldn't be the same, though, so I've adjusted to match the rest in ac3adf6.

charts/just/wait-for.just
ethanoroshiba: It seems `rollout status --watch` is better for `statefulset`s? I couldn't find a condition that would work correctly using `kubectl wait` when trying to change it

charts/just/wait-for.just
ethanoroshiba: Good catch! Done in ac3adf6

warhod (approved):
Thanks for addressing the feedback! I tried this out on my local cluster and the wait logic proves to be helpful Occasionally I ran into errors but that's due to existing cluster or other issues with my own environment.  For the purpose of this PR I think everything looks gtg. 👍

PR #2108, opened 2025-04-10
feat(sequencer): add name to validator update action

### LINKED ISSUE #1590: feat(sequencer): updates of validator names

ethanoroshiba:
Note: this PR is a continuation of #2089 with the commit history cleaned up and based on `main`.

## Summary
Adds a name field to `ValidatorUpdate` and includes this change as a part of the `Aspen` upgrade

## Background
Previously, validator names were not stored app-side, and names could only be set in CometBFT genesis.

### `proto`
- Created proprietary `ValidatorUpdate` message with a `name` field added to the tendermint type
- Added `GetValidatorName` rpc to sequencer service, along with request and response types. Reflected in generated files along with sequencer service implementations.

### in `astria-core`
- Use new proprietary `ValidatorUpdate` as raw action instead of tendermint type.
- Added `ValidatorName` which restricts the name to 32 bytes max.
- Added name of `ValidatorName` type to domain `ValidatorUpdate`.

### in `astria-cli`
- Added optional `name` to `sequencer sudo validator_update` action.

### in `astria-sequencer`
- Added storage keys, value impls, snapshots, and state read/write methods for new `ValidatorInfo` and `ValidatorCount` to accommodate for storing names in state.
- Added logic to migrate from the current method of storing the entire `ValidatorSet` in state to storing each validator's information keyed to its address as part of the Aspen upgrade.
- Added new gRPC query for validator name.
- Edited `ValidatorSet` to use `insert` instead of `push_update` so that it doesn't imply any ordering.

## Testing
- Added exhaustive tests for pre- and post-Aspen state changes in `authority/state_ext.rs`.
- Added tests for pre- and post-Aspen `ValidatorUpdate` execution with addition, removal, and change.
- Added tests for pre- and post-Aspen `AuthorityComponent::begin_block()` and `end_block()`.
- Added tests for gRPC query.
- Added validator updates to upgrade testing
- Added validator updates with names to upgrade system tests

## Changelogs
Changelogs updated

## Breaking Changelist
- Noting that this otherwise breaking change is specifically non-breaking since the old `ValidatorUpdate` PB message will result in a blank name and the new PB message will have its name ignored when using with the old action. Additionally, unless the Aspen upgrade has been specified and has passed, Sequencer will default to the old logic.
- Calling out specifically that `sequencerblock` proto changes are non-breaking, since we have only made additions.

## Related Issues
closes #1590


crates/astria-sequencer/src/action_handler/impls/validator_update.rs
```diff
@@ -40,47 +52,101 @@ impl ActionHandler for ValidatorUpdate {
...
             .wrap_err("failed to put validator updates in state")?;
+
         Ok(())
     }
 }
 
+pub(crate) async fn use_pre_aspen_validator_updates<S: StateRead>(state: &S) -> Result<bool> {
+    let pre_aspen_upgrade = state
+        .get_upgrade_change_info(&Aspen::NAME, &ValidatorUpdateActionChange::NAME)
+        .await?
```
noot, line 134: wrap err

crates/astria-sequencer/src/app/vote_extension.rs
```diff
@@ -529,6 +524,33 @@ async fn validate_vote_extensions<S: StateReadExt>(
     Ok(())
 }
 
+async fn verification_key<S: StateReadExt>(
+    state: &S,
+    address: &[u8; ADDRESS_LENGTH],
+) -> Result<VerificationKey> {
+    let verification_key = if use_pre_aspen_validator_updates(state)
```
noot, line 531: is this necessary? vote extensions are only enabled after aspen, so `use_pre_aspen_validator_updates` should always be false definitionally?

crates/astria-sequencer/src/action_handler/impls/validator_update.rs
ethanoroshiba: 3acc14e

crates/astria-sequencer/src/app/vote_extension.rs
ethanoroshiba: Thanks for the catch! Switched to erroring in that scenario in 3acc14e.

quasystaty1 (approved):
Infra approval

crates/astria-core/src/protocol/transaction/v1/action/mod.rs
```diff
@@ -675,117 +689,120 @@ enum ValidatorUpdateErrorKind {
...
-    derive(::serde::Deserialize, ::serde::Serialize),
-    serde(
-        into = "crate::generated::astria_vendored::tendermint::abci::ValidatorUpdate",
-        try_from = "crate::generated::astria_vendored::tendermint::abci::ValidatorUpdate",
-    )
-)]
 pub struct ValidatorUpdate {
     pub power: u32,
     pub verification_key: crate::crypto::VerificationKey,
+    pub name: String,
```
SuperFluffy, line 703: This type apparently has to uphold certain invariants (looking at the `MAX_VALIDATOR_LENGTH` const there). Because of that this field should not be publicly settable as a `String`.

I suggest creating a newtype wrapper `ValidatorUpdateName(String)` if all these fields should be kept public.

ethanoroshiba: Thanks for the catch! Done in 9af1012.

crates/astria-core/src/protocol/transaction/v1/action/mod.rs
```diff
@@ -647,6 +647,66 @@ enum TransferActionErrorKind {
...
+    )]
+    NameTooLong { length: usize },
+}
+
+/// Wrapper for validator name field of [`ValidatorUpdate`]. Cannot be longer than
+/// [`MAX_VALIDATOR_NAME_LENGTH`].
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct ValidatorUpdateName(String);
+
+impl ValidatorUpdateName {
```
SuperFluffy, line 676: IMO add a method `ValidatorUpdateName::into_inner` for those cases where you want to use the inner string and drop the object before (you have lots of `as_str().clone()` everywhere.

crates/astria-core/src/protocol/transaction/v1/action/mod.rs
```diff
@@ -647,6 +647,66 @@ enum TransferActionErrorKind {
...
+        Self(String::new())
+    }
+
+    #[must_use]
+    pub fn as_str(&self) -> &str {
+        &self.0
+    }
+}
+
+impl TryFrom<String> for ValidatorUpdateName {
```
SuperFluffy, line 688: IMO replace this by `FromStr for ValidatorUpdateName`. The `TryFrom` vs `FromStr` differentiation is a bit murky and has historical reasons, but for validating stringy data I'd use `FromStr` (you can then use `<your_string>.parse::<Target>()`, which is a nice API).

crates/astria-core/src/protocol/transaction/v1/action/mod.rs
```diff
@@ -689,18 +749,15 @@ enum ValidatorUpdateErrorKind {
     Secp256k1NotSupported,
     #[error("bytes stored in the .pub_key field could not be read as an ed25519 verification key")]
     VerificationKey { source: crate::crypto::Error },
-    #[error(
-        "validator name was {length} characters long, but must be {MAX_VALIDATOR_NAME_LENGTH} at \
-         most"
-    )]
-    NameTooLong { length: usize },
+    #[error(transparent)]
```
SuperFluffy, line 752: This is validating an actual protobuf, right? This should not be `transparent` then. Actually provide the field that you had to parse and that was bad. I.e.:

```suggestion
    #[error("field `.name` was invalid")]
```

crates/astria-core/src/protocol/transaction/v1/action/mod.rs
```diff
@@ -647,6 +647,66 @@ enum TransferActionErrorKind {
     FeeAsset(#[source] asset::ParseDenomError),
 }
 
+#[derive(Debug, thiserror::Error)]
+#[error(transparent)]
+pub struct ValidatorUpdateNameError(ValidatorUpdateNameErrorKind);
```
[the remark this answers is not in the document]
ethanoroshiba: Done in 336459b.

crates/astria-core/src/protocol/transaction/v1/action/mod.rs
```diff
@@ -647,6 +647,66 @@ enum TransferActionErrorKind {
...
+    #[error(
+        "validator name was {length} characters long, but must be {MAX_VALIDATOR_NAME_LENGTH} at \
+         most"
+    )]
+    NameTooLong { length: usize },
+}
+
+/// Wrapper for validator name field of [`ValidatorUpdate`]. Cannot be longer than
+/// [`MAX_VALIDATOR_NAME_LENGTH`].
+#[derive(Clone, Debug, PartialEq, Eq)]
```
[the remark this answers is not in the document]
ethanoroshiba: 336459b

crates/astria-core/src/protocol/transaction/v1/action/mod.rs
ethanoroshiba: Done in 336459b

crates/astria-core/src/protocol/transaction/v1/action/mod.rs
ethanoroshiba: Good call, done in 336459b.

crates/astria-core/src/protocol/transaction/v1/action/mod.rs
ethanoroshiba: Done in 336459b

joroshiba (approved):

PR #2109, opened 2025-04-10
chore(charts): add initial draft of upgrades.json files for running networks

Fraser999:
## Summary
Initial draft upgrades.json files

## Background
We need an upgrades.json file for each of devnet, testnet and mainnet.

## Changes
- Updated `dusk-11.upgrades.json` with an activation point estimated as 2025-04-14 16:00:00 UTC
- Updated `dawn-1.upgrades.json` with an activation point estimated as 2025-05-14 16:00:00 UTC
- Updated `mainnet.upgrades.json` with an activation point estimated as 2025-05-14 16:00:00 UTC

### DIFF
--- a/charts/sequencer/Chart.yaml
+++ b/charts/sequencer/Chart.yaml
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 2.1.0
+version: 2.1.1
 # This is the version number of the application being deployed. This version number should be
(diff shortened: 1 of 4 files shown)


joroshiba (commented):
Should we leave mainnet & dawn empty for time being? These will use different accounts. 

At a minimum maybe set a hilariously far forward activation height like 999_999_999. Makes it clear that it's not a real number, and makes it such that if used wouldn't accidentally activate?

Fraser999:
mainnet and dawn both have their relevant authorities (all three auths are different), but that will be getting removed once #2113 merges.

I force-pushed a change to set the activation height for mainnet and dawn as you suggested.

PR #2113, opened 2025-04-11
chore(core, sequencer): simplify price feed action

### LINKED ISSUE #2112: Flatten price feed action enum and remove authorities
We want the single new price feed action to be flattened into something more like:
- CurrencyPairsChange
  - addition
  - removal
- MarketsChange
  - creation
  - removal
  - update

We also want to simplify the auth for the action, so we can remove the `Params` type (contains admin and market authorities), and require all price feed actions to have the signer be the sudo address.

Fraser999:
## Summary
The new price feed action has been flattened into two new top-level actions, and authorization for these has been simplified to exclusively require the sudo address.

## Background
The action was structured in a similar way to corresponding objects in Skip's Connect project.  We want to flatten the action into the parent `Action` to keep it more in line with other existing actions.

We also want to simplify authorization for executing these actions.  Again we followed the design in Skip Connect, but this is not required in Astria, at least for now.

## Changes
- Removed `PriceFeed` action and price feed `Params` types.
- Made `CurrencyPairsChange` into a top-level action.
- Renamed `ChangeMarkets` to `MarketsChange` and made this into a top-level action.
- Renamed `MarketsChange` variants to `creation`, `removal` and `update`,
- Changed auth checks for these to require the tx signer be the sudo address.

## Testing
Existing tests were sufficient - these have just been refactored to match the new types.

## Changelogs
No updates required - these are covered by the existing entries around addition of price feed data.

## Breaking Changelist
- Despite changing some API types, these types have not yet been published, so this is effectively not a breaking change.

## Related Issues
Closes #2112.


proto/vendored/connect/marketmap/v2/query.proto
```diff
@@ -26,11 +25,6 @@ service Query {
   rpc LastUpdated(LastUpdatedRequest) returns (LastUpdatedResponse) {
     option (google.api.http) = {get: "/connect/marketmap/v2/last_updated"};
   }
-
-  // Params returns the current x/marketmap module parameters.
-  rpc Params(ParamsRequest) returns (ParamsResponse) {
```
noot, line 31: why was this removed?

noot (approved):
looks good to me!

quasystaty1 (approved):
Infra approval

crates/astria-core/src/oracles/price_feed/market_map.rs
```diff
@@ -1,10 +1,7 @@
 pub mod v2 {
-    use std::{
-        io::{
-            self,
-            Write,
-        },
-        str::FromStr,
+    use std::io::{
+        self,
+        Write,
```
SuperFluffy, line 4: Just a stylistic comment: I'd import `io` and then just refer to it as `io::Write` (I don't see the rest of the file, but I assume that somewhere there is an `impl Write for *`

SuperFluffy (approved):
Approving the changes to core. They seem minimal enough :)

crates/astria-core/src/oracles/price_feed/market_map.rs
Fraser999: Three more LoC deleted - nice! :)

Done in [7d16f72](https://github.com/astriaorg/astria/pull/2113/commits/7d16f72232eacf4103249ea655569663e419397e).

joroshiba (approved):

PR #2115, opened 2025-04-14
feat(account-monitor): initial account monitor service

quasystaty1:
## Summary
Implements account monitor service to continuously track account states on the sequencer network.

## Background
Current monitoring stack lacks the ability to continuously monitor account balances and create alerts based on such metrics. The binary queries a sequencer node using `astria-sequencer-client` for configured accounts information.

## Changes
- implements the new account monitor service

## Testing
Tested locally by running against a mainnet public rpc endpoint. Instructions can be found in the crate README.

## Metrics
- nonce_fetch_count: Counter,
- nonce_fetch_failure_count: Counter,
- balance_fetch_count: Counter,
- balance_fetch_failure_count: Counter,
- account_nonce: HashMap<String, Gauge>,
- account_balance: HashMap<String, Gauge>,
- bridge_last_transaction_height: HashMap<String, Gauge>,

### DIFF
--- a/crates/astria-account-monitor/local.env.example
+++ b/crates/astria-account-monitor/local.env.example
@@ -1,0 +1,62 @@
+# Configuration options of Astria Account Monitor.
+
+# Log level. One of debug, info, warn, or error
+ASTRIA_ACCOUNT_MONITOR_LOG="astria_account_monitor=info"
+
+# If true disables writing to the opentelemetry OTLP endpoint.
+ASTRIA_ACCOUNT_MONITOR_NO_OTEL=false
+
+# If true disables tty detection and forces writing telemetry to stdout.
+# If false span data is written to stdout only if it is connected to a tty.
+ASTRIA_ACCOUNT_MONITOR_FORCE_STDOUT=false
+
+# If set to any non-empty value removes ANSI escape characters from the pretty
+# printed output.
+NO_COLOR=
+
+# Address of the ABCI server for the sequencer chain  
+ASTRIA_ACCOUNT_MONITOR_SEQUENCER_ABCI_ENDPOINT="http://127.0.0.1:26657"
+
+# Chain ID of the sequencer chain which transactions are submitted to.
+ASTRIA_ACCOUNT_MONITOR_SEQUENCER_CHAIN_ID="astria-dev-1"
+
+# The prefix that will be used to construct bech32m sequencer addresses.
+ASTRIA_ACCOUNT_MONITOR_SEQUENCER_ADDRESS_PREFIX=astria
+
+# The sequencer asset for balance monitoring.
+ASTRIA_ACCOUNT_MONITOR_SEQUENCER_ASSET="nria"
+
+# A list of sequencer accounts to monitor the balance of in the format.
+# "address1,address2,..."
+ASTRIA_ACCOUNT_MONITOR_SEQUENCER_ACCOUNTS=""
+
+# A list of sequencer bridge account to keep track of last submission height.
+# expected to follow the format of "address1,address2,..."
+ASTRIA_ACCOUNT_MONITOR_SEQUENCER_BRIDGE_ACCOUNTS=""
+
+# The duration in milliseconds that monitor waits between query requests.
+ASTRIA_ACCOUNT_MONITOR_QUERY_INTERVAL_MS=2000
+
+# Set to true to enable prometheus metrics.
+ASTRIA_ACCOUNT_MONITOR_NO_METRICS=true
+
+# The address at which the prometheus HTTP listener will bind if enabled.
+ASTRIA_ACCOUNT_MONITOR_METRICS_HTTP_LISTENER_ADDR="127.0.0.1:9000"
+
+# The OTEL specific config options follow the OpenTelemetry Protocol Exporter v1
+# specification as defined here:
+# https://github.com/open-telemetry/opentelemetry-specification/blob/e94af89e3d0c01de30127a0f423e912f6cda7bed/specification/protocol/exporter.md
+
+# Sets the general OTLP endpoint.
+OTEL_EXPORTER_OTLP_ENDPOINT="http://localhost:4317"
+# Sets the OTLP endpoint for trace data. This takes precedence over `OTEL_EXPORTER_OTLP_ENDPOINT` if set.
+OTEL_EXPORTER_OTLP_TRACES_ENDPOINT="http://localhost:4317/v1/traces"
+# The duration in seconds that the OTEL exporter will wait for each batch export.
+OTEL_EXPORTER_OTLP_TRACES_TIMEOUT=10
+# The compression format to use for exporting. Only `"gzip"` is supported.
+# Don't set the env var if no compression is required.
+OTEL_EXPORTER_OTLP_TRACES_COMPRESSION="gzip"
+# The HTTP headers that will be set when sending gRPC requests.
+OTEL_EXPORTER_OTLP_HEADERS="key1=value1,key2=value2"
+# The HTTP headers that will be set when sending gRPC requests. This takes precedence over `OTEL_EXPORTER_OTLP_HEADERS` if set.
+OTEL_EXPORTER_OTLP_TRACE_HEADERS="key1=value1,key2=value2"
--- a/crates/astria-account-monitor/src/account_monitor/mod.rs
+++ b/crates/astria-account-monitor/src/account_monitor/mod.rs
@@ -1,0 +1,221 @@
+use std::{
+    str::FromStr,
+    time::Duration,
+};
+
+use astria_core::{
+    primitive::v1::{
+        Address,
+        asset::Denom,
+    },
+    protocol::account::v1::AssetBalance,
+};
+use astria_eyre::eyre::{
+    self,
+    OptionExt,
+    WrapErr as _,
+};
+use sequencer_client::{
+    Client,
+    SequencerClientExt as _,
+    tendermint::{
+        Hash,
+        hash::Algorithm,
+    },
+};
+use tokio::time::interval;
+use tokio_util::sync::CancellationToken;
+use tracing::{
+    debug,
+    info,
+    instrument,
+};
+
+use crate::{
+    config::Config,
+    metrics::Metrics,
+};
+
+pub struct AccountMonitor {
+    shutdown_token: ShutdownHandle,
+    sequencer_abci_client: sequencer_client::HttpClient,
+    accounts: Vec<Address>,
+    bridge_accounts: Vec<Address>,
+    sequencer_asset: Denom,
+    metrics: &'static Metrics,
+    interval: Duration,
+    // fields omitted
+}
+
+impl AccountMonitor {
+    pub fn new(cfg: Config, metrics: &'static Metrics) -> eyre::Result<Self> {
+        let shutdown_handle = ShutdownHandle::new();
+
+        let accounts = cfg.parse_accounts()?;
+        let bridge_accounts = cfg.parse_bridge_accounts()?;
+
+        let Config {
+            sequencer_abci_endpoint,
+            sequencer_asset,
+            ..
+        } = cfg;
+
+        let sequencer_cometbft_client =
+            sequencer_client::HttpClient::new(&*sequencer_abci_endpoint)
+                .wrap_err("failed constructing cometbft http client")?;
+
+        let sequencer_asset = Denom::from_str(sequencer_asset.as_str())
+            .map_err(|e| eyre::eyre!("failed to parse asset: {e}"))?;
+
+        let interval = Duration::from_millis(cfg.block_time_ms);
+        Ok(Self {
+            shutdown_token: shutdown_handle,
+            sequencer_abci_client: sequencer_cometbft_client,
+            accounts,
+            bridge_accounts,
+            sequencer_asset,
+            metrics,
+            interval,
+        })
+    }
+
+    pub async fn run_until_stopped(&self) -> eyre::Result<()> {
+        info!("starting account monitor");
+
+        let mut poll_timer = interval(self.interval);
+
+        loop {
+            // Check if shutdown signal has been received
+            if self.shutdown_token.token.is_cancelled() {
+                info!("received shutdown signal");
+                break;
+            }
+
+            // Wait for the next poll interval
+            poll_timer.tick().await;
+
+            for account in self.bridge_accounts.iter() {
+                let account = account.clone();
+                let last_tx_hash = self
+                    .sequencer_abci_client
+                    .get_bridge_account_last_transaction_hash(account)
+                    .await
+                    .wrap_err("failed to get last transaction hash")?
+                    .tx_hash;
+
+                if let Some(tx_hash) = last_tx_hash {
+                    let last_tx_height = self
+                        .sequencer_abci_client
+                        .tx(Hash::from_bytes(Algorithm::Sha256, &tx_hash)?, false)
+                        .await
+                        .wrap_err("failed query bridge tx by hash")?
+                        .height;
+
+                    self.metrics.set_bridge_last_transaction_height(
+                        account.to_string().as_str(),
+                        last_tx_height.into(),
+                    );
+                }
+            }
+
+            // Process regular accounts
+            for account in self.accounts.iter() {
+                let account = account.clone();
+
+                // Get latest nonce
+                self.metrics.increment_nonce_fetch_count();
+                match get_latest_nonce(&self.sequencer_abci_client, account).await {
+                    Ok(nonce) => {
+                        self.metrics
+                            .set_account_nonce(account.to_string().as_str(), nonce);
+                    }
+                    Err(e) => {
+                        debug!("failed to get latest nonce: {e}");
+                        self.metrics.increment_nonce_fetch_failure_count();
+                    }
+                };
+
+                // Get latest balance
+                self.metrics.increment_balance_fetch_count();
+                match get_latest_balance(
+                    &self.sequencer_abci_client,
+                    account,
+                    self.sequencer_asset.clone(),
+                )
+                .await
+                {
+                    Ok(balance) => {
+                        self.metrics
+                            .set_account_balance(account.to_string().as_str(), balance.balance);
+                    }
+                    Err(e) => {
+                        debug!("failed to get latest balance: {e}");
+                        self.metrics.increment_balance_fetch_failure_count();
+                    }
+                };
+            }
+        }
+
+        Ok(())
+    }
+}
+
+async fn get_latest_balance(
+    client: &sequencer_client::HttpClient,
+    account: Address,
+    asset: Denom,
+) -> eyre::Result<AssetBalance> {
+    let balances = client.get_latest_balance(account).await?;
+    balances
+        .balances
+        .into_iter()
+        .find(|b| b.denom == asset)
+        .ok_or_eyre("failed to find asset balance")
+}
+
+async fn get_latest_nonce(
+    client: &sequencer_client::HttpClient,
+    account: Address,
+) -> eyre::Result<u32> {
+    let nonce = client.get_latest_nonce(account).await?;
+    Ok(nonce.nonce)
+}
+
+/// A handle for instructing the [`Service`] to shut down.
+///
+/// It is returned along with its related `Service` from [`Service::new`].  The
+/// `Service` will begin to shut down as soon as [`ShutdownHandle::shutdown`] is called or
+/// when the `ShutdownHandle` is dropped.
+pub struct ShutdownHandle {
+    token: CancellationToken,
+}
+
+impl ShutdownHandle {
+    #[must_use]
+    fn new() -> Self {
+        Self {
+            token: CancellationToken::new(),
+        }
+    }
+
+    /// Returns a clone of the wrapped cancellation token.
+    #[must_use]
+    pub fn token(&self) -> CancellationToken {
+        self.token.clone()
+    }
+
+    /// Consumes `self` and cancels the wrapped cancellation token.
+    pub fn shutdown(self) {
+        self.token.cancel();
+    }
+}
+
+impl Drop for ShutdownHandle {
+    #[instrument(skip_all)]
+    fn drop(&mut self) {
+        if !self.token.is_cancelled() {
+            info!("shutdown handle dropped, issuing shutdown to all services");
+        }
+        self.token.cancel();
+    }
+}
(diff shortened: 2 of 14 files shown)


crates/astria-account-monitor/src/config.rs
```diff
@@ -0,0 +1,103 @@
...
+    pub block_time_ms: u64,
+
+    /// Forces writing trace data to stdout no matter if connected to a tty or not.
+    pub force_stdout: bool,
+
+    /// Disables writing trace data to an opentelemetry endpoint.
+    pub no_otel: bool,
+
+    /// Set to true to disable the metrics server
+    pub no_metrics: bool,
```
joroshiba, line 41: note: I know we do this everywhere else, but I think it might not make sense to run this binary at all with the metrics server disabled?

crates/astria-account-monitor/src/metrics.rs
```diff
@@ -0,0 +1,174 @@
...
+impl Metrics {
+    pub(crate) fn increment_nonce_fetch_count(&self) {
+        self.nonce_fetch_count.increment(1);
+    }
+
+    pub(crate) fn increment_nonce_fetch_failure_count(&self) {
+        self.nonce_fetch_failure_count.increment(1);
+    }
+
+    pub fn set_account_nonce(&self, account: &str, nonce: u32) {
```
joroshiba, line 33: should the account balance use different labels for different balances? (ie account can have utia, and umilktia balances?)

crates/astria-account-monitor/src/config.rs
```diff
@@ -0,0 +1,103 @@
...
+
+    /// The address prefix to use when constructing sequencer addresses using the signing key.
+    pub sequencer_address_prefix: String,
+
+    /// The addresses of the sequencer chain to monitor.
+    pub sequencer_accounts: String,
+
+    pub sequencer_bridge_accounts: String,
+    /// The asset ID of the sequencer chain to monitor.
+    pub sequencer_asset: String,
```
joroshiba, line 28: Should we just monitor all assets? We have a way to query all and we could report them by using different labels for each?

joroshiba (approved):
Ok on my end as is, a couple of suggestions here but with with them as follow up items. Prefer not to delay things too much.

crates/astria-account-monitor/src/metrics.rs
quasystaty1: what is the value of tracking all assets?

crates/astria-account-monitor/src/config.rs
quasystaty1: What is the value of tracking all assets?

crates/astria-account-monitor/src/account_monitor/mod.rs
```diff
@@ -0,0 +1,234 @@
...
+    }
+
+    /// Run the query loop, polling the sequencer for accounts information.
+    ///
+    /// # Errors
+    /// An error is returned if bridge last transaction height is not found.
+    pub async fn run(&self) -> eyre::Result<()> {
+        info!("starting account monitor");
+
+        let mut poll_timer = interval(self.interval);
```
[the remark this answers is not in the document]
quasystaty1: default behavior is ok here, burst should catch up as quickly as it can.

crates/astria-account-monitor/src/account_monitor/mod.rs
```diff
@@ -0,0 +1,234 @@
...
+                    }
+                };
+            }
+        }
+
+        Ok(())
+    }
+}
+
+async fn get_latest_balance(
```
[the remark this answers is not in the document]
quasystaty1: provided account as a field.

crates/astria-account-monitor/src/main.rs
```diff
@@ -0,0 +1,53 @@
...
+        .try_init(&cfg)
+        .wrap_err("failed to setup telemetry")
+    {
+        Err(_) => {
+            return ExitCode::FAILURE;
+        }
+        Ok(metrics_and_guard) => metrics_and_guard,
+    };
+
+    info!("initializing account monitor");
```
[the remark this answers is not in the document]
quasystaty1: Removed

crates/astria-account-monitor/local.env.example
```diff
@@ -0,0 +1,55 @@
...
+
+# The prefix that will be used to construct bech32m sequencer addresses.
+ASTRIA_ACCOUNT_MONITOR_SEQUENCER_ADDRESS_PREFIX=astria
+
+# The sequencer asset for balance monitoring.
+ASTRIA_ACCOUNT_MONITOR_SEQUENCER_ASSET="nria"
+
+# A list of sequencer accounts to monitor the balance of in the format.
+# "[address1,address2,...]"
+ASTRIA_ACCOUNT_MONITOR_SEQUENCER_ACCOUNTS=[]
```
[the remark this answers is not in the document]
quasystaty1: Had to change Deserializer a bit to support.

crates/astria-account-monitor/src/account_monitor.rs
```diff
@@ -0,0 +1,296 @@
...
+        let Config {
+            sequencer_abci_endpoint,
+            sequencer_asset,
+            sequencer_accounts,
+            ..
+        } = cfg;
+
+        let sequencer_cometbft_client =
+            sequencer_client::HttpClient::new(&*sequencer_abci_endpoint)
+                .wrap_err("failed constructing cometbft http client")?;
```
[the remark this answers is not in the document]
quasystaty1: I think it is being constructed rather than connected. as in `http://127.0.0.1:26657` is valid here even if no sequencer is running. Added the message nonetheless.

crates/astria-account-monitor/src/account_monitor.rs
```diff
@@ -0,0 +1,296 @@
...
+}
+
+async fn run_loop(
+    metrics: &'static Metrics,
+    client: &sequencer_client::HttpClient,
+    accounts: Vec<Account>,
+    denom: Denom,
+    pull_interval: Duration,
+) {
+    let mut poll_timer = interval(pull_interval);
```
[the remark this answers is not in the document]
quasystaty1: Changed to `MissedTickBehavior::Skip`

crates/astria-account-monitor/src/account_monitor.rs
```diff
@@ -0,0 +1,296 @@
...
+    };
+
+    AccountInfo {
+        nonce: account_nonce,
+        balance: account_balance,
+    }
+}
+
+#[derive(Debug, Clone)]
+pub enum QueryResponse<T> {
```
SuperFluffy, line 203: I like this ^.^ But don't make this `pub` (we try to keep the services very locked down); if you need to share it across the crate, `pub(crate)`.
```suggestion
enum QueryResponse<T> {
```

crates/astria-account-monitor/src/account_monitor.rs
```diff
@@ -0,0 +1,296 @@
...
+    Value(T),
+    Error,
+    Timeout,
+}
+
+#[derive(Debug, Clone)]
+pub struct AccountInfo {
```
SuperFluffy, line 210:
```suggestion
struct AccountInfo {
```

crates/astria-account-monitor/src/account_monitor.rs
```diff
@@ -0,0 +1,296 @@
...
+        let Config {
+            sequencer_abci_endpoint,
+            sequencer_asset,
+            sequencer_accounts,
+            ..
+        } = cfg;
+
+        let sequencer_cometbft_client =
+            sequencer_client::HttpClient::new(&*sequencer_abci_endpoint).wrap_err_with(|| {
+                format!("failed to create sequencer client for url {sequencer_abci_endpoint}")
```
SuperFluffy, line 71: I recommend backticks, because it can happen that you provide an empty string, leading to a very strange error:

```suggestion
format!("failed to create sequencer client for url `{sequencer_abci_endpoint}`")
```

crates/astria-account-monitor/src/account_monitor.rs
```diff
@@ -0,0 +1,296 @@
...
+    config::{
+        Asset,
+        Config,
+        SequencerAccountsToMonitor,
+    },
+    metrics::Metrics,
+};
+
+pub struct AccountMonitor {
+    shutdown_token: ShutdownHandle,
```
SuperFluffy, line 40: Don't give the the handle to the monitor - the handle is for the monitor after all and shouldn't be held by it.

(Also `shutdown_token.token` down below reads weird)

crates/astria-account-monitor/src/account_monitor.rs
```diff
@@ -0,0 +1,296 @@
...
+            interval,
+        })
+    }
+
+    /// Run the query loop, polling the sequencer for accounts information.
+    ///
+    /// # Errors
+    /// An error is returned if bridge last transaction height is not found.
+    pub async fn run(&self) -> eyre::Result<()> {
+        let Some(_res) = self
```
SuperFluffy, line 90: The return value is `()` right? Let's make that explicit. We had cases where we accidentally threw away a return value because we bound it to `_value`:
```suggestion
        let Some(()) = self
```

crates/astria-account-monitor/src/config.rs
```diff
@@ -0,0 +1,184 @@
...
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Account {
+    /// The address of the account to monitor.
+    pub address: Address,
+}
+
+impl Hash for Account {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        self.address.to_string().hash(state);
```
SuperFluffy, line 129: In `astria-core`, provide `#[derive(Hash)]` for `Address`. Then you can avoid this.

crates/astria-account-monitor/src/account_monitor.rs
```diff
@@ -0,0 +1,296 @@
...
+    metrics: &'static Metrics,
+    client: &sequencer_client::HttpClient,
+    accounts: &SequencerAccountsToMonitor,
+    asset: &Asset,
+) {
+    for account in accounts {
+        let client = client.clone();
+        let asset = asset.clone();
+        let address = account.address;
+        let _handle =
```
SuperFluffy, line 136: You don't need to `let _handle = ` I think - you should be able to just `tokio::spawn` without binding to a variable (or did clippy flag this with `must_use`?)

crates/astria-account-monitor/src/account_monitor.rs
```diff
@@ -0,0 +1,296 @@
...
+        )
+    );
+
+    let account_nonce = match nonce {
+        Ok(Ok(nonce)) => {
+            metrics.set_account_nonce(&address.into(), nonce);
+            QueryResponse::Value(nonce)
+        }
+        Ok(Err(err)) => {
+            warn!(%err, "failed to get nonce");
```
SuperFluffy, line 168: write our the field as `%error`

crates/astria-account-monitor/src/account_monitor.rs
```diff
@@ -0,0 +1,296 @@
...
+            metrics.increment_nonce_fetch_failure_count();
+            QueryResponse::Error
+        }
+        Err(err) => {
+            warn!(%err, "nonce query timed out");
```
SuperFluffy, line 173: `%error`

crates/astria-account-monitor/src/account_monitor.rs
```diff
@@ -0,0 +1,296 @@
...
+        }
+    };
+
+    let account_balance = match balance {
+        Ok(Ok(balance)) => {
+            metrics.set_account_balance(&address.into(), balance.balance);
+            QueryResponse::Value(balance.balance)
+        }
+        Ok(Err(err)) => {
+            warn!(%err, "failed to get balance");
```
SuperFluffy, line 185: `%error`

crates/astria-account-monitor/src/account_monitor.rs
```diff
@@ -0,0 +1,296 @@
...
+            metrics.increment_balance_fetch_failure_count();
+            QueryResponse::Error
+        }
+        Err(err) => {
+            warn!(%err, "balance query timed out");
```
SuperFluffy, line 190: `%error`

crates/astria-account-monitor/src/config.rs
```diff
@@ -0,0 +1,184 @@
...
+    }
+}
+
+impl FromStr for Account {
+    type Err = eyre::Report;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let address = s
+            .parse()
+            .map_err(|e| eyre::eyre!("failed to parse account address: {e}"))?;
```
SuperFluffy, line 147: I think I mentioned this before in an (invalidated?) comment: prefer `wrap_err` over explicitly formatting the error. Although in this case I think it'd be fine to just forward the `<Address as FromStr>::Err` without wrapping this in eyre.

crates/astria-account-monitor/src/config.rs
```diff
@@ -0,0 +1,184 @@
...
+        Ok(Self {
+            address,
+        })
+    }
+}
+
+impl Display for Account {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.address)
```
SuperFluffy, line 156:
```suggestion
        self.address.fmt(f)
```

crates/astria-account-monitor/src/config.rs
```diff
@@ -0,0 +1,184 @@
...
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let value = String::deserialize(deserializer)?;
+
+        let accounts: Result<Vec<Account>, _> = value
+            .split(',')
+            .map(str::trim)
+            .filter(|s| !s.is_empty())
```
SuperFluffy, line 92: I wouldn't filter these. If somebody provides `astriaxxxx,,astriaxxxxx` then that empty field is just a wrong address.

crates/astria-account-monitor/src/config.rs
```diff
@@ -0,0 +1,184 @@
...
+            .map(str::parse)
+            .collect();
```
SuperFluffy, line 94: Very mild nit: I prefer to have the collection that you collect into as part of the function signature so that you don't have to back up (if it's a more complicated thing):

```suggestion
.collect::<Result<Vec<Account>, _>();
```

SuperFluffy (approved):
Looks great, thank you for the work. I left a few nits that I would like to see addressed, and a few suggestions.

crates/astria-account-monitor/src/account_monitor.rs
```diff
@@ -0,0 +1,296 @@
...
+    client: &sequencer_client::HttpClient,
+    accounts: &SequencerAccountsToMonitor,
+    asset: &Asset,
+) {
+    for account in accounts {
+        let client = client.clone();
+        let asset = asset.clone();
+        let address = account.address;
+        let _handle =
+            tokio::spawn(async move { fetch_account_info(metrics, &client, address, asset).await });
```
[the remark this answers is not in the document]
quasystaty1: the client does not live long enough without it

crates/astria-account-monitor/src/account_monitor.rs
quasystaty1: Should the handle be created at the start of the run loop or should we move the block to `main.rs`

crates/astria-account-monitor/src/config.rs
quasystaty1: Had to provide `#[derive(Hash)]` to `Bech32m` enum in `astria-core-address`.

crates/astria-core-address/src/lib.rs
```diff
@@ -302,9 +302,9 @@ enum ErrorKind {
     },
 }
 
-#[derive(Clone, Copy, Debug)]
+#[derive(Clone, Copy, Debug, Hash)]
```
SuperFluffy, line 305: Hah, so I did derive `Hash for Address<T>`, but I didn't derive it for `Bech32` (and `Bech32m`), so you could in fact not use it for `Address` at all.

crates/astria-account-monitor/src/main.rs
```diff
@@ -27,8 +27,8 @@ async fn main() -> ExitCode {
         .try_init(&cfg)
         .wrap_err("failed to setup telemetry")
     {
-        Err(e) => {
-            eprintln!("failed to setup telemetry: {e}");
+        Err(error) => {
+            eprintln!("failed to setup telemetry: {error}");
```
SuperFluffy, line 31:
```suggestion
eprintln!("failed to setup telemetry:\n{error}");
```

crates/astria-account-monitor/src/main.rs
```diff
@@ -38,10 +38,11 @@ async fn main() -> ExitCode {
         Err(_) => return ExitCode::FAILURE,
         Ok(account_monitor) => account_monitor,
     };
+
     match account_monitor.run().await {
         Ok(()) => ExitCode::SUCCESS,
-        Err(e) => {
-            eprintln!("account monitor exited unexpectedly: {e}");
+        Err(error) => {
+            eprintln!("account monitor exited unexpectedly: {error}");
```
SuperFluffy, line 45:
```suggestion
eprintln!("account monitor exited unexpectedly:\n{error}");
```

crates/astria-account-monitor/src/config.rs
```diff
@@ -142,9 +129,7 @@ impl FromStr for Account {
     type Err = eyre::Report;
```
SuperFluffy, line 129: Re my comment from the previous iteration (I think I wasn't clear on this one): instead of `type Err = eyre::Report`, just use the error that `<Address as FromStr>::Err` is using. You can literally write this:

```suggestion
    type Err = <Address as FromStr>::Err;
```

crates/astria-account-monitor/src/account_monitor.rs
```diff
@@ -82,40 +98,70 @@ impl AccountMonitor {
...
-async fn run_loop(
-    metrics: &'static Metrics,
-    client: &sequencer_client::HttpClient,
-    accounts: &SequencerAccountsToMonitor,
-    asset: &Asset,
-    pull_interval: Duration,
-) {
-    let mut poll_timer = interval(pull_interval);
-    poll_timer.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
+impl AccountMonitor {
```
SuperFluffy, line 125: Group the impl blocks appropriately: `impl AccountMonitor` should be placed after its definition; all of its trait impls after the inherent impls.

Same for `Inner`

crates/astria-account-monitor/src/account_monitor.rs
```diff
@@ -82,40 +98,70 @@ impl AccountMonitor {
...
-        let Some(()) = shutdown_token
-            .token
-            .run_until_cancelled(run_loop(
+    async fn run(self) -> eyre::Result<()> {
+        let mut poll_timer = interval(self.interval);
+        poll_timer.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
+
+        loop {
+            if self.shutdown_token.is_cancelled() {
+                info!("received shutdown signal");
```
SuperFluffy, line 107: This event happens outside of an span. Either construct an adhoc span within which you emit this event - or just exit without an event.

crates/astria-account-monitor/src/account_monitor.rs
```diff
@@ -82,40 +98,70 @@ impl AccountMonitor {
...
-        let shutdown_token = ShutdownHandle::new();
-        let Some(()) = shutdown_token
-            .token
-            .run_until_cancelled(run_loop(
+    async fn run(self) -> eyre::Result<()> {
+        let mut poll_timer = interval(self.interval);
+        poll_timer.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
+
+        loop {
+            if self.shutdown_token.is_cancelled() {
```
SuperFluffy, line 106: You have brought back the behaviour we already fixed in the meantime. Use `CancellationToken::run_until_cancelled` or use `tokio::select!` with a `CancellationToken::cancelled` arm.

SuperFluffy (changes requested):
The recent changes have reintroduced behavior that was already reviewed and fixed.

crates/astria-account-monitor/src/account_monitor.rs
quasystaty1: removed

crates/astria-account-monitor/src/main.rs
```diff
@@ -0,0 +1,67 @@
...
+    tokio::select! {
+        _ = sigterm.recv() => {
+            account_monitor.shutdown();
+            ExitCode::SUCCESS
+        },
+        res = &mut account_monitor => {
+            match res {
+                Ok(()) => ExitCode::SUCCESS,
+                Err(error) => {
+                    eprintln!("account monitor exited unexpectedly:\n {error}");
```
SuperFluffy, line 61: Remove space and change the message:
```suggestion
                    eprintln!("account monitor exited with error:\n{error}");
```

crates/astria-account-monitor/src/main.rs
```diff
@@ -0,0 +1,67 @@
...
+        }
+        Ok(account_monitor) => account_monitor,
+    };
+
+    let mut sigterm = signal(SignalKind::terminate())
+        .expect("setting a SIGTERM listener should always work on Unix");
+
+    tokio::select! {
+        _ = sigterm.recv() => {
+            account_monitor.shutdown();
```
SuperFluffy, line 54: Right now, if we receive a shutdown signal this cancels the service, but then immediately exits. This means that if we ever have some cleanup logic that needs to run, it will not run.

Suggestion for an easy fix: turn `AccountMonitor::shutdown` into a `async fn` and then let it await itself, i.e. something like:

```rust
impl AccountMonitor {
    pub async fn shutdown(self) -> eyre::Result<()> {
        self.shutdown_token.cancel();
        self.await // using the fact that it's a future
    }
}
```

You can then have both `select!` arms returns the same information (so you don't duplicate code).

crates/astria-account-monitor/src/main.rs
```diff
@@ -0,0 +1,67 @@
...
+            ExitCode::SUCCESS
+        },
+        res = &mut account_monitor => {
+            match res {
+                Ok(()) => ExitCode::SUCCESS,
```
SuperFluffy, line 59: This should not be a success case because the service should only exit after a shtudown signal - this arm is actually `eprintln!("account monitor exited unexpectedly");`

SuperFluffy (approved):
Neat, good job!

Very minor nits. Please address them, but otherwise feel free to merge.

PR #2116, opened 2025-04-14
feat(charts): account monitor chart

quasystaty1:
## Summary
Adds helm chart to run the account monitor service in kubernetes environment.

## Background
Required to run in production and as part of the local stack

## Changes
- added helm chart for the account monitor service

## Testing
locally against a cluster

## Changelogs
No updates required.


charts/account-monitor/templates/deployment.yaml
```diff
@@ -0,0 +1,34 @@
...
+          stdin: {{ .Values.global.useTTY }}
+          tty: {{ .Values.global.useTTY }}
+          envFrom:
+            - configMapRef:
+                name: account-monitor-env
+          ports:
+            {{- if .Values.metrics.enabled }}
+            - containerPort: {{ .Values.ports.metrics }}
+              name: monitor-metric
+            {{- end }}
```
joroshiba, line 32: I know it's standard across other charts and binaries but perhaps this binary doesn't make much sense with metrics disabled? As that is kinda the point?

joroshiba (approved):

Issue #1958, opened 2025-02-10
Make CheckTx checks stricter


Fraser999:
We currently are too permissive in our CheckTx flow, allowing txs into the CometBFT and appside mempools which will subsequently fail execution via stateful checks in the `check_and_execute` flow.

This is problematic as such txs are not recorded in the sequencer block nor in the CometBFT txs.  This means they remain present in the mempools of all but the validator proposing the block for that round.

A design to improve this situation has been created [here](https://www.notion.so/astria-org/CheckTx-and-Tx-Execution-Rework-1986bd31a90c80608443c4ac4a2542ca).

This ticket will implement the first part of the suggested changes, i.e. stricter checks during CheckTx.

joroshiba:
This issue is stale because it has been open 45 days with no activity. Remove stale label or this issue
be closed in 7 days.

joroshiba:
This issue was closed because it was stale

PR #2132, opened 2025-04-24
chore(sequencer): remove extraneous check from action

Fraser999:
## Summary
Removed an extraneous check from the `CurrencyPairsChange` action.

## Background
The check fails if the action specifies more currency-pair removals than currently exist in storage.  However, given that removal of a non-existent pair is silently ignored, there is no point to this check.

## Changes
- Removed the check.

## Testing
No tests cover this currently.  (More extensive tests are being added as part of #1958.)

## Changelogs
No updates required.

## Breaking Changelist
- While this is technically a consensus-breaking change, the action in question has only been deployed to devnet as part of v3.0.0-rc.1, so it is relatively safe to include this without any special handling.


noot (approved):

Fraser999:
Closing in favour of approach specified in #2158.

Issue #2130, opened 2025-04-23
Remove correct tx ID when promotion fails in mempool


Fraser999:
This [line](https://github.com/astriaorg/astria/blob/cb262a84b0aeda2ced8489817b178175b32e04e2/crates/astria-sequencer/src/mempool/mod.rs#L284) has specified the wrong transaction ID to remove from the mempool's `contained_txs`.

`timemarked_tx` has just been added to the pending queue at this stage and has not yet been added to  `contained_txs`.

We should instead be using `tx_id` as this tx has just been removed from parked with the intent of adding it to pending, but adding to pending failed.

`tx_id` should also be added to the `comet_bft_removal_cache`, but given that this flow should never actually be hit (all txs picked from the parked should be chosen so that they _can_ be successfully added to pending) we'll need to introduce a new `RemovalReason::InternalError` or similar for this case.

PR #2135, opened 2025-04-24
fix(sequencer): remove failed promotable instead of inserted tx

ethanoroshiba:
## Summary
Fixes an error in mempool promotion logic after insertion.

## Background
Previously, any transactions which failed to be added to the `Pending` queue during promotion post-insertion would remove the just-inserted transaction ID from `contained_txs`. This is an error, but it hasn't caused any problems since this code path is meant to never happen. Additionally, any transactions which failed adding to pending should be added to the CometBFT removal cache.

## Changes
- In `Mempool::insert`, post-insertion, in failure case of adding promoted transaction to `pending`: remove `tx_id` from `contained_txs` instead of the transaction which was successfully added.
- Add failing transaction to CometBFT removal cache.
- Added `RemovalReason::InternalError`.

## Testing
Added unit test to ensure this functionality works as expected. The test scenario is quite contrived, but without direct manipulation of the `pending` and `parked` caches, we have no other way of creating this edge case.

## Changelogs
Changelog updated.

## Related Issues
closes  #2130

### DIFF
--- a/crates/astria-sequencer/CHANGELOG.md
+++ b/crates/astria-sequencer/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Fixed mempool post-insertion failure case during promotion [#2135](https://github.com/astriaorg/astria/pull/2135).
+
 ## [3.0.0-rc.1]
 
 ### Added
--- a/crates/astria-sequencer/CHANGELOG.md
+++ b/crates/astria-sequencer/CHANGELOG.md
@@ -11,7 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Fixed mempool post-insertion failure case during promotion [#2135](https://github.com/astriaorg/astria/pull/2135).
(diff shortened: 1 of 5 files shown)


Fraser999 (approved):

crates/astria-sequencer/CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Fixed mempool post-insertion failure case during promotion [#2135](https://github.com/astriaorg/astria/pull/2135).
```
SuperFluffy, line 14: You can also amend the changelog to have the (slightly) clearer PR title.

crates/astria-sequencer/src/app/test_utils.rs
```diff
@@ -119,6 +119,15 @@ impl MockTxBuilder {
         }
     }
 
+    pub(crate) fn new_with_chain_id(chain_id: String) -> Self {
```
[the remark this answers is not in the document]
ethanoroshiba: Removed in 20c6409

crates/astria-sequencer/CHANGELOG.md
ethanoroshiba: Done in 20c6409.

noot (approved):

PR #2139, opened 2025-04-28
feat(cli): generate signing key from mnemonic

quasystaty1:
## Summary
adds mnemonic field to `astria-cli sequencer account create` output.
## Background
Signing keys are generated from OsRng, which is not ideal for wallet backup. This PR switches to mnemonic usage, emitting random mnemonic used to generate signingKey, it can then be used to reconstruct the accout.

## Changes
- signing key in account create subcommand now generated from random mnemonic.
- added `--from-mnemonic` flag to generate an account from a specific mnemonic.

## Testing
Tested locally by running commands
* `astria-cli sequencer account create`
* `astria-cli sequencer account create --from-mnemonic`
two blackbox tests:
- should_reconstruct_account
- should_create_account

## Changelogs
Changelogs updated.

## Related Issues

closes


crates/astria-cli/tests/create_account.rs
```diff
@@ -0,0 +1,75 @@
...
+
+#[test]
+fn should_reconstruct_account() -> Result<(), Box<dyn std::error::Error>> {
+    let resources: Resources = Resources::new("create_account")?;
+    resources.check_reconstruct_account()?;
+    Ok(())
+}
+
+#[test]
+#[track_caller]
```
Fraser999, line 64: This is only worth adding to functions called from within the test, not the test itself.  I'd move this to `fn new_create_account_command()`.

crates/astria-cli/src/sequencer/account.rs
```diff
@@ -44,11 +44,44 @@ struct Create {
     /// The address prefix
     #[arg(long, default_value = "astria")]
     prefix: String,
+    /// Optional mnemonic to use for key generation.
+    #[arg(long = "from-mnemonic", default_value = None)]
+    mnemonic: Option<String>,
```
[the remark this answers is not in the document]
quasystaty1: A separate subcommand sounds nice.

crates/astria-cli/src/sequencer/account.rs
```diff
@@ -44,11 +44,44 @@ struct Create {
...
+    /// Mnemonic length.
+    #[arg(long, default_value = "24")]
```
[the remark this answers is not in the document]
quasystaty1: Got it! Also limited values to "12" or "24" which are the standard in the space

crates/astria-cli/src/sequencer/account.rs
```diff
@@ -25,30 +31,94 @@ impl Command {
...
+            .prefix(&self.prefix)
+            .try_build()?;
+
+        println!("Recover Sequencer Account");
+        println!();
+        // TODO: don't print private keys to CLI, prefer writing to file:
+        // https://github.com/astriaorg/astria/issues/594
+        println!("Private Key: {pretty_signing_key}");
+        println!("Public Key:  {pretty_verifying_key}");
+        println!("Address:     {pretty_address}");
```
SuperFluffy, line 86: Please replace this by a single print statement.

crates/astria-cli/src/sequencer/account.rs
```diff
@@ -25,30 +31,94 @@ impl Command {
...
+
+        let signing_key = SigningKey::from(seed_bytes);
+
+        let pretty_signing_key = hex::encode(signing_key.as_bytes());
+        let pretty_verifying_key = hex::encode(signing_key.verification_key().as_bytes());
+
+        let pretty_address: Address = Address::builder()
+            .array(signing_key.address_bytes())
+            .prefix(&self.prefix)
+            .try_build()?;
```
SuperFluffy, line 78: context

crates/astria-cli/src/sequencer/account.rs
```diff
@@ -25,30 +31,94 @@ impl Command {
...
+
 #[derive(Debug, clap::Args)]
 struct Create {
     /// The address prefix
     #[arg(long, default_value = "astria")]
     prefix: String,
+    /// Number of words to use in the mnemonic phrase
+    #[arg(
+        long,
+        short = 'l',
```
SuperFluffy, line 99: I know this clashes with @Fraser999's request above, but I'd rather stay away from short values for rarely used subcommands and keep the explicit `mnemonic-length` for now. We can always add short commands later - but once we have them we need to support them/

crates/astria-cli/src/sequencer/account.rs
```diff
@@ -25,30 +31,94 @@ impl Command {
...
+        default_value = "24",
+        value_parser(mnemonic_length_value_parser())
+    )]
+    mnemonic_length: u8,
```
SuperFluffy, line 103: Having said the above: is it even desirable to support both 12 and 24? Why not just choose 12 for now?

crates/astria-cli/src/sequencer/account.rs
```diff
@@ -61,6 +131,7 @@ impl Create {
         println!();
         // TODO: don't print private keys to CLI, prefer writing to file:
         // https://github.com/astriaorg/astria/issues/594
+        println!("Mnemonic:    {bip39_mnemonic}");
         println!("Private Key: {pretty_signing_key}");
         println!("Public Key:  {pretty_verifying_key}");
         println!("Address:     {pretty_address}");
```
SuperFluffy, line 137: Would be nice to use a single print statement for this, as above.

crates/astria-cli/src/sequencer/account.rs
```diff
@@ -126,3 +197,7 @@ struct ArgsInner {
     /// The address of the Sequencer account
     address: Address,
 }
+
+fn mnemonic_length_value_parser() -> impl TypedValueParser {
+    PossibleValuesParser::new(["12", "24"]).map(|v| v.parse::<u8>().unwrap())
```
SuperFluffy, line 202: No panicking, especially without a message. Have a look at this: https://docs.rs/clap/latest/clap/struct.Command.html#method.error

crates/astria-cli/src/sequencer/account.rs
```diff
@@ -6,12 +6,18 @@ use astria_sequencer_client::{
...
+use clap::{
+    builder::{
+        PossibleValuesParser,
+        TypedValueParser,
+    },
+    Subcommand,
+};
 use color_eyre::eyre::{
     self,
+    Ok,
```
SuperFluffy, line 18: What's this?

crates/astria-cli/src/sequencer/account.rs
```diff
@@ -25,30 +31,94 @@ impl Command {
...
+    /// The recovery mnemonic phrase, must be wrapped in quotes.
+    #[arg(long, short)]
+    mnemonic: String,
+}
+
+impl Recover {
+    fn run(self) -> eyre::Result<()> {
+        let bip39_mnemonic = bip39::Mnemonic::from_phrase(&self.mnemonic, bip39::Language::English)
+            .wrap_err("failed to recover mnemonic from phrase")?;
+        let seed = bip39::Seed::new(&bip39_mnemonic, "");
```
SuperFluffy, line 65: I wonder if providing a password here would actually be a nice feature...

crates/astria-cli/src/sequencer/account.rs
```diff
@@ -25,30 +31,94 @@ impl Command {
...
+        let pretty_address: Address = Address::builder()
+            .array(signing_key.address_bytes())
+            .prefix(&self.prefix)
+            .try_build()?;
+
+        println!("Recover Sequencer Account");
+        println!();
+        // TODO: don't print private keys to CLI, prefer writing to file:
+        // https://github.com/astriaorg/astria/issues/594
+        println!("Private Key: {pretty_signing_key}");
```
SuperFluffy, line 84: In regards to the password field that bip39 seeds in principle provide: we can completely avoid ever printing the private key (and just print the public key + the mnemonic) if we set that userprovided password that `tiny_bip39::Seed::new` takes.

crates/astria-cli/Cargo.toml
```diff
@@ -42,6 +43,11 @@ jiff.workspace = true
 tryhard.workspace = true
 serde_json.workspace = true
 futures.workspace = true
+tiny-bip39 = "2.0.0"
```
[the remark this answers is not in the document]
quasystaty1: Nothing specific, both appears to be widely used and I preferred `tiny-bip39`  interface. Ok with changing over.

crates/astria-cli/src/sequencer/account.rs
quasystaty1: Yeah that's a fair point.

crates/astria-cli/src/sequencer/account.rs
quasystaty1: 12 and 24 are more common, I'm ok with allowing only 12 as well.

crates/astria-cli/src/sequencer/account.rs
quasystaty1: Less common in space, however we can make it optional.

crates/astria-cli/src/sequencer/account.rs
quasystaty1: We don't support a local keyring at the moment, so for signing a transaction a user is required to provide a privatekey. So not printing the private key at the moment does not make sense.

crates/astria-cli/Cargo.toml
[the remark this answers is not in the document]
SuperFluffy: No need to change over. I was merely curious about using one over the other.

crates/astria-cli/src/sequencer/account.rs
```diff
@@ -48,22 +91,33 @@ struct Create {
...
+            .wrap_err("failed constructing a valid from address from the provided prefix")?;
 
-        println!("Create Sequencer Account");
-        println!();
         // TODO: don't print private keys to CLI, prefer writing to file:
         // https://github.com/astriaorg/astria/issues/594
-        println!("Private Key: {pretty_signing_key}");
-        println!("Public Key:  {pretty_verifying_key}");
-        println!("Address:     {pretty_address}");
+        println!(
```
SuperFluffy, line 115: IMO don't indent with spaces here. Just leave it with a single space separator. We need to find a more "sustainable" way to arrange the columns of the output (even beyond tabs, for example).

crates/astria-cli/src/sequencer/account.rs
```diff
@@ -25,20 +24,64 @@ impl Command {
...
+        let pretty_address: Address = Address::builder()
+            .array(signing_key.address_bytes())
+            .prefix(&self.prefix)
+            .try_build()
+            .wrap_err("failed constructing a valid from address from the provided prefix")?;
+
+        // TODO: don't print private keys to CLI, prefer writing to file:
+        // https://github.com/astriaorg/astria/issues/594
+        println!(
+            "Recover Sequencer Account\n\nPrivate Key: {pretty_signing_key}\nPublic Key:  \
```
SuperFluffy, line 77: As below, I wouldn't do the manual spacing here. Let's fix this in a followup by providing a mechanism to fix this across all outputs (e.g. so that adding a character or changing the variable output doesn't require manually re-indenting things).

crates/astria-cli/tests/create_and_recover_account.rs
```diff
@@ -0,0 +1,77 @@
+use std::{
+    fs,
+    path::Path,
+};
+
+use assert_cmd::Command;
+use astria_eyre::{
+    eyre::WrapErr,
```
SuperFluffy, line 8: `as _`, but very nit

crates/astria-cli/CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add mnemonic to `account create` subcommand [#2139](https://github.com/astriaorg/astria/pull/2139).
```
SuperFluffy, line 14: You also added a `recover` subcommand, if I see it correctly.

SuperFluffy (approved):
Good job

joroshiba:
closed in 7 days.

joroshiba:

PR #2140, opened 2025-04-28
fix(conductor): add explicit TLS config

mycodecrafting:
## Summary
Add explicit TLS config for conductor when the sequencer url is https

## Background
Tonic [removed implicit TLS configs](https://github.com/hyperium/tonic/pull/1731) in v0.12.0 which now causes TLS errors when conductor tries to connect to a remote sequencer network over TLS.

### DIFF
--- a/crates/astria-conductor/src/sequencer/client.rs
+++ b/crates/astria-conductor/src/sequencer/client.rs
@@ -39,10 +39,9 @@ impl SequencerGrpcClient {
         let uri: Uri = sequencer_uri
             .parse()
             .wrap_err("failed parsing provided string as Uri")?;
-        let mut endpoint = Endpoint::from(uri.clone());
-        if uri.scheme() == Some(&http::uri::Scheme::HTTPS) {
-            endpoint = endpoint.tls_config(ClientTlsConfig::new().with_enabled_roots())?;
-        }
+        let endpoint = Endpoint::from(uri.clone())
+            .tls_config(ClientTlsConfig::new().with_enabled_roots())
+            .wrap_err("failed to configure TLS for sequencer client")?;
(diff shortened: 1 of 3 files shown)


joroshiba (approved):

joroshiba:
Would be good to get a changelog entry w/ info on the fix.

crates/astria-conductor/src/sequencer/client.rs
```diff
@@ -38,7 +39,10 @@ impl SequencerGrpcClient {
         let uri: Uri = sequencer_uri
             .parse()
             .wrap_err("failed parsing provided string as Uri")?;
-        let endpoint = Endpoint::from(uri.clone());
+        let mut endpoint = Endpoint::from(uri.clone());
+        if uri.scheme() == Some(&http::uri::Scheme::HTTPS) {
```
SuperFluffy, line 43: Is this necessary? Would setting the tls config preclude us from using http? I think we should be able to do without?

Either way, if this does blockhttp, I think I'd prefer to have ane explicit flag that deactivates tls so that one doesn't accidentally provide http instead of https and risk exposing the service to an unchecked endpoint.

crates/astria-conductor/src/sequencer/client.rs
```diff
@@ -38,7 +39,10 @@ impl SequencerGrpcClient {
...
+            endpoint = endpoint.tls_config(ClientTlsConfig::new().with_enabled_roots())?;
```
SuperFluffy, line 44: This is missing error context. Report that you failed to set the tls config using `wrap_err`

SuperFluffy (approved):
Looks good. But seconding @joroshiba: let's add an entry to the changelog.

PR #2159, opened 2025-05-07
feat(sequencer): add tx result to "executed" transaction status response

### LINKED ISSUE #2153: Add tx results to transaction status proto
#2133 introduced the mempool API containing the `TransactionStatus` resource. One of its variants is `TransactionStatus::Executed`, which contains the height at which the transaction was included. It would be good to also include the `TxResults` with this response.

ethanoroshiba:
## Summary
Adds the `ExecTxResult` to `TransactionStatus::Executed` for the given transaction.

## Background
Previously, the `TransactionStatus::Executed` only contained the height at which the transaction was executed. It would be good UX to included the results of the transaction as well, which this change aims to address.

## Changes
- Adds `ExecTxResult`, `Event`, and `EventAttribute` to vendored protos, and added `result` field to `TransactionStatus::Executed`.
- Implements `Protobuf` for the Tendermint Rs counterparts to the new generated types.
- Rolls transaction IDs into the `tx_results` collections throughout `sequencer::App`, and includes the collection in `WriteBatch` to be passed into `run_maintenance`.

## Testing
Added the result to existing test for `TransactionStatus::Executed`.

## Changelogs
Changelog updated.

## Related Issues
closes #2153


crates/astria-sequencer/src/mempool/mod.rs
```diff
@@ -69,8 +70,14 @@ impl std::fmt::Display for RemovalReason {
...
             }
-            RemovalReason::IncludedInBlock(block_height) => {
-                write!(f, "included in sequencer block {block_height}")
+            RemovalReason::IncludedInBlock {
+                height,
+                result,
+            } => {
+                write!(
+                    f,
+                    "included in sequencer block {height} with result {result:?}"
```
Fraser999, line 79: I don't think we should use the debug form here.  Potentially JSON-encoding the results would be the most useful, but failing that, I'd suggest just doing a more human-readable form here, possibly skipping fields we never populate (e.g. `data`, `gas_wanted`, etc.)

proto/mempoolapis/astria/mempool/v1/transaction_status.proto
```diff
@@ -41,5 +41,33 @@ message TransactionStatus {
...
+    message ExecTxResult {
+      uint32 code = 1;
+      bytes data = 2;
+      string log = 3;
+      string info = 4;
+      int64 gas_wanted = 5;
+      int64 gas_used = 6;
+      repeated Event events = 7;
+      string codespace = 8;
+    }
```
Fraser999, line 71: I'm not sure if these would be better added to an existing vendored protobuf file, since they're lifted from `tendermint/abci/types.proto`?

crates/astria-sequencer/src/app/mod.rs
```diff
@@ -223,8 +223,9 @@ struct WriteBatch {
     /// of the current block being executed, created from the state delta,
     /// and set after `finalize_block`.
     write_batch: StagedWriteBatch,
-    /// The IDs of all transactions executed in the block for which this write batch is made.
-    executed_tx_ids: HashSet<TransactionId>,
+    /// The hashes and results of all transactions executed in the block for which this write batch
+    /// is made.
+    executed_txs: HashMap<TransactionId, ExecTxResult>,
```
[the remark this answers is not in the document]
ethanoroshiba: Done in 9d386ee.

crates/astria-sequencer/src/app/mod.rs
```diff
@@ -1328,12 +1326,15 @@ impl App {
...
                             AbciErrorCode::INTERNAL_ERROR
                         };
-                        tx_results.push(ExecTxResult {
-                            code: Code::Err(code.value()),
-                            info: code.info(),
-                            log: format!("{:#}", Report::new(error)),
-                            ..Default::default()
-                        });
+                        tx_results.push((
+                            None,
```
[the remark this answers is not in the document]
ethanoroshiba: Done in 9d386ee.

crates/astria-sequencer/src/mempool/mod.rs
ethanoroshiba: Used JSON-encoded result in 9d386ee.

proto/mempoolapis/astria/mempool/v1/transaction_status.proto
ethanoroshiba: It seems we have many mentions of tendermint proto types in our vendored package. Pulled the `Event` and `EventAttribute` out of `cosmos_sdk` in 9d386ee, and instead moved them as well as `ExecTxResult` into a new `tendermint` package, since their definitions are essentially lifted from Tendermint's Go repo. Let me know if this works for you!

Fraser999: Works for me thanks.  Other reviewers might disagree though :)

Fraser999 (approved):

proto/mempoolapis/astria/mempool/v1/transaction_status.proto
```diff
@@ -41,5 +43,15 @@ message TransactionStatus {
...
+
+    // The transaction result.
+    tendermint.abci.ExecTxResult result = 2;
+
+    // EventAttribute is a single key-value pair, associated with an event.
+    message EventAttribute {
+      string key = 1;
+      string value = 2;
+      bool index = 3;
+    }
```
joroshiba, line 55: Shouldn't be needed?

ethanoroshiba: Good catch, forgot to remove that one. Done in f3bc8d1.

PR #2160, opened 2025-05-07
feat(sequencer): add transaction fee gRPC query

### LINKED ISSUE #2152: Add gRPC transaction fee query
#2133 introduced the mempool API along with gRPC endpoints for `GetTransactionStatus` and `SubmitTransaction`. It would be good to also create an RPC for transaction fees similar to our ABCI query.

ethanoroshiba:
## Summary
Adds transaction fee gRPC query to the Mempool API.

## Background
We have an ABCI query for fees; now that we are shifting towards gRPC usage it would be good to have a gRPC endpoint for obtaining a transaction's fees as well. This change is meant to address this.

## Changes
- Added `GetTransactionFeeRequest` and `GetTransactionFeeResponse` to `mempoolapis`.
- Added `GetTransactionFees` to `TransactionService`.
- Added implementation of `get_transaction_fees` in the mempool server, utilizing the same logic as in `fees::query`.

## Testing
Added one failure case unit test and one success case test.

## Changelogs
Changelog updated.

## Related Issues
closes #2152

### DIFF
--- a/crates/astria-sequencer/src/grpc/mempool.rs
+++ b/crates/astria-sequencer/src/grpc/mempool.rs
@@ -24,7 +24,8 @@ use astria_core::{
         TransactionId,
         TRANSACTION_ID_LEN,
     },
-    protocol::transaction::v1::Transaction,
+    protocol::transaction::v1::TransactionBody,
+    Protobuf as _,
 };
 use bytes::Bytes;
 use cnidarium::Storage;
@@ -115,26 +116,27 @@ impl TransactionService for Server {
         self: Arc<Self>,
         request: Request<GetTransactionFeesRequest>,
     ) -> Result<Response<GetTransactionFeesResponse>, Status> {
-        let unchecked_tx: Transaction = request
-            .into_inner()
-            .transaction
-            .ok_or_else(|| Status::invalid_argument("transaction is empty"))?
-            .try_into()
-            .map_err(|err| {
-                Status::invalid_argument(format!("failed to decode transaction from raw: {err}"))
-            })?;
+        let tx_body = TransactionBody::try_from_raw(
+            request
+                .into_inner()
+                .transaction_body
+                .ok_or_else(|| Status::invalid_argument("transaction is empty"))?,
+        )
+        .map_err(|err| {
+            Status::invalid_argument(format!("failed to decode transaction from raw: {err}"))
+        })?;
 
         let snapshot = self.storage.latest_snapshot();
         let block_height = snapshot
             .get_block_height()
             .await
             .map_err(|err| Status::internal(format!("failed to get block height: {err}")))?;
-        let fees_with_ibc_denoms = total_fees(
-            unchecked_tx.actions().iter().map(ActionRef::from),
-            &snapshot,
-        )
-        .await
-        .map_err(|err| Status::internal(format!("failed to get fees for transaction: {err}")))?;
+        let fees_with_ibc_denoms =
+            total_fees(tx_body.actions().iter().map(ActionRef::from), &snapshot)
+                .await
+                .map_err(|err| {
+                    Status::internal(format!("failed to get fees for transaction: {err}"))
+                })?;
         let mut fees = Vec::with_capacity(fees_with_ibc_denoms.len());
         for (ibc_denom, value) in fees_with_ibc_denoms {
             let trace_denom = match snapshot.map_ibc_to_trace_prefixed_asset(&ibc_denom).await {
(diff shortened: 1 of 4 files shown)


proto/mempoolapis/astria/mempool/v1/get_transaction_fees_response.proto
```diff
@@ -0,0 +1,13 @@
+syntax = "proto3";
+
+package astria.mempool.v1;
+
+import "astria/protocol/fees/v1/types.proto";
+
+// A response to a request for the fees of a transaction on the Astria Sequencer.
+message GetTransactionFeesResponse {
+  // The height at which the fees were calculated.
+  uint64 height = 1;
```
Fraser999, line 10:
```suggestion
  // The block height at which the fees were calculated.
  uint64 block_height = 1;
```

crates/astria-sequencer/src/grpc/mempool.rs
```diff
@@ -98,6 +110,66 @@ impl TransactionService for Server {
             duplicate: submission_outcome.duplicate,
         }))
     }
+
+    async fn get_transaction_fees(
+        self: Arc<Self>,
+        request: Request<GetTransactionFeesRequest>,
+    ) -> Result<Response<astria_core::generated::mempool::v1::GetTransactionFeesResponse>, Status>
```
Fraser999, line 117:
```suggestion
    ) -> Result<Response<GetTransactionFeesResponse>, Status>
```

crates/astria-sequencer/src/grpc/mempool.rs
```diff
@@ -98,6 +110,66 @@ impl TransactionService for Server {
...
+                .encode_to_vec()
+                .into(),
+            &snapshot,
+        )
+        .await
+        .map_err(|err| {
+            Status::invalid_argument(format!(
+                "Failed to construct checked transaction from raw: {err}"
+            ))
+        })?;
```
Fraser999, line 134: I think we should just use an unchecked `Transaction` here to avoid having the server do extra work, and also so that users can check txs which maybe aren't necessarily valid under the current global state.

crates/astria-sequencer/src/grpc/mempool.rs
```diff
@@ -98,6 +110,66 @@ impl TransactionService for Server {
...
+
+        let snapshot = self.storage.latest_snapshot();
+        let height = snapshot
+            .get_block_height()
+            .await
+            .map_err(|err| Status::internal(format!("Failed to get block height: {err}")))?;
```
Fraser999, line 140: We have a mixture of status messages starting with a lowercase and uppercase letter.  Probably best to stick to lowercase everywhere for error messages, since I think that's the more prevalent form?

proto/mempoolapis/astria/mempool/v1/get_transaction_fees_response.proto
ethanoroshiba: Done in 08e2195.

crates/astria-sequencer/src/grpc/mempool.rs
ethanoroshiba: Done in 08e2195.

crates/astria-sequencer/src/grpc/mempool.rs
ethanoroshiba: I was debating the same thing, but wasn't sure since the the checked transaction construction also provides immutable logic such as checking transaction size. The more I think about it, though, I think you're right and it would be much better UX that way. Done in 08e2195.

crates/astria-sequencer/src/grpc/mempool.rs
ethanoroshiba: Ah yeah that's an oversight on my part, I guess I just went on autopilot. Good catch! Done in 08e2195.

Fraser999 (approved):

proto/mempoolapis/astria/mempool/v1/get_transaction_fees_request.proto
```diff
@@ -0,0 +1,10 @@
+syntax = "proto3";
+
+package astria.mempool.v1;
+
+import "astria/protocol/transaction/v1/transaction.proto";
+
+// A request to get the fees for the given transaction on the Astria Sequencer.
+message GetTransactionFeesRequest {
+  astria.protocol.transaction.v1.Transaction transaction = 1;
```
joroshiba, line 9: We should probably take a `TransactionBody` instead of a full `Transaction` here?

Signing a transaction should be considered authorization to execute, generally shouldn't be broadcast unless we are okay with it being executed. If I'm checking fees its probably fair to assume I'm not ready to authorize execution. If we need to account information at all for fees we could pass that as a seperate field, but I don't think we currently do need it?

ethanoroshiba: Agreed! Done in 9518ba3.

Issue #2157, opened 2025-05-06
Replace sequencer `app_legacy_execute_transactions_with_every_action_snapshot` test


Fraser999:
PR #2142 provides a new test fixture used extensively throughout the crate to initialize global state and provide checked transactions for use in tests.

That PR also renamed the snapshot test `app_execute_transaction_with_every_action_snapshot` to `app_legacy_execute_transactions_with_every_action_snapshot` and provided methods on the test fixture to support initializing global state in exactly the same way as before the PR, so that the snapshot test continued to pass.

Once that PR is merged to `main`, we can remove the legacy test and supporting helpers in the fixture, replacing it with a new snapshot test with the same end goal.

The new test should aim to execute every variant of every action, with only one action per tx in case we change which actions are bundleable in the future.  Note that the new fixture provides helpers to allow for an `IbcRelay::CreateClient` action to be executed (see [this unit test](https://github.com/astriaorg/astria/blob/003a1e4a21b02052970fae91d23dc66c84ddc275/crates/astria-sequencer/src/checked_actions/ibc_relay.rs#L219-L242) for details).  Testing this one variant of `IbcRelay` is probably enough, and is certainly an improvement over not testing any `IbcRelay` actions.

NOTE: There are a few places in the production code where an exception to the normal checking rules for actions has been added.  Each such code block has a comment mentioning `app_legacy_execute_transactions_with_every_action_snapshot`, so these code blocks should all be removed as part of this ticket's work too.

joroshiba:
This issue is stale because it has been open 45 days with no activity. Remove stale label or this issue
be closed in 7 days.

joroshiba:
This issue was closed because it was stale

PR #2171, opened 2025-05-13
chore(sequencer): update handling of addition and removal actions

### LINKED ISSUE #2158: Change behaviour of tx actions which add or remove items
Currently we have a variety of actions which add, remove or update data in global state.  We do not have a consistent approach to handling addition where the item already exists, or deletion where the item doesn't exist.

The following table describes the current state of play in this regard, and suggested changes to behaviour:

| Action | Current Behaviour | New Behaviour |
| :--- | :--- | :--- |
| `CurrencyPairsChange::Addition` | Ignore (skips all pairs which already exist) | Fail entire action if any pair already exists |
| `CurrencyPairsChange::Removal` | Ignore (skips all pairs which don't exist) | Fail entire action if any pair doesn't exist |
| `FeeAssetChange::Addition` | Ignore (doesn't care if asset already exists) | Fail if asset already exists |
| `FeeAssetChange::Removal` | Ignore (doesn't care if asset doesn't exist) | Fail if asset doesn't exist |
| `IbcRelayerChange::Addition` | Ignore (doesn't care if IBC relayer address already exists) | Fail if IBC relayer address already exists |
| `IbcRelayerChange::Removal` | Ignore (doesn't care if IBC relayer address doesn't exist) | Fail if IBC relayer address doesn't exist |
| `MarketsChange::Creation` | Fail if market for any pair already exists | No change to current |
| `MarketsChange::Removal` | Ignore (doesn't care if any market for pair doesn't exist) | Fail if any market for pair doesn't exist |
| `MarketsChange::Update` | Fail if market for any pair doesn't exist | No change to current |
| `ValidatorUpdate` with power != 0 | Ignore (doesn't care if validator already exists) | No change to current - is notionally an upsert rather than addition |
| `ValidatorUpdate` with power == 0 | Fail if validator doesn't exist | No change to current |

Fraser999:
## Summary
This changes how actions involving addition or removal of data are handled.

## Background
In some cases currently, we silently ignore additions where the data already exists, or removals where the data does not exist.  We want to always treat these as error conditions.

## Changes
- Changed `CurrencyPairsChange::Addition` action to fail if any currency pair to be added is already stored.
- Changed `CurrencyPairsChange::Removal` action to fail if any currency pair to be removed is not currently stored.
- Changed `FeeAssetChange::Addition` action to fail if the fee asset to be added is already stored.
- Changed `FeeAssetChange::Removal` action to fail if the fee asset to be removed is not currently stored.
- Changed `IbcRelayerChange::Addition` action to fail if the address to be added is already stored.
- Changed `IbcRelayerChange::Removal` action to fail if the address to be removed is not currently stored.
- Changed `MarketsChange::Removal` action to fail if any market to be removed is not currently stored.

## Testing
Added/updated unit tests for all of the changed behaviour.

## Changelogs
Changelogs updated.

## Related Issues
Closes #2158.

### DIFF
--- a/crates/astria-core/CHANGELOG.md
+++ b/crates/astria-core/CHANGELOG.md
@@ -58,7 +58,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `CheckedTransaction` from this [#2142](https://github.com/astriaorg/astria/pull/2142).
 - Rename ABCI error code 10 from `TRANSACTION_FAILED` to
   `TRANSACTION_FAILED_EXECUTION` [#2142](https://github.com/astriaorg/astria/pull/2142).
-- Change `CurrencyPairsChange` action to hold a `BTreeSet` rather than a `Vec`
+- Change `CurrencyPairsChange` action to hold an `IndexSet` rather than a `Vec`
   of currency pairs [#2171](https://github.com/astriaorg/astria/pull/2171).
 
 ### Removed
--- a/crates/astria-core/CHANGELOG.md
+++ b/crates/astria-core/CHANGELOG.md
@@ -58,6 +58,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `CheckedTransaction` from this [#2142](https://github.com/astriaorg/astria/pull/2142).
 - Rename ABCI error code 10 from `TRANSACTION_FAILED` to
   `TRANSACTION_FAILED_EXECUTION` [#2142](https://github.com/astriaorg/astria/pull/2142).
+- Change `CurrencyPairsChange` action to hold a `BTreeSet` rather than a `Vec`
+  of currency pairs [#2171](https://github.com/astriaorg/astria/pull/2171).
 
 ### Removed
 
--- a/crates/astria-core/src/protocol/transaction/v1/action/mod.rs
+++ b/crates/astria-core/src/protocol/transaction/v1/action/mod.rs
@@ -1,5 +1,4 @@
 use std::{
-    collections::BTreeSet,
     fmt::Display,
     str::FromStr,
 };
@@ -12,6 +11,7 @@ use ibc_types::{
     },
     IdentifierError,
 };
+use indexmap::IndexSet;
 use penumbra_ibc::IbcRelay;
 use prost::Name as _;
 
@@ -2580,8 +2580,8 @@ enum RecoverIbcClientErrorKind {
 
 #[derive(Debug, Clone)]
 pub enum CurrencyPairsChange {
-    Addition(BTreeSet<CurrencyPair>),
-    Removal(BTreeSet<CurrencyPair>),
+    Addition(IndexSet<CurrencyPair>),
+    Removal(IndexSet<CurrencyPair>),
 }
 
 impl Protobuf for CurrencyPairsChange {
@@ -2623,22 +2623,30 @@ impl Protobuf for CurrencyPairsChange {
             Some(raw::currency_pairs_change::Value::Addition(raw::CurrencyPairs {
                 pairs,
             })) => {
-                let pairs = pairs
+                let pairs_len = pairs.len();
+                let parsed_pairs: IndexSet<_> = pairs
                     .into_iter()
                     .map(CurrencyPair::try_from_raw)
                     .collect::<Result<_, _>>()
                     .map_err(Self::Error::invalid_currency_pair)?;
-                Ok(Self::Addition(pairs))
+                if pairs_len != parsed_pairs.len() {
+                    return Err(Self::Error::duplicate_currency_pair());
+                }
+                Ok(Self::Addition(parsed_pairs))
             }
             Some(raw::currency_pairs_change::Value::Removal(raw::CurrencyPairs {
                 pairs,
             })) => {
-                let pairs = pairs
+                let pairs_len = pairs.len();
+                let parsed_pairs: IndexSet<_> = pairs
                     .into_iter()
                     .map(CurrencyPair::try_from_raw)
                     .collect::<Result<_, _>>()
                     .map_err(Self::Error::invalid_currency_pair)?;
-                Ok(Self::Removal(pairs))
+                if pairs_len != parsed_pairs.len() {
+                    return Err(Self::Error::duplicate_currency_pair());
+                }
+                Ok(Self::Removal(parsed_pairs))
             }
             None => Err(Self::Error::unset()),
         }
@@ -2669,6 +2677,11 @@ impl CurrencyPairsChangeError {
     fn invalid_currency_pair(err: CurrencyPairError) -> Self {
         Self(CurrencyPairsChangeErrorKind::InvalidCurrencyPair(err))
     }
+
+    #[must_use]
+    fn duplicate_currency_pair() -> Self {
+        Self(CurrencyPairsChangeErrorKind::DuplicateCurrencyPair)
+    }
 }
 
 #[derive(Debug, thiserror::Error)]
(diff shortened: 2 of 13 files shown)


crates/astria-core/src/protocol/transaction/v1/action/mod.rs
```diff
@@ -2579,8 +2580,8 @@ enum RecoverIbcClientErrorKind {
 
 #[derive(Debug, Clone)]
 pub enum CurrencyPairsChange {
-    Addition(Vec<CurrencyPair>),
-    Removal(Vec<CurrencyPair>),
+    Addition(BTreeSet<CurrencyPair>),
```
SuperFluffy, line 2583: I wonder if this could be changed to use an `IndexSet` because these are usually faster than btrees. I think we are already having it as part of our public API anyway.

Fraser999: I went for the `BTreeSet` here since I thought having these ordered alphabetically would be more useful from a human-readable perspective.  `IndexSet` wouldn't do that (it retains the insertion order, more like a `Vec`) unless we explicitly call `sort_by` on it.

I doubt that performance is important here - we're likely dealing with small collections (so perf difference will be insignificant) and this action will likely be called very infrequently.

I could see a case to made for using `IndexSet` purely to retain the insertion order; it's maybe less surprising for a user creating an instance of this action.  But I tend to favour the alphabetical approach since it might be more obvious to the user that creating this action removes duplicates, which is important now that we fail execution for a duplicate entry.

crates/astria-core/src/protocol/transaction/v1/action/mod.rs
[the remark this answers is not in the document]
Fraser999: Done in [d4d2c04](https://github.com/astriaorg/astria/pull/2171/commits/d4d2c043ac2b4119979177f5b9d82cd4f59dfc9d).

crates/astria-core/src/protocol/transaction/v1/action/mod.rs
joroshiba: Is there implications on the serialization order vs the execution here, and is there anything there that might matter?

IE I have a proto with an addition/removal action and a list of currency pairs. Those will be added or removed in alphabetical order instead of the order I specified them in?

Fraser999: I think the order in which these are _actually_ applied is invisible to the user.  The whole action will either succeed or fail as part of a tx: there's no exposure to a partial change whereby some of the additions/removals have been applied and some have not.

The only thing which I think might be surprising to the user is that (assuming the action is created via the proto object, e.g. via JSON rather than via e.g. a new CLI command) then the original order is changed to alphabetical and probably more importantly, duplicates have been removed.  But I'm not sure if there's anywhere the user has access to the parsed `Transaction` rather than the proto one.

Fraser999: Potentially it might be more honest to have the decoding fail if the proto contains duplicates, rather than just silently removing them for the user.  If we want to do that, then I guess `IndexSet` is back on the table, since there'd be nothing to warn the user about - we're no longer removing duplicates.

Fraser999: Changed to `IndexSet` and now fails to parse if the action contains duplicates.  Done in [a3ee841](https://github.com/astriaorg/astria/pull/2171/commits/a3ee841369ed28d328258c420a3066203a9c6bec).

crates/astria-core/CHANGELOG.md
```diff
@@ -58,6 +58,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `CheckedTransaction` from this [#2142](https://github.com/astriaorg/astria/pull/2142).
 - Rename ABCI error code 10 from `TRANSACTION_FAILED` to
   `TRANSACTION_FAILED_EXECUTION` [#2142](https://github.com/astriaorg/astria/pull/2142).
+- Change `CurrencyPairsChange` action to hold a `BTreeSet` rather than a `Vec`
```
noot, line 61:
```suggestion
- Change `CurrencyPairsChange` action to hold a `IndexSet` rather than a `Vec`
```

crates/astria-sequencer/src/checked_actions/fee_asset_change.rs
```diff
@@ -66,6 +67,48 @@ impl CheckedFeeAssetChange {
             &sudo_address == self.tx_signer.as_bytes(),
             "transaction signer not authorized to change fee assets"
         );
+
+        // NOTE: To allow `app_legacy_execute_transactions_with_every_action_snapshot` to continue
+        // to pass, we need to make an exception for `test-0` here to allow the test tx to be
+        // constructed. This exception can be removed once the legacy test is removed.
+        #[cfg(test)]
```
noot, line 74: what is the legacy test for? is it to ensure backwards-compatibility? can we change the test to have `test-0` be an existing asset (which will change the snapshot but can add an explanation of why and that it's not breaking)

Fraser999: It was really to help ensure backwards-compatibility during the large changeset in #2142.  Now that that PR has merged, it's safe to replace that legacy test with something more convenient, and which will let us remove these workarounds altogether.

This is ticketed in #2157.

crates/astria-core/CHANGELOG.md
Fraser999: Fixed in [97a90d8](https://github.com/astriaorg/astria/pull/2171/commits/97a90d8e0738cea54d6556314fd276af53300510).

crates/astria-core/src/protocol/transaction/v1/action/mod.rs
```diff
@@ -2622,22 +2623,30 @@ impl Protobuf for CurrencyPairsChange {
...
-                let pairs = pairs
+                let pairs_len = pairs.len();
+                let parsed_pairs: IndexSet<_> = pairs
                     .into_iter()
                     .map(CurrencyPair::try_from_raw)
                     .collect::<Result<_, _>>()
                     .map_err(Self::Error::invalid_currency_pair)?;
-                Ok(Self::Addition(pairs))
+                if pairs_len != parsed_pairs.len() {
+                    return Err(Self::Error::duplicate_currency_pair());
```
SuperFluffy, line 2633: From a UX pov it'd be nice to know what the dupe was, but also no big deal

crates/astria-core/src/protocol/transaction/v1/action/mod.rs
```diff
@@ -2622,22 +2623,30 @@ impl Protobuf for CurrencyPairsChange {
...
-                let pairs = pairs
+                let pairs_len = pairs.len();
+                let parsed_pairs: IndexSet<_> = pairs
                     .into_iter()
                     .map(CurrencyPair::try_from_raw)
                     .collect::<Result<_, _>>()
                     .map_err(Self::Error::invalid_currency_pair)?;
-                Ok(Self::Removal(pairs))
+                if pairs_len != parsed_pairs.len() {
+                    return Err(Self::Error::duplicate_currency_pair());
```
SuperFluffy, line 2647: As above: not a blocker, but knowing the dupe(s) is a nice property

SuperFluffy (approved):

crates/astria-core/src/protocol/transaction/v1/action/mod.rs
Fraser999: Yeah - I considered that, but I figured the trade-off of less code was worth it since it'll only be sudo user creating these and the collection will likely be small enough that it's not too difficult to just spot the dupes.  If I was putting in handling for listing any dupes in the error, I'd want to list _all_ dupes, which is slightly more involved.

PR #2201, opened 2025-05-27
fix(sequencer)!: allow IBC Relay actions to fail execution

### LINKED ISSUE #2200: Allow failing IBC Relay actions
We utilize Hermes relayers to facilitate cross-chain transfers via IBC. One longstanding goal has been to operate multiple relayers per channel for redundancy and expediency. This scenario, however, creates a race condition where if both relayers have submitted transactions that end up being proposed in the same block, the first will be executed and the next will fail execution. On validator nodes, the transaction will eventually be cleared out of all mempools due to failed execution during `prepare_proposal`, but in non-validating full-nodes, however, the transaction will be stuck in `Pending` until it times out, causing one relayer to stall during this time.

Allowing inclusion of failed `IbcRelay` actions will fix this issue, and will be safe due to it being a whitelisted action.

ethanoroshiba:
## Summary
Allows for failing `IbcRelay` actions to still be included.

## Background
When two Hermes relayers are connected via the same channel on a single sequencer, a race condition is created wherein if both have their submitted transactions proposed in the same block, one will fail execution since it is a duplicate. If connected to a validator, this should be fine, since the transactions will be removed from the mempool's `Pending` cache after failing execution during `PrepareProposal`. If the relayers are connected to a non-validating full-node, however, the transaction will be stuck in `Pending` until it times out, causing one of the relayers to stall. Allowing failing `IbcRelay` actions to be included will fix this problem, and will not create an attack vector since these actions are already permissioned.

## Changes
- Allow `IbcRelay` actions to fail once their sender address has been validated to be an IBC Relayer.

## Testing
- Added E2E test with multiple relayers to ensure that all transfers are executed correctly and neither relayer stalls.

## Changelogs
Changelog updated.

## Metrics
- Added counter `INCLUDED_TRANSACTIONS_FAILED_EXECUTION`.

## Breaking Changelist

This is a protocol breaking change in the following scenario: the Proposer for the current round is updated to allow for failing `IbcRelay` actions, but Validator A is not. Proposer calls `PrepareProposal` and prepares a block with a failing `IbcRelay` action. Upon receiving this block in `ProcessProposal`, Validator A will fail.

Because of this, there are 3 ways this update could be applied:
1. If we are confident that submitted `IbcRelay` actions will not fail while we are only using 1 relayer, we can simply update all sequencers before spinning up other relayers.
2. If we are not confident in this, we can temporarily halt IBC relaying via `IbcRelayerChange::Remove` while the sequencers are updated.
3. We can schedule a new upgrade to include this change.

## Related Issues
closes #2200

### DIFF
--- a/system-tests/helpers/astria_cli.py
+++ b/system-tests/helpers/astria_cli.py
@@ -103,6 +103,20 @@ class Cli:
         except Exception as error:
             raise SystemExit(error)
 
+    def add_utia_asset(self):
+        try:
+            self._try_exec_sequencer_command_with_retry(
+                "sudo",
+                "fee-asset",
+                "add",
+                "--sequencer.chain-id=sequencer-test-chain-0",
+                "--private-key=2bd806c97f0e00af1a1fc3328fa763a9269723c8db8fac4f93af71db186d6e90",
+                "--asset=transfer/channel-0/utia",
+                sequencer_name="node0"
+            )
+        except Exception as error:
+            raise SystemExit(error)
+
     def address(self, sequencer_name, address_bytes):
         try:
             return self._try_exec_sequencer_command(
--- a/system-tests/multiple_hermes_relayer_test.py
+++ b/system-tests/multiple_hermes_relayer_test.py
@@ -78,6 +78,8 @@ if cli_image is None:
     cli_image = "latest"
 cli = Cli(cli_image)
 
+cli.add_utia_asset()
+
 # Check starting sequencer balance
 print(colored("Checking starting balance on sequencer...", "blue"))
 try:


system-tests/helpers/celestia_controller.py
```diff
@@ -0,0 +1,68 @@
...
+    IBC_TRANSFER_AMOUNT,
+)
+
+BRIDGE_ADDRESS = ""
+CELESTIA_DEV_ACCOUNT_ADDRESS = "celestia1m0ksdjl2p5nzhqy3p47fksv52at3ln885xvl96"
+CELESTIA_CHAIN_ID = "celestia-local-0"
+KEYRING_BACKEND = "test"
+IBC_TRANSFER_FEES = "26000"
+
+class CelestiaController:
```
ethanoroshiba, line 16: With the advent of the `CelestiaController`, should the other tests be switched to use it (instead of the `just deploy celestia-local` recipe) now or in a follow-up?

crates/astria-sequencer/src/checked_actions/ibc_relay.rs
```diff
@@ -86,11 +89,29 @@ impl CheckedIbcRelay {
     #[instrument(skip_all, err(level = Level::DEBUG))]
     pub(super) async fn execute<S: StateWrite>(&self, state: S) -> Result<()> {
         self.run_mutable_checks(&state).await?;
```
Fraser999, line 91: Might be worthwhile to also just bump the failure count here and return?  If we do that, then the whole `execute()` method can be changed to return nothing.

system-tests/helpers/celestia_controller.py
Fraser999: I'd suggest a follow-up for that.

crates/astria-sequencer/src/checked_actions/ibc_relay.rs
ethanoroshiba: I think it could still be important to fail in the case of the signer not matching the sudo address, otherwise block space could theoretically be taken up for free by non-sudo. Though, by this point it would've had to pass the checks during construction, which would indicate that it was sudo at some point, so it probably wouldn't matter too much. I can go ahead and make this change if you think it's still pertinent.

Fraser999: Yeah - probably doesn't matter much either way.  Happy for this to stay as is.

crates/astria-sequencer/src/checked_actions/ibc_relay.rs
```diff
@@ -86,11 +89,29 @@ impl CheckedIbcRelay {
...
+                .ephemeral_get_ibc_failure_count()
+                .unwrap_or_default()
+                .saturating_add(1);
+            state.ephemeral_put_ibc_failure_count(failure_count);
+            debug!(err = %e, "failed to execute IBC Relay action, still including in block");
+        } else {
+            let (mut state, events) = ibc_delta.apply();
+            events.into_iter().for_each(|event| state.record(event));
+        }
+        Ok(())
```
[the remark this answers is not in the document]
Fraser999: Good call.  Done in [8b78022](https://github.com/astriaorg/astria/pull/2201/commits/8b78022cc01dd636bdc4f4c7bcea9a5b639b1044).

crates/astria-sequencer/src/checked_actions/ibc_relay.rs
```diff
@@ -240,4 +270,40 @@ mod tests {
...
+            .unwrap()
+            .into();
+
+        assert!(checked_action
+            .action_with_handlers
+            .check_and_execute(fixture.state_mut())
+            .await
+            .is_err());
+        assert!(checked_action.execute(fixture.state_mut()).await.is_ok());
+        assert_eq!(
```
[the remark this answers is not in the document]
Fraser999: This test was removed in [8b78022](https://github.com/astriaorg/astria/pull/2201/commits/8b78022cc01dd636bdc4f4c7bcea9a5b639b1044).

aajimal (approved):

joroshiba (approved):
stamp

PR #2221, opened 2025-06-01
fix(telemetry)!: fix memory leak in metrics

Fraser999:
## Summary
This fixes an issue where no-op metrics consumed an every-increasing amount of memory without freeing it.

## Background
If metrics are not enabled (e.g. by setting `ASTRIA_SEQUENCER_NO_METRICS=true`), a no-op instance is created which is intended to do nothing when metrics are updated as the process runs.

The metrics returned from the `Metrics::noop_metrics()` function are registered via an instance of a `PrometheusRecorder` which is immediately dropped rather than being passed into `metrics::set_global_recorder` with the assumption that it will then render all subsequent calls to metric functions no-ops.

This is not the case though (which is obvious in hindsight - a handle to the recorder is held by each of the individual counters, gauges and histograms).

This alone didn't result in the leak.  The further problem is that when creating the `PrometheusRecorder` in `noop_metrics()`, we call `PrometheusBuilder::build_recorder()`.  [The docs](https://docs.rs/metrics-exporter-prometheus/latest/metrics_exporter_prometheus/#upkeep-and-maintenance) describe that using this function requires a maintenance task to be run; essentially garbage collection.

## Changes
- Provided a `NoopRecorder` and used this inside `Metrics::noop_metrics()`.
- Start the maintenance task for the case where metrics are enabled, but no listening address is provided.

## Testing
Load-tested a standalone sequencer configured to have metrics disabled, and observed the memory leak to have been fixed. The load test caused many calls to a histogram, and before the fix, this caused the memory usage to constantly increase.

## Changelogs
Changelogs updated.

## Breaking Changelist
- Changes the public Rust API of the `astria-telemetry` crate.

### DIFF
--- a/crates/astria-telemetry/src/metrics/builders.rs
+++ b/crates/astria-telemetry/src/metrics/builders.rs
@@ -8,12 +8,10 @@ use std::{
     net::SocketAddr,
 };
 
-use metrics::Recorder as _;
 use metrics_exporter_prometheus::{
     ExporterFuture,
     Matcher,
     PrometheusBuilder,
-    PrometheusRecorder,
 };
 
 #[cfg(doc)]
@@ -115,6 +113,16 @@ impl ConfigBuilder {
                 .map_err(|error| Error::StartListening(error.into()))?
         } else {
             let recorder = bucket_builder.builder.build_recorder();
+
+            let recorder_handle = recorder.handle();
+            tokio::spawn(async move {
+                let mut interval = tokio::time::interval(std::time::Duration::from_secs(5));
+                loop {
+                    interval.tick().await;
+                    recorder_handle.run_upkeep();
+                }
+            });
+
             let fut: ExporterFuture = Box::pin(async move { Ok(()) });
             (recorder, fut)
         };
@@ -194,14 +202,14 @@ impl BucketBuilder {
 /// A builder used to register individual metrics.
 ///
 /// It is constructed in [`ConfigBuilder::build`] and passed to [`Metrics::register`].
-pub struct RegisteringBuilder {
-    recorder: PrometheusRecorder,
+pub struct RegisteringBuilder<R> {
+    recorder: R,
     counters: HashSet<String>,
     gauges: HashSet<String>,
     histograms: HashSet<String>,
 }
 
-impl RegisteringBuilder {
+impl<R: metrics::Recorder> RegisteringBuilder<R> {
     /// Returns a new `CounterFactory` for registering [`Counter`]s under the given name.
     ///
     /// # Errors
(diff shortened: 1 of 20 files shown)


noot (approved):

crates/astria-telemetry/src/metrics/builders.rs
```diff
@@ -115,6 +113,16 @@ impl ConfigBuilder {
                 .map_err(|error| Error::StartListening(error.into()))?
         } else {
             let recorder = bucket_builder.builder.build_recorder();
+
+            let recorder_handle = recorder.handle();
+            tokio::spawn(async move {
+                let mut interval = tokio::time::interval(std::time::Duration::from_secs(5));
```
SuperFluffy, line 119: This likely is not an issue, but let's set this to to `MissedTickBehavior::Skip` to be conservative: https://docs.rs/tokio/1.45.1/tokio/time/struct.Interval.html#method.set_missed_tick_behavior

crates/astria-telemetry/src/metrics/builders.rs
```diff
@@ -194,14 +202,14 @@ impl BucketBuilder {
 /// A builder used to register individual metrics.
 ///
 /// It is constructed in [`ConfigBuilder::build`] and passed to [`Metrics::register`].
-pub struct RegisteringBuilder {
-    recorder: PrometheusRecorder,
+pub struct RegisteringBuilder<R> {
```
SuperFluffy, line 205: I don't like that this infects all instances where metrics are being instantiated. I haven't dug deep into this code, but can we type erase this recorder so that we don't have to thread the `R` everywhere?

crates/astria-telemetry/src/metrics/builders.rs
```diff
@@ -115,6 +113,16 @@ impl ConfigBuilder {
                 .map_err(|error| Error::StartListening(error.into()))?
         } else {
             let recorder = bucket_builder.builder.build_recorder();
+
```
SuperFluffy, line 116: Above `fn build` (the scope of the lines I am commenting on): can you add a section that this spawns a task on the tokio runtime and will fail if running outside of tokio?

SuperFluffy (approved):
Change looks fine, but from a usability perspective I wonder if making the recorder `Box<dyn Recorder>` (or maybe `Arc<dyn Recorder>` or `Box<dyn Recorder + Send>`, whatever is actually needed)?

The implementations allow for that: https://docs.rs/metrics/0.24.2/metrics/trait.Recorder.html#foreign-impls

crates/astria-telemetry/src/metrics/builders.rs
Fraser999: Done in [780a46b](https://github.com/astriaorg/astria/pull/2221/commits/780a46bca9606a0af9a70250c2e825d4e9e7a86a).

joroshiba:

PR #2225, opened 2025-06-03
fix(sequencer): fix race condition in mempool

### LINKED ISSUE #2131: Race condition in mempool
@ethanoroshiba has identified a race condition between `Mempool::insert` and `Mempool::run_maintenance`.

`insert` is called with values (`current_account_nonce`, `current_account_balances` and `transaction_cost`) drawn from a state snapshot which may be older than the one being used by `run_maintenance`.

For example, let's say the state changes from executing block 10 are about to be committed, and at the same time, we receive a call to CheckTx for a new tx.

CheckTx (which eventually calls through to `Mempool::insert`) will use the snapshot from block 9, while `run_maintenance` will be using the snapshot from block 10.  If `run_maintenance` happens to acquire the mempool locks first, it will cause `insert` to use outdated state: all of the passed-in values mentioned above could be incorrect as per the latest state.

Some of these problems will be resolved during the next `run_maintenance` (for example, the new tx could be inserted incorrectly into the parked queue, but would be promoted to pending on `run_maintenance`) causing at worst a delay in execution.  However, a tx with invalid associated costs could be held in parked until timed out and never executed.

There are potentially a handful of ways to resolve this.  At the moment, I think the best approach is to provide the mempool with a copy of `Storage::latest_snapshot()` as a member variable.  It will be provided via `Mempool::new()` and will be updated once per block via `Mempool::run_maintenance()`.  The signature of `insert` should be simplified to only take the new tx, and likewise `run_maintenance` should be simplified to only have the new `Snapshot` passed in.

Recommend blocking this on #1958 and #2119 to avoid unnecessary conflicts.

Fraser999:
## Summary
This fixes a race condition in the mempool.

## Background
Details of the issue are specified in #2131.  This fixes the problematic race condition by moving the retrieval of relevant data from global state to inside the mempool, where it is done while holding a lock on the mempool's `RwLock`.

## Changes
- Added a state `Snapshot` as a member variable of `Mempool`.  This is updated on every call to `run_maintenance`.
- Changed the insertion of tx to fetch the required data from this snapshot rather than having it being provided from outside of Mempool.

## Testing
Existing tests updated.

## Changelogs
Changelogs updated.

## Metrics
- Removed:
  - `astria_sequencer_check_tx_duration_seconds_fetch_nonce`
  - `astria_sequencer_check_tx_duration_seconds_fetch_balances`
  - `astria_sequencer_check_tx_duration_seconds_fetch_tx_cost`

## Related Issues
Closes #2131.

### DIFF
--- a/crates/astria-sequencer/CHANGELOG.md
+++ b/crates/astria-sequencer/CHANGELOG.md
@@ -14,6 +14,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add cache of recent execution results to mempool [#2163](https://github.com/astriaorg/astria/pull/2163).
 - Add tx result to `Executed` transaction status [#2159](https://github.com/astriaorg/astria/pull/2159).
 
+### Removed
+
+- Remove metrics:
+  - `astria_sequencer_check_tx_duration_seconds_fetch_nonce`
+  - `astria_sequencer_check_tx_duration_seconds_fetch_balances`
+  - `astria_sequencer_check_tx_duration_seconds_fetch_tx_cost` [#2225](https://github.com/astriaorg/astria/pull/2225).
+
+### Fixed
+
+- Fix race condition in mempool [#2225](https://github.com/astriaorg/astria/pull/2225).
+
 ## [3.0.0] - 2025-05-21
 
 ### Added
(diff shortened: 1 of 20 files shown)


joroshiba:

PR #2231, opened 2025-06-17
Add TLS config for composer when sequencer endpoint is tls

mycodecrafting:
## Summary
Add explicit TLS config for composer's sequencer connection

## Background
Tonic [removed implicit TLS configs](https://github.com/hyperium/tonic/pull/1731) in v0.12.0 which now causes TLS errors when conductor tries to connect to a remote sequencer network over TLS.

## Changelogs
Changelogs updated.

## Related Issues
Similar to fix done for conductor in [PR #2140](https://github.com/astriaorg/astria/pull/2140)

### DIFF
--- a/crates/astria-composer/src/executor/builder.rs
+++ b/crates/astria-composer/src/executor/builder.rs
@@ -17,6 +17,12 @@ use astria_eyre::eyre::{
 };
 use tokio::sync::watch;
 use tokio_util::sync::CancellationToken;
+use tonic::transport::{
+    Channel,
+    ClientTlsConfig,
+    Endpoint,
+    Uri,
+};
 
 use crate::{
     executor,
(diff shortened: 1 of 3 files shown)


crates/astria-composer/src/executor/builder.rs
```diff
@@ -102,13 +108,12 @@ fn read_signing_key_from_file<P: AsRef<Path>>(path: P) -> eyre::Result<SigningKe
...
-) -> eyre::Result<SequencerServiceClient<tonic::transport::Channel>> {
-    let uri: tonic::transport::Uri = grpc_endpoint
+fn connect_sequencer_grpc(grpc_endpoint: &str) -> eyre::Result<SequencerServiceClient<Channel>> {
+    let uri: Uri = grpc_endpoint
         .parse()
         .wrap_err("failed to parse endpoint as URI")?;
-    Ok(SequencerServiceClient::new(
-        tonic::transport::Endpoint::from(uri).connect_lazy(),
-    ))
+    let endpoint = Endpoint::from(uri.clone())
```
SuperFluffy, line 115: Unclear why the clone here is necessary. Just move the URI in there?

Alternatively remove the `let uri` binding altogether and chain the whole thing:

```suggestion
    let endpoint = grpc_endpoint.parse::<Endpoint>().wrap_err("failed to parse input as gRPC endpoint")?.tls_config(...) // etc
```

joroshiba (approved):

PR #2242, opened 2025-07-30
ci: try default cache in depot

joroshiba:
## Summary
Migrates away from buildjet cache, and updates to latest rust cache handler action. 

## Background
The buildjet cache seems to hit random slowdowns, and we aren't using their servers anymore. According to docs, depot runners default to using [local cache mechanisms](https://depot.dev/docs/github-actions/overview#faster-caching). This attempts using those. 

Testing on this PR show that it performs a little faster than latest runs on mainnet (5s on cached vs 8s on most recent main test run), and hopefully avoids the [occasional slowdown](https://github.com/astriaorg/astria/actions/runs/16631231960) we have seen where tests can take over 30m.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -34,9 +34,7 @@ jobs:
         uses: dtolnay/rust-toolchain@master
         with:
           toolchain: ${{ env.RUSTUP_TOOLCHAIN }}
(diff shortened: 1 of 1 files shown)


joroshiba:

PR #2258, opened 2025-10-02
fix(conductor): don't use last block validator set

joroshiba:
## Summary
Updates Conductor to use the validator set at same height as a given commit.

## Background
We were previously using the previous commit, which worked when validator didn't change quickly enough to affect quorum, and allowed the use the proposer priorities to determine the correct validating node. We no longer use this priority, and quick changes to the validator set have caused the quorum to be unreachable from previously blocks proposer set.

## Changes
- Use commit height instead of previous height for validator set.

## Testing
CI/CD updates + manual testing.

### DIFF
--- a/crates/astria-conductor/src/celestia/block_verifier.rs
+++ b/crates/astria-conductor/src/celestia/block_verifier.rs
@@ -314,7 +314,7 @@ mod tests {
         };
 
         (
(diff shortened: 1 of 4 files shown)


jbowen93 (approved):

jbowen93 (approved):

Fraser999 (approved):

joroshiba:
closed in 7 days.

joroshiba:

